PR #3, opened 2018-05-24
Fix support for --enable-ssl.

jkroonza:
This change enables the explicit use of --enable-ssl in such a way that
package managers such as portage (Gentoo) that explicitly does
--enable-ssl or --disable-ssl will get the results that it's looking
for.

Without this specifying --enable-ssl would end up actually disabling it.

Additionally, if --enable-ssl is specified but the script ends up being
unable to enable ssl it will fail.


ismangil:
We do not use pull requests, please suggest patches on the mailing list. Thank you.

PR #4, opened 2018-07-01
Vphone pjproject base/base rebase

fhsiao:


mingteluu:
Hi Frank,

First, we do not use pull requests. Patches are normally submitted via our
mailing list.

Also, looking at the patch, it seems that there's still a lot of work to be
done before it can be integrated into the trunk:
- The patch is specific for Android only, it will break other platforms.
- There are a lot of redundancies between timer.c and timer_slnk.c, so we
may need to refactor this (i.e. putting the shared code in one file such as
timer.c, and only the different ones in timer_common.c and timer_slnk.c).

But before we proceed further, can you please tell us about the background
of the patch (i.e. why is it necessary, what does it fix/improve, etc)?

Best regards,
Ming

PR #8, opened 2019-01-14
Update comment in configure-android

bokic:
Since `grep 'g++'` command finds both g++ or clang++ just like expected, reflect it in the comment.


ismangil:
We are not using PRs. Please submit to our mailing list.

PR #12, opened 2019-06-19
Update sip_msg.h

GilRichardIntertalk:
Valid Expiry values are between 0 and 2**32-1 [RFC 3261 20.19].  Changing the ivalue of pjsip_generic_int_hdr to pj_int64_t will widen the variable to prevent overflowing and maintain the historical signed nature of the variable.


sauwming:
Fixed in #2209

PR #2285, opened 2020-02-18
Added initial .gitignore

nanangizz:
Please try put the file on your PJSIP root dir on your environment, feel free to add more ignore patterns :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,11 @@ output/
 .vs/
 **/build/**/*.vcproj.*.user
 UpgradeLog.htm
+# VS 2015 stuff
+*.pdb
+*.db
+*.opendb
+samples.log
 
 # GNU stuff
 config.log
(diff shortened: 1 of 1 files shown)


sauwming:
gak bisa di-review?

okay ignore file-nya, tapi koq di tempatku gitignore-nya sendiri gak
di-ignore ya? :D
"Untracked files:

  (use "git add <file>..." to include in what will be committed)


.gitignore

"

nanangizz:
> why can't I review this?

check commit 90d50c298c67d2e3d9f8f574950c94c91fa85b3e & write any note there?

> the ignore file seems okay, but why the gitignore itself is not ignored? 😆 "Untracked files: (use "git add <file>..." to include in what will be committed) .gitignore "

haha.. it will be 'ignored' (tracked) once pushed/merged.

.gitignore
```diff
@@ -2,6 +2,7 @@ pjlib/include/pj/config_site.h
 lib/
 bin/
 output/
+Debug/
```
nanangizz: what about "Release/" or any other build configs?
Or even better if Debug, Release, etc can be set to be under "output".

trengginas: It's actually used for "pjsip-apps\build\Debug" folder which contains some swig_java_pjsua2 output files. Need to change the project properties (swig_java_pjsua2) apparently

PR #2286, opened 2020-02-18
Add API for sending an RTCP Feedback PLI

saghul:


pjmedia/src/pjmedia/vid_stream.c
```diff
@@ -2392,6 +2392,22 @@ PJ_DEF(pj_status_t) pjmedia_vid_stream_send_rtcp_bye(
...
 
+/*
+ * Send RTCP PLI.
+ */
+PJ_DEF(pj_status_t) pjmedia_vid_stream_send_rtcp_pli(
+						pjmedia_vid_stream *stream)
+{
+    PJ_ASSERT_RETURN(stream, PJ_EINVAL);
+
+    if (stream->enc && stream->transport) {
```
nanangizz: as it is supposed to be decoder that sends PLI, shouldn't it be 'stream->dec' instead of 'stream->enc'?

saghul: TBH I copied it from the `pjmedia_vid_stream_send_rtcp_bye` function. Come to think of it, I think we only need to check for stream->transport, since none enc or dec are used in send_rtcp. What do you thing?

saghul:
I'll drop the second commit as per our discussion on the mailing list.

pjmedia/src/pjmedia/vid_stream.c
nanangizz: Just checked send_rtcp(), it seems to use stream->enc for logging purpose, not really sure why, will try to find out later. Anyway, only checking for stream->transport sounds fine. Thanks.

saghul:
@nanangizz Updated, please take a look.

pjmedia/include/pjmedia/vid_stream.h
```diff
@@ -454,6 +454,18 @@ PJ_DECL(pj_status_t) pjmedia_vid_stream_send_rtcp_sdes(
 PJ_DECL(pj_status_t) pjmedia_vid_stream_send_rtcp_bye(
 						pjmedia_vid_stream *stream);
 
+
+/**
+ * Send RTCP PLI for the media stream.
```
sauwming:
```suggestion
 * Send RTCP PLI for the video stream.
```

pjmedia/include/pjmedia/vid_stream.h
```diff
@@ -454,6 +454,18 @@ PJ_DECL(pj_status_t) pjmedia_vid_stream_send_rtcp_sdes(
...
+ *
+ * @param stream	The media stream.
```
sauwming:
```suggestion
 * @param stream	The video stream.
```

pjmedia/include/pjmedia/vid_stream.h
sauwming: 2 API above this (send_rtcp_sdes() and _bye()) also uses media stream, probably because it's copied from audio stream, but the rest already more appropriately and specifically use the words "video stream".

pjmedia/include/pjmedia/vid_stream.h
sauwming: the param, too?

saghul:
Updated again, this time I added a new commit fixing the other parameters, while we are here.

PR #2287, opened 2020-02-19
Transaction module shutdown may leak resources when there is any transaction with pending transport

nanangizz:
Currently when transaction module is shutdown, any transactions with pending transport (e.g: sending in progress) will be terminated and destroyed but its sending counter (transaction group lock ref counter) may never reach zero, so there will be leak due to unfreed tsx group lock and unreleased tsx memory pool.

This is because the sending counter will be decreased only by send_msg_callback() with condition that transaction module is not shutdown. This pull request will change the transaction destroy behavior that sending counter will be decreased when tsx state shifting to terminated, also revert back latest changes in tsx destroy test in pjsip-test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,38 @@
+pjlib/include/pj/config_site.h
+lib/
+bin/
+output/
+
+# VS stuff
+*.ncb
+*.suo
+.vs/
+**/build/**/*.vcproj.*.user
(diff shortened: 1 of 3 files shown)


nanangizz:
> looks okay. gitignore accidentally included?

ya, next time should have cleaned up the workdir somehow before creating new branch.

PR #2297, opened 2020-02-21
Add TLS proto for TLS 1.3

nanangizz:


nanangizz:
Tested using GnuTLS 3.6.12 (self build) on Ubuntu, added checks to make sure TLS 1.3 is included in the build:
```
#if GNUTLS_VERSION_NUMBER >= 0x030605
    if (GNUTLS_TLS_VERSION_MAX>=GNUTLS_TLS1_3 &&
	ssock->param.proto & PJ_SSL_SOCK_PROTO_TLS1_3)
    {
        pj_strcat2(&priority, "+VERS-TLS1.3:");
    }
#endif
```
Unfortunately cannot make it work:
```14:43:01.234                        ..echo test w/ compatible proto: server TLSv1.2+1.3 vs client TLSv1.3
14:43:01.235      ssl0x559e9cc59e70 Priority string: NORMAL:+VERS-TLS1.3:+VERS-TLS1.2:%LATEST_RECORD_VERSION:+COMP-NULL:%SERVER_PRECEDENCE
14:43:01.235        ssl_sock_gtls.c gnutls_priority_set_direct return -50
14:43:01.235      ssl0x559e9cc59e70 Handshake failed in accepting 127.0.0.1:45779: Invalid value or argument (PJ_EINVAL)
14:43:01.235      ssl0x559e9c7fa600 Priority string: NORMAL:+VERS-TLS1.3:%LATEST_RECORD_VERSION:+COMP-NULL
14:43:01.235        ssl_sock_gtls.c gnutls_priority_set_direct return -50
14:43:01.235                   test ...ERROR ssl_on_connect_complete(): [pj_status_t=70004] Invalid value or argument (PJ_EINVAL)
```

Currently our implementation will always set GnuTLS ["NORMAL"](https://gnutls.org/manual/html_node/Priority-Strings.html) priority string, which seems include all secure ciphers (which should implicitly also include their protocol versions), so good news is TLS 1.3 should be implicitly included (yet cannot set to only use TLS 1.3).

sauwming:
there seems to be only one or two who uses gnutls backend, so probably okay just to leave it as it is.
But just wondering, what does the error -50 mean?

nanangizz:
perhaps this:
`#define GNUTLS_E_INVALID_REQUEST -50`

PR #2301, opened 2020-02-24
Limit log printing if video conference is connected to a sink port that's not running

sauwming:
Re: [r6138](https://github.com/pjsip/pjproject/commit/d5ec2f45920a852cd28ea44822d9320bc3caec84), but that revision was for the encoding direction.

### DIFF
--- a/pjmedia/src/pjmedia/vid_conf.c
+++ b/pjmedia/src/pjmedia/vid_conf.c
@@ -57,8 +57,6 @@ struct pjmedia_vid_conf
     pj_mutex_t		 *mutex;	/**< Conference mutex.		    */
     struct vconf_port	**ports;	/**< Array of ports.		    */
     pjmedia_clock	 *clock;	/**< Clock.			    */
-    pj_status_t		 *last_err;	/**< Last error status.		    */
-    unsigned		 *last_err_cnt;	/**< Last error count.		    */
 };
 
 
@@ -104,6 +102,9 @@ typedef struct vconf_port
     pj_pool_t	       **render_pool;	/**< Array of pool for render state */
     render_state       **render_states;	/**< Array of render_state (one for
 					     each transmitter).		    */
+
+    pj_status_t		  last_err;	/**< Last error status.		    */
+    unsigned		  last_err_cnt;	/**< Last error count.		    */
 } vconf_port;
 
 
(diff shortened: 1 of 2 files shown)


pjmedia/src/pjmedia/vid_conf.c
```diff
@@ -54,6 +57,8 @@ struct pjmedia_vid_conf
     pj_mutex_t		 *mutex;	/**< Conference mutex.		    */
     struct vconf_port	**ports;	/**< Array of ports.		    */
     pjmedia_clock	 *clock;	/**< Clock.			    */
+    pj_status_t		 *last_err;	/**< Last error status.		    */
+    unsigned		 *last_err_cnt;	/**< Last error count.		    */
```
nanangizz: putting them as part of vid conf port?

pjmedia/src/pjmedia/vid_conf.c
```diff
@@ -771,10 +786,22 @@ static void on_clock_tick(const pj_timestamp *now, void *user_data)
...
 	}
 	status = pjmedia_port_put_frame(sink->port, &frame);
-	if (status != PJ_SUCCESS) {
-	    PJ_PERROR(5, (THIS_FILE, status,
-			  "Failed to put frame to port [%s]!",
-			  sink->port->info.name.ptr));
+	if (got_frame && status != PJ_SUCCESS) {
+	    if (vid_conf->last_err[i] != status ||
+	        vid_conf->last_err_cnt[i] > MAX_ERR_COUNT ||
+	        status != PJ_EINVALIDOP)
```
nanangizz: - exclude this condition 'status != PJ_EINVALIDOP'?
- also print the error counter

pjsip/src/pjsua-lib/pjsua_vid.c
```diff
@@ -994,6 +994,8 @@ static pj_status_t setup_vid_capture(pjsua_call_media *call_med)
     /* Done */
     inc_vid_win(wid);
     call_med->strm.v.cap_win_id = wid;
+    PJ_LOG(4,(THIS_FILE, "Call %d: video capture set up with dev %d, wid=%d",
+	      call_med->call->index, call_med->strm.v.cap_dev, wid));
```
nanangizz: better put the initial patch in a separate commit

sauwming: #2303

nanangizz: Meant to be separate commits in the same PR (not sure how to do that, need to find out later perhaps). No probs with a new PR anyway.

pjmedia/src/pjmedia/vid_conf.c
nanangizz: what about printing the error counter (if it is more than 1)? should be helpful to identify 'persistent' errors?

sauwming: currently the error counter will always be reset.
unless we change it to:
`sink->last_err_cnt % MAX_ERR_COUNT == 0`

there are a couple of drawbacks:
- require division
- can overflow

or perhaps we don't need the actual value of the counter itself. then we can simply print:
`failed again to ....`

nanangizz: For now, we need to identify any persistent/recurring error, so being reset or overflow is not really a problem I guess, e.g: 50 same errors should already present it.

Using 'again' sounds like just showing a second occurrence (after few minutes), it is fine as long as we can remember what it really means (like next month :), as generally I prefer to remember less things, I'd vote for simple logging the counter number as is, IMO any not small number (e.g: > 10) will do.

Issue #2311, opened 2020-02-27
PJSIP undefine standard UNIX macros s_addr and s6_addr


aberaud:
In `sock.h` and `compat/socket.h`,  PJSIP undefines crucial macros from the UNIX standard library:

```c
#undef s_addr
#undef s6_addr
#undef sin_zero
```

This breaks any program using PJSIP and making use of these standard-library macros/fields.

For instance it makes impossible to use PJSIP along with ASIO, because ASIO makes use of these standard-library macros:

```
/include/asio/ip/impl/address_v6.ipp:56:16: error: no member named 's6_addr' in 'in6_addr'
  memcpy(addr_.s6_addr, bytes.data(), 16);
```
PJSIP, like any proper C/C++ program or library, should never undefine or redefine UNIX or standard library macros and symbols.

aberaud:
A solution might be to define a pjsip version of the macro only if the macro is undefined ?

sauwming:
PJSIP library itself seems to build just fine if I remove those macro undefinitions.
However, if in my app, I include headers which define those macros, such as `<netinet/in.h>` then it will fail to compile with an error:
```
../include/pj/sock.h:549:17: error: expected ';' at end of declaration list
    pj_uint8_t  s6_addr[16];   /**< 8-bit array */
                ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/netinet6/in6.h:159:28: note: 
      expanded from macro 's6_addr'
#define s6_addr   __u6_addr.__u6_addr8
```

The reason we undefine those macros is because we use it do declare our IPv4/6 address structure, such as:
```
/* Must undefine s_addr because of pj_in_addr below */
//#undef s_addr
typedef struct pj_in_addr {
    pj_uint32_t	s_addr;		/**< The 32bit IP address.	    */
...
#undef s6_addr
typedef union pj_in6_addr {
    pj_uint8_t  s6_addr[16];   /**< 8-bit array */
```

So the solution seems to be to rename the struct member names to something else. But then, I'm not sure why we insisted to use such conflicting names in the first place. Perhaps @nanangizz can comment on this issue?

sauwming:
[...]

PR #2313, opened 2020-02-28
Change struct member names s_addr and s6_addr since it may conflict with predefined macros

sauwming:
Re: #2311

### DIFF
--- a/pjlib/include/pj/compat/socket.h
+++ b/pjlib/include/pj/compat/socket.h
@@ -61,7 +61,7 @@
  */
 #if defined(_MSC_VER) && defined(PJ_HAS_IPV6) && PJ_HAS_IPV6!=0
 #   ifndef s_addr
-#	define s_addr  S_un.S_addr
+#	define s_addr  S_un.addr
 #   endif
 
 #   if !defined(IPPROTO_IPV6) && (_WIN32_WINNT == 0x0500)
@@ -179,10 +179,12 @@
 
 /*
  * And undefine these..
+ * Note: unfortunately, this may cause build failure to anyone who
+ * uses these standard macros.
  */
-#undef s_addr
-#undef s6_addr
-#undef sin_zero
+// #undef s_addr
+// #undef s6_addr
+// #undef sin_zero
 
 /*
  * This will finally be obsoleted, since it should be declared in
--- a/pjlib/include/pj/compat/socket.h
+++ b/pjlib/include/pj/compat/socket.h
@@ -61,7 +61,7 @@
  */
 #if defined(_MSC_VER) && defined(PJ_HAS_IPV6) && PJ_HAS_IPV6!=0
 #   ifndef s_addr
-#	define s_addr  S_un.addr
+#	define s_addr  S_un.S_addr
 #   endif
 
 #   if !defined(IPPROTO_IPV6) && (_WIN32_WINNT == 0x0500)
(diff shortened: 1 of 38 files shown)


pjlib/include/pj/sock.h
```diff
@@ -484,15 +484,26 @@ typedef enum pj_socket_sd_type
...
+ */
+//#undef s_addr
+
+/* This macro is used only for backward compatibility in apps which
+ * access pj_in_addr.addr . If it causes any issue, comment this
+ * and access pj_in_addr.addr directly.
+ */
+#ifndef s_addr
+    #define s_addr addr
+#endif
```
nanangizz: mention the issue link/number somewhere?

pjlib/include/pj/sock.h
```diff
@@ -535,16 +546,27 @@ struct pj_sockaddr_in
...
+//#undef s6_addr
 
-#undef s6_addr
+/* This macro is used only for backward compatibility in apps which
+ * access pj_in_addr.s6_addr . If it causes any issue, comment this
+ * and access pj_in6_addr.addr directly.
+ */
+#ifndef s6_addr
+    #define s6_addr addr
+#endif
```
nanangizz: also this, link.

sauwming:
The important changes are the ones in `pjlib/include/pj/sock.h` and `pjlib/include/pj/compat/socket.h`

Should I change the field name `sin_zero` as well? Unlike `s*addr`, `sin_zero` undefinition can only be found in `compat`, but not in `sock.h`, and it builds fine here on Mac without changing it. Here's the definition of `sin_zero`:
`#define sin_zero __pad /* for BSD UNIX comp`

If it's needed, what should I change it to?

pjlib/include/pj/compat/socket.h
```diff
@@ -61,7 +61,7 @@
  */
 #if defined(_MSC_VER) && defined(PJ_HAS_IPV6) && PJ_HAS_IPV6!=0
 #   ifndef s_addr
-#	define s_addr  S_un.S_addr
+#	define s_addr  S_un.addr
```
nanangizz: this should not be changed?

pjlib/include/pj/compat/socket.h
```diff
@@ -179,10 +179,12 @@
 
 /*
  * And undefine these..
+ * Note: unfortunately, this may cause build failure to anyone who
+ * uses these standard macros.
```
nanangizz: link to issue

pjlib/include/pj/compat/socket.h
sauwming: oh no, that's the evil of using batch replace. Need to recheck everything.

nanangizz:
re: renaming field sin_zero, there seems to be no problem on this, so I think it's better to let's just leave it.

sauwming:
Note: This still causes build failure on Windows:
```
`1>Compiling...
1>ssl_sock_ossl.c
1>c:\program files (x86)\microsoft visual studio 8\vc\platformsdk\include\winsock2.h(403) : error C2079: 'sin_addr' uses undefined struct 'in_addr'
1>c:\program files (x86)\microsoft visual studio 8\vc\platformsdk\include\ws2tcpip.h(29) : error C2079: 'imr_multiaddr' uses undefined struct 'in_addr'
```

sauwming:
Closing this PR as we adopt a different approach in #2423.

PR #2317, opened 2020-03-02
Indicate that video format change is from decoding stream.

trengginas:
Currently on_call_media_event() might be called multiple times from various video components although it is the same event.
This patch will add information to indicate that the event publisher is stream decoder. This is useful for app to identify the correct event to watch.

### DIFF
--- a/pjmedia/include/pjmedia/event.h
+++ b/pjmedia/include/pjmedia/event.h
@@ -125,6 +125,10 @@ typedef struct pjmedia_event_fmt_changed_data
 
     /** The new media format. */
     pjmedia_format	new_fmt;
+
+    /** Event publisher is decoding stream. */
+    pj_bool_t		is_dec_stream;
+
 } pjmedia_event_fmt_changed_data;
 
 /**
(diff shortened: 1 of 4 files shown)


nanangizz:
not so sure we can go with this approach, it may solve an issue, but what if app wants to watch for the event from other video comp? also in the future, there may be more video components broadcasting media events that pjsua2 app wants to watch and I don't think we can add flag 'is_<vid_com>' as many.

trengginas:
We could introduce new enum to represent the event publisher type, e.g: `pjmedia_event_pub_type`. 
This will lead to API changes, e.g: `pjmedia_event_init2()` or `pjmedia_event_publish2()`.
Another approach is to use new field to indicate that the event is actually republished event.

nanangizz:
Yes, adding publisher/source type info can be useful, but I don't think it is really needed for now, too much with API change.

IIRC, FMT_CHANGED is originally from video codec, so stream & renderer may both republish it. If app sees that both events have the field set, it can be confusing for app.

An idea, is it possible for pjsua to check duplicated format change events? e.g: simply comparing old vs new format, if yes, perhaps pjsua can filter such duplicated events.

trengginas:
Okay then, app just needs to compare the format changed of the events. If there's no change, app can ignore and handle it as duplicate.

nanangizz:
well, actually I was thinking to help app by doing the filter in pjsua (lib, not app), but leaving it to app should be okay for now.

PR #2324, opened 2020-03-03
SDL full-screen mode fixes

saghul:
This PR fixes 2 main issues with the current implementation of full-screen:

* It changes the videomode, which users most likely don't want
* It doesn't work properly if the video is resized while in full-screen

### DIFF
--- a/pjmedia/src/pjmedia-videodev/sdl_dev.c
+++ b/pjmedia/src/pjmedia-videodev/sdl_dev.c
@@ -704,7 +704,7 @@ static pj_status_t sdl_create_window(struct sdl_stream *strm,
         if ((strm->param.flags & PJMEDIA_VID_DEV_CAP_OUTPUT_FULLSCREEN) &&
             strm->param.window_fullscreen)
         {
-            flags |= SDL_WINDOW_FULLSCREEN;
+            flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
         }
 
 #if PJMEDIA_VIDEO_DEV_SDL_HAS_OPENGL
@@ -1111,7 +1111,7 @@ static pj_status_t get_cap(void *data)
 	return PJ_SUCCESS;
     } else if (cap == PJMEDIA_VID_DEV_CAP_OUTPUT_FULLSCREEN) {
 	Uint32 flag = SDL_GetWindowFlags(strm->window);
-	*((pj_bool_t *)pval) = (flag & SDL_WINDOW_FULLSCREEN)? PJ_TRUE: PJ_FALSE;
+	*((pj_bool_t *)pval) = (flag & SDL_WINDOW_FULLSCREEN_DESKTOP)? PJ_TRUE: PJ_FALSE;
 	return PJ_SUCCESS;
     }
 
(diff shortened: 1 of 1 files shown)


sauwming:
What if the user wants the real fullscreen mode? Won't this remove the feature?

While for the fake fullscreen, I believe it can still be done on the application level instead, i.e. as the SDL doc says, by resizing the video window to the desktop size.

And as for the resizing issue, I think it's the expected behavior that resizing can't be done in real fullscreen mode. But if you use fake fullscreen mode, it shouldn't be an issue.

saghul:
Hey @sauwming, thanks for taking a look!

> What if the user wants the real fullscreen mode? Won't this remove the feature?

No user wants "real" fullscreen mode. Nobody expects their system resolution changed just because you want see the video take all the screen space available. That is used mostly in games, which revert the mode at exit, but it's not an expectation of regular desktop applications.

With this fixes the full-screen functionality will behave the same, but this time properly. I submitted the patch which implemented the full-screen API in the first place, but I've realized it had problems which this PR solves, sorry I didn't catch them earlier!

> While for the fake fullscreen, I believe it can still be done on the application level instead, i.e. as the SDL doc says, by resizing the video window to the desktop size.

Sure, you can do that, but that's what SDL_WINDOW_FULLSCREEN_DESKTOP already does, in addition to taking care of the border, which is done automatically.

> And as for the resizing issue, I think it's the expected behavior that resizing can't be done in real fullscreen mode. But if you use fake fullscreen mode, it shouldn't be an issue.

It's still an issue, I tested it and that's why I'm submutting this fix :-)

nanangizz:
What if we support both modes? With the same capability `PJMEDIA_VID_DEV_CAP_OUTPUT_FULLSCREEN`, the mode is configurable via the existing param `param.window_fullscreen`, just need to change the type from bool to unsigned and add enum flags for its valid value def (e.g: to maintain backward comp, `PJMEDIA_VID_DEV_OUT_WINDOWED=0, PJMEDIA_VID_DEV_OUT_FULLSCREEN=1, PJMEDIA_VID_DEV_OUT_FULLSCREEN_DESKTOP=2`).

saghul:
@nanangizz I thought of it, but I see no reason why anyone would want the videomode changed. Do you?

nanangizz:
Well, perhaps it is not so popular for desktop SIP apps, but some embedded SIP apps may prefer it (at least it could save some CPU from avoiding video resize). Providing alternatives is generally better, it is simple to implement and low overhead.

saghul:
Ok, makes sense I guess. I'll try to add that in the upcoming week.

PS: Please excuse my tone here, I just can't get over the fact that something which didn't exist (literally) a month ago cannot be changed now.

saghul:
Just a heads up. I have not forgotten about this, but attention has been shifted due to the COVID-19 situation. I'll get back to this at some point.

nanangizz:
Hi @saghul . Tried to implement it here, turned out the change affects more files than I expected, e.g: pjsua app, SWIG files symbol.i, new method in PJSUA2. Anyway, here is the patch (based on jitsi/fix-fullscreen branch): [fullscreen_modes.txt](https://github.com/pjsip/pjproject/files/5595837/fullscreen_modes.txt). Please check and git-push it so it can be reviewed here, thanks.

Issue #2316, opened 2020-02-28
pjsua_call.c  .Incoming Request msg INVITE/cseq=2477 (rdata0x10884bb30) Assertion failed: (pjsua_var.acc_cnt!=0), function pjsua_acc_find_for_incoming, file ../src/pjsua-lib/pjsua_acc.c, line 2990.


Rushang08:
**Describe the bug**
[iOS]Application getting a crash when PJSIP receives a call while disconnecting from asterisks.
 
**To Reproduce**

1. Put disconnect asterisks method on did enter background. 
2. Goes to the background and do call at the time. 

I have a 5-year-old PJSIP version and also I am not able to update new pjsip to my existing project.

trengginas:
Try updating the library. If you still having issue, provide us with the information required from the issue template. To update the library, refer to #2315

PR #2328, opened 2020-03-04
Add support for wildcard certificates.

andreas-wehrmann:
Believe it or not, but some (huge) providers use these and we need to accept them if we want to be able to talk to them with server certificate verification turned on.
One of these "huge" providers sits in Germany and there's a big "T" in its name...

### DIFF
--- a/pjsip/include/pjsip/sip_config.h
+++ b/pjsip/include/pjsip/sip_config.h
@@ -382,6 +382,22 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 #   define PJSIP_DONT_SWITCH_TO_TCP	0
 #endif
 
+/**
+ * Even though RFC 5922 (section 7.2) forbids the use of wildcard certificates,
+ * some providers use them anyway and they won't or
+ * (by the way their systems are designed) cannot change to non-wildcard certificates.
+ *
+ * This option will specify whether wildcard certificates are accepted.
+ *
+ * If this option is disabled, you can still connect to endpoints
+ * that present wildcard certificates, provided that server verification is turned off.
+ *
+ * Default is 0 (no).
+ */
+#ifndef PJSIP_ACCEPT_TLS_WILDCARD_CERTIFICATE
+#   define PJSIP_ACCEPT_TLS_WILDCARD_CERTIFICATE	0
+#endif
+
 
 /**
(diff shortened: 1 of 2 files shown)


nanangizz:
For this scenario, actually we recommend to handle it in app. First, please check about TLS verification here: #2316 

Basically PJSIP TLS transport allows app to fine tune the TLS verification in app level via PJSUA callback `on_transport_state()`. To do this, server verification should be disabled, so TLS transport won't automatically disconnect the transport due to verification error. Then when a transport is connected (`on_transport_state()` is invoked with `state==PJSIP_TP_STATE_CONNECTED`), application can inspect the initial verification result done by the library (sample code [here](https://github.com/pjsip/pjproject/blob/master/pjsip-apps/src/pjsua/pjsua_app.c#L836)). If `PJ_SSL_CERT_EIDENTITY_NOT_MATCH` is the only verification error, app can check further for wildcard cert. If app eventually decides to abort the connection, e.g: cert identity still does not match after wildcard check or verification error is not (only) `PJ_SSL_CERT_EIDENTITY_NOT_MATCH`, it can simply shutdown the transport.

andreas-wehrmann:
I didn't realize I could implement verification that way.

I've now implemented the identity check via PJSUA's on_transport_state() callback,
however I don't know how I can figure out whether I actually need to do verification
for the specific transport I'm getting the callback for.
In my app, I might have several transports, and I want this specific check routine to run only for specific ones.

What I need is some kind of 'verify_server' flag that I can set when creating the transport
and then the ability to query for that inside the on_transport_state() callback.

Maybe I overlooked something, but is there some way to set user data on a transport via PJSUA?
Getting the transport ID in the callback that PJSUA assigned during transport creation would also be okay.

nanangizz:
There is no straightforward way to get PJSUA transport listener instance ID from a PJSIP transport client instance (note that they are actually different instances, and a PJSUA app normally does not create PJSIP transport client instance), but for whatever your need, perhaps you can make use of `transport->remote_name` or perhaps `transport->factory`.

Anyway, the transport selection/filtering sounds like a different issue, so I guess we can close the TLS cert wildcard check discussion now.

andreas-wehrmann:
Thanks for your help and the insight.
For the moment I'll leave my patch and comment out the callback until there is a way to filter/get the correct settings for the transport.

PR #2329, opened 2020-03-05
Add more CI tests on Ubuntu and Mac

sauwming:
Continuation of #2308 .

Disabled test:
* vid_codec_test because there's no SDL renderer

Expected test durations:
- pjlib & pjlib-util ~5m
- pjmedia 2m
- pjnath 44m
- pjsip 26m
- pjsua ~15m

The tests, in particular pjlib-util-test, pjsip-test, and pjsua-test, don't seem to have a 100% success rate.


pjlib/include/pj/config_site_test_full.h
```diff
@@ -0,0 +1,12 @@
...
+#define PJMEDIA_CODEC_L16_HAS_8KHZ_MONO 1
+#define PJMEDIA_CODEC_L16_HAS_8KHZ_STEREO 1
+#define PJMEDIA_CODEC_L16_HAS_16KHZ_MONO 1
+#define PJMEDIA_CODEC_L16_HAS_16KHZ_STEREO 1
+#define PJMEDIA_CODEC_L16_HAS_48KHZ_MONO 1
+#define PJMEDIA_CODEC_L16_HAS_48KHZ_STEREO 1
+#define PJMEDIA_HAS_G7221_CODEC 1
+#define PJMEDIA_HAS_G722_CODEC 1
+#define PJ_EXCLUDE_BENCHMARK_TESTS 1
+#define PJ_CLI_MAX_CHOICE_VAL 64
```
nanangizz: was thinking to put 64 as PJ_CLI_MAX_CHOICE_VAL default? and I guess it is okay to make it in this same PR.

.github/workflows/ccpp.yml
```diff
@@ -1,28 +1,311 @@
 name: C/C++ CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
-  build:
+  build-ubuntu-default:
```
nanangizz: `default` is supposed to be about checking pure lib source distribution with plain configure & make. So can we remove '-fPIC' and swig here?

also, adding simple description/comment for each job can be useful for future maintenance, e.g:
```
jobs:
  build-ubuntu-default:
  # checking pure lib source distribution with plain configure & make
  ...
```

.github/workflows/ccpp.yml
```diff
@@ -1,28 +1,311 @@
...
+  build-ubuntu-video-openh264-1:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: install dependencies
+      run: sudo apt-get install -y swig nasm sip-tester libvpx-dev libopencore-amrnb-dev
+    - name: get openh264
+      run: git clone --single-branch --branch openh264v2.1.0 https://github.com/cisco/openh264.git
+    - name: build openh264
+      run: cd openh264 && make && sudo make install && sudo rm /usr/local/lib/libopenh264*.so*
```
nanangizz: why libopenh264.so is removed here? better add note/comment about it?
if only static lib is wanted, perhaps there is configure option for that, e.g: usually kind of `--disabled-shared`?

pjlib/include/pj/config_site_test_video.h
```diff
@@ -0,0 +1,5 @@
+#define PJMEDIA_HAS_VIDEO 1
+#define PJMEDIA_SRTP_HAS_DTLS 1
+#define PJMEDIA_HAS_G7221_CODEC 1
+#define PJ_EXCLUDE_BENCHMARK_TESTS 1
+#define PJ_CLI_MAX_CHOICE_VAL 64
```
nanangizz: perhaps we can replace this file with config_site_test_full.h+PJMEDIA_HAS_VIDEO?

PR #2336, opened 2020-03-09
Add thread locking callback when using OpenSSL.

trengginas:
It is reported that while using PJSIP with 8 worker threads and 1.0.2 openssl with FIPS enabled,
during runtime FIPS Random is stalled with a error message 'Selftest failed'.
```
SSL ECDH initialized (automatic), faster PFS ciphers enabled
SSL SSL_ERROR_SSL (Handshake): Level: 0 err: <755437702> <FIPS routines-FIPS_drbg_generate-selftest failed> len: 0
SSL SSL_ERROR_SSL (Handshake): Level: 1 err: <336285997> <SSL routines-ssl_get_new_session-ssl session id callback failed> len: 0
Handshake failed in accepting 136.243.39.78:53371: Unknown OpenSSL error 134
```
Once this happens, no further openssl function will work due to the protection in FIPS container.
Based on [this](https://www.openssl.org/docs/man1.0.2/man3/CRYPTO_set_locking_callback.html)
on multiple thread application, locking function need to be set otherwise random crash will happen. 

Notes:
- There is dynamic lock mechanism which is not included on this patch. From the docs:
```
Also, dynamic locks are currently not used internally by OpenSSL, but may do so in the future.
```

### DIFF
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -1024,6 +1024,15 @@
 #  define PJ_SSL_SOCK_MAX_CURVES   32
 #endif
 
+/**
+ * Use SSL backend thread locking.
+ *
+ * Default: 0 (No)
+ */
+#ifndef PJ_SSL_SOCK_USE_THREAD_LOCK
+#  define PJ_SSL_SOCK_USE_THREAD_LOCK   0
+#endif
+
 
 /**
  * Disable WSAECONNRESET error for UDP sockets on Win32 platforms. See
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -1025,12 +1025,15 @@
 #endif
 
 /**
- * Use SSL backend thread locking.
+ * Use OpenSSL thread locking callback. This is only applicable for OpenSSL
+ * version prior to 1.1.0
  *
- * Default: 0 (No)
+ * Default: 1 (enabled)
  */
-#ifndef PJ_SSL_SOCK_USE_THREAD_LOCK
-#  define PJ_SSL_SOCK_USE_THREAD_LOCK   0
+#ifndef PJ_SSL_SOCK_OSSL_USE_THREAD_CB
+#   define PJ_SSL_SOCK_OSSL_USE_THREAD_CB   1
+#else
+#   define PJ_SSL_SOCK_OSSL_USE_THREAD_CB   0
 #endif
 
 
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -504,8 +504,34 @@ static void ossl_lock(int mode, int id, const char *file, int line)
     }
 }
 
+static void release_thread_cb()
+{
+    unsigned i = 0;
+
+#if     OPENSSL_VERSION_NUMBER >= 0x10000000
+    CRYPTO_THREADID_set_callback(NULL);
+#else
+    CRYPTO_set_id_callback(NULL);
+#endif
+    CRYPTO_set_locking_callback(NULL);
+
+    for (; i < ossl_num_locks; ++i) {
+        if (ossl_locks[i]) {
+            pj_lock_destroy(ossl_locks[i]);
+            ossl_locks[i] = NULL;
+        }
+    }
+    if (lock_pool) {
+        pj_pool_release(lock_pool);
+        lock_pool = NULL;
+    }
+    pj_caching_pool_destroy(&cp);
+}
+
 static pj_status_t init_ossl_lock()
 {
+    pj_status_t status = PJ_SUCCESS;
+
     pj_caching_pool_init(&cp, NULL, 0);
 
     lock_pool = pj_pool_create(&cp.factory,
@@ -533,12 +559,16 @@ static pj_status_t init_ossl_lock()
 #else
         CRYPTO_set_id_callback(ossl_thread_id);
 #endif
-
         CRYPTO_set_locking_callback(ossl_lock);
+        status = pj_atexit(&release_thread_cb);
+        if (status != PJ_SUCCESS) {
+            PJ_PERROR(1, (THIS_FILE, status,
+                      "Cannot set OpenSSL thread callback unrelease method"));
+        }
     } else {
-        return PJ_ENOMEM;
+        status = PJ_ENOMEM;
     }
-    return PJ_SUCCESS;
+    return status;
 }
 
 #endif
@@ -686,41 +716,10 @@ static pj_status_t init_openssl()
     return status;
 }
 
-
 /* Shutdown OpenSSL */
-static void shutdown_openssl(pj_ssl_sock_t *ssock)
+static void shutdown_openssl()
 {
-    PJ_UNUSED_ARG(ssock);
-#if defined(PJ_SSL_SOCK_OSSL_USE_THREAD_CB) && \
-    PJ_SSL_SOCK_OSSL_USE_THREAD_CB != 0 && OPENSSL_VERSION_NUMBER < 0x10100000L
-
-    if (openssl_init_count) {
-        unsigned i = 0;
-
-#if     OPENSSL_VERSION_NUMBER >= 0x10000000
-        CRYPTO_THREADID_set_callback(NULL);
-#else
-        CRYPTO_set_id_callback(NULL);
-#endif
-        CRYPTO_set_locking_callback(NULL);
-
-        for (; i < ossl_num_locks; ++i) {
-            if (ossl_locks[i]) {
-                pj_lock_destroy(ossl_locks[i]);
-                ossl_locks[i] = NULL;
-            }
-        }
-        if (lock_pool) {
-            pj_pool_release(lock_pool);
-            lock_pool = NULL;
-        }
-        pj_caching_pool_destroy(&cp);
-
-        openssl_init_count = 0;
-    }
-#endif
 }
-
 
 /* SSL password callback. */
 static int password_cb(char *buf, int num, int rwflag, void *user_data)
@@ -1333,7 +1332,7 @@ static void ssl_destroy(pj_ssl_sock_t *ssock)
     /* Potentially shutdown OpenSSL library if this is the last
      * context exists.
      */
-    shutdown_openssl(ssock);
+    shutdown_openssl();
 }
 
 
@@ -1373,8 +1372,8 @@ static void ssl_reset_sock_state(pj_ssl_sock_t *ssock)
 static void ssl_ciphers_populate()
 {
     if (ssl_cipher_num == 0 || ssl_curves_num == 0) {
-	init_openssl(NULL);
-	shutdown_openssl(NULL);
+	init_openssl();
+	shutdown_openssl();
     }
 }
 
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -455,11 +455,96 @@ static int openssl_init_count;
 /* OpenSSL application data index */
 static int sslsock_idx;
 
+#if defined(PJ_SSL_SOCK_USE_THREAD_LOCK) && PJ_SSL_SOCK_USE_THREAD_LOCK != 0 \
+    && OPENSSL_VERSION_NUMBER < 0x10100000
+
+/* OpenSSL thread lock initialization counter. */
+static int openssl_lock_init_count;
+
+/* OpenSSL locking list. */
+pj_lock_t **ossl_locks;
+
+/* OpenSSL number locks. */
+unsigned ossl_num_locks;
+
+#if     OPENSSL_VERSION_NUMBER >= 0x10000000
+static void ossl_set_thread_id(CRYPTO_THREADID *id)
+{
+    CRYPTO_THREADID_set_numeric(id,
+                     (unsigned long)pj_thread_get_os_handle(pj_thread_this()));
+}
+
+#else
+
+static unsigned long ossl_thread_id(void)
+{
+    return ((unsigned long)pj_thread_get_os_handle(pj_thread_this()));
+}
+#endif
+
+static void ossl_lock(int mode, int id, const char *file, int line)
+{
+    PJ_UNUSED_ARG(file);
+    PJ_UNUSED_ARG(line);
+    if (mode & CRYPTO_LOCK) {
+        if (ossl_locks[id]) {
+            //PJ_LOG(6, (THIS_FILE, "Lock File (%s) Line(%d)", file, line));
+            pj_lock_acquire(ossl_locks[id]);
+        }
+    } else {
+        if (ossl_locks[id]) {
+            //PJ_LOG(6, (THIS_FILE, "Unlock File (%s) Line(%d)", file, line));
+            pj_lock_release(ossl_locks[id]);
+        }
+    }
+}
+
+static pj_status_t init_ossl_lock(pj_pool_t* pool) {    
+    if (openssl_lock_init_count == 0) {        
+        ossl_num_locks = CRYPTO_num_locks();
+        ossl_locks = (pj_lock_t **)pj_pool_calloc(pool,
+                                                  ossl_num_locks,
+                                                  sizeof(pj_lock_t*));
+
+        if (ossl_locks) {
+            unsigned i = 0;
+            for (; i < ossl_num_locks; ++i) {
+                pj_lock_create_simple_mutex(pool, "ossl_lock%p",
+                                            &ossl_locks[i]);
+            }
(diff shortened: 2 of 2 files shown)


pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -455,12 +455,97 @@ static int openssl_init_count;
 /* OpenSSL application data index */
 static int sslsock_idx;
 
+#if defined(PJ_SSL_SOCK_USE_THREAD_LOCK) && PJ_SSL_SOCK_USE_THREAD_LOCK != 0 \
+    && OPENSSL_VERSION_NUMBER < 0x10100000
```
nanangizz: as mentioned above, IMO better check OpenSSL version number check in the config.h.

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -455,12 +455,97 @@ static int openssl_init_count;
...
+    }
+    openssl_lock_init_count++;
+    return PJ_SUCCESS;
+}
+
+#endif
 
 /* Initialize OpenSSL */
-static pj_status_t init_openssl(void)
+static pj_status_t init_openssl(pj_ssl_sock_t *ssock)
```
nanangizz: - init_openssl() context is supposed to be for global initialization, so I don't think the SSL socket param is appropriate here.
- is it safe to use SSL socket pool for allocating global var? i.e: if SSL socket is destroyed, won't the global var memory slot be released too?
  an idea, perhaps just use OpenSSL API to allocate memory (can't remember the func name)? and don't forget to release it later.

pjlib/include/pj/config.h
```diff
@@ -1024,6 +1024,15 @@
...
 #endif
 
+/**
+ * Use SSL backend thread locking.
+ *
+ * Default: 0 (No)
+ */
+#ifndef PJ_SSL_SOCK_USE_THREAD_LOCK
+#  define PJ_SSL_SOCK_USE_THREAD_LOCK   0
+#endif
```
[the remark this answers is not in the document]
trengginas: > I think this is openssl specific, so description and naming should specifically mention about it? Or other supported backends (gnutls, etc) have similar mechanism?

gnutls has different mechanism : https://gnutls.org/manual/html_node/Thread-safety.html

> also, perhaps better to enable it by default if openssl version is 1.0.x?

Should be check against 1.1.0. The thread callback is supported for version < 1.1.0

pjlib/src/pj/ssl_sock_ossl.c
trengginas: ok

pjlib/src/pj/ssl_sock_ossl.c
trengginas: > * init_openssl() context is supposed to be for global initialization, so I don't think the SSL socket param is appropriate here.
There are 2 places called init_openssl():
* ```ssl_create()``` which accepting socket param
* ```ssl_ciphers_populate()```
> * is it safe to use SSL socket pool for allocating global var? i.e: if SSL socket is destroyed, won't the global var memory slot be released too?
If we intend to use pj_lock_t, then pj_pool_t is required to create them. That is why we use ```openssl_lock_init_count``` to count the number of times the callback init is called. We can only release the callback after the counter set to 0. 
>   an idea, perhaps just use OpenSSL API to allocate memory (can't remember the func name)? and don't forget to release it later.
If we intend to use pj_lock_t for locking, then pj_pool_t is required.

pjlib/include/pj/config.h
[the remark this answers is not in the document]
nanangizz: ok, so please rename, and update docs.

re: 1.1.0, ok.

pjlib/src/pj/ssl_sock_ossl.c
nanangizz: > > * init_openssl() context is supposed to be for global initialization, so I don't think the SSL socket param is appropriate here.
> >   There are 2 places called init_openssl():
> 
> * `ssl_create()` which accepting socket param
> * `ssl_ciphers_populate()`
> 

ideally it is PJSIP endpoint init that calls init_openssl(), but it is not necessary for now, so inserting it in ssl_create() is sufficient.

so, still init_openssl() is for global context, not SSL socket specific, don't put the SSL socket as param.

> > * is it safe to use SSL socket pool for allocating global var? i.e: if SSL socket is destroyed, won't the global var memory slot be released too?
>
> If we intend to use pj_lock_t, then pj_pool_t is required to create them. That is why we use `openssl_lock_init_count` to count the number of times the callback init is called. We can only release the callback after the counter set to 0.

if the SSL socket is destroyed, its pool will be released too, right? then what will happen with the pj_lock_t (that may still be used by other any SSL sockets)? that's why it is not safe.

>   an idea, perhaps just use OpenSSL API to allocate memory (can't remember the func name)? and don't forget to release it later.
>  >  If we intend to use pj_lock_t for locking, then pj_pool_t is required.

ya that is okay, but don't use 'short lived' pool such as SSL socket's.

pjlib/src/pj/ssl_sock_ossl.c
trengginas: version number check shouldn't be on config.h since it needed to include ```opensslv.h``` and other version checking (existing ones) is done on ssl_sock_ossl.c

pjlib/src/pj/ssl_sock_ossl.c
trengginas: > > > * init_openssl() context is supposed to be for global initialization, so I don't think the SSL socket param is appropriate here.
> > >   There are 2 places called init_openssl():
> > 
> > 
> > 
> > * `ssl_create()` which accepting socket param
> > * `ssl_ciphers_populate()`
> 
> ideally it is PJSIP endpoint init that calls init_openssl(), but it is not necessary for now, so inserting it in ssl_create() is sufficient.
> 
> so, still init_openssl() is for global context, not SSL socket specific, don't put the SSL socket as param.
> 
okay
> > > * is it safe to use SSL socket pool for allocating global var? i.e: if SSL socket is destroyed, won't the global var memory slot be released too?
> > 
> > 
> > If we intend to use pj_lock_t, then pj_pool_t is required to create them. That is why we use `openssl_lock_init_count` to count the number of times the callback init is called. We can only release the callback after the counter set to 0.
> 
> if the SSL socket is destroyed, its pool will be released too, right? then what will happen with the pj_lock_t (that may still be used by other any SSL sockets)? that's why it is not safe.
> 
okay
> > an idea, perhaps just use OpenSSL API to allocate memory (can't remember the func name)? and don't forget to release it later.
> > > If we intend to use pj_lock_t for locking, then pj_pool_t is required.
> 
> ya that is okay, but don't use 'short lived' pool such as SSL socket's.

will create a new pool then.

pjlib/include/pj/config.h
```diff
@@ -1024,6 +1024,18 @@
...
+ * Use OpenSSL thread locking callback. This is only applicable for OpenSSL
+ * version prior to 1.1.0
+ *
+ * Default: 1 (enabled)
+ */
+#ifndef PJ_SSL_SOCK_OSSL_USE_THREAD_CB
+#   define PJ_SSL_SOCK_OSSL_USE_THREAD_CB   1
+#else
+#   define PJ_SSL_SOCK_OSSL_USE_THREAD_CB   0
+#endif
```
nanangizz: hmm, why not enable by default if version < 1.1.0, e.g:
```
#ifndef PJ_SSL_SOCK_OSSL_USE_THREAD_CB
#  if OPENSSL_VERSION_NUMBER < 0x10100000L
#    define PJ_SSL_SOCK_OSSL_USE_THREAD_CB    1
#  else
#    define PJ_SSL_SOCK_OSSL_USE_THREAD_CB    0
#  endif
#endif
```

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -588,14 +675,50 @@ static pj_status_t init_openssl(void)
...
     return status;
 }
 
 
 /* Shutdown OpenSSL */
-static void shutdown_openssl(void)
+static void shutdown_openssl(pj_ssl_sock_t *ssock)
 {
-    PJ_UNUSED_ARG(openssl_init_count);
+    PJ_UNUSED_ARG(ssock);
```
nanangizz: remove `ssock` param?

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1210,7 +1333,7 @@ static void ssl_destroy(pj_ssl_sock_t *ssock)
     /* Potentially shutdown OpenSSL library if this is the last
      * context exists.
      */
-    shutdown_openssl();
+    shutdown_openssl(ssock);
```
nanangizz: as SSL socket instance count may reach zero in the middle of lib life time, so shutdown+reinit can be called multiple times and I think it will be inefficient (it is okay now because shutdown is actually no-op). An alternative is to schedule `shutdown_openssl()` using `pj_atexit()` so it will be called only when PJLIB is destroyed (note: must log when pj_atexit() return error!).

pjlib/src/pj/ssl_sock_ossl.c
trengginas: done

pjlib/src/pj/ssl_sock_ossl.c
trengginas: okay, will use ```pj_atexit()```

pjlib/include/pj/config.h
trengginas: version number check shouldn't be on config.h since it needed to include opensslv.h (openssl header) and other version checking (existing ones) is done on ssl_sock_ossl.c

nanangizz: ah okay. just a note, if openssl >= 1.1.0 supports this locking mech, we may need to include opensslv.h here, should be fine if openssl is enabled.

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -455,9 +455,126 @@ static int openssl_init_count;
...
+
+/* Thread lock pool.*/
+pj_caching_pool 	 cp;
+pj_pool_t 		*lock_pool;
+
+/* OpenSSL locking list. */
+pj_lock_t **ossl_locks;
+
+/* OpenSSL number locks. */
+unsigned ossl_num_locks;
```
nanangizz: put 'static' for those vars?

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -455,9 +455,126 @@ static int openssl_init_count;
...
+        status = PJ_ENOMEM;
+    }
+    return status;
+}
+
+#endif
 
 /* Initialize OpenSSL */
-static pj_status_t init_openssl(void)
+static pj_status_t init_openssl()
```
nanangizz: revert this line modification (put back 'void')?

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -588,17 +705,22 @@ static pj_status_t init_openssl(void)
...
     return status;
 }
 
-
 /* Shutdown OpenSSL */
-static void shutdown_openssl(void)
+static void shutdown_openssl()
 {
-    PJ_UNUSED_ARG(openssl_init_count);
 }
```
nanangizz: please put back 'void' & unused_arg

trengginas: done

pjlib/src/pj/ssl_sock_ossl.c
trengginas: done

pjlib/src/pj/ssl_sock_ossl.c
trengginas: done

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -455,6 +455,123 @@ static int openssl_init_count;
...
+        if (ossl_locks[i]) {
+            pj_lock_destroy(ossl_locks[i]);
+            ossl_locks[i] = NULL;
+        }
+    }
+    if (lock_pool) {
+        pj_pool_release(lock_pool);
+        lock_pool = NULL;
+    }
+    pj_caching_pool_destroy(&cp);
```
sauwming: just asking, is it okay to destroy a caching_pool that's never initialised?

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -455,6 +455,123 @@ static int openssl_init_count;
...
+    CRYPTO_set_id_callback(NULL);
+#endif
+    CRYPTO_set_locking_callback(NULL);
+
+    for (; i < ossl_num_locks; ++i) {
+        if (ossl_locks[i]) {
+            pj_lock_destroy(ossl_locks[i]);
+            ossl_locks[i] = NULL;
+        }
+    }
```
sauwming: ossl_locks = NULL;
ossl_num_locks = 0;

trengginas: done

pjlib/src/pj/ssl_sock_ossl.c
trengginas: moved to above block.

sauwming: If moved to the block, it's the other way around, there is a possibility of the caching pool being initialised but never destroyed (it is an edge case though).
If I have to choose, I will still prefer the earlier version, i.e. destroying an uninitialised caching pool, provided that it's proven to be safe.

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -455,6 +455,125 @@ static int openssl_init_count;
...
+        CRYPTO_set_id_callback(ossl_thread_id);
+#endif
+        CRYPTO_set_locking_callback(ossl_lock);
+        status = pj_atexit(&release_thread_cb);
+        if (status != PJ_SUCCESS) {
+            PJ_PERROR(1, (THIS_FILE, status,
+                      "Cannot set OpenSSL thread callback unrelease method"));
+        }
+    } else {
+        status = PJ_ENOMEM;
```
nanangizz: release the pool and destroy cp here?

pjlib/src/pj/ssl_sock_ossl.c
nanangizz: the destructor will only be called if initializer is called (because atexit() is called by initializer), so there should be no destroy without init, right? but as atexit() may return error, destructor may not be called.

sauwming: right. shouldn't we call atexit() earlier, right after caching_pool_init()?

nanangizz: the convention is to avoid occupying an atexit slot if possible. so atexit() should be called last after everything else goes well (just as is). OTOH, if any of the step fail, cleanup everything before returning error, print warning (please add), don't call atexit().

pjlib/src/pj/ssl_sock_ossl.c
trengginas: done

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -455,6 +455,139 @@ static int openssl_init_count;
...
+        CRYPTO_THREADID_set_callback(ossl_set_thread_id);
+#else
+        CRYPTO_set_id_callback(ossl_thread_id);
+#endif
+        CRYPTO_set_locking_callback(ossl_lock);
+        status = pj_atexit(&release_thread_cb);
+        if (status != PJ_SUCCESS) {
+            PJ_PERROR(1, (THIS_FILE, status,
+                  "Cannot set OpenSSL lock thread callback unrelease method"));
+            release_thread_cb();
```
nanangizz: hmm.. not sure if we should uninstall the callbacks when atexit() fails, just print warning I guess (btw, perhaps it is good idea to explicitly put keyword 'Warning!' in the log), usually app is about to be destroyed too after the library is uninitialized?

trengginas: if we are allowing the thread lock to continue it's work without having the chance to release the resources  (e.g: pool, locks) properly, wouldn't it be considered a leak?

nanangizz: yes, aware of the leak, but perhaps here are things we can consider:
- atexit failure will not disrupt the lib operations, while cancelled thread lock can be harmful (btw, please add description into PR about the background or reported problem).
- as mentioned before, usually lib destroy (where the leak happens) will be followed by app destroy soon (where OS will resolve the leak).

PR #2353, opened 2020-03-23
Update pjsua-test module call to wait for ICE & DTLS-SRTP nego before checking media

nanangizz:
This should fix intermittent failure of some test scenario involving ICE, e.g: `tests/pjsua/scripts-call/305_ice_comp_*.py`.

Also add a couple of tests for DTLS-SRTP.

### DIFF
--- a/tests/pjsua/mod_call.py
+++ b/tests/pjsua/mod_call.py
@@ -33,6 +33,13 @@ def test_func(t):
         time.sleep(1)
     if caller.inst_param.have_reg:
         time.sleep(1)
+        
+    # Check if ICE is used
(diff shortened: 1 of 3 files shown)


tests/pjsua/mod_call.py
```diff
@@ -75,6 +82,17 @@ def test_func(t):
     ##time.sleep(0.1)
     caller.sync_stdout()
     callee.sync_stdout()
+    
+    # Wait ICE nego before checking media
+    if caller_use_ice:
```
sauwming: shouldn't this be if both caller and callee use_ice?

Issue #2061, opened 2017-11-10
Unable to destroy certain PJSIP transports


pjsipbot:
_2017-11-10 03:20:52: @mingteluu_ _created the issue on trac ticket [2061](https://trac.pjsip.org/repos/ticket/2061)_

Once a pjsip transport is removed from the hash table, it cannot be manually destroyed.

Scenario:
1. pjsip_transport_register() can move a transport from the hash table to tp_list.
 See ticket #1774.

2. Calling pjsip_transport shutdown() to that transport will not destroy it since pjsip_transport_add_ref() and pjsip_transport_dec_ref() will have no effect, due to is_transport_valid() check.

Thanks to Joshua Colp for the patch.

___
_2017-11-10 03:21:54: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2017-11-10 03:21:54: @mingteluu_ _set **resolution** to **fixed**_

___
_2017-11-10 03:21:54: @mingteluu_ _commented_

In [r5690](../commit/39b6c288775b07a944750902742a956e1662c93e):
Fixed #2061: Unable to destroy certain PJSIP transports

PR #2357, opened 2020-03-26
Fix leak of ioqueue resource (PJ_ETOOMANY)

Bjoe:
Galaxy S7
Android 8
Pjsip 2.7.2

Scenario:

A pjsip tls transport is getting closed because for example of some network issue.
We create a new one, with the same remote address and tranport type because of register expire timeout.

Following happens in a rare case if that is happen very quickly in a row:

To create a new connection, [pjsip_transport_register()](https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L1071) is called.
Because of that the "old closing" transport is still "valid" this transport still exist in the `pjsip_tpmgr::table`.  https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L1092 will found the "old" tranport, move it to `pjsip_tpmgr::tp_list` https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L1103 and delete https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L1104 the "old" transport entry in the `pjsip_tpmgr::table`
This has the effect that the tranport will not get in the `PJSIP_TP_STATE_DESTROY` state to call [pj_ioqueue_unregister()](https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjlib/src/pj/ioqueue_select.c#L467) to unregister handle from ioqueue, because `istransportvalid()` https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L977 is false. This function is used in [pjsip_transport_dec_ref()](https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L1018) to verify if the transport is valid https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L1038. The [transport_idle_callback()](https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L1058) is never schedule.

I added in `istransportvalid()` when the transport is not found in `pjsip_tpmgr::table` it should verify if the transport is in `pjsip_tpmgr::tp_list`. This also effects [pjsip_transport_add_ref()](https://github.com/pjsip/pjproject/blob/63b8ea5c14bff15fd6b080cef632ebf203998a8f/pjsip/src/pjsip/sip_transport.c#L983). I'm not sure if this is a good idea but it looks more consistent.

It was not very helpful to see in the log that a transport get a PJ_ETOOMANY state without to see from where this comes. So I also added a log error to see from where PJ_ETOOMANY is coming. I see there are many places where PJ_ETOOMANY is set, so I think it should be a coding convention to log an error in the same place.

Btw. coding convention, I didn't find your format convention, so it might be broken. I will fix this if somebody can send me the formatting convention.

### DIFF
--- a/pjlib/src/pj/ioqueue_select.c
+++ b/pjlib/src/pj/ioqueue_select.c
@@ -339,19 +339,19 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock2(pj_pool_t *pool,
      * an fd that is higher than FD_SETSIZE.
      */
     if (sizeof(fd_set) < FD_SETSIZE && sock >= FD_SETSIZE) {
-    PJ_LOG(1, ("pjlib", "Failed to register socket to ioqueue because "
-		   	    "socket fd is too big (fd=%d/FD_SETSIZE=%d)",
-		   	    sock, FD_SETSIZE));
-    	return PJ_ETOOBIG;
+	PJ_LOG(1, ("pjlib", "Failed to register socket to ioqueue because "
+			    "socket fd is too big (fd=%d/FD_SETSIZE=%d)",
+		   sock, FD_SETSIZE));
+	return PJ_ETOOBIG;
     }
 
     pj_lock_acquire(ioqueue->lock);
 
     if (ioqueue->count >= ioqueue->max) {
-        PJ_LOG(1, ("pjlib", "Failed to register socket %d to ioqueue because "
-                   "current key count %d reaches max key count %d (PJ_ETOOMANY)",
-                   sock, ioqueue->count, ioqueue->max));
-        rc = PJ_ETOOMANY;
+	PJ_LOG(1, ("pjlib", "Failed to register socket %d to ioqueue because "
+			    "current key count %d reaches max key count %d (PJ_ETOOMANY)",
+		   sock, ioqueue->count, ioqueue->max));
+	rc = PJ_ETOOMANY;
 	goto on_return;
     }
 
@@ -365,7 +365,7 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock2(pj_pool_t *pool,
 
     pj_assert(!pj_list_empty(&ioqueue->free_list));
     if (pj_list_empty(&ioqueue->free_list)) {
-        PJ_LOG(1, ("pjlib", "Failed to get free list of ioqueue key (PJ_ETOOMANY)"));
+	PJ_LOG(1, ("pjlib", "Failed to get free list of ioqueue key (PJ_ETOOMANY)"));
 	rc = PJ_ETOOMANY;
 	goto on_return;
     }
--- a/pjlib/src/pj/ioqueue_select.c
+++ b/pjlib/src/pj/ioqueue_select.c
@@ -339,7 +339,7 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock2(pj_pool_t *pool,
      * an fd that is higher than FD_SETSIZE.
      */
     if (sizeof(fd_set) < FD_SETSIZE && sock >= FD_SETSIZE) {
-	PJ_LOG(4, ("pjlib", "Failed to register socket to ioqueue because "
+    PJ_LOG(1, ("pjlib", "Failed to register socket to ioqueue because "
 		   	    "socket fd is too big (fd=%d/FD_SETSIZE=%d)",
 		   	    sock, FD_SETSIZE));
     	return PJ_ETOOBIG;
@@ -348,6 +348,9 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock2(pj_pool_t *pool,
     pj_lock_acquire(ioqueue->lock);
 
     if (ioqueue->count >= ioqueue->max) {
+        PJ_LOG(1, ("pjlib", "Failed to register socket %d to ioqueue because "
+                   "current key count %d reaches max key count %d (PJ_ETOOMANY)",
+                   sock, ioqueue->count, ioqueue->max));
         rc = PJ_ETOOMANY;
 	goto on_return;
     }
@@ -362,6 +365,7 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock2(pj_pool_t *pool,
 
     pj_assert(!pj_list_empty(&ioqueue->free_list));
     if (pj_list_empty(&ioqueue->free_list)) {
+        PJ_LOG(1, ("pjlib", "Failed to get free list of ioqueue key (PJ_ETOOMANY)"));
 	rc = PJ_ETOOMANY;
 	goto on_return;
     }
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -974,7 +974,21 @@ static pj_bool_t is_transport_valid(pjsip_transport *tp, pjsip_tpmgr *tpmgr,
 				    const pjsip_transport_key *key,
 				    int key_len)
 {
-    return (pj_hash_get(tpmgr->table, key, key_len, NULL) == (void*)tp);
+    pj_bool_t result = PJ_FALSE;
+    if(pj_hash_get(tpmgr->table, key, key_len, NULL) == (void*)tp) {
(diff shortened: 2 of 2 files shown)


trengginas:
Thanks for the patch. For coding convention we are using column size 80.  And for indentation setting please refer to https://trac.pjsip.org/repos/wiki/Using_Eclipse_with_PJSIP#SettingUpEclipse

Bjoe:
@trengginas Thanks for your link to the coding convention. I will fix this, give me some time.

Bjoe:
@trengginas I fixed the source code format. Hopefully it is now correct because I use QtCreator with clang format. I tried to set up a clang format that conforms to your coding convention.
Btw.I register to Bitrise to see what is failing to fix the error. But I get always a 404 when I use the link from the "Summary" page.

Bjoe:
Btw. I found that this was also fixed in the issue #2061

nanangizz:
Well, as the issue has actually been fixed, I guess we should close the PR. But, the additional log lines and readjusting log verbosity level are still good idea, we'd be glad to merge it upstream, but I am afraid that the verbosity level jump from 4 to 1 is perhaps too far, I don't think it is a critical error as it will recover after any transport is released, so log level 2 or 3 is perhaps more appropriate.

Bjoe:
@nanangizz  Yes I will close this PR.

First of all, in my opinion, an error should not occur frequently. If it is occur frequently, then it is not an error for me "ala" an exception should be an exception :-)

I will explain in short what was happen to have a clear picture:
In my case following happens:
In a release version we decrease the log to error level. So I've seen only following log line from my user/customer:
`pjsua_acc.c Unable to create/send REGISTER: Too many objects of the specified type (PJ_ETOOMANY) [status=70010]`
It happens on every register and not recoverable.

When I got a log with an error from a user/customer, first I try is to understand what the error means and why it is occurs. Here it is clear, it cannot register, but why `PJ_ETOOMANY` and what this means was not clear to me.
So I try to reproduced this issue with a debug version, but it was not possible.
Next step, search `PJ_ETOOMANY` in the source code: found about 90 times.
It takes some time to have only an idea of the root cause of `PJ_ETOOMANY`
It would have been easier, if I see in the log line like "if is not possible to create ioqueue resources because it reaches max value" ... 
It is more "verbose" but in this case I want it to be "verbose" to find/see the root cause. This "verbosity" is here only happen when this ioqueue is leaking. This should never happen in a normal case so it is an exception/error for me.

I change also change also this https://github.com/pjsip/pjproject/blob/98fbfc48cf082bc0b75cc40fa75850c27e24b3a4/pjlib/src/pj/ioqueue_select.c#L342 from 4 to 1 because it is similar for me.
In this case I will found an error log line like:
`pjsua_acc.c Unable to create/send REGISTER: Size is too big (PJ_ETOOBIG)`
The question is here, which size?

Even, when we set this for example to 3, it is not intuitive for me that this "info" line is the root cause of the "error" log line. It should be logged on the "same" level. If you said, if it is revocable and it is only a 2 or 3 then also this line https://github.com/pjsip/pjproject/blob/1062a90f569d486644166fa42be786cbc9eb8d46/pjsip/src/pjsua-lib/pjsua_acc.c#L2756 should be log on the same level.

But anyways, at the end `pj_ioqueue_register_sock2()` is failing with `PJ_ETOOBIG` and this should be only an exception and not occur frequently like `PJ_ETOOMANY`, or?

That's my opinion.

nanangizz:
Well, first, as can be seen in `pjsua_logging_config` docs [here](https://www.pjsip.org/pjsip/docs/html/structpjsua__logging__config.htm#aef4b96f808214ab8e273b69f3e8efa54), the default value is 4-5 (from max 6), and the default value is generally the recommended value for app.

Second, generally 5 is for info, 4 is for warning/light/recoverable error, 3 or less are for errors and we don't really put a very strict rule on which error should use 1/2/3, just a convention that the less value should be less critical (while `pjsua_error()` is an exception, all error is treated as 1).

Basically we don't expect app to use verbosity level less than 4 and using it to investigate/debug an issue. For issue reporting, we usually ask for log file with verbosity level 5 for investigation. So unless app is considered very stable or very tight on storage, generally we'd not recommend app to use log verbosity level as low as 1.

Bjoe:
> Second, generally 5 is for info, 4 is for warning/light/recoverable error, 3 or less are for errors 

Hm I think you mean 5 is for Debug and 4 for Info and 3 is for warning/light/recoverable error. Or maybe I missed something, but what I see in the code is following:
https://github.com/pjsip/pjproject/blob/0077afc2aa876da2a9fc0452b6a4962452f512ef/pjlib/src/pj/log.c#L329-L330
https://github.com/pjsip/pjproject/blob/0077afc2aa876da2a9fc0452b6a4962452f512ef/pjlib/include/pj/config.h#L401-L408
https://github.com/pjsip/pjproject/blob/0077afc2aa876da2a9fc0452b6a4962452f512ef/pjlib/src/pj/log.c#L347-L348

> and we don't really put a very strict rule on which error should use 1/2/3, just a convention that the less value should be less critical

Ok.

> For issue reporting, we usually ask for log file with verbosity level 5 for investigation. So unless app is considered very stable or very tight on storage

Yep :+1: with a release version I mean a considered stable version :-) and for reporting a issue for sure you will have a debug log.

Ok I think I will close this PR.
Thanks for you clearification.

nanangizz:
> https://github.com/pjsip/pjproject/blob/0077afc2aa876da2a9fc0452b6a4962452f512ef/pjlib/include/pj/config.h#L401-L408
> 
> 
> https://github.com/pjsip/pjproject/blob/0077afc2aa876da2a9fc0452b6a4962452f512ef/pjlib/src/pj/log.c#L347-L348
> 

Yes, unfortunately somehow many error logs were printed using verbosity level > 1. Perhaps this is one of the reason: as the lib comprises of different level of libraries (e.g: PJSUA built on PJSIP, PJMEDIA, etc, which built on PJLIB-UTIL, so on), there was an idea that the lower the library level the less verbose the log tends to be (e.g: errors in PJLIB tend to use level 3-4 instead of 1).

tangxiangpi:
> > Second, generally 5 is for info, 4 is for warning/light/recoverable error, 3 or less are for errors
> 
> Hm I think you mean 5 is for Debug and 4 for Info and 3 is for warning/light/recoverable error. Or maybe I missed something, but what I see in the code is following:
> https://github.com/pjsip/pjproject/blob/0077afc2aa876da2a9fc0452b6a4962452f512ef/pjlib/src/pj/log.c#L329-L330
> 
> 
> https://github.com/pjsip/pjproject/blob/0077afc2aa876da2a9fc0452b6a4962452f512ef/pjlib/include/pj/config.h#L401-L408
> 
> 
> https://github.com/pjsip/pjproject/blob/0077afc2aa876da2a9fc0452b6a4962452f512ef/pjlib/src/pj/log.c#L347-L348
> 
> > and we don't really put a very strict rule on which error should use 1/2/3, just a convention that the less value should be less critical
> 
> Ok.
> 
> > For issue reporting, we usually ask for log file with verbosity level 5 for investigation. So unless app is considered very stable or very tight on storage
> 
> Yep 👍 with a release version I mean a considered stable version :-) and for reporting a issue for sure you will have a debug log.
> 
> Ok I think I will close this PR.
> Thanks for you clarification.
Hey guy i meat the same problem, https://github.com/pjsip/pjproject/issues/2400
did you found the reason cause the  problem?

Issue #2351, opened 2020-03-19
Semaphores in sandboxed Mac app


eofster:
I was debugging the same problem of PJMEDIA event manager not starting in recent PJSIP version in a sandboxed Mac app as reported in #2331. With the same error “Error creating PJMEDIA event manager: Operation not permitted [status=120001]”.

The last version that works is 2.7.2. It was “broken” by f61de805c8ef2af6d450c701f06902143e26369c, specifically by adding the event manager creation to `pjsip/src/pjsua-lib/pjsua_media.c`.

I discovered that the problem is in POSIX semaphores. They [don’t work](https://developer.apple.com/library/archive/documentation/Security/Conceptual/AppSandboxDesignGuide/AppSandboxInDepth/AppSandboxInDepth.html#//apple_ref/doc/uid/TP40011183-CH3-SW24) in a sandboxed Mac app by default. To make them work, apparently one has to create a so-called app group. App groups are normally used to share data between apps of the same developer. And POSIX semaphores work in a sandboxed environment only when their names are composed in a special way containing app group names. Also, there is a limit of 31 bytes for the semaphore name, which puts additional restriction to the app group name.

I think semaphores never worked in sandboxed Mac apps. The problem wasn’t coming up only by coincidence, probably because no semaphore code was running in a typical PJSUA API usage. I’ve been using it for years without encountering this problem!

One possible solution to this is to make the app developer create an artificial app group and to make this name configurable in PJSIP.

Another solution would be to use semaphores from the dispatch library (`dispatch_semaphore_create(3)`). This is a newer Apple’s approach to multitasking. I’ve already implemented it in PJSIP and can share results, if you’re interested.

sauwming:
Thanks for the information!

The second solution sounds better if it can work seamlessly without requiring users to configure semaphore names.

Please do share with us the modifications needed, we really appreciate it. Or even better, if you think it's ready to be merged into the master branch, feel free to create a PR.

eofster:
@sauwming what is the correct place to add a compile-time setting like `PJ_SEMAPHORE_USE_DISPATCH_SEM`?

sauwming:
It should be here:
https://github.com/pjsip/pjproject/blob/master/pjlib/include/pj/config.h#L806

eofster:
Thanks! If I wanted to enable it by default on Darwin, what would be the best place to do it?
[...]

PR #2362, opened 2020-03-27
Semaphores in sandboxed Mac app 2351

Slash333:
Here is resolving of this issue: https://github.com/pjsip/pjproject/issues/2351


sauwming:
I would suggest not to remove the current implementation. Perhaps you can add a compile-time setting `PJ_SEMAPHORE_USE_DISPATCH_SEM`, i.e.
```
#if PJ_SEMAPHORE_USE_DISPATCH_SEM
... // your patch which uses dispatch_semaphore here
#else
... // current implementation which uses sem_open
#endif
```
Also avoid any unnecessary modification such as indentation/white space changes.

And btw, you need to sign the CLA before we can merge this later.

sauwming:
There's another PR #2411 addressing the same issue, which has now been merged into master. Unfortunately we can't proceed with this one without the CLA, but we thank and appreciate your contribution.

PR #2363, opened 2020-03-27
ice: rfc6544 support

AmarOk1412:
This patch is an implementation proposal of the RFC 6544 into PJNATH.
This allow PJNATH to support TCP ICE candidates and open a direct TCP
connection between peers.

- BUG fix also the on_data_sent in turn_sock.c

Written by
Sébastien Blin <sebastien.blin@savoirfairelinux.com>
on behalf of Savoir-faire Linux.

Rebased for pjsip 2.10 by Peymane Marandi
<paymon@savoirfairelinux.com>
on behalf of Savoir-faire Linux.

--------------------------------------

Hi,

Since a long time, we have some patches for pjsip we use and a bit hard to maintain.

**The patch is far from finished/perfect, lot of dirty things/hacks, but is working** We use it since 1 year and a half on a lot of clients. The idea is to support ICE over TCP (RFC6544).

Ideally, if we can work together to merge this upstream, I think it will be great for both us and pjproject.

You already have my contributor agreement.

Best regards,
Sébastien

### DIFF
--- a/pjnath/include/pjnath/config.h
+++ b/pjnath/include/pjnath/config.h
@@ -383,6 +383,15 @@
 #   define ICE_CONTROLLED_AGENT_WAIT_NOMINATION_TIMEOUT	10000
 #endif
 
+/**
+ * For TCP transport, this timer is time that a controlling agent must wait for
+ * incoming checks if the local candidate is of type "passive" or "s-o".
+ *
+ * Default: 10000 (milliseconds)
+ */
+#ifndef ICE_CONTROLLING_PASSIVE_TIMEOUT
+#   define ICE_CONTROLLING_PASSIVE_TIMEOUT	10000
+#endif
 
 /**
  * For controlling agent if it uses regular nomination, specify the delay to
--- a/pjnath/include/pjnath/ice_session.h
+++ b/pjnath/include/pjnath/ice_session.h
@@ -163,15 +163,58 @@ typedef enum pj_ice_cand_type
 
 } pj_ice_cand_type;
 
+/**
+ * ICE candidates types like described by RFC 6544.
+ */
+typedef enum pj_ice_cand_transport {
+    /**
+     * Candidates UDP compatible
+     */
+    PJ_CAND_UDP,
+    /**
+     * Candidates sending outgoing TCP connections
+     */
+    PJ_CAND_TCP_ACTIVE,
+    /**
+     * Candidates accepting incoming TCP connections
+     */
+    PJ_CAND_TCP_PASSIVE,
+    /**
+     * Candidates capable of receiving incoming connections and sending
+     * connections
+     */
+    PJ_CAND_TCP_SO
+} pj_ice_cand_transport;
+
+/**
+ * ICE transport types, which will be used both to specify the connection
+ * type for reaching candidates and other client
+ */
+typedef enum pj_ice_tp_type {
+    /**
+     * UDP transport, which value corresponds to IANA protocol number.
+     */
+    PJ_ICE_TP_UDP = 17,
+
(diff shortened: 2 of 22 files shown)


AmarOk1412:
Updated. Our current stack is based on pjsip 2.11 with 19 patches (Some of them are merged upstream in 2.12)

nanangizz:
Thank you for updating this, will try to get a time to check it.

sauwming:
Closing this since it now has too many conflicts with the current master. Besides, the unfinished state makes this too difficult to integrate.

AmarOk1412:
> Closing this since it now has too many conflicts with the current master.

Actually I still maintain the patch and rebase it. https://github.com/pjsip/pjproject/commit/ed0ffc9f0e9be11155b7eb1d5512ec32470aeaed

> Besides, the unfinished state makes this too difficult to integrate.

I never got any review/feedback so difficult to know what is unfinished as it's not in your Roadmap. Unit-testing for sure, the rest is pretty functional and should not break current things.

sauwming:
The reason the patch didn't manage to undergo a review is due to the patch description itself: `The patch is far from finished/perfect, lot of dirty things/hacks`. So we thought it was still in progress or there would still be some finishing changes to remove the hacks and perfect things up. Not to mention that the sheer size of the patch itself makes it a daunting task to review (this may be by far the largest contributed patch we've ever received).

While the reason it's not in the roadmap is because there doesn't seem to be any other request for this feature.

So perhaps you can clarify the state of the patch itself, i.e. which parts are still unfinished/todo, and which parts are "hackish". Although I still can't promise that the PR will be accepted, at least it will make the review task easier for us to know what still needs to be done before this can be integrated.

AmarOk1412:
`The patch is far from finished/perfect, lot of dirty things/hacks`

For this, it's mostly because pjsip is a large project and I'm not a pjproject developer. So, the architecture may have to change. Also there is no unit-tests. But never got any feedback, so can't know if there is anything wrong with the design I did.

But we use this stack for years in our every day life, so the code is functional and maintained. My original comment is from 2020, but everything is working since the comment.

So yeah, for me the patch is ready since my PR and I hoped any feedback to be able to fix it. For the hacky/dirty part, I can't really say as it mostly depends on the architecture. The only dirty logic is the `PJ_ICE_SESS_CHECK_STATE_NEEDS_RETRY` I don't like on a connection reset while the peer didn't authorize the ip to connect and that we don't use trickle ICE ourself.

I understand that the patch is not in your roadmap and fine with this reason. But for the un-finished state I was always open to discuss about this as I said in my first comment.

sauwming:
Since we just released 2.14, I suppose we have plenty of time to revisit and work on this again.

A few initial comments:
- Please push the latest version to resolve the conflicts.
- Modifying existing API is generally not recommended due to backward compatibility reason. So API such as `pj_stun_sock_create()` must not be changed. You either need to put the new field inside `pj_stun_sock_cfg`, or create a new API `pj_stun_sock_create2()`.
- Do you plan to extend this to pjsua? One of the problem with this feature is that it can't be easily tested. According to section 4.1:
```
Providers of real-time communications services may decide that it is
   preferable to have no media at all rather than to have media over
   TCP.  To allow for choice, it is RECOMMENDED that it be possible to
   configure agents to either obtain or not obtain TCP candidates for
   real-time media.
```
So it would be great if there's option in pjsua to enable ICE over TCP (default: false), and provide option in pjsua app (such as `--enable-ice-tcp`). This way, a lot of PJSIP users would be able to try out this feature and benefit from it.

sauwming:
Closing this due to code conflicts and lack of update.

Issue #2334, opened 2020-03-06
PJCamera.java bug (Android)


scottishdeveloper:
1. Rotate the camera to back facing
2. Destroy the call (hangup)
3. Destroy camera preview
4. PJCamera -> Stop() is never called

Subsequent attempts to acquire the camera with further calls will fail with a -10 error, due to Camera in use.

This results in the Camera being locked forever and never "released", until an application force restart.

All versions up to 2.9,  not tested latest dev branches.

sauwming:
Please check #2368 and let us know if the issue is not resolved by that PR.

PR #2368, opened 2020-03-31
VideoPreview not updating its device ID after device fast-switched

nanangizz:
When application instantiates a VideoPreview then performs camera switch that is done using fast-switching method (e.g: on Android and iOS), VideoPreview::stop() will fail and it fails silently (no exception thrown).

Thanks to Dmytrii Gonchar for the report.


sauwming:
Does this fix, or is somehow related to, #2334 as well?

nanangizz:
> Does this fix, or is somehow related to, #2334 as well?

Yes, looks like the same symptom.

PR #2381, opened 2020-04-10
Status phrases

gaaf:
Update list of status-phrases:

- Add new response codes
- Update existing phrases to match the spelling in the RFC's


gaaf:
The changes to the existing phrases are so that they are easier recognized and/or searched for and make it a bit less obvious to an observer that PJSIP is in play with their unique phrasing.

gaaf:
CI failure seems unrelated.

andreas-wehrmann:
Hey there, it'd be nice if the newly added codes would also show up in the pjsip_status_code enum.
See here:
https://github.com/pjsip/pjproject/blob/b8eea5d015e21030cdf2e78f965e1576f8fa1257/pjsip/include/pjsip/sip_msg.h#L403
https://github.com/pjsip/pjproject/blob/b8eea5d015e21030cdf2e78f965e1576f8fa1257/pjsip-apps/src/swig/symbols.i#L488

gaaf:
> Hey there, it'd be nice if the newly added codes would also show up in the pjsip_status_code enum.
> See here:
> 
> https://github.com/pjsip/pjproject/blob/b8eea5d015e21030cdf2e78f965e1576f8fa1257/pjsip/include/pjsip/sip_msg.h#L403

Done.



> https://github.com/pjsip/pjproject/blob/b8eea5d015e21030cdf2e78f965e1576f8fa1257/pjsip-apps/src/swig/symbols.i#L488

I don't think I could/should add it there. Top of file says:
```// This file is autogenerated by importsym script, do not modify!```

And the script that is supposed to create that file, can only be run on one system in the world:
```
python importsym.py 
Error: couldn't find pycparser utils in '/home/bennylp/Desktop/opt/src/pycparser-master'
```

andreas-wehrmann:
Good catch; I simply grepped for the enum's...
Maybe the generated file is a leftover?
I'm wondering why an autogenerated file is submitted at all.

sauwming:
I updated `symbols.i`.
Although it is generated by a script, it's not automatically generated every time the project is built (due to the requirement of pycparser as you have found out, as well as the fact that the symbols very rarely change). So that's why it's included in the project.

gaaf:
> due to the requirement of pycparser

That is not the problem I had, having pycpraser is easy. The problem is the hardcoded path to pycparser which only a single person will have in that location.

sauwming:
Thanks @vasilevalex for noticing this. I fixed this in https://github.com/pjsip/pjproject/commit/b53b8e8651fd86ef0beb3fa9e3f6937ca7f9a71c.

PR #2383, opened 2020-04-10
pjsip: De-obfuscate prefixing the branch param with "Pj"

gaaf:
The target is a char*, put chars in it, not ints or the result of some
const calculation.


sauwming:
It was intentionally created that way :)

gaaf:
> It was intentionally created that way :)

Care to explain why?


And maybe even share the reason the 'Pj' is there in the first place? It adds no entropy at all to a field that is supposed to be random.

sauwming:
It's **Pj**sip's hidden "fingerprint" :)

gaaf:
> It's Pjsip's hidden "fingerprint" :)

I understand the intended "use", but that still does not explain why the code adding it is obfuscated.

I think having code that is more easy to read instead of (intentionally) obfuscated might help prevent issues and make is easier for new developers to understand the code.

PR #2388, opened 2020-04-14
pjmedia: Populate missing info fields from pjmedia_conf_get_port_info

maxtruxa:
`pjmedia_conf_get_port_info` failed to populate the `format` member in `pjmedia_conf_port_info`.

### DIFF
--- a/pjmedia/src/pjmedia/conference.c
+++ b/pjmedia/src/pjmedia/conference.c
@@ -1278,6 +1278,12 @@ PJ_DEF(pj_status_t) pjmedia_conf_get_port_info( pjmedia_conf *conf,
 
     info->slot = slot;
     info->name = conf_port->name;
+    if (conf_port->port) {
+        pjmedia_format_copy(&info->format, &conf_port->port->info.fmt);
+    } else {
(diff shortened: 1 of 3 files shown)


pjmedia/src/pjmedia/conference.c
```diff
@@ -1278,6 +1278,12 @@ PJ_DEF(pj_status_t) pjmedia_conf_get_port_info( pjmedia_conf *conf,
 
     info->slot = slot;
     info->name = conf_port->name;
+    if (conf_port->port) {
+        pjmedia_format_copy(&info->format, &conf_port->port->info.fmt);
+    } else {
+        pj_bzero(&info->format, sizeof(info->format));
+        info->format.id = (pj_uint32_t)-1;
```
sauwming, line 1285: Do we need to introduce a new format in format.h, such as
```
typedef enum pjmedia_format_id
{
...
PJMEDIA_FORMAT_INVALID = 0xFFFFFFFF
```

nanangizz: agree, sounds better than -1 constant.

PR #2394, opened 2020-04-21
the simpleua app segfaults because the memory pool for the pjmedia ev…

byoungdale:
…ent manager is never created. This fixes issue #2391. I initially thought it was only segfaulting on macos, but I tested on debian as well, and it will segfaulted. It is just because the memory pull ws not being created, so when passed to `pj_event_mgr_create`, it crashed.


sauwming:
The issue occurred because pool is not created when `PJMEDIA_HAS_VIDEO` is disabled. So I proposed the following fix: https://github.com/pjsip/pjproject/pull/2394/commits/4c5d1a1223c9681011d0e55cdb8109d3741c4e56 instead.

Issue #2389, opened 2020-04-14
Increasing memory usage with UDP transport


andreas-wehrmann:
I face the following problem:
When stress-testing my application, I can observe a dramatic increase in memory usage.
The memory is never free'd while the application is running.
Valgrind reports no leak however, it seems to be stuck in some memory pool
even though the high memory usage is not reflected in the output of pjsua_dump(1);

I was recently stress-testing my app which uses the PJSUA library.
It works more or less like this:
- There are two sides: internal and external
- When a call from one side comes in, a call for the other side is made
- The calls will be connected once both are confirmed

For stress-testing the app, I wrote a little testclient (also based on PJSUA)
which works like this:
- Make call
- When call confirmed: hangup
- When a call has been disconnected: make call again

The apps run on Ubuntu 16.04.
When I run this test I noted that memory usage of my app kept increasing dramatically.
It would consume hundreds of MBytes after running less than half a minute.

Interestingly, the rate of growth does not always seem to be the same.
Sometimes it consumes hundreds of MBytes after a few seconds and other times memory
usage grows fast and slows down only to speed up later.
What I also observed is this behaviour:
- Memory usage increases fast (couple of hundred MB, maybe 1G)
- Memory usage stays constant at that point, and never decreases
- I stop the test and let the app sit for a while
- Restart the test
- Memory usage will increase again

After too much time, (I think) I narrowed it down the the SIP-UDP transport.
When I step through all the shutdown functions; memory would always be given back
when the memory pool of the UDP transport is released in udp_on_destroy
(even though the statistics of PJ do not show the huge amounts of memory apparently consumed).

I then tried a TCP only setup and couldn't observe the same issues, memory usage would always be low.
After that, I tried to see how the PJSUA app behaves when I run the test against it and could observe the same behaviour.

For this, I run the app like this:
./pjsua-x86_64-unknown-linux-gnu --bound-addr=<eth0-ip> --no-tcp --null-audio --auto-answer=200 --auto-loop

And then start my test program to bombard it with calls.

I attached a massif output file from my app (it includes the app shutdown, which is why you can see the memory usage drop after the spike).
I recommend massiv-visualizer for reading the file; Bummer is I cannot run my app under massif for too long because it'll eventually slow down to almost a halt.

Also, I attached the config_site.h which I used for building the pjsua app.

Steps to reproduce:

[...]

PR #2395, opened 2020-04-22
PJ adjustable pool sizes

andreas-wehrmann:
This can be used to deal with memory fragmentation issues in long running applications that encounter temporary high loads.

For reference, see this issue:
https://github.com/pjsip/pjproject/issues/2389


pjmedia/include/pjmedia/config.h
```diff
@@ -44,6 +44,34 @@
...
+#   define PJMEDIA_POOL_EVTMGR_LEN		500
+#endif
+
+/**
+ * Memory increment for evnt manager.
+ */
+#ifndef PJMEDIA_POOL_EVTMGR_INC
+#   define PJMEDIA_POOL_EVTMGR_INC		500
+#endif
+
```
nanangizz, line 74: Sorry just got a chance to revisit this PR.

IIRC we tried to reproduce the memory fragmentation issue (can't remember the result, can't find any doc/mail history either :D ). Anyway, some pools have been configurable already, so I think it is okay to integrate the patch, just perhaps better to try to maintain a consistent config name pattern, which mostly seem to use: `[MODULE]_POOL_LEN/INC_[OBJECT]`, e.g: `PJSIP_POOL_LEN/INC_TRANSPORT`.

andreas-wehrmann: Hello there,

I tried to adhere to the name patterns I found the in modules in which I made the changes; but I'll go and revisit the changes to see if I missed something.

As for the fragmentation issue:
I specifically made only the pools configurable which I found to be problematic in my case.
I experimented with cleaning up pools under certain conditions (which definitely helped with the fragmentation issue)
but I wasn't happy with what I came up and stayed with the mitigation enabled by the changes proposed in this PR.

pjmedia/src/pjmedia-codec/g722.c
```diff
@@ -174,8 +183,8 @@ PJ_DEF(pj_status_t) pjmedia_codec_g722_init( pjmedia_endpt *endpt )
     g722_codec_factory.endpt = endpt;
     g722_codec_factory.pcm_shift = PJMEDIA_G722_DEFAULT_PCM_SHIFT;
 
-    g722_codec_factory.pool = pjmedia_endpt_create_pool(endpt, "g722", 1000, 
-						        1000);
+    g722_codec_factory.pool = pjmedia_endpt_create_pool(endpt, "g722", G722_POOL_INIT_SIZE, 
+						        G722_POOL_INC_SIZE);
```
nanangizz, line 187: Sorry, forgot to mention this earlier.
- Perhaps better to put this setting in `pjmedia/include/pjmedia-codec/config.h`? Named `PJMEDIA_CODEC_POOL_LEN/INC_G722`?
- Also for all changes in the library code, please use max column length of 80.

Thank you.

andreas-wehrmann: Putting a general define like `PJMEDIA_CODEC_POOL_LEN` in `pjmedia/include/pjmedia-codec/config.h` would imply, that every codec uses it, no?
If I introduce this define, should I update every codec impl to make use of it?
If every codec uses it, what would a good default value be?

nanangizz: > Putting a general define like `PJMEDIA_CODEC_POOL_LEN` in `pjmedia/include/pjmedia-codec/config.h` would imply, that every codec uses it, no?

Was thinking for G722 only, as shown in the patch. But, having a common setting for all codecs sounds like a good idea.

> If I introduce this define, should I update every codec impl to make use of it?

Yes, unfortunately.

> If every codec uses it, what would a good default value be?

Assuming each codec instance will need to allocate about 100-500bytes (mostly for PJMEDIA codec structures, as some codec backends allocate their states by theirselves), IMO default value in the range of 1-5kB should be fine, and 2kB is perhaps reasonable as max calls in PJSUA by default is 4 (so 4*500B = 2kB).

Feel free if you have any opinions or suggestions.

andreas-wehrmann: Moved and renamed the defines to pjmedia-codec's config.h now;
Having a define for every codec implementation allows for more fine tuning, if that should become necessary.

andreas-wehrmann: Did you get a chance to look at the new changes?
I'm not sure whether I need to explicitly request a review or you get notified automatically.

nanangizz: Yes, we get notified automatically. Sorry for the delay, must be in the middle of something when the notification came, but then forgot to come back to check it.

Things look okay to me, need to wait for another reviewer (from PJSIP team) to be able to merge.

Issue #2398, opened 2020-04-22
Illegal flags in the pkgconfig file


H5117:
**Steps to reproduce the behavior**
1. Compile pjsip in a such way:
```bash
CPPFLAGS=" -I/tmp/contrib/x86_64-pc-linux-gnu/include -fPIC" CFLAGS=" -I/tmp/contrib/x86_64-pc-linux-gnu/include -DNDEBUG=1 -O3  -fPIC" CXXFLAGS=" -I/tmp/contrib/x86_64-pc-linux-gnu/include -DNDEBUG=1 -O3 -std=c++14 -fPIC" LDFLAGS=" -L/tmp/contrib/x86_64-pc-linux-gnu/lib" EXCLUDE_APP=1 ./aconfigure --build="x86_64-pc-linux-gnu" --host="x86_64-pc-linux-gnu" --target="x86_64-pc-linux-gnu" --program-prefix="" --disable-dependency-tracking --enable-static --disable-shared --with-pic --disable-sound --enable-video --enable-ext-sound --disable-speex-aec --disable-g711-codec --disable-l16-codec --disable-gsm-codec --disable-g722-codec --disable-g7221-codec --disable-speex-codec --disable-ilbc-codec --disable-opencore-amr --disable-silk --disable-sdl --disable-ffmpeg --disable-v4l2 --disable-openh264 --disable-resample --disable-libwebrtc --enable-epoll
make
make install
```
2. Get the following libpjproject.pc:
```
# Package Information for pkg-config

prefix=/usr/local
exec_prefix=${prefix}
libdir=/usr/local/lib
includedir=/usr/local/include

Name: libpjproject
Description: Multimedia communication library
URL: http://www.pjsip.org
Version: 2.10
[...]

PR #2403, opened 2020-04-23
libpjproject.pc: correct the field 'cflags'

H5117:
Fixes issue #2398.


trengginas:
Tested the patch here without any issue.
The patch still missing `aconfigure` file. You should be able to generate it using `"autoconf aconfigure.ac > aconfigure"`.

H5117:
Done.

PR #2408, opened 2020-04-27
Skip unsupported digest algorithm

sauwming:
Report and patch by Alexander Traud:
" (some minor edits without altering its content)
See <https://stackoverflow.com/q/60683714>, found independently.

PJSIP does not support SHA-256 but only MD5 (and AKAv1-MD5; pjsip/sip_auth_client.c:respond_digest). However, the Backus-Naur-Form in RFC 3261 allows several WWW-Authenticate lines and the implementation has to pick its supported algorithm. RFC 3261 even known about unknown algorithms (token).

Currently, PJSIP 2.10, goes for the first WWW-Authenticate line only. If that is not supported, no progress (PJSIP_EINVALIDALGORITHM)

With RFC 7616, algorithms got added like SHA-256. Recently, the SIP service at sip.linphone.org added that. Give it a try and use my account "traud", password does not matter. You get two WWW-Authenticate lines, the first is SHA-256 and the second one is MD5.

I am *not* asking for supporting "SHA-256". I am asking for picking the right, your supported algorithm. This was required in RFC 3261 already. And actually my 12-year-old Nokia Mobile Phone picks MD5.

Note: pjsua command line tool faces the same issue.
"

This current patch has two limitations:
1. WWW-Authenticate without algorithm is ignored.
2. WWW-Authenticate with AKAv1-MD5 is ignored [1].
Both cases work(ed) further down in the stack in respond_digest(.).

### DIFF
--- a/pjsip/src/pjsip/sip_auth_client.c
+++ b/pjsip/src/pjsip/sip_auth_client.c
@@ -1062,6 +1062,14 @@ static pj_status_t process_auth( pj_pool_t *req_pool,
 		    hdr = hdr->next;
 		    pj_list_erase(sent_auth);
 		    continue;
+		} else
+		if (pj_stricmp(&sent_auth->scheme, &pjsip_DIGEST_STR)==0 &&
+		    pj_stricmp(&sent_auth->credential.digest.algorithm,
+		               &hchal->challenge.digest.algorithm)==0)
+		{
+		    /* Same 'digest' scheme but different algo */
+		    hdr = hdr->next;
+		    continue;
 		} else {
 		    /* Found previous authorization attempt */
 		    break;
@@ -1155,7 +1163,7 @@ PJ_DEF(pj_status_t) pjsip_auth_clt_reinit_req(	pjsip_auth_clt_sess *sess,
 {
     pjsip_tx_data *tdata;
     const pjsip_hdr *hdr;
-    unsigned chal_cnt;
+    unsigned chal_cnt, auth_cnt;
     pjsip_via_hdr *via;
     pj_status_t status;
 
@@ -1178,6 +1186,7 @@ PJ_DEF(pj_status_t) pjsip_auth_clt_reinit_req(	pjsip_auth_clt_sess *sess,
      */
     hdr = rdata->msg_info.msg->hdr.next;
     chal_cnt = 0;
+    auth_cnt = 0;
     while (hdr != &rdata->msg_info.msg->hdr) {
 	pjsip_cached_auth *cached_auth;
 	const pjsip_www_authenticate_hdr *hchal;
@@ -1194,10 +1203,6 @@ PJ_DEF(pj_status_t) pjsip_auth_clt_reinit_req(	pjsip_auth_clt_sess *sess,
 	    break;
 
 	hchal = (const pjsip_www_authenticate_hdr*)hdr;
-	if (pj_stricmp(&hchal->challenge.digest.algorithm, &pjsip_MD5_STR)) {
-	    hdr = hdr->next;
-	    continue;
-	}
 	++chal_cnt;
 
 	/* Find authentication session for this realm, create a new one
@@ -1226,8 +1231,11 @@ PJ_DEF(pj_status_t) pjsip_auth_clt_reinit_req(	pjsip_auth_clt_sess *sess,
 	 */
 	status = process_auth(tdata->pool, hchal, tdata->msg->line.req.uri,
 			      tdata, sess, cached_auth, &hauth);
-	if (status != PJ_SUCCESS)
-	    return status;
+	if (status != PJ_SUCCESS) {
+	    /* Process next header. */
+	    hdr = hdr->next;
+	    continue;
+	}
 
 	if (pj_pool_get_used_size(cached_auth->pool) >
 	    PJSIP_AUTH_CACHED_POOL_MAX_SIZE) 
@@ -1240,11 +1248,16 @@ PJ_DEF(pj_status_t) pjsip_auth_clt_reinit_req(	pjsip_auth_clt_sess *sess,
 
 	/* Process next header. */
 	hdr = hdr->next;
+	auth_cnt++;
     }
 
     /* Check if challenge is present */
     if (chal_cnt == 0)
 	return PJSIP_EAUTHNOCHAL;
+
+    /* Check if any authorization header has been created */
+    if (auth_cnt == 0)
+	return PJSIP_EAUTHNOAUTH;
 
     /* Remove branch param in Via header. */
     via = (pjsip_via_hdr*) pjsip_msg_find_hdr(tdata->msg, PJSIP_H_VIA, NULL);
--- a/pjsip/src/pjsip/sip_auth_client.c
+++ b/pjsip/src/pjsip/sip_auth_client.c
@@ -1042,7 +1042,7 @@ static pj_status_t process_auth( pj_pool_t *req_pool,
     pjsip_hdr *hdr;
     pj_status_t status;
 
-    /* See if we have sent authorization header for this realm */
+    /* See if we have sent authorization header for this realm (and scheme) */
     hdr = tdata->msg->hdr.next;
     while (hdr != &tdata->msg->hdr) {
 	if ((hchal->type == PJSIP_H_WWW_AUTHENTICATE &&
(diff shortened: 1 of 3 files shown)


nanangizz:
`pjsip_auth_clt_reinit_req()` iterates all auth headers already, it just returns error immediately when finding an algo is not supported, so next auth header is not evaluated. So, I think we'd need to consult the standard first: when receiving multiple auth headers (with same realm?), should PJSIP respond to all algo (current behavior) or supported algo only?

traud:
https://tools.ietf.org/html/rfc7616#section-3.9.1
example is explained in general in chapter 3.7. Furthermore, there is the definition of ‘algorithm’: 
> If the algorithm is not understood, the challenge SHOULD be ignored (and a different one used, if there is more than one).

This sentence is the same in the new RFC 7616 and the old RFC 2617.

nanangizz:
Just proposed an alternate solution, please check :)

traud:
No, that approach that does not work (correctly). Just tested, to double-check.

I used that approach (or at least a similar one) at start as well. That approach creates additional SIP requests. For example, pjsua invites someone and my server has MD5 as the second algorithm: pjsua creates not two but *four* INVITEs. Not the second but (only) the *fourth* INVITE contains a header `Authorization: Digest …`. Although that authenticates then successfully, such a behavior might have triggered suspicious filterers like [fail2ban](https://en.wikipedia.org/wiki/Fail2ban).

To be honest, I do not know what exactly those two additional SIP requests create. Therefore, I cannot say what is wrong with your approach. At start, just reading the code, I went for the same approach like you. Did you test it against _sip.linphone.org_? What is wrong with my approach? I can enhance it for the two cases no-algorithm and AKAv1-MD5. I did not do it because I am not that steady in PJStr comparison (and I am not sure where to move that AKAv1-MD5 constant then).

nanangizz:
Just tested with _sip.linphone.org_, but unfortunately it responded (for both `REGISTER` & `INVITE` requests) with one authenticate header line only (?). The patch work fine with just added SIPp test scripts though, no multiple `REGISTER`.

Your patch is generally fine (minor issue, missing scheme check), but it adds more place doing scheme/supported-algo checks, currently it is done in `process_auth()` (and also further down in `auth_respond()` & `respond_digest()`). Basically, just feel better to leave the algo check in `process_auth()` as is now.

traud:
Go for another, a third
`WWW-Authenticate: Digest … algorithm=MD2 …`
Yes, this MD-Two is a fictive but possible algorithm with the same name length as MD-Five.

nanangizz:
Good test, found a bug with the MD2! The Authorization header generated was okay (REGISTER retried with only one Auth header, etc), but the error message was `Authorization failed ... server rejected with stale=false` while it should be `Unsupported digest algorithm "MD2"`.

traud:
Works and passes my tests. Magically (?) those additional SIP requests are gone as well.

Nevertheless, I do not understand that whole code file. For example:
1. `process_auth(.)` returns many more status than just `PJSIP_EINVALIDALGORITHM`. But *all* of them seem to be silenced now after your patch.
2. What happens if the *realms* are different? If I am not totally misreading the code for `cached_auth =`, the realm of the first challenge is used always.

I have not tested this, yet. I understood my patch. It made sense. Now with the current patch, I do not understand anything. I cannot judge on yours, except that it works now for my test cases. Sorry. Cannot help here further.

nanangizz:
> Works and passes my tests. Magically (?) those additional SIP requests are gone as well.

Great! Not sure what might cause the additional SIP requests, the code seems to generate only one request to me.

> Nevertheless, I do not understand that whole code file. For example:
> 
> 1. `process_auth(.)` returns many more status than just `PJSIP_EINVALIDALGORITHM`. But _all_ of them seem to be silenced now after your patch.

Yes, but as the function is actually still invoked for each auth challenge, any log will still be printed. And as the point seems to generate a valid auth respond, ignoring error of other (unsupported) challenge should be fine, and when no algo is supported, it will return `PJSIP_EAUTHNOAUTH` instead of `PJSIP_EAUTHNOCHAL` (with the original patch). Note that with the original patch, the function will be completely skipped for any unsupported algo.

> 2. What happens if the _realms_ are different? If I am not totally misreading the code for `cached_auth =`, the realm of the first challenge is used always.

Not sure which part of the code points to the first challenge.

> I have not tested this, yet. I understood my patch. It made sense. Now with the current patch, I do not understand anything. I cannot judge on yours, except that it works now for my test cases. Sorry. Cannot help here further.

Basically the idea is the same, it's just that the new patch is letting the algo check (among other checks) to be done inside `process_auth()`.

Thank you for the report, the patch, and all the helps.

traud:
Not sure if you learned whether such discussion|sentence style is successful and therefore you do it. Are you interested why exactly the four SIP requests were created? I am not sure if that statement of you was curiosity = question; or simply a do not care anymore.

> Which part of the code points to the first challenge?

`cached_auth = … hchal->challenge.common.realm …`
I hope at least that was a question. I tested your change, and it does not use the first realm over the wire. However, I do not understand what happens there in that code part. `cached_auth` is initialized with the first, possible wrong challenge. Where is that overridden? That was the reason I did my change before `cached_auth` gets assigned.

> log […] still be printed

No question either. Are you interested? Now, I have to _defend_ my curiosity. Wrong, isn’t it?

I am not about the log messages but the return/status value. Previously, the status of `process_auth(.)` was forwarded, for example, `pjsip_auth_clt_reinit_req(.)` did return `PJSIP_EINVALIDALGORITHM`, which still can happen. Now, you changed the API to return just 
a) `PJ_SUCCESS`
b) `PJSIP_EAUTHNOCHAL`
c) `PJSIP_EAUTHNOAUTH`
The latter is even a new status which was only used in `pjsip_auth_srv_verify(.)` before.

This could affect upstream projects, for example, Digium Asterisk creates [own log messages …](https://github.com/asterisk/asterisk/blob/master/res/res_pjsip_outbound_authenticator_digest.c#L153)

nanangizz:
> Not sure if you learned whether such discussion|sentence style is successful and therefore you do it. Are you interested why exactly the four SIP requests were created? I am not sure if that statement of you was curiosity = question; or simply a do not care anymore.
> 

Actually with your last sentence in [this comment](https://github.com/pjsip/pjproject/pull/2408#issuecomment-622012362), I thought it was you that didn't want to continue the discussion (for whatever reason), hence I said thanks and did such sentence style :)

The four SIP request was not reproduced here, so as mentioned before, I checked the code, but could not see how four SIP requests may be generated. If you provide a log to investigate or even have a ready explanation, I am definitely open and all ears, simple, no need to say things with such tone.

I decided to propose an alternative patch because problems in the original patch:
- more places doing supported algo check. If SHA-256 is implemented in the future, with original patch, both challenges MD5 & SHA-265 will go through `cached_auth=...`, `process_auth()` and so on, the same way as the alternative patch, so why not pass all challenges and let `process_auth()` to filter for things such as supported algo (as it is also already doing it).
- minor: limitation you mentioned (AKA-MD5)
- minor: directly accessing digest without scheme check

Please note that I don't mind at all to go back with the approach in the original patch if it is indeed the correct one.

> > Which part of the code points to the first challenge?
> 
> `cached_auth = … hchal->challenge.common.realm …`
> I hope at least that was a question. I tested your change, and it does not use the first realm over the wire. However, I do not understand what happens there in that code part. `cached_auth` is initialized with the first, possible wrong challenge. Where is that overridden? That was the reason I did my change before `cached_auth` gets assigned.
> 

AFAIK, `cached_auth` is supposed to be unique by realm and it initializes/updates its challenge info (e.g: nonce, nc) when that challenge is accepted (authorization header is generated for that challenge). Should you see any problem with patch doing on `cached_auth`, feel free to describe.

I think `pjsip_auth_clt_reinit_req()` should only accept one challenge for each realm, and this is still missing from the code (and any patch). Any comment on this?

> > log […] still be printed
> 
> No question either. Are you interested? Now, I have to _defend_ my curiosity. Wrong, isn’t it?
> 
> I am not about the log messages but the return/status value. Previously, the status of `process_auth(.)` was forwarded, for example, `pjsip_auth_clt_reinit_req(.)` did return `PJSIP_EINVALIDALGORITHM`, which still can happen. Now, you changed the API to return just
> a) `PJ_SUCCESS`
> b) `PJSIP_EAUTHNOCHAL`
> c) `PJSIP_EAUTHNOAUTH`
> The latter is even a new status which was only used in `pjsip_auth_srv_verify(.)` before.
> 
> This could affect upstream projects, for example, Digium Asterisk creates [own log messages …](https://github.com/asterisk/asterisk/blob/master/res/res_pjsip_outbound_authenticator_digest.c#L153)

Ok, got your point here. If no auth hdr is generated/updated, returning any last error should do it I guess.

sauwming:
The latest patch was confirmed to work in an independent test by George Joseph (Asterisk).

nanangizz (approved):

trengginas (approved):

PR #2410, opened 2020-04-27
Suppressed various compilation warnings

sauwming:
- conversion from string literal to 'char *' warning
- format-truncation warning

### DIFF
--- a/pjlib/include/pj/string.h
+++ b/pjlib/include/pj/string.h
@@ -69,11 +69,17 @@ PJ_BEGIN_DECL
  *  - @ref page_pjlib_string_test
  */
 
-/* Check if a string is truncated and if yes, put a suffix of ".."
+/**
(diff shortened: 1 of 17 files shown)


pjlib/include/pj/string.h
```diff
@@ -69,6 +69,12 @@ PJ_BEGIN_DECL
  *  - @ref page_pjlib_string_test
  */
 
+/* Check if a string is truncated and if yes, put a suffix of ".."
```
nanangizz, line 72: start with double asterisks for doxygen?

pjlib/include/pj/string.h
```diff
@@ -69,6 +69,12 @@ PJ_BEGIN_DECL
...
+ * to indicate the truncation.
+ */
```
nanangizz, line 74: also mention that the macro is for checking truncated string in pj_ansi_snprintf()?

PR #2411, opened 2020-04-28
Dispatch semaphores on Darwin (#2351)

### LINKED ISSUE #2351: Semaphores in sandboxed Mac app

eofster:
Dispatch semaphore is a more modern semaphore implementation on Apple platforms. Apple claims it is more efficient. A call down to kernel is only needed when the calling thread needs to be blocked.

Dispatch semaphores work in sandboxed Mac apps, but POSIX semaphores don’t.

Closes #2351

### DIFF
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -804,6 +804,19 @@
 #  define PJ_HAS_SEMAPHORE	    1
 #endif
 
+/**
+ * Use dispatch semaphores on Darwin.
+ *
+ * Default: 1 on Darwin, 0 otherwise
+ */
+#ifndef PJ_SEMAPHORE_USE_DISPATCH_SEM
+#   if defined(PJ_DARWINOS)
+#	define PJ_SEMAPHORE_USE_DISPATCH_SEM	1
+#   else
+#	define PJ_SEMAPHORE_USE_DISPATCH_SEM	0
+#   endif
+#endif
+
 
 /**
  * Event object (for synchronization, e.g. in Win32)
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -810,7 +810,7 @@
  * Default: 1 on Darwin, 0 otherwise
  */
 #ifndef PJ_SEMAPHORE_USE_DISPATCH_SEM
-#   if defined(PJ_DARWINOS)
+#   if defined(PJ_DARWINOS) && PJ_DARWINOS != 0
 #	define PJ_SEMAPHORE_USE_DISPATCH_SEM	1
 #   else
 #	define PJ_SEMAPHORE_USE_DISPATCH_SEM	0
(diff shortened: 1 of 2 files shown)


eofster:
Are there any tests to run that touch semaphore code? For example, I don’t think `pj_sem_trywait()` has ever been called in my manual testing by running the app using the PJSUA API.

sauwming:
For reference, here's Apple's doc regarding [dispatch_semaphore](https://developer.apple.com/documentation/dispatch/dispatch_semaphore_t?language=objc).

I tested here on Mac OS and iOS and it worked fine.

There's pjlib-test for semaphore, albeit a very simple one, i.e. only trying to call the semaphore APIs (including `trywait`) and checking the return value.

pjlib/include/pj/config.h
```diff
@@ -804,6 +804,19 @@
 #  define PJ_HAS_SEMAPHORE	    1
 #endif
 
+/**
+ * Use dispatch semaphores on Darwin.
+ *
+ * Default: 1 on Darwin, 0 otherwise
+ */
+#ifndef PJ_SEMAPHORE_USE_DISPATCH_SEM
+#   if defined(PJ_DARWINOS)
```
nanangizz, line 813: a minor, for safety and consistency, please update to `#if defined(PJ_DARWINOS) && PJ_DARWINOS!=0`

PR #2415, opened 2020-05-01
Update sample app for video calls for Qt5

cristeab:


sauwming:
Tested on Mac with Qt 5.14.2

PR #2419, opened 2020-05-04
Fixed assertion when receving an INVITE response during an UPDATE

### LINKED ISSUE #2413: Assertion failed: (neg->state == PJMEDIA_SDP_NEG_STATE_DONE)
**Describe the bug**
When making a call to some toll free numbers using Asterisk 16.9.0, while trying to bridge the call, I PJ crashes on the client side:

Assertion failed: (neg->state == PJMEDIA_SDP_NEG_STATE_DONE), function pjmedia_sdp_neg_modify_local_offer2, file ../src/pjmedia/sdp_neg.c, line 316.

**To Reproduce**

1. run pjsua app and register with asterisk
2. make a single call: "sip:8003161044@myasteriskserver.domain.tld"
3. app crashes while trying to bridge the audio

**Expected behavior**
I would expect the call to complete allow me to talk.

**Desktop/Smartphone (please complete the following information):**

- macOS 10.15.3
- PJSIP version: 2.10
 
aconfigure --prefix=/Users/franke/pjproject --exec-prefix=/Users/franke/pjproject --disable-video --disable-ssl

**Additional context**
If I comment out the PJ_ASSERT_RETURN the call completes successfully.

sauwming:
Fixed #2413 .

This issue happens because we're receiving a response to INVITE in the middle of UPDATE when dialog is in early state, i.e. as in [RFC 3311 call flow example](https://tools.ietf.org/html/rfc3311#section-8), but (9) occurs after (5) instead.

Since we have just sent an UPDATE, neg->state is at `PJMEDIA_SDP_NEG_STATE_LOCAL_OFFER`, which causes the assertion `(neg->state == PJMEDIA_SDP_NEG_STATE_DONE)` when processing the INVITE response.

`
frame #4: 0x0000000100164243 pjsua-x86_64-apple-darwin19.3.0pjmedia_sdp_neg_modify_local_offer2.cold.1 + 35 frame #5: 0x0000000100091448 pjsua-x86_64-apple-darwin19.3.0pjmedia_sdp_neg_modify_local_offer2 + 632
frame #6: 0x000000010003ceb3 pjsua-x86_64-apple-darwin19.3.0inv_check_sdp_in_incoming_msg + 499 frame #7: 0x000000010003dde2 pjsua-x86_64-apple-darwin19.3.0inv_on_state_early + 274
`

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -2061,6 +2061,15 @@ static pj_status_t inv_check_sdp_in_incoming_msg( pjsip_inv_session *inv,
 	{
 	    const pjmedia_sdp_session *reoffer_sdp = NULL;
 
+	    if (pjmedia_sdp_neg_get_state(inv->neg) !=
+	    	PJMEDIA_SDP_NEG_STATE_DONE)
+	    {
+	    	PJ_LOG(4,(inv->obj_name, "Ignoring %s response "
+		          "because SDP negotiation is in progress",
+		          (st_code/10==18? "early" : "final" )));
+	    	return PJ_EINVALIDOP;
+	    }
+
 	    PJ_LOG(4,(inv->obj_name, "Received %s response "
 		      "after SDP negotiation has been done in early "
 		      "media. Renegotiating SDP..",
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -2064,10 +2064,10 @@ static pj_status_t inv_check_sdp_in_incoming_msg( pjsip_inv_session *inv,
 	    if (pjmedia_sdp_neg_get_state(inv->neg) !=
 	    	PJMEDIA_SDP_NEG_STATE_DONE)
 	    {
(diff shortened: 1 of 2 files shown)


sauwming:
The PR includes a SIPP scenario, which will be reverted before merging.

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -2061,6 +2061,15 @@ static pj_status_t inv_check_sdp_in_incoming_msg( pjsip_inv_session *inv,
 	{
 	    const pjmedia_sdp_session *reoffer_sdp = NULL;
 
+	    if (pjmedia_sdp_neg_get_state(inv->neg) !=
+	    	PJMEDIA_SDP_NEG_STATE_DONE)
+	    {
+	    	PJ_LOG(4,(inv->obj_name, "Ignoring %s response "
+		          "because SDP negotiation is in progress",
+		          (st_code/10==18? "early" : "final" )));
+	    	return PJ_EINVALIDOP;
```
nanangizz, line 2070: - shouldn't it be `Ignoring message body in %s response` instead of `Ignoring %s response`?
- in the `else` block (which also ignoring message body), it returns `PJ_SUCCESS`, any particular reason it is not the case here?

Issue #2414, opened 2020-04-30
Deadlock between  PJSUA_LOCK and event manager mutex


sauwming:
Reported by Антон Шудяев via PJSIP mailing list.

Stack trace:
```
Thread 1 (Thread 0x7fb2d05cb700 (LWP 2016)):
#2  0x00007fb2f3440edb in pj_mutex_lock (mutex=0x66991b5cf6b8) at ../src/pj/os_core_unix.c:1281
#3  0x00007fb2f339e546 in pjmedia_event_subscribe (mgr=0x66991b5cedc8, mgr@entry=0x0, cb=cb@entry=0x7fb2f33b9690 <vidstream_event_cb>, user_data=user_data@entry=0x66991af85c18, epub=0x7fb28804c928) at ../src/pjmedia/event.c:255
#4  0x00007fb2f33baaa4 in pjmedia_vid_port_create (pool=0x66991af85b70, prm=prm@entry=0x7fb2d05c4aa0, p_vid_port=p_vid_port@entry=0x7fb2f39ae820 <pjsua_var+2355296>) at ../src/pjmedia/vid_port.c:606
#5  0x00007fb2f3353eb5 in create_vid_win (type=type@entry=PJSUA_WND_TYPE_PREVIEW, fmt=0x7fb28842cc50, rend_id=4, cap_id=0, show=show@entry=0, wnd_flags=<optimized out>, wnd=0x0, id=0x7fb2d05c54fc) at ../src/pjsua-lib/pjsua_vid.c:742
#6  0x00007fb2f335457b in setup_vid_capture (call_med=call_med@entry=0x7fb2f3781f90 <pjsua_var+75728>) at ../src/pjsua-lib/pjsua_vid.c:940
#7  0x00007fb2f3354b21 in pjsua_vid_channel_update (call_med=call_med@entry=0x7fb2f3781f90 <pjsua_var+75728>, tmp_pool=0x66991e94b340, si=si@entry=0x7fb2d05c5fc0, local_sdp=local_sdp@entry=0x66991ec52be8, remote_sdp=remote_sdp@entry=0x7fb288421dc8) at ../src/pjsua-lib/pjsua_vid.c:1195
#8  0x00007fb2f3346e1a in pjsua_media_channel_update (call_id=<optimized out>, local_sdp=0x66991ec52be8, remote_sdp=0x7fb288421dc8) at ../src/pjsua-lib/pjsua_media.c:3501
[...]

PR #2421, opened 2020-05-04
Potential deadlock when app invokes PJSUA API from format changed event callback

nanangizz:
The event is published under video conference's lock and PJSUA API usually acquires PJSUA lock. If at the same time another thread is invoking PJSUA API that accesses video conference, such as `pjsua_vid_preview_start()` or `pjsua_call_set_vid_strm()`, deadlock may occur.

Not really sure why the `PJMEDIA_EVENT_FMT_CHANGED` event is currently published synchronously. Perhaps it wants the renderer/video-conf (the component downstream) to readjust format info and any frame buffer size before the frame being fetched is sent to it, to avoid possible buffer overflow for example, so changing it to be published asynchronously is perhaps not a good idea.

However, after investigation, for H264 where format change event is usually occurred, video stream's format is initialized with the highest size possible based on profile level ID, and any downstream components (e.g: video conf, renderer) will also initialize its format (and frame buffer size) based on the video stream's format, and format changed event should only reduce the video size, so buffer overflow should not happen. Moreover, currently PJSUA seem to only reconnect video conf connections (which actually does not readjust buffer sizes) upon format changed event.

After some tests, publishing the event asynchronously do not seem to cause any problem (crash, bad video rendering, etc).

### DIFF
--- a/pjmedia/src/pjmedia/vid_stream.c
+++ b/pjmedia/src/pjmedia/vid_stream.c
@@ -1527,7 +1527,8 @@ static pj_status_t get_frame(pjmedia_port *port,
 			stream->dec : stream->enc,
 		       "changed");
 
-	pjmedia_event_publish(NULL, port, &stream->fmt_event, 0);
+	pjmedia_event_publish(NULL, port, &stream->fmt_event,
+			      PJMEDIA_EVENT_PUBLISH_POST_EVENT);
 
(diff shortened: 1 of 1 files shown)


sauwming:
This should fix issue #2414.

nanangizz:
Alternative approach is to post PJSUA callback `on_call_media_event()` to a separate thread, e.g: worker thread via timer.

Issue #1941, opened 2016-07-04
Review iOS 10 integration to PJSIP


pjsipbot:
_2016-07-04 05:51:22: @nanangizz_ _created the issue on trac ticket [1941](https://trac.pjsip.org/repos/ticket/1941)_

iOS 10 is coming with new features such as CallKit, framework changes such as in AVFoundation, and also API deprecations. This ticket will review and apply any integrations needed in PJSIP library.

== PushKit guide==

What your application needs to do:
1. Set up PushKit according to [Apple's doc](https://developer.apple.com/library/content/documentation/Performance/Conceptual/EnergyGuide-iOS/OptimizeVoIP.html).
2. Upon startup, request for push tokens.
3. Send REGISTER to SIP server with the push token info, the info may be added as a specific header, Contact header params, or Contact URI params, to tell the SIP server about the PN (Push Notification) info (server URI, tokens, etc). You can use `pjsua_acc_config.reg_hdr_list`, `pjsua_acc_config.contact_uri_params`, or `pjsua_acc_config.reg_contact_params` (the later is available in ticket #1965).<BR />Note: [​RFC 8599: Push Notifications in the Session Initiation Protocol (SIP)](https://tools.ietf.org/html/rfc8599) suggests to put it in Contact URI params, for example:
```
Contact: <sip:alice@alicemobile.example.com;
     pn-type=acme;pn-methods="INVITE";
     pn-uri="https://pn.acme.example.com/ZTY4ZDJlMzODE1NmUgKi0K">
```
4. After a successful registration, SIP client can go background and will be woken up by the OS upon receiving push notification.
Note: starting from iOS 13, there's a new requirement:
```Apps receving VoIP pushes must post an incoming call (via CallKit or IncomingCallNotifications) in the same run loop as   pushRegistry:didReceiveIncomingPushWithPayload:forType:[withCompletionHandler:] without delay. Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Killing app because it never posted an incoming call to the system after receiving a PushKit VoIP push callback.'```
In order to make it work with the normal SIP flow which may require you to wait for some time to receive the INVITE message, please look at Apple's recommendation in its [developer forum](https://forums.developer.apple.com/thread/117939).

Configure your SIP server:
1. Parse the PN info in registration.
2. Upon receiving an incoming SIP INVITE, SIP server should contact PN server as specified via PN URI and tokens.
[...]

PR #2422, opened 2020-05-05
Fix short noise at the beginning of call when audio device started later.

nanangizz:
Reported that short clicking noise right was heard on callee's side after the call is connected after implementing iOS 13 support as outlined in issue #1941.

After investigation, the click noise seems to be caused by JB discard to adjust latency as the provided log shows a regular interval of missing frames and the codec PLC may not perform well (in maintaining audio signal continuation in back-to-back frames generated by decoder-PLC-decoder). And some tests showed that resetting jitter buffer does solve the clicking noise issue.

The patch introduces a new feature in stream to automatically reset jitter buffer in the beginning of stream playback (stream's `get_frame()`) and return silent audio frames to allow a short prefetching effect to avoid empty frame returned by just resetted jitter buffer in the next immediate `get_frame()`.

Thanks to Marcus Froeschl for the report.

### DIFF
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -1428,6 +1428,24 @@
 
 
 /**
+ * Reset jitter buffer and return silent audio on stream playback start
+ * (first get_frame()). This is useful to avoid possible noise that may be
+ * introduced by discard algorithm and neutralize latency when audio device
+ * is started later than the stream.
+ *
+ * Set this to N>0 to allow N silent audio frames returned on stream playback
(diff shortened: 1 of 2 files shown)


sauwming:
How is the default value (1) selected? Is one frame enough to get rid of the clicking noise, which, I assume, may/should last longer than that?

nanangizz:
The suspect of the cause of the clicking noise is signal continuation (just updated the PR desc), not really the PLC (also PLC should generate silent in the beginning of decoding), one frame should be enough to mimic the normal scenario (where the issue does not occur): audio device started about the same time as the stream.

PR #2423, opened 2020-05-05
Avoid undefinition of standard macros s_addr and s6_addr

### LINKED ISSUE #2311: PJSIP undefine standard UNIX macros s_addr and s6_addr

sauwming:
To fix #2311. 

There's a different approach suggested in PR #2313, but it has the drawbacks of rather major API break as well as portability issue .

Thanks to @traud for suggesting this approach instead.

Tested on Windows, Mac, Ubuntu, iOS, and Android (there is a possibility this change may break PJSIP build on some other platform, though).
Passes pjlib and pjlib-util-test.

Note: minor API break: `struct pj_in_addr` no longer exists.

### DIFF
--- a/pjlib-util/src/pjlib-util-test/resolver_test.c
+++ b/pjlib-util/src/pjlib-util-test/resolver_test.c
@@ -37,6 +37,9 @@
 #define ACTION_IGNORE	-1
 #define ACTION_CB	-2
 
+#undef s6_addr32
+#define s6_addr32(addr, idx) *((pj_uint32_t *)(addr.s6_addr + idx*4))
+
 static struct server_t
 {
     pj_sock_t	     sock;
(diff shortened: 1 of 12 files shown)


nanangizz:
Error compile on MSVC2005:
`c:\program files (x86)\microsoft visual studio 8\vc\platformsdk\include\wspiapi.h(44) : error C2894: templates cannot be declared to have 'C' linkage`
After moving `#include <pj/compat/socket.h>` to outside of `PJ_BEGIN_DECL` block, e.g: right after `#include <pj/types.h>`, the error disappears.

Issue #2424, opened 2020-05-05
Error parsing DTMF signal and duration in INFO message


Truong-Thanh-Quang:
pjsip\src\pjsua-lib\pjsua_call.c -> pjsua_call_on_tsx_state_changed():
When parsing DTMF in INFO message, it compares exactly "Signal=" and "Duration=". 
In many cases, there is one space before/after equal sign.
For example:
    _Signal= #
    Duration= 120_
Theresult in that cases, digit= (space), duration=0.

PR #2425, opened 2020-05-05
Fix dtmf signal &duration parsing in dtmf-relay INFO message

Truong-Thanh-Quang:


Truong-Thanh-Quang:
https://github.com/pjsip/pjproject/issues/2424

nanangizz:
setting this PR as `enhancement` instead of `bug` because the [draft](https://tools.ietf.org/html/draft-kaplan-dispatch-info-dtmf-package-00#section-5.2) actually says:
```
   Dtmf-relay-body = digit-line CRLF duration-line

   digit-line      = "Signal" EQUAL SP button
   button          = DIGIT / "A" / "B" / "C" / "D" / "*" / "#"

   duration-line   = "Duration" EQUAL SP msecs
   msecs           = 1*4(DIGIT)  ;100-5000 millisecs
```

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -5576,13 +5576,18 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
...
 		    char *val;
 
 		    val = pj_strstr(&input, &STR_SIGNAL);
 		    if (val) {
-			info.digit = *(val+STR_SIGNAL.slen);
+				pj_ssize_t idx1 = val - input.ptr;
+				pj_ssize_t idx2 = 0;
+				while ( (*(val + STR_SIGNAL.slen + idx2) == 32 || *(val + STR_SIGNAL.slen + idx2) == 61) &&
+					(idx1 + idx2 < input.slen))
+					idx2++;
```
nanangizz, line 5589: unaligned indentation? note: PJSIP uses 8 chars tab-size, please check [this](https://docs.pjsip.org/en/latest/consider.html#coding-style) for more info.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -5576,13 +5576,18 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
...
+		    const pj_str_t STR_DURATION = { "Duration", 8 };
 		    char *val;
 
 		    val = pj_strstr(&input, &STR_SIGNAL);
 		    if (val) {
-			info.digit = *(val+STR_SIGNAL.slen);
+				pj_ssize_t idx1 = val - input.ptr;
+				pj_ssize_t idx2 = 0;
+				while ( (*(val + STR_SIGNAL.slen + idx2) == 32 || *(val + STR_SIGNAL.slen + idx2) == 61) &&
+					(idx1 + idx2 < input.slen))
```
nanangizz, line 5588: shouldn't loop condition `(idx1 + idx2 < input.slen)` include `STR_SIGNAL.slen`, i.e: `(idx1 + STR_SIGNAL.slen + idx2 < input.slen)`?
or alternatively, initialize `idx1` with `val - input.ptr + STR_SIGNAL.slen`?

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -5576,13 +5576,18 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
...
+					idx2++;
+			info.digit = *(val+STR_SIGNAL.slen + idx2);
```
nanangizz, line 5590: what if there is no `=` in the input string?

Truong-Thanh-Quang:
> setting this PR as `enhancement` instead of `bug` because the [draft](https://tools.ietf.org/html/draft-kaplan-dispatch-info-dtmf-package-00#section-5.2) actually says:
> 
> ```
>    Dtmf-relay-body = digit-line CRLF duration-line
> 
>    digit-line      = "Signal" EQUAL SP button
>    button          = DIGIT / "A" / "B" / "C" / "D" / "*" / "#"
> 
>    duration-line   = "Duration" EQUAL SP msecs
>    msecs           = 1*4(DIGIT)  ;100-5000 millisecs
> ```

In the draft, there is one space after equal sign. The code takes this space for digit.
Same for duration-line.

nanangizz:
> > setting this PR as `enhancement` instead of `bug` because the [draft](https://tools.ietf.org/html/draft-kaplan-dispatch-info-dtmf-package-00#section-5.2) actually says:
> > ```
> >    Dtmf-relay-body = digit-line CRLF duration-line
> > 
> >    digit-line      = "Signal" EQUAL SP button
> >    button          = DIGIT / "A" / "B" / "C" / "D" / "*" / "#"
> > 
> >    duration-line   = "Duration" EQUAL SP msecs
> >    msecs           = 1*4(DIGIT)  ;100-5000 millisecs
> > ```
> 
> In the draft, there is one space after equal sign. The code takes this space for digit.
> Same for duration-line.

Agree, I think it is a bug.

pjsip/src/pjsua-lib/pjsua_call.c
Truong-Thanh-Quang: I have thought of this case. But I don't know what to return in case of error format.

pjsip/src/pjsua-lib/pjsua_call.c
Truong-Thanh-Quang: Yes, It is better.

pjsip/src/pjsua-lib/pjsua_call.c
nanangizz: In case of error format, I think we should not call the `on_dtmf_digit2()` callback (and leave `is_handled` to false so 400 response will be sent) and write some log info?

> I have thought of this case. But I don't know what to return in case of error format.

Truong-Thanh-Quang:
I've just commited changes.

https://github.com/Truong-Thanh-Quang/pjproject/commit/9f72732a54e7e58aea1deba654804b76e2fc099d

- Reformat regarding the coding style
- Check equal sign existence
- Write log if error occured

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -5576,26 +5576,54 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
...
+		    pj_ssize_t count_equal_sign;
 
 		    val = pj_strstr(&input, &STR_SIGNAL);
 		    if (val) {
-			info.digit = *(val+STR_SIGNAL.slen);
-			is_handled = PJ_TRUE;
+			idx = 0;
+			count_equal_sign = 0;
+			while ((*(val + STR_SIGNAL.slen + idx) == 32 || *(val + STR_SIGNAL.slen + idx) == 61) &&
+			    (val - input.ptr + STR_SIGNAL.slen + idx < input.slen)) {
```
nanangizz, line 5590: Looks much better now, just few minor suggestions.

Perhaps better to rearrange it like this:
```
while ((val - input.ptr + STR_SIGNAL.slen + idx < input.slen) && 
      (*(val + STR_SIGNAL.slen + idx) == 32 || *(val + STR_SIGNAL.slen + idx) == 61))
{
```
The updates are:
- coding style: align multiline condition, new line for bracket `{` when condition is long/multiline
- better check validity first before accessing/evaluating the pointer content.

Furthermore, as `val + STR_SIGNAL.slen + idx` seems to be repeatedly used, perhaps it'll be more readable to have a new temporary var, e.g:
```
char *p = val + STR_SIGNAL.slen;
while ((p - input.ptr < input.slen) &&  (*p == ' ' || *p == '=')) {
   ...
   ++p;
}
...
info.digit = *p;
```

Truong-Thanh-Quang: I've updated as your suggestion. Thanks!
https://github.com/Truong-Thanh-Quang/pjproject/commit/838091b37926032546f5f3fe32ff4fa2523b3b58

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -5576,26 +5576,51 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
...
+			        val_str.ptr = p;
+			        val_str.slen = input.slen - (p - input.ptr);
+			        info.duration = pj_strtoul(&val_str);
+			    } else {
+				is_handled = PJ_FALSE;
+				PJ_LOG(2, (THIS_FILE, "Invalid dtmf-relay format"));
+			    }
 			}
 		    	info.method = PJSUA_DTMF_METHOD_SIP_INFO;
 			(*pjsua_var.ua_cfg.cb.on_dtmf_digit2)(call->index,
```
nanangizz, line 5626: missed this earlier, perhaps invoke the callback and send 200 response only when `is_handled` is true, or otherwise send 400 response?

Truong-Thanh-Quang: Oh, big mistake.
Check is_handled before calling callback.
https://github.com/Truong-Thanh-Quang/pjproject/commit/757199fb6ae91fa5857142007e53aa6697329994

PR #2429, opened 2020-05-13
Fixed assertion in WebRTC AEC if clock rate > 16000

### LINKED ISSUE #2404: Assertion in WebRTC AEC on desktop when clock rate is higher than 16000 Hz
Assertion line:
```../../webrtc/src/webrtc//modules/audio_processing/aec/aec_core.c:1765: WebRtcAec_ProcessFrames: Assertion `aec->num_bands == num_bands' failed.```

1. run pjsua app with parameter: `--snd-clock-rate=48000`
1. start sound device, e.g: `cc 0 0`
1. assertion occurs!

After investigation, there seem to be a bug in our AEC wrapper, it wrongly passes channel count to `WebRtcAec_Process()`, while it actually expects number of bands (number of bands is not channel count). The WebRTC expects that audio samples splitted into high bands and low band when clock rate is higher than 16000 Hz. The audio band splitting can be done via WebRTC [AudioBuffer](https://chromium.googlesource.com/external/webrtc/+/refs/heads/master/modules/audio_processing/audio_buffer.h).

Thanks to Marcus Froeschl, Thomas Hackl, and Arkadiusz Wroński for the reports.

### PR DESCRIPTION
**Pull request by sauwming**
Fixed #2404 .

Another approach that uses `SplittingFilter`, which doesn't work, can be found in #2428.

### DIFF
--- a/pjmedia/src/pjmedia/echo_webrtc.c
+++ b/pjmedia/src/pjmedia/echo_webrtc.c
@@ -59,7 +59,7 @@
 #endif
 
 #define NUM_BANDS		3
-#define NUM_SAMPLES		80
+#define NUM_SAMPLES		160
 #define BUF_LEN			NUM_SAMPLES * NUM_BANDS
 
 /* Set this to 0 to disable metrics calculation. */
@@ -324,7 +324,7 @@ PJ_DEF(pj_status_t) webrtc_aec_cancel_echo( void *state,
 #if PJMEDIA_WEBRTC_AEC_USE_MOBILE
         status = WebRtcAecm_Process(echo->AEC_inst, &rec_frm[frm_idx],
         			    (echo->NS_inst? buf_ptr: NULL),
-        			    out_buf_ptr, FRAME_LEN,
+        			    out_buf_ptr, NUM_SAMPLES,
         			    echo->tail);
 #else
         {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pjmedia/src/pjmedia/echo_webrtc.c [hunk 1]
```diff
@@ -72,6 +74,7 @@ typedef struct webrtc_ec
     unsigned	tail;
     unsigned    clock_rate;
     unsigned	channel_count;
+    unsigned	num_bands;
     unsigned    subframe_len;
     sample      tmp_buf[BUF_LEN];
     sample      tmp_buf2[BUF_LEN];
```
nanangizz, line 80: perhaps we can apply some optimization to allocate `tmp_buf & tmp_buf2` dynamically, i.e: allocate `(echo->num_bands*NUM_SAMPLES)`, instead of `NUM_BANDS*NUM_SAMPLES`, as usually `echo->num_bands` should be 1 as default clock rate is 16khz?

nanangizz:
assertion has been fixed. now, any info about how the AEC performing on >16khz clock rate?

sauwming:
Putting this on hold for now. Further testing shows that this patch doesn't work either. Not sure how it was fine the day before, but all the git branch switches must have corrupted the build.

mfroeschl:
@sauwming Just to let you know, we were also testing this PR and for us it completely destroyed audio quality when using WebRTC on win32 platform (horrible stutter & noise).

Please let us know when there is an update.

Issue #1725, opened 2014-01-24
ACK is not sent upon receiving 200/OK retransmission if re-INVITE is sent


pjsipbot:
_2014-01-24 09:04:23: @bennylp_ _created the issue on trac ticket [1725](https://trac.pjsip.org/repos/ticket/1725)_

ACK is not sent when receiving INVITE 200/OK response, if a re-INVITE is sent after the first ACK is sent. Scenario:

```
-----> INVITE    ---->
<----- 200/OK    -----
------ ACK       ---->  (lost)

-----> re-INVITE --->

<----- 200/OK    -----  (retransmission)

(no ACK is sent)

```

___
_2014-01-24 11:28:36: @bennylp_ _changed **status** from **new** to **closed**_

___
_2014-01-24 11:28:36: @bennylp_ _set **resolution** to **fixed**_

___
_2014-01-24 11:28:36: @bennylp_ _commented_

In [r4716](../commit/b02eb708c8e56d385102737083615382ece9c127):
Fixed #1725: ACK is not sent upon receiving 200/OK retransmission if re-INVITE is sent

PR #2432, opened 2020-05-15
Possible leak tdata when receiving 200 OK response after the invite session is destroyed.

trengginas:
Consider this scenario:
```
-------> INVITE [1]
<------  200 OK INVITE [2]
-------> ACK [3]
-------> BYE [4]
<------  200 OK BYE [5]
<------  200 OK INVITE [6]
-------> ACK [7]
Dialog destroyed [8]
```
On normal condition, tdata created to send ACK [3] will be released when INVITE session was destroyed [5]. 
However If a 200 OK INVITE response was received [6] before dialog is destroyed, then the tdata created to send ACK [7] will not be released.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -503,6 +503,10 @@ static pj_status_t inv_send_ack(pjsip_inv_session *inv, pjsip_event *e)
     if (inv->state < PJSIP_INV_STATE_CONFIRMED) {
 	inv_set_state(inv, PJSIP_INV_STATE_CONFIRMED, &ack_e);
     } else if (inv->state == PJSIP_INV_STATE_DISCONNECTED) {
+        /* Avoid possible leaked tdata when invite session is already
+         * destroyed.
+         * https://github.com/pjsip/pjproject/pull/2432
+         */
         pjsip_tx_data_dec_ref(inv->last_ack);
         inv->last_ack = NULL;
     }
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -502,6 +502,9 @@ static pj_status_t inv_send_ack(pjsip_inv_session *inv, pjsip_event *e)
      */
     if (inv->state < PJSIP_INV_STATE_CONFIRMED) {
 	inv_set_state(inv, PJSIP_INV_STATE_CONFIRMED, &ack_e);
+    } else if (inv->state == PJSIP_INV_STATE_DISCONNECTED) {
+        pjsip_tx_data_dec_ref(inv->last_ack);
+        inv->last_ack = NULL;
     }
 
     return PJ_SUCCESS;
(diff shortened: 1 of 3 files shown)


pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -502,6 +502,9 @@ static pj_status_t inv_send_ack(pjsip_inv_session *inv, pjsip_event *e)
      */
     if (inv->state < PJSIP_INV_STATE_CONFIRMED) {
 	inv_set_state(inv, PJSIP_INV_STATE_CONFIRMED, &ack_e);
+    } else if (inv->state == PJSIP_INV_STATE_DISCONNECTED) {
```
nanangizz, line 505: please put some info/comment or simply a link to this PR for future ref.

btw, the issue is reproducible and the patch does fix it, right?
also, I've found another scenario that may lead to leaked tdata, will add more info & perhaps patch later.

nanangizz:
Found another scenario that may lead to leaked tdata:
1. initiates a call
2. receives `200 response` for the initial `INVITE`
3. sends `ACK`, but somehow it is not delivered within ~32 secs (tsx lifetime)
4. when `ACK` is actually sent, app does not receives sent notification which causing `ACK tdata` ref cnt not decremented.

This is perhaps also the scenario occurred in [this report](http://lists.pjsip.org/pipermail/pjsip_lists.pjsip.org/2019-September/041490.html).

Initial patch [here](https://github.com/pjsip/pjproject/files/4641724/tp-send-timeout-not-notify-app.diff.txt).

pjsip/src/pjsip-ua/sip_inv.c
trengginas: Yes, it's reproducible and the patch fix it.

trengginas:
Do you want to incorporate the patch on this PR?

nanangizz:
> Do you want to incorporate the patch on this PR?

yes, but please do some further checks first, make sure it is not harmful.

sauwming:
Btw, do we need to send ACK for the 2nd 200/OK for the INVITE? Perhaps the fix should be to avoid sending one?

i.e. If the scenario is like this instead, what should happen?
-------> INVITE [1]
<------  200 OK INVITE [2]
-------> ACK [3]
<------  200 OK INVITE [4]

trengginas:
> Btw, do we need to send ACK for the 2nd 200/OK for the INVITE? Perhaps the fix should be to avoid sending one?
> 
> i.e. If the scenario is like this instead, what should happen?
> -------> INVITE [1]
> <------ 200 OK INVITE [2]
> -------> ACK [3]
> <------ 200 OK INVITE [4]

https://github.com/pjsip/pjproject/issues/1725 is to handle this scenario

trengginas:
> tcp_flush_pending_tx

Tested here sending INVITE and REGISTER, both will leak tdata without the patch.

nanangizz:
> > tcp_flush_pending_tx
> 
> Tested here sending INVITE and REGISTER, both will leak tdata without the patch.

and with the patch, no leak and no crash/assert/other-bad-thing?

trengginas:
> > > tcp_flush_pending_tx
> > 
> > 
> > Tested here sending INVITE and REGISTER, both will leak tdata without the patch.
> 
> and with the patch, no leak and no crash/assert/other-bad-thing?

No issue (assert/crash) found. Pass pjsip-test also

Issue #2263, opened 2020-01-28
Add callback to notify when first valid pair is found on ICE negotiation


pjsipbot:
_2020-01-28 03:50:31: @trengginas_ _created the issue on trac ticket [2263](https://trac.pjsip.org/repos/ticket/2263)_

This ticket will add callback to notify when first valid pair is found on ICE negotiation.
Thanks to Ben Ford for the patch.
___
_2020-01-28 03:51:52: @trengginas_ _changed **status** from **new** to **closed**_

___
_2020-01-28 03:51:52: @trengginas_ _set **owner** to **@trengginas**_

___
_2020-01-28 03:51:52: @trengginas_ _set **resolution** to **fixed**_

___
_2020-01-28 03:51:52: @trengginas_ _commented_

In [r6139](../commit/71ff0065108470602af063d631e0affad38efc04):
Close #2263: Add callback to notify when first valid pair is found on ICE negotiation.

PR #2433, opened 2020-05-15
ice_stran: add on_valid_pair() to callbacks

257:
Hello,
This is related to ```71ff0065108470602af063d631e0affad38efc04```.

### DIFF
--- a/pjnath/include/pjnath/ice_strans.h
+++ b/pjnath/include/pjnath/ice_strans.h
@@ -182,6 +182,14 @@ typedef struct pj_ice_strans_cb
 			    pj_ssize_t sent);
 
     /**
+     * An optional callback that will be called by the ICE transport when a
+     * valid pair has been found during ICE negotiation.
+     *
+     * @param ice_st	    The ICE stream transport.
+     */
+    void (*on_valid_pair)(pj_ice_strans *ice_st);
+
+    /**
      * Callback to report status of various ICE operations.
      * 
      * @param ice_st	    The ICE stream transport.
--- a/pjnath/src/pjnath/ice_strans.c
+++ b/pjnath/src/pjnath/ice_strans.c
@@ -89,6 +89,7 @@ static pj_uint8_t srflx_pref_table[PJ_ICE_CAND_TYPE_MAX] =
 
 
 /* ICE callbacks */
+static void	   on_valid_pair(pj_ice_sess *ice);
 static void	   on_ice_complete(pj_ice_sess *ice, pj_status_t status);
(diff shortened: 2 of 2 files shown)


sauwming:
This is related to issue #2263 .

It seems okay, but do you need the parameter op in the `on_valid_pair()` callback? (since it seems that it's always PJ_ICE_STRANS_OP_FIRST_CAND anyway?)

257:
> This is related to issue #2263 .
> 
> It seems okay, but do you need the parameter op in the `on_valid_pair()` callback? (since it seems that it's always PJ_ICE_STRANS_OP_FIRST_CAND anyway?)

Yeah, sticky. I don't have a perfect solution.

My thinking was that it would help applications that already have an ```on_ice_complete```; they can reuse the same function, possibly modifying it a bit, and react accordingly based on ```op```.

Here is an example of it I made for Jami here:
https://review.jami.net/c/ring-daemon/+/13902/7/src/ice_transport.cpp#513

sauwming:
I would prefer to remove op parameter from the callback since it doesn't seem to be necessary.

As for your application, if you want to use the same callback handler function, I believe you can do like this?
```
#define PJ_ICE_STRANS_OP_FIRST_CAND [some_value]

tr->onComplete(ice_st, PJ_ICE_STRANS_OP_FIRST_CAND, PJ_SUCCESS);
```

257:
> I would prefer to remove op parameter from the callback since it doesn't seem to be necessary.
> 
sure, updated in a rush; no mistakes, hopefully :)

btw, code style is diverging more and more. what do you think of a .clang-format? i can work on it if need be.

pjnath/include/pjnath/ice_strans.h
```diff
@@ -181,6 +181,15 @@ typedef struct pj_ice_strans_cb
     void    (*on_data_sent)(pj_ice_strans *sock,
 			    pj_ssize_t sent);
 
+    /**
+     * An optional callback that will be called by the ICE transport when a
+     * valid pair has been found during ICE negotiation.
+     *
+     * @param ice_st	    The ICE stream transport.
+     * @param status	    Operation status.
```
sauwming, line 189: remove this non-existent param

sauwming:
Not sure what you're referring to, but the coding style seems fine?

The patch looks okay, only a minor fix needed. Will also need to wait for others to review it as well.

pjnath/include/pjnath/ice_strans.h
257: oops! fixed.

PR #2434, opened 2020-05-18
Call in early state gets disconnected on IP change

nanangizz:
Add configurable setting for INVITE transaction to terminate or maintain the transaction on transport down (such as in IP change scenario). Currently a transport failure (either error or down) will terminate all transactions using that transport. And when the transaction is terminated, the INVITE session will be terminated too. While some applications prefer to be able to leave the call remaining active.

Caveat:
Some servers such as Kamailio, FreeSwitch may still try to route the response via the old/down transport instead of the new one (used by re-REGISTER).

### DIFF
--- a/pjsip/include/pjsip/sip_config.h
+++ b/pjsip/include/pjsip/sip_config.h
@@ -1019,6 +1019,17 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 #   define PJSIP_TSX_1XX_RETRANS_DELAY	60
 #endif
 
+/**
+ * Terminate INVITE transaction when TCP/TLS transport is disconnected.
+ * This setting does not apply for disconnection during sending. For initial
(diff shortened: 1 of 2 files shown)


sauwming:
So should we proceed with this PR? Considering that even though the call doesn't get disconnected, it's still unlikely to get through anyway because any answer from remote may still be routed to the old address.

nanangizz:
Right, perhaps let's just drop this PR for now, diff attached [here](https://github.com/pjsip/pjproject/files/4727857/2434.diff.txt) (branch may be deleted and just in case the patch will be needed in the future).

Also for future reference, some efforts have been done to update the server about our new address/connection:
- unregister the old contact and re-register a new contact
- send call UPDATE, to update Contact and kind of telling the server that new connection is established for the existing call

And on Kamailio, those do not seem to work.

Some forum discussion suggest to just cancel the call and create a new one.

Issue #2438, opened 2020-05-19
pjmedia_conf_disconnect_port affect active connections


half-left:
**Describe the bug**
Conference bridge is compiled as pjmedia_conf implementation. Call of pjmedia_conf_disconnect_port change level adjustment for another listeners of source port.

**To Reproduce**

1. make three calls
2. connect via pjsua_conf_connect2 media port of call 1 to media port of call 2 with level 0.0 
3. connect via pjsua_conf_connect2 media port of call 1 to media port of call 3 with level 1.0 
4. disconnect via pjsua_conf_disconnect media port of call 1 from media port of call 2 
5. call 3 doesn't hear call 1

**Expected behavior**
call 3 hears call 1

**Desktop/Smartphone (please complete the following information):**

- Debian 10
- PJSIP
  - version: 2.9

sauwming:
I believe you have created a patch for this in the mailing list. Would you be willing to submit a PR so we can review it? Thanks.

PR #2439, opened 2020-05-20
Fix bug: level is not removed form src_port->listener_adj_level in pjmedia_conf_disconnect_port

half-left:
Fix for https://github.com/pjsip/pjproject/issues/2438


sauwming:
Tested here and the patch works fine.

PR #2441, opened 2020-05-20
Update video port to publish media events from videodev streams asynchronously

half-left:
Fix for https://github.com/pjsip/pjproject/issues/2440


sauwming:
We also have another fix to the issue in #2446. Since the issue can potentially happen for other media events as well, we think it would be better to fix it in pjsua-lib instead, i.e. by calling the `on_call_media_event()` asynchronously.

PR #2446, opened 2020-05-22
Call on_call_media_event() callback asynchronously to avoid deadlock

sauwming:
This is to prevent issues such as #2414 and #2440 .

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -1513,6 +1513,20 @@ pj_status_t on_media_event(pjmedia_event *event, void *user_data)
     return status;
 }
 
+/* Call on_call_media_event() callback using timer */
+void call_med_event_cb(void *user_data)
+{
+    pjsua_event_list *eve = (pjsua_event_list *)user_data;
+    
+    (*pjsua_var.ua_cfg.cb.on_call_media_event)(eve->call_id,
+					       eve->med_idx,
+					       &eve->event);
+
+    pj_mutex_lock(pjsua_var.timer_mutex);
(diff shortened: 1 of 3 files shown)


pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -1624,14 +1638,29 @@ pj_status_t call_media_on_event(pjmedia_event *event,
...
-	if (call) {
-	    (*pjsua_var.ua_cfg.cb.on_call_media_event)(call->index,
-						       call_med->idx, event);
-	} else {
+	pjsua_event_list *eve = NULL;
+ 
+    	pj_mutex_lock(pjsua_var.timer_mutex);
+
+    	if (pj_list_empty(&pjsua_var.event_list)) {
+            eve = PJ_POOL_ALLOC_T(pjsua_var.pool, pjsua_event_list);
```
nanangizz, line 1646: after a quick check, access to `pjsua_var.pool` should be protected by PJSUA lock, but unfortunately locking PJSUA lock is the cause of the deadlock, so the lock must be avoided. A solution is to use a specific pool which can be shared with pjsua timer (note that the pjsua timer also uses `pjsua_var.pool`, so it is buggy, but it can simply use PJSUA lock without a risk of deadlock).

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -1513,6 +1513,20 @@ pj_status_t on_media_event(pjmedia_event *event, void *user_data)
     return status;
 }
 
+/* Call on_call_media_event() callback using timer */
+void call_med_event_cb(void *user_data)
```
nanangizz, line 1517: static?

Issue #2436, opened 2020-05-18
Duration of DTMF digits received via RTP


maxtruxa:
**Is your feature request related to a problem? Please describe.**
When receiving telephone events indicating DTMF digits via RTP, there is currently no way to determine the duration of the event. The DTMF callback of the stream is only called once when the first packet indicating the event is received. [Further packets are filtered](https://github.com/pjsip/pjproject/blob/a35a53b4d9da3fb95aa078170aa38b3eb1f43bbe/pjmedia/src/pjmedia/stream.c#L1716) to prevent duplicate indications of the same event (which is good obviously).
The downside of this behavior is that while the events duration might be continously extended via RTP, the application has no way of knowing that.

**Describe the solution you'd like**
**Solution 1:**
Add a new set of callbacks - similar to `on_dtmf_digit2` - throughout the stack that allows the application to get notified about "DTMF continuations".
**Solution 2:**
Add a flag (compile time or runtime?), indicating that the application expects the DTMF callback to be called everytime a new packet is received via RTP (effectively disabling the filtering done by RTP stream handling).

**Describe alternatives you've considered**
There is no alternative. We need some way to determine the duration of DTMF digits.

**Additional context**
Not really relevant, but:
Our application is more or less a PBX, sitting in front of other legacy (non-VoIP-aware) applications consuming the audio streams. The duration of DTMF digits is significant to those applications. Since the software has no concept of telehone-events, we are re-inserting the DTMF digits back into the audio stream.

I'm absolutely willing to work on this, but I don't want to invest time without knowing that a PR addressing this issue would be accepted.

sauwming:
It sounds like a good idea. Please go ahead.

PR #2447, opened 2020-05-25
Add callbacks to receive RTP DTMF events

maxtruxa:
First attempt at implementing #2436.

## Goal

Enable the application to determine the actual duration of an DTMF event in real-time.

I'm not completely happy with the abstraction provided but that was the best I could come up with, without violating the design constraints:
- It is obviously not possible to indicate the correct event duration right away, since it is not known.
- Buffering the event until the full duration is known and only then indicating the event is not feasible since it can cause huge delays - especially for longer DTMF digits.
- Fully managing the events in pjmedia is surprisingly hard and error-prone:
  When does an event end? RTP packets indicating the end of an event might be dropped. Using just a naive timer to bail out if no "end" packet is received can easily lead to false-positives, if event packets extending the event are dropped.
  The current approach leaves these decisions to the application.

## Overview

RTP events are basically just forwarded to the application using a new callback. Minimal state information is tracked by pjmedia and indicated to the application using flags:
- Is this an update? Events are considered an update, if the event has been indicated previously and this is just an update to the duration.
- Is this the end of the event? This information is directly derived from the "E" flag of the RTP payload. Packets with the "E" flag are typically sent multiple times to counter packet loss. Only the first packet with the "E" flag set results in an event indication, duplicates are absorbed by pjmedia.

## Changes

- pjmedia:
  - Added `struct pjmedia_stream_dtmf_event` for holding DTMF event information.
  - Added `pjmedia_stream_set_dtmf_event_callback()` for registering a callback to receive DTMF event information.
  - Modified `handle_incoming_dtmf()` to invoke callback registered by `pjmedia_stream_set_dtmf_event_callback()`. If it was not registered, there is no change in behavior and the old callback is invoked.
- pjsua-lib:
  - Added macro `PJSUA_UNKNOWN_DTMF_DURATION` to indicate that an event's duration is unknown.
  - Added `struct pjsua_dtmf_event` for holding DTMF event information (basically an extension of `pjsua_dtmf_info`).
  - Added callback `on_dtmf_event()` to `struct pjsua_callback`.
  - Modified `pjsua_aud_channel_update()` to register DTMF callback using `pjmedia_stream_set_dtmf_event_callback()` instead of `pjmedia_stream_set_dtmf_callback()` if callback `on_dtmf_event()` is set.
  - DTMF digits received via SIP INFO are passed to `on_dtmf_event()` as well (if set).
- pjsua2-lib
  - Add `pj::Call::onDtmfEvent` and `struct OnDtmfEventParam`.

_Edit: Changed the names of various identifiers as discussed below._

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_call.c
+++ b/pjsip/src/pjsua-lib/pjsua_call.c
@@ -5589,7 +5589,9 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
 	    pj_status_t status;
 	    pj_bool_t is_handled = PJ_FALSE;
 
-	    if (pjsua_var.ua_cfg.cb.on_dtmf_digit2) {
+	    if (pjsua_var.ua_cfg.cb.on_dtmf_digit2 ||
+                pjsua_var.ua_cfg.cb.on_dtmf_event)
+            {
 		pjsua_dtmf_info info;
 		pj_str_t delim, token, input;
 		pj_ssize_t found_idx;
@@ -5641,6 +5643,7 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
 			        val_str.slen = input.slen - (p - input.ptr);
 			        info.duration = pj_strtoul(&val_str);
 			    } else {
+                                info.duration = PJSUA_UNKNOWN_DTMF_DURATION;
 				is_handled = PJ_FALSE;
 				PJ_LOG(2, (THIS_FILE, "Invalid dtmf-relay format"));
 			    }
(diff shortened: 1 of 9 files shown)


pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -5596,10 +5598,25 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
 			    val_str.ptr = val + STR_DURATION.slen;
 			    val_str.slen = input.slen - STR_DURATION.slen;
 			    info.duration = pj_strtoul(&val_str);
+			} else {
+			    info.duration = PJSUA_UNKNOWN_DTMF_DURATION;
```
maxtruxa, line 5602: Kind of forgot about this. This should have been a separate PR.

`info.duration` was left uninitialized if the duration was not specified in the message body, making the field basically useless (no way of knowing whether the field is valid or not).

sauwming:
I have to applaud the quality of your PR. You seem to understand PJSIP so well, with macros and enums neatly separated and good knowledge of different levels of PJSIP API (pjmedia, pjsua-lib, pjsua2).

The only thing that's (slightly) bothering me is the name of the new callback. In pjmedia it is `dtmf2_callback`, in pjsua it's `on_call_dtmf3`, then in pjsua2 it's back to using number 2 as the suffix again: `onDtmfDigit2`. It can be confusing, though it's not your fault of course. But how about changing it to `dtmf_event`, or `dtmf_update` instead?

maxtruxa:
Well, thank you! I've worked quite a bit with the C/C++ components of PJSIP so far.

Regarding the "numbered callbacks" I have to agree. It bothered me as well. I really like the idea of `on_dtmf_event`, but would probably want to change the callback behavior slightly:

If you register `on_dtmf_digit2`, `on_dtmf_digit` is not being called anymore. This behavior is well documented. I kept this pattern with `on_dtmf_digit3`, basically replacing `on_dtmf_digit2`.
With the new callback renamed, it seems counter-intuitive that registering `on_dtmf_event` prevents `on_dtmf_digit2` from being called (even if it's documented), as there is no obvious dependency between the two names.
What I'm getting at: Would it make more sense to always call the "old" `on_dtmf_digit2`/`on_dtmf_digit` callbacks regardless of wether `on_dtmf_event` is registered or not? I can't think of a use case where it would make sense to register both callbacks, but allowing that would IMO be less surprising.

maxtruxa:
I went ahead and renamed the callbacks without changing behavior. The dependencies between the callbacks are clearly documented and most applications won't be using the DTMF events directly anyways, since it only addresses a very specific edge case.

maxtruxa:
Commit 811cdc1 adds a timestamp to all events which identifies the start time of the event in units of milliseconds. While the timestamp has no immediate relation to wallclock time, this is useful for calculating the exact "off" time between two tones (i.e. `off_time = start_of_tone_2 - start_of_tone_1 - duration_of_tone_1`).
For DTMF events sent via SIP INFO the timestamp is synthesized using the local tick count from `pj_get_timestamp`.

sauwming:
There's a conflict with #2425, so could you please help resolve it before we can merge this? Thanks.

maxtruxa:
Fixed the merge conflict and squashed commits in this PR.

PR #2451, opened 2020-06-01
Allow demote SRTP call to RTP when using re-INVITE with PJSUA_CALL_REINIT_MEDIA

trengginas:
Although the standards doesn't prohibit this, currently demoting a SRTP call to RTP call is not allowed when `PJSUA_CALL_REINIT_MEDIA` is specified.
This PR will allow demoting such scenario provided that `PJSUA_CALL_REINIT_MEDIA` is used when modifying the call session (using re-INVITE or UPDATE). 

Upon testing, a bug is also found when using DTLS-SRTP with this scenario:
1. Caller send INVITE
2. Callee send 200/OK
3. Caller send re-INVITE with modified media transport address (`PJSUA_CALL_REINIT_MEDIA` is specified). 
4. Callee doesn't reinitiate DTLS renegotiation triggering `PJMEDIA_SRTP_EKEYNOTREADY` error, since on the caller side it has re-initialize the media.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -39,7 +39,7 @@
 #endif
 
 /* Set to 1 to enable DTLS-SRTP debugging */
-#define DTLS_DEBUG  0
+#define DTLS_DEBUG  1
 
 /* DTLS-SRTP transport op */
 static pj_status_t dtls_media_create  (pjmedia_transport *tp,
@@ -898,10 +898,11 @@ static pj_status_t parse_setup_finger_attr(dtls_srtp *ds,
     return PJ_SUCCESS;
 }
 
-
 static pj_status_t get_rem_addrs(dtls_srtp *ds,
 				 const pjmedia_sdp_session *sdp_remote,
-				 unsigned media_index)
+				 unsigned media_index,
+                                 pj_sockaddr *rem_rtp,
+                                 pj_sockaddr *rem_rtcp)
 {
     pjmedia_sdp_media *m_rem = sdp_remote->media[media_index];
     pjmedia_sdp_conn *conn;
@@ -918,7 +919,7 @@ static pj_status_t get_rem_addrs(dtls_srtp *ds,
 	}
     }
     if (af != pj_AF_UNSPEC()) {
-	pj_sockaddr_init(af, &ds->rem_addr, &conn->addr,
+	pj_sockaddr_init(af, rem_rtp, &conn->addr,
 			 m_rem->desc.port);
     } else {
 	return PJ_EAFNOTSUP;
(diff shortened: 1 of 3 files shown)


pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1247,12 +1249,31 @@ static pj_status_t dtls_encode_sdp( pjmedia_transport *tp,
 		    (ds->setup==DTLS_SETUP_ACTIVE? &ID_ACTIVE:&ID_PASSIVE));
 	pjmedia_sdp_media_add_attr(m_loc, a);
 
+        if (last_setup != DTLS_SETUP_UNKNOWN && sdp_remote) {
+            pj_sockaddr rem_rtp;
+            pj_sockaddr rem_rtcp;
+
+            pj_bzero(&rem_rtp, sizeof(rem_rtp));
+            pj_bzero(&rem_rtcp, sizeof(rem_rtcp));
```
nanangizz, line 1257: unnecessary? `pj_sockaddr_init()` will be used in `get_rem_addrs()`

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3001,6 +3001,7 @@ pj_status_t pjsua_media_channel_deinit(pjsua_call_id call_id)
 	    call_med->tp = call_med->tp_orig = NULL;
 	}
         call_med->tp_orig = NULL;
+        call_med->rem_srtp_use = PJ_FALSE;
```
nanangizz, line 3004: should be set to `PJMEDIA_SRTP_DISABLED`? or create new enum `UNKNOWN` (==`DISABLED`) & set to this.

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1492,8 +1515,10 @@ static pj_status_t dtls_media_start( pjmedia_transport *tp,
...
 	    /* Attach ourselves to member transport for DTLS nego. */
-	    if (!pj_sockaddr_has_addr(&ds->rem_addr))
-		get_rem_addrs(ds, sdp_remote, media_index);
+            if (!pj_sockaddr_has_addr(&ds->rem_addr)) {
+                get_rem_addrs(ds, sdp_remote, media_index, &ds->rem_addr,
+                              &ds->rem_rtcp);
+            }
 
 	    if (pj_sockaddr_has_addr(&ds->rem_addr))
 		pj_sockaddr_cp(&ap.rem_addr, &ds->rem_addr);
```
nanangizz, line 1524: for consistency, shouldn't we also set `peer_use` (to `MANDATORY`) in `media_start()`?

pjmedia/src/pjmedia/transport_srtp_dtls.c
trengginas: ok

pjmedia/src/pjmedia/transport_srtp_dtls.c
trengginas: ok

pjsip/src/pjsua-lib/pjsua_media.c
trengginas: ok

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1268,7 +1265,7 @@ static pj_status_t dtls_encode_sdp( pjmedia_transport *tp,
         }
 
 	/* Check if remote signals DTLS re-nego by changing its
-	 * setup/fingerprint in SDP.
+	 * setup/fingerprint in SDP or changed in media transport address.
```
nanangizz, line 1268: perhaps "... setup/fingerprint or media transport address in SDP"?

sauwming:
Make sure it passes pjsua test first before merging since I saw several failures in srtp tests

trengginas:
> Make sure it passes pjsua test first before merging since I saw several failures in srtp tests

Test it here using and ubuntu and passed the tests. Tried testing from a fork with the patch, passed the test.

PR #2466, opened 2020-06-21
Blocker bug,Fixed a bug when a rtp pkt timestamp is 0 but not more th…

Ratelsun:
Blocker bug,Fixed a bug when a rtp pkt timestamp is 0 but not more than 0, then frist and second frame be made together for a new wrong frame, and it will make decoder die

### DIFF
--- a/pjmedia/src/pjmedia/vid_stream.c
+++ b/pjmedia/src/pjmedia/vid_stream.c
@@ -1279,7 +1279,7 @@ static pj_status_t decode_frame(pjmedia_vid_stream *stream,
                                 pjmedia_frame *frame)
 {
     pjmedia_vid_channel *channel = stream->dec;
-    pj_uint32_t last_ts = 0, frm_ts = 0;
+    pj_uint32_t last_ts = -1, frm_ts = 0;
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia/vid_stream.c
```diff
@@ -1279,7 +1279,7 @@ static pj_status_t decode_frame(pjmedia_vid_stream *stream,
                                 pjmedia_frame *frame)
 {
     pjmedia_vid_channel *channel = stream->dec;
-    pj_uint32_t last_ts = 0, frm_ts = 0;
+    pj_uint32_t last_ts = -1, frm_ts = 0;
```
nanangizz, line 1282: perhaps it's better to introduce a new boolean flag, e.g: `last_ts_set`, because:
- negative value does not fit unsigned type,
- RTP ts is 32 bits, so any value (including `0` & `-1`) may still match a real RTP ts.

nanangizz:
also, please stop spamming other PR/issue for attention!

nanangizz:
Continued in #2584.

PR #2468, opened 2020-06-23
Call state not progressing due to ICE callback not propagated

nanangizz:
In PJSIP internal, the ICE initialization callback may never propagated to PJSUA due to failure in acquiring PJSUA mutex, as shown in the log line:
```
Timed-out trying to acquire PJSUA mutex (possibly system has deadlocked) in ice_init_complete_cb
```
The problem is in the `ice_init_complete_cb()` implementation in `pjsua_media.c`, it simply returns when `acquire_call()` fails, because it wrongly assumes that the call has already ended.

Thanks to Mark Woodard for the report.


Ratelsun:
Can you help me check or review  #2466 ?

Issue #2430, opened 2020-05-13
Possible crash if PJSIP_TRANSPORT_IDLE_TIME is set to zero


Bjoe:
As the ticket [#267 (Configurable PJSIP_TRANSPORT_IDLE_TIME)](https://trac.pjsip.org/repos/ticket/267)  for example describe, to allow transports to be destroyed immediately they should set idle time to zero.

We use PjSip version 2.7.2 and I compare 2.7.2 with 2.10. It looks like the same issue is also in PJSip 2.10.

We set `PJSIP_TRANSPORT_IDLE_TIME` to `0` and we got sometimes a crash in following scenario:

If the sip password is changed of an account with a configured TCP transport, we call this function [pjsua_acc_modify()](https://github.com/pjsip/pjproject/blob/3e7b75cb2e482baee58c1991bd2fa4fb06774e0d/pjsip/src/pjsua-lib/pjsua_acc.c#L808) .  Because of the [credential change](https://github.com/pjsip/pjproject/blob/3e7b75cb2e482baee58c1991bd2fa4fb06774e0d/pjsip/src/pjsua-lib/pjsua_acc.c#L1238) `update_reg` and `unreg_first` is set to true. This means the SIP stack [send first an unregister](https://github.com/pjsip/pjproject/blob/3e7b75cb2e482baee58c1991bd2fa4fb06774e0d/pjsip/src/pjsua-lib/pjsua_acc.c#L1412) and[calls pjsip_regc_destroy()](https://github.com/pjsip/pjproject/blob/3e7b75cb2e482baee58c1991bd2fa4fb06774e0d/pjsip/src/pjsua-lib/pjsua_acc.c#L1424). Immediately after that the SIP stack [update the registration](https://github.com/pjsip/pjproject/blob/2.10/pjsip/src/pjsua-lib/pjsua_acc.c#L1438) and [calls pjsua_acc_set_registration()](https://github.com/pjsip/pjproject/blob/3e7b75cb2e482baee58c1991bd2fa4fb06774e0d/pjsip/src/pjsua-lib/pjsua_acc.c#L1441).

Following happens:

[...]

PR #2470, opened 2020-06-24
Race condition between transport destroy and acquire

nanangizz:
The race is between `transport_idle_callback()` and `pjsip_tpmgr_acquire_transport2()`:
- when `destroy` locks the transport manager mutex first, things seems to be okay, transport will never be acquired.
- when `acquire` locks the transport manager mutex first (and the transport is acquired successfully), but `destroy` will keep on going (as it does not re-check the transport ref count), this is the problem.

So the solution should be about:
- `transport_idle_callback()` must stop the `destroy` if the transport is re-acquired (i.e: transport `ref_cnt` is not zero).
- `pjsip_tpmgr_acquire_transport2()` must not return a transport that is being destroyed.

See also discussions in #2430.

### DIFF
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -2411,7 +2411,7 @@ static void tsx_update_transport( pjsip_transaction *tsx,
 	pjsip_transport_add_ref(tp);
 	pjsip_transport_add_state_listener(tp, &tsx_tp_state_callback, tsx,
 					    &tsx->tp_st_key);
-        if (tp->is_shutdown) {
+	if (tp->is_shutdown || tp->is_destroying) {
 	    pjsip_transport_state_info info;
 
 	    pj_bzero(&info, sizeof(info));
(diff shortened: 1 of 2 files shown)


Bjoe:
PR looks promise.

I've been thinking what are about following function/lines:

`tsx_update_transport()`
https://github.com/pjsip/pjproject/blob/ce6c72b578b140d521083a25c0ff817cf490b103/pjsip/src/pjsip/sip_transaction.c#L2414-L2421

`pjsip_transport_shutdown2()`
https://github.com/pjsip/pjproject/blob/ce6c72b578b140d521083a25c0ff817cf490b103/pjsip/src/pjsip/sip_transport.c#L1405-L1409

`pjsip_transport_add_state_listener()`
https://github.com/pjsip/pjproject/blob/ce6c72b578b140d521083a25c0ff817cf490b103/pjsip/src/pjsip/sip_transport.c#L2645-L2648

they verify if the transport "is in shutdown state". Maybe they should also verify if the transport "is in destroying state"?

trengginas (approved):

PR #2471, opened 2020-06-25
Fix sip transaction to send pending message

sauwming:
Example scenario:
1. RX Invite
2. 100 initial answer
3. Answer with 200 while 2 is still pending.
4. After 100 completed, the 200/OK never got sent and the call just stalled.

After investigation, there are two identified problems:
* There is no mechanism to send any pending message in transport_callback(), unlike in send_msg_callback().
* In sip invite, pjsip_inv_answer() will call pjsip_tx_data_clone() of the last response. If the last response is still pending, the resulting cloned tdata will also be marked pending, which will never get cleared since it's not actually being sent (The discussion to clone field is_pending can be found in our internal discussion for [CR issue 38787003](https://cr.teluu.com/38787003/#msg15)). As a result, tsx_send_msg() will always fail with `"Unable to send %s: message is pending".`

### DIFF
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -2042,6 +2042,38 @@ static void transport_callback(void *token, pjsip_tx_data *tdata,
      */
     pj_grp_lock_acquire(tsx->grp_lock);
     tsx->transport_flag &= ~(TSX_HAS_PENDING_TRANSPORT);
+
+    if (sent > 0) {
+	/* Pending destroy? */
+	if (tsx->transport_flag & TSX_HAS_PENDING_DESTROY) {
+	    tsx_set_state( tsx, PJSIP_TSX_STATE_DESTROYED,
+			   PJSIP_EVENT_UNKNOWN, NULL, 0 );
+	    pj_grp_lock_release(tsx->grp_lock);
+	    return;
+	}
+
+	/* Need to transmit a message? */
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsip/sip_transaction.c
```diff
@@ -2042,6 +2042,38 @@ static void transport_callback(void *token, pjsip_tx_data *tdata,
...
+	    tsx_set_state( tsx, PJSIP_TSX_STATE_DESTROYED,
+			   PJSIP_EVENT_UNKNOWN, NULL, 0 );
+	    pj_grp_lock_release(tsx->grp_lock);
+	    return;
+	}
+
+	/* Need to transmit a message? */
+	if (tsx->transport_flag & TSX_HAS_PENDING_SEND) {
+	    tsx->transport_flag &= ~(TSX_HAS_PENDING_SEND);
+	    tsx_send_msg(tsx, tsx->last_tx);
```
nanangizz, line 2058: perhaps better to have some check/protection?
```
if (tsx->last_tx && tsx->last_tx != tdata)
  tsx_send_msg();
```

sauwming: The block is actually identical to the one in `send_msg_callback()`. So should I add the check in both? Or is there any reason why the check is only applicable for this callback but not the other?

nanangizz: ah wasn't aware of that, as so far `send_msg_callback()` has been proven to be safe, I think this callback should be safe too without the checks.

Issue #2460, opened 2020-06-15
Imposible to create `pjmedia_stereo_port` with options


goro-gh:
If StereoPort with `PJMEDIA_STEREO_DONT_DESTROY_DN` option is created, `pjmedia_convert_channel_1ton()` asserts with

```
../include/pjmedia/stereo.h:132: pjmedia_convert_channel_1ton: Assertion `options == 0' failed.
```

trengginas:
> PJMEDIA_STEREO_DONT_DESTROY_DN

Thanks for the report.
Could you test the patch?

[patch_stereo_port.txt](https://github.com/pjsip/pjproject/files/4809716/patch_stereo_port.txt)

PR #2474, opened 2020-06-29
Stereo port fix

goro-gh:
In this series of commits i try to address 4 issues.
- Impossibility to create StereoPort with options (like `PJMEDIA_STEREO_DONT_DESTROY_DN`); #2460 
- Bad sounding in Stereo to Mono conversion (half signaled, half zeroed frames in mono side)
- Unwanted signal amplification in mix algorithm (R+L instead of L+R/2)
- An app memory corruption

All of the problem are detected in setup: StereoConfBridge ==> StereoPort --> MonoSndPort --> AlsaDev

### DIFF
--- a/pjmedia/include/pjmedia/stereo.h
+++ b/pjmedia/include/pjmedia/stereo.h
@@ -93,6 +93,7 @@ PJ_INLINE(pj_status_t) pjmedia_convert_channel_nto1(pj_int16_t mono[],
 	    for(j = 0; j < channel_count; ++j)
 		tmp += multi[i+j];
 
+	    tmp = tmp / (int)channel_count;
 	    if (tmp > 32767) tmp = 32767;
 	    else if (tmp < -32768) tmp = -32768;
 	    *mono = (pj_int16_t) tmp;
--- a/pjmedia/include/pjmedia/stereo.h
+++ b/pjmedia/include/pjmedia/stereo.h
@@ -129,7 +129,6 @@ PJ_INLINE(pj_status_t) pjmedia_convert_channel_1ton(pj_int16_t multi[],
 
     PJ_ASSERT_RETURN(mono && multi && channel_count && samples_per_frame, 
 		     PJ_EINVAL);
-    PJ_ASSERT_RETURN(options == 0, PJ_EINVAL);
 
     PJ_UNUSED_ARG(options);
 
--- a/pjmedia/src/pjmedia/stereo_port.c
+++ b/pjmedia/src/pjmedia/stereo_port.c
@@ -183,12 +183,10 @@ static pj_status_t stereo_get_frame(pjmedia_port *this_port,
     if (status != PJ_SUCCESS)
 	return status;
 
+    // If downstream provides non PCM frames, skip conversion and return silence
+    // For example, empty ConfBridge gives PJMEDIA_FRAME_TYPE_NONE
     if (tmp_frame.type != PJMEDIA_FRAME_TYPE_AUDIO) {
-	frame->type = tmp_frame.type;
-	frame->timestamp = tmp_frame.timestamp;
-	frame->size = tmp_frame.size;
-	if (tmp_frame.size && tmp_frame.buf == sport->get_buf)
-	    pj_memcpy(frame->buf, tmp_frame.buf, tmp_frame.size);
+	pj_bzero(frame->buf, frame->size);
 	return PJ_SUCCESS;
     }
 
--- a/pjmedia/src/pjmedia/stereo_port.c
+++ b/pjmedia/src/pjmedia/stereo_port.c
@@ -196,7 +196,7 @@ static pj_status_t stereo_get_frame(pjmedia_port *this_port,
 	pjmedia_convert_channel_nto1((pj_int16_t*)frame->buf, 
 				     (const pj_int16_t*)tmp_frame.buf,
 				     dn_afd->channel_count,
-				     PJMEDIA_AFD_SPF(s_afd),
+				     PJMEDIA_AFD_SPF(dn_afd),
 				     (sport->options & PJMEDIA_STEREO_MIX), 0);
     } else {
 	pjmedia_convert_channel_1ton((pj_int16_t*)frame->buf, 


goro-gh:
Supper strange CICD failure!

pjmedia/include/pjmedia/stereo.h
```diff
@@ -93,6 +93,7 @@ PJ_INLINE(pj_status_t) pjmedia_convert_channel_nto1(pj_int16_t mono[],
 	    for(j = 0; j < channel_count; ++j)
 		tmp += multi[i+j];
 
+	    tmp = tmp / (int)channel_count;
```
sauwming, line 96: I understand the unwanted signal amplification. But unfortunately, if a sound is only played in one of the channels, dividing it by the channel count will cause unwanted signal reduction instead?

pjmedia/include/pjmedia/stereo.h
```diff
@@ -129,7 +130,6 @@ PJ_INLINE(pj_status_t) pjmedia_convert_channel_1ton(pj_int16_t multi[],
 
     PJ_ASSERT_RETURN(mono && multi && channel_count && samples_per_frame, 
 		     PJ_EINVAL);
-    PJ_ASSERT_RETURN(options == 0, PJ_EINVAL);
```
sauwming, line 132: Use Riza's patch to fix the assertion instead. There's no use in passing non-zero options here anyway.

pjmedia/src/pjmedia/stereo_port.c
```diff
@@ -183,20 +183,18 @@ static pj_status_t stereo_get_frame(pjmedia_port *this_port,
...
 
+    // If downstream provides non PCM frames, skip conversion and return silence
+    // For example, empty ConfBridge gives PJMEDIA_FRAME_TYPE_NONE
     if (tmp_frame.type != PJMEDIA_FRAME_TYPE_AUDIO) {
-	frame->type = tmp_frame.type;
-	frame->timestamp = tmp_frame.timestamp;
-	frame->size = tmp_frame.size;
-	if (tmp_frame.size && tmp_frame.buf == sport->get_buf)
-	    pj_memcpy(frame->buf, tmp_frame.buf, tmp_frame.size);
+	pj_bzero(frame->buf, frame->size);
```
sauwming, line 189: This will basically zero all non-audio frames. If you only want to handle TYPE_NONE, I would suggest changing it to:
```
    if (tmp_frame.type != PJMEDIA_FRAME_TYPE_AUDIO) {
	frame->type = tmp_frame.type;
	if (tmp_frame.type == PJMEDIA_FRAME_TYPE_NONE) {
	    pj_bzero(frame->buf, frame->size);
	} else {
	    frame->timestamp = tmp_frame.timestamp;
	    frame->size = tmp_frame.size;
	    if (tmp_frame.size && tmp_frame.buf == sport->get_buf)
	        pj_memcpy(frame->buf, tmp_frame.buf, tmp_frame.size);
	}
	return PJ_SUCCESS;
    }
```

pjmedia/include/pjmedia/stereo.h
sauwming: As for the solution: if you're sure that all your channels play the same sound, perhaps you can adjust the signal level instead?

goro-gh: There are two strategies in `pjmedia_convert_channel_nto1` for dealing with stereo channels.

- mono = L | R (the user select witch channel wants to mono side)
- mono = L + R (the user wants them summed)

```c
    if (mix==PJ_FALSE) {
        // mono is L or R
    } else {
        // mono is L + R 
    }
```

When stereo channels are **same**, **phase inverted** or **one of them is zeroed**, then the user should use (L | R) strategy.

When stereo channels are real asymmetric channels, most of the users choice is (L + R) so, no signal get lost.

On (L+R) strategy, if the (L+R) signal is not divided by channel count, then it easily get clipped by 16bit boundaries.

```c
    tmp = tmp / (int)channel_count;
    
    if (tmp > 32767) tmp = 32767;
    else if (tmp < -32768) tmp = -32768;
    
    *mono = (pj_int16_t) tmp;
```

So, not dividing by `channel_count`, give us twice amplification on signals with amplitude peaks below -6db, and distortion effect on signals above -6db.


ps. 
this `if (mix==PJ_FALSE)` really should be `if (options & PJMEDIA_STEREO_DONT_MIX)` but that is another story.

pjmedia/src/pjmedia/stereo_port.c
goro-gh: This was my first guess, but then I wondered who would put StereoPort on non PCM path ?
It can't deal others than TYPE_AUDIO frames.

pjmedia/include/pjmedia/stereo.h
goro-gh: Ok, but this assertion will be removed anyway, if somebody try to make `channel_1ton` and `channel_nto1` fully symmetric, implementing the missing part in `channel_1ton`

say 
- stereo = (mono, mono)
- stereo = (mono, 0)
- stereo = (0, mono)

And when make it, should go and revert Riza's patch.
I think, this will be my second contribution, because I need it. I will open feature request for that, to discuss it.

ps.
Is there an unwritten rule: Don't touch header files. :) 
I know there is some pain in compile time when touching headers, but if no API/ABI changes are introduced all be fine.

pjmedia/include/pjmedia/stereo.h
sauwming: Personally, I prefer your approach, especially after reading several references on the net such as in the [DSP forum here](https://www.dsprelated.com/showthread/comp.dsp/106421-1.php).

My only concern is the change of behaviour that this introduces, especially after many years that the code has been this way. So let's hear what the other reviewers think about this.

pjmedia/include/pjmedia/stereo.h
sauwming: There's no such rule. It's just that the assertion is not wrong, and it's exactly according to the spec above it (`Options for conversion, currently must be zero`), so there's no reason to remove it. It's the API user that is at fault and that's exactly what the patch fixes.

But if you have plan to extend its features and allow non-zero options, then sure, removing the assertion does make sense.

pjmedia/src/pjmedia/stereo_port.c
sauwming: I wonder the same thing, but same as above, when the code has been around for so many years, I tend to try not to change it unless it's broken (i.e. only fix the broken part and leave the unbroken one as is).

pjmedia/include/pjmedia/stereo.h
nanangizz: As there seems to be some approaches to do N-to-1 conversion, I think it is better for the stereo port, as a base/lower-level module, to provide options that may accommodate those approaches, so the application will make use the options according to its needs. In this case, instead of replacing the existing approach with a new approach, perhaps we should add a new option for (L+R)/2.

Regarding the original issue: signal amplification in the setup of `StereoConfBridge ==> StereoPort --> MonoSndPort --> AlsaDev`, perhaps it is the application (i.e: conference bridge) that need to be 'fixed' or updated (e.g: replace `PJMEDIA_STEREO_MIX` param in its `pjmedia_convert_channel_nto1()` invocation, or make it configurable to avoid behavior change). Perhaps we can safely assume that stereo signal usually consists of the same signals in L & R, so we can just choose any channel to generate the mono channel.

SebaLedesma:
Hi! I want to add some info.
I work for a company called Solidyne that makes audio mixers for broadcasting (https://www.linkedin.com/in/sebastian-ledesma-3b02881b/  and http://www.solidynepro.com/ ).

Long story short: when you have to mix L & R into a mono signal the algorythm is: M = ( L / 2 + R / 2 )
The other aproach is simple a bug : M = L + R as it will caused a unintended 6db amplification and also a possible clipping (and there is no return from that).

To choice M = L, it's not mixing a signal, it's called switcher, you pick left or right channel and discard the other. 

Extra info: we have digital consoles with several channels. For example in a 16 channel console, each channel it's divided by 16 before enter into the sum, The hardware actually have 20 bits conversors , and internally it use more bits to keep resolution and to have some headroom. But the case it's pretty the same, when mixing several channels, each channel enter into the mix divided by the number of channels, no matter if the channel it's on or off.  
PS: I'm working on include pjsip capabilities to this console https://www.youtube.com/watch?v=bP9CqapmvjE

goro-gh:
> Long story short: when you have to mix L & R into a mono signal the algorythm is: M = ( L / 2 + R / 2 )

I like the short story. That way there is no signal reduction, as @sauwming worries, when one of the chanells is 0.

nanangizz:
> > Long story short: when you have to mix L & R into a mono signal the algorythm is: M = ( L / 2 + R / 2 )
> 
> I like the short story. That way there is no signal reduction, as @sauwming worries, when one of the chanells is 0.

I remember how the conference bridge doing mixing back then, there was volume reduction indeed when division was applied, e.g: `mix = (source1/3 + source2/3 + source3/3)`, the more source number, the lower the volume became, so now it uses 'mix = source1 + source2 + source3'. Yes, there is a risk of clipping, so the bridge is equipped with a simple AGC to handle that, which works by lowering volume of each source until the clipping is avoided (the mixing buffer uses 32 bit sample) then increasing back the volume slowly.

So if one channel is silent, there will be reduction, similar way as amplification in `mix = L + R` when L == R. So I think `mix = L/2 + R/2` works fine (no amplification) when both channels L & R has same signals, which perhaps a common case for stereo -> mono conversion.

sauwming:
Closing this due to lack of recent update. Pls reopen/recreate if needed.

PR #2490, opened 2020-07-16
PJSUA call: fix on_pager_statusX() callbacks for in-dialog requests

andreas-wehrmann:
on_pager_status() was called twice: once for event TSX_STATE and later for event TIMER

on_pager_status2() was not called at all

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_call.c
+++ b/pjsip/src/pjsua-lib/pjsua_call.c
@@ -5454,8 +5454,12 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
 			       &inv->dlg->local.info_str, rdata);
 
     }
-    else if (tsx->role == PJSIP_ROLE_UAC &&
-	     pjsip_method_cmp(&tsx->method, &pjsip_message_method)==0)
+    else if (e->type == PJSIP_EVENT_TSX_STATE &&
+	     tsx->role == PJSIP_ROLE_UAC &&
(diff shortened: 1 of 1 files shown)


andreas-wehrmann:
Sorry, I sometimes forget to switch settings.
Also, I replaced the tabs with spaces; spaces are the way to go, yes?
I'm asking because in some places there seems to be a mixture of tabs and spaces.

nanangizz:
Actually for indentation, usually it is mixed, but it is okay with spaces only.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -5454,32 +5454,57 @@ static void pjsua_call_on_tsx_state_changed(pjsip_inv_session *inv,
...
+                                                            &tsx->status_text);
+                }
+
+                if (pjsua_var.ua_cfg.cb.on_pager_status2) {
+                    pjsip_rx_data* rdata;
+
+                    if (e->body.tsx_state.type == PJSIP_EVENT_RX_MSG)
+                    rdata = e->body.tsx_state.src.rdata;
+                    else
+                    rdata = NULL;
```
sauwming, line 5494: minor: indentation, i.e.
if ()
    statement;
else
    statement;

nanangizz: as this is minor, leave it that way should be fine for now as the "if" structure is quite simple, or we can fix it in Miscellaneous issue later perhaps.

PR #2492, opened 2020-07-17
No keyframe missing event from OpenH264 decoder after many decoding error

nanangizz:
Currently decoder is invoked twice: on each NAL unit and later invoked for the whole packet, but only the latter (for whole packet) can trigger missing-keyframe event. Unfortunately, sometimes it is the NAL decoding that reports error while the whole packet decoding does not, this causes false error-free decoding and missing keyframe event is not generated.

The patch is to also perform check for any error in NAL decoding and generate missing keyframe error if any.

### DIFF
--- a/pjmedia/src/pjmedia-codec/openh264.cpp
+++ b/pjmedia/src/pjmedia-codec/openh264.cpp
@@ -972,6 +972,7 @@ static pj_status_t oh264_codec_decode(pjmedia_vid_codec *codec,
     const pj_uint8_t nal_start[] = { 0, 0, 1 };
     SBufferInfo sDstBufInfo;
     pj_bool_t has_frame = PJ_FALSE;
+    pj_bool_t kf_requested = PJ_FALSE;
     unsigned buf_pos, whole_len = 0;
     unsigned i, frm_cnt;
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia-codec/openh264.cpp
```diff
@@ -1056,7 +1057,18 @@ static pj_status_t oh264_codec_decode(pjmedia_vid_codec *codec,
 	start = oh264_data->dec_buf + buf_pos;
 
 	/* Decode */
-	oh264_data->dec->DecodeFrame2( start, frm_size, pData, &sDstBufInfo);
+	ret = oh264_data->dec->DecodeFrame2( start, frm_size, pData,
+					     &sDstBufInfo);
+
+	if (ret != dsErrorFree && !kf_requested) {
```
sauwming, line 1063: why is !kf_requested check necessary here since the value must be FALSE? for future use?

nanangizz: just to avoid multiple events per frame decoding, note that this decode is inside a loop.

PR #2501, opened 2020-08-03
libpjproject.pc: Move external libraries to Libs.private.

Apteryks:
This is done so users of a shared pjproject library do not get link
directives (provided via pkg-config file of pjproject) to link against
pjproject's own library dependencies.

Static linking with pjproject is preserved by moving the external
libraries link directives to the Libs.private field, which is
specifically intended for this purpose, via for example:

  $ pkg-config --static --libs libpjproject

* build.mak.in (PJ_INSTALL_LDFLAGS_PRIVATE): New variable.
(PJ_INSTALL_LDFLAGS): Filter out PJ_INSTALL_LDFLAGS_PRIVATE entries.
* libpjproject.pc.in (Libs.private): New field.
* Makefile: Substitute PJ_INSTALL_LDFLAGS_PRIVATE as the value for the
Libs.private field.


nanangizz:
This looks similar to #2403, with additional change of stripping out third party libs. Shouldn't app still need to specify those third-party libs in its build config? Or perhaps you find a specific case that the third-party libs cause problem in building app?

H5117:
@Apteryks First of all, your patch brokes linking with statically compiled version of pjproject. You are right that pjproject's dependencies should not be placed in `Libs:`, but according to specification they should be placed in `Libs.private:`. In this case calling `pkg-config --static --libs libpjproject` will return a concatenation of `Libs:` and `Libs.private:`, and linking will work for both static and shared versions of pjproject.

Secondly, there are no separate properties in .pc files for C and C++ projects. So IMO the .pc file should be generated for C++ projects, to include all libraries of pjproject.

Apteryks:
Hello!

> This looks similar to #2403, with additional change of stripping out third party libs.

Hello! It's not just similar to #2403, it's based on it :-). I see it has now been merged, should I've rebased this branch on it. 

> Shouldn't app still need to specify those third-party libs in its build config? Or perhaps you find a specific case that the third-party libs cause problem in building app?

Working in the context of pjproject built as a shared library, with a few optional dependencies such as `gsm`, I was surprised that the application I'm working with (`libring`, a.k.a. `jami-daemon`) wanted to link with `gsm` as well (and failed to, being built in a minimal container without `gsm` present).

Comments from H5117 below point to a better solution, which I'll experiment with (although I'm surprised the CI passes if this change truly breaks static linking with pjproject!).

Apteryks:
> @Apteryks First of all, your patch brokes linking with statically compiled version of pjproject. You are right that pjproject's dependencies should not be placed in `Libs:`, but according to specification they should be placed in `Libs.private:`. 

Interesting! I somehow thought that static libraries were self-contained, i.e., they bundled all their transitive dependencies in their own archive. Isn't it the case? I'll read more about Libs.private and send an updated patch, thanks for the pointer!

> In this case calling `pkg-config --static --libs libpjproject` will return a concatenation of `Libs:` and `Libs.private:`, and linking will work for both static and shared versions of pjproject.

So linking with a static library always require to also link with the dependencies of said static library? That's new to me, but then, I don't know much about static libraries :-).

> Secondly, there are no separate properties in .pc files for C and C++ projects. So IMO the .pc file should be generated for C++ projects, to include all libraries of pjproject.

I'm not sure I understand your comment here; what are you proposing? We're already using APP_LDXXLIBS for the LDFLAGS, which include the C++ libraries.

Apteryks:
So, I've built the updated patch with @H5117's suggestion, and the libpjproject.pc file looks like this:

```
cat /gnu/store/l47dj1r9fdbg0c1madj9gqg4qrks359c-pjproject-2.10-1.a9f4e1a/lib/pkgconfig/libpjproject.pc 
# Package Information for pkg-config

prefix=/gnu/store/l47dj1r9fdbg0c1madj9gqg4qrks359c-pjproject-2.10-1.a9f4e1a
exec_prefix=${prefix}
libdir=/gnu/store/l47dj1r9fdbg0c1madj9gqg4qrks359c-pjproject-2.10-1.a9f4e1a/lib
includedir=/gnu/store/l47dj1r9fdbg0c1madj9gqg4qrks359c-pjproject-2.10-1.a9f4e1a/include

Name: libpjproject
Description: Multimedia communication library
URL: http://www.pjsip.org
Version: 2.10-dev
Libs: -L/gnu/store/l47dj1r9fdbg0c1madj9gqg4qrks359c-pjproject-2.10-1.a9f4e1a/lib -lpjsua2 -lstdc++ -lpjsua -lpjsip-ua -lpjsip-simple -lpjsip -lpjmedia-codec -lpjmedia -lpjmedia-videodev -lpjmedia-audiodev -lpjmedia -lpjnath -lpjlib-util -lpj
Libs.private:  -lsrtp2 -lgsm -lspeex -lspeexdsp -lportaudio -lbcg729 -lopus -lsamplerate -lm -lrt -lpthread    -L/gnu/store/0f6agk6xk8rcz99907d39nd1vzmc4abq-gnutls-3.6.14/lib -lgnutls
Cflags: -I/gnu/store/l47dj1r9fdbg0c1madj9gqg4qrks359c-pjproject-2.10-1.a9f4e1a/include -DPJ_AUTOCONF=1  -DPJ_IS_BIG_ENDIAN=0 -DPJ_IS_LITTLE_ENDIAN=1
```

And for the invocation of pkg-config:
```
pkg-config --libs libpjproject
-L/gnu/store/l47dj1r9fdbg0c1madj9gqg4qrks359c-pjproject-2.10-1.a9f4e1a/lib -lpjsua2 -lstdc++ -lpjsua -lpjsip-ua -lpjsip-simple -lpjsip -lpjmedia-codec -lpjmedia -lpjmedia-videodev -lpjmedia-audiodev -lpjmedia -lpjnath -lpjlib-util -lpj
```
```
pkg-config --static --libs libpjproject
-L/gnu/store/l47dj1r9fdbg0c1madj9gqg4qrks359c-pjproject-2.10-1.a9f4e1a/lib -L/gnu/store/0f6agk6xk8rcz99907d39nd1vzmc4abq-gnutls-3.6.14/lib -lpjsua2 -lstdc++ -lpjsua -lpjsip-ua -lpjsip-simple -lpjsip -lpjmedia-codec -lpjmedia -lpjmedia-videodev -lpjmedia-audiodev -lpjmedia -lpjnath -lpjlib-util -lpj -lsrtp2 -lgsm -lspeex -lspeexdsp -lportaudio -lbcg729 -lopus -lsamplerate -lm -lrt -lpthread -lgnutls
```
So, it looks good!

H5117:
@Apteryks
> So linking with a static library always require to also link with the dependencies of said static library?

Yes, it is. A static library becomes a part of your project (executable or shared library), so you have to link your project with all inherited dependencies also.

> I'm not sure I understand your comment here; what are you proposing? We're already using APP_LDXXLIBS for the LDFLAGS, which include the C++ libraries.

You declared PJ_INSTALL_LDFLAGS and PJ_INSTALL_LDXXFLAGS (or something) in the description of your first patch. I meant that this is odd.

> Working in the context of pjproject built as a shared library, with a few optional dependencies such as gsm, I was surprised that the application I'm working with (libring, a.k.a. jami-daemon) wanted to link with gsm as well (and failed to, being built in a minimal container without gsm present).

Jami can't be used with the official version of pjproject, because it requires patches that have not been merged upstream (for example, https://github.com/pjsip/pjproject/pull/2363). So you should build a static version of pjproject from `<jami-daemon>/contrib/` :-).

The patch looks good to me (but I haven't tested it).

Apteryks:
@H5117 

> Jami can't be used with the official version of pjproject, because it requires patches that have not been merged upstream (for example, #2363). So you should build a static version of pjproject from `<jami-daemon>/contrib/` :-).

I'm aware of this. Still, I've made it so that the pjproject-jami package in Guix shares the same definition with the vanilla pjproject package in Guix. They only differs in those custom patches being applied:

```
(define-public pjproject-jami
  (package
    (inherit pjproject)
    (name "pjproject-jami")
    (native-inputs
     `(("sfl-patches" ,(jami-source))
       ,@(package-native-inputs pjproject)))
    (arguments
     (substitute-keyword-arguments (package-arguments pjproject)
       ((#:phases phases '%standard-phases)
        `(modify-phases ,phases
           (add-after 'make-source-files-writable 'apply-patches
             (lambda* (#:key inputs #:allow-other-keys)
               (,jami-apply-dependency-patches
                #:inputs inputs
                #:dep-name "pjproject"
                #:patches
                '("0001-rfc6544"
                  "0002-rfc2466"
                  "0003-add-tcp-keep-alive"
                  "0004-multiple_listeners"
                  "0005-fix_ebusy_turn"
                  "0006-ignore_ipv6_on_transport_check"
                  "0007-pj_ice_sess"
                  "0008-fix_ioqueue_ipv6_sendto"
                  "0009-add-config-site"
                  ;; Note: The base pjproject is already patched with
                  ;; "0010-fix-pkgconfig".
                  "0011-fix-tcp-death-detection"
                  "0012-fix-turn-shutdown-crash"))
               #t))))))))
```
This makes maintenance easy.

nanangizz (approved):

sauwming (approved):

PR #2507, opened 2020-08-13
Implement audio stream configuration callback

maxtruxa:
Currently it is not possible to modify jitter buffer settings per stream. The global jitter buffer settings [are applied to all streams created by pjsua](https://github.com/pjsip/pjproject/blob/32153443e7452ff790e4ef04ef5c1834b9da3e2a/pjsip/src/pjsua-lib/pjsua_aud.c#L645) (and hence pjsua2).
This PR introduces the callbacks `pjsua_callback::on_stream_configure` (pjsua) and `pj::Call::onStreamConfigure` (pjsua2) which enable on-demand modification of the `pjmedia_stream_info` structure before the stream is actually created.

Changes affect pjsua and pjsua2.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -531,6 +531,58 @@ typedef struct pjsua_reg_info
 
 
 /**
+ * Media stream info.
+ */
+typedef struct pjsua_stream_info
+{
+    /** Media type of this stream. */
+    pjmedia_type type;
+
+    /** Stream info (union). */
+    union {
+	/** Audio stream info */
+	pjmedia_stream_info	aud;
+
+	/** Video stream info */
+	pjmedia_vid_stream_info	vid;
+    } info;
+
+} pjsua_stream_info;
+
+
+/**
+ * Media stream statistic.
+ */
+typedef struct pjsua_stream_stat
+{
+    /** RTCP statistic. */
+    pjmedia_rtcp_stat	rtcp;
+
+    /** Jitter buffer statistic. */
+    pjmedia_jb_state	jbuf;
+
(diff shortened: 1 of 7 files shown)


sauwming:
I'm generally okay with this PR, with a few notes:
- Too bad we can't use the existing on_stream_created2() callback. I'm not too fond of the callback name on_stream_configure(), but then I can't think of a better name as well (not sure if on_stream_precreate() is better). But in any case (whether we change the name or not), I think some additional docs is helpful to differentiate with the existing callbacks, i.e.:
```
about to be created (as compared to the callback on_stream_created() or created2() which will be called after the session has been created).
```
- Let's see first whether other reviewers are okay with the addition of the new callback.

nanangizz:
Generally okay too.

`on_stream_precreate()` sounds better so it can be used for more than 'configurer' callback (e.g: simply notifier or any other purpose in the future). Also, with param `pjsua_on_stream_precreate_param` (just like `on_stream_created2`), to allow adding new params in the future without introducing another new PJSUA callback.

Please also implement PJSUA2 version of `pjmedia_stream_info`, trimmed/simpler one should be okay for now.

Thanks.

maxtruxa:
Those were all valid concerns. Thanks for your feeedback!

Relevant changes:
- Renamed the callbacks to `pjsua_callback::on_stream_precreate`/`pj::Call::onStreamPreCreate`
- Wrapped arguments to `pjsua_callback::on_stream_precreate` in new type `pjsua_on_stream_precreate_param`
- Extended the existing type `pj::StreamInfo` with some of the missing fields from `pjsua_stream_info`

sauwming (approved):

nanangizz (approved):

PR #2518, opened 2020-09-09
Option to send keep alive periodically when stream is started for NAT hole punching

trengginas:
Consider a call involving a media server. 
When the call is connected, it will not send any data before it receive any data from remote.
Current behavior:
- keep-alive packet is sent when the stream is created. Since it was sent before the 200/OK is sent, the media server might ignore this.
- keep-alive packet is sent after ```PJMEDIA_STREAM_KA_INTERVAL``` is passed. This means that when the call is set to "recvonly", the call will receive the data from the server after that interval pass. (default: 5s)

This ticket will introduce new build settings ```PJMEDIA_STREAM_START_KA_CNT``` and ```PJMEDIA_STREAM_START_KA_NTERVAL_MSEC``` with their runtime counterpart ```pjmedia_stream_ka_config```. This will allow user to specify the number of keep-alive packet sent and it's interval when the stream is started.

### DIFF
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -1625,10 +1625,10 @@
  *
  * Setting this to 0 will disable it.
  *
- * Default : 0
+ * Default : 2
  */
 #ifndef PJMEDIA_STREAM_START_KA_CNT
-#   define PJMEDIA_STREAM_START_KA_CNT	0
+#   define PJMEDIA_STREAM_START_KA_CNT	2
 #endif
 
 
@@ -1638,8 +1638,8 @@
  *
  * Default : 1000
  */
-#ifndef PJMEDIA_STREAM_START_KA_NTERVAL_MSEC
-#   define PJMEDIA_STREAM_START_KA_NTERVAL_MSEC  1000
+#ifndef PJMEDIA_STREAM_START_KA_INTERVAL_MSEC
+#   define PJMEDIA_STREAM_START_KA_INTERVAL_MSEC  1000
 #endif
 
 
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -1262,6 +1262,30 @@
 
 
 /**
+ * Specify the number of keep-alive needed to be sent after the stream is
+ * created.
+ *
+ * Setting this to 0 will disable it.
+ *
+ * Default : 2
+ */
+#ifndef PJMEDIA_STREAM_START_KA_CNT
+#   define PJMEDIA_STREAM_START_KA_CNT	2
+#endif
+
+
+/**
+ * Specify the interval to send keep-alive after the stream is created,
+ * in msec.
+ *
+ * Default : 1000
+ */
+#ifndef PJMEDIA_STREAM_START_KA_INTERVAL_MSEC
+#   define PJMEDIA_STREAM_START_KA_INTERVAL_MSEC  1000
+#endif
+
+
+/**
  * Specify the number of identical consecutive error that will be ignored when 
  * receiving RTP/RTCP data before the library tries to restart the transport.
  *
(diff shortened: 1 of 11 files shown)


pjmedia/include/pjmedia/config.h
```diff
@@ -1619,6 +1619,29 @@
...
+
+
+/**
+ * Specify the interval to send keep-alive after the stream is created,
+ * in msec.
+ *
+ * Default : 1000
+ */
+#ifndef PJMEDIA_STREAM_START_KA_NTERVAL_MSEC
+#   define PJMEDIA_STREAM_START_KA_NTERVAL_MSEC  1000
```
nanangizz, line 1642: shouldn't it `INTERVAL` instead of `NTERVAL`?

pjmedia/src/pjmedia/stream.c
```diff
@@ -1324,16 +1332,30 @@ static pj_status_t put_frame_imp( pjmedia_port *port,
...
-	}
+            pj_get_timestamp(&now);
+            elapse_time = pj_elapsed_msec(&stream->last_start_ka_tx, &now);
+	    if ((elapse_time > stream->start_ka_interval))
+	    {
+	        send_keep_alive_packet(stream);
+	        stream->last_start_ka_tx = now;
+                stream->start_ka_count--;
+	    }
+        } else {
```
nanangizz, line 1347: If outgoing RTP has been started, is there a chance that KA & RTP packets will be sent at the same time? may it cause any issue (e.g: remote jitter buffer may overwrite audio payload with KA/empty payload)?
Anyway, perhaps it can be minimized if the start-KA also consider `last_frm_ts_sent`.

pjmedia/include/pjmedia/config.h
```diff
@@ -1619,6 +1619,29 @@
...
+/**
+ * Specify the number of keep-alive needed to be sent after the stream is
+ * created.
+ *
+ * Setting this to 0 will disable it.
+ *
+ * Default : 0
+ */
+#ifndef PJMEDIA_STREAM_START_KA_CNT
+#   define PJMEDIA_STREAM_START_KA_CNT	0
```
nanangizz, line 1631: perhaps setting the default to 1-3 is good (e.g: NAT hole punch is good for video to avoid missing keyframe)?

pjmedia/src/pjmedia/stream.c
```diff
@@ -1324,16 +1332,30 @@ static pj_status_t put_frame_imp( pjmedia_port *port,
      */
     if (stream->use_ka)
     {
-	pj_uint32_t dtx_duration;
+        if (stream->start_ka_count) {
```
sauwming, line 1335: since ka already moved here, do we still need the one in stream_create()?

pjmedia/include/pjmedia/config.h
trengginas: done

pjmedia/include/pjmedia/config.h
trengginas: done

pjmedia/src/pjmedia/stream.c
trengginas: I suggest keep sending keepalive if start_ka_count is 0.

pjmedia/src/pjmedia/stream.c
trengginas: use ```last_frm_ts_sent``` into interval calculation

pjmedia/src/pjmedia/stream.c
sauwming: Okay. But will need to explain it in the doc.

nanangizz: The first KA here seems to be sent only after the configured interval time has elapsed (cmiiw), if that's the case, it won't do hole punching (before any real RTP packet is sent), so I think the one in stream_create() should still be sent even when start_ka_count==0, its main purpose is hole punching not purely KA.

nanangizz: oops sorry, it should be `even when start_ka_count != 0`. Hole punching should be done as early as possible, and it does exactly that, IMO.

Btw, the PR title/desc should mention about hole punching too somewhere perhaps, as the main purpose seems about hole punching instead of KA (just piggyback-ing KA mechanism)?

nanangizz (approved):

sauwming (approved):

PR #2523, opened 2020-09-15
Create codeql-analysis.yml

ismangil:
Setting up code scanning provided by GitHub

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         # Override automatic language detection by changing the below list
         # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
-        language: ['cpp', 'python', 'csharp', 'java']
+        language: ['cpp', 'python']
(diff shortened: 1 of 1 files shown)


ismangil:
GitHub is offering [code scanning](https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning) for security.

sauwming:
The Csharp analysis failed, but perhaps we can remove it. We only have one C# sample code in our library.
For Java though, not sure why it failed.

ismangil:
> 
> 
> The Csharp analysis failed, but perhaps we can remove it. We only have one C# sample code in our library.

Disabled.

> For Java though, not sure why it failed.

It needs a build command, it can't figure out the autobuild. Disabled for now.

sauwming (approved):

nanangizz:
Cannot find the build log, how can we know whether the code is built properly or whether it analyses the correct code. From the log, the cpp build took about 4 minutes, seems about right though.

nanangizz (approved):

Issue #416, opened 2007-11-18
Allow application to handle sending ACK manually


pjsipbot:
_2007-11-18 15:13:28: @bennylp_ _created the issue on trac ticket [416](https://trac.pjsip.org/repos/ticket/416)_

Please see "3PCC call transfer implementation" thread on PJSIP mailing list archive for discussion (http://lists.pjsip.org/pipermail/pjsip_lists.pjsip.org/2007-November/000713.html)
___
_2007-11-25 04:41:35: @bennylp_ _changed **status** from **new** to **closed**_

___
_2007-11-25 04:41:35: @bennylp_ _set **resolution** to **fixed**_

___
_2007-11-25 04:41:35: @bennylp_ _commented_

Implemented in [r1598](../commit/d5f9f42a79e0c32de63934fbd7df9e883e462e8f):

 - Added `on_send_ack()` callback in `pjsip_inv_callback` to let application handle ACK transmission manually.
 - Added `pjsip_inv_create_ack()` to create the ACK request.

___
_2007-11-25 04:46:35: @bennylp_ _commented_

Here's a snippet to use the callback and to send ACK:

```
static int saved_cseq;

static void on_send_ack(pjsip_inv_session *inv, pjsip_rx_data *rdata)
{
    if (saved_cseq == 0)
	saved_cseq = rdata->msg_info.cseq->cseq;
}

static void send_the_ack(pjsip_inv_session *inv)
{
  pjsip_tx_data *ack;
  pj_status_t status;

  // See if we need to send SDP with the ACK	
  if (pjmedia_sdp_neg_get_state(inv->neg)==PJMEDIA_SDP_NEG_STATE_REMOTE_OFFER) {
    pjsip_inv_set_sdp_answer(inv, answer);
  }

  // Create and send ACK
  status = pjsip_inv_create_ack(inv, saved_cseq, &ack);
  if (status == PJ_SUCCESS)
     pjsip_inv_send_msg(inv, ack);
}
```

PR #2528, opened 2020-09-21
Invite state not shifting to CONFIRMED state after sending ACK manually

nanangizz:
Feature of allowing app to send ACK manually was added in #416. Unfortunately, there seem to be a bug in the feature, the invite session will not go to the CONFIRMED state and stay in the CONNECTING state after sending the manual ACK. **Only after** an incoming retransmission of the final response (so the library automatically sends the cached ACK), the invite session state will shift to the CONFIRMED state.

Thanks to Thomas Cresson for the report.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -3283,6 +3283,32 @@ PJ_DEF(pj_status_t) pjsip_inv_send_msg( pjsip_inv_session *inv,
 	    goto on_error;
 	}
 
+	/* Check if this is delayed manual ACK (see #416) */
+	if (tdata->msg->line.req.method.id == PJSIP_ACK_METHOD &&
+	    mod_inv.cb.on_send_ack && tdata == inv->last_ack)
+	{
+		pjsip_dlg_inc_lock(inv->dlg);
+
+		/* Set state to CONFIRMED (if we're not in CONFIRMED yet).
+		 * But don't set it to CONFIRMED if we're already DISCONNECTED
+		 * (this may have been a late 200/OK response.
+		 */
+		if (inv->state < PJSIP_INV_STATE_CONFIRMED) {
+			pjsip_event ack_e;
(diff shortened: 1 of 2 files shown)


sauwming (approved):

trengginas (approved):

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -3283,6 +3283,32 @@ PJ_DEF(pj_status_t) pjsip_inv_send_msg( pjsip_inv_session *inv,
 	    goto on_error;
 	}
 
+	/* Check if this is delayed manual ACK (see #416) */
+	if (tdata->msg->line.req.method.id == PJSIP_ACK_METHOD &&
+	    mod_inv.cb.on_send_ack && tdata == inv->last_ack)
```
thomasCresson: Will the condition **tdata == inv->last_ack** work if the ack request is created without using **pjsip_inv_create_ack** ?

thomasCresson (commented):
just a small comment on the condition transitioning the session to CONFIRMED.

pjsip/src/pjsip-ua/sip_inv.c
nanangizz: Good point.

No, it won't. ACK is supposed to be created using `pjsip_inv_create_ack()`, where `inv->last_ack` will be set and used by ACK auto retransmission, unfortunately the `on_send_ack` docs does not mention it either (to be updated).

Btw, is there any sample case that app does not want to (or even cannot) use `pjsip_inv_create_ack()`?

thomasCresson: I cannot think of any case where you wouldn't use the provided API to construct such a message. The only thing is that if it **has** to be used, then, IMHO, it should be mentioned somewhere.

But as you said in your comment, if the doc is updated with this piece of information, then it's OK !

thomasCresson (approved):

PR #2534, opened 2020-10-01
Modify on_new_session callback as optional.

trengginas:
Currently ```on_new_session()``` of ```pjsip_inv_callback``` is treated as mandatory forcing it to be implemented.
However, it's not yet implemented and would be safe to set it as optional

### DIFF
--- a/pjsip/include/pjsip-ua/sip_inv.h
+++ b/pjsip/include/pjsip-ua/sip_inv.h
@@ -131,7 +131,7 @@ typedef struct pjsip_inv_callback
      * This callback is called when the invite usage module has created 
      * a new dialog and invite because of forked outgoing request.
      *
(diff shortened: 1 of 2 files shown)


nanangizz:
Perhaps it is intentional? Perhaps forking should be handled by app as no default action will/can be provided by invite session module.

Note that UA layer seems to provide callback for forked dialog, i.e: `on_dlg_forked`. If it is not implemented, the dialog module (and dialog usage including invite session) may need to handle it.

nanangizz:
Forking scenario does not seem to be a common issue so far, so I think we can merge this. Just perhaps update the docs with something like "this callback is not implemented yet".

nanangizz (approved):

sauwming (approved):

PR #2537, opened 2020-10-01
Add ios arm64e arch support

efendizadeh:
${ARCH} env variable sets with ${CFLAGS} -arch parameter also --host parameter for cross-compile. Apple's new arm64e --host parameter is the same as arm64 --host param but CFLAGS -arch param is arm64e. This pr adds condition if ARCH is arm64e to set correct --host param for aconfigure script


sauwming:
We need to test if PJSIP can run on this architecture.
https://developer.apple.com/documentation/security/preparing_your_app_to_work_with_pointer_authentication

sauwming (approved):

nanangizz (approved):

PR #2538, opened 2020-10-02
Add option to disable ice mismatch check.

trengginas:
RFC [5245 ](https://tools.ietf.org/html/rfc5245#section-5.1) states that:
> The agent will proceed with the ICE procedures defined in this
   specification if, for each media stream in the SDP it received, the
   default destination for each component of that media stream appears
   in a candidate attribute.  For example, in the case of RTP, the IP
   address and port in the c and m lines, respectively, appear in a
   candidate attribute and the value in the rtcp attribute appears in a
   candidate attribute.

If the IP address and port in c and m lines doesn't appears in the candidate attribute, 
then ice-mismatch attribute will be returned.

The updated RFC [8455](https://tools.ietf.org/html/rfc8445#section-5.4) states that:
> Certain middleboxes, such as ALGs, can alter signaling information in
   ways that break ICE (e.g., by rewriting IP addresses in SDP).  This
   is referred to as "ICE mismatch".  If the using protocol is
   vulnerable to ICE mismatch, the responding agent needs to be able to
   detect it and inform the peer ICE agent about the ICE mismatch.

In some scenario, the IP address and port check is not required and might break ICE.

This ticket will allow to disable the check so that the ICE can continue.

### DIFF
--- a/pjmedia/include/pjmedia/transport_ice.h
+++ b/pjmedia/include/pjmedia/transport_ice.h
@@ -157,10 +157,11 @@ enum pjmedia_transport_ice_options
     /**
      * The standard (rfc5245) specify that ice-mismatch attribute is used
      * due to a mismatch of candidates with the default destination for media
-     * signaled in the SDP. However rfc8445 specify that ALGs, can alter
-     * signaling information in ways that break ICE and triggered ice mismatch.
-     * (e.g., by rewriting IP addresses in SDP). So disabling ice mismatch
-     * is required on some scenario.
+     * signaled in the SDP. The purpose is to identify some poorly ALGs that
+     * alter signaling information in ways that break ICE
+     * (e.g., by rewriting IP addresses in SDP). Specifying this option is
+     * to disable the ice mismatch check and allow ICE to continue
+     * if such scenario occur.
      */
     PJMEDIA_ICE_DISABLE_ICE_MISMATCH = 2
 };
--- a/pjmedia/include/pjmedia/transport_ice.h
+++ b/pjmedia/include/pjmedia/transport_ice.h
@@ -152,7 +152,17 @@ enum pjmedia_transport_ice_options
      * are different after several packets are received.
      * Specifying this option will disable this feature.
      */
-    PJMEDIA_ICE_NO_SRC_ADDR_CHECKING = 1
+    PJMEDIA_ICE_NO_SRC_ADDR_CHECKING = 1,
+
+    /**
+     * The standard (rfc5245) specify that ice-mismatch attribute is used
+     * due to a mismatch of candidates with the default destination for media
+     * signaled in the SDP. However rfc8445 specify that ALGs, can alter
+     * signaling information in ways that break ICE and triggered ice mismatch.
+     * (e.g., by rewriting IP addresses in SDP). So disabling ice mismatch
+     * is required on some scenario.
+     */
+    PJMEDIA_ICE_DISABLE_ICE_MISMATCH = 2
 };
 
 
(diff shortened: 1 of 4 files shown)


pjmedia/include/pjmedia/transport_ice.h
```diff
@@ -152,7 +152,17 @@ enum pjmedia_transport_ice_options
...
-    PJMEDIA_ICE_NO_SRC_ADDR_CHECKING = 1
+    PJMEDIA_ICE_NO_SRC_ADDR_CHECKING = 1,
+
+    /**
+     * The standard (rfc5245) specify that ice-mismatch attribute is used
+     * due to a mismatch of candidates with the default destination for media
+     * signaled in the SDP. However rfc8445 specify that ALGs, can alter
+     * signaling information in ways that break ICE and triggered ice mismatch.
+     * (e.g., by rewriting IP addresses in SDP). So disabling ice mismatch
+     * is required on some scenario.
```
nanangizz, line 163: The rfc5245 seems to have identified the ALG issues as well:
> Unfortunately, many ALGs are known to work poorly in these corner
>    cases.  ICE does not try to work around broken ALGs, as this is
>    outside the scope of its functionality.  ICE can help diagnose these
>    conditions, which often show up as a mismatch between the set of
>    candidates and the m and c lines and rtcp attributes.  The ice-
>    mismatch attribute is used for this purpose.

And rfc8445 does not seems to cancel ice-mismatch rule.

So I think the docs needs to be updated.

nanangizz (changes requested):
also the patch seems to fail in the sipp tests, please check.

pjmedia/src/pjmedia/transport_ice.c
```diff
@@ -1070,17 +1072,17 @@ static pj_status_t verify_ice_sdp(struct transport_ice *tp_ice,
 
 	if (!comp1_found && cand.comp_id==COMP_RTP)
 	{
-            if ((tp_ice->options & PJMEDIA_ICE_DISABLE_ICE_MISMATCH) ||
-                ((tp_ice->options & PJMEDIA_ICE_DISABLE_ICE_MISMATCH == 0)
-                 && pj_sockaddr_cmp(&rem_conn_addr, &cand.addr) == 0))
+            if ((disable_ice_mismatch) ||
+                (!disable_ice_mismatch &&
```
nanangizz, line 1076: perhaps a bit optimization, by removing `!disable_ice_mismatch` condition?

pjmedia/src/pjmedia/transport_ice.c
```diff
@@ -1070,17 +1072,17 @@ static pj_status_t verify_ice_sdp(struct transport_ice *tp_ice,
...
             {
                 comp1_found = PJ_TRUE;
             }
 	} else if (!comp2_found && cand.comp_id==COMP_RTCP)
 	{
-            if ((tp_ice->options & PJMEDIA_ICE_DISABLE_ICE_MISMATCH) ||
-                ((tp_ice->options & PJMEDIA_ICE_DISABLE_ICE_MISMATCH == 0)
-                 && pj_sockaddr_cmp(&rtcp_addr, &cand.addr) == 0))
+            if ((disable_ice_mismatch) ||
+                (!disable_ice_mismatch &&
```
nanangizz, line 1084: same as above

nanangizz (approved):
just a minor/optional note.

pjmedia/src/pjmedia/transport_ice.c
trengginas: done

pjmedia/src/pjmedia/transport_ice.c
trengginas: done

sauwming (approved):

PR #2539, opened 2020-10-02
Server Name Indication on TLS doesn't allow IP literal.

trengginas:
According to [RFC3546](https://tools.ietf.org/html/rfc3546#section-3.1)

> Literal IPv4 and IPv6 addresses are not permitted in "HostName"

This ticket will will set server name only if it is not IP literal.

Thanks to Guru Thodime for the report and suggested fix

### DIFF
--- a/pjsip/src/pjsip/sip_transport_tls.c
+++ b/pjsip/src/pjsip/sip_transport_tls.c
@@ -171,28 +171,6 @@ static void tls_on_destroy(void *arg);
 
 static void wipe_buf(pj_str_t *buf);
 
-/*
- * Internal:
- *  determine if an address is a valid IP address, and if it is,
- *  return the IP version (4 or 6).
- */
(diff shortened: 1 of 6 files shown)


nanangizz (commented):
Shouldn't the fix be in the PJLIB SSL socket (for all backends unfortunately) instead of SIP TLS transport?

pjsip/src/pjsip/sip_transport_tls.c
```diff
@@ -171,7 +171,6 @@ static void tls_on_destroy(void *arg);
 
 static void wipe_buf(pj_str_t *buf);
 
-
```
nanangizz, line 174: clean up the patch a bit, supposed to be no change in this file?

nanangizz (approved):
just a minor.

sauwming (approved):

Issue #2547, opened 2020-10-12
run it Centos or debian error....Assertion `frame->size == conf->samples_per_frame ....but Ubuntu is OK!


netyuchangfa:
Ubuntu os RUN OK!
But Centos or debian error...
pjsua: ../src/pjmedia/conference.c:1895: get_frame: Assertion `frame->size == conf->samples_per_frame * conf->bits_per_sample / 8' failed.

Issue #2401, opened 2020-04-23
Pjsip 2.10 crashes when running pjsua2_demo or making a real call.


Syirrus:
**The Problem**
When attempted to make a call (m) via the URL (i.e. sip:user@domain) or running the pjsua2_demo, I get the following error:
` pjsua-armv6l-unknown-linux-gnueabihf: ../src/pjmedia/conference.c:1895: get_frame: Assertion frame->size == conf->samples_per_frame * conf->bits_per_sample / 8' failed.
Aborted`

**To Reproduce**
1. run pjsua2_dmeo app without parameters or make a call (m) via URL. 
2. Error occurs.

**Expected behavior**
I'm expecting the demo to not crash with this error. When actually trying to call someone I get the same error. 

**Environment**
- Raspian armv6 on dietpi
- PJSIP
  - version: 2.10
  - configured with the following: ./configure && make dep && make && make install
  - `config_site.h` contents: no additions

**Smartphone (please also complete the following information):**
- Device: Raspberry Pi Zero W

**Additional context**
Everything compiled successfully and I have compiled multiple times.  The only reference I see to this issue was posted back in 2012 and the suggested patch has already been implemented in 2.10
http://lists.pjsip.org/pipermail/pjsip_lists.pjsip.org/2012-May/034967.html

The sound card and mic work in command line just fine.  
I'm using https://github.com/respeaker/seeed-voicecard

sauwming:
You probably need to enable resampling or make sure that the sound device and conference's clock rates are the same to avoid resampling.

luismartingil:
+1 Getting same error in a Centos8 box

```
sudo yum install -y alsa-lib alsa-lib-devel alsa-utils binutils boost-devel bzip2 gcc gcc-c++ libevent-devel make openssl-devel nv-codec-headers platform-python-devel python3-gevent
pushd /tmp
VER=2.10
wget -qO- https://github.com/pjsip/pjproject/archive/$VER.tar.gz | tar zxv
cd pjproject-*
export CFLAGS="$CFLAGS -fPIC" && ./configure --disable-libyuv
make dep
make
sudo make install
sudo cp /tmp/pjproject-*/pjsip-apps/bin/pjsua-x86_64-unknown-linux-gnu /usr/bin/pjsua
sudo cp /tmp/pjproject-*/pjsip-apps/bin/pjsystest-x86_64-unknown-linux-gnu /usr/bin/pjsystest
popd
[...]

Issue #2345, opened 2020-03-13
pjsua get_frame fails


ForeverACE:
I am encountering an error that seems similar to one that has already been resolved (i.e. ([https://trac.pjsip.org/repos/ticket/1510](https://trac.pjsip.org/repos/ticket/1510)). When I try to run pjsua, I get the following error:

**pjsua-armv7l-unknown-linux-gnueabihf: ../src/pjmedia/conference.c:1895: get_frame: Assertion 'frame->size == conf->samples_per_frame * conf->bits_per_sample / 8' failed.**

**Background**
- Hardware: Raspberry Pi 3B
- OS: Official Raspbian 10 (Buster) minimal image
- Audio H/W: plugable USB audio adapter (uses CMedia's HS-100B chipset - [https://www.cmedia.com.tw/products/USB20_FULL_SPEED/HS100B](https://www.cmedia.com.tw/products/USB20_FULL_SPEED/HS100B))
- PJSIP version 2.10

**Compile Steps**
`./configure --disable-video --disable-ffmpeg --disable-v4l2 && make dep && make`

**Contents of custom.cfg**
`--id=sip:300@10.0.0.150
--registrar=sip:10.0.0.150
--realm=*
--username=300
--password=my_secret_password
--local-port=5060
--auto-answer=200
--auto-loop
--max-calls=1
--capture-dev=14
--playback-dev=14
`

**Command**
`./pjsua-armv7l-unknown-linux-gnueabihf --config-file=custom.cfg`

**Output/Console**
--------Console Output Start--------
13:55:13.429 os_core_unix.c !pjlib 2.10 for POSIX initialized
13:55:13.432 sip_endpoint.c  .Creating endpoint instance...
13:55:13.433          pjlib  .select() I/O Queue created (0x4e1310)
13:55:13.433 sip_endpoint.c  .Module "mod-msg-print" registered
13:55:13.434 sip_transport.  .Transport manager created.
13:55:13.434   pjsua_core.c  .PJSUA state changed: NULL --> CREATED
13:55:13.434 sip_endpoint.c  .Module "mod-pjsua-log" registered
13:55:13.434 sip_endpoint.c  .Module "mod-tsx-layer" registered
13:55:13.434 sip_endpoint.c  .Module "mod-stateful-util" registered
13:55:13.434 sip_endpoint.c  .Module "mod-ua" registered
13:55:13.435 sip_endpoint.c  .Module "mod-100rel" registered
[...]

Issue #2339, opened 2020-03-09
USB Soundcards + choppy playback + no capture (Linux, arm + x86_64)


forca83:
**Describe the bug**
With any extern usb audio card we did try:
`Jabra Speak 410 ,  0b0e:0412 GN`
`Netcom, Native Union, 0c76:1607 JMTek, LLC. audio controller`
 we have "choppy playback" + "no capture" with pjsip v2.10

The caller-side has many suspected flooding messages:
`Master/sound  Underflow, buf_cnt=0, will generate 1 frame`

_The issue is, usb sound cards have "choppy playback" and "not working mic"._
_The onboard soundcard on our imx6 board works well._

**Describe the hw/sw**
SW: For testing used pjsua-app from pjsip v2.10
HW: imx6 + Yocto Poky 3.0.2

**To Reproduce**
start on caller side (side 1) pjsua-app:
`./pjsua-arm-poky-linux-gnu --log-level=5 --add-codec=OPUS --stereo --snd-auto-close=-1 --no-vad --playback-dev=2 --capture-dev=2 --ec-tail=0 --ec-opt=0 sip:abc@10.1.63.6`

start on calle side (side 2) pjsua-app:
`./pjsua-arm-poky-linux-gnu --log-level=3 --add-codec=OPUS --stereo --snd-auto-close=4 --clock-rate=48000 --snd-clock-rate=48000 --ec-tail=30 --ec-opt=322 --no-vad --no-udp --playback-dev=0 --capture-dev=0`

**Logs/Screenshots (caller, side 1)**
**----------------------------------------------------------------------------------------------------------------**

```
./pjsua-arm-poky-linux-gnu --log-level=5 --add-codec=OPUS --stereo --snd-auto-close=-1 --no-vad --playback-dev=2 --capture-dev=2 --ec-tail=0 --ec-opt=0 sip:abc@10.1.63.6
17:17:00.944 os_core_unix.c !pjlib 2.10 for POSIX initialized
17:17:00.947 sip_endpoint.c  .Creating endpoint instance...
17:17:00.947          pjlib  .select() I/O Queue created (0x5de668)
17:17:00.947 sip_endpoint.c  .Module "mod-msg-print" registered
17:17:00.947 sip_transport.  .Transport manager created.
17:17:00.947   pjsua_core.c  .PJSUA state changed: NULL --> CREATED
17:17:00.948 sip_endpoint.c  .Module "mod-pjsua-log" registered
[...]

PR #2549, opened 2020-10-13
ALSA dev incorrectly sets the number of samples per frame

sauwming:
The fix of issue #2223 seems to be causing an adverse effect, which is that it will set the number of samples per frame incorrectly. This will trigger an assertion in the conference, as reported in:
https://github.com/pjsip/pjproject/issues/2345, https://github.com/pjsip/pjproject/issues/2401, and https://github.com/pjsip/pjproject/issues/2547

It also causes audio issue such as reported in:
https://github.com/pjsip/pjproject/issues/2339

Managed to reproduce the issue on CentOS 8, which will return period size of 341 for samples per frame of 320 (16 KHz 20 ms). Reverting issue #2223 works, and doesn't seem to cause the problem described there (i.e. crash due to incorrect buffer size). `snd_pcm_readi()` and `snd_pcm_writei()` has `nframes` as a parameter, so I'm not clear how or why ALSA can cause buffer overflow.

Original report for issue #2223 can be found here:
http://lists.pjsip.org/pipermail/pjsip_lists.pjsip.org/2019-August/041426.html

So for now, I propose to revert issue #2223 first, until a proper, or better fix is found.

### DIFF
--- a/pjmedia/src/pjmedia-audiodev/alsa_dev.c
+++ b/pjmedia/src/pjmedia-audiodev/alsa_dev.c
@@ -702,6 +702,9 @@ static pj_status_t open_playback (struct alsa_stream* stream,
 	tmp_buf_size = (rate / 1000) * param->output_latency_ms;
     else
 	tmp_buf_size = (rate / 1000) * PJMEDIA_SND_DEFAULT_PLAY_LATENCY;
+    if (tmp_buf_size < tmp_period_size * 2)
+        tmp_buf_size = tmp_period_size * 2;
+    tmp_buf_size *= param->channel_count;
     snd_pcm_hw_params_set_buffer_size_near (stream->pb_pcm, params,
 					    &tmp_buf_size);
     stream->param.output_latency_ms = tmp_buf_size / (rate / 1000);
@@ -825,6 +828,9 @@ static pj_status_t open_capture (struct alsa_stream* stream,
 	tmp_buf_size = (rate / 1000) * param->input_latency_ms;
     else
 	tmp_buf_size = (rate / 1000) * PJMEDIA_SND_DEFAULT_REC_LATENCY;
+    if (tmp_buf_size < tmp_period_size * 2)
+        tmp_buf_size = tmp_period_size * 2;
+    tmp_buf_size *= param->channel_count;
     snd_pcm_hw_params_set_buffer_size_near (stream->ca_pcm, params,
 					    &tmp_buf_size);
     stream->param.input_latency_ms = tmp_buf_size / (rate / 1000);
--- a/pjmedia/src/pjmedia-audiodev/alsa_dev.c
+++ b/pjmedia/src/pjmedia-audiodev/alsa_dev.c
@@ -689,8 +689,11 @@ static pj_status_t open_playback (struct alsa_stream* stream,
     tmp_period_size = stream->pb_frames;
     snd_pcm_hw_params_set_period_size_near (stream->pb_pcm, params,
 					    &tmp_period_size, NULL);
-    stream->pb_frames = tmp_period_size > stream->pb_frames ? tmp_period_size : 
-                                                              stream->pb_frames;					    					    
+    /* Commenting this as it may cause the number of samples per frame
+     * to be incorrest.
+     */  
+    // stream->pb_frames = tmp_period_size > stream->pb_frames ?
+    //			tmp_period_size : stream->pb_frames;					    					    
     TRACE_((THIS_FILE, "open_playback: period size set to: %d",
 	    tmp_period_size));
 
@@ -809,8 +812,11 @@ static pj_status_t open_capture (struct alsa_stream* stream,
     tmp_period_size = stream->ca_frames;
     snd_pcm_hw_params_set_period_size_near (stream->ca_pcm, params,
 					    &tmp_period_size, NULL);
-    stream->ca_frames = tmp_period_size > stream->ca_frames ? tmp_period_size : 
-                                                              stream->ca_frames;
+    /* Commenting this as it may cause the number of samples per frame
+     * to be incorrest.
+     */
+    // stream->ca_frames = tmp_period_size > stream->ca_frames ?
+    //			tmp_period_size : stream->ca_frames;
     TRACE_((THIS_FILE, "open_capture: period size set to: %d",
 	    tmp_period_size));
 


nanangizz:
From [this article](https://www.alsa-project.org/main/index.php/FramesPeriods), `period_size` seems to be the number of samples to be populated before notifying app (via hardware interrupt), so if it is 341, the app will be notified every time samples reach multiplication of 341. If the app read 320 samples only (less than 341) in each notification, it may eventually lead to buffer overflow in ALSA capture buffer (due to lower app consuming rate)? Our code uses `snd_pcm_readi()` instead of notification, but still not sure whether the function will return based on notification or buffer availability. From other articles/posts, ALSA behavior seems vary among audio hardware drivers.

Perhaps it is safer to regard the returned period size, here is a raw idea. Employ a (ring) buffer to bridge ALSA & PJMEDIA period/frame size. So when ALSA returns period size of 341 or 300 (for a requested size 320), we read/write 341 or 320 samples from/to ALSA (which one is bigger, as in the current code), while we invoke callbacks to PJMEDIA by 320 samples (or the configured samples per frame).

Also note that `snd_pcm_hw_params_set_period_size_near()` has `dir` param, we always set it to `NULL`, while perhaps it's better to set it appropriately, so the period/frame size may be different between capture and playback.

sauwming:
During my testing, `snd_pcm_readi()` seems to behave according to the doc https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#ga4c2c7bd26cf221268d59dc3bbeb9c048:
```If the blocking behaviour was selected and it is running, then routine waits until all requested frames are filled. The returned number of frames can be less only if a signal or underrun occurred.```
So it returned based on buffer availability and in my case, always returned exactly 320 frames as requested.

As for the dir, I also tried it before and it seemed to be an output-only parameter. If dir is 1, it means the actual period size is > than requested, 0 exact, -1 smaller. I forced set it to 0 and -1 on input, but it had no effect, and always returned 1.

The article gave me an idea on what could be the problem with the issue reported in #2223. Note that It mentions that ALSA's internal ring buffer size is commonly 2 * period size. It's also this way in the official example:
https://www.alsa-project.org/alsa-doc/alsa-lib/_2test_2latency_8c-example.html#a11
But in that example, as well as in our implementation, it can also be set according to the desired latency. So, if, say, someone aims for a minimum latency that is smaller than 2 * period size, it may cause problem with ALSA's internal ring buffer since it can't accommodate more than one period.

So perhaps the fix should be:
```tmp_buf_size = max(tmp_buf_size, tmp_period_size * 2)
snd_pcm_hw_params_set_buffer_size_near(tmp_buf_size);
```
This way, I believe we can utilize ALSA's internal ring buffer effectively and do not need to waste memory/performance to create a duplicate ring buffer in our PJMEDIA.

sauwming:
In my environment, ALSA refused the buffer to be set lower than 2 * period size (so the minimum is 341 * 2 = 682). So the issue could be due to a bug (or different behaviour) in the original reporter's ALSA driver instead.

nanangizz:
> During my testing, `snd_pcm_readi()` seems to behave according to the doc https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#ga4c2c7bd26cf221268d59dc3bbeb9c048:
> `If the blocking behaviour was selected and it is running, then routine waits until all requested frames are filled. The returned number of frames can be less only if a signal or underrun occurred.`
> So it returned based on buffer availability and in my case, always returned exactly 320 frames as requested.

Okay, so the docs explicitly says it is based on buffer availability.

> As for the dir, I also tried it before and it seemed to be an output-only parameter. If dir is 1, it means the actual period size is > than requested, 0 exact, -1 smaller. I forced set it to 0 and -1 on input, but it had no effect, and always returned 1.

Ha, thought `dir` was for specifying capture or playback.

> The article gave me an idea on what could be the problem with the issue reported in #2223. Note that It mentions that ALSA's internal ring buffer size is commonly 2 * period size. It's also this way in the official example:
> https://www.alsa-project.org/alsa-doc/alsa-lib/_2test_2latency_8c-example.html#a11
> But in that example, as well as in our implementation, it can also be set according to the desired latency. So, if, say, someone aims for a minimum latency that is smaller than 2 * period size, it may cause problem with ALSA's internal ring buffer since it can't accommodate more than one period.
> 
> So perhaps the fix should be:
> 
> ```
> tmp_buf_size = max(tmp_buf_size, tmp_period_size * 2)
> snd_pcm_hw_params_set_buffer_size_near(tmp_buf_size);
> ```
> 
> This way, I believe we can utilize ALSA's internal ring buffer effectively and do not need to waste memory/performance to create a duplicate ring buffer in our PJMEDIA.

Good idea.
Also, the latency/`tmp_buf_size` calculation perhaps needs to include channel count.

Btw, better to also update the reporter of #2223 about the patch reversion & new fix, also asking some info about the setups (e.g: Unix/linux distro/version, ALSA version, audio hardware type, audio params, latency setting) & the numbers such as period size requested vs returned, nframes for read/write, at least for our docs.

nanangizz (approved):

trengginas (approved):

pjmedia/src/pjmedia-audiodev/alsa_dev.c
```diff
@@ -699,6 +702,9 @@ static pj_status_t open_playback (struct alsa_stream* stream,
 	tmp_buf_size = (rate / 1000) * param->output_latency_ms;
     else
 	tmp_buf_size = (rate / 1000) * PJMEDIA_SND_DEFAULT_PLAY_LATENCY;
+    if (tmp_buf_size < tmp_period_size * 2)
+        tmp_buf_size = tmp_period_size * 2;
+    tmp_buf_size *= param->channel_count;
```
mike2307, line 707: In my opinion this is **not** correct.

The function `snd_pcm_hw_params_set_buffer_size_near()` expects the buffer size in frames. In ALSA terms a frame is the sum of all samples for on given point in time. e.g. For a stereo device 1 frame would consist of 2 samples. [This article](https://www.linuxjournal.com/article/6735) gives a good overview.

Because `snd_pcm_hw_params_set_buffer_size_near()` expects the number of frames as argument, there's no point in multiplying the buffer size with the channel count. The effect will be an increased buffer latency, dependent on the channel count, which is obviously not on purpose.

The same applies to line 833.

sauwming: You're right, we missed it. The article: https://www.alsa-project.org/main/index.php/FramesPeriods mentioned by @nanangizz above actually also says the same thing.
"1 frame represents 1 analog sample from all channels".

Thanks for noticing. Created PR #2549 for this.

nanangizz: Good catch! Thanks @mike2307.

Btw, the new PR ID should be #2556?

Issue #2246, opened 2019-10-16
Specify audio/video direction when making/answering call


pjsipbot:
_2019-10-16 04:35:55: @nanangizz_ _created the issue on trac ticket [2246](https://trac.pjsip.org/repos/ticket/2246)_

Currently audio/video direction in call initialization is always send & receive, and from time to time, users have asked how to change the behavior. Currently a not-straightforward way is by modifying the SDP directly via PJSUA callback `on_call_sdp_created()` or PJSUA2 callback `Call::onCallSdpCreated()`. So adding settings for that should be useful.

andreas-wehrmann:
I'd really like to have this feature, as I'm currently doing it in the SDP callbacks "by hand".

Another feature that might come in really handy would be the ability
to "filter" codecs to be used in a call (on a per call basis).

For outgoing calls, one could limit the set of codecs to be proposed.
One might do this because:
- They might want to avoid resampling and choose the codecs accordingly
- Use special codecs for special endpoints

For incoming calls, it would be great if PJSUA could somehow provide the list of codecs proposed
with the ability to let the user influence which codecs get passed on for further negotiation.
For this, a 'prefilter' would need to occur, to see which codecs are currently enabled
and could theoretically be accepted before passing it to the user.

I realize this doesn't fit this issue, but I just wanted to throw this in somewhere.

PR #2550, opened 2020-10-14
Fix bad Video SDP when using 3rdPartyMedia and PJMEDIA_DIR_ENCODING only

inf265:
When using PJMEDIA_DIR_ENCODING with 3rd Party Media in the video structure, the SDP part of the INVITE is not set properly. This commit omits the direction to be PJMEDIA_DIR_ENCODING_DECODING.

### DIFF
--- a/pjmedia/src/pjmedia/endpoint.c
+++ b/pjmedia/src/pjmedia/endpoint.c
@@ -773,9 +773,8 @@ PJ_DEF(pj_status_t) pjmedia_endpt_create_video_sdp(pjmedia_endpt *endpt,
 	    break;
 	}
 
-	/* Must support RTP packetization and bidirectional */
-	if ((codec_info[i].packings & PJMEDIA_VID_PACKING_PACKETS) == 0 ||
-	    codec_info[i].dir != PJMEDIA_DIR_ENCODING_DECODING)
+        /* Must support RTP packetization */
+        if ((codec_info[i].packings & PJMEDIA_VID_PACKING_PACKETS) == 0)
 	{
 	    continue;
 	}


sauwming:
Checking the code, it seems that the snippet is there to check if the video codec supports packing packetization and can encode and decode (i.e. bidirectional). If not, it will be skipped (not displayed in the SDP). We have quite a number of users that use third party media and so far this has never been an issue. So are you registering custom video codecs which only supports one direction?

inf265:
Yes we have only one direction, it's about a doorbell who only sends video. BTW, I attached the alt_pjsua_vid.c so that you can see  how an example struct alt_codec_desc looks like. I added also Makefile since the make in pjproject/pjsip-apps/src/3rdparty_media_sample throws linker errors.
[alt_pjsua_vid.c.txt](https://github.com/pjsip/pjproject/files/5415534/alt_pjsua_vid.c.txt)
[Makefile.txt](https://github.com/pjsip/pjproject/files/5415537/Makefile.txt)

Another thing is also, when I fix this issue with the commit, I would expect in SDP the direction to be only "sendonly" but it is "sendrecv". I can do a workround in onCallSdpCreated but it's not nice.

inf265:
How is the further procedure for this ?

sauwming (approved):

sauwming:
Sorry for the delay. After an internal discussion, we agree that codec can be unidirectional and the assertion can be safely removed. As for the video direction, it won't be determined by the codec direction and shall be fixed in another issue #2246 .

nanangizz (approved):

PR #2551, opened 2020-10-15
Check hostname during transport selection

sauwming:
Currently, the selection criteria to reuse transport is only based on IP address + port + protocol. In this PR, we add checking hostname as well.

Thanks to Lauri Vänskä for the report and suggestion.


nanangizz (approved):

trengginas (approved):

sauwming:
Please check https://github.com/pjsip/pjproject/security/advisories/GHSA-8hcp-hm38-mfph instead.

PR #2552, opened 2020-10-16
Support Android AMediaCodec

trengginas:
This ticket will implement AMRNB, AMRWB audio codec and AVC (H264), VP8, VP9 video codec supported by Android AMediaCodec.

On Android, AMediaCodec support is auto-detected enabled, represented by these lines on ```config.log``` or output of ```configure-android``` command.
```
Checking if Android AMediaCodec library is available... yes
```
To disable it, specify ```--disable-android-mediacodec``` on ```configure-android``` command.

By enabling the above, by default AMRNB, AMRWB, H264, VP8 and VP9 codec will also be enabled.
They are controllable by these settings.
```
PJMEDIA_HAS_AND_MEDIA_AMRNB
PJMEDIA_HAS_AND_MEDIA_AMRWB
PJMEDIA_HAS_AND_MEDIA_H264
PJMEDIA_HAS_AND_MEDIA_VP8
PJMEDIA_HAS_AND_MEDIA_VP9
```
Notes:
- Most APIs used are supported since for Android-21. Older devices is not recommended to enable this.
- On older devices, some codecs are not supported. e.g: VP9 codec are not supported older devices (Android oreo devices).

### DIFF
--- a/pjmedia/include/pjmedia-codec.h
+++ b/pjmedia/include/pjmedia-codec.h
@@ -42,6 +42,7 @@
 #include <pjmedia-codec/speex.h>
 #include <pjmedia-codec/vid_toolbox.h>
 #include <pjmedia-codec/vpx.h>
+#include <pjmedia-codec/and_aud_mediacodec.h>
 #include <pjmedia-codec/and_vid_mediacodec.h>
 
 #endif	/* __PJMEDIA_CODEC_PJMEDIA_CODEC_H__ */
--- a/pjmedia/include/pjmedia-codec.h
+++ b/pjmedia/include/pjmedia-codec.h
@@ -42,6 +42,7 @@
 #include <pjmedia-codec/speex.h>
 #include <pjmedia-codec/vid_toolbox.h>
 #include <pjmedia-codec/vpx.h>
+#include <pjmedia-codec/and_vid_mediacodec.h>
 
 #endif	/* __PJMEDIA_CODEC_PJMEDIA_CODEC_H__ */
 
--- a/pjmedia/include/pjmedia-codec.h
+++ b/pjmedia/include/pjmedia-codec.h
@@ -24,7 +24,8 @@
  * @file pjmedia-codec.h
  * @brief Include all codecs API in PJMEDIA-CODEC
  */
-
+#include <pjmedia-codec/and_aud_mediacodec.h>
+#include <pjmedia-codec/and_vid_mediacodec.h>
 #include <pjmedia-codec/audio_codecs.h>
 #include <pjmedia-codec/bcg729.h>
 #include <pjmedia-codec/ffmpeg_vid_codecs.h>
@@ -42,8 +43,6 @@
 #include <pjmedia-codec/speex.h>
 #include <pjmedia-codec/vid_toolbox.h>
 #include <pjmedia-codec/vpx.h>
-#include <pjmedia-codec/and_aud_mediacodec.h>
-#include <pjmedia-codec/and_vid_mediacodec.h>
 
 #endif	/* __PJMEDIA_CODEC_PJMEDIA_CODEC_H__ */
 
--- a/pjmedia/include/pjmedia-codec/and_aud_mediacodec.h
+++ b/pjmedia/include/pjmedia-codec/and_aud_mediacodec.h
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C)2020 Teluu Inc. (http://www.teluu.com)
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifndef __PJMEDIA_CODEC_AND_AUD_MEDIACODEC_H__
+#define __PJMEDIA_CODEC_AND_AUD_MEDIACODEC_H__
+
+/**
+ * @file and_aud_mediacodec.h
+ * @brief Android audio MediaCodec codec.
+ */
+
+#include <pjmedia-codec/types.h>
+
+PJ_BEGIN_DECL
+
+/**
+ * Initialize and register Android audio MediaCodec factory to pjmedia
+ * endpoint.
+ *
+ * @param endpt		The pjmedia endpoint.
+ *
+ * @return		PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t) pjmedia_codec_and_media_aud_init( pjmedia_endpt *endpt );
+
+/**
+ * Unregister Android audio MediaCodec factory from pjmedia endpoint 
+ * and deinitialize the codec library.
+ *
+ * @return	    PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t) pjmedia_codec_and_media_aud_deinit( void );
+
+PJ_END_DECL
+
+/**
+ * @}
+ */
+
+#endif	/* __PJMEDIA_CODEC_AND_AUD_MEDIACODEC_H__ */
--- a/pjmedia/include/pjmedia-codec/and_aud_mediacodec.h
+++ b/pjmedia/include/pjmedia-codec/and_aud_mediacodec.h
@@ -45,6 +45,8 @@
  * General codec settings for this codec such as VAD and PLC can be
  * manipulated through the <tt>setting</tt> field in #pjmedia_codec_param.
  * Please see the documentation of #pjmedia_codec_param for more info.
+ * Note that MediaCodec doesn't provide internal VAD/PLC feature, they will be
+ * provided by PJMEDIA instead.
  *
  * \subsubsection bitrate Bitrate
  *
--- a/pjmedia/include/pjmedia-codec/and_aud_mediacodec.h
+++ b/pjmedia/include/pjmedia-codec/and_aud_mediacodec.h
@@ -21,10 +21,84 @@
 
 /**
  * @file and_aud_mediacodec.h
- * @brief Android audio MediaCodec codec.
+ * @brief Android audio MediaCodec codecs.
  */
 
 #include <pjmedia-codec/types.h>
+
+/**
+ * @defgroup PJMEDIA_CODEC_AUD_MEDIACODEC Audio MediaCodec Codec
+ * @ingroup PJMEDIA_CODEC_CODECS
+ * @{
+ *
+ * Audio MediaCodec codec wrapper for Android.
+ *
+ * This codec wrapper contains varius codecs: i.e: AMR and AMR-WB.
+ *
+ * \section pjmedia_codec_mediacodec_AMR MediaCodec AMRNB/AMR-WB
+ *
+ * MediaCodec AMR supports 16-bit PCM audio signal with sampling rate 8000Hz,
+ * 20ms frame length and producing various bitrates that ranges from 4.75kbps
+ * to 12.2kbps.
+ * \subsection codec_setting Codec Settings
+ *
+ * General codec settings for this codec such as VAD and PLC can be
+ * manipulated through the <tt>setting</tt> field in #pjmedia_codec_param.
+ * Please see the documentation of #pjmedia_codec_param for more info.
+ *
+ * \subsubsection bitrate Bitrate
+ *
+ * By default, encoding bitrate is 7400bps. This default setting can be
+ * modified using #pjmedia_codec_mgr_set_default_param() by specifying
+ * prefered AMR bitrate in field <tt>info::avg_bps</tt> of
+ * #pjmedia_codec_param. Valid bitrates could be seen in
+ * #pjmedia_codec_amrnb_bitrates.
+ *
+ * \subsubsection payload_format Payload Format
+ *
+ * There are two AMR payload format types, bandwidth-efficient and
+ * octet-aligned. Default setting is using octet-aligned. This default payload
+ * format can be modified using #pjmedia_codec_mgr_set_default_param().
+ *
+ * In #pjmedia_codec_param, payload format can be set by specifying SDP
+ * format parameters "octet-align" in the SDP "a=fmtp" attribute for
+ * decoding direction. Valid values are "0" (for bandwidth efficient mode)
+ * and "1" (for octet-aligned mode).
+ *
+ * \subsubsection mode_set Mode-Set
+ *
+ * Mode-set is used for restricting AMR modes in decoding direction.
+ *
+ * By default, no mode-set restriction applied. This default setting can be
+ * be modified using #pjmedia_codec_mgr_set_default_param().
+ *
+ * In #pjmedia_codec_param, mode-set could be specified via format parameters
+ * "mode-set" in the SDP "a=fmtp" attribute for decoding direction. Valid
+ * value is a comma separated list of modes from the set 0 - 7, e.g:
+ * "4,5,6,7". When this parameter is omitted, no mode-set restrictions applied.
+ *
+ * Here is an example of modifying AMR default codec param:
+ \code
+    pjmedia_codec_param param;
+
+    pjmedia_codec_mgr_get_default_param(.., &param);
+    ...
+    // set default encoding bitrate to the highest 12.2kbps
+    param.info.avg_bps = 12200;
+
+    // restrict decoding bitrate to 10.2kbps and 12.2kbps only
+    param.setting.dec_fmtp.param[0].name = pj_str("mode-set");
+    param.setting.dec_fmtp.param[0].val  = pj_str("6,7");
+
+    // also set to use bandwidth-efficient payload format
+    param.setting.dec_fmtp.param[1].name = pj_str("octet-align");
+    param.setting.dec_fmtp.param[1].val  = pj_str("0");
+
+    param.setting.dec_fmtp.cnt = 2;
+    ...
+    pjmedia_codec_mgr_set_default_param(.., &param);
+ \endcode
+ */
 
 PJ_BEGIN_DECL
 
--- a/pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
+++ b/pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
@@ -1,0 +1,71 @@
+/* $Id$ */
+/*
+ * Copyright (C)2020 Teluu Inc. (http://www.teluu.com)
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifndef __PJMEDIA_CODEC_AND_VID_MEDIACODEC_H__
+#define __PJMEDIA_CODEC_AND_VID_MEDIACODEC_H__
+
+#include <pjmedia-codec/types.h>
+#include <pjmedia/vid_codec.h>
+
+/**
+ * @file pjmedia-codec/and_vid_mediacodec.h
+ * @brief Android Mediacodec video codec
+ */
+
+PJ_BEGIN_DECL
+
+/**
+ * @defgroup PJMEDIA_HAS_ANDROID_MEDIACODEC Android Mediacodec Codec
+ * @ingroup PJMEDIA_CODEC_VID_CODECS
+ * @{
+ *
+ * Android Mediacodec (AVC/H264+VP8+VP9) codec wrapper for Android.
+ */
+
+/**
+ * Initialize and register Android Mediacodec video codec factory.
+ *
+ * @param mgr	    The video codec manager instance where this codec will
+ * 		    be registered to. Specify NULL to use default instance
+ * 		    (in that case, an instance of video codec manager must
+ * 		    have been created beforehand).
+ * @param pf	    Pool factory.
+ *
+ * @return	    PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t) pjmedia_codec_anmed_vid_init(pjmedia_vid_codec_mgr *mgr,
+                                                  pj_pool_factory *pf);
+
+/**
+ * Unregister Android Mediacodec video codecs factory from the video codec
+ * manager and deinitialize the codec library.
+ *
+ * @return	    PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t) pjmedia_codec_anmed_vid_deinit(void);
+
+
+/**
+ * @}
+ */
+
+
+PJ_END_DECL
+
+#endif	/* __PJMEDIA_CODEC_AND_VID_MEDIACODEC_H__ */
--- a/pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
+++ b/pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
@@ -1,4 +1,3 @@
-/* $Id$ */
 /*
  * Copyright (C)2020 Teluu Inc. (http://www.teluu.com)
  *
@@ -49,8 +48,9 @@ PJ_BEGIN_DECL
  *
  * @return	    PJ_SUCCESS on success.
  */
-PJ_DECL(pj_status_t) pjmedia_codec_anmed_vid_init(pjmedia_vid_codec_mgr *mgr,
-                                                  pj_pool_factory *pf);
+PJ_DECL(pj_status_t) pjmedia_codec_and_media_vid_init(
+				    pjmedia_vid_codec_mgr *mgr,
+                                    pj_pool_factory *pf);
 
 /**
  * Unregister Android Mediacodec video codecs factory from the video codec
@@ -58,7 +58,7 @@ PJ_DECL(pj_status_t) pjmedia_codec_anmed_vid_init(pjmedia_vid_codec_mgr *mgr,
  *
  * @return	    PJ_SUCCESS on success.
  */
-PJ_DECL(pj_status_t) pjmedia_codec_anmed_vid_deinit(void);
+PJ_DECL(pj_status_t) pjmedia_codec_and_media_vid_deinit(void);
 
 
 /**
--- a/pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
+++ b/pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
@@ -34,7 +34,10 @@ PJ_BEGIN_DECL
  * @ingroup PJMEDIA_CODEC_VID_CODECS
  * @{
  *
- * Android Mediacodec (AVC/H264+VP8+VP9) codec wrapper for Android.
+ *
+ * Video MediaCodec codec wrapper for Android.
+ *
+ * This codec wrapper contains varius codecs: i.e: H.264/AVC, VP8 and VP9.
  */
 
 /**
--- a/pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
+++ b/pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
@@ -24,7 +24,7 @@
 
 /**
  * @file pjmedia-codec/and_vid_mediacodec.h
- * @brief Android Mediacodec video codec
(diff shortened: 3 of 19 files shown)


sauwming (commented):
- anmed doesn't sound obvious what it is an abbreviation of. Perhaps a simple `android` or `and_media` would be better.
- can app use this with a fallback option to other codecs (such as OpenH264 or VPX)?
  * If yes, need the doc how to do that, and better, put it in Android sample app.

nanangizz:
> To add AMediaCodec support, include this on the `config_site.h` file.
> 
> ```
> #define PJMEDIA_HAS_ANDROID_MEDIACODEC 1

Why not simply have this setting auto-detect & auto-set ?

> #define PJMEDIA_HAS_ANMED_AVC      /* AVC/H264 codec */

- replace 'AVC' with 'H264'?
- not set to 1?

> #define PJMEDIA_HAS_ANMED_VP8 1.   /*. VP8 codec */
> #define PJMEDIA_HAS_ANMED_VP9 1.   /* VP9 codec */

remove '.'

> ```

Usually native codecs are preferred, so if this is stable enough and no compatibility issue with OpenH264, perhaps better to make them (or at least H264) enabled by default?

pjmedia/include/pjmedia-codec/and_vid_mediacodec.h
```diff
@@ -0,0 +1,71 @@
+/* $Id$ */
```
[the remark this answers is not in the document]
trengginas: > * anmed doesn't sound obvious what it is an abbreviation of. Perhaps a simple `android` or `and_media` would be better.

Change to and_media prefix

> * can app use this with a fallback option to other codecs (such as OpenH264 or VPX)?
>   
>   * If yes, need the doc how to do that, and better, put an example in Android sample app.

The codec (H264/VP8/VP9) will be enabled if the  encoder/decoder is successfully created/supported.
However if they are supported and OpenH264/VPX is enabled then they will be present in the SDP.
This is the same with vidtoolbox (H264) codec, if it's enabled along with OpenH264. Then both codec will be active.

trengginas:
> > To add AMediaCodec support, include this on the `config_site.h` file.
> > ```
> > #define PJMEDIA_HAS_ANDROID_MEDIACODEC 1
> > ```
> 
> Why not simply have this setting auto-detect & auto-set ?

Cons of enabling Android MediaCodec by default:
- Most API used are supported only on API 21. Older device will have building issue.
- Some codec are supported on newer devices. e.g: VP9 is not yet supported on Android 7.0

> 
> > #define PJMEDIA_HAS_ANMED_AVC      /* AVC/H264 codec */
> 
> * replace 'AVC' with 'H264'?

Done

> * not set to 1?

If PJMEDIA_HAS_ANDROID_MEDIACODEC is enabled, then PJMEDIA_HAS_AND_MEDIA_H264 by default enabled.

> 
> > #define PJMEDIA_HAS_ANMED_VP8 1.   /*. VP8 codec _/
> > #define PJMEDIA_HAS_ANMED_VP9 1.   /_ VP9 codec */
> 
> remove '.'
> 

Done

> > 
> 
> Usually native codecs are preferred, so if this is stable enough and no compatibility issue with OpenH264, perhaps better to make them (or at least H264) enabled by default?

Anroid MediaCodec supports software and hardware codec access. Based on tests, the most stable configuration is to use 
software encoder and hardware decoder. e.g: qcom SoC doesn't provides Exynos or other codec.
Currently, the encoder/decoder names are hardcoded and will be enabled if its able to be created.
Last test is for qcom (qualcomm) and Exynos SoC.

On runtime, the logcat showed this message on encoding process (repeatedly):
```
E/NdkMediaCodec: sf error code: -13
```
However, there's no error was returned from the code and the video was successfully rendered on the remote side.

sauwming:
- Are the interoperability tests vs OpenH264 and vs libvpx okay?
- For VPX, how's the encoding and decoding performance compared to libvpx? Can it support high resolution without lagging?

trengginas:
> * Are the interoperability tests vs OpenH264 and vs libvpx okay?

Tested with OpenH264, vidtoolbox (H264) and libvpx with okay result.

> * For VPX, how's the encoding and decoding performance compared to libvpx? Can it support high resolution without lagging?

Using both parties using Android MediaCodec wiith 1280x720. VP8 showed minimal lagging, however VP9 lagging is very noticeable.

pjmedia/src/pjmedia-codec/and_aud_mediacodec.cpp
```diff
@@ -0,0 +1,1347 @@
...
+
+    PJ_LOG(5,(THIS_FILE, "Opening codec.."));
+
+    codec_data->vad_enabled = (attr->setting.vad != 0);
+    codec_data->plc_enabled = (attr->setting.plc != 0);
+    and_media_data->clock_rate = attr->info.clock_rate;
+
+#if PJMEDIA_HAS_AND_MEDIA_AMRNB
+    if (and_media_data->pt == PJMEDIA_RTP_PT_AMR ||
+	and_media_data->pt == PJMEDIA_RTP_PT_AMRWB)
```
nanangizz, line 836: The usage of enum/hardcoded payload type (e.g: `PJMEDIA_RTP_PT_AMR/WB`) for codec identification should be avoided as much as possible (please check other codec impls, e.g: in ilbc.c PT is used for codec_info.pt field requirement but effectively not really used anywhere else), note that the hardcoded PT may be deprecated in the future, currently in SDP offer generation, the default PT may be changed by PJMEDIA endpoint, and for answerer, it is set to match to PT in the SDP offer. Hardcoded PT may also introduce confusion in debugging code (when the PT are different between SDP & hardcoded).

nanangizz:
> > > To add AMediaCodec support, include this on the `config_site.h` file.
> > > ```
> > > #define PJMEDIA_HAS_ANDROID_MEDIACODEC 1
> > > ```
> > 
> > 
> > Why not simply have this setting auto-detect & auto-set ?
> 
> Cons of enabling Android MediaCodec by default:
> 
> * Most API used are supported only on API 21. Older device will have building issue.

Not enabled-by-default, but auto-detect (so it will be enabled only when supported, and the auto detect can be turn off via configure script param e.g: --no-android-mediacodec).


> * Some codec are supported on newer devices. e.g: VP9 is not yet supported on Android 7.0

Perhaps the common scenario among app developers is this: use native codecs when supported, otherwise fallback to third-party (e.g: OpenH264, vpx.c). So for each codec, ideally, unless disabled at compile time, the native codec availability is detected on run-time, at compile-time user can still enable both native & third-party (e.g: OpenH264), so when the native impl is not available on the device, PJMEDIA codec will fallback to the third-party impl.

Perhaps this can be simply done by reordering the codec registration (native codecs are registered first)?


> > Usually native codecs are preferred, so if this is stable enough and no compatibility issue with OpenH264, perhaps better to make them (or at least H264) enabled by default?
> 
> Anroid MediaCodec supports software and hardware codec access. Based on tests, the most stable configuration is to use
> software encoder and hardware decoder. e.g: qcom SoC doesn't provides Exynos or other codec.
> Currently, the encoder/decoder names are hardcoded and will be enabled if its able to be created.
> Last test is for qcom (qualcomm) and Exynos SoC.
> 
> On runtime, the logcat showed this message on encoding process (repeatedly):
> 
> ```
> E/NdkMediaCodec: sf error code: -13
> ```
> 
> However, there's no error was returned from the code and the video was successfully rendered on the remote side.

Please clarify about 'software encoder' and 'hardware decoder', how such configuration (or any other configuration such as all software/hardware) can be configured?

pjmedia/include/pjmedia-codec/and_aud_mediacodec.h
```diff
@@ -0,0 +1,55 @@
...
+ * @return	    PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t) pjmedia_codec_and_media_aud_deinit( void );
+
+PJ_END_DECL
+
+/**
+ * @}
+ */
+
```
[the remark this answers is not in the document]
trengginas: > * as this wrapper may contain multiple codecs, better list the supported codecs in the docs (plus some info about it), please check `ipp_codecs.h` as reference

Done

> * for AMR, shouldn't be there some configurable settings, e.g: octet-align, bitrate? if it is to be configurable via SDP fmtp, please add info (better also add sample code) in the docs above.

Done

pjmedia/src/pjmedia-codec/and_aud_mediacodec.cpp
trengginas: use new callback to predecode

pjmedia/include/pjmedia-codec.h
```diff
@@ -42,6 +42,8 @@
 #include <pjmedia-codec/speex.h>
 #include <pjmedia-codec/vid_toolbox.h>
 #include <pjmedia-codec/vpx.h>
+#include <pjmedia-codec/and_aud_mediacodec.h>
+#include <pjmedia-codec/and_vid_mediacodec.h>
```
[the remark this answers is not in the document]
trengginas: > > > > To add AMediaCodec support, include this on the `config_site.h` file.
> > > > ```
> > > > #define PJMEDIA_HAS_ANDROID_MEDIACODEC 1
> > > > ```
> > > 
> > > 
> > > Why not simply have this setting auto-detect & auto-set ?
> > 
> > 
> > Cons of enabling Android MediaCodec by default:
> > 
> > * Most API used are supported only on API 21. Older device will have building issue.
> 
> Not enabled-by-default, but auto-detect (so it will be enabled only when supported, and the auto detect can be turn off via configure script param e.g: --no-android-mediacodec).

Okay. Now it's auto-detect. With new option "--disable-android-mediacodec"

> 
> > * Some codec are supported on newer devices. e.g: VP9 is not yet supported on Android 7.0
> 
> Perhaps the common scenario among app developers is this: use native codecs when supported, otherwise fallback to third-party (e.g: OpenH264, vpx.c). So for each codec, ideally, unless disabled at compile time, the native codec availability is detected on run-time, at compile-time user can still enable both native & third-party (e.g: OpenH264), so when the native impl is not available on the device, PJMEDIA codec will fallback to the third-party impl.
> 
> Perhaps this can be simply done by reordering the codec registration (native codecs are registered first)?

Done.

> 
> > > Usually native codecs are preferred, so if this is stable enough and no compatibility issue with OpenH264, perhaps better to make them (or at least H264) enabled by default?
> > 
> > 
> > Anroid MediaCodec supports software and hardware codec access. Based on tests, the most stable configuration is to use
> > software encoder and hardware decoder. e.g: qcom SoC doesn't provides Exynos or other codec.
> > Currently, the encoder/decoder names are hardcoded and will be enabled if its able to be created.
> > Last test is for qcom (qualcomm) and Exynos SoC.
> > On runtime, the logcat showed this message on encoding process (repeatedly):
> > ```
> > E/NdkMediaCodec: sf error code: -13
> > ```
> > 
> > 
> > However, there's no error was returned from the code and the video was successfully rendered on the remote side.
> 
> Please clarify about 'software encoder' and 'hardware decoder', how such configuration (or any other configuration such as all software/hardware) can be configured?

The codec name indicates it is a software/hardware encoder/decoder.
e.g: codec prefix with "OMX.google" or "c2.android" is software codec.
Hardware decoder contains the SoC names on the prefix.
e.g: "OMX.qcom." or "OMX.exynos." 

https://stackoverflow.com/questions/37715529/how-to-know-android-decoder-mediacodec-createdecoderbytypetype-is-hardware-or

pjsip/src/pjsua-lib/pjsua_vid.c
```diff
@@ -1064,9 +1079,7 @@ pj_status_t pjsua_vid_channel_update(pjsua_call_media *call_med,
 
 #if defined(PJMEDIA_STREAM_ENABLE_KA) && PJMEDIA_STREAM_ENABLE_KA!=0
 	/* Enable/disable stream keep-alive and NAT hole punch. */
-	si->use_ka = acc->cfg.use_stream_ka;
-
-        si->ka_cfg = acc->cfg.stream_ka_cfg;
+	si->use_ka = pjsua_var.acc[call->acc_id].cfg.use_stream_ka;
```
nanangizz, line 1082: looks a bit strange, is this something related to the codec impl? also the changes below.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -0,0 +1,1908 @@
...
+	if (write_output)
+	    break;
+
+	buf_pos += frm_size;
+    }
+
+    return PJ_SUCCESS;
+}
+
+static pj_status_t vpx_unpacketize(struct and_media_codec_data *and_media_data,
```
sauwming, line 1691: if there are a lot of similarities with libvpx, perhaps better to refactor it to another file, similar to h263/h264_packetizer?

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -0,0 +1,1908 @@
...
+}
+
+static pj_status_t configure_encoder(and_media_codec_data *and_media_data)
+{
+    media_status_t am_status;
+    AMediaFormat *vid_fmt;
+    pjmedia_vid_codec_param *param = and_media_data->prm;
+
+    vid_fmt = AMediaFormat_new();
+    if (!vid_fmt) {
```
sauwming, line 606: need log as well, similar to below?

pjsip/src/pjsua-lib/pjsua_vid.c
```diff
@@ -1100,7 +1114,7 @@ pj_status_t pjsua_vid_channel_update(pjsua_call_media *call_med,
 	    }
 	}
 
-        if (pjsua_var.ua_cfg.cb.on_stream_precreate) {
+        if (!call->hanging_up && pjsua_var.ua_cfg.cb.on_stream_precreate) {
```
nanangizz, line 1117: This does not look related to the codec? If it is a fix for another issue, please put it in another PR or misc PR.

pjsip/src/pjsua-lib/pjsua_vid.c
```diff
@@ -1851,7 +1865,7 @@ static void call_get_vid_strm_info(pjsua_call *call,
 
 /* Send SDP reoffer. */
 static pj_status_t call_reoffer_sdp(pjsua_call_id call_id,
-				    const pjmedia_sdp_session *sdp)
+				    pjmedia_sdp_session *sdp)
```
nanangizz, line 1868: This does not look related to the codec? If it is a fix for another issue, please put it in another PR or misc PR.

pjsip/src/pjsua-lib/pjsua_vid.c
```diff
@@ -1868,6 +1882,13 @@ static pj_status_t call_reoffer_sdp(pjsua_call_id call_id,
 	return PJSIP_ESESSIONSTATE;
     }
 
+    /* Notify application */
+    if (!call->hanging_up && pjsua_var.ua_cfg.cb.on_call_sdp_created) {
+	(*pjsua_var.ua_cfg.cb.on_call_sdp_created)(call_id, sdp,
+						   call->inv->pool_prov,
+						   NULL);
+    }
+
```
nanangizz, line 1891: This does not look related to the codec? If it is a fix for another issue, please put it in another PR or misc PR.

pjmedia/src/pjmedia-codec/vpx_packetizer.c
```diff
@@ -0,0 +1,215 @@
...
+	cfg->fmt_id != PJMEDIA_FORMAT_VP9)
+    {
+	return PJ_ENOTSUP;
+    }
+
+    p_ = PJ_POOL_ZALLOC_T(pool, pjmedia_vpx_packetizer);
+    if (cfg) {
+	pj_memcpy(&p_->cfg, cfg, sizeof(*cfg));
+    } else {
+	p_->cfg.fmt_id = PJMEDIA_FORMAT_VP8;
```
nanangizz, line 62: `mtu` should be initialized too?
even better if there is `pjmedia_vpx_packetizer_cfg_default()`.

pjmedia/src/pjmedia-codec/and_aud_mediacodec.cpp
nanangizz: Predecode callback does avoid using PT for ID, but still many places using it, like the one here.

My point is, for new codecs like this, using PT (`PJMEDIA_RTP_PT_*` or even keyword `pt/PT`) for identifying codec should be avoided (except to be assigned to `pjmedia_codec_info`), for reasons described above, so please use other identification, simple internal enum should be fine, just avoid `PT`.

pjmedia/src/pjmedia-codec/and_aud_mediacodec.cpp
```diff
@@ -0,0 +1,1366 @@
...
+    }
+    AMediaFormat_setString(aud_fmt, AND_MEDIA_KEY_MIME,
+                           and_media_codec[idx].mime_type);
+    AMediaFormat_setInt32(aud_fmt, AND_MEDIA_KEY_PCM_ENCODING, 2);
+    AMediaFormat_setInt32(aud_fmt, AND_MEDIA_KEY_CHANNEL_COUNT,
+                          and_media_codec[idx].channel_count);
+    AMediaFormat_setInt32(aud_fmt, AND_MEDIA_KEY_SAMPLE_RATE,
+			  and_media_codec[idx].clock_rate);
+    AMediaFormat_setInt32(aud_fmt, AND_MEDIA_KEY_BITRATE,
+			  and_media_codec[idx].def_bitrate);
```
nanangizz, line 369: If there are any settings for VAD, DTX, PLC, etc, better configure them here too.

Btw, can we enable/disable VAD/PLC for AMR/AMR-WB? If not, the docs better mention it.

trengginas: There's no VAD, DTX, PLC feature from MediaCodec. Add documentation on ```and_aud_mediacodec.h```

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -0,0 +1,1846 @@
...
+    pktz_cfg.unpack_nal_start = 4;
+    /* Packetization mode */
+    if (h264_fmtp.packetization_mode == 0)
+        pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_SINGLE_NAL;
+    else if (h264_fmtp.packetization_mode == 1)
+        pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_NON_INTERLEAVED;
+    else
+        return PJ_ENOTSUP;
+
+    pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_NON_INTERLEAVED;
```
nanangizz, line 1406: Looks like we need to mention somewhere in the docs that Android H264 only support packetization mode non-interleaved. Also just checked vid_toolbox impl, it seems to have the same issue.

Ideally the codec implementation can verify H264 fmtp setting from `pjmedia_codec_mgr_set_default_param()` to make sure that SDP fmtp `packetization-mode` is set to 1. Unfortunately this cannot be done for now.

Note that when the SDP nego result uses `packetization mode=0` (single-nal) and we keep sending/encoding with non-interleaved RTP packetization, remote will not be able to unpack our RTP payload and it won't be able to decode our video properly.

pjmedia/include/pjmedia-codec.h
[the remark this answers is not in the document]
nanangizz: So is it possible that a device supports all combinations, i.e: hardware & software for both encoder & decoder? If yes, how an app that wants to have 'the most stable configuration', which mentioned above: software encoder and hardware decoder, can configure that?

nanangizz (approved):
Looks okay, just need to specify in docs (vid codec header file) about H264 packetization-mode limitation.

trengginas:
> Looks okay, just need to specify in docs (vid codec header file) about H264 packetization-mode limitation.

Further testing shows that the video codec drops its performance on a bad quality network. In this case (e.g: using hardware decoder on `Exynos SoC`) there would be runtime issues and the codec unable to recover until it freeze the app entirely.  Due to this, software decoder should be prioritize for performance. Maybe we should also consider registering non-native codec implementation (`OpenH264`, `VPX`) first. For example, `vid_toolbox` codec implementation, needed to enabled manually although it's auto-detected.

sauwming:
> Further testing shows that the video codec drops its performance on a bad quality network. In this case (e.g: using hardware decoder on `Exynos SoC`) there would be runtime issues and the codec unable to recover until it freeze the app entirely. Due to this, software decoder should be prioritize for performance. Maybe we should also consider registering non-native codec implementation (`OpenH264`, `VPX`) first. For example, `vid_toolbox` codec implementation, needed to enabled manually although it's auto-detected.

That's strange. I would imagine that the codec decoder should have an easier task if there are lost or corrupted packets since it only needs to report decode failure or missing frames. While the encoder should be unaffected.
So perhaps more investigation is needed as to why the performance could drop?

trengginas:
> > Further testing shows that the video codec drops its performance on a bad quality network. In this case (e.g: using hardware decoder on `Exynos SoC`) there would be runtime issues and the codec unable to recover until it freeze the app entirely. Due to this, software decoder should be prioritize for performance. Maybe we should also consider registering non-native codec implementation (`OpenH264`, `VPX`) first. For example, `vid_toolbox` codec implementation, needed to enabled manually although it's auto-detected.
> 
> That's strange. I would imagine that the codec decoder should have an easier task if there are lost or corrupted packets since it only needs to report decode failure or missing frames. While the encoder should be unaffected.
> So perhaps more investigation is needed as to why the performance could drop?

Unfortunately, the issue couldn't be reproduce. Did a rebuild and the codec manage to recover and not freeze.
Tried with `tx-drop-pct` and `rx-drop-pct` set upto 80 and the codec didn't freeze.

sauwming (approved):

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -0,0 +1,1923 @@
...
+			   pj_str_t **codec_name,
+			   unsigned *codec_num)
+{
+    pj_bool_t use_sw_enc = PJMEDIA_AND_MEDIA_PRIO_SW_VID_ENC;
+    pj_bool_t use_sw_dec = PJMEDIA_AND_MEDIA_PRIO_SW_VID_DEC;
+
+    *codec_num = 0;
+
+    switch (fmt_id) {
+
```
ArtsemKurantsou, line 641: Looks like an additional compile-time check required here to avoid `use of undeclared identifier 'H264_sw_encoder'` compilation error.
```suggestion
#if PJMEDIA_HAS_AND_MEDIA_H264
```
The same thing for other codecs

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -0,0 +1,1923 @@
...
+    pj_uint8_t         keyframe_interval; /* Keyframe interval.              */
+
+    open_cb            open_codec;
+    process_encode_cb  process_encode;
+    encode_more_cb     encode_more;
+    decode_cb          decode;
+
+    pjmedia_codec_fmtp dec_fmtp;	  /* Decoder's fmtp params.	     */
+}
+and_media_codec[] {
```
ArtsemKurantsou, line 316: I got a compilation error here `definition of variable with array type needs an explicit size or an initializer and_media_codec[] {`. Not sure if it because of a compiler.
```suggestion
and_media_codec[] = {
```

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -0,0 +1,1929 @@
...
+
+    pjmedia_codec_fmtp dec_fmtp;	  /* Decoder's fmtp params.	     */
+}
+and_media_codec[] = {
+#if PJMEDIA_HAS_AND_MEDIA_H264
+    {0, "H264",	"Android MediaCodec H264 codec", "video/avc",
+        NULL, NULL,
+        PJMEDIA_RTP_PT_H264, PJMEDIA_FORMAT_H264, KEYFRAME_INTERVAL,
+        &open_h264, &process_encode_h264, &encode_more_h264, &decode_h264,
+        {2, {{{(char *)"profile-level-id", 16}, {(char *)"42e01e", 6}},
```
ArtsemKurantsou, line 322: Is there any reason to use `42e01e`by default instead of `640028` like it done in `openh264.cpp`?

ArtsemKurantsou:
Hi @trengginas. Would like to ask one question: it looks like both openh264 and AMediaCodec one will have the same id - `H264/96`. So it looks like there won't be a way to change the priority for the openh264 implementation at runtime. Am I correct?

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
trengginas: `42e01e` also used on `openh264.cpp`  https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-codec/openh264.cpp#L284

trengginas:
> Hi @trengginas. Would like to ask one question: it looks like both openh264 and AMediaCodec one will have the same id - `H264/96`. So it looks like there won't be a way to change the priority for the openh264 implementation at runtime. Am I correct?

That's correct, native codec will be prioritize if enabled. If you want to use openh264, you would need to disable the AMediaCodec

nanangizz:
> > Hi @trengginas. Would like to ask one question: it looks like both openh264 and AMediaCodec one will have the same id - `H264/96`. So it looks like there won't be a way to change the priority for the openh264 implementation at runtime. Am I correct?
> 
> That's correct, native codec will be prioritize if enabled. If you want to use openh264, you would need to disable the AMediaCodec

Interesting question. Here is an idea, if AMediaCodec uses default unique payload types for its H264 codecs (and none of them using 96), the codec IDs of H264 backends will be unique, and app can change their priorities at run time. Note that for video codecs, codec ID is generated from codec name & default payload type.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
ArtsemKurantsou: Sorry. Forgot that I've changed it in my version.

trengginas:
> > > Hi @trengginas. Would like to ask one question: it looks like both openh264 and AMediaCodec one will have the same id - `H264/96`. So it looks like there won't be a way to change the priority for the openh264 implementation at runtime. Am I correct?
> > 
> > 
> > That's correct, native codec will be prioritize if enabled. If you want to use openh264, you would need to disable the AMediaCodec
> 
> Interesting question. Here is an idea, if AMediaCodec uses default unique payload types for its H264 codecs (and none of them using 96), the codec IDs of H264 backends will be unique, and app can change their priorities at run time. Note that for video codecs, codec ID is generated from codec name & default payload type.

Currently this doesn't work. When setting codec priority, only the codec desc is sorted out and not the factory [(ref).](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/vid_codec.c#L473)
This will affect when allocating the correct codec, since now it only matches the fmt id and not the pt (e.g: [openh264](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-codec/openh264.cpp#L250)).
So for this to work, we need to also sort the codec factory when setting the codec prio. Or match the pt on `test_alloc()`.

nanangizz:
> Currently this doesn't work. When setting codec priority, only the codec desc is sorted out and not the factory [(ref).](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/vid_codec.c#L473)
> This will affect when allocating the correct codec, since now it only matches the fmt id and not the pt (e.g: [openh264](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-codec/openh264.cpp#L250)).
> So for this to work, we need to also sort the codec factory when setting the codec prio. Or match the pt on `test_alloc()`.

For now, setting different default PTs for all H264 backends and matching default PT in `test_alloc()` seem to be the simplest way. 

Sorting codec factories may be not applicable, here is a sample scenario: factory A has codec 1 & 2, factory B has codec 3 & 4, when app sets codec priority to: 1, 3, 2, 4, not sure how the factories should be sorted. Perhaps the `pjmedia_vid_codec_mgr_alloc_codec` should iterate the sorted `codec_desc` instead of the factories, but this is perhaps a major change (not for now).

IMO, to be able to set priority for H264 codec backends at run-time is rather urgent for now, especially on Android where there are choices of OS native backends (with limitations for real-time streaming purpose) and matured OpenH264.

trengginas:
> > Currently this doesn't work. When setting codec priority, only the codec desc is sorted out and not the factory [(ref).](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/vid_codec.c#L473)
> > This will affect when allocating the correct codec, since now it only matches the fmt id and not the pt (e.g: [openh264](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-codec/openh264.cpp#L250)).
> > So for this to work, we need to also sort the codec factory when setting the codec prio. Or match the pt on `test_alloc()`.
> 
> For now, setting different default PTs for all H264 backends and matching default PT in `test_alloc()` seem to be the simplest way.
> 
> Sorting codec factories may be not applicable, here is a sample scenario: factory A has codec 1 & 2, factory B has codec 3 & 4, when app sets codec priority to: 1, 3, 2, 4, not sure how the factories should be sorted. Perhaps the `pjmedia_vid_codec_mgr_alloc_codec` should iterate the sorted `codec_desc` instead of the factories, but this is perhaps a major change (not for now).
> 
> IMO, to be able to set priority for H264 codec backends at run-time is rather urgent for now, especially on Android where there are choices of OS native backends (with limitations for real-time streaming purpose) and matured OpenH264.

Should create a new PR for this? It applies to other codecs as well

nanangizz:
> Should create a new PR for this? It applies to other codecs as well

Yes, a separate PR should be cleaner.

nanangizz (approved):

PR #2560, opened 2020-10-23
Don't infinite loop when searching CLI history list

### LINKED ISSUE #2559: Telnet cli of pjsua runs into infinite loop after several new history entries
**Describe the bug**
After entering several (about 15) different commands in the telnet cli the linked list of the history is corrupted and has no end. Entering the next new command, the command is searched in the history. The search comes not to the end and the client consumes a lot of CPU.

**To Reproduce**
1. start the pjsua app with a configured telnet cli. 
2. connect to the telnet cli
3. performe different commands (e.g. acc show, call list, call new, call hangup)
4. the commands must differ in a way that new history entries will be created (e.g. calls to different numbers)
5. after some (about 15) commands the command will not return an a lot of cpu is consumed. The process hangs in the loop to find a matching history entry (debugged with additional temporary logs).

**Expected behavior**
The client accepts the command.

**Logs/Screenshots**
no logs

**Desktop/Smartphone (please complete the following information):**

- OS: debian 9
- PJSIP
  - version: 2.10
07:52:51.571         os_core_unix.c !pjlib 2.10 for POSIX initialized
07:52:51.571         sip_endpoint.c  .Creating endpoint instance...
07:52:51.571                  pjlib  .select() I/O Queue created (0x557795745b60)
07:52:51.571         sip_endpoint.c  .Module "mod-msg-print" registered
07:52:51.571        sip_transport.c  .Transport manager created.
07:52:51.571           pjsua_core.c  .PJSUA state changed: NULL --> CREATED
07:52:51.571               config.c  PJLIB (c)2008-2016 Teluu Inc.
07:52:51.571               config.c  Dumping configurations:
07:52:51.571               config.c   PJ_VERSION                : 2.10
07:52:51.571               config.c   PJ_M_NAME                 : x86_64
07:52:51.571               config.c   PJ_HAS_PENTIUM            : 0
07:52:51.571               config.c   PJ_OS_NAME                : x86_64-unknown-linux-gnu
07:52:51.571               config.c   PJ_CC_NAME/VER_(1,2,3)    : gcc-6.3.0
07:52:51.571               config.c   PJ_IS_(BIG/LITTLE)_ENDIAN : little-endian
07:52:51.571               config.c   PJ_HAS_INT64              : 1
07:52:51.571               config.c   PJ_HAS_FLOATING_POINT     : 1
07:52:51.571               config.c   PJ_DEBUG                  : 1
07:52:51.571               config.c   PJ_FUNCTIONS_ARE_INLINED  : 0
07:52:51.571               config.c   PJ_LOG_MAX_LEVEL          : 5
07:52:51.571               config.c   PJ_LOG_MAX_SIZE           : 4000
07:52:51.571               config.c   PJ_LOG_USE_STACK_BUFFER   : 1
07:52:51.571               config.c   PJ_POOL_DEBUG             : 0
07:52:51.571               config.c   PJ_HAS_POOL_ALT_API       : 0
07:52:51.571               config.c   PJ_HAS_TCP                : 1
07:52:51.571               config.c   PJ_MAX_HOSTNAME           : 128
07:52:51.571               config.c   ioqueue type              : select
07:52:51.571               config.c   PJ_IOQUEUE_MAX_HANDLES    : 64
07:52:51.571               config.c   PJ_IOQUEUE_HAS_SAFE_UNREG : 1
07:52:51.571               config.c   PJ_HAS_THREADS            : 1
07:52:51.571               config.c   PJ_LOG_USE_STACK_BUFFER   : 1
07:52:51.571               config.c   PJ_HAS_SEMAPHORE          : 1
07:52:51.571               config.c   PJ_HAS_EVENT_OBJ          : 1
07:52:51.571               config.c   PJ_ENABLE_EXTRA_CHECK     : 1
07:52:51.571               config.c   PJ_HAS_EXCEPTION_NAMES    : 1
07:52:51.571               config.c   PJ_MAX_EXCEPTION_ID       : 16
07:52:51.571               config.c   PJ_EXCEPTION_USE_WIN32_SEH: 0
07:52:51.571               config.c   PJ_TIMESTAMP_USE_RDTSC:   : 0
07:52:51.571               config.c   PJ_OS_HAS_CHECK_STACK     : 0
07:52:51.571               config.c   PJ_HAS_HIGH_RES_TIMER     : 1
07:52:51.571               config.c   PJ_HAS_IPV6               : 0
**Smartphone (please also complete the following information):**

- Device: desktop

**Additional context**
none

seanbright:
The easiest way to clean up the history list when moving an item is to
just remove it and reinsert, so use pj_list_erase().

Fixes #2559

### DIFF
--- a/pjlib-util/src/pjlib-util/cli_telnet.c
+++ b/pjlib-util/src/pjlib-util/cli_telnet.c
@@ -385,13 +385,13 @@ static pj_status_t insert_history(cli_telnet_sess *sess,
 {
     cmd_history *in_history;
     pj_str_t cmd;
-    cmd.ptr = cmd_val;
-    cmd.slen = pj_ansi_strlen(cmd_val)-1;
-
-    if (cmd.slen == 0)
-	return PJ_SUCCESS;
 
     PJ_ASSERT_RETURN(sess, PJ_EINVAL);
+
+    cmd = pj_str(cmd_val);
+    pj_strtrim(&cmd);
+    if (cmd.slen == 0)
+	return PJ_SUCCESS;
 
     /* Find matching history */
     in_history = pj_list_search(sess->history, (void*)&cmd, compare_str);
@@ -409,10 +409,10 @@ static pj_status_t insert_history(cli_telnet_sess *sess,
 	    in_history = sess->history->prev;
 	    pj_list_erase(in_history);
 	}
+	pj_strncpy(&in_history->command, &cmd, PJ_CLI_MAX_CMDBUF);
     } else {
-	pj_list_insert_nodes_after(in_history->prev, in_history->next);
+	pj_list_erase(in_history);
     }
-    pj_strcpy(&in_history->command, pj_strtrim(&cmd));
     pj_list_push_front(sess->history, in_history);
     sess->active_history = sess->history;
 
--- a/pjlib-util/src/pjlib-util/cli_telnet.c
+++ b/pjlib-util/src/pjlib-util/cli_telnet.c
@@ -407,6 +407,7 @@ static pj_status_t insert_history(cli_telnet_sess *sess,
 	} else {
 	    /* Get the oldest history */
 	    in_history = sess->history->prev;
+	    pj_list_erase(in_history);
 	}
     } else {
 	pj_list_insert_nodes_after(in_history->prev, in_history->next);


nanangizz (approved):

sauwming (commented):
Looking at the spec and implementation of pj_strtrim() doesn't give me assurance that it's safe for zero-length pj_str that has uninitialised ptr.
So perhaps move the strtrim() below the if?

nanangizz:
The spec does not seem to specify anything about it, but looking at the implementation, the uninitalised ptr should never be dereferenced when the string is zero-length, so it should be safe? Or it is unsafe because of something else?

OTOH the original code seems to assume that the string length is always greater than zero: `cmd.slen = pj_ansi_strlen(cmd_val)-1`.

It should be better if the string can be trimmed first so we get a 'clean' string when evaluating the length (`if (cmd.slen == 0)`).

sauwming:
The code that's bothering me is:
`register char *p = end - 1;` which, theoretically, when the pj_str is zero length, allows the pointer to be outside the buffer and  potentially underflow. Practically, it may be extremely rare, if possible at all, though.

Ah right, you want to check the length after trimming. I initially thought there's no point in doing so and that's why I suggest to move it below.

sauwming (approved):

nanangizz (approved):

PR #2567, opened 2020-10-29
Fix possible INVITE response retransmission with wrong code

nanangizz:
Scenario:
1. `pjsip_inv_answer()` is called with response code `180` and the returned message is sent.
2. `pjsip_inv_answer()` is called again with response code `200`, unfortunately it fails after the response code is updated to `200`, e.g: due to `process_answer()` returning non-success.
3. retransmission timer is called to retransmit last response, instead of retransmitting `180` response, it transmits `200` response (likely without SDP).

Proposed fix is to delay `inv->last_answer` update until after `process_answer()` is done successfully.


sauwming:
This seems to be a duplicate of #2472. Better combine that as well, i.e. the part when it releases `last_res` upon failure.

nanangizz:
Yes, the issue described here should be the effect of the first bug described in #2472. Also I think the fix in #2472 should be the correct fix.

sauwming:
I just realized my comment for that PR that I wrote long time ago was never submitted.

nanangizz:
Duplicate of #2472

PR #2570, opened 2020-11-02
Added underlying buffer configurability for splitcomb

mike2307:
This pull request adds the possibility to handover options for the underlying buffer when creating an instance of splitcomb.

### DIFF
--- a/pjmedia/include/pjmedia/splitcomb.h
+++ b/pjmedia/include/pjmedia/splitcomb.h
@@ -113,7 +113,10 @@ PJ_DECL(pj_status_t) pjmedia_splitcomb_set_channel(pjmedia_port *splitcomb,
  * @param options	    Normally is zero, but the lower 8-bit of the 
  *			    options can be used to specify the number of 
  *			    buffers in the circular buffer. If zero, then
- *			    default number will be used (default: 8).
+ *			    default number will be used (default: 8). The second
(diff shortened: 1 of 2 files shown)


sauwming:
Since `pjmedia_delay_buf` flag is currently only 1 bit, I wonder whether using 8 bits is too much. Perhaps 2, or at most 4 bits, should be more than sufficient.

mike2307:
The options argument is a 32bit unsigned. So there should be plenty of space left. I thought I use a whole byte to make the usage easier.

sauwming:
I believe unsigned can be 16 or 32-bit, depending on the machine or compiler option.

mike2307:
Yes, but is pjproject supporting any 16 bit architectures?

sauwming:
PJSIP supports various architectures and platforms, including embedded devices. Even if it's not directly supported, PJSIP can be ported to another OS by porting PJLIB, as it has been done by our community in the past:
https://www.pjsip.org/docs/book-latest/html/consider.html#porting-to-other-embedded-os
https://www.pjsip.org/docs/latest/pjlib/docs/html/porting_pjlib_pg.htm

I understand though that 16-bit arch may be rare nowadays, so I don't have any major objection for this.

sauwming (approved):

trengginas (approved):

nanangizz (approved):

PR #2571, opened 2020-11-02
Added function pjmedia_tonegen_stop_loop()

mike2307:
This pull request adds the function pjmedia_tonegen_stop_loop() to tonegen, which can be used to stop looping of the current tone set.

### DIFF
--- a/pjmedia/src/pjmedia/tonegen.c
+++ b/pjmedia/src/pjmedia/tonegen.c
@@ -522,7 +522,7 @@ PJ_DEF(pj_status_t) pjmedia_tonegen_stop_loop(pjmedia_port *port)
     TRACE_((THIS_FILE, "tonegen_stop_loop()"));
 
     pj_lock_acquire(tonegen->lock);
-    tonegen->playback_options &= !PJMEDIA_TONEGEN_LOOP;
+    tonegen->playback_options &= ~PJMEDIA_TONEGEN_LOOP;
     pj_lock_release(tonegen->lock);
 
     return PJ_SUCCESS;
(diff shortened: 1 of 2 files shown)


sauwming:
Can you let us know the use case, i.e. when app/user will prefer to continue playing the tonegen until the end instead of stopping it immediately with `pjmedia_tonegen_stop()` ?

mike2307:
The reason is, that this way will ensure a "soft" stop. e.g. the generation of a busy phone tone can then be stopped in the silence phase instead of in the middle of the tone.

sauwming (approved):

pjmedia/src/pjmedia/tonegen.c
```diff
@@ -511,6 +511,22 @@ PJ_DEF(pj_status_t) pjmedia_tonegen_stop(pjmedia_port *port)
...
+ */
+PJ_DEF(pj_status_t) pjmedia_tonegen_stop_loop(pjmedia_port *port)
+{
+    struct tonegen *tonegen = (struct tonegen*) port;
+    PJ_ASSERT_RETURN(port->info.signature == SIGNATURE, PJ_EINVAL);
+
+    TRACE_((THIS_FILE, "tonegen_stop_loop()"));
+
+    pj_lock_acquire(tonegen->lock);
+    tonegen->playback_options &= !PJMEDIA_TONEGEN_LOOP;
```
nanangizz, line 525: instead of `!PJMEDIA_TONEGEN_LOOP`, shouldn't it be `~PJMEDIA_TONEGEN_LOOP` (logical vs bitwise op)?

mike2307: Yes, you're right. I will correct that.

nanangizz (approved):

Issue #1049, opened 2010-03-29
PJSUA-LIB should report disconnection event immediately after pjsua_call_hangup() is called


pjsipbot:
_2010-03-29 08:05:15: @bennylp_ _created the issue on trac ticket [1049](https://trac.pjsip.org/repos/ticket/1049)_

There are several scenarios in which call DISCONNECTED state will only be reported later, if at all, after `pjsua_call_hangup()` is issued:
 - when INVITE is not responded with any responses yet (not even provisional responses), which in this case, PJSIP will continue retransmitting the request until either a response is received (and if it's not a final response, CANCEL then will be issued), or retransmission times out 32 seconds after call is initiated.
 - when BYE is issued against peer which doesn't exist; in this case, the DISCONNECTED state will be issued after the BYE transaction times out, 32 seconds after it is issued
 - ~~when BYE is answered with failure final response; in this case, most likely call will stay as is. Note that 408 and 481 response would cause call to be terminated though.~~ (in `inv_handle_bye_response()`, any response other than 401 and 407 will end the session).
 - when 200/OK for CANCEL is received, but the final response for the INVITE transaction never arrives; in this case, call will remain as is indefinitely (!) (done in #503)
 - when ACK is not received yet for 200/OK, UAS is not able to send BYE. The stack ideally should send the BYE automatically once ACK is received (done in #1712)

In all cases above, it is desirable to have the application receives the DISCONNECTED event immediately and let PJSUA-LIB handle the call under the hood. Several scenarios that must be handled by PJSUA-LIB include:
 - sending BYE if call isn't disconnected after CANCEL is issued:
    - it is possible that the CANCEL request and 200/OK to INVITE had crossed each other in the wire; in this case, the call will get CONNECTED and the CANCEL will be rejected with 481 (done in #1817)
    - peer may have been malfunctioning, not disconnecting the call even after sending 200/OK response to CANCEL. We may need to have a timer for this. (done in #503)
 - ~~retry sending the BYE if previous ones fail and call remains connected. If this keeps happening (the BYE request gets rejected), we'll probably need to hard-terminate the call, that is to forcefully terminate the call without relying for the BYE to succeed.~~ (this doesn't seem necessary, in `inv_handle_bye_response()`, any response other than 401 and 407 will end the session).
 - when ACK is not received yet for 200/OK, UAS is not able to send BYE. The stack ideally should send the BYE automatically once ACK is received (done in #1712)
[...]

Issue #1627, opened 2013-02-26
Callback on_tsx_state_changed() not invoked on call disconnection


pjsipbot:
_2013-02-26 11:52:58: @nanangizz_ _created the issue on trac ticket [1627](https://trac.pjsip.org/repos/ticket/1627)_

When a call is disconnected, SIP invite module will call `on_call_state(DISCONNECTED)` first before `on_tsx_state_changed()`. However, in PJSUA, the `on_call_state(DISCONNECTED)` will reset PJSUA call states and the next incoming `on_tsx_state_changed()` will not be forwarded to the application, as it will not make sense for the application to still receive `on_tsx_state_changed()` after the call is disconnected (and the PJSUA call states are resetted).

An easy or low-risk workaround will be 'manually' invoking `on_tsx_state_changed()` from `on_call_state(DISCONNECTED)`.
___
_2013-02-26 12:26:11: @nanangizz_ _edited the issue description_

___
_2013-02-26 12:38:29: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2013-02-26 12:38:29: @nanangizz_ _set **resolution** to **fixed**_

___
_2013-02-26 12:38:29: @nanangizz_ _commented_

In [r4371](../commit/581cf335e5871bb26ab349b7c41bf8576d18728f):
Close #1627: invoke on_tsx_state_changed() callback from on_call_state(DISCONNECTED).

___
_2013-02-27 15:05:30: @trengginas_ _commented_

In [r4408](../commit/61d29933fb9b2faecf5d8bc0777f2c299e68f493):
Re #1627: backported to 1.x

PR #2582, opened 2020-11-17
Report disconnection event immediately when hanging up a call

sauwming:
To continue the work in #1049 and close it.

* Add a new API `pjsua_call_hangup2()` which allows user the option to immediately hangup a call. The call slot can then be reused almost immediately while the hangup process itself will be handled in the background. The only exception is if the asynchronous media transport creation is in progress during initial call setup. In this case, the hangup will be pending upon the completion of the media transport. Note that any media transport (re-)creation, such as during reinit media, once the call has been confirmed will not pend the hangup process, since the media tp recreation will be synchronous while acquiring dialog lock.

* Application will not receive any PJSUA call related callback once call is declared to be disconnected.

* Retry creating and sending BYE after a period of `CALL_HANGUP_RETRY_INTERVAL` if it fails. If this keeps happening after `CALL_HANGUP_MAX_RETRY` times, we will hard-terminate the call, i.e. cleanup the call without sending any BYE to the remote.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -5574,6 +5574,40 @@ PJ_DECL(pj_status_t) pjsua_call_hangup(pjsua_call_id call_id,
 				       const pjsua_msg_data *msg_data);
 
 /**
+ * Hangup call by using method that is appropriate according to the
+ * call state. This function is different than answering the call with
+ * 3xx-6xx response (with #pjsua_call_answer()), in that this function
+ * will hangup the call regardless of the state and role of the call,
+ * while #pjsua_call_answer() only works with incoming calls on EARLY
+ * state.
+ *
+ * If parameter immediate is set to PJ_TRUE, call DISCONNECTED state
+ * will be immediately reported to app and the call slot can be reused.
+ * The only exception is if media transport creation is in progress
+ * during call setup. In this case, the hangup will be pending upon
+ * the completion of the media transport.
+ * The library will then process the call hangup, such as notifying
+ * the remote and cleaning up used resources, in the background.
+ *
+ * @param call_id	Call identification.
+ * @param immediate	Set to PJ_TRUE to hangup the call immediately.
+ * @param code		Optional status code to be sent when we're rejecting
+ *			incoming call. If the value is zero, "603/Decline"
+ *			will be sent.
+ * @param reason	Optional reason phrase to be sent when we're rejecting
+ *			incoming call.  If NULL, default text will be used.
+ * @param msg_data	Optional list of headers etc to be added to outgoing
+ *			request/response message.
+ *
+ * @return		PJ_SUCCESS on success, or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsua_call_hangup2(pjsua_call_id call_id,
+				        pj_bool_t immediate,
+				        unsigned code,
+				        const pj_str_t *reason,
+				        const pjsua_msg_data *msg_data);
+
+/**
  * Accept or reject redirection response. Application MUST call this function
  * after it signaled PJSIP_REDIRECT_PENDING in the \a on_call_redirected() 
  * callback, to notify the call whether to accept or reject the redirection
--- a/pjsip-apps/src/pjsua/pjsua_app_legacy.c
+++ b/pjsip-apps/src/pjsua/pjsua_app_legacy.c
@@ -911,7 +911,7 @@ static void ui_hangup_call(char menuin[])
 	pjsua_call_hangup_all();
     } else {
 	/* Hangup current calls */
-	pjsua_call_hangup(current_call, 0, NULL, NULL);
+	pjsua_call_hangup2(current_call, PJ_TRUE, 0, NULL, NULL);
     }
 }
 
--- a/pjsip/include/pjsua-lib/pjsua_internal.h
+++ b/pjsip/include/pjsua-lib/pjsua_internal.h
@@ -491,6 +491,8 @@ struct pjsua_data
     unsigned		 call_cnt;		/**< Call counter.	*/
     pjsua_call		 calls[PJSUA_MAX_CALLS];/**< Calls array.	*/
     pjsua_call_id	 next_call_id;		/**< Next call id to use*/
+    pjsua_call		 hangup_calls[PJSUA_MAX_CALLS];/**< Calls in
+    						     hangup process.	*/
 
     /* Buddy; */
     unsigned		 buddy_cnt;		    /**< Buddy count.	*/
@@ -694,7 +696,7 @@ pj_status_t pjsua_media_channel_create_sdp(pjsua_call_id call_id,
 pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 				       const pjmedia_sdp_session *local_sdp,
 				       const pjmedia_sdp_session *remote_sdp);
-pj_status_t pjsua_media_channel_deinit(pjsua_call_id call_id);
+pj_status_t pjsua_media_channel_deinit(pjsua_call *call);
 
 /*
  * Error message when media operation is requested while another is in progress
@@ -712,7 +714,7 @@ pj_status_t pjsua_call_media_init(pjsua_call_media *call_med,
 void pjsua_call_cleanup_flag(pjsua_call_setting *opt);
 void pjsua_set_media_tp_state(pjsua_call_media *call_med, pjsua_med_tp_st tp_st);
 
-void pjsua_media_prov_clean_up(pjsua_call_id call_id);
+void pjsua_media_prov_clean_up(pjsua_call *call);
 
 /* Callback to receive media events */
 pj_status_t on_media_event(pjmedia_event *event, void *user_data);
--- a/pjsip/include/pjsua-lib/pjsua_internal.h
+++ b/pjsip/include/pjsua-lib/pjsua_internal.h
@@ -182,6 +182,7 @@ struct pjsua_call
             struct {		
                 call_answer      answers;/**< A list of call answers.       */
 		pj_bool_t	 hangup;/**< Call is hangup?		    */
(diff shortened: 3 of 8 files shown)


pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -5573,6 +5573,40 @@ PJ_DECL(pj_status_t) pjsua_call_hangup(pjsua_call_id call_id,
...
+ * 3xx-6xx response (with #pjsua_call_answer()), in that this function
+ * will hangup the call regardless of the state and role of the call,
+ * while #pjsua_call_answer() only works with incoming calls on EARLY
+ * state.
+ *
+ * If parameter immediate is set to PJ_TRUE, call DISCONNECTED state
+ * will be immediately reported to app and the call slot can be reused.
+ * The only exception is if media transport creation is in progress
+ * during call setup. In this case, the hangup will be pending upon
+ * the completion of the media transport.
```
nanangizz, line 5588: For the exception condition, will the app still be notified immediately?

sauwming: No, the app will be notified upon completion of media transport.

nanangizz: Why is that?

Not sure if this is relevant. With trickle ICE, media transport completion may be reported immediately, while internally ICE is actually still gathering candidates. So for this case, which completion will the pjsua-lib wait for?

sauwming: It is the current behavior of call hangup, i.e. to delay the hangup until media transport creation completes:
https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_call.c#L2760

The reason this behavior is not changed is because of the complexity:
In order to support immediate hangup, we move the current `pjsua_var.call` instance into another variable `pjsua_var.hangup_call` and handle the hangup process from this new variable. For media related callbacks however, since currently we are using `&call.media[i]` as the user data, this will point to the old call media instance. It is possible to overcome this issue, by:
- Creating the user data in the heap (using dialog's pool?), perhaps a struct of { dialog and media index }.
- Modifying `pjsua_media.c` callbacks to use this new user data.

But there is still the complexity of race condition, i.e. since the media transport creation is async, there is a possibility that the media callback will happen at the same time with call hangup. This may require us to hold dialog lock in all media related callbacks. So I'm not sure if the potentially major modifications required is worth the advantage, i.e. to be able to immediately hangup the call when async med tp creation is in progress in the beginning of the call (again, note that sync media reinit in the middle of the call should be unaffected by this).

As for trickle ICE, I believe it will be upon the report of completion, because the hangup will proceed in `pjsua_call.c`'s `on_make/incoming_call_med_tp_complete()`. I'm not sure if this can cause any complications though, since trickle ICE is a new feature.

nanangizz: As the ticket is about immediate hangup notification and handle the rest in background, I am thinking that media transport creation completion should be part of things to be done in background.

Re: `&call.media[i]` user data in media callbacks, have not checked the patch details, but perhaps the problem raises because the call instance may now be reused, right? What if it cannot be reused until the hangup process completed, e.g: simply add a flag `hangingup` in `pjsua_call` and alloc_call_id() filters out call slot with that flag set. This may be a completely different approach compared to this current patch (which seems to move the hanging up call instance to `pjsua_var.hangup_call`, cmiiw).

Re: trickle ICE, from PJSUA point of view, create_ice_media_transport() will set `call_med->tp_ready` to `PJ_SUCCESS` (instead of `PJ_EPENDING` as in current behavior). So perhaps PJSUA will try to destroy the `call_med->tp` immediately on call hangup, may be safe though.

sauwming: Alright, so that's another alternative we need to consider, i.e. only to report the DISCONNECTED state immediately while the call instance itself still remains. And this seems a hybrid approach between what the PR proposes and the current state (which is that application is recommended to treat the call as disconnected immediately after hanging up (i.e. report as such to the user, clear the call from the app UI, etc), even though it hasn't received DISCONNECTED call state and the hangup process is still underway).

There may be a problem with the alternative approach, i.e. when the call slots are full because of the pending hangups, but users can't make new calls (returning PJ_ETOOMANY) even though `pjsua_call_get_count()` seems to indicate that there are few/no active calls. Even worse, compared to current situation, the user will no longer be notified when the call will finally be cleared.
A solution for this that I can think of is probably to add a new API `pjsua_call_get_free_count()` and introduce a new state `CALL_STATE_DESTROYED`.

There is also another option that we should consider, which is to stick with the current approach. If both alternatives do not actually give a clear advantage over the current one, and only adds unnecessary complexity, I guess it makes sense to leave it as is?

nanangizz: The complexity should be the reason the ticket have not been implemented, yet we can't close the ticket because requests keep coming from time to time :) So let's do this I think.

Re: call slot full in the alternative approach, I guess it is understandable, even it can also happen with `pjsua_var.hangup_call` approach. But yes, app should be able to check for any free call slot, either via notification or function call (not sure why we need a new inv state DESTROYED, better not touching PJSIP level if possible).

Some checklist (for all approach, perhaps already handled by this patch, not checked):
- PJSUA still should invoke callback `on_stream_destroyed()`, app may need it to destroy its custom component for example. So perhaps PJSUA should also deinit media immediately as if call is already disconnected.
- Handle/auto-reject incoming re-INVITE/UPDATE/other request silently (don't invoke PJSUA callback).

sauwming: Okay, I will do this in another branch and create a different PR since it is quite a different solution compared to this PR.

sauwming: A few notes when trying to implement the new approach:
- In general, there's a problem with the requirement in #1049: cease any PJSUA call related callback once call is declared to be disconnected. More about this below.
- `on_stream_destroyed()` will be called after the DISCONNECTED `on_call_state()`. But then, so is the current behaviour. Except that since we now report the DISCONNECTED state much earlier, the delay can be longer between the two callbacks. Trying to call `on_stream_destroyed()` earlier poses two problems:
  * The media will already be deactivated upon call disconnection, resulting in call dump not showing any useful end-of-call statistics. But this is minor and the workaround is probably to do the call dump before the media deinit.
  * The bigger problem is the async media transport creation, so no matter what, we still need to wait for it to complete. 
- How about `on_call_tsx_state()`? In #1627 (Callback on_tsx_state_changed() not invoked on call disconnection), the workaround is to call it before `on_call_state(DISCONNECTED)`, but since we now report the DISCONNECTED state earlier in `pjsua_call_hangup()`, we don't have the `pjsip_event`.

nanangizz: > A few notes when trying to implement the new approach:
> 
> * In general, there's a problem with the requirement in #1049: cease any PJSUA call related callback once call is declared to be disconnected. More about this below.
> * `on_stream_destroyed()` will be called after the DISCONNECTED `on_call_state()`. But then, so is the current behaviour. Except that since we now report the DISCONNECTED state much earlier, the delay can be longer between the two callbacks. Trying to call `on_stream_destroyed()` earlier poses two problems:
>   
>   * The media will already be deactivated upon call disconnection, resulting in call dump not showing any useful end-of-call statistics. But this is minor and the workaround is probably to do the call dump before the media deinit.

Ic.

>   * The bigger problem is the async media transport creation, so no matter what, we still need to wait for it to complete.

If `on_stream_created()` is not called yet (currently this should be the case when med tp is not completed yet), we can skip `on_stream_destroyed()' perhaps?

> * How about `on_call_tsx_state()`? In #1627 (Callback on_tsx_state_changed() not invoked on call disconnection), the workaround is to call it before `on_call_state(DISCONNECTED)`, but since we now report the DISCONNECTED state earlier in `pjsua_call_hangup()`, we don't have the `pjsip_event`.

Perhaps we can invoke the callback using user event?

sauwming: > >   * The media will already be deactivated upon call disconnection, resulting in call dump not showing any useful end-of-call statistics. But this is minor and the workaround is probably to do the call dump before the media deinit.

Calling `pjsua_call_dump()` in the application's `on_stream_destroyed()` seems to solve this issue.

> If `on_stream_created()` is not called yet (currently this should be the case when med tp is not completed yet), we can skip `on_stream_destroyed()' perhaps?

Yes.

> > * How about `on_call_tsx_state()`? In #1627 (Callback on_tsx_state_changed() not invoked on call disconnection), the workaround is to call it before `on_call_state(DISCONNECTED)`, but since we now report the DISCONNECTED state earlier in `pjsua_call_hangup()`, we don't have the `pjsip_event`.
> Perhaps we can invoke the callback using user event?

`on_call_tsx_state()` also has `pjsip_transaction *` parameter. Passing NULL may cause application to crash.

nanangizz: > `on_call_tsx_state()` also has `pjsip_transaction *` parameter. Passing NULL may cause application to crash.

Ah okay.
For immediate hangup, perhaps we should just skip `on_call_tsx_state()` altogether as any tsx state change, if any, happens in background. And the docs should explicitly specify this.

sauwming:
Closing this. Used a different approach in #2600.

pjsip/include/pjsua-lib/pjsua.h
calanor: About calling pjsua_call_dump() in on_stream_destroyed().
With this change, if we receive the hang up, the call_dump appears as "DISCONNCTD", but if we are the ones who hang up the call, it appears as "CONFIRMED".
Before, in on_call_state, in all cases the call dump appears as "DISCONNCTD".

sauwming: Will continue this discussion in the relevant PR #2600 instead.

PR #2588, opened 2020-11-20
Implement trickle ICE

### LINKED ISSUE #2090: Support trickle ICE
_2018-02-12 11:11:15: @nanangizz_ _created the issue on trac ticket [2090](https://trac.pjsip.org/repos/ticket/2090)_

https://tools.ietf.org/html/draft-ietf-ice-trickle-16:
```
   ... an extension to the
   Interactive Connectivity Establishment (ICE) protocol that enables
   ICE agents to send and receive candidates incrementally rather than
   exchanging complete lists.  With such incremental provisioning, ICE
   agents can begin connectivity checks while they are still gathering
   candidates and considerably shorten the time necessary for ICE
   processing to complete.
```

nanangizz:
### PJSUA
- enable and configure ICE as usual.
- global setting: set `pjsua_media_config.ice_opt.trickle` to `PJ_ICE_SESS_TRICKLE_HALF` or `PJ_ICE_SESS_TRICKLE_FULL`.
- account setting:
  - set `pjsua_acc_config.ice_cfg_use` to `PJSUA_ICE_CONFIG_USE_CUSTOM`
  - set `pjsua_acc_config.ice_cfg.ice_opt.trickle` to `PJ_ICE_SESS_TRICKLE_HALF` or `PJ_ICE_SESS_TRICKLE_FULL`.

### PJSUA2
- enable and configure ICE as usual.
- set `AccountConfig.natConfig.iceTrickle` to `PJ_ICE_SESS_TRICKLE_HALF` or `PJ_ICE_SESS_TRICKLE_FULL`.

### PJSUA sample app
- enable and configure ICE as usual, e.g: `--use-ice`, STUN & TURN settings.
- use pjsua param `--ice-trickle=N` (N=0:disabled, 1:half, 2:full). 

### ICE session
- `pj_ice_sess_options.trickle`
- `pj_ice_sess_update_check_list()`
### ICE transport
- `pj_ice_strans_cb.on_new_candidate()`
- `pj_ice_strans_update_check_list()`
### Media transport ICE
- `pjmedia_ice_cb.on_new_candidate()`
- `pjmedia_ice_sdp_has_trickle()`
- `pjmedia_ice_trickle_encode_sdp()` & `pjmedia_ice_trickle_decode_sdp()`
- `pjmedia_ice_trickle_send_local_cand()`
- `pjmedia_ice_trickle_update()`

## References
- ICE: https://tools.ietf.org/html/rfc8445
  trickle ICE is based on RFC8445 instead of RFC5245, this PR does not update the PJNATH ICE implementation to fully comply with this RFC though.
- Trickle ICE: https://tools.ietf.org/html/draft-ietf-ice-trickle-21
- Trickle ICE on SIP: https://tools.ietf.org/html/draft-ietf-mmusic-trickle-ice-sip-18
- SIP INFO Package: https://tools.ietf.org/html/rfc6086
- SDP for ICE: https://tools.ietf.org/html/draft-ietf-mmusic-ice-sip-sdp-39
- SDP media ID: https://tools.ietf.org/html/rfc5888

Note that not all of the above RFC/drafts are fully implemented, only the parts that are related to trickle ICE.
This PR will close #2090.


pjnath/include/pjnath/ice_strans.h
```diff
@@ -999,16 +1061,19 @@ PJ_DECL(pj_status_t) pj_ice_strans_sendto(pj_ice_strans *ice_st,
...
- * If ICE negotiation is in progress, this function will send the data 
- * to the destination address. Otherwise if ICE negotiation has completed
- * successfully, this function will send the data to the nominated remote 
- * address, as negotiated by ICE.
+ * the address in SDP c= and m= lines, or in a=rtcp attribute).
+ * If ICE negotiation is in progress, this function will try to send the data
+ * via any valid candidate pair (which has passed ICE connectivity test).
+ * If ICE negotiation has completed successfully, this function will send
+ * the data to the nominated remote address, as negotiated by ICE.
+ * If the ICE negotiation fails or valid candidate pair is not available,
```
sauwming, line 1074: not yet available?

pjmedia/src/pjmedia/transport_ice.c
```diff
@@ -773,6 +1132,31 @@ static pj_status_t encode_session_in_sdp(struct transport_ice *tp_ice,
...
+
+	ice_state = pj_ice_strans_get_state(tp_ice->ice_st);
+	end_of_cand = (ice_state == PJ_ICE_STRANS_STATE_READY ||
+		       ice_state == PJ_ICE_STRANS_STATE_SESS_READY);
+
+	status = pjmedia_ice_trickle_encode_sdp(sdp_pool, sdp_local,
+						&tp_ice->sdp_mid, NULL, NULL,
+						0, NULL, end_of_cand);
+	if (status != PJ_SUCCESS) {
+	    pj_assert(!"pjmedia_ice_trickle_encode_sdp() failed");
```
sauwming, line 1155: Not sure why we need to assert here.
There are already assertions in place inside the function itself, which is the more appropriate place, and there's also a legitimate error as well (if too many media) although it's a rare case.

pjnath/src/pjnath/ice_session.c
```diff
@@ -1723,60 +1838,279 @@ PJ_DEF(pj_status_t) pj_ice_sess_create_check_list(
...
+	return status;
+    }
 
     /* Log checklist */
     dump_checklist("Checklist created:", ice, clist);
 
     pj_grp_lock_release(ice->grp_lock);
 
-    return PJ_SUCCESS;
+    return status;
```
sauwming, line 2059: it seems like it always returns PJ_SUCCESS here

pjnath/src/pjnath/ice_session.c
```diff
@@ -1723,60 +1838,279 @@ PJ_DEF(pj_status_t) pj_ice_sess_create_check_list(
...
+    if (trickle_done) {
+	LOG5((ice->obj_name, "Trickling done."));
+	ice->is_trickling = PJ_FALSE;
+    }
+
+    /* Verify remote ufrag & passwd, if remote candidate specified */
+    if (rem_cand_cnt && (pj_strcmp(&ice->tx_ufrag, rem_ufrag) ||
+			 pj_strcmp(&ice->tx_pass, rem_passwd)))
+    {
+	status = PJ_EINVAL;
```
sauwming, line 2100: Add log. And better return status (if possible)

pjnath/src/pjnath/ice_session.c
```diff
@@ -2396,6 +2738,47 @@ static void on_stun_request_complete(pj_stun_session *stun_sess,
 	{
 	    /* Match */
 	    lcand = &ice->lcand[i];
+
+#if 0
```
sauwming, line 2742: why is this disabled?

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -4147,6 +4171,452 @@ static pj_status_t process_pending_reinvite(pjsua_call *call)
...
+	      call->index, call->trickle_ice.retrans18x_count));
+
+    pjsip_tx_data_add_ref(tdata);
+    pjsip_tsx_retransmit_no_state(call->inv->invite_tsx, tdata);
+
+    /* Schedule next retransmission */
+    if (call->trickle_ice.retrans18x_count < 6) {
+	delay.sec = 0;
+	delay.msec = (1 << call->trickle_ice.retrans18x_count)*
+		     pjsip_cfg()->tsx.t1;
```
sauwming, line 4210: From Code scanning
"Multiplication result converted to larger type
Multiplication result may overflow 'unsigned int' before it is converted to 'long'.
"

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -4147,6 +4171,452 @@ static pj_status_t process_pending_reinvite(pjsua_call *call)
...
+    pj_str_t CONTENT_DISP_STR	= {"Content-Disposition", 19};
+    pj_str_t INFO_PKG_STR	= {"Info-Package", 12};
+    pj_str_t TRICKLE_ICE_STR	= {"trickle-ice", 11};
+
+    PJ_UNUSED_ARG(th);
+
+    PJSUA_LOCK();
+
+    /* Check provisional media or active media to use */
+    use_med_prov = call->med_prov_cnt > call->med_cnt;
```
sauwming, line 4312: I'm wondering if this is the right check? If media is not being added, is it correct to always use the active media?

Btw, there's already an internal function: pjsua_call_media_is_changing() for the same check.

sauwming (commented):
Ups, I didn't realise I was sending out the comment one by one.
Anyway, the first iteration of my review is done.

pjsip/src/pjsua-lib/pjsua_call.c
nanangizz: not sure the correct solution, will just copy the pattern of [this code](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_transaction.c#L2345).

pjsip/src/pjsua-lib/pjsua_call.c
nanangizz: Actually I cannot remember exactly the lifetime (or the validity period) of the provisional media, while currently the `trickle_ice_send_sip_info()` does not acquire call/dlg lock, so its execution may be started before SDP nego but ended after SDP nego. The call media is needed only to make sure that the media transport base is ICE, changing transport base does not seem to be a common case. So active media should be the safer choice?

Or perhaps it is better to always protect the function with call/dlg lock (a bit costly considering the function will be called every 100ms until all ICE candidates are gathered) and evaluate provisional media whenever it is valid (med_prov_cnt>0?) rather than active media as generally provisional media is more updated.

pjmedia/src/pjmedia/transport_ice.c
nanangizz: Replaced the assertion with an error log.

pjnath/src/pjnath/ice_session.c
nanangizz: done (added comment)

pjsip/src/pjsua-lib/pjsua_call.c
sauwming: If only to check the transport type, I think we can leave it as it is now.

sauwming (approved):

trengginas (approved):

pjnath/include/pjnath/config.h
```diff
@@ -286,7 +286,7 @@
 /**
  * Maximum number of ICE components.
  */
-#define PJ_ICE_MAX_COMP				    (2<<PJ_ICE_COMP_BITS)
+#define PJ_ICE_MAX_COMP				    (1<<PJ_ICE_COMP_BITS)
```
AmarOk1412, line 289: @nanangizz question. What motivated this change?

nanangizz: Shouldn't it the correct formula? E.g: if `COMP_BITS` is 1, it can only represent maximum of 2 components, the prev formula will result in 4 components.

Anything becomes not working or having a problem after that change? :)

AmarOk1412: Yeah you are right. In our case it COMP_BITS should be 4

For the little story, I started to rebase our code on pjsip 2.11 (we're using 2.10) and got some assertions failure when the call starts with
https://review.jami.net/c/ring-daemon/+/18710/3/contrib/src/pjproject/0007-upnp-srflx-nat-assisted-cand.patch

But our ICE uses 4 components for medias, so COMP_BITS should be set to 2

PR #2589, opened 2020-11-21
Fix building with video support (including dshow camera) on mingw

zhubr:
This patchset allows building with dshow camera support in most current msys/minw64-i686 environment and also fixes some issues with SDL2 linking on mingw. Running the produced pjmedia-test on windows xp shows some webcam windows with apparently correct picture and finally says "Looks like everything is okay!". Note1: this behaviour is only activated if user explicitely specifies --enable-video=yes, otherwise behaviour is unchanged on mingw; Note2: successfully building on msys may also require some trivial corrections to Makefile which are unrelated to video devices and are not present in this patchset.


pjsip/build/Makefile
```diff
@@ -271,15 +271,15 @@ $(TEST_EXE): $(PJSIP_UA_LIB) $(PJSIP_UA_SONAME)
 $(TEST_EXE): $(PJSIP_SIMPLE_LIB) $(PJSIP_SIMPLE_SONAME)
 $(TEST_EXE): $(PJSUA_LIB_LIB) $(PJSUA_LIB_SONAME)
 $(TEST_EXE):
-	$(MAKE) -f $(RULES_MAK) APP=TEST app=pjsip-test $(subst /,$(HOST_PSEP),$(BINDIR)/$@)
+	CFLAGS="$(CFLAGS) $(SDL_CFLAGS)" $(MAKE) -f $(RULES_MAK) APP=TEST app=pjsip-test $(subst /,$(HOST_PSEP),$(BINDIR)/$@)
```
sauwming, line 274: shouldn't SDL_CFLAGS already be included in PJ_VIDEO_CFLAGS?

In any case, the addition of flags for such specific library is better not to be put here. Same for below, as well as for Samples.mak above (find out why the existing PJ_VIDEO_CFLAGS is not working and how to use it, if possible).

zhubr: Yes I see your point and I'll comment a bit on this part. Thing is, SDL on windows has a specific peculiarity that "main" needs to be redefined to "SDL_main" for the SDL library to get its necessary hook for some internal initialization and finalization (in its "substitute" "main"). To achieve that, SDL usually pushes "-Dmain=SDL_main" into its cflags (See e.g. sdl2.pc) and it is then supposed to be picked up by an application. Now because in pjproject most applications/examples do not apparently suppose any sort of SDL playback (on windows anyway), I felt a bit reluctant to somehow get all the execatables affected globally by this messy SDL hackery and tried to only fix those specific executables which needed it. Maybe it could be somehow arranged better but I'm not exactly sure yet. Please let me know what you think.

nanangizz:
Is this for [Mingw-w64](http://mingw-w64.org/)? Officially we don't support that platform (yet). Currently we only support the standard msys/MinGW and without video, so it is interesting to have video working on Mingw platform family. You also mention about 'trivial corrections to Makefile required', is it also specifically for Mingw-w64? Thank you for the patch anyway.

zhubr:
> Is this for [Mingw-w64](http://mingw-w64.org/)? Officially we don't support that platform (yet). Currently we only support the standard msys/MinGW and without video, so it is interesting to have video working on Mingw platform family. You also mention about 'trivial corrections to Makefile required', is it also specifically for Mingw-w64? Thank you for the patch anyway.

Yes, it is for mingw-w64, although 32-bit version. Actually I just supposed mingw-w64 is what everybody have already long switched to.
My additional corrections were as follows. The first one fixes a problem with path translation (one extra slash somehow confused msys path substitution on my system, but it might be that DESTDIR and/or libdir were set incorrectly), the second avoids direct usage of ld in favour of gcc for linking (because otherwise ld was missing some arguments and/or paths and AFAIK gcc is now the recommended method of linking on modern gnu platforms anyway) and prefers static linking so that resulting executables were independent of mingw libraries at runtime, as by default this might now always be the case. These changes need some more checking I think.
<pre>
--- a/Makefile	Thu Nov 12 15:16:20 2020
+++ b/Makefile	Thu Nov 12 15:45:19 2020
@@ -130,7 +130,7 @@
 		sed -e "s!@LIBDIR@!$(libdir)!" | \
 		sed -e "s/@PJ_VERSION@/$(PJ_VERSION)/" | \
 		sed -e "s!@PJ_INSTALL_LDFLAGS@!$(PJ_INSTALL_LDFLAGS)!" | \
-		sed -e "s!@PJ_INSTALL_CFLAGS@!$(PJ_INSTALL_CFLAGS)!" > $(DESTDIR)/$(libdir)/pkgconfig/libpjproject.pc
+		sed -e "s!@PJ_INSTALL_CFLAGS@!$(PJ_INSTALL_CFLAGS)!" > $(DESTDIR)$(libdir)/pkgconfig/libpjproject.pc
 
 uninstall:
 	$(RM) $(DESTDIR)$(libdir)/pkgconfig/libpjproject.pc
--- a/aconfigure.ac	Mon Nov 16 20:56:40 2020
+++ b/aconfigure.ac	Mon Nov 16 21:34:33 2020
@@ -44,6 +44,12 @@
 AC_PROG_RANLIB
 AC_CHECK_TOOLS([AR], [ar gar], :)
 
+case $target in
+ *mingw*)
+	LD="$CC -static"
+	;;
+esac
+
 if test "$AR_FLAGS" = ""; then AR_FLAGS="rv"; fi
 AC_SUBST(AR_FLAGS)
 if test "$LD" = ""; then LD="$CC"; fi
</pre>

pjsip/build/Makefile
sauwming: I don't know a solution for this either. All I know is that it seems out of place.

nanangizz:
> Yes, it is for mingw-w64, although 32-bit version. Actually I just supposed mingw-w64 is what everybody have already long switched to.

Thanks for the info, will try to build PJSIP on mingw-w64 later.

pjsip/build/Makefile
zhubr: I was thinking about maybe introducing some additional PJ_SDL_EXE_EXTRA specifically for this. But looking at root build.mak again it seems APP_CFLAGS and APP_CXXFLAGS should already include the necessary SDL_CFLAGS. Unless they are redefined elsewhere or ignored. The variables definition and parsing system seems complicated and flexible already, I'd suppose it could allow fixing this SDL thing in a better way, but I need some hints.

zhubr:
> Thanks for the info, will try to build PJSIP on mingw-w64 later.

Great! For reference, I think my initial environment preparation (just after running mingw-w64 installer) was basically this:
<pre>
pacman -S base-devel mingw-w64-i686-toolchain
</pre>
Main thing to keep in mind is to not ocasionally pick wrong flavour of standard tools such as "make". In my experience those from "msys" group (base-devel package) are closer to native linux versions and are easier to deal with, as compared to those from "mingw64" group. HTH.

pjsip/build/Makefile
sauwming: After looking at it again, I believe the modification should be as follows instead:
`export PJSUA2_TEST_CXXFLAGS = $(_CXXFLAGS) $(PJSUA2_LIB_CFLAGS)  $(PJ_VIDEO_CFLAGS)`
and
`export TEST_CFLAGS += $(_CFLAGS) $(PJ_VIDEO_CFLAGS)`

And for the one in `Samples.mak`:
```
export _CFLAGS 	:= $(PJ_CFLAGS) $(CFLAGS) $(PJ_VIDEO_CFLAGS)
export _CXXFLAGS:= $(PJ_CXXFLAGS) $(CFLAGS) $(PJ_VIDEO_CFLAGS)
```

Test it there and let me know if it works.

zhubr: Yes it does! Thank you!
Now should I push such corrections to this PR myself?
(Please excuse my ignorance, I'm not very familiar with github's workflow)

sauwming: Yes please.

zhubr: Corrections have been pushed successfully.

sauwming (approved):

nanangizz:
@zhubr Re: Mingw-w64 support, just created a new PR #2598 for it (only based on my tests). Please check and feel free to drop any feedbacks. You may notice that the PR does not include `LD="$CC -static"`, still not sure if it is necessary, because here in my msys2 env, `LD` is already using `gcc`, and for the `-static` part, perhaps let's just leave it to user preference as in other GNU platforms, e.g: user may set it via `LDFLAGS` (let's continue the discussion in the new PR, if you want to). Thanks.

aconfigure.ac
```diff
@@ -823,6 +825,18 @@ else
...
 	;;
+  *mingw*)
+	if test "$enable_video" = "yes"; then
+	  ac_pjmedia_video=windows_os
+	  AC_SUBST(ac_pjmedia_video_dev_has_dshow)
+	  ac_pjmedia_video_dev_has_dshow=yes
+	  AC_SUBST(ac_v4l2_cflags)
+	  ac_v4l2_cflags="-DPJMEDIA_HAS_VIDEO=1 -DPJMEDIA_VIDEO_DEV_HAS_DSHOW=1"
+	  AC_SUBST(ac_v4l2_ldflags)
+	  ac_v4l2_ldflags=" -lstdc++ -lquartz -lole32 -loleaut32 -lrpcrt4 -lwinmm -luuid -lmingwex -lstrmiids "
```
zhubr, line 836: I'd like to draw a bit of attention that I'm somewhat abusing the v4l2 variables here, for the lack of e.g. ac_dshow_cflags. This might look unclean and/or confusing, but I think it is basically a small consmetic issue, if one thinks of "ac_v4l2_cflags" as like just "ac_camera_cflags". It does exactly that job, and obviously directshow camera and v4l2 camera can not possibly get in conflict in any way, so supposedly no breakage would happen.

sauwming: It would be better to create a separate variable ac_dshow_flags & ldflags. Just search the entire pjsip project for the string `ac_v4l2_cflags` to find out all the places it's being used and add the new variable.

nanangizz:
And one more thing, could you please also include the updated `aconfigure` script in the patch. Thanks.

aconfigure.ac
zhubr: Ok, done.

zhubr:
> And one more thing, could you please also include the updated `aconfigure` script in the patch. Thanks.
Done.
I'm now observing a failure in turn_sock_test at build-mac-video-openh264-2, whereas the same test succeeds e.g. at build-mac-default-full-bundle-2 (and all others actually). It looks like maybe some ocasional problem on a build machine, not sure. This specific test seems totally unrelated of my changes anyway.

nanangizz:
> I'm now observing a failure in turn_sock_test at build-mac-video-openh264-2, whereas the same test succeeds e.g. at build-mac-default-full-bundle-2 (and all others actually). It looks like maybe some ocasional problem on a build machine, not sure. This specific test seems totally unrelated of my changes anyway.

Yes, it is occasional problem and totally unrelated indeed, so don't worry about that.

Just FYI, in the other dev branch (trickle ICE), tried to add more debugging info and looks like the problem is about "address already in use". The test app performs many iterations reusing the same port number, even if the socket is closed properly (from app PoV) in each iteration, the next iteration may sometime still fail with "address already in use" error. This specific issue and location should have been fixed, in that branch only though.

trengginas (approved):

PR #2591, opened 2020-11-23
Support new Mac M1 ARM64 architecture

sauwming:
Currently, PJSIP fails to build on the new Mac ARM64 architecture because it selects SSE2 for building WebRTC AEC. This PR will fix this and select the right architecture.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -182,7 +182,7 @@ jobs:
 
   build-mac-default:
   # checking pure lib source distribution with plain configure & make
-    runs-on: macos-latest
+    runs-on: macos-11.0
     steps:
     - uses: actions/checkout@v2
     - name: configure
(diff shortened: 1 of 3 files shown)


nanangizz:
The title is quite generic but the desc only mentions WebRTC AEC, is it the only component that is failing? Or perhaps add info about tested configs & external libs.

If GitHub CI supports this platform already, perhaps we could also add this new platform.

sauwming:
Yes, only WebRTC AEC fails in test build and run pjsua audio+video (Video Toolbox + SDL).

GitHub supports the new macOS (macOS 11.0, although currently still for preview) but I doubt they're using ARM64 machine. Nevertheless, let me try.

sauwming:
They're using the usual x86_64 machine.
`checking host system type... x86_64-apple-darwin20.1.0`

nanangizz (approved):

trengginas (approved):

nfranke:
I was able to build it for arm64, but I can't actually use it in a project since these symbols are undefined: 

#define _ARM_ 1
#define PJ_IS_LITTLE_ENDIAN 1
#define PJ_IS_BIG_ENDIAN 0

Issue #2594, opened 2020-11-25
Pass non-enumerated ciphers to openssl, i.e. SECLEVEL=n


naf419:
Users of pjlib cannot configure the openssl tls connections to change openssl "security level" other than by changing the system default. Within openssl, the "security level" specifies (among other things) the minimum certificate key length. This means to allow a pjlib-linked application to initiate a tls connection with a 1024-bit key, the only way is to lower security for the entire system. It would be better if pjlib could be configured in a way that it would pass the a cipher like the common "DEFAULT@SECLEVEL=1" when creating the openssl tls connection.

--Details--
Openssl provides atleast 3 ways to specify the security level: the openssl_conf config file, using SSL_set_security_level() and through SSL_set_cipher_list() via the special "@SECLEVEL=n" cipher:
- pjsip doesn't honor the OPENSSL_CONF environment variable to specify an application-specific configuration file, so the system default is always used (honoring OPENSSL_CONF would be *better* because the env variable doesn't affect the global config, but still not connection-by-connection specific)
- pjsip doesn't use SSL_set_security_level() (and probably shouldn't - its very openssl-specific)
- pjsip does use SSL_set_cipher_list(), but the cipher list is pre-verified to only contain items contained in the "ALL:COMPLEMENTOFALL" cipher list. This prevents using the special "@SECLEVEL=n" specifier. Not sure how to make that happen with the pjsip interface using an array of enumerated ciphers instead of the openssl string style (realizing that openssl isn't the only tls backend)

sauwming:
We are open to any contribution for this :)

PR #2596, opened 2020-11-27
Add support for SECLEVEL=n

naf419:
Here's my suggested method of adding support for the SECLEVEL psuedo-cipher in openssl tls connections. (#2594)

Note that the setting the cipher list has been moved to be before adding certificates, because the success of adding the certificates depends on the current security level.

### DIFF
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -394,6 +394,22 @@ static pj_str_t ssl_strerror(pj_status_t status,
     return errstr;
 }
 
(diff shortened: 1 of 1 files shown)


sauwming:
It failed to build on Mac. You can check the CI result above for the details.

naf419:
Ya, looks like all openssl version don't provide SSL_CTX_get_ciphers().
No worries, we conveniently have all the ciphers cached already so we don't need it anyway.

nanangizz (approved):

sauwming (approved):

PR #2598, opened 2020-12-01
Supporting mingw-w64

nanangizz:
This is the initial works for supporting [Mingw-w64](http://mingw-w64.org/).

### Steps
1. Install [msys2](https://www.msys2.org/).
2. Open just installed app `MSYS2 MinGW 64-bit`, terminal should be shown. Alternatively, you can use `MSYS2 MinGW 32-bit` for 32 bit target, don't forget to also use the appropriate msys2 package version (e.g: with `i686` suffix instead of `x86_64`).
3. Install at least `gcc` & `make` (and optionally other dev tools as needed), e.g:
    ```
    pacman -S mingw-w64-x86_64-gcc
    pacman -S make
    pacman -S autoconf # this is optional
    ```
    Alternatively, install base development and mingw-w64 toolchains packages:
    ```
    pacman -S base-devel mingw-w64-x86_64-toolchain
    ```
4. Verify the environment:
   - `uname` command should print something like `MINGW64_NT-x-y` (instead of `MSYS_NT-x-y`).
   - `gcc -dumpmachine` command should print `x86_64-w64-mingw32` (instead of `x86_64-pc-msys`).
5. Get [PJSIP source code](https://trac.pjsip.org/repos/wiki/Getting-Started/Download-Source) and follow the [Getting Started](https://trac.pjsip.org/repos/wiki/Getting-Started/Autoconf) wiki page.
6. Apply patch from this pull request (if you are using 2.10 or older).

### Video support
1. Check the [Video Support](https://trac.pjsip.org/repos/wiki/Getting-Started/Autoconf#VideoSupportfor2.0andabove) section in the Getting Started wiki page.
2. Install dependency packages, e.g:
    ```
    pacman -S mingw-w64-x86_64-SDL2
    pacman -S mingw-w64-x86_64-openh264
    ``` 
3. Check pull request #2589 and apply its patch (if you are using 2.10 or older).
4. Run `configure --enable-video=yes`, make sure the dependency packages are correctly detected, then `make` as usual.

### Credits
Thanks to Nikolai ZHUBR (@zhubr) for the heads-up and the initial patch for supporting Mingw-w64 platform.

### DIFF
--- a/aconfigure.ac
+++ b/aconfigure.ac
@@ -165,7 +165,7 @@ case $target in
 	AC_DEFINE(PJ_WIN32_WINNT,0x0400)
 	AC_DEFINE(WIN32_LEAN_AND_MEAN)
 	case $target in
-	    *w64-mingw* )
+	    *_64-w64-mingw* )
 		AC_DEFINE(PJ_WIN64,1)
 	    ;;
 	esac
--- a/aconfigure.ac
+++ b/aconfigure.ac
@@ -164,6 +164,11 @@ case $target in
 	AC_DEFINE(PJ_WIN32,1)
 	AC_DEFINE(PJ_WIN32_WINNT,0x0400)
 	AC_DEFINE(WIN32_LEAN_AND_MEAN)
+	case $target in
+	    *w64-mingw* )
+		AC_DEFINE(PJ_WIN64,1)
+	    ;;
+	esac
 	;;
     *darwin*)
 	AC_DEFINE(PJ_DARWINOS,1)
--- a/pjnath/src/pjnath-test/stun_sock_test.c
+++ b/pjnath/src/pjnath-test/stun_sock_test.c
@@ -25,8 +25,7 @@ enum {
     RESPOND_STUN    = 1,
     WITH_MAPPED	    = 2,
     WITH_XOR_MAPPED = 4,
-
-    ECHO	    = 8
+    SEND_ECHO	    = 8
 };
 
 /*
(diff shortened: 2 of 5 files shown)


sauwming (approved):

aconfigure.ac
```diff
@@ -164,6 +164,11 @@ case $target in
 	AC_DEFINE(PJ_WIN32,1)
 	AC_DEFINE(PJ_WIN32_WINNT,0x0400)
 	AC_DEFINE(WIN32_LEAN_AND_MEAN)
+	case $target in
+	    *w64-mingw* )
```
zhubr, line 168: I think this should be something like *_64-w64-mingw* instead. Because otherwise "i686-w64-mingw32" here also sets PJ_WIN64 to 1 while being 32-bit windows. And while apparently this does not break the build, I suppose the intent was to only set PJ_WIN64 for a 64-bit system.

pjnath/src/pjnath-test/stun_sock_test.c
```diff
@@ -25,8 +25,7 @@ enum {
     RESPOND_STUN    = 1,
     WITH_MAPPED	    = 2,
     WITH_XOR_MAPPED = 4,
-
-    ECHO	    = 8
+    SEND_ECHO	    = 8
```
zhubr, line 28: Looks like name clash prevention, but apparently unrelated of mingw64? It compiles fine both midified and unmodified. The only place I see ECHO defined is netmon.h of i686-w64-mingw32, but apparently netmon.h is not used in pjsip? Maybe this change was intended for some other PR?

nanangizz: Got compile error previously (yes, name clash), but it was perhaps because of using wrong headers/toolchains (was using msys's instead of mingw's at some point). Anyway, just tested again without the change and it compiles fine now indeed, so will revert the change. Thanks.

trengginas (approved):

PR #2600, opened 2020-12-02
Report disconnection event immediately when hanging up a call

sauwming:
To continue the work in #1049 and close it.
* When calling `pjsua_call_hangup()`, application will immediately receive `on_call_state()` callback with DISCONNECTED state. No further call related callbacks will be received after this. The call hangup process itself (sending BYE, waiting for the response, and resource cleanup) will continue in the background and the call slot can be reused only after this process is completed.
* Retry creating and sending BYE after a period of `CALL_HANGUP_RETRY_INTERVAL` if it fails. If this keeps happening after `CALL_HANGUP_MAX_RETRY` times, we will hard-terminate the call, i.e. cleanup the call without sending any BYE to the remote.

This will introduce some behavioural changes though:
- Media will be deinitialised immediately upon hanging up, before disconnected state is reported. So media callbacks such as `on_stream_destroyed()` will be called before `on_call_state()`. Previously, it was the reverse.
- Application can no longer dump end-of-call media stats upon receiving disconnection report. This is because media has been deactivated by then. One alternative is to dump media stats in `on_stream_destroyed()`, but note that `on_stream_destroyed()` can also be called during media update, can be called multiple times if there are more than one audio streams, and won't be called if there is no audio (such as in video-only calls).

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_call.c
+++ b/pjsip/src/pjsua-lib/pjsua_call.c
@@ -1037,7 +1037,7 @@ static pj_status_t process_incoming_call_replace(pjsua_call *call,
     replaced_call = (pjsua_call*) replaced_dlg->mod_data[pjsua_var.mod.id];
 
     /* Notify application */
-    if (pjsua_var.ua_cfg.cb.on_call_replaced)
+    if (!replaced_call->hanging_up && pjsua_var.ua_cfg.cb.on_call_replaced)
 	pjsua_var.ua_cfg.cb.on_call_replaced(replaced_call->index,
 					     call->index);
 
@@ -1506,13 +1506,17 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
 	pjsua_call_cleanup_flag(&call->opt);
 
 	/* Notify application */
-	if (pjsua_var.ua_cfg.cb.on_call_replace_request) {
+	if (!replaced_call->hanging_up &&
+	    pjsua_var.ua_cfg.cb.on_call_replace_request)
+	{
 	    pjsua_var.ua_cfg.cb.on_call_replace_request(replaced_call->index,
 							rdata,
 							&st_code, &st_text);
 	}
 
-	if (pjsua_var.ua_cfg.cb.on_call_replace_request2) {
+	if (!replaced_call->hanging_up &&
+	    pjsua_var.ua_cfg.cb.on_call_replace_request2)
+	{
 	    pjsua_var.ua_cfg.cb.on_call_replace_request2(replaced_call->index,
 							 rdata,
 							 &st_code, &st_text,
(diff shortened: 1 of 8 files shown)


pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -2811,16 +2778,175 @@ PJ_DEF(pj_status_t) pjsua_call_hangup(pjsua_call_id call_id,
...
+	PJSIP_EVENT_INIT_USER(user_event, 0, 0, 0, 0);
+    	if (pjsua_var.ua_cfg.cb.on_call_state) {
+	    (*pjsua_var.ua_cfg.cb.on_call_state)(call->index,
+	    					 &user_event);
+	}
+
+    }
+
+    if (call->inv)
+    	call_inv_end_session(call, code, reason, msg_data);
```
nanangizz, line 2948: If `pjsua_call_hangup()` is called multiple times (for the same call), is it handled already (e.g: avoid rescheduling an active timer)? The scenario is perhaps not uncommon, e.g: hangup button in the UI not disabled immediately so the user may press it multiple times (by accident or not) and the app code simply invokes `pjsua_call_hangup()` whenever the button is pressed.

sauwming: Interesting scenario.

Here's the result of my test:
When `pjsua_call_hangup() is called more than once, each will send a 'BYE' request. This is consistent with the current behaviour.
But with this patch, the scenario (trying to hangup call more than once) can be easily avoidable since app will get the DISCONNECTED state immediately during the first hangup attempt.

I have added the timer check for more robustness.

nanangizz:
Just recalled about the idea of adding an API to get info about free call slot count. But then perhaps it can be calculated from `pjsua_call_get_max_count()-pjsua_call_get_count()`, so I guess it is not necessary? Or perhaps we should add info about it in the `pjsua_call_hangup()` docs.

nanangizz (approved):

trengginas (approved):

sauwming:
Comment from @calanor in #2582
> About calling pjsua_call_dump() in on_stream_destroyed().
> With this change, if we receive the hang up, the call_dump appears as "DISCONNCTD", but if we are the ones who hang up the call, it appears as "CONFIRMED".
> Before, in on_call_state, in all cases the call dump appears as "DISCONNCTD".

Yes, that's true. I was thinking of introducing a flag to detect this, but was not sure if it's necessary or if it's going to be useful. Let us know if this causes any complications.

calanor:
No, it is not a problem for me. I was just saying it for your information.

MasterG1S:
I am developing a sip application for Android. Faced such a situation: 
I am starting to display preview and incoming video. At the same time, I am receiving BYE from the other side. 
It turns out at the same time I start the video and try to finish it. The library at this moment freezes the codec: Encoder dequeueInputBuffer failed [-1]. Hang occurs in pjsua_vid_stop_stream(...) when unsubscribed from all events.
I would like to know how to immediately get a callback when a BYE is received, so as not to start trying to preview or incoming video display.

Regards, Ivan.

PR #2606, opened 2020-12-11
TLS failed to load password-protected private key from buffer

nanangizz:
When loading TLS private key from buffer, password callback is not set.

Thank you Marcus Froeschl for the report and the analysis.

This PR also do the following:
- add more logs in loading certificate/key from, e.g: info about successful/failure, from files or buffer.
- update TLS certificate in SSL socket unit test.


sauwming (approved):

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1141,13 +1159,17 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
 								  NULL, NULL);
 
 		if (inf != NULL) {
-		    int i = 0;		    
+		    int i = 0, cnt = 0;
 		    for (; i < sk_X509_INFO_num(inf); i++) {
 			X509_INFO *itmp = sk_X509_INFO_value(inf, i);
 			if (itmp->x509) {
 			    X509_STORE_add_cert(cts, itmp->x509);
```
mfroeschl, line 1166: @nanangizz Maybe check return value of `X509_STORE_add_cert()` as well?

trengginas (approved):

PR #2609, opened 2020-12-17
Use SRTP attribute specified on SDP when starting SRTP-SDES transport

trengginas:
There's a requirement to change SRTP-SDES policy before the media transport is started. 
e.g: offerer side change the SRTP key with the key specified on the SDP answer.
In this case, it is recommended the use of Transport Adapter (TA) to modify the key.
The general flow would be:
```
TA::encode_sdp() {
   parent/srtp->encode_sdp(&local_sdp); // call parent's encode_sdp() before updating local SDP
   if (answerer) {
     update_local_srtp_key(local_sdp);
   }
}
TA::media_start() {
  if (offerer) {
     update_local_srtp_key(local_sdp);
  }
  parent/srtp->media_start(); // call parent's media_start() after updating local SDP
}
```

This PR will check for changes on SRTP policy based on the local SDP supplied.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp_sdes.c
+++ b/pjmedia/src/pjmedia/transport_srtp_sdes.c
@@ -647,6 +647,40 @@ static pj_status_t sdes_media_start( pjmedia_transport *tp,
 	    srtp->peer_use = PJMEDIA_SRTP_OPTIONAL;
     }
 
+    /* For answerer side, SRTP crypto policies have been populated in
+     * media_encode_sdp(). Check if the key changes on the local SDP.
+     */
+    if (!srtp->offerer_side) {
+        if (srtp->tx_policy_neg.name.slen == 0)
+            return PJ_SUCCESS;
+
+        /* Get the local crypto. */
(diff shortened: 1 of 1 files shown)


nanangizz:
Looks like it always fails on pjsua test scripts-sipp_uac-srtp-sdes-reinv-dtls, please check.

pjmedia/src/pjmedia/transport_srtp_sdes.c
```diff
@@ -680,6 +674,31 @@ static pj_status_t sdes_media_start( pjmedia_transport *tp,
...
+                {
+                    pj_strdup(pool, &srtp->setting.crypto[i].key,
+                              &loc_crypto[0].key);
+                }
+            }
+	}
+
+        return PJ_SUCCESS;
+    }
+
```
nanangizz, line 701: It should be safer to put this block in its original place, note that the `srtp->setting.use` condition checks+actions are originally for offerer side (so action may be redundant or even inappropriate). If the block needs to call `fill_local_crypto()`, just call it by itself (e.g: with additional safety check `srtp->tx_policy_neg.key.slen > 0`).

nanangizz (approved):

sauwming (approved):

PR #2612, opened 2020-12-23
Fix sound output issue when running iOS apps on Mac

sauwming:
With the new Apple's ARM-based M1 Mac, iOS apps can now run natively on those ARM Mac machines. However, when running pjsip's iOS sample app on Mac, there will be no sound output. This is because on Mac, Voice Processing IO audio unit requires float audio samples. This PR will add the check to detect if we are running iOS apps on Mac.


trengginas (approved):

pjsip-apps/src/pjsua/ios/ipjsua/ipjsuaAppDelegate.m
```diff
@@ -132,8 +132,26 @@ static void pjsuaOnAppConfigCb(pjsua_app_config *cfg)
...
+        "--username=404",
+        "--password=pw404",
+        "--registrar=sip:pjsip.org",
+        "--proxy=sip:sip.pjsip.org;transport=tcp",
+        "--stun-srv=stun.pjsip.org",
+        "--use-ice",
+        "--auto-answer=200",
+//           "--video",
+        NULL };
+    int argc = PJ_ARRAY_SIZE(argv) -1;
```
nanangizz, line 154: accidentally included?

nanangizz (approved):

PR #2618, opened 2020-12-31
Fix PCMU/A support in AVI player

nanangizz:
The AVI player tries to support audio with PCMU/PCMA format (as seen in in the code, it finds for raw PCM and PCMU/A formats, the support is also described in the specs/doxygen). However, the check itself does not seem to be correct (should be checking format tag), so PCMU/A does not seem fully supported so far. Moreover, PJSUA only supports AVI file with PCM format.

This PR will fix the PCMU/A support in AVI player:
- fix the AVI header check for PCMU/A
- automatically decode returned audio frames (to 16-bit PCM), so no changes needed in pjsua app to support AVI file with PCMU/A.


ismangil (approved):

sauwming (approved):
okay. let's wait for the reporter's feedback as well.

PR #2619, opened 2021-01-06
Fix ai_socktype to be zero when passed to getaddrinfo

alexkalmuk:
`getaddrinfo()` does not accept a bitwise mask as `ai_socktype`, this field represents a single socket type. If we want to resolve the address for multiple socket types we should pass zero instead.  https://pubs.opengroup.org/onlinepubs/9699919799/functions/getaddrinfo.html

`(pj_SOCK_DGRAM() | pj_SOCK_STREAM())` is equal to `SOCK_RAW=3` in Linux, so that's why it was working.

### DIFF
--- a/pjlib/src/pj/addr_resolv_sock.c
+++ b/pjlib/src/pj/addr_resolv_sock.c
@@ -178,7 +178,9 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
     /* Call getaddrinfo() */
     pj_bzero(&hint, sizeof(hint));
     hint.ai_family = af;
-    hint.ai_socktype = pj_SOCK_DGRAM() | pj_SOCK_STREAM();
+    /* Zero value of ai_socktype means the implementation shall attempt
+     * to resolve the service name for all supported socket types */
+    hint.ai_socktype = 0;
 
     rc = getaddrinfo(nodecopy, NULL, &hint, &res);
     if (rc != 0)
@@ -191,6 +193,11 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
 	/* Ignore unwanted address families */
 	if (af!=PJ_AF_UNSPEC && res->ai_family != af)
 	    continue;
+
+	if (res->ai_socktype != pj_SOCK_DGRAM()
+                    && res->ai_socktype != pj_SOCK_STREAM()) {
+	        continue;
+	}
 
 	/* Store canonical name (possibly truncating the name) */
 	if (res->ai_canonname) {
--- a/pjlib/src/pj/addr_resolv_sock.c
+++ b/pjlib/src/pj/addr_resolv_sock.c
@@ -190,6 +190,9 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
 
     /* Enumerate each item in the result */
     for (i=0; i<*count && res; res=res->ai_next) {
+	unsigned j;
+	pj_bool_t duplicate_found = PJ_FALSE;
+
 	/* Ignore unwanted address families */
 	if (af!=PJ_AF_UNSPEC && res->ai_family != af)
 	    continue;
@@ -197,6 +200,18 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
 	if (res->ai_socktype != pj_SOCK_DGRAM()
                     && res->ai_socktype != pj_SOCK_STREAM()) {
 	        continue;
+	}
+
+	/* Add current address in the resulting list if there
+	 * is no duplicates only. */
+	for (j = 0; j < i; j++) {
+	    if (!pj_memcmp(&ai[j].ai_addr, res->ai_addr, res->ai_addrlen)) {
+		duplicate_found = PJ_TRUE;
+		break;
+	    }
+	}
+	if (duplicate_found) {
+	    continue;
 	}
 
 	/* Store canonical name (possibly truncating the name) */


sauwming:
I can't find any reference that specifically says `getaddrinfo` doesn't accept bitwise `ai_socktype`, but then I can't find any reference which says that it does so as well (only `ai_flags` is explicitly mentioned can be bitwise).

I'm wondering why on every platform other than Linux, this has also worked so far.

nanangizz:
Checking some source codes of `socket.h`, e.g: [here](https://github.com/openbsd/src/blob/master/sys/sys/socket.h#L60), the socket type/`SOCK_*` values do not seem bitwise-able indeed, so I don't think `ai_socktype` accepts a bitwise-mask/multiple socket types.

Also I noticed a couple things (not directly related to the PR):
- As recommended by the [getaddrinfo docs](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getaddrinfo.html) in `APPLICATION USAGE` section, `hint` initialization should not use `memset()` for better portability (the code uses `bzero()`, but I assume they are similar).
- Redundancies returned by `pj_getaddrinfo()`? The function does not accept socket/protocol type filters in input and `pj_addrinfo` output does not have socket/protocol type info, so looks like the returned server addresses will always contain redundancies (which may lead to inefficiencies, e.g: connection retry will try to contact same servers).

alexkalmuk:
> I'm wondering why on every platform other than Linux, this has also worked so far.

I think it's because `SOCK_STREAM=1`, `SOCK_DGRAM=2`, and `SOCK_RAW=3` almost everywhere. At least, I see these values in [glibc](https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/bits/socket_type.h.html#__socket_type) and [newlib](https://chromium.googlesource.com/native_client/nacl-newlib/+/master/newlib/libc/sys/linux/sys/socket.h#70). But I cannot say about Symbian:

```cpp
const pj_uint16_t PJ_SOCK_STREAM= KSockStream;
const pj_uint16_t PJ_SOCK_DGRAM = KSockDatagram;
const pj_uint16_t PJ_SOCK_RAW   = 0xFFFF;
```

So, it this case `(PJ_SOCK_STREAM | PJ_SOCK_DGRAM) != PJ_SOCK_RAW` probably.

> Redundancies returned by pj_getaddrinfo()? The function does not accept socket/protocol type filters in input and pj_addrinfo output does not have socket/protocol type info, so looks like the returned server addresses will always contain redundancies (which may lead to inefficiencies, e.g: connection retry will try to contact same servers).

It looks like adding `pj_SOCK_DGRAM() | pj_SOCK_STREAM()` was an attempt to reduce the amount of information returned. It worked because in such a case `getaddrinfo()` works only with `SOCK_RAW` inside and returns only one socket per server address. I think we can do this "trick" if we know the required socket type at the moment of calling getaddrinfo(). In other cases, if we want multiple socket types to be processed, it's normal to have "duplicates" in the information returned. I see two possible ways here:
1. Specify the required socket type like `hint.ai_socktype = pj_SOCK_DGRAM();` or `hint.ai_socktype = pj_SOCK_STREAM();`. By the way, it is not done because PJSIP can open TCP or UDP socket and we do not know for sure which one when resolving server address?
2. Handle all sockets types with `hint.ai_socktype = 0;`. We will have redundancies in this case, but the caller should know about that.

alexkalmuk:
Alternatively, I could just change `pj_SOCK_DGRAM() | pj_SOCK_STREAM()` to `pj_SOCK_RAW()`, describing why it is used. It will fit POSIX and keep things working at the same time.

nanangizz:
> It looks like adding `pj_SOCK_DGRAM() | pj_SOCK_STREAM()` was an attempt to reduce the amount of information returned. It worked because in such a case `getaddrinfo()` works only with `SOCK_RAW` inside and returns only one socket per server address. I think we can do this "trick" if we know the required socket type at the moment of calling getaddrinfo(). In other cases, if we want multiple socket types to be processed, it's normal to have "duplicates" in the information returned (I mean the same IP address/port for two different socket types). I see two possible ways here:

Tried creating a simple `getaddrinfo()` app on MacOS with `hint.ai_socktype = SOCK_RAW` (which should be similar to `pj_SOCK_DGRAM() | pj_SOCK_STREAM()`), it turned out that there were still duplicates: each IP address+port listed twice, one for each socket type of STREAM & DGRAM (so `SOCK_RAW` resolved to `SOCK_STREAM` & `SOCK_DGRAM`), so the duplicates should have been already there all along (and it seems that the duplicates need to be removed manually, but as mentioned before, this seems to be a different issue, except if you are willing to incorporate it into this PR, of course :).

> 1. Specify the required socket type like `hint.ai_socktype = pj_SOCK_DGRAM();` or `hint.ai_socktype = pj_SOCK_STREAM();`. By the way, it is not done because PJSIP can open TCP or UDP socket and we do not know for sure which one when resolving server address?

Yes, sometimes `pj_getaddrinfo()` is used in the place of DNS SRV resolution (i.e: when DNS server is not specified).

> 2. Handle all sockets types with `hint.ai_socktype = 0;`. We will have redundancies in this case, but the caller should know about that.

Like what the current patch does, right? Well, as I believe `hint.ai_socktype` should not accept a bitwise mask, IMO this should be the correct approach (the closest to the original code objective).

> Alternatively, I could just change `pj_SOCK_DGRAM() | pj_SOCK_STREAM()` to `pj_SOCK_RAW()`, describing why it is used. It will fit POSIX and keep things working at the same time.

Not sure if `SOCK_RAW` can resolve to `SOCK_DGRAM` & `SOCK_STREAM` only, I guess the previous approach above should be cleaner.

alexkalmuk:
> Tried creating a simple getaddrinfo() app on MacOS with hint.ai_socktype = SOCK_RAW (which should be similar to pj_SOCK_DGRAM() | pj_SOCK_STREAM()), it turned out that there were still duplicates: each IP address+port listed twice, one for each socket type of STREAM & DGRAM (so SOCK_RAW resolved to SOCK_STREAM & SOCK_DGRAM),

Nice, thank you. btw, I tried the same under Linux, and addresses were resolving to SOCK_RAW only (without duplicates). But anyway, this approach doesn't seem to be portable enough.

> and it seems that the duplicates need to be removed manually, but as mentioned before, this seems to be a different issue, except if you are willing to incorporate it into this PR, of course :)

Yes, I agree, it looks like a slightly different issue. Maybe I would try a bit later. :)

> Not sure if SOCK_RAW can resolve to SOCK_DGRAM & SOCK_STREAM only, I guess the previous approach above should be cleaner.

I think so.

nanangizz:
> btw, I tried the same under Linux, and addresses were resolving to SOCK_RAW only (without duplicates). But anyway, this approach doesn't seem to be portable enough.

Good to know it, so there is no duplicate on Linux in the current code (`hint.ai_socktype=3`) and after changed to `hint.ai_socktype=0`, there will be duplicate. Looks like the need of duplication removal is increasing now :)

alexkalmuk:
> Looks like the need of duplication removal is increasing now :)

Then I will try to remove them and update the current PR.

alexkalmuk:
@nanangizz I added straightforward filtering of duplicate addresses to `pj_getaddrinfo()`. Could you take a look, please?

nanangizz:
> @nanangizz I added straightforward filtering of duplicate addresses to `pj_getaddrinfo()`. Could you take a look, please?

It looks good to me, thanks.

nanangizz (approved):

sauwming (approved):

PR #2630, opened 2021-01-25
Add --disable-pjsua2 option for targets which lacks C++

alexkalmuk:
Optionally disable C++ part (pjsua2 libs and pjsua2_demo) with the following option:

```
./configure --disable-pjsua2
```

pjsua2 is enabled by default of course.

Possible use case: Currently we are running PJSIP on STM32 MCUs with 320 - 512 Kbytes RAM, so it's a resource-constrained device. If one doesn't need the C++ part of PJSIP, it's better to disable it explicitly. Another point is that pjsua2 and other PJSIP's code is very well separated already.

What do you think?


sauwming:
The idea sounds fine to me.
As for the patch, is it missing` aconfigure.ac`?

alexkalmuk:
@sauwming Thanks. I pushed `aconfigure.ac` (force-pushed), but removed updated `aconfigure` from the changes because I have a difference between the original `aconfigure` and the generated one even in the master branch. When I do

```
$ autoconf --version
autoconf (GNU Autoconf) 2.69
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+/Autoconf: GNU GPL version 3 or later
<http://gnu.org/licenses/gpl.html>, <http://gnu.org/licenses/exceptions.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by David J. MacKenzie and Akim Demaille.

$ autoconf aconfigure.ac > aconfigure
```

I get the following `diff`:
```diff
diff --git a/aconfigure b/aconfigure
index 7070292fb..094610136 100755
--- a/aconfigure
+++ b/aconfigure
@@ -766,6 +766,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -893,6 +894,7 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
+runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE_TARNAME}'
@@ -1145,6 +1147,15 @@ do
   | -silent | --silent | --silen | --sile | --sil)
     silent=yes ;;
 
+  -runstatedir | --runstatedir | --runstatedi | --runstated \
+  | --runstate | --runstat | --runsta | --runst | --runs \
+  | --run | --ru | --r)
+    ac_prev=runstatedir ;;
+  -runstatedir=* | --runstatedir=* | --runstatedi=* | --runstated=* \
+  | --runstate=* | --runstat=* | --runsta=* | --runst=* | --runs=* \
+  | --run=* | --ru=* | --r=*)
+    runstatedir=$ac_optarg ;;
+
   -sbindir | --sbindir | --sbindi | --sbind | --sbin | --sbi | --sb)
     ac_prev=sbindir ;;
   -sbindir=* | --sbindir=* | --sbindi=* | --sbind=* | --sbin=* \
@@ -1282,7 +1293,7 @@ fi
```

I see that my Autoconf version is 2.69, so it's exactly the same as the original one. Probably, Autoconfs under Ubuntu and OSX differ, or I didn't provide some options.

alexkalmuk:
I also pushed a small change https://github.com/pjsip/pjproject/pull/2630/commits/0187fa1ccacf29b2da1369b1c1603ba5fc98e918 that fix `make uninstall` in the master branch.

sauwming (approved):
Looks okay to me.
An internal note: if approved, we need to add the file `aconfigure` before merging it.

nanangizz (approved):

PR #2633, opened 2021-02-03
Remove ioqueue key from set when calling pj_ioqueue_post_completion()

trengginas:
The call to `pj_ioqueue_post_completion()` will find and remove the operation from the ioqueue operation list.
However, it wouldn't clear the key from the ioqueue descriptor set.
This might lead to `pj_ioqueue_poll()` to busy loop, and spike the CPU usage.

This is printed on the trace log, which indicates that there are still event's based on descriptor set, but no pending operation is found.
```
poll: count=2 events=0 processed=0
poll: count=2 events=0 processed=0
```

### DIFF
--- a/pjlib/src/pj/ioqueue_common_abs.c
+++ b/pjlib/src/pj/ioqueue_common_abs.c
@@ -1326,6 +1326,13 @@ PJ_DEF(pj_status_t) pj_ioqueue_post_completion( pj_ioqueue_key_t *key,
         op_rec = op_rec->next;
     }
 
+    /* Clear connecting operation. */
+    if (key->connecting) {
+        key->connecting = 0;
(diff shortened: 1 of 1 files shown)


sauwming:
How about exception event?
Some CI tests are failing, doesn't seem related though.

trengginas:
> How about exception event?
> Some CI tests are failing, doesn't seem related though.

One time EXCEPTION_EVENT added to FD set is on pj_ioqueue_connect() and it didn't add anything to the operation list.

mschlapfer:
When/how does fd for in exception set ever get removed after connect?

nanangizz (approved):

sauwming (approved):

PR #2635, opened 2021-02-04
Use CoreAudio for Mac for macCatalyst

weissinet:
Recently a pull Request was accepted, that enabled Audio for iOS App on Mac

https://github.com/pjsip/pjproject/pull/2612

There is a second Option where the CoreAudio Setup is needed for Apps Running the Mac Catalyst Option.

### DIFF
--- a/pjmedia/src/pjmedia-audiodev/coreaudio_dev.m
+++ b/pjmedia/src/pjmedia-audiodev/coreaudio_dev.m
@@ -1475,18 +1475,25 @@ static pj_status_t create_audio_unit(AudioComponent io_comp,
     if (dir & PJMEDIA_DIR_PLAYBACK) {
 	AURenderCallbackStruct output_cb;
 	AudioStreamBasicDescription streamFormat = strm->streamFormat;
-	BOOL isiOSAppOnMac = false;
+	BOOL isiOSAppOnMacOrMacCatalyst = false;
 
 #ifdef __IPHONE_14_0
 	if (@available(iOS 14.0, *)) {
-    	    isiOSAppOnMac = [NSProcessInfo processInfo].isiOSAppOnMac;
+        if ([NSProcessInfo processInfo].isiOSAppOnMac) {
+            isiOSAppOnMacOrMacCatalyst = true;
+        }
 	}
+    if (@available(iOS 13.0, *)) {
+            if ([NSProcessInfo processInfo].isMacCatalystApp) {
+                isiOSAppOnMacOrMacCatalyst = true;
+            }
+    }
 #endif
 
 	/* Set the stream format */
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia-audiodev/coreaudio_dev.m
```diff
@@ -1475,18 +1475,25 @@ static pj_status_t create_audio_unit(AudioComponent io_comp,
...
+	BOOL isiOSAppOnMacOrMacCatalyst = false;
 
 #ifdef __IPHONE_14_0
 	if (@available(iOS 14.0, *)) {
-    	    isiOSAppOnMac = [NSProcessInfo processInfo].isiOSAppOnMac;
+        if ([NSProcessInfo processInfo].isiOSAppOnMac) {
+            isiOSAppOnMacOrMacCatalyst = true;
+        }
 	}
+    if (@available(iOS 13.0, *)) {
```
sauwming, line 1486: since iOS 13 < 14, this code will always override the above. so shouldn't this be "else if"?

weissinet: Made some additional Research:

isiOSAppOnMac is true for an App build for iOS running on an arm64 mac (Flag available since iOS 13)
isMacCatalystApp is true for an App build as Mac Catalyst or an iOS App running an an arm64 mac (Flag available since iOS 13) 
"To distinguish between an iOS app running on Apple silicon and a Mac app built with Mac Catalyst, use the isiOSAppOnMac property."

I Think we do not need to distinguish so I will remove the "isiOSAppOnMac"

sauwming: Okay, but please add some comment in the code since it's not obvious that isMacCatalystApp also applies for iOS apps.

And just a minor correction: isiOSAppOnMac is available on iOS 14+
https://developer.apple.com/documentation/foundation/processinfo/3608556-isiosapponmac

sauwming: I don't know why there needs to be a specific reference to UIKit since this detection also applies for iOS/iPaad apps in general, such as the ones using SwiftUI. So I just pushed a minor correction of the doc.

sauwming (approved):
I tested it here with this new flag and it seems to work.

You need to sign the CLA though before we can merge this PR.

weissinet:
Hi,

i signed it, but the commits have no E-Mail address, 
How can we fix this?

nanangizz (approved):

sauwming:
> i signed it, but the commits have no E-Mail address,
> How can we fix this?

We'll handle this on our end. Thanks for the patch.

PR #2638, opened 2021-02-10
Use UTC time in PJSIP presence PIDF timestamp

nanangizz:
The timestamp value in PJSIP presence PIDF is actually in local time instead of UTC although it has "Z" suffix. This PR will fix the timestamp value to be in UTC.

Thanks to Christian Becker for the report.

### DIFF
--- a/pjsip/src/pjsip-simple/presence_body.c
+++ b/pjsip/src/pjsip-simple/presence_body.c
@@ -117,9 +117,18 @@ PJ_DEF(pj_status_t) pjsip_pres_create_pidf( pj_pool_t *pool,
 	  int tslen = 0;
 	  pj_time_val tv;
 	  pj_parsed_time pt;
+	  pj_status_t s;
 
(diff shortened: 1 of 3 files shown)


pjsip/src/pjsip-simple/presence_body.c
```diff
@@ -119,7 +119,7 @@ PJ_DEF(pj_status_t) pjsip_pres_create_pidf( pj_pool_t *pool,
 	  pj_parsed_time pt;
 
 	  pj_gettimeofday(&tv);
-	  /* TODO: convert time to GMT! (unsupported by pjlib) */
+	  pj_time_local_to_gmt(&tv);
```
sauwming, line 122: minor: probably add a log or a doc either here or in os_time implementation, that this can be unsupported?

sauwming (approved):

trengginas (approved):

PR #2643, opened 2021-02-16
Add Embox RTOS to supported targets in README.txt

alexkalmuk:
Hello,

We finally finished porting PJSIP (v2.10) to [Embox RTOS ](https://github.com/embox/embox)and verified it with `simple_pjsua` for STM32.

Here is a short [video](https://youtu.be/fJ-HNopjQkM) where we show a simple SIP Phone on STM32. Here is an [article](https://alexkalmuk.medium.com/sip-phone-with-gui-on-stm32-1c3b4abf7ed4) describing this in more detail. Also, we added a description of how to use PJSIP with Embox to our [wiki](https://github.com/embox/embox/wiki/PJSIP).

We would like to mention this in your README.txt if you like this use case.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -404,6 +404,7 @@ Creating config_site.h file
      * mingw (i386),
      * FreeBSD and maybe other BSD's (i386, Opteron, etc.),
      * RTEMS with cross compilation (ARM, powerpc),
+     * Embox RTOS (i386, ARM including STM32),
      * etc.
 
 


sauwming:
Great job for the successful porting and thank you very much for providing the detailed steps by steps. Just wondering whether you manage to run the unit tests (pjlib-test, pjsip-test, etc) and if they are all successful.

alexkalmuk:
@sauwming Good idea to try the tests! I didn't launch them either on qemu/i386 or stm32 yet. Only standard apps/demos. 
I am not sure that running all these tests is possible under stm32 (maybe a limited subset only), but I hope they should run under qemu/i386. Would it be nice if we run the tests and report here in the pull request about their success/failure?

sauwming:
Oh yes, please. Feel free to post the result here. In particular, the PJLIB test should tell us which parts of PJSIP is supported:
https://www.pjsip.org/porting.htm#mozTocId50420

alexkalmuk:
Hi,

We added `pjlib-test` and `pjsip-test` to GH Actions CI. Both tests passed well, we only added a small patch for the pjsip-test because sprintf doesn't support non-ascii symbols (https://github.com/embox/embox/issues/2376). You can see the results [here](https://github.com/embox/embox/runs/2111631268).

nanangizz (approved):

sauwming (approved):

trengginas (approved):

Issue #2620, opened 2021-01-09
Answering re-invite sends a modified SDP file without incrementing version


zecke:
In Asterisk we see the following issue: https://issues.asterisk.org/jira/browse/ASTERISK-28452. It seems that when the remote answer is processed the active_sdp might not be the same as the one offered. The next time it is used (e.g. in answering a re-INVITE) it will be sent without having increased the version number.

How should this work and what role does asterisk have? A work-around seems to increment the version number on active after successfully processing the answer.

https://github.com/pjsip/pjproject/blob/3e8211db8b817e2dc2a9ef03701f40d9309b0589/pjmedia/src/pjmedia/sdp_neg.c#L1564

silentindark:
Hello, @sauwming . Could you please help with this topic?

sauwming:
In the issue page, I believe Ivan has explained it well. Since app/user modifies the local SDP, I think it should be okay to also increase the version. Or is there any reason not to?

Note that I tested the scenario here with pjsua app and without the SDP modification, everything seems to work correctly.
Attached is the sipp scenario I used for testing:
[uac-reinv-no-sdp.xml.txt](https://github.com/pjsip/pjproject/files/5993427/uac-reinv-no-sdp.xml.txt)

jcolp:
There are actually two cases that I'm aware of. The case mentioned by Ivan is certainly one under the control of Asterisk. There is a second one, however:

In a scenario where PJSIP sends an initial INVITE with a set of codecs (ulaw and alaw) and the remote side answers with a single codec (ulaw) the SDP negotiator will change the active local SDP to include only that single codec (ulaw). This is done without incrementing the session version number.

If a user of PJSIP INVITE session does not implement the "on_create_offer" callback then when a re-INVITE without SDP is received the active local SDP is used, without the session version number incremented.

This doesn't impact pjsua as it appears to implement the on_create_offer callback which produces the SDP instead of using the active local SDP. I'm also implementing this callback in Asterisk which should fix both Ivan's case as well as the scenario I've just described.

It does feel as though there is a bug though if the on_create_offer callback is not implemented. Perhaps in INVITE session if it uses the active local SDP in that case it should increment the session version number.

sauwming:
Other than the patch in #2646, let us know if there's anything else that needs to be fixed (or changed) on PJSIP's side in relation to this issue.

jcolp:
[...]

PR #2646, opened 2021-02-18
Increase SDP version when receiving re-INVITE without SDP

sauwming:
Related to #2620 

In a scenario described by @jcolp, if app doesn't implement sip_inv's `on_create_offer()` optional callback:
* PJSIP sends an initial INVITE offer
* remote answers (the answer is different than the offer, for example, it will choose only one out of all the offered codecs)
* PJSIP receives re-INVITE without SDP
* Since `on_create_offer()` is not implemented, PJSIP will answer with the active local SDP, which is different than its initial offer, but without the session version number incremented.

### DIFF
--- a/pjmedia/src/pjmedia/sdp_neg.c
+++ b/pjmedia/src/pjmedia/sdp_neg.c
@@ -449,16 +449,16 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_send_local_offer( pj_pool_t *pool,
 						       neg->active_local_sdp);
 
 #if PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION
-    	if (pjmedia_sdp_session_cmp(neg->active_local_sdp, 
+    	if (pjmedia_sdp_session_cmp(neg->neg_local_sdp, 
     				    neg->initial_sdp, 0) != PJ_SUCCESS)
     	{
-	    neg->active_local_sdp->origin.version++;
+	    neg->neg_local_sdp->origin.version++;
     	}    
 #else
-    	neg->active_local_sdp->origin.version++;
+    	neg->neg_local_sdp->origin.version++;
 #endif
 
-	*offer = neg->active_local_sdp;
+	*offer = neg->neg_local_sdp;
 
     } else {
 	/* We assume that we're in STATE_LOCAL_OFFER.
--- a/pjmedia/src/pjmedia/sdp_neg.c
+++ b/pjmedia/src/pjmedia/sdp_neg.c
@@ -447,6 +447,17 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_send_local_offer( pj_pool_t *pool,
 	neg->state = PJMEDIA_SDP_NEG_STATE_LOCAL_OFFER;
 	neg->neg_local_sdp = pjmedia_sdp_session_clone(pool, 
 						       neg->active_local_sdp);
+
+#if PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -5154,8 +5154,10 @@ static void inv_on_state_confirmed( pjsip_inv_session *inv, pjsip_event *e)
 		    status = pjmedia_sdp_neg_send_local_offer(inv->pool_prov,
 							      inv->neg, 
 							      &active_sdp);
-		    if (status == PJ_SUCCESS)
+		    if (status == PJ_SUCCESS) {
 			sdp = (pjmedia_sdp_session*) active_sdp;
+			sdp->origin.version++;
+		    }
```
nanangizz, line 5160: Shouldn't it be done in `pjmedia_sdp_neg_send_local_offer()`? Most SDP version increases are handled there. Also it should compare active SDP vs initial SDP first if `PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION` is set, similar to `pjmedia_sdp_neg_modify_local_offer2()` that compares new offer vs initial SDP.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -447,6 +447,17 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_send_local_offer( pj_pool_t *pool,
...
+    	if (pjmedia_sdp_session_cmp(neg->active_local_sdp, 
+    				    neg->initial_sdp, 0) != PJ_SUCCESS)
+    	{
+	    neg->active_local_sdp->origin.version++;
+    	}    
+#else
+    	neg->active_local_sdp->origin.version++;
+#endif
+
 	*offer = neg->active_local_sdp;
```
nanangizz, line 461: Hmm.. this is a different but related issue. Just realized that the offer output is the active_local_sdp, while it perhaps should be neg_local_sdp, note that the future sdp nego will only use neg_local_sdp instead of active_local_sdp. So far this works fine because they are identic, but now they are no longer identic as the version is different (so perhaps the future active local SDP, nego result, will have the wrong version? anyway, theoretically returning active_local_sdp here is wrong).

In other hands, not sure if changing the output offer to neg_local_sdp is safe, need some tests (e.g: GitHub CI).

sauwming: Local unit tests are okay here when returning `neg_local_sdp`.

Also, what if here we get rid of the check `PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION` and always compare the SDP? This is because in my understanding, one of the purpose of introducing the setting is for backward compatibility and that's why the default is zero (i.e. previously we always increase the version). But here is the other way around, we currently do not increase the version and the change will do it.

nanangizz: Always increasing the version is safe (standard wise) and faster (for local endpoint), hence we use it as default. But then some apps want to increase only when something really changes (so remote won't need to renegotiate SDP & re-init media due to different SDP version, when actually nothing changes), this is the purpose of the macro.

Here `pjmedia_sdp_neg_send_local_offer()` can always increase the version (without check for changes), it is safe & faster. But then ... as described above. So the macro check purpose still applies here I guess?

jcolp:
I've pulled this into my test build with the original issues I described and confirmed that it is not impacting my fix involving the create_offer callback or any other usage - which makes sense since it is isolated to pjmedia_sdp_neg_send_local_offer. I will still be implementing that callback regardless, since for the unhold case it better fits things.

nanangizz (approved):

trengginas (approved):

PR #2647, opened 2021-02-18
Update pjsua_core.c

SalihPalamut:
limited just 4 calls. Now limited by PJSUA_MAX_CALLS definations.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -2005,10 +2005,9 @@ typedef struct pjsua_config
 
     /** 
      * Maximum calls to support (default: 4). The value specified here
-     * must be smaller than the compile time maximum settings 
-     * PJSUA_MAX_CALLS, which by default is 32. To increase this 
-     * limit, the library must be recompiled with new PJSUA_MAX_CALLS
-     * value.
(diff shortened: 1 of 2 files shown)


sauwming:
Should we?

The original modification, and some feedback, can be found here:
https://github.com/pjsip/pjproject/commit/6627f3e64b479dee8640b8a757b2b7680bb0f5ca

nanangizz:
Okay with me. Few notes:
- The proposed patch is logically wrong when `PJSUA_MAX_CALLS < 4`, perhaps simply `max_calls = PJSUA_MAX_CALLS`
- Change default value of `PJSUA_MAX_CALLS` to 4, because now 32 is a bit too high for most apps I guess and 4 is the current run-time default value.
- While in the area, the `max_calls` docs need to be updated:
   ```
     * must be smaller than the compile time maximum settings 
     * PJSUA_MAX_CALLS
   ```
   should be `smaller than or equal to`?

sauwming:
I took the liberty to modify it according to @nanangizz's suggestion.

Let us know if you have any objection or feedback.

SalihPalamut:
Thanks for the solution.

sauwming:
We actually haven't merged the PR. So if you want to reopen the PR and restore the branch, please do so. If this is not possible, or you'd rather us open a new PR for this, let us know.

PR #2651, opened 2021-02-22
Allow EC to be used when sound device is opened on speaker mode only

trengginas:
Consider the scenario:
On a PTT (push to talk) app, sound device is switched regularly between speaker only and speaker+mic mode.
On iOS this means the audio component will switch between VPIO (speaker+mic) and remote IO (speaker only).
When opening the sound device in speaker only mode, currently ec is disabled and audio component switch is forced.
However when using remote IO, lower volume playback is noticeable.

This ticket will made these changes:
- Don't reset sound device capability and ec option when it is opened on speaker mode. 
- Remove assertion that only allow sound device is opened full duplex when ec is enabled

### DIFF
--- a/pjmedia/include/pjmedia-audiodev/config.h
+++ b/pjmedia/include/pjmedia-audiodev/config.h
@@ -113,6 +113,18 @@ PJ_BEGIN_DECL
 #endif
 
 
+/**
+ * This setting controls coreaudio should always use VPIO.
+ * Audio component might be switched from VPIO if ec is disabled,
+ * which has lower volume output. Use this to force using VPIO
+ * regardless the ec settings.
+ *
+ */
+#ifndef PJMEDIA_AUDIO_DEV_COREAUDIO_ALWAYS_USE_VPIO
+#   define PJMEDIA_AUDIO_DEV_COREAUDIO_ALWAYS_USE_VPIO	0
+#endif
+
+
  /**
   * This setting controls whether WMME support should be included.
   */
--- a/pjmedia/include/pjmedia-audiodev/config.h
+++ b/pjmedia/include/pjmedia-audiodev/config.h
@@ -113,18 +113,6 @@ PJ_BEGIN_DECL
 #endif
 
 
-/**
- * This setting controls coreaudio should always use VPIO.
- * Audio component might be switched from VPIO if ec is disabled,
- * which has lower volume output. Use this to force using VPIO
- * regardless the ec settings.
- *
- */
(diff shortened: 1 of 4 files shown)


sauwming:
Not sure why we need to introduce the compile-time setting.
If there's a problem with switching from VPIO to RemoteIO, then we should try to fix this playback volume issue instead. If it can't be fixed (such as due to a bug from the OS itself), then there's already a workaround of not disabling the EC setting, which means there's no need for another setting, is there?

trengginas:
> Not sure why we need to introduce the compile-time setting.
> If there's a problem with switching from VPIO to RemoteIO, then we should try to fix this playback volume issue instead. If it can't be fixed (such as due to a bug from the OS itself), then there's already a workaround of not disabling the EC setting, which means there's no need for another setting, is there?

Consider on PTT scenario, app will switch between "mic+speaker" and "speaker only" regularly.
On "speaker only" mode, ec is disabled by default which changes the audio component [(ref).](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_aud.c#L1879) and [(ref)](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/sound_port.c#L444).
This will result in a noticeable changes in playback volume when switching modes.

Another alternative probably is to allow `ec_options` to be set on "speaker only" mode.

sauwming:
> Another alternative probably is to allow `ec_options` to be set on "speaker only" mode.

This sounds better, IMO.

trengginas:
> > Another alternative probably is to allow `ec_options` to be set on "speaker only" mode.
> 
> This sounds better, IMO.

It seems that when sound device is opened/started with PJMEDIA_DIR_PLAYBACK only, ec will always be disabled ([ref](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/sound_port.c#L253)) regardless the ec option on sound port param.

sauwming:
> It seems that when sound device is opened/started with PJMEDIA_DIR_PLAYBACK only, ec will always be disabled ([ref](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/sound_port.c#L253)) regardless the ec option on sound port param.

Then this needs to be changed as well. And I believe the current code isn't quite correct since aud dev capability is not limited to capture dev capability only, so by right it shouldn't be set to 0.

nanangizz (approved):

sauwming (approved):

PR #2652, opened 2021-02-23
Add option to disable auto re-registration.

trengginas:
Currently (on `pjsua` API) when registration is successful, the library will send auto registration refresh after a successful registration.
On some cases (e.g: Push Notification use), app might require to disable it to save resources.
This patch will add the option (`PJSUA_REG_AUTO_REG_REFRESH`) to disable auto registration refresh.
If this is disabled, app will still be able to send registration refresh periodically or before making a call manually.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -3357,15 +3357,17 @@ PJ_DECL(pj_status_t) pjsua_transport_lis_start( pjsua_transport_id id,
 #endif
 
 /**
- * Set this to 0 to disable auto re-registration. This is useful for app
- * that uses Push Notification and doesn't require auto re-registration.
- * App can periodically send refresh registration or send it before making
- * a call. See https://github.com/pjsip/pjproject/pull/2652 for more info.
+ * When the registration is successfull, the auto registration refresh will
+ * be sent before it expires. Setting this to 0 will disable it.
+ * This is useful for app that uses Push Notification and doesn't require auto
+ * registration refresh. App can periodically send refresh registration or
+ * send it before making a call.=
+ * See https://github.com/pjsip/pjproject/pull/2652 for more info.
  *
  * Default: 1 (enabled)
  */
-#ifndef PJSUA_REG_AUTO_REREG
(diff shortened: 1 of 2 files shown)


pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -3356,6 +3356,17 @@ PJ_DECL(pj_status_t) pjsua_transport_lis_start( pjsua_transport_id id,
...
 
+/**
+ * Set this to 0 to disable auto re-registration. This is useful for app
+ * that uses Push Notification and doesn't require auto re-registration.
+ * App can periodically send refresh registration or send it before making
+ * a call. See https://github.com/pjsip/pjproject/pull/2652 for more info.
+ *
+ * Default: 1 (enabled)
+ */
+#ifndef PJSUA_REG_AUTO_REG
```
sauwming, line 3367: AUTO_REREG should be more self-explanatory.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -2364,13 +2364,22 @@ static void regc_cb(struct pjsip_regc_cbparam *param)
...
 	    update_service_route(acc, param->rdata);
 
-	    PJ_LOG(3, (THIS_FILE, 
-		       "%s: registration success, status=%d (%.*s), "
-		       "will re-register in %d seconds", 
-		       pjsua_var.acc[acc->index].cfg.id.ptr,
-		       param->code,
-		       (int)param->reason.slen, param->reason.ptr,
-		       param->expiration));
+            if (PJSUA_REG_AUTO_REG) {
```
sauwming, line 2367: since it's a compile-time setting, why is the check done runtime?

nanangizz:
The `auto rereg` term is being used by the reregistration retry after a temporary failure, while this reregistration is for the normal registration refresh before it expires (I think the new macro docs need to clarify this too). Wouldn't it be better to avoid `auto rereg` term to avoid ambiguity, perhaps something like `auto_reg_refresh`?

nanangizz (approved):

sauwming (approved):
Don't forget to change the PR desc as well

PR #2655, opened 2021-02-28
Update alsa_dev.c

SalihPalamut:
Useless resources have been removed. These resources can be used directly by pjsua_set_snd_dev.

### DIFF
--- a/pjmedia/src/pjmedia-audiodev/alsa_dev.c
+++ b/pjmedia/src/pjmedia-audiodev/alsa_dev.c
@@ -400,7 +400,7 @@ static pj_status_t alsa_factory_refresh(pjmedia_aud_dev_factory *f)
     while (*n != NULL) {
 	char *name = snd_device_name_get_hint(*n, "NAME");
 	if (name != NULL) {
-	    if (0 != strcmp("null", name))
+	     if (0 == strncmp("plughw", name,6)||0 == strncmp("default", name,7))
 		add_dev(af, name);
 	    free(name);
 	}


nanangizz:
Could you also describe more about the problem (or enhancement?) the PR trying to fix? The patch seems to filter device name to only `plughw` & `default`, why? Other devices are 'useless'?

SalihPalamut:
Because all inputs are virtual and your maximum device is 32. So we added a new usb sound card, it does not show any external card.
#define MAX_DEVICES			32

nanangizz:
Not very familiar with ALSA, but some hardware device may be named `hw` instead of `plughw`, also some app may need to use virtual ALSA devices too. So I guess the patch can be improved by:
- make the `MAX_DEVICES` configurable, e.g: redefine it to `PJMEDIA_AUD_MAX_DEVS` & make `PJMEDIA_AUD_MAX_DEVS` configurable (just realized that `PJMEDIA_AUD_MAX_DEVS` is not really configurable as it is not protected with `#ifdef`)
- make the filter configurable, instead of hardcoded to `plughw` & `default`.

What do you think?

SalihPalamut:
Maybe I tried changing the MAX_DEVICES size was not good enough, I also tried to open other defined cards and it didn't open correctly.
So we can only show alsamixer based cards or 
alsa
only inputs with input + output greater than 0 can be used,removed (0/0)
auddemo
My solution 
my_sol

nanangizz:
Let me try to rephrase. The patch may introduce problems, e.g: some hardware device may be named `hw` instead of `plughw`, also some app may need to use virtual ALSA devices too. Unless they are addressed, I don't think we could merge this.

SalihPalamut:
Ok, I understand.

Maybe this functionality can be changed or a filter function can be added for more flexibility.

pj_status_t alsa_factory_refresh(pjmedia_aud_dev_factory *f) 
change to
pj_status_t alsa_factory_refresh(pjmedia_aud_dev_factory *f,char *hw_filter)

and 
  if (0 == strcmp(hw_filter, name))

This idea can be developed.

nanangizz:
Yes, the filter should be configurable.

> pj_status_t alsa_factory_refresh(pjmedia_aud_dev_factory *f)
> change to
> pj_status_t alsa_factory_refresh(pjmedia_aud_dev_factory *f,char *hw_filter)

Unfortunately the function signature should match to `pjmedia_aud_dev_factory_op.refresh`, so adding a new param is not allowed there.

An approach is by introducing a new compile-time setting for the filter, e.g:
```
/**
 * Specifies ALSA device names filter. If set, ALSA will only list the devices whose names
 * containing tokens specified in the macro settings, tokens are separated by colon (":"),
 * e.g: "plughw:default:hw".
 * 
 * By default, no filter is applied, so all ALSA devices will be listed. Note that currently
 * ALSA will only list up to 32 devices.
 */
#ifndef PJMEDIA_ALSA_DEV_NAME_FILTER
#  undef PJMEDIA_ALSA_DEV_NAME_FILTER
#endif
```

So, would you be interested in updating the patch?

nanangizz:
I assume this PR will not be updated anytime soon, so will close this PR for now.

PR #2656, opened 2021-03-01
Differentiate video codec ID by assigning each codec backend a specific PT number

trengginas:
Currently the library supports multiple codec backend implementation (e.g: H264 using `OpenH264`, `ffmpeg` or `vid_toolbox`).
When enabling those codec backend, user cannot set their priority and only one video codec will be active.
This is useful when user want prioritize or disable one codec backend. (e.g: on iOS user want to prioritize `vid_toolbox` over `OpenH264`).
Video codec implementation uses PT as part of its ID, so assigning each video codec implementation specific PT will allow user will to choose which codec implementation to use.

### DIFF
--- a/pjmedia/include/pjmedia-codec/types.h
+++ b/pjmedia/include/pjmedia-codec/types.h
@@ -126,7 +126,11 @@ enum pjmedia_video_pt
      PJMEDIA_RTP_PT_H264_RSV4,
 
      PJMEDIA_RTP_PT_VP8,
+     PJMEDIA_RTP_PT_VP8_RSV1,
+     PJMEDIA_RTP_PT_VP8_RSV2,
      PJMEDIA_RTP_PT_VP9,
+     PJMEDIA_RTP_PT_VP9_RSV1,
+     PJMEDIA_RTP_PT_VP9_RSV2,
 
      /* Caution!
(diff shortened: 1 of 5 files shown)


pjmedia/include/pjmedia-codec/types.h
```diff
@@ -126,7 +126,11 @@ enum pjmedia_video_pt
      PJMEDIA_RTP_PT_H264_RSV4,
 
      PJMEDIA_RTP_PT_VP8,
+     PJMEDIA_RTP_PT_VP8_RSV1,
+     PJMEDIA_RTP_PT_VP8_RSV2,
      PJMEDIA_RTP_PT_VP9,
+     PJMEDIA_RTP_PT_VP9_RSV1,
+     PJMEDIA_RTP_PT_VP9_RSV2,
```
nanangizz, line 133: A short comment for each PT (non-doxygen) about the 'owner' of the PT slot perhaps help tracking in the future, e.g:
```PJMEDIA_RTP_PT_H264, /* used by OpenH264 */```

nanangizz (approved):
Just a minor inline note, also don't forget to update the PT & `test_alloc()` checks in android native codec too.

sauwming (approved):

PR #2661, opened 2021-03-05
Fix python print into with bracket

jeesang7:
- For use python sample in also python3, fixed python print code to with bracket `(` `)`.
- It is also effective to reduce warnings at python editor or IDE.
- Reference: https://docs.python.org/3.0/whatsnew/3.0.html#what-s-new-in-python-3-0
- If this PR is OK, I'll try to this work that print with bracket about other python samples(pjsip-apps/src/python/samples).

### DIFF
--- a/pjsip-apps/src/python/samples/simplecall.py
+++ b/pjsip-apps/src/python/samples/simplecall.py
@@ -24,7 +24,7 @@ import pjsua as pj
 
 # Logging callback
 def log_cb(level, str, len):
-    print str,
+    print(str)
 
 # Callback to receive events from Call
 class MyCallCallback(pj.CallCallback):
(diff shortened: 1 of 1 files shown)


sauwming (approved):

nanangizz:
The Python port in `pjsip-apps/src/python` is actually deprecated, please check [here](https://trac.pjsip.org/repos/wiki/Python_SIP_Tutorial) and no longer maintained for quite long time, in fact it misses a lot of new APIs such as video. I am afraid that updating something in there may give an impression that it is still maintained or may attract new users.

sauwming:
Ah, I forgot about that. In that case, then I think we can close this PR.

PR #2663, opened 2021-03-09
Fix secure transport checking

sauwming:
Related to commit https://github.com/pjsip/pjproject/commit/67e46c1ac45ad784db5b9080f5ed8b133c122872 of https://github.com/pjsip/pjproject/security/advisories/GHSA-8hcp-hm38-mfph : fix the check of whether a transport type is secure.

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -2335,7 +2335,10 @@ PJ_DEF(pj_status_t) pjsip_tpmgr_acquire_transport2(pjsip_tpmgr *mgr,
 		    if (!tp_iter->tp->is_shutdown &&
 			!tp_iter->tp->is_destroying)
 		    {
-			if ((type & PJSIP_TRANSPORT_SECURE) && tdata) {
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsip/sip_transport.c
```diff
@@ -2335,7 +2335,10 @@ PJ_DEF(pj_status_t) pjsip_tpmgr_acquire_transport2(pjsip_tpmgr *mgr,
 		    if (!tp_iter->tp->is_shutdown &&
 			!tp_iter->tp->is_destroying)
 		    {
-			if ((type & PJSIP_TRANSPORT_SECURE) && tdata) {
+			if (tdata &&
+			    (pjsip_transport_get_flag_from_type(type) &
+			     PJSIP_TRANSPORT_SECURE))
```
nanangizz, line 2340: minor: a bit optimization, perhaps get the type flag before the iteration?

nanangizz (approved):
just a minor below.

trengginas (approved):

PR #2667, opened 2021-03-12
More works on Trickle ICE

nanangizz:
This PR will address some issues & enhancements in trickle ICE:
- Improve trickling state management (fix no SIP INFO when initial INVITE responded immediately with 200, strayed SIP INFO after trickling is done, etc).
- Fix issues when rtcp-mux is enabled.
- Allow process incoming SIP INFO before receiving remote SDP (as it seems to be allowed by the standard), initially we don't want to process it because remote ICE ufrag is still unknown, now the ufrag from incoming SIP INFO will be used as remote ufrag.
- Use regular ICE on re-INVITE (with reinit media flag).

### DIFF
--- a/pjnath/include/pjnath/config.h
+++ b/pjnath/include/pjnath/config.h
@@ -470,12 +470,16 @@
 
 /**
  * For trickle ICE, this macro specifies the maximum time of waiting for
- * end-of-candidates indication from remote after remote first signal of its
- * intention of using trickle ICE (e.g: via SDP "a=ice-options:trickle"),
- * in seconds.
+ * end-of-candidates indication from remote once ICE connectivity checks
+ * is started, in seconds. When the timer expires, ICE will assume that
+ * end-of-candidates indication is received so any further remote candidate
+ * update will be ignored.
  *
- * When the timer expires, ICE assumes that end-of-candidates indication
- * has been received, and any further remote candidate update will be ignored.
+ * Note that without remote end-of-candidates indication, ICE will not be
+ * able to conclude the ICE negotiation. Also note that the ICE connectivity
+ * checks should only be started after both agents have started trickling
+ * ICE candidates (e.g: both have sent their SDPs, either via normal SDP
+ * offer/answer or SIP INFO).
  *
  * Default: 60 seconds.
  */
--- a/pjnath/include/pjnath/config.h
+++ b/pjnath/include/pjnath/config.h
@@ -476,15 +476,19 @@
  * update will be ignored.
  *
  * Note that without remote end-of-candidates indication, ICE will not be
- * able to conclude the ICE negotiation. Also note that the ICE connectivity
- * checks should only be started after both agents have started trickling
- * ICE candidates (e.g: both have sent their SDPs, either via normal SDP
- * offer/answer or SIP INFO).
+ * able to conclude that the ICE negotiation has failed when all pair checks
+ * are completed but there is no valid pair (on the other hand, the ICE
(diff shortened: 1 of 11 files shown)


pjsip/include/pjsua-lib/pjsua_internal.h
```diff
@@ -118,6 +118,14 @@ typedef struct call_answer
 } call_answer;
 
 
+/* Generic states */
```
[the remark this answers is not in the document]
nanangizz: > I'm not sure why this has to be a generic state while the use is only for trickle ICE.

There seem to be some specific-but-semantically-generic state enums already in the library, e.g: `pjsua_state, pj_ice_sess_checklist_state`, so I was thinking why introduce another specific one instead of a generic one, just in case it can be reusable by other feature in the future.

> And can this be put in pjnath instead so that it can be used by pjnath user?

Actually it is not part of trickle ICE itself, but part of the using protocol (i.e: specific for SIP), other using protocol may not need it.

> 
> After I said this, I suddenly remember, will it be possible to update `icedemo` as well to provide an example for pjnath user who wants to use trickle ICE directly without pjsip/pjsua?
> (if this is quite major, perhaps need to schedule it for the next release instead).

Yes, it is perhaps a major change for icedemo. Actually there is one 'sample' already in pjnath-test, instead of using SDP, it conveys new candidates to other UA via shared global var (btw, haven't checked it with recent modifications), I think it can also be used as sample reference.

pjnath/include/pjnath/config.h
```diff
@@ -468,6 +468,26 @@
...
+ * end-of-candidates indication is received so any further remote candidate
+ * update will be ignored.
+ *
+ * Note that without remote end-of-candidates indication, ICE will not be
+ * able to conclude the ICE negotiation. Also note that the ICE connectivity
+ * checks should only be started after both agents have started trickling
+ * ICE candidates (e.g: both have sent their SDPs, either via normal SDP
+ * offer/answer or SIP INFO).
+ *
+ * Default: 60 seconds.
```
sauwming, line 484: 60 sec seems too long? Is there any recommendation for this?

sauwming (approved):

trengginas (approved):

pjnath/include/pjnath/config.h
nanangizz: I can't find any recommendation for this.

The total timeout (with retransmissions+response timeout) of a STUN tsx is ~15s, the default DNS query timeout is ~10s, so a STUN tsx may take ~25s. A TURN allocation may require few STUN tsx (e.g: retry next server), so I guess 60s is a fair value (about 1 DNS query + 3 retry)?

sauwming: By 30 sec, half of the period, I imagine most users will have already concluded that there is no media flow and hang up the call.
But this is minor only, app can simply change the default timeout as it wishes.

nanangizz: Hmm.. okay, will change to 40 sec (1 DNS query + 2 retry/signalling delay), to give a TURN backup server a better chance, and perhaps to accommodate a few patient users :)

Also updated the docs to clarify about the relation of the end-of-cand indication & ICE negotiation conclusion.

nanangizz: Also note that the timer can be started since receiving 180 response (instead of 200), as SIP INFO (with trickle ICE SDP) traffics should be started after dialog is established (response with To tag).

PR #2670, opened 2021-03-18
Fix installation of libpjsua2 shared libraries

hlobit:
Hi, while trying to make and install from release 2.11, I noticed `libpjsua2` shared libraries were not correctly installed.

Here is the patch on the Makefile that I needed to apply to make it back again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,9 +121,10 @@ cmp_wav:
 
 install:
 	mkdir -p $(DESTDIR)$(libdir)/
-	cp -af $(APP_LIB_FILES) $(DESTDIR)$(libdir)/
 	if [ "$(PJ_EXCLUDE_PJSUA2)x" = "x" ] ; then \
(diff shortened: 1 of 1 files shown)


trengginas:
> Hi, while trying to make and install from release 2.11, I noticed `libpjsua2` shared libraries were not correctly installed.
> 
> Here is the patch on the Makefile that I needed to apply to make it back again.

Thank you for the patch! This is related to #2630 , have you tried `--disable-pjsua2` option ? Making sure that the correct files were copied when the option is used.

hlobit:
> > Hi, while trying to make and install from release 2.11, I noticed `libpjsua2` shared libraries were not correctly installed.
> > Here is the patch on the Makefile that I needed to apply to make it back again.
> 
> Thank you for the patch! This is related to #2630 , have you tried `--disable-pjsua2` option ? Making sure that the correct files were copied when the option is used.

Yes it looks OK to me, when configured with `--disable-pjsua2`, pjsua2 is not built and not installed.

trengginas (approved):

nanangizz (approved):

sauwming (approved):

PR #2671, opened 2021-03-19
Check error returned from decode callback of the decoding process using video toolbox codec

trengginas:
Currently when decode callback returns error, it is not conveyed to the decoding process handle it as success.
This will lead to misinformation on the log:

`Error: not enough buffer for decoded frame ()`

This patch will consider the error from decode callback and so that it will be treated as decode error.

### DIFF
--- a/pjmedia/src/pjmedia-codec/vid_toolbox.m
+++ b/pjmedia/src/pjmedia-codec/vid_toolbox.m
@@ -169,6 +169,7 @@ typedef struct vtool_codec_data
     pj_uint8_t			*dec_buf;
     unsigned			 dec_buf_size;
     CMFormatDescriptionRef	 dec_format;
+    OSStatus             dec_status;
 
     unsigned			 dec_sps_size;
     unsigned			 dec_pps_size;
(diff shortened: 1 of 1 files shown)


sauwming:
Since `VTDecompressionSessionDecodeFrame` is called synchronously, the callback has been called before the function returns. So I'm wondering what error it is that occurs only in the callback but not conveyed to the return status of the decode API?

trengginas:
> Since `VTDecompressionSessionDecodeFrame` is called synchronously, the callback has been called before the function returns. So I'm wondering what error it is that occurs only in the callback but not conveyed to the return status of the decode API?

Testing with remote using OpenH264, status `kVTVideoDecoderBadDataErr` (-12909) was returned.

sauwming:
> Testing with remote using OpenH264, status `kVTVideoDecoderBadDataErr` (-12909) was returned.

I see. And `VTDecompressionSessionDecodeFrame` returned `noErr`?

sauwming (approved):

nanangizz (approved):

PR #2678, opened 2021-03-25
Allow change Capture Device to *any* valid one

trandi:
call_change_cap_dev() should allow *any* valid capture device, ie. including PJMEDIA_DIR_CAPTURE_PLAYBACK for example


nanangizz (approved):
Currently there is no video device that can do both capture & playback (in a single device ID), but perhaps there will be in the future (e.g: bidirectional streaming video device).

sauwming (approved):

trandi:
> Currently there is no video device that can do both capture & playback (in a single device ID), but perhaps there will be in the future (e.g: bidirectional streaming video device).

Thank you @nanangizz for the comment.
I found this problem indeed while using a custom video device.

PR #2680, opened 2021-03-29
Add missing route-set updates

hlobit:
Hi,

While testing with outgoing proxy in PJSUA2, I noticed 2 places where the route-set is not updated before SIP requests.

Here is a patch I use to overcome this.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_pres.c
+++ b/pjsip/src/pjsua-lib/pjsua_pres.c
@@ -1950,8 +1950,13 @@ static void unsubscribe_buddy_presence(pjsua_buddy_id buddy_id)
     pjsua_buddy *buddy;
     pjsip_tx_data *tdata;
     pj_status_t status;
+    int acc_id;
+    pjsua_acc *acc;
 
     buddy = &pjsua_var.buddy[buddy_id];
+    acc_id = pjsua_acc_find_for_outgoing(&buddy->uri);
+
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsua-lib/pjsua_pres.c
```diff
@@ -1964,6 +1969,11 @@ static void unsubscribe_buddy_presence(pjsua_buddy_id buddy_id)
     PJ_LOG(5,(THIS_FILE, "Buddy %d: unsubscribing..", buddy_id));
     pj_log_push_indent();
 
+    /* Set route-set */
+    if (!pj_list_empty(&acc->route_set)) {
+	pjsip_dlg_set_route_set(buddy->dlg, &acc->route_set);
+    }
```
sauwming, line 1975: Shouldn't this already be set when subscribing the presence?

hlobit: If the `buddy->dlg` is the same and if route set is already set, yes it should be.
I just checked again, and the `Route` header is not set in the un`SUBSCRIBE` request.

Maybe this is not the proper way to fix it : I can investigate when I have more time if the dialog is the same, and if it is, why the route set is unset.

sauwming:
Closing this due to lack of follow-up. Please reopen if there is any update in your investigation. Thanks.

PR #2683, opened 2021-03-31
Allow SIP transaction to continue for certain case despite transport error

sauwming:
Normally, SIP transaction will always terminate upon transport error. But it may be desirable for certain cases (such as for INVITE UAS tsx that hasn't been confirmed) that the transaction be allowed to continue. So we introduce a compile-time setting for this.

### DIFF
--- a/pjsip/include/pjsip/sip_config.h
+++ b/pjsip/include/pjsip/sip_config.h
@@ -1045,6 +1045,18 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 #   define PJSIP_TSX_1XX_RETRANS_DELAY	60
 #endif
 
+/**
+ * Setting to determine if SIP transaction will always terminate upon
+ * transport error. If disabled, then there are certain cases (such as
+ * for INVITE UAS tsx that hasn't been confirmed) when the transaction
+ * can be allowed to continue.
+ *
+ * Default: 1 (transaction always terminated)
+ */
+#ifndef PJSIP_TSX_ALWAYS_TERMINATE_ON_TP_ERROR
+#   define PJSIP_TSX_ALWAYS_TERMINATE_ON_TP_ERROR 1
+#endif
+
 #define PJSIP_MAX_TSX_KEY_LEN		(PJSIP_MAX_URL_SIZE*2)
 
 /* User agent. */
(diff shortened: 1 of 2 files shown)


sauwming:
This patch is from an old CR http://cr.teluu.com/231677003/ (this CR link can only be accessed by internal developers only).

I can't remember, and can't find any information from our internal chat, if this is overlooked or if there's any other reason why it hasn't been integrated.

nanangizz:
Perhaps retrying is actually the correct behavior, or at least a better behavior as I am not sure if the retry procedure is inline with the standard. So what if we modify the behavior (break backward compatibility), and the macro is provided for app wanting the old behavior. Note: the macro name is perhaps too generic, while it is only applicable for UAS tsx.

In https://tools.ietf.org/html/rfc3261#section-17.2.4:
```
   When the server transaction sends a response to the transport layer
   to be sent, the following procedures are followed if the transport
   layer indicates a failure.

   First, the procedures in [4] are followed, which attempt to deliver
   the response to a backup.
```
Then in https://tools.ietf.org/html/rfc3263#section-5 (previously referred as "procedures in [4]"):
```
   A server, according to RFC 3261 [1], will send a response on the
   connection it arrived on (in the case of reliable transport
   protocols), and for unreliable transport protocols, to the source
   address of the request, and the port in the Via header field.  The
   procedures here are invoked when a server attempts to send to that
   location and that response fails (the specific conditions are
   detailed in RFC 3261). "Fails" is defined as any closure of the
   transport connection the request came in on before the response can
   be sent, or communication of a fatal error from the transport layer.

   In these cases, the server examines the value of the sent-by
   construction in the topmost Via header.  If it contains a numeric IP
   address, the server attempts to send the response to that address,
   using the transport protocol from the Via header, and the port from
   sent-by, if present, else the default for that transport protocol.
```

nanangizz (approved):

trengginas (approved):

nanangizz:
Perhaps need to note the behavior change (potential backward compatibility issue) in the milestone 2.12 page?

sauwming:
Sure. Note that Github milestone page is simple, so we can't make something like https://trac.pjsip.org/repos/milestone/release-2.10

PR #2689, opened 2021-04-12
Update pjsua.py hdr_list documentation

feczo:
The current documentation is misleading eg. make_call(self, dst_uri, cb=None, hdr_list=None)
hdr_list -- Optional list of headers to be sent with outgoing INVITE

Based on this one would think what need to be passed on is ["Privacy: id", "My-other-header-key: value"]

However once one reads the C library code its apparent that a list of tuples is expected:

https://github.com/pjsip/pjproject/blob/ea7105c222d657d3c1f7fb18ff9130450e2e40e6/pjsip-apps/src/py_pjsua/py_pjsua.c#L469

PyObject * tuple = PyList_GetItem(py_hdr_list, i);

so the hdr_list should be  [("Privacy","id"), ("My-other-header-key", "value")]

This caused me a day worth of head ache and Im sure it would for others too.

### DIFF
--- a/pjsip-apps/src/python/pjsua.py
+++ b/pjsip-apps/src/python/pjsua.py
@@ -1271,8 +1271,9 @@ class Account:
                     created Call object. If this CallCallback is not
                     specified (i.e. None is given), it must be installed
                     later using call.set_callback().
-        hdr_list -- Optional list of headers to be sent with outgoing
-                    INVITE
+        hdr_list -- Optional list of header (key, value) tuples
+                    to be sent with outgoing INVITE
+
 
         Return:
             Call instance.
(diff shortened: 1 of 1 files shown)


nanangizz:
Sorry for the troubles and thank you for the patch. But I am afraid that we won't be able to apply any update on this module. 

In case you haven't noticed, the PJSUA Python module is actually deprecated (see [here](https://trac.pjsip.org/repos/wiki/Python_SIP_Tutorial)) and may be removed in the future release. As described in the link, it is replaced by [PJSUA2 Python via SWIG](https://www.pjsip.org/docs/book-latest/html/intro_pjsua2.html#building-python-and-java-swig-modules).

PR #2703, opened 2021-04-22
Fixed remote address change checking in DTLS transport

sauwming:
Currently the remote address change check in DTLS transport doesn't take into account rtcp multiplexing. Therefore, it will incorrectly think that the remote rtcp address has changed and try to renegotiate DTLS key.

This can cause no media when both DTLS and rtcp multiplexing are enabled due to DTLS key negotiation never completed. One can know this from the log which will output:
```Error sending RTP: SRTP parameters negotiation still in progress (PJMEDIA_SRTP_EKEYNOTREADY)```
even after some time has passed.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -117,7 +117,6 @@ typedef struct dtls_srtp
     pj_bool_t		 got_keys;	    /* DTLS nego done & keys ready  */
     pjmedia_srtp_crypto	 tx_crypto;
     pjmedia_srtp_crypto	 rx_crypto;
-    pj_bool_t		 use_rtcp_mux;
 
     char		 buf[PJMEDIA_MAX_MTU];
     pjmedia_clock	*clock;		    /* Timer workaround for retrans */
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by ismangil — CHANGES_REQUESTED** [review]
The DTLS debug flag defaults to 1 in this code. It needs to be kept at 0.

File: pjmedia/src/pjmedia/transport_srtp_dtls.c [hunk 1]
```diff
@@ -949,6 +950,12 @@ static pj_status_t get_rem_addrs(dtls_srtp *ds,
 			  pj_sockaddr_get_addr_len(rem_rtp));
 	    }
 	}
+    } else {
+    	a = pjmedia_sdp_attr_find2(m_rem->attr_count, m_rem->attr,
+			       	   "rtcp-mux", NULL);
+	if (a) {
+	    pj_sockaddr_cp(rem_rtcp, rem_rtp);
+	}
```
nanangizz, line 958: What if remote SDP has both attributes, rtcp-mux & rtcp? Shouldn't it check for rtcp-mux first?

nanangizz (approved):

trengginas (approved):

PR #2705, opened 2021-04-30
Specify media direction when making, answering, or updating call

### LINKED ISSUE #2246: Specify audio/video direction when making/answering call

sauwming:
To fix #2246

In this PR, we add the flag `PJSUA_CALL_SET_MEDIA_DIR` and `media_dir[]` as part of `pjsua_call_setting` to specify the media direction. The setting will be used when initializing or updating media channel, so it's also applicable for sending/receiving re-INVITE/UPDATE.

Spec:
```
    /**
     * Media direction. This setting will only be used if the flag
     * PJSUA_CALL_SET_MEDIA_DIR is set, and it will persist for subsequent
     * offers or answers. 
     * For example, a media that is set as PJMEDIA_DIR_ENCODING can only
     * mark the stream in the SDP as sendonly or inactive, but will not
     * become sendrecv in subsequent offers and answers.
     * Application can update the media direction in any API or callback
     * that accepts pjsua_call_setting as a parameter, such as via
     * pjsua_call_reinvite/update() or in on_call_rx_offer/reinvite()
     * callback.
     *
     * The index of the media dir will correspond to the provisional media
     * in pjsua_call_info.prov_media.
     * For offers that involve adding new medias (such as initial offer),
     * the index will correspond to all new audio media first, then video.
     * For example, for a new call with 2 audios and 1 video, media_dir[0]
     * and media_dir[1] will be for the audios, and media_dir[2] video.
     *
     * Default: PJMEDIA_DIR_ENCODING_DECODING
     */
    pjmedia_dir	     media_dir[PJMEDIA_MAX_SDP_MEDIA];
```

Example 1:
```
    /* Initiate a call as sendonly audio */
    pjsua_call_setting_default(&call_opt);
    call_opt.flag |= PJSUA_CALL_SET_MEDIA_DIR;
    call_opt.media_dir[0] = PJMEDIA_DIR_ENCODING;
    pjsua_call_make_call(current_acc, &dest, &call_opt, NULL, NULL, &current_call);
    ...
    /* After some time, update the audio to bidirectional */
    call_opt.flag |= PJSUA_CALL_SET_MEDIA_DIR;
    call_opt.media_dir[0] = PJMEDIA_DIR_ENCODING_DECODING;
    pjsua_call_reinvite2(current_call, &call_opt, NULL);
```
Example 2:
```
/* Answer video call as recvonly because we only have a display and no camera */
static void on_incoming_call(...)
{
    pjsua_call_get_info(call_id, &call_info);
    pjsua_call_setting_default(&call_opt);
    call_opt.flag |= PJSUA_CALL_SET_MEDIA_DIR;
    for (mi = 0; mi < call_info.prov_media_cnt; mi++) {
        if (call_info.prov_media[mi].type == PJMEDIA_TYPE_VIDEO)
            call_opt.media_dir[mi] = PJMEDIA_DIR_DECODING;
    }
    pjsua_call_answer2(call_id, &call_opt, 200, NULL, NULL);
```

### DIFF
--- a/pjmedia/include/pjmedia/endpoint.h
+++ b/pjmedia/include/pjmedia/endpoint.h
@@ -60,6 +60,17 @@ typedef enum pjmedia_endpt_flag
 
 } pjmedia_endpt_flag;
 
+/**
+ * This structure specifies various settings that can be passed when creating
+ * audio/video sdp.
+ */
+typedef struct pjmedia_endpt_create_sdp_param
+{
+    /**
+     * Direction of the media.
+     */
+    pjmedia_dir dir;
+} pjmedia_endpt_create_sdp_param;
 
 /**
  * Type of callback to register to pjmedia_endpt_atexit().
@@ -296,16 +307,17 @@ PJ_DECL(pj_status_t) pjmedia_endpt_create_base_sdp(pjmedia_endpt *endpt,
  * @param endpt		The media endpoint.
  * @param pool		Pool to allocate memory from.
  * @param si		Socket information.
- * @param options	Option flags, to indicate the audio direction.
+ * @param options	Options parameter, can be NULL.
  * @param p_m		Pointer to receive the created SDP media.
  *
  * @return		PJ_SUCCESS on success, or the appropriate error code.
  */
-PJ_DECL(pj_status_t) pjmedia_endpt_create_audio_sdp(pjmedia_endpt *endpt,
-                                                    pj_pool_t *pool,
-                                                    const pjmedia_sock_info*si,
-                                                    unsigned options,
-                                                    pjmedia_sdp_media **p_m);
+PJ_DECL(pj_status_t)
+pjmedia_endpt_create_audio_sdp(pjmedia_endpt *endpt,
+                               pj_pool_t *pool,
+                               const pjmedia_sock_info *si,
+                               const pjmedia_endpt_create_sdp_param *options,
+                               pjmedia_sdp_media **p_m);
 
 /**
  * Create SDP media line for video media.
@@ -313,16 +325,17 @@ PJ_DECL(pj_status_t) pjmedia_endpt_create_audio_sdp(pjmedia_endpt *endpt,
  * @param endpt		The media endpoint.
  * @param pool		Pool to allocate memory from.
  * @param si		Socket information.
- * @param options	Option flags, to indicate the video direction.
+ * @param options	Options parameter, can be NULL.
  * @param p_m		Pointer to receive the created SDP media.
  *
  * @return		PJ_SUCCESS on success, or the appropriate error code.
  */
-PJ_DECL(pj_status_t) pjmedia_endpt_create_video_sdp(pjmedia_endpt *endpt,
-                                                    pj_pool_t *pool,
-                                                    const pjmedia_sock_info*si,
-                                                    unsigned options,
-                                                    pjmedia_sdp_media **p_m);
+PJ_DECL(pj_status_t)
+pjmedia_endpt_create_video_sdp(pjmedia_endpt *endpt,
+                               pj_pool_t *pool,
+                               const pjmedia_sock_info *si,
+                               const pjmedia_endpt_create_sdp_param *options,
+                               pjmedia_sdp_media **p_m);
 
 /**
  * Dump media endpoint capabilities.
--- a/pjmedia/include/pjmedia/endpoint.h
+++ b/pjmedia/include/pjmedia/endpoint.h
@@ -296,7 +296,7 @@ PJ_DECL(pj_status_t) pjmedia_endpt_create_base_sdp(pjmedia_endpt *endpt,
  * @param endpt		The media endpoint.
  * @param pool		Pool to allocate memory from.
  * @param si		Socket information.
- * @param options	Option flags, must be zero for now.
+ * @param options	Option flags, to indicate the audio direction.
  * @param p_m		Pointer to receive the created SDP media.
  *
  * @return		PJ_SUCCESS on success, or the appropriate error code.
@@ -313,7 +313,7 @@ PJ_DECL(pj_status_t) pjmedia_endpt_create_audio_sdp(pjmedia_endpt *endpt,
  * @param endpt		The media endpoint.
  * @param pool		Pool to allocate memory from.
  * @param si		Socket information.
- * @param options	Option flags, must be zero for now.
+ * @param options	Option flags, to indicate the video direction.
  * @param p_m		Pointer to receive the created SDP media.
  *
  * @return		PJ_SUCCESS on success, or the appropriate error code.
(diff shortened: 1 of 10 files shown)


pjmedia/include/pjmedia/endpoint.h
```diff
@@ -313,7 +313,7 @@ PJ_DECL(pj_status_t) pjmedia_endpt_create_audio_sdp(pjmedia_endpt *endpt,
  * @param endpt		The media endpoint.
  * @param pool		Pool to allocate memory from.
  * @param si		Socket information.
- * @param options	Option flags, must be zero for now.
+ * @param options	Option flags, to indicate the video direction.
```
nanangizz, line 316: Similar to above.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -620,15 +620,23 @@ on_make_call_med_tp_complete(pjsua_call_id call_id,
...
 {
+    unsigned i;
+
     opt->flag &= ~(PJSUA_CALL_UNHOLD | PJSUA_CALL_UPDATE_CONTACT |
 		   PJSUA_CALL_NO_SDP_OFFER | PJSUA_CALL_REINIT_MEDIA |
 		   PJSUA_CALL_UPDATE_VIA);
+
+    for (i = 0; i < PJSUA_MAX_CALL_MEDIA; i++) {
+    	opt->aud_dir[i] = opt->vid_dir[i] = PJMEDIA_DIR_ENCODING_DECODING;
+    }
```
nanangizz, line 639: Please make sure if direction needs to be reset by this function, note that this function is to reset temporary/one-off settings/options (rather than setting default values), while media direction is something to maintain over the subsequent in/outgoing re-INVITE/UPDATE, app should manually set the direction setting to change it and should not be surprised by auto-reset after sending/receiving re-INVITE.

sauwming: The specification of the issue is to specify audio/video direction when making/answering call, so I believe it's a one-time setting (will probably need add doc to make this clearer), rather than something that needs to be maintained.

There are some complications and complexities regarding this media direction, especially if it needs to be maintained. This is caused by which index refers to which media?

Note that the settings of audio and video count only specifies **active** audio/video streams. For example: `aud_cnt` is 1 and `aud_dir[0]` is `PJMEDIA_DIR_ENCODING`. Then there are 2 audio SDP, first one inactive, second one active. `aud_dir[0]` will refer to the first active media, i.e. the second media.

But the status and property of the media can change over time. Some of the scenario that can happen:
- the first media activated
- a third audio media added
- the second one becomes inactive after double hold
- remote can change the property of any of the media by adding srtp (note that when answering we sort the media based on the "quality", so the first media is not necessarily the first active media in the SDP).
- and although this is illegal by the RFC, remote can also remove media.

Since it's no longer clear `aud_dir[0]` should refer to which media, that's why I made it a one-time setting and reset it every time. The cons of one time setting is that for every re-INVITE/UPDATE, app will need to specify the media direction again if desired. But I believe this is still much less complicated than trying to understand the index referencing.

Another alternative that I can think of is to make `aud_dir[i]` always refers to the i-th audio media in the SDP. The drawback of this alternative is that there will be a disconnect between `aud_cnt` and `aud_dir`.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -321,6 +321,13 @@ typedef struct pj_stun_resolve_result pj_stun_resolve_result;
 #   define PJSUA_ACC_MAX_PROXIES    8
 #endif
 
+/**
+ * Maximum number of call medias to be supported.
+ */
+#ifndef PJSUA_MAX_CALL_MEDIA
+#   define PJSUA_MAX_CALL_MEDIA		PJMEDIA_MAX_SDP_MEDIA
+#endif
```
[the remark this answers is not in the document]
nanangizz: Yes, noticed that, but it was internal and not configurable, now it becomes public and configurable.

pjmedia/include/pjmedia/endpoint.h
```diff
@@ -296,7 +296,7 @@ PJ_DECL(pj_status_t) pjmedia_endpt_create_base_sdp(pjmedia_endpt *endpt,
  * @param endpt		The media endpoint.
  * @param pool		Pool to allocate memory from.
  * @param si		Socket information.
- * @param options	Option flags, must be zero for now.
+ * @param options	Option flags, to indicate the audio direction.
```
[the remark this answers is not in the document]
nanangizz: Yes, but I think it is okay to change it to pointer, app should not see any problem with such change as currently it must be set to zero.

pjsip/src/pjsua-lib/pjsua_call.c
nanangizz: Agree, I think it needs a specification doc please.

Another idea is to introduce a flag in call setting to specify that directions settings are set and to be applied by the following op. So the `clear_flag()` function will just clear the flag and other op will not reset the media direction.

Regarding media dir indexing, just an idea, perhaps we can just make it one array only for all media types, similar to media info in call info. Yes, a disconnect between `aud_cnt` and `aud_dir`/`med_dir` is okay, just the spec needs to be clear about this, e.g: for making call, audio will take the first part, then video, so med_dir[0] will refer to audio (if any) and so on, for other ops (incoming call, re-INVITE, etc), it seems to be simpler.

Let's discuss this further over Slack if needed.

sauwming: One array for all media is a good idea indeed. It avoids the index complication and is much simpler to implement. So please review the latest change. Thanks.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -1055,6 +1055,25 @@ typedef struct pjsua_call_setting
      */
     unsigned         vid_cnt;
 
+    /**
+     * Media direction. This setting will only be used if the flag
+     * PJSUA_CALL_SET_MEDIA_DIR is set, and it will persist for subsequent
+     * offers or answers.
```
nanangizz, line 1061: - "it will persist for subsequent offers or answers" may cause misinterpretation that the media dir setting cannot be updated later?
- To make directions persistence, shouldn't `PJSUA_CALL_SET_MEDIA_DIR` be reset in `pjsua_call_cleanup_flag()`?

nanangizz (commented):
Please make sure all re-INVITE scenarios are also tested, i.e: with/without `SET_MEDIA_DIR` flag, incoming/outgoing re-INVITE.

nanangizz (approved):
- It would be great to have specifications & how-to-s/sample code in the PR desc.
- Tests.

trengginas (approved):

PR #2708, opened 2021-05-07
Do not redefine SSL_CIPHER_get_id() for LibreSSL

botovq:
Both SSL_CIPHER_get_id() and SSL_set_session() are available in all
LibreSSL versions. SSL_CIPHER will become opaque in LibreSSL 3.4.x,
so this redefinition reaching into the struct will break the build.

### DIFF
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -130,7 +130,7 @@ static unsigned get_nid_from_cid(unsigned cid)
 #     define X509_get_notBefore(x)  X509_get0_notBefore(x)
 #     define X509_get_notAfter(x)   X509_get0_notAfter(x)
 #  endif
-#else
+#elif !USING_LIBRESSL
 #  define SSL_CIPHER_get_id(c)	    (c)->id
 #  define SSL_set_session(ssl, s)   (ssl)->session = (s)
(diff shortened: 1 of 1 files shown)


sauwming:
Since the latest release is 3.3.3 and the doc only specifies change logs up to that version, can you point out the 3.4 roadmap plan or the commit, since I can't seem to find it?

botovq:
On Sun, May 09, 2021 at 07:35:03PM -0700, sauwming wrote:
> Since the latest release is 3.3.3 and the doc only specifies change logs up to that version, can you point out the 3.4 roadmap plan or the commit, since I can't seem to find it?

The change making SSL_CIPHER and SSL_SESSION opaque not yet committed.
There is no publicly available roadmap plan.  I can post a link to the
commit here when that happens.

However, you can find both SSL_CIPHER_get_id() and SSL_set_session() in
ssl.h and see from their blames that they have always been available:

https://github.com/libressl-portable/openbsd/blob/b70510b0c39be0ec556017b70b6e4c0b253e2d5c/src/lib/libssl/ssl.h#L1314
https://github.com/libressl-portable/openbsd/blob/b70510b0c39be0ec556017b70b6e4c0b253e2d5c/src/lib/libssl/ssl.h#L1404

nanangizz (approved):

Issue #865, opened 2009-05-22
More clever RTP transport remote address switch


pjsipbot:
_2009-05-22 16:24:15: @bennylp_ _created the issue on trac ticket [865](https://trac.pjsip.org/repos/ticket/865)_

Currently the UDP and ICE media transport has the capability to switch destination RTP/RTCP address to the source address of the RTP/RTCP packets, if they are different than the original address set in SDP received from remote.

It may be better to put this functionality in the stream instead, for the following reason:
 1. the stream has more information about the packet, such as RTP pt, SSRC, and info from RTCP statistic
 1. some application that uses media transport wants to inspect and possibly filter the source address
 1. some firewall such as Microsoft Forefront Threat Management Gateway wants us to send the RTP to the address in SDP, regardless of the source address of the RTP packet

___
_2009-05-22 16:56:37: @bennylp_ _edited the issue description_

___
_2009-05-23 14:42:36: @bennylp_ _changed **priority** from **normal** to **major**_

___
_2009-05-30 09:42:36: @bennylp_ _changed **milestone** from **Sipit24** to **release-1.3**_

___
_2009-06-17 11:00:44: @bennylp_ _changed **milestone** from **release-1.3** to **release-1.4**_

___
_2009-06-17 11:00:44: @bennylp_ _commented_

There is a temptation to support multiple callback users, so that multiple applications (e.g. stream and PJSUA-LIB) can peek the incoming RTP/RTCP packets. Some sample usage of this includes:
 - supporting STUN keep-alive (done by pjsua-lib)
 - packet filtering
 - PoC (by capturing RTCP traffic)

Sure this can be done by creating transport adapter, but utilizing multiple callback users will make it lot easier for application (e.g. application doesn't need to handle transport creation in PJSUA-LIB).

Pending for further review. Moving to 1.4 milestone.

___
_2009-08-13 00:17:02: @bennylp_ _changed **milestone** from **release-1.4** to **release-1.5**_

___
_2009-10-29 04:59:50: @bennylp_ _changed **milestone** from **release-1.5** to **release-1.6**_

___
_2010-01-12 09:43:22: @bennylp_ _changed **milestone** from **release-1.6** to **Known-Issues-and-Ideas**_

___
_2011-09-08 07:53:41: @bennylp_ _commented_

Enhancement of this behavior in #1366

___
_2018-03-08 01:55:12: @mingteluu_ _edited the issue description_

___
[...]

PR #2713, opened 2021-05-19
Filter RTP based on source address early to avoid invalid RTP session update

trengginas:
Changes from #865 affect how filtering RTP packet is processed.
- Pre #865, the RTP filter by source address is done on the transport media layer (e.g: transport UDP). RTP session updating should be done on filtered packet.
- Post #865, RTP filter is done on stream. RTP session update is done before packet filtering. If both sources somehow are using the same SSRC, RTP session updating will no longer be valid. 

This ticket will move the RTP filtering early to avoid the above issue.


pjmedia/src/pjmedia/rtp.c
```diff
@@ -250,7 +251,7 @@ PJ_DEF(void) pjmedia_rtp_session_update2( pjmedia_rtp_session *ses,
         ses->peer_ssrc = pj_ntohl(hdr->ssrc);
 
     if (pj_ntohl(hdr->ssrc) != ses->peer_ssrc) {
-	seq_st.status.flag.badssrc = 1;
+	flag_badssrc = 1;
```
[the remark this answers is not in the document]
trengginas: the badssrc was actually set for `p_seq_st` which is the output param. I agree it should use the original var to set.

pjmedia/src/pjmedia/stream.c
```diff
@@ -1888,64 +1889,28 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
...
-    if (seq_st.status.flag.bad) {
-	pkt_discarded = PJ_TRUE;
-	goto on_return;
-    }
-
     /* Ignore if payloadlen is zero */
     if (payloadlen == 0) {
 	pkt_discarded = PJ_TRUE;
 	goto on_return;
     }
```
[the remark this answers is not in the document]
trengginas: Right, it should be reported to the RTP session.

pjmedia/src/pjmedia/rtp.c
```diff
@@ -233,7 +233,7 @@ PJ_DEF(void) pjmedia_rtp_session_update2( pjmedia_rtp_session *ses,
 					  pjmedia_rtp_status *p_seq_st,
 					  pj_bool_t check_pt)
 {
-    pjmedia_rtp_status seq_st;
+    pjmedia_rtp_status seq_st;
```
nanangizz, line 236: Clean up please.

pjmedia/src/pjmedia/rtp.c
```diff
@@ -260,7 +260,7 @@ PJ_DEF(void) pjmedia_rtp_session_update2( pjmedia_rtp_session *ses,
 	if (p_seq_st) {
 	    p_seq_st->status.value = seq_st.status.value;
 	    p_seq_st->status.flag.bad = 1;
-	    p_seq_st->status.flag.badpt = 1;
+	    p_seq_st->status.flag.badpt = 1;
```
nanangizz, line 263: Clean up please.

pjmedia/src/pjmedia/rtp.c
```diff
@@ -281,6 +281,7 @@ PJ_DEF(void) pjmedia_rtp_session_update2( pjmedia_rtp_session *ses,
     if (p_seq_st) {
 	p_seq_st->status.value = seq_st.status.value;
 	p_seq_st->diff = seq_st.diff;
+	p_seq_st->status.flag.badssrc = seq_st.status.flag.badssrc;
```
nanangizz, line 284: Copying `status.value` above should be sufficient to copy all flags including `badssrc`, shouldn't it?

nanangizz (commented):
> Changes from #865 affect how unwanted RTP packet is processed.
> Discarding invalid RTP packet was done after RTP session was updated.
> This in turn will affect jitter buffer operation since bad RTP data was also processed.
> 
> This patch will move discarding unwanted RTP packet before updating RTP session to make sure
> the RTP session was for a valid one.

It looks like [RTP packet discard](https://github.com/pjsip/pjproject/pull/2713/files#diff-93eb538d1b4b08a610a83e9d36387763ff20184cfe5946313dcf2a19707a5bb3R1996) is still done after updating RTP session (`pjmedia_rtp_session_update2`). To me, the patch seems to be about moving out SSRC update in the RTP session to app (i.e: stream), and not sure how it affects the stream?

Btw, could you be a bit more detail about the problem this PR wants to fix (e.g: what is invalid RTP here, how jitter buffer is affected by that, what is the expected behavior).

pjmedia/src/pjmedia/stream.c
```diff
@@ -1888,64 +1890,22 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
...
-    }
-
-    /* Ignore if payloadlen is zero */
-    if (payloadlen == 0) {
-	pkt_discarded = PJ_TRUE;
-	goto on_return;
+    if (pj_ntohl(hdr->ssrc) != channel->rtp.peer_ssrc) {
+	flag_badssrc = 1;
+	if (!channel->rtp.has_peer_ssrc)
+	    channel->rtp.peer_ssrc = pj_ntohl(hdr->ssrc);
```
nanangizz, line 1900: looks like `channel->rtp` is manually updated by app (also in previous change block), is this really necessary?

pjmedia/src/pjmedia/stream.c
```diff
@@ -2176,6 +2188,9 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
     }
 
 on_return:
+    if (!update_rtcp)
+	return;
+
```
nanangizz, line 2193: This return does not seem to only skip RTCP update, but it also skips sending initial RTCP & RTCP-FB, so the variable name `update_rtcp` is perhaps a bit misleading.

Even perhaps we should refactor this, e.g: some cases (e.g: valid but zero payload length) still trigger RTCP update & sending RTCP/RTCP-FB, and other cases simply return?

pjmedia/src/pjmedia/stream.c
trengginas: It was taken from [here](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/rtp.c#L248).
Set the peer_ssrc : 
- If remote doesn't specify its ssrc and peer_ssrc is not yet set
- If remote doesn't specify its ssrc and peer_ssrc is set however diferent than the one received, allow the ssrc to be changed.

nanangizz: Yes, aware of where it was taken from actually. But, is it necessary or what for?

pjmedia/src/pjmedia/stream.c
trengginas: Discarded packet cause:
1. invalid source address 
2. sequence status bad flag (`seq_st.status.flag.bad`) is set
3. empty payload

Only empty payload (3) will need to process/update RTCP.

nanangizz: Agree, I think the change can be included in this PR too (e.g: simply by introducing another goto label).

pjmedia/src/pjmedia/stream.c
nanangizz: Trying to rephrase here, if app needs to assign `flag_badssrc` (btw, is it _bad_ or just _unrecognized_?), it is okay. But unless it is really necessary, don't update RTP session too (`channel->rtp`) as it should be done via `pjmedia_rtp_session_update2()` which is called later.

pjmedia/src/pjmedia/stream.c
trengginas: Actually there are other scenario which packets are ignored:
1. Error decoding RTP ([ref](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream.c#L1874)). This will update rtcp stats. If source address is not valid, no need to update rtcp stats.
2. RTCP data ([ref](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream.c#L1882)). No changes needed
3. Channel is paused ([ref](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream.c#L1888))
If source address is not valid, process/update RTCP.

Re: new goto label
There's no other processing needed, I think we can just return from the method?

nanangizz: > 3. Channel is paused ([ref](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream.c#L1888))
>    If source address is not valid, process/update RTCP.

Okay. Just make sure that regular RTCP RR is still sent when the stream is paused.

> 
> Re: new goto label
> There's no other processing needed, I think we can just return from the method?

Perhaps, but RTCP session may still need to be updated (via `pjmedia_rtcp_rx_rtp2`, note the `discarded` param) regardless RTP is bad/not, please check RTCP-XR specs.

pjmedia/src/pjmedia/stream.c
```diff
@@ -1843,8 +1843,8 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
     unsigned payloadlen;
     pjmedia_rtp_status seq_st;
     pj_bool_t check_pt;
-    pj_status_t status;
-    pj_bool_t pkt_discarded = PJ_FALSE;
+    pj_status_t status;    
+    pj_bool_t pkt_discarded = PJ_FALSE;
```
nanangizz, line 1847: clean up.

nanangizz (approved):
just a minor.

sauwming (approved):

PR #2715, opened 2021-05-24
Possible DTLS nego failure when using ICE

nanangizz:
When ICE has host and relay candidates, the default address in SDP will be the relay address. However when eventually ICE nego resulted in a valid pair with the host candidate, it will modify the default address in SDP (via SIP UPDATE/re-INVITE). Unfortunately the peer endpoint will restart the DTLS (invoking `ssl_destroy()` and so on) when it sees remote default address change, and eventually this will cause DTLS nego to fail.

The proposed solution is to get remote address (to detect any change) using info from ICE transport info instead of SDP. Note that DTLS will not be started before ICE nego is completed, so initially remote address info from ICE will be invalid and invalid remote address will not trigger DTLS restart.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -909,6 +909,7 @@ static pj_status_t get_rem_addrs(dtls_srtp *ds,
     pjmedia_sdp_conn *conn;
     pjmedia_sdp_attr *a;
     int af = pj_AF_UNSPEC();
+    pj_bool_t use_ice_info = PJ_FALSE;
 
     /* Init RTP & RTCP address */
     pj_bzero(rem_rtp, sizeof(*rem_rtp));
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -910,7 +910,28 @@ static pj_status_t get_rem_addrs(dtls_srtp *ds,
...
+	pjmedia_transport_get_info(ds->srtp->member_tp, &info);
+	ice_info = (pjmedia_ice_transport_info*)
+		   pjmedia_transport_info_get_spc_info(
+				    &info, PJMEDIA_TRANSPORT_TYPE_ICE);
+	if (ice_info) {
+	    *rem_rtp = ice_info->comp[0].rcand_addr;
+	    if (ice_info->comp_cnt > 1)
+		*rem_rtcp = ice_info->comp[1].rcand_addr;
+
+	    goto check_rtcp_mux;
```
sauwming, line 932: so far we use goto for purposes such as `on_return` or `on_error`, so perhaps it's better to avoid this. especially since here it seems we can just use `else` instead.

sauwming (approved):

pjmedia/src/pjmedia/transport_srtp_dtls.c
nanangizz: The goto is for avoiding 'noise' in the diff (no change other than indentation). When examining change history, sometime we need to compare changes line by line and as there will be ~45 lines to be shifted right, I think it is annoying to compare all 45 lines to eventually see nothing changed there. Should it be changed to using `else`?

sauwming: If only the indentation changes, I believe the CR will only mark it with `>>>`, or is this only valid for the old CR (I can't remember)? But anyway, since the codebase clarity and cleanliness is more important than the cleanliness of CR diff, I vote for `else`. Otherwise, we will potentially see a lot of `goto` in the future in order to minimise patch changes.

sauwming: Btw, I just tried it: the option "Hide white space changes" (available in the "Files changed" tab) does the trick to hide indentation modifs.

sauwming (approved):

trengginas (approved):

PR #2716, opened 2021-05-25
Race condition in SSL socket server

nanangizz:
There were a couple reports about crash related to SSL socket. In the investigation, we found a silly bug in SSL socket server code: the configured group lock is not applied to underlying socket (active socket), so race condition may happen between callback and destroy.

Also, call stack in one report points to certificate verification callback (`verify_cb()` function), which may also be caused by above bug or something else (e.g: OpenSSL application data index acquired via `SSL_get_ex_new_index()` somehow gets corrupted), so this PR also address the issue by adding normal validation check of SSL socket instance (was only assertion).


sauwming (approved):

trengginas (approved):

trengginas:
Should not reopen this, use https://github.com/pjsip/pjproject-ghsa-cv8x-p47p-99wr/pull/1 for merging.

PR #2718, opened 2021-05-31
Ignore 192.0.0.0/29 address

sauwming:
According to [RFC 7335](https://datatracker.ietf.org/doc/html/rfc7335#section-4), the address `192.0.0.0/29` can't receive any data and is used for ipv4/ipv6 translation. So it's probably better to remove this address when enumerating interfaces, as it will not work anyway.

Thanks to Sebastien Blin for the patch.


nanangizz (approved):

trengginas (approved):

AmarOk1412:
(I'm sebastien) for info, it's behalf of Savoir-Faire Linux :)

PR #2719, opened 2021-05-31
Update doc for ignoring stream keep alive

sauwming:
When sending stream keep alive using empty RTP (i.e. by setting `PJMEDIA_STREAM_ENABLE_KA` to `PJMEDIA_STREAM_KA_EMPTY_RTP`), the size of keep-alive packet sent will be equal to the `sizeof(pjmedia_rtp_hdr)`, but the check of ignoring it seem to expect packets strictly smaller than this, so keep-alive packets will never be ignored.

**Update:** Since the behavior seems to be okay, we update the doc instead, so it's clear that we only ignore non-RTP keep alive packets.

### DIFF
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -1865,7 +1865,7 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
     }
 
     /* Ignore keep-alive packets */
-    if (bytes_read < (pj_ssize_t) sizeof(pjmedia_rtp_hdr))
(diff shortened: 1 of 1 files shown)


nanangizz:
I think the comment of "Ignore keep-alive packets" is for other than emtpy-RTP KA packets, e.g: user KA, because all RTP packets, including empty/KA, should be validated and reported to RTP & RTCP stats, otherwise the stats will be skewed, e.g: the empty RTPs will be counted as lost packets. So the code should be fine, perhaps just update the comment?

nanangizz (approved):

trengginas (approved):

PR #2722, opened 2021-06-01
Add support for WebRTC AEC3

sauwming:
The current WebRTC source included in PJSIP is already quite old (from Oct 2015) and only supports a clock rate of 16 KHz (see #2404). So in this PR, we add the latest WebRTC AEC3 into PJSIP.

Requirement: `C++17`
Due to this requirement, and the complexity needed to build the AEC3 component (in my local Mac machine, PJSIP build time nearly doubles from ~35 sec to ~65 sec if AEC3 is included), by default AEC3 will be disabled and the old WebRTC AEC is still made available.

Licenses: check `third_party/webrtc_aec3/PJSIP_NOTES` about the licenses of WebRTC itself and external third party components required, such as `abseil`, `rnnoise`, and `pffft`.

Features:
- AEC3
- Noise Suppressor
  Pass the option `PJMEDIA_ECHO_USE_NOISE_SUPPRESSOR` to enable it.
- AGC2
  The next generation AGC functionality. AGC2 aims to automatically adjust levels by acting on the microphone gain and/or applying digital gain.
  Pass the option `PJMEDIA_ECHO_USE_GAIN_CONTROLLER` to enable it.

To support WebRTC AEC3, enable it when configuring PJSIP, i.e.:
`./configure --enable-libwebrtc-aec3`
And check if it can be built:
`checking if WebRtc AEC3 can be compiled with C++17... yes`

To use it, specify `PJMEDIA_ECHO_WEBRTC_AEC3` in `pjsua_media_config.ec_options` or `MediaConfig::ecOptions`.
You can test with our sample app pjsua, by pass the number 4 as the EC option desired, i.e..:
`./pjsua --ec-opt=4`
You can check the EC statistics using the API `pjsua_get_ec_stat()` or from `pjsua_call_dump()`. It will output information like this:
`EC stat: WebRTC AEC3 metrics: delay=12 ms, return loss=-7.30, return loss enh=1.09`


ismangil (approved):
Also update detailed Licensing wiki

nanangizz (approved):
As briefly discussed before, there seem to be no microphone gain adjustment? So perhaps we should remove or add some notes about "acting on the microphone gain" in the PR desc?

sauwming:
> As briefly discussed before, there seem to be no microphone gain adjustment? So perhaps we should remove or add some notes about "acting on the microphone gain" in the PR desc?

It's the original wording in their documentation, taken from:
https://github.com/pjsip/pjproject/blob/webrtc-aec3/third_party/webrtc_aec3/src/modules/audio_processing/gain_controller2.h#L28

And the dictionary says that "acting on" means "to use (something, such as a feeling or suggestion) as a reason or basis for doing something", so what the doc says is that it uses microphone gain information and then apply digital gain.

trengginas (approved):

maneeshj:
I did the above steps to build AEC3, but my logs output AEC3 stats as all 0 numbers. I am not sure if AEC is actually kicked in or not. Does O in return loss, delay indicate it is not being used ?

monocyte:
Having the same issue: followed build instructions on macOS and Windows, but every call stats show zeroes in AEC3 stats on both OS:
`EC stat: WebRTC AEC3 metrics: delay=0 ms, return loss=0.00, return loss enh=0.00`

fbreton06:
Hi guys, sorry it's probably not the better location to post that. 
I've write a python program to establish a SIP audio communication on my Raspberry4.
It's based on pjsua2 module from PJSIP V1.13.1 release.
I'm trying to suppress echo with activating the PJMEDIA_ECHO_WEBRTC_AEC3, but result is disappointing.
Has anyone advices to improve that?
cheers,

PR #2723, opened 2021-06-02
swift-C++ binding & swift pjsua2 sample app 

emretufekci:
Swift & C++ Bindings & Sample App for 64 bit Simulator  (PJSua2 Level)


sauwming:
Thank you very much for providing an iOS Swift sample app using pjsua2.

I tested it here and I had to fix a few things to make it run (such as by defining PJ_AUTOCONF=1), but after that, it can successfully run without any issue on arm64, so I'm not sure why you have to limit your target to the simulator.

Also, the registration login and call destination are still hard coded, so you may need to remove it.

emretufekci:
> Thank you very much for providing an iOS Swift sample app using pjsua2.
> 
> I tested it here and I had to fix a few things to make it run (such as by defining PJ_AUTOCONF=1), but after that, it can successfully run without any issue on arm64, so I'm not sure why you have to limit your target to the simulator.
> 
> Also, the registration login and call destination are still hard coded, so you may need to remove it.

Hardcoded parts removed, i forgot to remove it 👍 
I tried to add PJ_AUTOCONF=1 but couldn't run via that config, that's why pushed it as is. Looking for it. If you have it already, could you share it?

sauwming:
Here's the modification of the pbxproj file that I use:
[pbxproj_diff.txt](https://github.com/pjsip/pjproject/files/6595213/pbxproj_diff.txt)

emretufekci:
Thank you! worked like a charm 👍

Pushed latest changes as well.

sauwming:
- Can you please move the directory to a new one `pjsip-apps/src/pjsua2/ios-swift` (the app name can remain the same) instead of `pjsip-apps/src/samples`? This is because `samples` directory are meant for simple samples only that consist of a single file or dummy samples (in the case of Android).

- For the source files *.swift, *.h/hpp, *.cpp, *.mm, we'd like to ask your permission if you're okay if we change the copyright info to look like https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-codec/silk.c . If you agree, you can change this yourself, or we can change it for you after merging later, that's up to you.

emretufekci:
Sure, no problem at all i added copyright message.

I created pjsua2 directory same level with pjsua folder and moved sample app to there. But i am not sure is it proper way for you.

sauwming:
The part "Contributed by", that's where you can put your name :)
Or you can remove it if you prefer to remain anonymous.

emretufekci:
I missed the part now added it :)
I guess finally it's okay

.gitignore
```diff
@@ -45,6 +45,7 @@ pjsip-apps/src/pjsua/ios*/lib*.a
 pjsip-apps/src/pjsua/ios/ipjsua.xcodeproj/project.xcworkspace/
 pjsip-apps/src/pjsua/ios/ipjsua.xcworkspace/xcshareddata
 pjsip-apps/src/pjsua/ios-swift/Libraries/
+pjsip-apps/samples/pjsua2/ios-swift-pjsua2/Libraries/
```
sauwming, line 48: need to change this directory, also below

pjsip-apps/build/Makefile
```diff
@@ -71,11 +71,13 @@ $(PJSUA_EXE):
 	$(MAKE) -f $(RULES_MAK) APP=PJSUA app=pjsua $(subst /,$(HOST_PSEP),$(BINDIR)/$@)
 	@if echo "$(TARGET_NAME)" | grep -q "apple-darwin_ios$$"; then \
 	  mkdir ../src/pjsua/ios-swift/Libraries ; \
+	  mkdir ../src/samples/ios-swift-pjsua2/Libraries ; \
```
sauwming, line 74: here

pjsip-apps/build/Makefile
```diff
@@ -71,11 +71,13 @@ $(PJSUA_EXE):
...
 	  for F in $(filter %$(TARGET_NAME).a,$(PJ_LIBXX_FILES)); do \
 	    if test -f $$F; then \
 	      tmp=`echo $${F##*/} | sed -e "s/\-$(TARGET_NAME)\.a/.a/"`; \
 	      ln -sf $$F ../src/pjsua/ios/$$tmp; \
 	      ln -sf $$F ../src/pjsua/ios-swift/Libraries/$$tmp; \
+	      ln -sf $$F ../src/samples/ios-swift-pjsua2/Libraries/$$tmp; \
```
sauwming, line 80: and here

emretufekci:
Solved with e77c002698ede7a8cdddd7154436628162d69145 👍

sauwming (approved):

nanangizz (approved):

trengginas (approved):

PR #2725, opened 2021-06-04
Fix PJMEDIA_STREAM_CHECK_RTP_PT flag usage

mscdex:
The existing `PJMEDIA_STREAM_CHECK_RTP_PT` conditional(s) are not enough because `pjmedia_rtp_session_update2()` actually checks `out_pt`, so setting `#define PJMEDIA_STREAM_CHECK_RTP_PT 0` has no effect in some cases.

In my particular case the answerer in SDP exchanges was replying with a different payload type, despite the payload description matching one that was offered. From what I understand in the SIP/SDP RFC, this is a bug with the answerer's implementation. While I would prefer being more selective for ignoring the payload type check (e.g. only for dynamic payload type situations -- e.g. opus) and would prefer having it be a runtime option rather than compile time, this at least allows me to work around the buggy implementation.


pjmedia/src/pjmedia/stream.c
```diff
@@ -1891,7 +1891,11 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
     /* Update RTP session (also checks if RTP session can accept
      * the incoming packet.
      */
-    check_pt = (hdr->pt != stream->rx_event_pt) && PJMEDIA_STREAM_CHECK_RTP_PT;
```
sauwming, line 1894: doesn't this already mean the same thing as below? i.e. if (PJMEDIA_STREAM_CHECK_RTP_PT==0) check_pt = (... && 0) = 0. And if  (PJMEDIA_STREAM_CHECK_RTP_PT==0) check_pt = (hdr->pt != stream->rx_event_pt) && 1;

mscdex:
I've updated the PR with the actual fix that I've now verified no longer results in "Bad RTP pt" messages.

pjmedia/src/pjmedia/stream.c
```diff
@@ -1893,8 +1893,8 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
      */
     check_pt = (hdr->pt != stream->rx_event_pt) && PJMEDIA_STREAM_CHECK_RTP_PT;
     pjmedia_rtp_session_update2(&channel->rtp, hdr, &seq_st, check_pt);
-#if !PJMEDIA_STREAM_CHECK_RTP_PT 
-    if (!check_pt && hdr->pt != channel->rtp.out_pt &&
+#if PJMEDIA_STREAM_CHECK_RTP_PT 
+    if (check_pt && hdr->pt != channel->rtp.out_pt &&
```
mscdex, line 1897: I changed the `check_pt` here to match how the check is made in rtp.c, I'm not sure what the original intention was though on this particular line.

nanangizz: To me, the intention is to assign `flag.badpt` when `PJMEDIA_STREAM_CHECK_RTP_PT` is not set. Note that if `PJMEDIA_STREAM_CHECK_RTP_PT` is set, `pjmedia_rtp_session_update2()` will handle the `badpt` assignment. So I think the existing code is already correct.

mscdex: The problem is that merely setting

```c
#define PJMEDIA_STREAM_CHECK_RTP_PT 0
```

in the config_site.h alone is not enough because with that defined that way, the few lines above effectively become:


```c
check_pt = PJ_FALSE;
// does not mark badpt
pjmedia_rtp_session_update2(&channel->rtp, hdr, &seq_st, check_pt);
// erroneously marks badpt, despite `check_pt` being false
if (hdr->pt != channel->rtp.out_pt &&
	hdr->pt != stream->rx_event_pt) 
```

nanangizz: Yes, then the `badpt` will be set to 1, that's actually the purpose of the `if !PJMEDIA_STREAM_CHECK_RTP_PT` block. Setting the `badpt` without `bad` won't cause the RTP packet to be dropped, it is just for logging purpose.

Issue #2717, opened 2021-05-27
Deadlock between pjsua_handle_ip_change and processing of incoming message


dstd:
I having the deadlock if the network change occurs during the call settling. 

In this case first thread performs `pjsua_handle_ip_change`, where `PJSUA_LOCK` executed, and then waiting for transport restarting completion. 

At the same time, in another thread, an incoming message comes, and during the processing, in one of functions `PJSUA_LOCK` trying to execute as well. 

So that we got a deadlock because first thread is still `PSJUA_LOCK`-ed and waiting until all transports complete their processing.

pjsip 2.10 @ android

Stack traces:

```
Thread 1:

pj_thread_sleep 		– os_core_unix.c:781
pjsip_udp_transport_restart2 	– sip_transport_udp.c:1262
restart_listener 		– pjsua_core.c:3739
pjsua_handle_ip_change 		– pjsua_core.c:3852
SomeClientApp::HandleNetworkChange()
...
```

```
Thread 2:

pj_mutex_lock                   - os_core_unix.c:1281
PJSUA_LOCK                      - pjsua_internal.h:593
pjsua_enum_calls                - pjsua_call.c:275
pjsua_check_snd_dev_idle        - pjsua_aud.c:408
pjsua_aud_stop_stream           - pjsua_aud.c:547
stop_media_stream               - pjsua_media.c:2886
pjsua_media_channel_update      - pjsua_media.c:3372
pjsua_call_on_media_update      - pjsua_call.c:4488
inv_negotiate_sdp               - sip_inv.c:1959
inv_check_sdp_in_incoming_msg   - sip_inv.c:2186
inv_on_state_confirmed          - sip_inv.c:5169
mod_inv_on_tsx_state            - sip_inv.c:726
pjsip_dlg_on_tsx_state          - sip_dialog.c:2129
tsx_set_state                   - sip_transaction.c:1272
tsx_on_state_proceeding_uac     - sip_transaction.c:0
tsx_on_state_calling            - sip_transaction.c:2599
pjsip_tsx_recv_msg              - sip_transaction.c:1832
mod_tsx_layer_on_rx_response    - sip_transaction.c:893
pjsip_endpt_process_rx_data     - sip_endpoint.c:938
endpt_on_rx_msg                 - sip_endpoint.c:1080
pjsip_tpmgr_receive_packet      - sip_transport.c:2182
udp_on_read_complete            - sip_transport_udp.c:192
[...]

PR #2731, opened 2021-06-10
Avoid deadlock when restarting SIP UDP transport due to holding pjsua lock

trengginas:
This ticket is to address #2717 by not holding PJSUA lock when calling `pjsip_udp_transport_restart2()`.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_core.c
+++ b/pjsip/src/pjsua-lib/pjsua_core.c
@@ -3781,6 +3781,7 @@ static pj_status_t restart_listener(pjsua_transport_id id,
 	int i = 0;
 	pj_bool_t all_done = PJ_TRUE;
 
+	PJSUA_LOCK();
 	pjsua_var.tpdata[id].is_restarting = PJ_FALSE;
 	pjsua_var.tpdata[id].restart_status = status;
 	if (pjsua_var.ua_cfg.cb.on_ip_change_progress) {
@@ -3803,6 +3804,7 @@ static pj_status_t restart_listener(pjsua_transport_id id,
 		break;
 	    }
 	}
+	PJSUA_UNLOCK();
 	if (all_done)
 	    status = handle_ip_change_on_acc();
     }
@@ -3846,12 +3848,12 @@ PJ_DEF(pj_status_t) pjsua_handle_ip_change(const pjsua_ip_change_param *param)
 		pjsua_var.tpdata[i].restart_status = PJ_EUNKNOWN;
 	    }
 	}
+	PJSUA_UNLOCK();
 	for (i = 0; i < PJ_ARRAY_SIZE(pjsua_var.tpdata); ++i) {
 	    if (pjsua_var.tpdata[i].data.ptr != NULL) {
 		status = restart_listener(i, param->restart_lis_delay);
 	    }
 	}
-        PJSUA_UNLOCK();
     } else {
 	for (i = 0; i < PJ_ARRAY_SIZE(pjsua_var.tpdata); ++i) {
 	    if (pjsua_var.tpdata[i].data.ptr != NULL) {
--- a/pjsip/src/pjsua-lib/pjsua_core.c
+++ b/pjsip/src/pjsua-lib/pjsua_core.c
@@ -3734,20 +3734,27 @@ static pj_status_t restart_listener(pjsua_transport_id id,
 
     switch (tp_info.type) {
     case PJSIP_TRANSPORT_UDP:
-    case PJSIP_TRANSPORT_UDP6:
+    case PJSIP_TRANSPORT_UDP6:    
     {
-	pjsip_udp_transport_res_param param;
+	unsigned num_locks = 0;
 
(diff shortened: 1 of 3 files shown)


nanangizz:
Theoretically PJSUA lock must be held while accessing any PJSUA state (e.g: `pjsua_var`), so better try to find another approach find, e.g: unlock temporarily in the loop spin.

pjsip/src/pjsua-lib/pjsua_core.c
```diff
@@ -3735,12 +3735,20 @@ static pj_status_t restart_listener(pjsua_transport_id id,
...
+	param.option = PJSIP_UDP_TRANSPORT_DESTROY_SOCKET;
+	param.sock = PJ_INVALID_SOCKET;
+	param.local = &bind_addr;
+	param.unlock = &PJSUA_UNLOCK;
+	param.lock = &PJSUA_LOCK;
+
+	status = pjsip_udp_transport_restart3(
+					    pjsua_var.tpdata[id].data.tp,
+					    &param);
+    }
```
nanangizz, line 3751: Why introduce new APIs, instead of simply unlocking PJSUA lock before invoking `pjsip_udp_transport_restart*()` and re-locking later after?

nanangizz:
Hmm.. I didn't analyze the problem in detail, so I don't really know whether callbacks in the previous patch are really needed or not. Also no idea whether the simpler approach is better.

So, could you describe the previous approach further. E.g: Is there any mutexes other than PJSUA, perhaps for other apps? And does it need to be unlocked right before the spin loop (as opposed to before `pjsip_udp_transport_restart()`)?

trengginas:
The deadlock issue is because prior to calling `pjsip_udp_transport_restart2()`, `pjsua_handle_ip_change()` will acquire pjsua lock.
When restarting UDP transport, it will wait for the read spin loop to finish. This will trigger a deadlock if there's a new incoming message which in turn tried to acquire the pjsua lock.
The alternatives:
- Make sure that there's no locks held before calling `pjsip_udp_transport_restart2()`
- Provide a mechanism to release the lock before waiting for the read spin loop to finish.

The new API on the previous alternative will provide mechanism for user to be able to release the lock before waiting for the read loop spin to finish.
However, since currently only `pjsua_handle_ip_change()` that will call `pjsip_udp_transport_restart2()`, releasing the pjsua lock before restarting the transport should be simpler and provide a solution to #2717.

nanangizz:
> The deadlock issue is because prior to calling `pjsip_udp_transport_restart2()`, `pjsua_handle_ip_change()` will acquire pjsua lock.
> When restarting UDP transport, it will wait for the read spin loop to finish. This will trigger a deadlock if there's a new incoming message which in turn tried to acquire the pjsua lock.
> The alternatives:
> 
> * Make sure that there's no locks held before calling `pjsip_udp_transport_restart2()`
> * Provide a mechanism to release the lock before waiting for the read spin loop to finish.
> 
> The new API on the previous alternative will provide mechanism for user to be able to release the lock before waiting for the read loop spin to finish.
> However, since currently only `pjsua_handle_ip_change()` that will call `pjsip_udp_transport_restart2()`, releasing the pjsua lock before restarting the transport should be simpler and provide a solution to #2717.

As `pjsip_udp_transport_restart2()` is public API, I don't think it is safe to assume that `pjsua_handle_ip_change()` is the only user. So, the question is: can we just suggest app (i.e: via specs docs) to release any locks before invoking `pjsip_udp_transport_restart2()`? Or it should be in callback form for some reason (please explain if it should)?

sauwming:
Since there's already a delay restart mechanism in `restart_listener()`, perhaps we can just use this?
i.e. if there's read spin loop, return PJ_EBUSY and try restarting it again later.

trengginas:
Personally, I think adding a warning in the doc should be sufficient.

trengginas:
If we return to the read spin loop, then there's a chance that the restart
will need to be retried indefinitely provided incoming messages are still
going on.

nanangizz:
> Personally, I think adding a warning in the doc should be sufficient.

Okay, so just need to update the `pjsip_udp_transport_restart*()` docs.

nanangizz:
Just did a quick check in `pjsip_udp_transport_restart2()` there is also a call to `pj_ioqueue_unregister()`, can this cause deadlock too, e.g: wait for ioqueue key lock being held by the read op?

trengginas:
It shouldn't cause deadlock for `pjsip_udp_transport_restart*()`, since any locks acquired in `pj_ioqueue_unregister()` will be released once the read loop wait is executed. 
Or perhaps you meant deadlock inside `pj_ioqueue_unregister()` ?

nanangizz:
Ah right, as now any mutex should be released by app prior to invoking `pjsip_udp_transport_restart*()`, `pj_ioqueue_unregister()` eventually will be executed.

nanangizz (approved):
It'd be better to wait for confirmation from reporter before merging this.

sauwming (approved):

dstd:
yep, it looks fixed now – deadlocks didn't appear anymore

PR #2737, opened 2021-06-16
Avoid call disconnection due to request timeout when network is changing

nanangizz:
Some requests (within dialog) may be in progress when network is changing, they may eventually get timed out and cause call disconnection.

This PR introduces a new PJSIP endpoint setting: `pjsip_cfg_t.endpt.keep_inv_after_tsx_timeout`, default is `PJ_FALSE` (maintaining the existing behavior). The IP change processing in PJSUA temporarily changes this setting value to `PJ_TRUE` to avoid call disconnection due to request timeout.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -4082,8 +4082,8 @@ static pj_bool_t handle_uac_tsx_response(pjsip_inv_session *inv,
     if (inv->state != PJSIP_INV_STATE_DISCONNECTED &&
 	((tsx->status_code == PJSIP_SC_CALL_TSX_DOES_NOT_EXIST &&
 	    tsx->method.id != PJSIP_CANCEL_METHOD) ||
-	 tsx->status_code == PJSIP_SC_REQUEST_TIMEOUT ||
-	 tsx->status_code == PJSIP_SC_TSX_TIMEOUT))
+	 (tsx->status_code == PJSIP_SC_REQUEST_TIMEOUT &&
+	  !pjsip_cfg()->endpt.keep_inv_after_tsx_timeout)))
     {
 	pjsip_tx_data *bye;
 	pj_status_t status;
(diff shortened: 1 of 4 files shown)


pjsip/src/pjsua-lib/pjsua_core.c
```diff
@@ -3817,6 +3817,15 @@ static void restart_listener_cb(void *user_data)
 }
 
 
+static void ip_change_put_back_inv_config(void *user_data)
+{
+    PJ_UNUSED_ARG(user_data);
+
+    PJ_LOG(4,(THIS_FILE,"IP change stops ignoring request timeout"));
+    pjsip_cfg()->endpt.keep_inv_after_tsx_timeout = PJ_FALSE;
```
sauwming, line 3825: Shouldn't we revert it to the previous value instead?

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -4082,8 +4082,8 @@ static pj_bool_t handle_uac_tsx_response(pjsip_inv_session *inv,
     if (inv->state != PJSIP_INV_STATE_DISCONNECTED &&
 	((tsx->status_code == PJSIP_SC_CALL_TSX_DOES_NOT_EXIST &&
 	    tsx->method.id != PJSIP_CANCEL_METHOD) ||
-	 tsx->status_code == PJSIP_SC_REQUEST_TIMEOUT ||
-	 tsx->status_code == PJSIP_SC_TSX_TIMEOUT))
```
sauwming, line 4086: so PJSIP_SC_TSX_TIMEOUT is removed?

sauwming (approved):

pjsip/src/pjsua-lib/pjsua_core.c
nanangizz: The timer is scheduled only when the value is `PJ_FALSE`.

pjsip/src/pjsip-ua/sip_inv.c
nanangizz: They are the same, see [here](https://github.com/pjsip/pjproject/blob/master/pjsip/include/pjsip/sip_msg.h#L490), and as it is defined in a public header file, it should not be changed.

sauwming (approved):

trengginas (approved):

Issue #2745, opened 2021-06-24
DTLS crashes when receiving "Encrypted Alert" after end of call


SamVanBelle:
There's a crash in ssl_on_recv_packet (pjmedia\transport_srtp_dtls.c) when invoked after ssl_destroy, because the SSL BIOs are deleted without resetting the pointers.
This can occur when a DTLS Encrypted Alert message arrives after SIP BYE/OK (invoking pjmedia_transport_media_stop).

Crash callstack:

```
 	libcrypto-1_1-x64.dll!bio_write_intern(bio_st * b, const void * data, unsigned __int64 dlen, unsigned __int64 * written) Line 328	C
 	libcrypto-1_1-x64.dll!BIO_write(bio_st * b, const void * data, int dlen) Line 363	C
>	SipStackTest.exe!ssl_on_recv_packet(dtls_srtp * ds, const void * data, unsigned __int64 len) Line 974	C
 	SipStackTest.exe!dtls_on_recv_rtp(pjmedia_transport * tp, const void * pkt, unsigned __int64 size) Line 1114	C
 	SipStackTest.exe!pjmedia_transport_send_rtp(pjmedia_transport * tp, const void * pkt, unsigned __int64 size) Line 843	C
 	SipStackTest.exe!srtp_rtp_cb(pjmedia_tp_cb_param * param) Line 1366	C
 	SipStackTest.exe!ice_on_rx_data(pj_ice_strans * ice_st, unsigned int comp_id, void * pkt, unsigned __int64 size, const void * src_addr, unsigned int src_addr_len) Line 2559	C
 	SipStackTest.exe!ice_rx_data(pj_ice_sess * ice, unsigned int comp_id, unsigned int transport_id, void * pkt, unsigned __int64 size, const void * src_addr, unsigned int src_addr_len) Line 2268	C
 	SipStackTest.exe!pj_ice_sess_on_rx_pkt(pj_ice_sess * ice, unsigned int comp_id, unsigned int transport_id, void * pkt, unsigned __int64 pkt_size, const void * src_addr, int src_addr_len) Line 3640	C
 	SipStackTest.exe!stun_on_rx_data(pj_stun_sock * stun_sock, void * pkt, unsigned int pkt_len, const void * src_addr, unsigned int addr_len) Line 2345	C
 	SipStackTest.exe!on_data_recvfrom(pj_activesock_t * asock, void * data, unsigned __int64 size, const void * src_addr, int addr_len, int status) Line 1004	C
[...]

PR #2746, opened 2021-06-24
Reset SSL BIOs (#2745) 

SamVanBelle:
to avoid crash when receiving DTLS packets after ssl_destroy.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -693,6 +693,8 @@ static pj_status_t ssl_flush_wbio(dtls_srtp *ds)
     pj_size_t len;
     pj_status_t status = PJ_SUCCESS;
 
+	if (ds->ossl_wbio == NULL) return PJ_EGONE;
+
     /* Check whether there is data to send */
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -528,6 +528,9 @@ static void ssl_destroy(dtls_srtp *ds)
 	}
 	SSL_free(ds->ossl_ssl); /* this will also close BIOs */
 	ds->ossl_ssl = NULL;
+	/* thus reset the BIOs as well */
+	ds->ossl_rbio = NULL;
+	ds->ossl_wbio = NULL;
     }
```
sauwming, line 534: Shouldn't some check be added to ssl_on_recv_packet() as well?
i.e.
`if (ds->ossl_rbio == NULL) return some error (PJ_EGONE?)`

SamVanBelle: This check is already available inside bio_write_intern
But I'll add it to PjSip as well to be independent

SamVanBelle:
Extra check added

sauwming (approved):

nanangizz (approved):

trengginas (approved):

PR #2759, opened 2021-07-07
SDL full-screen mode fixes and enhancement

nanangizz:
Continuation of PR #2324. This PR includes the original PR and implements few additional stuffs, e.g: instead of replacing fullscreen with fullscree desktop, this PR provides both modes, as discussed in the original PR.

So this PR covers:
- add fullscreen mode `PJMEDIA_VID_DEV_FULLSCREEN_DESKTOP` (no video mode change), which is mapped to SDL flag `SDL_WINDOW_FULLSCREEN_DESKTOP`,
- fix resizing while in full-screen,
- also update PJSUA, PJSUA2 & pjsua app to accomodate the new fullscreen mode setting, e.g: fullscreen setting was boolean (fullscreen enabled or disabled), now it is enum (disabled, fullscreen, or fullscreen desktop).

### DIFF
--- a/pjmedia/include/pjmedia/videodev.h
+++ b/pjmedia/include/pjmedia/videodev.h
@@ -163,6 +163,30 @@ enum pjmedia_vid_dev_std_index
 
 
 /**
+ * Enumeration of window fullscreen flags.
+ */
+typedef enum pjmedia_vid_dev_fullscreen
+{
+    /**
+     * Windowed or disable fullscreen.
+     */
+    PJMEDIA_VID_DEV_WINDOWED = 0,
+
+    /**
+     * Fullscreen enabled, video mode may be changed.
+     */
+    PJMEDIA_VID_DEV_FULLSCREEN = 1,
+
+    /**
+     * Fullscreen enabled by resizing video frame to match to the desktop,
+     * video mode will not be changed.
+     */
+    PJMEDIA_VID_DEV_FULLSCREEN_DESKTOP = 2
+
+} pjmedia_vid_dev_fullscreen;
+
+
+/**
  * This enumeration identifies various video device capabilities. These video
  * capabilities indicates what features are supported by the underlying
  * video device implementation.
@@ -469,10 +493,11 @@ typedef struct pjmedia_vid_dev_param
     unsigned window_flags;
 
     /**
-     * Video window's full screen status. This setting is optional, and will only be
-     * used if PJMEDIA_VID_DEV_CAP_OUTPUT_FULLSCREEN is set in the flags.
+     * Video window's fullscreen status. This setting is optional, and will
+     * only be used if PJMEDIA_VID_DEV_CAP_OUTPUT_FULLSCREEN is set in the
+     * flags.
      */
-    pj_bool_t window_fullscreen;
+    pjmedia_vid_dev_fullscreen window_fullscreen;
 
 } pjmedia_vid_dev_param;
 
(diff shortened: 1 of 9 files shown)


pjmedia/include/pjmedia/videodev.h
```diff
@@ -162,6 +162,30 @@ enum pjmedia_vid_dev_std_index
 };
 
 
+/**
+ * Enumeration of window fullscreen flags.
+ */
+typedef enum pjmedia_vid_dev_fullscreen
```
sauwming, line 168: Suggestion:
Change `pjmedia_vid_dev_fullscreen` enum to `pjmedia_vid_dev_fullscr_flag` or `pjmedia_vid_dev_fscreen_flag` or `pjmedia_vid_dev_fscr_flag`

pjsip/include/pjsua2/media.hpp
```diff
@@ -1880,6 +1880,16 @@ class VideoWindow
...
 
+    /**
+     * Set video window full-screen. This operation is valid only when the
+     * underlying video device supports PJMEDIA_VID_DEV_CAP_OUTPUT_FULLSCREEN
+     * capability. Currently it is only supported on SDL backend.
+     *
+     * @param mode		Fullscreen mode, see
+     *				pjmedia_vid_dev_fullscreen.
+     */
+    void setFullScreen2(pjmedia_vid_dev_fullscreen mode) PJSUA2_THROW(Error);
```
sauwming, line 1891: Why in pjsua2 we create a new API to avoid api-break but not in pjsua?

nanangizz: AFAIK in C++ assigning an integer value to an enum type is illegal, e.g: `setFullScreen2(PJ_TRUE)` will cause compile error. But it is legal in C, so no need to introduce a new API in pjsua.

pjmedia/include/pjmedia/videodev.h
nanangizz: Okay. Perhaps `pjmedia_vid_dev_fullscreen_flag`? Just ~3 chars more :)

sauwming: you mean 5 more? ;) as long as it contains "flag", it's okay.

pjsip/include/pjsua2/media.hpp
sauwming: in pjsua, there will be API break though, is this okay?

nanangizz: Existing pjsua app (using C, not C++) calling `pjsua_vid_win_set_fullscreen(PJ_TRUE/FALSE)` should not see any behavior change, so no API break?

pjsip-apps/src/swig/symbols.lst
```diff
@@ -12,7 +12,7 @@ pjmedia/echo.h                  pjmedia_echo_flag
 pjmedia/event.h                 pjmedia_event_type
 pjmedia/transport_srtp.h	pjmedia_srtp_use pjmedia_srtp_crypto_option pjmedia_srtp_keying_method
 pjmedia/vid_stream.h		pjmedia_vid_stream_rc_method
-pjmedia-videodev/videodev.h	pjmedia_vid_dev_index pjmedia_vid_dev_std_index pjmedia_vid_dev_cap
+pjmedia-videodev/videodev.h	pjmedia_vid_dev_index pjmedia_vid_dev_std_index pjmedia_vid_dev_cap pjmedia_vid_dev_fullscreen
```
sauwming, line 15: this needs to be changed as well

sauwming (approved):

trengginas (approved):

saghul:
Thanks for doing this!

nanangizz:
> Thanks for doing this!

Hi @saghul, glad to hear from you :)

PR #2762, opened 2021-07-08
Fix 's default constructor to do what the documentation says.

trandi:
We currently get `UndefinedBehaviorSanitizer: invalid-bool-load pjsip/src/pjsua2/account.cpp:639:40` when trying to create an `Account` with a default constructed `AccountConfig` which uses an `AccountNatConfig`, whose default constructor does NOT initialise its fields as per the documentation.

`AccountConfig`'s default constructor should set up the default values as it calls `pjsua_acc_config_default(&acc_cfg);`.

HOWEVER, `fromPj()` only populates some of the fields depending on the `ice_cfg_use` flag.

Regardless, given what the documentation for individual `AccountNatConfig` fields says, its default constructor should comply.

### DIFF
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -598,7 +598,7 @@ struct AccountNatConfig : public PersistentObject
      *
      * See also contactRewriteMethod field.
      *
-     * Default: TRUE
+     * Default: 1 (PJ_TRUE / yes)
      */
     int			contactRewriteUse;
 
@@ -623,7 +623,7 @@ struct AccountNatConfig : public PersistentObject
      * TCP socket when it is still connecting. In these cases, this
      * feature will also be turned off.
      *
-     * Default: 1 (yes).
+     * Default: 1 (PJ_TRUE / yes).
      */
     int			contactUseSrcPort;
 
@@ -633,7 +633,7 @@ struct AccountNatConfig : public PersistentObject
      * the REGISTER request, as long as the request uses the same transport
      * instance as the previous REGISTER request.
      *
-     * Default: TRUE
+     * Default: 1 (PJ_TRUE / yes)
      */
     int			viaRewriteUse;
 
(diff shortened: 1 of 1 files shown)


pjsip/include/pjsua2/account.hpp
```diff
@@ -706,9 +706,27 @@ struct AccountNatConfig : public PersistentObject
...
+      iceEnabled(false),
+      iceTrickle(PJ_ICE_SESS_TRICKLE_DISABLED),
+      iceMaxHostCands(-1),
+      iceAggressiveNomination(true),
+      iceNominatedCheckDelayMsec(PJ_ICE_NOMINATED_CHECK_DELAY),
+      iceWaitNominationTimeoutMsec(ICE_CONTROLLED_AGENT_WAIT_NOMINATION_TIMEOUT),
+      iceNoRtcp(false),
+      iceAlwaysUpdate(true),
+      turnConnType(PJ_TURN_TP_UDP),
+      contactRewriteUse(true),
```
nanangizz, line 720: As the type is `int`, should be better to initialized with `1` (or perhaps `PJ_TRUE`), this is similar to few fields below.

Optionally, while in this area, also update the docs of all `int` members in this class if their default value are not integer, e.g: changing `TRUE` to `1 (yes)`. Thanks.

pjsip/include/pjsua2/account.hpp
```diff
@@ -706,9 +706,27 @@ struct AccountNatConfig : public PersistentObject
...
+      contactRewriteMethod(PJSUA_CONTACT_REWRITE_METHOD),
+      contactUseSrcPort(PJ_TRUE),
+      viaRewriteUse(PJ_TRUE),
+      sdpNatRewriteUse(PJ_FALSE),
+      sipOutboundUse(PJ_TRUE),
+      sipOutboundInstanceId(""),
+      sipOutboundRegId(""),
```
nanangizz, line 727: No need to init string members with empty string (as done in some other string fields)?

nanangizz (approved):
Please also check few minor notes below.

pjsip/include/pjsua2/account.hpp
trandi: `iceNoRtcp` and `iceAlwaysUpdate` are `bool`s 
Good point re `contactRewriteUse` I've updated to `PJ_TRUE`

Have updated the docs to the int fields.

pjsip/include/pjsua2/account.hpp
trandi: Makes sense, removed

sauwming (approved):

Issue #1463, opened 2012-03-09
Separation of PJMEDIA specific implementation to support third party media stack/library in PJSUA-LIB


pjsipbot:
_2012-03-09 11:11:55: @bennylp_ _created the issue on trac ticket [1463](https://trac.pjsip.org/repos/ticket/1463)_

This ticket implements support for integrating third party media stack into PJSUA-LIB. By following these steps below, application can use third party media stack to perform audio and video functionality while still making use of the full SIP, NAT, and security (including SRTP) features provided by PJSUA-LIB API.

By disabling PJMEDIA, the following features will not be available in PJSUA-LIB (unless the equivalent implementation is provided by the third party media library):
 - sound device management
 - echo cancellation
 - codecs
 - jitter buffer
 - RTP and RTCP
 - WAV playback and recording
 - conference bridge
 - DTMF with RFC 2833
 - and so on, except explicitly mentioned below

The following features will still be available:
 - all SIP features, including SIP registration, etc.
 - SDP and SDP negotiation
 - NAT traversal features (including ICE)
 - security features including TLS and SRTP
 - media transports

Follow these steps to integrate third party media library with PJSUA-LIB:

 1. Declare this in `pjlib/include/pj/config_site.h`:
  ```
#define PJSUA_MEDIA_HAS_PJMEDIA    0
  ```
 to exclude PJMEDIA specific implementation from PJSUA-LIB library. Understandably you will loose all media features in PJSUA-LIB (this will be handled by your third party media stack).
 1. Also copy suggested settings from `pjsip-apps/src/3rdparty_media_sample/config_site.h` into `pjlib/include/pj/config_site.h`. These settings are mostly used to exclude unneeded media components from the link process.
 1. Build the libraries, but this time using 
  ```
$ make lib
  ```
  instead of just **`make`** or **`make all`**. This is because most samples will no longer build due to missing media in PJSUA-LIB, hence normal `make` will fail on these apps. The `make lib` command only builds the libraries and unit tests for the libraries.
 1. Go to directory **`pjsip-apps/src/3rdparty_media_sample`**. This is a sample application with hook points to integrate third party media library. Fill in the media implementation in the alt_pjsua_xxx.c files, following the "TODO" notes. Run `make` to build the application. Once it's built, run *alt_pjsua* just as you run the usual *pjsua* application (it's essentially the same app!).

___
_2012-03-09 12:07:47: @bennylp_ _edited the issue description_

___
[...]

PR #2770, opened 2021-07-21
Fix compilation of 3rdparty_media_sample (introduced by #1463)

wosrediinanatour:
The 3rd-party media sample does not compile anymore (following steps in  #1463), since pjsua_app has changed.
This PR fixes the Makefile.


nanangizz (approved):
Just checked the issue and tested the patch. The issue is valid and the patch fixes it.

Thank you.

sauwming (approved):

PR #2666, opened 2021-03-11
Fixed race condition between media event callback and stopping video

sauwming:
This is a more general case of #2645 .

To avoid deadlock, we usually avoid holding locks when calling a callback. But if the callback happens at the same time with a destroy process, it can potentially cause problems as the callback may access objects being, or already, destroyed.

For the case of media event callback and video stoppage, there are a couple of problems found:
1. In pjsua callback
To call `on_call_media_event()` callback, we use a timer (to avoid deadlock):
```
   if (pjsua_var.ua_cfg.cb.on_call_media_event) {
        ...
    	pjsua_schedule_timer2(&call_med_event_cb, eve, 1);
   }
```
If there’s a race with `pjsua_vid_stop_stream()`, it can cause crash/assertion if inside the callback, user tries to access objects destroyed in `pjsua_vid_stop_stream()`, such as video window or stream.
Example:
In `pjsua_vid_stop_stream()`:
```
    /* Decrement ref count of stream video window */
    dec_vid_win(call_med->strm.v.rdr_win_id);
    call_med->strm.v.rdr_win_id = PJSUA_INVALID_ID;
```
In `on_call_media_event()`:
```
    wid = ci.media[med_idx].stream.vid.win_in;
    pjsua_vid_win_get_info(wid, &win_info);
```
One scenario to reproduce: make a video call (you can make more than one calls at the same time to increase the likelihood of the issue happening), then hangup right at the format change event (usually only about 1-2s after making the calls, just before the video is about to show up).

2. pjmedia event callback
pjmedia event callback is called without holding event manager’s lock, so even though `pjmedia_event_unsubscribe()` has been called, the callback can still be in the process of being invoked.
Example:
In `pjmedia_vid_port_destroy()` (some codes are removed for clarity):
```
   pjmedia_event_unsubscribe(NULL, &vidstream_event_cb, vp, vp->strm);
   pjmedia_vid_dev_stream_destroy(vp->strm);
   vp->strm = NULL;
        
   pjmedia_event_unsubscribe(NULL, &client_port_event_cb, vp,
                                  vp->client_port);
   pjmedia_port_destroy(vp->client_port);
   vp->client_port = NULL;
```
So when `client_port_event_cb()` is called, it’s possible `vp->strm` or `vp->client_port` has been destroyed.
The same scenario above can be used to trigger the issue.

The proposed fixes:
[...]

PR #2771, opened 2021-07-22
Avoid deadlock on unsubscribe event

trengginas:
The patch in #2666 ensures that event callback doesn't access to a destroyed object. 
However it introduce new issue.

Scenario:
1) Alice and Bob are in an established video call.
2) Alice stops its video
3) Alice starts its video
4) Alice keeps doing 2 and 3 until 2 freezes

Thread 1 : 
```
- pjsua_vid_preview_stop() > pjsua lock hold
- pjmedia_vid_port_destroy()
- pjmedia_event_unsubscribe() > try to acquire cb_mutex
```
Thread 2 :
```
- video decoding error > raise keyframe request event
- event distribute > cb_mutex  hold
- callback event called, send SIP INFO to request keyframe > try to acquire pjsua lock
```

Thread 2 will not be able to send the keyframe request and Thread 1 will stuck on this loop 
```
    while(1) {
    	pj_mutex_lock(mgr->mutex);
    	if (pj_mutex_trylock(mgr->cb_mutex) == PJ_SUCCESS)
    	    break;
    	
    	/* The callback is currently being called, wait for a while. */
    	pj_mutex_unlock(mgr->mutex);
    	pj_thread_sleep(10);
    }
```

To avoid this deadlock, this patch will unlock pjsua lock when calling ```pjmedia_event_unsubscribe()```.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_vid.c
+++ b/pjsip/src/pjsua-lib/pjsua_vid.c
@@ -34,6 +34,10 @@
 static pjsua_vid_win_id vid_preview_get_win(pjmedia_vid_dev_index id,
                                             pj_bool_t running_only);
 static void free_vid_win(pjsua_vid_win_id wid);
+static pj_status_t media_event_unsubscribe(pjmedia_event_mgr* mgr,
+					   pjmedia_event_cb* cb,
+					   void* user_data,
+					   void* epub);
 
 /*****************************************************************************
  * pjsua video subsystem.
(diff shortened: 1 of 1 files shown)


sauwming:
Now I realize that introducing `cb_mutex` probably wasn't a good idea. As a result, `pjmedia_event_subscribe()` has a requirement that it should not be called while holding any locks (to be more precise, any locks that can be acquired inside the event callback itself).

The workaround in this PR looks okay, but I believe we eventually need to fix `pjmedia_event`.

nanangizz (approved):
Releasing PJSUA lock temporarily may introduce another issue, e.g: some PJSUA objects referred in the function, e.g: `call_med`, may be changed by other threads during the lock release. But then this can be just fine as unsubscribing event usually done in media deinitialization.

sauwming (approved):

Issue #1325, opened 2011-07-14
Signature for pjmedia objects


pjsipbot:
_2011-07-14 05:52:25: @bennylp_ _created the issue on trac ticket [1325](https://trac.pjsip.org/repos/ticket/1325)_

PJMEDIA object will be given object signature. This is needed for the event framework so that application can at least distinguish the sender of the event.

___
_2011-07-14 05:56:09: @bennylp_ _changed **status** from **new** to **closed**_

___
_2011-07-14 05:56:09: @bennylp_ _set **resolution** to **fixed**_

___
_2011-07-14 05:56:09: @bennylp_ _commented_

(In [3640]) Fixed #1325: signature fix pjmedia objects

___
_2011-07-14 07:34:30: @bennylp_ _commented_

(In [3641]) More re #1325 (pjmedia signature): invented pjmedia PJMEDIA_FOURCC macro, and change signature and pjmedia_format_id to use this object. Added pjmedia_fourcc_name( ) too

___
_2011-07-15 03:00:38: @nanangizz_ _commented_

(In [3644]) Re #1325:
 - Added signature for directshow video device.
 - Added signatures.h into pjmedia VS2005 project.

___
_2011-07-15 03:04:04: @mingteluu_ _commented_

(In [3645]) Re #1325:

- Add signature for QT and iOS dev.

PR #2780, opened 2021-07-30
Fix signature checking code

k-marciniak:
#### Reference issues:

None

#### What does this implement/fix?:

Signature checking code is almost unused, but original
version code had some errors in byte checking of subclasses
(no byte masking after bit shifts).

When PJMEDIA_FOURCC was implemented, PJMEDIA_SIGNATURE was changed
to use it, but used wrong order of arguments. Current version uses
original order which now works with class checking macros.

#### Other comments

Those issues cropped up during porting of one application when I've
implemented my own port and needed some checking. Everything worked
on version 1.8.5, but when ported to 2.10, signature checking broke.

Because PJMEDIA_FOURCC and signature checking only uses bit shifts, it should
work on any endianness.

### DIFF
--- a/pjmedia/include/pjmedia/signatures.h
+++ b/pjmedia/include/pjmedia/signatures.h
@@ -112,7 +112,7 @@ PJ_INLINE(const char*) pjmedia_sig_name(pjmedia_obj_sig sig, char buf[])
 /**
  * Macro to generate signature from four ASCII letters.
  */
-#define PJMEDIA_SIGNATURE(a,b,c,d)	PJMEDIA_FOURCC(a,b,c,d)
+#define PJMEDIA_SIGNATURE(a,b,c,d)	PJMEDIA_FOURCC(d,c,b,a)
 
 /*************************************************************************
  * Codec signature ('Cxxx'). Please keep the constant names sorted.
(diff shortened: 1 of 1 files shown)


pjmedia/include/pjmedia/signatures.h
```diff
@@ -169,7 +169,7 @@ PJ_INLINE(const char*) pjmedia_sig_name(pjmedia_obj_sig sig, char buf[])
  * Video ports signatures ('PVxx'). Please keep the constant names sorted.
  */
 #define PJMEDIA_SIG_CLASS_PORT_VID(c,d)	PJMEDIA_SIG_CLASS_PORT('V',c,d)
-#define PJMEDIA_SIG_IS_CLASS_PORT_VID(s) ((s)>>24=='P' && (s)>>16=='V')
+#define PJMEDIA_SIG_IS_CLASS_PORT_VID(s) ((s)>>24=='P' && (((s)>>16)&0xff)='V')
```
sauwming, line 172: missing one `=`

k-marciniak: Thank you, fixed.

sauwming:
> When PJMEDIA_FOURCC was implemented, PJMEDIA_SIGNATURE was changed
> to use it, but used wrong order of arguments. Current version uses
> original order which now works with class checking macros.

Just for our reference. It happened here:
https://github.com/pjsip/pjproject/issues/1325
https://trac.pjsip.org/repos/ticket/1325

nanangizz (approved):

sauwming (approved):

PR #2781, opened 2021-07-30
Added pjsip_tsx_set_timers to change timers at runtime

k-marciniak:
Added new function pjsip_tsx_set_timers in sip_transaction.c
which allows to change session timers during runtime.
It also allows to change timer values independently,
currently all timers are set at various ratios from
t1 during init. This was required for server which could
change timeout configuration on runtime, but could be
usable in other projects.

### DIFF
--- a/pjsip/include/pjsip/sip_transaction.h
+++ b/pjsip/include/pjsip/sip_transaction.h
@@ -437,7 +437,8 @@ PJ_DECL(pj_status_t) pjsip_tsx_set_timeout(pjsip_transaction *tsx,
 
 /*
  * Change timer values used by transaction layer. Currently scheduled
- * timers will not be changed.
+ * timers will not be changed. Any value set to 0 will be left
+ * unchanged.
  *
  * @param t1 - Transaction T1 timeout, in msec
  * @param t2 - Transaction T2 timeout, in msec
@@ -445,6 +446,11 @@ PJ_DECL(pj_status_t) pjsip_tsx_set_timeout(pjsip_transaction *tsx,
  * @param td - Transaction completed timer for INVITE, in msec
  */
 PJ_DECL(void) pjsip_tsx_set_timers(unsigned t1, unsigned t2, unsigned t4, unsigned td);
+
+/*
+ * (Re)Initializes timer values from `pjsip_cfg()`.
+ */
+PJ_DECL(void) pjsip_tsx_initialize_timer_values(void);
 
 /**
  * Get the transaction instance in the incoming message. If the message
--- a/pjsip/include/pjsip/sip_transaction.h
+++ b/pjsip/include/pjsip/sip_transaction.h
@@ -435,6 +435,16 @@ PJ_DECL(pj_status_t) pjsip_tsx_stop_retransmit(pjsip_transaction *tsx);
 PJ_DECL(pj_status_t) pjsip_tsx_set_timeout(pjsip_transaction *tsx,
 					   unsigned millisec);
 
+/*
+ * Change timer values used by transaction layer. Currently scheduled
+ * timers will not be changed.
+ *
+ * @param t1 - Transaction T1 timeout, in msec
+ * @param t2 - Transaction T2 timeout, in msec
+ * @param t4 - Transaction completed timer for non-INVITE, in msec
+ * @param td - Transaction completed timer for INVITE, in msec
+ */
+PJ_DECL(void) pjsip_tsx_set_timers(unsigned t1, unsigned t2, unsigned t4, unsigned td);
 
 /**
  * Get the transaction instance in the incoming message. If the message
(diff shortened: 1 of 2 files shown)


nanangizz:
Yes, the ability to change tsx timeout timer at run time can be useful, e.g: in IP change scenario, some app need to make use a shorter than usual timeout `REGISTER` tsx to detect if SIP network path has been reconnected. However, this approach may introduce inconsistencies with pjsip config settings (i.e: `pjsip_cfg()->tsx.t1/t2/t4/td`).

As the tsx timer settings in pjsip config are currently configurable already (by modifying `pjsip_cfg()->tsx.t1/t2/t4/td` directly), but it misses the capability to apply the re-configured tsx timer settings to the actual run-time settings (i.e: static vars `t1/t2/t4/td/timeout_timer_val`), perhaps the function should just apply/re-calculate the actual tsx timer settings from pjsip config. Or make the function `pjsip_tsx_set_timeout()` to also modify pjsip settings (also, it would be nice if app can modify partial settings only, for example modifying tsx timeout setting only while leaving t1, etc as is (e.g: '0' value means leaving it as is) ).

wosrediinanatour:
What happen, if the timers are set (or configured) with an ongoing session/transaction? Does it interfere with?

k-marciniak:
When timers are set through config, currently at runtime nothing changes, config values in milliseconds are used to initialize static timer values (`t1_timer_val` etc.) when you call `pjsip_tsx_layer_init_module`. If you call `pjsip_tsx_set_timers`, those values are reinitialized and used in any newly instantiated timers. Currently running timers expire with original times. Function uses mutex when setting values but maybe I should add locking in `tsx_on_state_proceeding_uac` when they are used?

I will change this implementation, but I would like to know which options you would consider as best architecturally:

1. remove direct collective setting of timer though `pjsip_tsx_set_timers`
2. add setting `pjsip_config` values from `pjsip_tsx_set_timers`
3. add function `pjsip_tsx_config_timers(pjsip_cfg_t *config)` which will recalc timer values (function name probably could be better )
4. add functions like `pjsip_tsx_set_t1_timer(unsigned t1)` for each time value

I propose 3 and 4, because setting pjsip_config from module which only used this config one way seems a little backwards and setting all timers in one go is nice, because that functions require mutex.

nanangizz:
> Function uses mutex when setting values but maybe I should add locking in `tsx_on_state_proceeding_uac` when they are used?

If this is about adding locking using `mod_tsx_layer.mutex`, please no, it would degrade the library performance a lot.

Actually I am not sure how locking `mod_tsx_layer.mutex` in `pjsip_tsx_set_timers` can be useful, as `mod_tsx_layer.mutex` is used for tsx instance registration & search, the mutex is not really used in the tsx operations where the timers are set, tsx operations use tsx mutex instead.

> I will change this implementation, but I would like to know which options you would consider as best architecturally:
> 
> 1. remove direct collective setting of timer though `pjsip_tsx_set_timers`

Not really sure what this mean, but setting all timers t1, t2, etc at once (using a single function) should be fine.

Or if this means that `pjsip_tsx_set_timers` should not take any param and take the `t1/t2/etc` settings from `pjsip_cfg()->tsx.t1/t2/t4/td`, then this is okay for me.

Note that my main concern is to make sure the static vars `t1/t2/t4/td/timeout_timer_val` are sync'ed to the global pjsip config settings `pjsip_cfg()->tsx.t1/t2/t4/td`. Also note that some tsx timers are set using `pjsip_cfg()->tsx.t1/t2/t4/td` directly instead of using the static vars `t1/t2/t4/td/timeout_timer_val`, e.g: [here](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_transaction.c#L2374). So I think synchronizing the timers setting in the pjsip global settings and the timer static vars are mandatory.

> 2. add setting `pjsip_config` values from `pjsip_tsx_set_timers`

This option is okay too. To make sure we are on the same page, this will mean adding lines such as `pjsip_cfg()->tsx.t1 = t1` inside `pjsip_tsx_set_timers`, so both settings (pjsip global and static vars) are synchronized.

> 3. add function `pjsip_tsx_config_timers(pjsip_cfg_t *config)` which will recalc timer values (function name probably could be better )

Not sure about the `config` param there, I guess this is not related to my concern above.

> 4. add functions like `pjsip_tsx_set_t1_timer(unsigned t1)` for each time value

This is actually okay too (as long as it also update the pjsip global setting), so app can set only some timer settings while leaving other settings as is. Another approach is, as mentioned before, add a specification to the function that when any param `t1/t2/etc` is set to zero, the corresponding setting will not be modified.

k-marciniak:
After comments, current version:

* `pjsip_tsx_set_timers` updates `pjsip_cfg()` values and accepts zero as "no change".
* add `pjsip_tsx_initialize_timer_values` which sets values from `pjsip_cfg()`
* remove mutexes

Previously suggested function `pjsip_tsx_config_timers` would accept pointer returned from `pjsip_cfg()` or to another struct with timer values (alternative config), but because it must be synchronized with original `pjsip_cfg()`, I've just made  function which doesn't accept any pointer and just sets from original config. I've also changed `pjsip_tsx_layer_init_module` to use that function, so implementation is in one place.

Proposition 4 was not implemented, because `pjsip_tsx_set_timers` can already do this anyway.

nanangizz (approved):

trengginas (approved):

sauwming (approved):

Issue #1726, opened 2014-01-29
Assertion in ICE connectivity check (thanks Amit Chowdhary for the report)


pjsipbot:
_2014-01-29 06:28:27: @nanangizz_ _created the issue on trac ticket [1726](https://trac.pjsip.org/repos/ticket/1726)_

This seems to be related to changeset [r4652](../commit/f68beb968246ef1ce125d836a88243a9fa96803b), timer was always cancelled, but now it is not when mod_count==TRUE.

Reproducing step:
 1. run 2 pjsua with `--use-ice` param, no STUN/TURN
 2. make call from one to another, and just accept the call on the receiving side
 3. assertion happens on both sides once ICE connectivity check is started.

Assertion line:
```
Assertion failed: (tsx->retransmit_timer.id == 0 || !tsx->require_retransmit), function tsx_transmit_msg, file ../src/pjnath/stun_transaction.c, line 204.
```
___
_2014-01-29 06:33:39: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2014-01-29 06:33:39: @nanangizz_ _set **resolution** to **fixed**_

___
_2014-01-29 06:33:39: @nanangizz_ _commented_

In [r4717](../commit/bd187283ee845184fd47278399093238e5bb7a49):
Fix #1726: Added check in assertion line for transmitting STUN message without incrementing retransmission count.

PR #2782, opened 2021-08-02
Reset ICE flag when re-using srtp transport

SamVanBelle:
Follow-up of #2726
Reset ICE flag when re-using srtp transport to allow for remote address update in dtls_on_rcv_rtp.


sauwming:
I tested it here with the same scenario as #1726 (i.e. re-INVITE without SDP to restart ICE) and it worked fine.

nanangizz (approved):

sauwming (approved):

PR #2783, opened 2021-08-04
Add checks and assertion if PJSIP string length is negative

sauwming:
Even though PJSIP string is not supposed to have negative length ever, a runtime bug or improper initialization can cause this to happen. And it may produce issues such as memory overwrite or crash, since the string implementation rarely checks for strings with negative length.

In this PR, we will add handling for string inputs that have negative length, as well as add assertion to allow apps to catch this issue during debugging/development mode.

### DIFF
--- a/pjlib/include/pj/string_i.h
+++ b/pjlib/include/pj/string_i.h
@@ -147,7 +147,7 @@ PJ_IDEF(pj_str_t*) pj_strncpy_with_null( pj_str_t *dst, const pj_str_t *src,
     pj_assert(max > 0);
 
     if (max <= src->slen)
-	max = max-1;
+	max = (max > 0)? max-1: 0;
     else
 	max = (src->slen < 0)? 0: src->slen;
 
--- a/pjlib/include/pj/string_i.h
+++ b/pjlib/include/pj/string_i.h
@@ -33,15 +33,17 @@ PJ_IDEF(pj_str_t*) pj_strdup(pj_pool_t *pool,
 			      pj_str_t *dst,
 			      const pj_str_t *src)
 {
+    pj_assert(src->slen >= 0);
+
     /* Without this, destination will be corrupted */
     if (dst == src)
 	return dst;
 
-    if (src->slen) {
+    if (src->slen > 0) {
 	dst->ptr = (char*)pj_pool_alloc(pool, src->slen);
 	pj_memcpy(dst->ptr, src->ptr, src->slen);
     }
-    dst->slen = src->slen;
+    dst->slen = (src->slen < 0)? 0: src->slen;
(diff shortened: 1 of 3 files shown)


pjlib/include/pj/string_i.h
```diff
@@ -118,25 +130,29 @@ PJ_IDEF(pj_str_t*) pj_strcpy2(pj_str_t *dst, const char *src)
...
     return dst;
 }
 
 PJ_IDEF(pj_str_t*) pj_strncpy_with_null( pj_str_t *dst, const pj_str_t *src,
 					 pj_ssize_t max)
 {
+    pj_assert(src->slen >= 0);
     pj_assert(max > 0);
 
     if (max <= src->slen)
```
nanangizz, line 149: max < 0 case is not handled yet?

pjlib/src/pj/string.c
```diff
@@ -143,6 +148,8 @@ PJ_DEF(char*) pj_strstr(const pj_str_t *str, const pj_str_t *substr)
 {
     const char *s, *ends;
 
+    PJ_ASSERT_RETURN(str->slen >= 0 && substr->slen >= 0, NULL);
+
     /* Special case when substr is zero */
     if (substr->slen == 0) {
```
nanangizz, line 154: - change condition to `substr->slen <=0 || substr->slen > str->slen` ?
- also check for `str->slen <= 0`?

pjlib/src/pj/string.c
```diff
@@ -162,6 +169,8 @@ PJ_DEF(char*) pj_stristr(const pj_str_t *str, const pj_str_t *substr)
...
```
nanangizz, line 175: same as above

pjlib/src/pj/string.c
sauwming: ah, I keep forgetting that `PJ_ASSERT_RETURN` doesn't translate to `return` if `PJ_ENABLE_EXTRA_CHECK` is disabled.

ok, added check for `if (str->slen <= 0) return NULL;`.  There will be a minor behavior change though: `pj_strstr(empty string, empty string)` will now return NULL; whereas previously it will return `str->ptr`. But I feel the `NULL` return should be better since `str->ptr` may point to some arbitrary memory for empty string.

for the case `substr->slen > str->slen`, I don't think it's correct to return `str->ptr` since the substring clearly cannot exist in the string. The current behavior is already correct, to return `NULL`. If `substr->slen > str->slen`, `ends` will point to memory outside the string buffer, but it should be harmless since no memory access will be performed.

nanangizz (approved):

trengginas (approved):

PR #2785, opened 2021-08-04
Adds support for Digest AKA Authentication in PJSUA2

wosrediinanatour:
Similar to PJSUA, PJSIP_HAS_DIGEST_AKA_AUTH has to be defined. It then calls the responsible callback, if a SIP 401 is received as response of a REGISTER.

### DIFF
--- a/pjsip/include/pjsua2/endpoint.hpp
+++ b/pjsip/include/pjsua2/endpoint.hpp
@@ -437,6 +437,150 @@ struct OnMediaEventParam
     MediaEvent      ev;
 };
 
+/**
+ * This structure describes authentication challenge used in Proxy-Authenticate
+ * or WWW-Authenticate for digest authentication scheme.
+ */
+struct DigestChallenge
+{
+    /**
+     * Realm for the challenge.
+     * */
+    std::string realm;
+
+    /*
+     * Other parameters.
+     * */
+    StringToStringMap otherParam;
+
+    /**
+     * Domain.
+     */
+    std::string domain;
+
+    /**
+     * Nonce challenge.
+     */
+    std::string nonce;
+
+    /**
+     * Opaque value.
+     */
+    std::string opaque;
+
+    /**
+     * Stale parameter.
+     */
+    int  stale;
+
+    /**
+     * Algorithm parameter.
+     */
+    std::string algorithm;
+
+    /**
+     * Quality of protection.
+     */
+    std::string qop;
+
+    /**
+     * Convert from pjsip
+     */
+    void fromPj(const pjsip_digest_challenge &prm);
+
+    /**
+     * Convert to pjsip
+     */
+    pjsip_digest_challenge toPj() const;
+};
+
+/**
+ * This structure describe credential used in Authorization and
+ * Proxy-Authorization header for digest authentication scheme.
+ */
+struct DigestCredential
+{
+    /**
+     *Realm of the credential
+     */
+    std::string realm;
+
+    /**
+     *Other parameters.
+     */
+    StringToStringMap otherParam;
+
+    /**
+     *Username parameter.
+     */
+    std::string username;
+
+    /**
+     *Nonce parameter.
+     */
+    std::string nonce;
+
+    /**
+     *URI parameter.
+     */
+    std::string uri;
+
+    /**
+     *Response digest.
+     */
+    std::string response;
+
+    /**
+     *Algorithm.
+     */
+    std::string algorithm;
+
+    /**
+     *Cnonce.
+     */
+    std::string cnonce;
+
+    /**
+     *Opaque value.
+     */
+    std::string opaque;
+
+    /**
+     *Quality of protection.
+     */
+    std::string qop;
+
+    /**
+     *Nonce count.
+     */
+    std::string nc;
+
+    /**
+     * Convert from pjsip
+     */
+    void fromPj(const pjsip_digest_credential &prm);
+
+    /**
+     * Convert to pjsip
+     */
+    pjsip_digest_credential toPj() const;
+};
+
+
+/**
+ * Parameters for onCredAuth account method.
+ */
+struct OnCredAuthParam
+{
+    DigestChallenge digestChallenge;
+
+    AuthCredInfo credentialInfo;
+
+    std::string method;
+
+    DigestCredential digestCredential;
+};
+
 //////////////////////////////////////////////////////////////////////////////
 /**
  * SIP User Agent related settings.
@@ -1606,8 +1750,6 @@ public:
      *    (configurable via \a AccountConfig.ipChangeConfig.reinviteFlags).
      *
      * @param param	The IP change parameter, have a look at #IpChangeParam.
-     *
-     * @return		PJ_SUCCESS on success, other on error.
      */
     void handleIpChange(const IpChangeParam &param) PJSUA2_THROW(Error);
 
@@ -1695,6 +1837,39 @@ public:
      */
     virtual void onMediaEvent(OnMediaEventParam &prm)
     { PJ_UNUSED_ARG(prm); }
+
+    /**
+     * Callback for computation of the digest credential.
+     *
+     * Usually, an application does not need to implement (overload) this callback.
+     * Use it, if your application needs to support Digest AKA authentication without 
+     * the default digest computation back-end (i.e: using <b>libmilenage</b>).
+     *
+     * To use Digest AKA authentication, add \a PJSIP_CRED_DATA_EXT_AKA flag in the
+     * AuthCredInfo's \a dataType field of the AccountConfig, and fill up other
+     * AKA specific information in AuthCredInfo:
+     *  - If PJSIP_HAS_DIGEST_AKA_AUTH is disabled, you have to overload this callback
+     *    to provide your own digest computation back-end.
+     *  - If PJSIP_HAS_DIGEST_AKA_AUTH is enabled, <b>libmilenage</b> library from 
+     *    \a third_party directory is linked, and this callback returns PJ_ENOTSUP,
+     *    then the default digest computation back-end is used.
+     *
+     * @param prm.digestChallenge	The authentication challenge sent by server in 401
+     *		    or 401 response, as either Proxy-Authenticate or
+     *		    WWW-Authenticate header.
+     * @param prm.credentialInfo	    The credential to be used.
+     * @param method    The request method.
+     * @param prm.digestCredential	    The digest credential where the digest response
+     *		    will be placed to. Upon calling this function, the
+     *		    nonce, nc, cnonce, qop, uri, and realm fields of
+     *		    this structure must have been set by caller. Upon
+     *		    return, the \a response field will be initialized
+     *		    by this function.
+     *
+     * @return PJ_ENOTSUP is the default. If you overload this callback,
+     *		    return PJ_SUCCESS on success. 
+     */
+    virtual pj_status_t onCredAuth(OnCredAuthParam &prm);
 
 private:
     static Endpoint		*instance_;	// static instance
@@ -1869,6 +2044,15 @@ private:
     on_ip_change_progress(pjsua_ip_change_op op,
 			  pj_status_t status,
 			  const pjsua_ip_change_op_info *info);
+
+    static pj_status_t on_auth_create_aka_response_callback(
+					     pj_pool_t *pool,
+					     const pjsip_digest_challenge*chal,
+					     const pjsip_cred_info *cred,
+					     const pj_str_t *method,
+					     pjsip_digest_credential *auth);
+    friend class Account;
+
 
 private:
     void clearCodecInfoList(CodecInfoVector &codec_list);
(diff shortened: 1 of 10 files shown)


wosrediinanatour:
Does somebody understand, why one build failed? ( C/C++ CI / build-windows-default (pull_request) Failing after 4m — build-windows-default )

sauwming:
Don't worry about the build failure. The CI machine is under a very heavy load, so sometimes it can fail for an unrelated cause.

pjsip/src/pjsua2/account.cpp
```diff
@@ -595,6 +595,9 @@ void AccountConfig::toPj(pjsua_acc_config &ret) const
 	dst->ext.aka.k	= str2Pj(src.akaK);
 	dst->ext.aka.op	= str2Pj(src.akaOp);
 	dst->ext.aka.amf= str2Pj(src.akaAmf);
+#if PJSIP_HAS_DIGEST_AKA_AUTH
+	dst->ext.aka.cb = &pjsip_auth_create_aka_response;
```
sauwming, line 599: toPj() is supposed to just do conversion, it shouldn't do a new assignment.

You should add a new field akaCb, then here in toPj() will be just:
```
dst->ext.aka.cb = src.akaCb;
```

sauwming:
I don't think the latest revision will work since `src` is of type `AuthCredInfo` and it has no field akaCb.

It should be something like this:
[pjsua2_aka_patch.txt](https://github.com/pjsip/pjproject/files/6942271/pjsua2_aka_patch.txt)

I also provide an example on how to use it in an application, in pjsua2_demo.cpp.
There's still one problem though, this may only work in C++.

wosrediinanatour:
I have to put this PR on hold - since I cannot test it at the moment.

wosrediinanatour:
I want to continue with this PR but am struggling on the build system...

For the PJSUA (C) part without your patch, building for DIGEST_AKA_AUTH worked only by 
* reverting https://trac.pjsip.org/repos/changeset/4720 via a patch
* Setting `C(XX)FLAGS*  to `-DPJSIP_HAS_DIGEST_AKA_AUTH=1` 

@sauwming Is there any easier solution?

sauwming:
In `pjlib/include/pj/config_site.h`, add:
`#define PJSIP_HAS_DIGEST_AKA_AUTH   1`

In `user.mak`, in PJSIP root directory, add:
`export LDFLAGS += -lmilenage-arm-apple-darwin`
(it may be different in your system, depending on the library name)

wosrediinanatour:
> I don't think the latest revision will work since `src` is of type `AuthCredInfo` and it has no field akaCb.
> 
> It should be something like this: [pjsua2_aka_patch.txt](https://github.com/pjsip/pjproject/files/6942271/pjsua2_aka_patch.txt)
> 


I updated this PR by reworking the commit following your patch.
For testing, I used `pjsua2_demo.cpp` except our credencials/URIs:
```

21:43:24.217            pjsua_acc.c  ..Acc 0: Registration sent
...
21:43:24.505           pjsua_core.c  .RX 825 bytes Response msg 401/REGISTER/cseq=61044 (rdata0x7ff420000b88) from UDP 10.17.20.200:5060:
SIP/2.0 401 Unauthorized - Challenging the UE
...
21:43:24.505      sip_auth_client.c  ...Digest algorithm is "AKAv1-MD5"
21:43:24.505           pjsua_core.c  ....TX 1017 bytes Request msg REGISTER/cseq=61045 (tdta0x113e348) to UDP 10.17.20.200:5060:
...
21:43:24.546           pjsua_core.c  .RX 406 bytes Response msg 100/REGISTER/cseq=61045 (rdata0x7ff420000b88) from UDP 10.17.20.200:5060:
...
21:43:24.672           pjsua_core.c  .RX 1015 bytes Response msg 200/REGISTER/cseq=61045 (rdata0x7ff420000b88) from UDP 10.17.20.200:5060:
...
21:43:24.673            pjsua_acc.c  ....sip:USER@DOMAIN: registration success, status=200 (OK), will re-register in 270 seconds
21:43:24.673            pjsua_acc.c  ....Keep-alive timer started for acc 0, destination:10.17.20.200:5060, interval:15s
*** Register: code=200

```

Is it possible to use some pjsip.org credencial to have a working test case?
(btw: I followed the used spaces/tabs to be consistent with current code. But it is generelly not consistent in pjsua2_demo.cpp)

wosrediinanatour:
> Currently the PJSIP server is for internal testing purposes only, but you can probably find a test server by Googling around.


I asked, because the credencials for in the psjua2_demo.cpp seems to work (without Digest AKA) and thought, maybe it would be possible for Digest AKA as well for internal tests.

> One important thing is how to make this work on other languages, such as Java and Python as well.

At the moment, using PJSIP_HAS_DIGEST_AKA_AUTH for JAVA/Python does not work at all, so my commit does not change anything in this regard.

As far as I understand,  current commit only calls a PJSIP callback in case of a `SIP/2.0 401 Unauthorized - Challenging the UE` responses, which then will retransmit a modified SIP REGISTER. When its response comes back then PJSUA2's `onRegState(...)` is called as usual. Hence, I think there is nothing needed for JAVA/Python, which are based on PJSUA2.

wosrediinanatour:
How can we proceed? From my point of view, the commit is ok.

nanangizz:
Hi, let me join the discussion.

SWIG ports (e.g: Python & Java) need to be covered indeed, so here are perhaps the alternatives:

- The callback is hardcoded to `pjsip_auth_create_aka_response()`, so app doesn't really need to specify callback in PJSUA2. For example when `PJSIP_CRED_DATA_EXT_AKA` is set in data type, set the callback too, e.g:
   ```
   AccountConfig::toPj()
   {
      ...
      if (src.dataType & PJSIP_CRED_DATA_EXT_AKA) {
         dst->ext.aka.cb= &pjsip_auth_create_aka_response;
      }
  }
  ```
   IMO this approach should be okay as it can be extended later to a proper callback solution (see below) whenever needed.

- Proper callback: to allow callback redirection in Python & Java, the callback needs to be set as a (virtual) method of a class, for example `Endpoint::onCredAuth(OnCredAuthParam &prm)` (please check other callbacks for reference), it may need some new PJSUA2 classes, e.g: for `pjsip_digest_challenge` and `pjsip_digest_credential`.

wosrediinanatour:
@nanangizz, I did similar to Alternative 1, but then @sauwming noted that "toPj() is supposed to just do conversion, it shouldn't do a new assignment." This sound reasonable, so I reworked my pull request.

Alternative 2 is no alternative for me, only an add-on: 
 -  Having a callback in PJSUA2 is always good. But what would be the goal? Since it makes no sense to do the "second" [1] SIP REGISTER and computations by the application, I think, the callback would be interesting for reporting only.
 - Without PJSIP_HAS_DIGEST_AKA_AUTH defined, there is already Callback onRegState()  called reporting SIP 401 [1]. So it may make more sense to do a re-register from there. But again: I don't think that anyone wants to implement it on application level.

My conclusion: PJSUA does not need some communication with application layer. Hence, PJSUA2, does not need it, and hence, Java & Python do not need it.

[1] "Standard ping-pong": SIP REGISTER -> SIP 401 -> SIP REGISTER -> SIP 200 OK

nanangizz:
> @nanangizz, I did similar to Alternative 1, but then @sauwming noted that "toPj() is supposed to just do conversion, it shouldn't do a new assignment." This sound reasonable, so I reworked my pull request.

True, but for this workaround (until the callback is properly supported), I think that's okay to make an exception.

> * Having a callback in PJSUA2 is always good. But what would be the goal? Since it makes no sense to do the "second" [1] SIP REGISTER and computations by the application, I think, the callback would be interesting for reporting only.

The goal of the callback should be to allow app using an alternative backend (other than `libmilenage`) for the auth calculation.

> My conclusion: PJSUA does not need some communication with application layer. Hence, PJSUA2, does not need it, and hence, Java & Python do not need it.

PJSUA app needs to set the callback actually, e.g: [here](https://github.com/pjsip/pjproject/blob/master/pjsip-apps/src/pjsua/pjsua_app_config.c#L904).

wosrediinanatour:
I think I now understand better what's your concern. Will try to incorporate.

wosrediinanatour:
Dear all! 
I reworked the whole pull request. Now there is a new Callback `Endpoint::onCredAuth(OnCredAuthParam &prm)`. The user can decide, if (s)he wants to use the libmilenage and can use a new `Endpoint::createAkaResponse(OnCredAuthParam &prm)`.

pjsip/include/pjsua2/endpoint.hpp
```diff
@@ -1870,6 +2063,15 @@ class Endpoint
 			  pj_status_t status,
 			  const pjsua_ip_change_op_info *info);
 
+    static pj_status_t pjsip_auth_create_aka_response_callback(
```
nanangizz, line 2066: minor: most callbacks in Endpoint has `on_` prefix, so perhaps better to rename  `pjsip_auth_create_aka_response_callback` --> `on_auth_create_aka_response()`?

pjsip/src/pjsua2/endpoint.cpp
```diff
@@ -583,6 +663,12 @@ void Endpoint::utilLogWrite(LogEntry &entry)
     }
 }
 
+pj_status_t Endpoint::onCredAuth(OnCredAuthParam &prm)
+{
+    PJ_UNUSED_ARG(prm);
+    return PJ_FALSE;
```
nanangizz, line 669: Return type is `pj_status_t`, better returning `PJ_ENOTSUP` instead of `PJ_FALSE`. Also note that: `PJ_FALSE == PJ_SUCCESS == 0 `.

pjsip/src/pjsua2/endpoint.cpp
```diff
@@ -2493,3 +2579,63 @@ void Endpoint::handleIpChange(const IpChangeParam &param) PJSUA2_THROW(Error)
...
+	    pj_strdup(pool, &auth->response, &auth_new.response);
+	    pj_strdup(pool, &auth->algorithm, &auth_new.algorithm);
+	    pj_strdup(pool, &auth->cnonce, &auth_new.cnonce);
+	    pj_strdup(pool, &auth->opaque, &auth_new.opaque);
+	    pj_strdup(pool, &auth->qop, &auth_new.qop);
+	    pj_strdup(pool, &auth->nc, &auth_new.nc);
+	    pjsip_param_clone(pool, &auth->other_param, &auth_new.other_param);
+    }
+    return status;
+}
```
nanangizz, line 2641: For simplicity (specifical for PJSUA2), what if we directly call `pjsip_auth_create_aka_response()` from this PJSIP callback when `onCredAuth()` returning `PJ_ENOTSUP` (as in base implementation), so we can omit `createAkaResponse()`. Need to specify this behavior in the `onCredAuth()` docs/spec though.

pjsip/include/pjsua2/endpoint.hpp
```diff
@@ -437,6 +437,150 @@ struct OnMediaEventParam
...
+{
+    /**
+     *Realm of the credential
+     */
+    std::string realm;
+
+    /**
+     *Other parameters.
+     */
+    std::map<std::string, std::string> otherParam;
```
nanangizz, line 511: re: `std::map<std::string, std::string>`, for SWIG interfacing purpose, better to use predefined type (similar to `StringVector`) in `types.hpp`, e.g:
```
/** Map string to string */
typedef std::map<std::string, std::string> StringToStringMap;
```

and then in `pjsip-apps/src/swig/pjsua.i`, add (see lines started with `+`):
```
//
// STL stuff.
//
%include "std_string.i"
%include "std_vector.i"
+ %include "std_map.i"

%template(StringVector)			std::vector<std::string>;
%template(IntVector) 			std::vector<int>;
+ %template(StringToStringMap)		map<string, string>;
```

nanangizz (commented):
Wow, a proper callback! Thanks :)

Just a few comments below, feel free to discuss :)

pjsip/src/pjsua2/endpoint.cpp
wosrediinanatour: So opt-out instead of opt-in. Sounds good.

wosrediinanatour:
Thank you for the good review!
I modified the commit accordingly.

pjsip/include/pjsua2/endpoint.hpp
```diff
@@ -1696,6 +1838,33 @@ class Endpoint
     virtual void onMediaEvent(OnMediaEventParam &prm)
     { PJ_UNUSED_ARG(prm); }
 
+    /**
+     * Callback for computation of the digest credential.
```
nanangizz, line 1842: Perhaps a bit more info about when app needs to implement this callback, for example:
```Normally application does not need to implement this callback. Currently, this callback needs to be implemented only if application needs to support Digest AKA auth AND does not want to use the default digest computation backend (i.e: using libmilenage)```.

nanangizz (approved):
Looks okay. Just a minor (about docs) below.

pjsip/include/pjsua2/endpoint.hpp
wosrediinanatour: Updated commit by an improved comment.

sauwming (approved):

wosrediinanatour:
@trengginas, @nanangizz: Could you take a look?

What is the next step? I see that some PRs have label, some not (like this). Is there a work flow somewhere, which I have missed?

PR #2786, opened 2021-08-05
Add minimum interval for keyframe missing events in Openh264

sauwming:
Currently when there are consecutive decode failures, Openh264 will keep generating missing keyframe events. This can be problematic because there may not be enough time for the previous event to be processed (i.e. we haven't event sent out a keyframe request to remote), and for the remote to respond and send the requested keyframe. In the worst case, the event queue can be fully filled and it will be forced to discard any new event:
```
08:15:04.655                event.c  Lost event IFRM from publisher [0x0x7ff7f80dfeb8] due to full queue.
08:15:04.706                event.c !Lost event IFRM from publisher [0x0x7ff7f79c1cb8] due to full queue.
08:15:04.740                event.c !Lost event IFRM from publisher [0x0x7ff7f79c1cb8] due to full queue.
08:15:04.803                event.c  Lost event FMCH from publisher [0x0x7ff7ff02b378] due to full queue.
```

### DIFF
--- a/pjmedia/src/pjmedia-codec/openh264.cpp
+++ b/pjmedia/src/pjmedia-codec/openh264.cpp
@@ -59,11 +59,11 @@
 /* OpenH264 default PT */
 #define OH264_PT                PJMEDIA_RTP_PT_H264
 
-/* Minimum interval between generating two missing keyframe events.
+/* Minimum interval (in msec) between generating two missing keyframe events.
  * This is to avoid sending too many events during consecutive decode
  * failures.
  */
-#define EVENT_KF_MIN_INTERVAL	15
+#define MISSING_KEYFRAME_EV_MIN_INTERVAL	1000
 
 /*
  * Factory operations.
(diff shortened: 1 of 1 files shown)


nanangizz (commented):
As perhaps the usual/practical video FPS for VoIP is ranging between 5 to 30 fps, having interval setting in frames unit may result in a quite broad range of interval in time unit, e.g: 3 to 0.5 seconds. What if this setting is in millisecond unit instead?

Additional info which may be related, in PJSUA there is `PJSUA_VID_REQ_KEYFRAME_INTERVAL` for sending SIP INFO interval and `PJMEDIA_RTCP_FB_INTERVAL` for sending RTCP-FB interval (including PLI). And for encoder side, there is `PJMEDIA_VID_STREAM_MIN_KEYFRAME_INTERVAL_MSEC` to avoid rapid keyframe generations.

nanangizz (approved):

trengginas (approved):

BaybaraEvgeny:
Actually it happens for CB event too!

PR #2791, opened 2021-08-11
Add AAC Codec

dickytamara:
add AAC stream support via fdk-aac library


dickytamara:
Test with pjsua
--stereo --clock-rate=48000 
--snd-clock-rate=48000 
--no-vad 
--ec-tail=0

and 
# define PJMEDIA_CONF_USE_AGC 0

PR #2793, opened 2021-08-12
Allow macos to share from USB camera

sean-horton:
Goal: Allow macOS to share video from attached USB web cams. Three things were done. 

1. Add `AVCaptureDeviceTypeExternalUnknown` when fetching cameras from the system. This will allow pjsip to see USB cameras.
2. Moved contents of `darwin_factory_init` into `darwin_factory_refresh`. As such, `init` now calls `refresh`. This allows pjsip to see added/removed USB camera devices when refreshing devs. 
3. I noticed while sharing a USB camera into a call and unplugging the camera, pjsip would throw an assert. This is because `session_runtime_error` is calling `PJ_LOG` from a non-registered thread in `NSNotificationCenter`. I've commented the LOG statement, but maybe there is a better option to handle this.

### DIFF
--- a/pjmedia/src/pjmedia-videodev/darwin_dev.m
+++ b/pjmedia/src/pjmedia-videodev/darwin_dev.m
@@ -286,7 +286,7 @@ static pj_status_t darwin_factory_destroy(pjmedia_vid_dev_factory *f)
 /* API: refresh the list of devices */
 static pj_status_t darwin_factory_refresh(pjmedia_vid_dev_factory *f)
 {
-struct darwin_factory *qf = (struct darwin_factory*)f;
+    struct darwin_factory *qf = (struct darwin_factory*)f;
     struct darwin_dev_info *qdi;
     unsigned i, l, first_idx, front_idx = -1;
     enum { MAX_DEV_COUNT = 8 };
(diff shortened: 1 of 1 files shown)


trengginas:
Thanks for the patch!

> 3. I noticed while sharing a USB camera into a call and unplugging the camera, pjsip would throw an assert. This is because `session_runtime_error` is calling `PJ_LOG` from a non-registered thread in `NSNotificationCenter`. I've commented the LOG statement, but maybe there is a better option to handle this.

Does the `session_runtime_error` called from the same thread each time?
If so, you can add checks and register the thread like [here](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-videodev/sdl_dev.c#L334)

sauwming:
Alternatively, replacing it with NSLog() should be fine, IMO, similar to here:
https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-videodev/ios_opengl_dev.m#L171

The downside is that this log won't be output to the PJSIP log file.

sean-horton:
Adding the below to `session_runtime_error` does resolve the crash and I see PJ_LOG log successfully. From my testing each time `session_runtime_error` is called it seems to be the same thread from `NSNotificationCenter`.

Does this look appropriate for registering the thread?
```
    if (!pj_thread_is_registered()) {
        pj_thread_t *thread;
        static pj_thread_desc thread_desc;
        pj_bzero(thread_desc, sizeof(pj_thread_desc));
        pj_thread_register("NSNotificationCenter", thread_desc, &thread);
    }
```

sauwming (approved):

nanangizz (approved):

trengginas (approved):

PR #2797, opened 2021-08-13
Replace deprecated Android Camera framework with Camera2

nanangizz:
Camera2 is the latest Android camera framework API that replaces the deprecated camera framework libraries, as described [here](https://developer.android.com/training/camera2). This framework is added in API level 21 (or Android 5.0 released in Oct 2014). In this framework, enumerating cameras and their capabilities (e.g: format, size) does not seem to need camera permission, so application can request for user permission only when about to use it (instead of in the library initialization).

PS: there is also [CameraX](https://developer.android.com/training/camerax) framework as jetpack support library, which seems to be a wrapper of both Camera & camera2 frameworks, but it seems to be still in alpha version. Also it does not seem to work (straightforwardly) with `SurfaceView` for video preview (initially this is the main reason of not using this framework, but lately realized that we use `SurfaceView` for debugging purpose only).

An interesting fact when testing camera2 in Samsung S10, it does not report NV21 format capability as in older framework Camera, it reports I420/YUV_420_888 capability (along with few private formats). However, when capturing using I420 format, the returned frame buffer is actually in NV21 format, but represented as I420 via camera2 Plane structure, which has 'pixel stride' attribute (note: it is different from 'row stride').

### DIFF
--- a/pjmedia/src/pjmedia-videodev/android/PjCamera2.java
+++ b/pjmedia/src/pjmedia-videodev/android/PjCamera2.java
@@ -1,0 +1,307 @@
+/* $Id$ */
+/*
+ * Copyright (C) 2021 Teluu Inc. (http://www.teluu.com)
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+package org.pjsip;
+
+import android.hardware.camera2.CameraCaptureSession;
+import android.hardware.camera2.CameraDevice;
+import android.hardware.camera2.CameraManager;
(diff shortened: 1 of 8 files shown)


pjmedia/src/pjmedia-videodev/android/PjCamera2.java
```diff
@@ -0,0 +1,307 @@
+/* $Id$ */
```
sauwming, line 1: No longer needed?
Also below.

pjmedia/src/pjmedia-videodev/android_dev.c
```diff
@@ -1109,6 +1143,162 @@ static pj_status_t and_stream_destroy(pjmedia_vid_dev_stream *s)
...
+    f.timestamp.u64 = strm->frame_ts.u64;
+    f.buf = data_buf = p0;
+
+    Y = (pj_uint8_t*)f.buf;
+    U = Y + strm->vafp.plane_bytes[0];
+    V = U + strm->vafp.plane_bytes[1];
+
+    /* Check if we need conversion here, this is the tricky part of camera2.
+     * When we request I420, the returned buffer may not be actually I420,
+     * for example it is NV21 in Samsung S10. The camera2 'cheats' us via
```
sauwming, line 1198: Maybe it's better to move device-specific testing to the issue desc/comment instead. One reason is that there will be timestamp that can be easily tracked in the issue page.

sauwming (approved):

pjmedia/src/pjmedia-videodev/android_dev.c
nanangizz: The code comment is about a sample case & details of the buffer trick by camera2 which is closely related to the conversion checks done in the following code, instead of a specific device test result. Anyway, the device may exhibit different behavior in the future which can be confusing, so I just stripped it out.

Issue #2181, opened 2019-03-05
Video conference implementation


pjsipbot:
_2019-03-05 06:11:20: @nanangizz_ _created the issue on trac ticket [2181](https://trac.pjsip.org/repos/ticket/2181)_

This ticket will implement video conference using centralized approach which is very similar to the existing audio conference, i.e: there will be an endpoint that act as the conference manager that is capable to establish multiple video calls, mix video from the participants (managed via connecting/disconnecting ports), and then send the mixing result to each participant. The video mixing operation is done by video conference bridge, which will use very similar APIs to the audio conference bridge.

### Video conference bridge specifications

 - Video timing clock, just like audio, the video conference bridge will have a clock that schedules video flow between ports.
 - All registered ports must be in passive mode (at least for now), so the conference bridge will actively pull video frames from source ports and push video frames to sink ports.
 - Ports can be bidirectional (they are PJMEDIA ports anyway) and just like in audio, ports connection is unidirectional, from a source port to a sink port.
 - A source port may send to multiple sink ports (splitting), and a sink port may receive from multiple source ports (mixing). For splitting, there is no sink port number limitation, but for mixing currently it will only support up to four video sources (actually no limitation, but only the first four will be rendered).
 - Video mixing layout. For now it will be static and no customization, e.g:
   - two source ports will be rendered into left/right layout for a landscape sink port, or top/bottom layout for a portrait sink
   - three source ports will be rendered into left+(top/bottom) layout for a landscape sink, or top+(left/right) layout for a portrait sink
   - four source ports will be rendered into 2x2 layout (equally divided slices) for a landscape sink, or four vertically stacked slices for a portrait sink.

### Implementation

Basically, the video conference implementation shares the same concept as the audio conference, so for applications that already have audio conference feature, integrating video conference feature should be relatively simple.

For participant, no modification is needed, it will see the video conference as a usual P2P video call, the video from other participants will be rendered into the same video window of the remote side of the video call (the conference manager).

For conference manager, it can setups the video conference using a very similar way as setting up an audio conference (with additional efforts for managing the UI perhaps). A sample code for this can be checked below.

### Conference ports in a video window in PJSUA

A video window may be related to two or more conference ports, application should be aware of this so it can manage the conference properly.

A video window represents a video renderer device and is automatically registered to the video conference bridge. The conference slot ID can be queried via `pjsua_vid_win_get_info()`, the slot ID should be in `pjsua_vid_win_info.slot_id`. This is a sink port.

[...]

PR #2798, opened 2021-08-16
High video latency when decoder frame rate is initialized lower than actually received

nanangizz:
When the video receiver endpoint is initialized with 15 frames per second (default setting in OpenH264) and the sender actually sends 30 frames per second, video latency experienced by the receiver will keep increasing until jitter buffer capacity maxed out. 

After investigation, the video conference only configures the sink port's frame rate once, i.e: when it is registered. While in the event of frame rate changed (or format changed), PJSUA-LIB only reconnect the stream decoder to renderer, which won't update the stream decoder's frame rate in the video conference bridge. This PR's patch introduces new API to update/refresh port states in the video conference bridge, which is invoked by PJSUA-LIB when it receives format change event. Note that with the new API, reconnecting stream decoder to renderer is no longer necessary, also any other connection involving stream decoder or renderer should also be updated (also without reconnecting).

This issue should not be occurred prior to video conference implementation (#2181).

### DIFF
--- a/pjmedia/src/pjmedia/vid_conf.c
+++ b/pjmedia/src/pjmedia/vid_conf.c
@@ -87,6 +87,7 @@ typedef struct vconf_port
     unsigned		 idx;		/**< Port index.		    */
     pj_str_t		 name;		/**< Port name.			    */
     pjmedia_port	*port;		/**< Video port.		    */
+    pjmedia_format	 format;	/**< Copy of port format info.	    */
     pj_uint32_t		 ts_interval;	/**< Port put/get interval.	    */
     pj_timestamp	 ts_next;	/**< Time for next put/get_frame(). */
     void		*get_buf;	/**< Buffer for get_frame().	    */
@@ -269,6 +270,7 @@ PJ_DEF(pj_status_t) pjmedia_vid_conf_add_port( pjmedia_vid_conf *vid_conf,
     /* Set pool, port, index, and name */
     cport->pool = pool;
     cport->port = port;
+    cport->format = port->info.fmt;
     cport->idx  = index;
     pj_strdup_with_null(pool, &cport->name, name);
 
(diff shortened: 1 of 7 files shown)


sauwming:
Would this be better if there's a new API instead, such as `pjsua_vid_conf_update_port()`?

nanangizz:
> Would this be better if there's a new API instead, such as `pjsua_vid_conf_update_port()`?

Good idea.

pjmedia/src/pjmedia/vid_conf.c
```diff
@@ -1109,4 +1111,117 @@ static pj_status_t render_src_frame(vconf_port *src, vconf_port *sink,
...
+	pjmedia_video_apply_fmt_param vafp;
+	pj_status_t status;
+	unsigned i;
+
+	vfi = pjmedia_get_video_format_info(NULL, new_fmt.id);
+	if (!vfi) {
+	    PJ_LOG(1,(THIS_FILE, "pjmedia_vid_conf_update_port(): "
+				 "unrecognized format %04X",
+				 new_fmt.id));
+	    return PJMEDIA_EBADFMT;
```
sauwming, line 1179: pj_mutex_unlock

pjmedia/src/pjmedia/vid_conf.c
```diff
@@ -1109,4 +1111,117 @@ static pj_status_t render_src_frame(vconf_port *src, vconf_port *sink,
...
+	}
+
+	pj_bzero(&vafp, sizeof(vafp));
+	vafp.size = new_fmt.det.vid.size;
+	status = (*vfi->apply_fmt)(vfi, &vafp);
+	if (status != PJ_SUCCESS) {
+	    PJ_LOG(1,(THIS_FILE, "pjmedia_vid_conf_update_port(): "
+				 "Failed to apply format %04X",
+				 new_fmt.id));
+	    return status;
```
sauwming, line 1189: pj_mutex_unlock

sauwming (approved):

trengginas (approved):

PR #2799, opened 2021-08-17
Added Function pjmedia_sdp_media_print(...).

wosrediinanatour:
Similar as pjmedia_sdp_print(), but only for media description.

### DIFF
--- a/pjmedia/include/pjmedia/sdp.h
+++ b/pjmedia/include/pjmedia/sdp.h
@@ -484,6 +484,17 @@ struct pjmedia_sdp_media
  */
 typedef struct pjmedia_sdp_media pjmedia_sdp_media;
 
+/**
+ * Print media description to a buffer.
+ *
+ * @param media	    The media description.
+ * @param buf	    The buffer.
+ * @param size	    The buffer length.
+ *
+ * @return	    the length printed, or -1 if the buffer is too
+ *		    short.
+ */
+PJ_DECL(int) pjmedia_sdp_media_print(const pjmedia_sdp_media *media, char *buf, pj_size_t size);
 
 /** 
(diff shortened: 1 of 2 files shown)


pjmedia/include/pjmedia/sdp.h
```diff
@@ -484,6 +484,17 @@ struct pjmedia_sdp_media
...
+ * Print media description to a buffer.
+ *
+ * @param media	 The media description.
+ * @param buf	    The buffer.
+ * @param size	    The buffer length.
+ *
+ * @return	    the length printed, or -1 if the buffer is too
+ *		    short.
+ */
+PJ_DEF(int) pjmedia_sdp_media_print(const pjmedia_sdp_media *media, char *buf, pj_size_t size);
```
nanangizz, line 497: - For declaration in header file, it should use `PJ_DECL` instead of `PJ_DEF`.
- The tabs/spaces in the param docs do not look properly aligned, but perhaps just the github diff preview.

wosrediinanatour:
It seems that the two failing checks have a different reason.

nanangizz (approved):

nanangizz:
> It seems that the two failing checks have a different reason.

Yes, don't worry about those. As the patch is currently not really used in the lib/test scenarios, I guess we'd just need to check for build errors.

FYI, some test scenarios are a bit sensitive to environment dynamics, e.g: it does not retry to listen on another port when a socket port is not available, so some failures can be ignored. Also, tests are a bit redundant (e.g: ffmpeg & openh264 variants), usually we can use that as a 'gauge' too.

trengginas (approved):

sauwming (approved):

PR #2806, opened 2021-08-24
Resort ICE checklist upon entering nomination stage

sauwming:
It is reported that in some situations, the successful peer reflexive (prflx) ICE candidates are not nominated, even though having the highest priority. This is because the prefix checks discovered when handling incoming checks (handle_incoming_check()) are added at the end of the check list. A check is performed right away on this new candidate, but the check list is not re-sorted.

The consequence is when the negotiation process enters the nomination stage when the timer `TIMER_START_NOMINATED_CHECK` fires, the checks with prflx candidates will have the wrong order in the list, which can cause the nomination of a check with a lower priority. And since the negotiation may complete as soon as all the components have a nominated check, the prflx may never get the chance to be nominated because they have the wrong order in the check list.

Note that the issue is more obvious with regular nomination, and also, if the Ta timer is larger than the default (20ms) (The reporter tested with a Ta=50ms as recommended in RFC-8445).

The proposed fix is to re-sort the check list each time a new pair is added.

Thanks to Mohamed Chibani for the report and the initial patch.

### DIFF
--- a/pjnath/src/pjnath/ice_session.c
+++ b/pjnath/src/pjnath/ice_session.c
@@ -2424,6 +2424,11 @@ static void start_nominated_check(pj_ice_sess *ice)
 	pj_timer_heap_cancel_if_active(ice->stun_cfg.timer_heap, &ice->timer,
 	                               TIMER_NONE);
     }
+	
+    /* Sort the list since some checks (with prflx cand in particular) 
+     * may have been added since the last sort.
+     */
+    sort_checklist(ice, &ice->clist);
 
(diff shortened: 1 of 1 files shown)


nanangizz:
Good catch!

I think the checklist resort should be done every time a new pair is added, and for this case it should be done in (the end of) `handle_incoming_check()`. Otherwise, aggressive nomination will suffer from the same issue.

nanangizz (approved):

trengginas (approved):

PR #2775, opened 2021-07-27
Support webrtc-aec3 on windows

trengginas:
PR #2722 add support for WebRTC AEC3. 
This ticket will add the support for Visual Studio (windows).

Requirement:
- Visual Studio 2015 or later with C++17 support

Changes:
- Some source files needs to be renamed.  Visual Studio will choose one implementation if source code are using the same name, which might lead to linker error.
- Create new project to Visual Studio solution (libwebrtc_aec3). By default it is not build since it will add considerable time to build. e.g: On my machine it will take ~2 minutes to build the single project. Compared to ~50s to build all the projects (without libwebrtc_aec3) in the solution.
- Define some build macro specific to building WebRTC-AEC3 on windows.
- Add echo_webrtc_aec3.c to pjmedia project.

Enabling WebRTC-AEC3 on Visual Studio 2015 or later:
1. Define `PJMEDIA_HAS_WEBRTC_AEC3` on `config_site.h` file
2. Change the BuildToolset configuration on build/vs/pjproject-vs14-common-config.props to v142.
```
     <When Condition="'$(API_Family)'=='WinDesktop'">
       <PropertyGroup>
-        <BuildToolset>v140</BuildToolset>
+        <BuildToolset>v142</BuildToolset>
```
3. Enable buildng libwebrtc-aec3 project from solution configuration manager.
4. When building samples project, you need to link with libwebrtc_aec3 library by editing Samples-vc.mak file.
```
+WEBRTC_AEC3_LIB = ..\..\third_party\lib\libwebrtc_aec3-$(TARGET)$(LIBEXT)
 
 THIRD_PARTY_LIBS = $(GSM_LIB) $(ILBC_LIB) $(RESAMPLE_LIB) \
 		   $(SPEEX_LIB) $(SRTP_LIB) $(G7221_LIB) $(BASECLASSES_LIB) \
-		   $(YUV_LIB) $(WEBRTC_LIB)
+		   $(YUV_LIB) $(WEBRTC_LIB) $(WEBRTC_AEC3_LIB)
```

PR #2807, opened 2021-08-24
Add support to windows on ARM build

trengginas:
Windows on ARM machine allow x86 win32 app to run without modification through emulation ([ref](https://docs.microsoft.com/en-us/windows/uwp/porting/apps-on-arm)).
This ticket will add support building the library using ARM64 specific configuration.

Requirement:
- VS2017 or newer (Visual Studio with Build toolset version v142)

Steps to build:
- The current solution also supports VS2015, and uses v140 build toolset version. You can change the build toolset from `pjproject-vs14-common-config.props` to v142.
```
    <DefaultToolset>v142</DefaultToolset>
    ....
        <BuildToolset>v142</BuildToolset>
```
- Choose ARM64 from the solution platform dropdown option.

Note:
- On VS2015, when opening the solution this warning will be printed.
`warning  : Platform '[ARM64, 0]' referenced in the project file cannot be found.`
  This is safe to ignore.
- When adding third party library (e.g: OpenSSL),  they also has to be built using ARM64 configuration.


nanangizz:
> Windows 10 on ARM64 machine allow running Win32 app through emulation.

Should not this is for ARM64 native (not emulated)?

When building for win32/x64 on VS2019, there are errors like:
--
third_party\webrtc_aec3\src\rtc_base/checks.h(248): error C3546: '...': there are no parameter packs available to expand
--

trengginas:
> > Windows 10 on ARM64 machine allow running Win32 app through emulation.
> 
> Should not this is for ARM64 native (not emulated)?

Yes, this is specific for ARM64 build. Perhaps we should change the wording above. The point is, that building the library using Win32 configuration should work on ARM64 machine.
> 
> When building for win32/x64 on VS2019, there are errors like:
> 
> ```
> third_party\webrtc_aec3\src\rtc_base/checks.h(248): error C3546: '...': there are no parameter packs available to expand
> ```

Building webrtc_aec3 require to use buildtoolset v142 (#2775). I tried here using win32 setup and it builds succesfully.

nanangizz:
> > > Windows 10 on ARM64 machine allow running Win32 app through emulation.
> > 
> > 
> > Should not this is for ARM64 native (not emulated)?
> 
> Yes, this is specific for ARM64 build. Perhaps we should change the wording above. The point is, that building the library using Win32 configuration should work on ARM64 machine.

Actually not sure about "Win32" term either as it can be ambiguous with "Win32" platform in VS config manager. Perhaps let's just use "Windows on ARM" which seems to be commonly used?

Also, please update the Getting Started trac wiki too about this new platform support.

> 
> > When building for win32/x64 on VS2019, there are errors like:
> > ```
> > third_party\webrtc_aec3\src\rtc_base/checks.h(248): error C3546: '...': there are no parameter packs available to expand
> > ```
> 
> Building webrtc_aec3 require to use buildtoolset v142 (#2775). I tried here using win32 setup and it builds succesfully.

Ah okay. Actually I didn't want to build libwebrtc_aec3 (no PJMEDIA_HAS_WEBRTC_AEC3 defined), but since there were about 9 linking errors with error message like below, I clicked build libwebrtc_aec3.
```LINK : fatal error LNK1104: cannot open file 'pjproject\third_party\lib\libwebrtc_aec3-i386-Win32-vc14-Debug.lib'```

So now there is this linking error issue. FYI, I just get the lib source, create an empty config_site.h, then build. Tried with both build toolsets v140 & v142.

trengginas:
> > > > Windows 10 on ARM64 machine allow running Win32 app through emulation.
> > > 
> > > 
> > > Should not this is for ARM64 native (not emulated)?
> > 
> > 
> > Yes, this is specific for ARM64 build. Perhaps we should change the wording above. The point is, that building the library using Win32 configuration should work on ARM64 machine.
> 
> Actually not sure about "Win32" term either as it can be ambiguous with "Win32" platform in VS config manager. Perhaps let's just use "Windows on ARM" which seems to be commonly used?
> 
> Also, please update the Getting Started trac wiki too about this new platform support.
Okay
> 
> > > When building for win32/x64 on VS2019, there are errors like:
> > > ```
> > > third_party\webrtc_aec3\src\rtc_base/checks.h(248): error C3546: '...': there are no parameter packs available to expand
> > > ```
> > 
> > 
> > Building webrtc_aec3 require to use buildtoolset v142 (#2775). I tried here using win32 setup and it builds succesfully.
> 
> Ah okay. Actually I didn't want to build libwebrtc_aec3 (no PJMEDIA_HAS_WEBRTC_AEC3 defined), but since there were about 9 linking errors with error message like below, I clicked build libwebrtc_aec3.
> `LINK : fatal error LNK1104: cannot open file 'pjproject\third_party\lib\libwebrtc_aec3-i386-Win32-vc14-Debug.lib'`
> 
> So now there is this linking error issue. FYI, I just get the lib source, create an empty config_site.h, then build. Tried with both build toolsets v140 & v142.

Right, #2775 misses something here. By default libwebrtc_aec3 is not included to be build. However libpjproject still reference libwebrtc_aec3 project. Since it is not build, linker error is raised. To fix the error, by default libpjproject will also need to remove libwebrtc_aec3 reference. User require to add the reference to libwebrtc_aec3 manually when when needed.

trengginas:
> > > > > Windows 10 on ARM64 machine allow running Win32 app through emulation.
> > > > 
> > > > 
> > > > Should not this is for ARM64 native (not emulated)?
> > > 
> > > 
> > > Yes, this is specific for ARM64 build. Perhaps we should change the wording above. The point is, that building the library using Win32 configuration should work on ARM64 machine.
> > 
> > 
> > Actually not sure about "Win32" term either as it can be ambiguous with "Win32" platform in VS config manager. Perhaps let's just use "Windows on ARM" which seems to be commonly used?
> > Also, please update the Getting Started trac wiki too about this new platform support.
> > Okay
> > > > When building for win32/x64 on VS2019, there are errors like:
> > > > ```
> > > > third_party\webrtc_aec3\src\rtc_base/checks.h(248): error C3546: '...': there are no parameter packs available to expand
> > > > ```
> > > 
> > > 
> > > Building webrtc_aec3 require to use buildtoolset v142 (#2775). I tried here using win32 setup and it builds succesfully.
> > 
> > 
> > Ah okay. Actually I didn't want to build libwebrtc_aec3 (no PJMEDIA_HAS_WEBRTC_AEC3 defined), but since there were about 9 linking errors with error message like below, I clicked build libwebrtc_aec3.
> > `LINK : fatal error LNK1104: cannot open file 'pjproject\third_party\lib\libwebrtc_aec3-i386-Win32-vc14-Debug.lib'`
> > So now there is this linking error issue. FYI, I just get the lib source, create an empty config_site.h, then build. Tried with both build toolsets v140 & v142.
> 
> Right, #2775 misses something here. By default libwebrtc_aec3 is not included to be build. However libpjproject still reference libwebrtc_aec3 project. Since it is not build, linker error is raised. To fix the error, by default libpjproject will also need to remove libwebrtc_aec3 reference. User require to add the reference to libwebrtc_aec3 manually when when needed.

The alternative would be to include an empty dummy file (e.g: used by libpjproject) when toolset is set to v140. And include the actual source when toolset is set to v142. This means by default, the libwebrtc_aec3 will be included to be built.

PR #2809, opened 2021-08-31
Allow registration refresh to continue when REGISTER response contains expires parameter set to 0

trengginas:
Consider this scenario:
- Client is behind a firewall/NAT and send registration and responded with 200/OK
- The firewall is restarted. And the client send refresh registration
- Server adds a new Contact header to the 200/OK response since it detects a different Via address on the request
- The first binding is never refreshed and eventually expires. This is reflected on the 200/OK REGISTER response
- When the client received the expires parameter to 0, it stops refreshing the registration although the registration should still be active

In this case, the client should be allowed to continue the refresh registration. If the server receive an expired refresh registration, it should send error response instead of a 200/OK response.

Note that this scenario can be avoided by setting `pjsua_acc_config.allow_contact_rewrite` set to PJ_TRUE. This way the library will send an updated contact to the server when it detects any IP/port changes on the REGISTER response.

### DIFF
--- a/pjsip/include/pjsip/sip_config.h
+++ b/pjsip/include/pjsip/sip_config.h
@@ -1244,15 +1244,14 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 
 
 /**
- * Specify the default expire refresh value when the registers sent a Contact
+ * Allow client to send refresh registration when the registrar sent a Contact
  * header with expire parameter 0 in the 200/OK REGISTER response.
- * This value will be used when the REGISTER request doesn't include a 
- * Expire header.
+ * Refer to https://github.com/pjsip/pjproject/pull/2809 for more info.
  *
- * Default is 30.
+ * Default is 1.
  */
-#ifndef PJSIP_REGISTER_EXP_REFRESH
-#   define PJSIP_REGISTER_EXP_REFRESH	30
+#ifndef PJSIP_REGISTER_ALLOW_EXP_REFRESH
+#   define PJSIP_REGISTER_ALLOW_EXP_REFRESH	1
 #endif
 
 
--- a/pjsip/include/pjsip/sip_config.h
+++ b/pjsip/include/pjsip/sip_config.h
@@ -1242,6 +1242,20 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 #   define PJSIP_REGISTER_CLIENT_ADD_XUID_PARAM	0
 #endif
 
+
+/**
+ * Specify the default expire refresh value when the registers sent a Contact
+ * header with expire parameter 0 in the 200/OK REGISTER response.
+ * This value will be used when the REGISTER request doesn't include a 
+ * Expire header.
+ *
+ * Default is 30.
+ */
+#ifndef PJSIP_REGISTER_EXP_REFRESH
+#   define PJSIP_REGISTER_EXP_REFRESH	30
+#endif
+
+
 /**
  * Maximum size of pool allowed for auth client session in pjsip_regc.
  * After the size exceeds because of Digest authentication processing,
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsip-ua/sip_reg.c
```diff
@@ -1359,6 +1359,14 @@ static void regc_tsx_callback(void *token, pjsip_event *event)
 						       PJSIP_REGC_MAX_CONTACT,
 						       contact);
 
+	    if (expiration == 0 && regc->current_op != REGC_UNREGISTERING) {
+		if (regc->expires_hdr && regc->expires_hdr->ivalue) {
+		    expiration = regc->expires_hdr->ivalue;
+		} else {
+		    expiration = PJSIP_REGISTER_EXP_REFRESH;
+		}
+	    }
```
nanangizz, line 1368: As this code handles a corner case:
- print some log?
- put a comment too, perhaps simply a reference to this PR

And as it introduces behavioral change/backward incompatibility:
  - consider a macro setting to disable this perhaps
  - please don't forget to mention it in the milestone after merged

nanangizz (approved):
Just few notes below

pjsip/include/pjsip/sip_config.h
```diff
@@ -1242,6 +1242,20 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
...
+ * Specify the default expire refresh value when the registers sent a Contact
+ * header with expire parameter 0 in the 200/OK REGISTER response.
+ * This value will be used when the REGISTER request doesn't include a 
+ * Expire header.
+ *
+ * Default is 30.
+ */
+#ifndef PJSIP_REGISTER_EXP_REFRESH
+#   define PJSIP_REGISTER_EXP_REFRESH	30
+#endif
```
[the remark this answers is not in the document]
sauwming: Isn't 3600 (sec) a bit too long?
More about this below.

pjsip/src/pjsip-ua/sip_reg.c
sauwming: So suppose the registrar responses with:
case 1:
```
Contact: <sip:url1>;expires=0
```
case 2:
```
Contact: <sip:url1>;expires=0
Contact: <sip:url1>;expires=180
```

What expiration do we actually want to set, which will determine when we will try to reregister?
If it's too long (such as 3600), in case 2, it will have already expired, won't It?

In my opinion, it should be:
if server specifies expiration, min(PJSIP_REGISTER_EXP_REFRESH, the lowest non-zero expiration), else min (PJSIP_REGISTER_EXP_REFRESH, regc->expires_hdr->ivalue)

pjsip/include/pjsip/sip_config.h
```diff
@@ -1242,6 +1242,20 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 #   define PJSIP_REGISTER_CLIENT_ADD_XUID_PARAM	0
 #endif
 
+
+/**
+ * Specify the default expire refresh value when the registers sent a Contact
```
sauwming, line 1247: registrar

pjsip/src/pjsip-ua/sip_reg.c
nanangizz: Basically I am okay with any value, as this is for handling a corner case only, not really a general default value. So I just tried to point out there is already 'default value' (for cases that there is no expire value), not sure if we should have another and configurable (while the others are not). Also, 3600 seems to be the implicitly-recommended default value for non-determinable value (e.g: [here](https://datatracker.ietf.org/doc/html/rfc3261#section-10.2)). So IMO if we need to have a default value, I guess 3600 should be reasonable (btw, 30 is too small too, isn't it?).

For case 2 above, perhaps ideally we can use the 180s, but the modification needs to be done inside `calculate_response_expiration()`, e.g: skip the expire=0 in the iteration. But again I am okay too with the current patch :)

The sip_reg.c seems to have an 'algo' to decide the expire value if there is none, e.g: 
https://github.com/pjsip/pjproject/blob/ccc0644df69b7f9cf00de22cb1e4dec25d01413c/pjsip/src/pjsip-ua/sip_reg.c#L1025

sauwming (approved):

nanangizz (approved):

PR #2810, opened 2021-09-01
Add a new ioqueue API pj_ioqueue_clear_key()

nanangizz:
The API will reset ioqueue key states and cancel any outstanding key operations. Currently it is only used in UDP media transport to replace `pj_ioqueue_post_completion()` in `transport_media_stop()`, which has been reported to have infinite loop issue (to be investigated).

### DIFF
--- a/pjlib/include/pj/ioqueue.h
+++ b/pjlib/include/pj/ioqueue.h
@@ -589,6 +589,19 @@ PJ_DECL(pj_status_t) pj_ioqueue_post_completion( pj_ioqueue_key_t *key,
                                                  pj_ssize_t bytes_status );
 
 
+/**
+ * Clear ioqueue key states. This function will cancel any outstanding
+ * operations on that key, without invoking any completion callback.
+ * After calling this function, application should reinit its all operation
+ * keys, i.e: using pj_ioqueue_op_key_init(), before reusing them.
+ *
+ * @param key           The key.
+ *
+ * @return		PJ_SUCCESS on success or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pj_ioqueue_clear_key( pj_ioqueue_key_t *key );
+
+
 
 #if defined(PJ_HAS_TCP) && PJ_HAS_TCP != 0
 /**
--- a/pjmedia/src/pjmedia/transport_udp.c
+++ b/pjmedia/src/pjmedia/transport_udp.c
@@ -37,6 +37,13 @@
 /* Maximum pending write operations */
 #define MAX_PENDING 4
 
+#if 1
+#  define TRACE_(expr)
+#else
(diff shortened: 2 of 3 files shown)


pjlib/include/pj/ioqueue.h
```diff
@@ -589,6 +589,19 @@ PJ_DECL(pj_status_t) pj_ioqueue_post_completion( pj_ioqueue_key_t *key,
                                                  pj_ssize_t bytes_status );
 
 
+/**
+ * Clear ioqueue key states. This function will cancel any outstanding
+ * operations on that key, without invoking any completion callback.
+ * After calling this function, application should reinit its all operation
+ * keys, i.e: using pj_ioqueue_op_key_init(), before reusing them.
```
sauwming, line 596: So this means that the ioq_op_key_init() needs to be moved from transport_attach() to _start(), right?

pjmedia/src/pjmedia/transport_udp.c
```diff
@@ -1218,11 +1215,8 @@ static pj_status_t transport_media_stop(pjmedia_transport *tp)
     if (!udp->started)
 	return PJ_SUCCESS;
 
-    pj_ioqueue_post_completion(udp->rtp_key, &udp->rtp_read_op,
-			       -PJ_ECANCELLED);
-
-    pj_ioqueue_post_completion(udp->rtcp_key, &udp->rtcp_read_op,
-			       -PJ_ECANCELLED);
+    pj_ioqueue_clear_key(udp->rtp_key);
```
sauwming, line 1218: With this, app will no longer receive callback with cancelled status. Wondering if this can cause any problem, such as in app that expects to receive such callback.

pjlib/include/pj/ioqueue.h
nanangizz: ah right, missed that.

pjmedia/src/pjmedia/transport_udp.c
nanangizz: Currently the using app is just UDP media transport and I think it is fine without callback.

Moreover, callback is bound to operation key, so there should be two callbacks if there are two outstanding ops and for the infinite issue, where the operation lists seem to be in bad state, we cannot iterate the list.

trengginas: Perhaps add an option to maintain the old behavior?

nanangizz: The modification only affects UDP media transport internally, there should not be any behavior change for app.

trengginas (approved):

sauwming (approved):
Since the patch is meant for an older PJSIP version, it seems to still have conflict with the latest master though.

Issue #2064, opened 2017-11-14
Fix return code in pjsip_find_msg()


pjsipbot:
_2017-11-14 11:06:23: @trengginas_ _created the issue on trac ticket [2064](https://trac.pjsip.org/repos/ticket/2064)_

The default return code for ```pjsip_find_msg()``` was ```PJ_SUCCESS``` so if
a ```Content-Length``` header wasn't found at all, ```pjsip_find_msg()``` was
returning ```PJ_SUCCESS``` instead of ```PJSIP_EMISSINGHDR```.

Also added the volatile keyword to a few variables what are used
both inside and outsude the ```PJ_TRY/PJ_CATCH``` block.

Thanks to George Joseph for the patch.
___
_2017-11-14 11:08:13: @trengginas_ _changed **status** from **new** to **closed**_

___
_2017-11-14 11:08:13: @trengginas_ _set **resolution** to **fixed**_

___
_2017-11-14 11:08:13: @trengginas_ _commented_

In [r5694](../commit/975aa1047a4dcc439315d9180c0f850e855c37d0):
Fixed #2064: Fix return code in pjsip_find_msg().

Issue #2056, opened 2017-11-08
Add validity checking for numeric header values


pjsipbot:
_2017-11-08 02:53:09: @trengginas_ _created the issue on trac ticket [2056](https://trac.pjsip.org/repos/ticket/2056)_

Parsing the numeric header fields like cseq, ttl, port, etc. all
had the potential to overflow, either causing unintended values to
be captured or, if the values were subsequently converted back to
strings, a buffer overrun. This will lead to a potential exploit using carefully crafted invalid values.
To address this, new "strto" functions
have been created that do range checking and those functions are
used wherever possible in the parser.

This patch will:
 - Created ```pjlib/include/limits.h``` and ```pjlib/include/compat/limits.h```
   to either include the system ```limits.h``` or define common numeric
   limits if there is no system ```limits.h```.
 - Created ```strtoi_validate``` functions in ```sip_parser``` that take bounds
   and on failure call the ```on_str_parse_error``` function which prints
   an error message and calls ```PJ_THROW```.
 - Updated ```sip_parser``` to validate the numeric fields.
 - Fixed an issue in ```sip_transport``` that prevented error messages
   from being properly displayed.
 - Added "volatile" to some variables referenced in ```PJ_CATCH``` blocks
   as the optimizer was sometimes optimizing them away.
 - Fixed length calculation in ```sip_transaction```/```create_tsx_key_2543```
   to account for signed ints being 11 characters, not 9.

Thanks to:
- Youngsung Kim at LINE Corporation for the report
- George Joseph ```<gjoseph@digium.com>``` for the patch
___
_2017-11-08 02:58:19: @trengginas_ _changed **status** from **new** to **closed**_

___
_2017-11-08 02:58:19: @trengginas_ _set **resolution** to **fixed**_

___
_2017-11-08 02:58:19: @trengginas_ _commented_

In [r5682](../commit/37f0826f9f89e2255332afa1ffa5d72dab4a149d):
Closed #2056: Add validity checking for numeric header values.

___
_2017-11-17 09:41:53: @trengginas_ _edited the issue description_

PR #2813, opened 2021-09-03
When optimization is enabled, local assignment might be lost on exception

trengginas:
It is reported that when using gcc 10 with -02, assignment prior to `PJ_CATCH_ANY` will be lost.
This is likely because the values are saved in registers when setjmp (`PJ_TRY`) is called, and restored when longjmp is called (`PJ_THROW`), which is why all assignments to local variables that occur between setjmp and longjmp might be discarded (with optimizations turned on).

This might raised an assertion which checks the error code assigned prior to raising the exception.

### DIFF
--- a/pjmedia/src/pjmedia/sdp.c
+++ b/pjmedia/src/pjmedia/sdp.c
@@ -50,21 +50,21 @@ typedef struct parse_context
 /*
  * Prototypes for line parser.
  */
-static void parse_version(pj_scanner *scanner, parse_context *ctx);
+static void parse_version(pj_scanner *scanner, volatile parse_context *ctx);
 static void parse_origin(pj_scanner *scanner, pjmedia_sdp_session *ses,
-			 parse_context *ctx);
+			 volatile parse_context *ctx);
 static void parse_time(pj_scanner *scanner, pjmedia_sdp_session *ses,
-		       parse_context *ctx);
+		       volatile parse_context *ctx);
 static void parse_generic_line(pj_scanner *scanner, pj_str_t *str,
-			       parse_context *ctx);
+			       volatile parse_context *ctx);
 static void parse_connection_info(pj_scanner *scanner, pjmedia_sdp_conn *conn,
-				  parse_context *ctx);
+				  volatile parse_context *ctx);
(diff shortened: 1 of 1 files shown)


nanangizz (approved):

sauwming:
Have we tried testing it ourselves with the sample code provided by the reporter?
Compare between -O0 and -O2 and perhaps another optimisation level. Then with a few, and the latest, gcc versions.

nanangizz:
> Have we tried testing it ourselves with the sample code provided by the reporter?
> Compare between -O0 and -O2 and perhaps another optimisation level. Then with a few, and the latest, gcc versions.

I think this is a generic `setjmp/longjmp` issue, not specific to gcc. We have seen similar issues before, e.g: #2056, #2064 (different approach though, added 'volatile' to local vars to avoid reverted/undetermined values).

sauwming (approved):

pjmedia/src/pjmedia/sdp.c
```diff
@@ -1433,6 +1431,7 @@ PJ_DEF(pj_status_t) pjmedia_sdp_parse( pj_pool_t *pool,
     }
     PJ_CATCH_ANY {
 	
+	ctx.last_error = PJ_GET_EXCEPTION();
```
nanangizz, line 1434: Some functions throw `SYNTAX_ERROR`, so `ctx.last_error` may be assigned  to `SYNTAX_ERROR` and returned by `pjmedia_sdp_parse()`, while app will expect the function to return a PJMEDIA error code.

I guess the alternatives are:
- change all throws to throw proper PJMEDIA error code, or
- using volatile approach (please test first), `parse_context` var in the function must be volatile, also all functions accepting and modifying `parse_context` param, must declare it as volatile too, e.g:
   ```
   static void parse_media(pj_scanner *scanner, pjmedia_sdp_media *med,
			volatile parse_context *ctx)
   ```

trengginas: The issue is reproducible by specifying `-O2` on the `CFLAGS` and assertion raised when there's a SDP parse error (exception raised). Modification using volatile approach eliminate the assertion and make sure the error code was set correctly.
- Tested using clang (macOS) with and without -O2 CFLAGS
- Tested using Visual Studio

PR #2818, opened 2021-09-06
build script for dev and sim, apple backend

oliverepper:
This scripts creates an xcframework containing all the required headers and static libraries for device and simulator. It's build for Apple backend (Network.framework and Security.framework). Doing this is not clear from the documentation so I though this script might help someone.


oliverepper:
I add a Mac build and some more improvements.

Issue #1965, opened 2016-10-07
Add support to specify Contact params specific to REGISTER requests


pjsipbot:
_2016-10-07 07:04:51: @mingteluu_ _created the issue on trac ticket [1965](https://trac.pjsip.org/repos/ticket/1965)_

This feature could be useful for specifying PN (Push Notification) information to the server, such as the one described in the draft [Push Notifications in the Session Initiation Protocol (SIP)](https://tools.ietf.org/html/draft-ivanov-sipcore-pnsip-01)

For example:
```
Contact: <sip:alice@alicemobile.example.com;
     pn-type=acme;pn-methods="INVITE";
     pn-uri="https://pn.acme.example.com/ZTY4ZDJlMzODE1NmUgKi0K">
```

___
_2016-10-07 08:42:23: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2016-10-07 08:42:23: @mingteluu_ _set **resolution** to **fixed**_

___
_2016-10-07 08:42:23: @mingteluu_ _commented_

In [r5455](../commit/9160ae5ad7abc2df48faead5d8eaca25eff138c1):
Fixed #1965: Add support to specify Contact params specific to REGISTER requests

PR #2819, opened 2021-09-07
Add support to specify Contact URI params specific to REGISTER requests

nanangizz:
In #1965, app can add Contact header params specific to REGISTER requests to specify Push Notification info. However, instead of in the Contact header params, actually the RFC recommends the PN info to be specified in the Contact URI params. So this PR introduces `pjsua_acc_config.reg_contact_uri_params` for PJSUA-LIB and `AccountRegConfig.contactUriParams` for PJSUA2 to allow app to specify Contact URI params specific to REGISTER requests.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -90,6 +90,8 @@ PJ_DEF(void) pjsua_acc_config_dup( pj_pool_t *pool,
     pj_strdup_with_null(pool, &dst->force_contact, &src->force_contact);
     pj_strdup_with_null(pool, &dst->reg_contact_params,
 			&src->reg_contact_params);
+    pj_strdup_with_null(pool, &dst->reg_contact_uri_params,
+			&src->reg_contact_uri_params);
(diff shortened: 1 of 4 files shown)


pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -1561,13 +1575,52 @@ static void update_regc_contact(pjsua_acc *acc)
...
+		pj_assert(uri != NULL);
+		uri = (pjsip_sip_uri*) pjsip_uri_get_uri(uri);
+		tmp_len = pjsip_uri_print(PJSIP_URI_IN_CONTACT_HDR,
+					  uri, tmp_uri,
+					  sizeof(tmp_uri));
+		pj_assert(tmp_len > 0);
+		pj_pool_release(pool);
+
+		/* Regenerate Contact */
+		reg_contact.slen = pj_ansi_snprintf(
```
sauwming, line 1613: There are a couple of warnings here

sauwming (approved):

trengginas (approved):

PR #2820, opened 2021-09-08
Avoid PJ_ERESOLVE when using IPv6

m-elagin:
This patch fixed the following:

when running pjsua with the parameters --ipv6  --ip-addr=xxx , the following error is being thrown:

`pjsua_core.c  Unable to resolve transport public address: gethostbyname() has returned error (PJ_ERESOLVE) [status=70018]
`
This is due to the wrong (ipv4) code branch being taken, as the value of app_config.ipv6 is being ignored.

### DIFF
--- a/pjsip-apps/src/pjsua/pjsua_app.c
+++ b/pjsip-apps/src/pjsua/pjsua_app.c
@@ -1643,7 +1643,7 @@ static pj_status_t app_init(void)
     }
 
     /* Add UDP transport unless it's disabled. */
-    if (!app_config.no_udp) {
+    if (!app_config.no_udp && !app_config.ipv6) {
 	pjsua_acc_id aid;
 	pjsip_transport_type_e type = PJSIP_TRANSPORT_UDP;
 


silentindark:
@sauwming @nanangizz Could you please take a look?

sauwming:
I try it here and get the same error as well.
Unfortunately you cannot disable the creation of UDP IPv4 transport just because `--ipv6` option is passed.

The proper fix should check whether the `public_addr` and `bound_addr` in `udp_cfg` is the correct IP version.
Something like this:
```
	pjsua_transport_config udp_cfg = app_config.udp_cfg;
        if (udp_cfg.public_addr.slen > 0 && get_ip_addr_ver(&udp_cfg.public_addr) != 4) {
             udp_cfg.public_addr.slen = 0;
        }
        if (udp_cfg.bound_addr.slen > 0 && get_ip_addr_ver(&udp_cfg.bound_addr) != 4) {
             udp_cfg.bound_addr.slen = 0;
        }
```

I tested it here and the above fixes the UDP resolution error, but there will be TCP listener creation failure. So the fix also needs to be applied to TCP, TLS, and RTP as well (and for each IP version (IPv4/IPv6)). Hence quite a lot and currently I am occupied with other things and can't create the PR myself.

sauwming:
Alternatively, perhaps easier, is to add `--no-ipv4` option to disable the creation of all (UDP, TCP, TLS) IPv4 transports.

sauwming:
Closing the PR since the proposed patch is incomplete to fix the issue. I have created #2983 instead to document the problem. Thanks.

PR #2822, opened 2021-09-10
Miscelaneous changes on PJSUA2 audio device APIs

nanangizz:
- Fix audio device still be opened when `setSndDevMode(PJSUA_SND_DEV_NO_IMMEDIATE_OPEN)` is called after `setNoDev()` & `setPlaybackDev(-1)`.
- Change behavior of `setPlaybackDev()/setCaptureDev()` to not change/remove current mode. I think it is more intuitive.
- Introduce a new PJSUA API `pjsua_get_snd_dev2()` to query current device IDs and mode.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_aud.c
+++ b/pjsip/src/pjsua-lib/pjsua_aud.c
@@ -2094,42 +2094,56 @@ PJ_DEF(pj_status_t) pjsua_set_snd_dev(int capture_dev,
 				      int playback_dev)
 {
     pjsua_snd_dev_param param;
-
-    pjsua_snd_dev_param_default(&param);
-
+    pjsua_get_snd_dev2(&param);
     param.capture_dev = capture_dev;
     param.playback_dev = playback_dev;
+
     /* Always open the sound device. */
     param.mode = 0;
 
     return pjsua_set_snd_dev2(&param);
 }
 
+
+PJ_DEF(pj_status_t) pjsua_get_snd_dev2(pjsua_snd_dev_param *snd_param)
+{
+    PJ_ASSERT_RETURN(snd_param, PJ_EINVAL);
+
+    PJSUA_LOCK();
+    snd_param->capture_dev = pjsua_var.cap_dev;
+    snd_param->playback_dev = pjsua_var.play_dev;
+    snd_param->mode = pjsua_var.snd_mode;
+    PJSUA_UNLOCK();
+
+    return PJ_SUCCESS;
+}
+
+
 /*
  * Select or change sound device. Application may call this function at
  * any time to replace current sound device.
  */
-PJ_DEF(pj_status_t) pjsua_set_snd_dev2(pjsua_snd_dev_param *snd_param)
+PJ_DEF(pj_status_t) pjsua_set_snd_dev2(const pjsua_snd_dev_param *snd_param)
 {
     unsigned alt_cr_cnt = 1;
     unsigned alt_cr[] = {0, 44100, 48000, 32000, 16000, 8000};
     unsigned i;
     pj_status_t status = -1;
     unsigned orig_snd_dev_mode = pjsua_var.snd_mode;
-    pj_bool_t no_change = (pjsua_var.snd_is_on || (!pjsua_var.snd_is_on &&
-			   (snd_param->mode & 
-			    PJSUA_SND_DEV_NO_IMMEDIATE_OPEN)));
 
-    PJ_LOG(4,(THIS_FILE, "Set sound device: capture=%d, playback=%d",
-	      snd_param->capture_dev, snd_param->playback_dev));
+    PJ_ASSERT_RETURN(snd_param, PJ_EINVAL);
+
+    PJ_LOG(4,(THIS_FILE, "Set sound device: capture=%d, playback=%d, mode=%d",
+	      snd_param->capture_dev, snd_param->playback_dev,
+	      snd_param->mode));
+
     pj_log_push_indent();
 
     PJSUA_LOCK();
 
     if (pjsua_var.cap_dev == snd_param->capture_dev &&
 	pjsua_var.play_dev == snd_param->playback_dev &&
-	pjsua_var.snd_mode == snd_param->mode &&
-	!pjsua_var.no_snd && no_change)
+	pjsua_var.snd_mode == snd_param->mode)
     {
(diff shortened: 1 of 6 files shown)


pjsip/src/pjsua2/endpoint.cpp
```diff
@@ -60,13 +60,21 @@ void TlsInfo::fromPj(const pjsip_tls_state_info &info)
...
     cipher 	= ssock_info->cipher;
-    cipherName	= pj_ssl_cipher_name(ssock_info->cipher);
+    cipher_name = pj_ssl_cipher_name(ssock_info->cipher);
+    if (cipher_name) {
+	cipherName = cipher_name;
+    } else {
+	char tmp[32];
+	pj_ansi_snprintf(tmp, sizeof(tmp), "Cipher 0x%x", cipher);
+	cipherName = tmp;
+    }
```
trengginas, line 77: mixed with other PR?

nanangizz: yes, wrong branch parent, just fixed it.

trengginas (approved):

sauwming:
It is reported that normal call audio doesn't work with this patch.
I tested using pjsua with the `master` branch vs `aud-dev-misc` and the report seems valid (i.e. no audio for normal call in `aud-dev-misc`).

nanangizz:
> It is reported that normal call audio doesn't work with this patch.
> I tested using pjsua with the `master` branch vs `aud-dev-misc` and the report seems valid (i.e. no audio for normal call in `aud-dev-misc`).

Well, I removed `no_change` variable as I was not sure what it is for (well, the var name is a bit confusing as it does not explicitly represent a comparison result). So, just put it back with comments and a slight structure change, hopefully it will improve the readability.

sauwming:
There's still a problem. If `PJSUA_SND_DEV_NO_IMMEDIATE_OPEN` mode is set, it turns out that the sound device is never opened during normal call.

nanangizz:
> There's still a problem. If `PJSUA_SND_DEV_NO_IMMEDIATE_OPEN` mode is set, it turns out that the sound device is never opened during normal call.

Just to clarify, is this a new issue (also occurs in master) or specific to this PR? Also, as `pjsua_call_make_call()`  & `pjsua_set_snd_dev()` use `pjsua_set_snd_dev()`, which always clears the mode, not sure when/where the `PJSUA_SND_DEV_NO_IMMEDIATE_OPEN` flag is set, so could you specify the reproducing steps too?

sauwming:
The scenario, as reported by the user, is as follows:
```
    ep.audDevManager().setNoDev();
    ep.audDevManager().setCaptureDev(1);
    ep.audDevManager().setPlaybackDev(0);
    ep.audDevManager().setSndDevMode(PJSUA_SND_DEV_SPEAKER_ONLY | PJSUA_SND_DEV_NO_IMMEDIATE_OPEN);
```
I also tried it:
a. without the SPEAKER_ONLY flag
b. without setting the capture device.

pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -2094,62 +2094,100 @@ PJ_DEF(pj_status_t) pjsua_set_snd_dev(int capture_dev,
...
 
     pjsua_var.snd_mode = snd_param->mode;
 
-    if (!pjsua_var.no_snd && !pjsua_var.snd_is_on &&
+    /* Just update the IDs if app does not want to open now and currently
+     * audio is off.
+     */
+    if (!pjsua_var.snd_is_on &&
 	(snd_param->mode & PJSUA_SND_DEV_NO_IMMEDIATE_OPEN))
     {
```
sauwming, line 2192: I think it's missing resetting the no sound in this block:
`    pjsua_var.no_snd = PJ_FALSE;`

sauwming:
> Just to clarify, is this a new issue (also occurs in master) or specific to this PR? Also, as `pjsua_call_make_call()` & `pjsua_set_snd_dev()` use `pjsua_set_snd_dev()`, which always clears the mode, not sure when/where the `PJSUA_SND_DEV_NO_IMMEDIATE_OPEN` flag is set, so could you specify the reproducing steps too?

The call to `pjsua_set_snd_dev()` in `make_call()` is problematic because it resets the mode to 0. So the speaker only mode is gone. Perhaps we should only remove the flag `IMMEDIATE_OPEN` in `set_snd_dev()`. Or should we use `pjsua_set_snd_dev2()` instead in `make_call()`?

nanangizz:
> > Just to clarify, is this a new issue (also occurs in master) or specific to this PR? Also, as `pjsua_call_make_call()` & `pjsua_set_snd_dev()` use `pjsua_set_snd_dev()`, which always clears the mode, not sure when/where the `PJSUA_SND_DEV_NO_IMMEDIATE_OPEN` flag is set, so could you specify the reproducing steps too?
> 
> The call to `pjsua_set_snd_dev()` in `make_call()` is problematic because it resets the mode to 0. So the speaker only mode is gone. Perhaps we should only remove the flag `IMMEDIATE_OPEN` in `set_snd_dev()`. Or should we use `pjsua_set_snd_dev2()` instead in `make_call()`?

Yes, seen it too & trying to fix it now. Perhaps the first approach, modifying set_snd_dev(), is a bit more intuitive, i.e: open immediately but maintaining other mode?

nanangizz:
> > > Just to clarify, is this a new issue (also occurs in master) or specific to this PR? Also, as `pjsua_call_make_call()` & `pjsua_set_snd_dev()` use `pjsua_set_snd_dev()`, which always clears the mode, not sure when/where the `PJSUA_SND_DEV_NO_IMMEDIATE_OPEN` flag is set, so could you specify the reproducing steps too?
> > 
> > 
> > The call to `pjsua_set_snd_dev()` in `make_call()` is problematic because it resets the mode to 0. So the speaker only mode is gone. Perhaps we should only remove the flag `IMMEDIATE_OPEN` in `set_snd_dev()`. Or should we use `pjsua_set_snd_dev2()` instead in `make_call()`?
> 
> Yes, seen it too & trying to fix it now. Perhaps the first approach, modifying set_snd_dev(), is a bit more intuitive, i.e: open immediately but maintaining other mode?

Perhaps 'intuitive' is not the correct word. `set_snd_dev()` is used in three places in the lib with immediate open behavior expected, so perhaps some apps out there also expect the same behavior, so I think it's better not to change it.

sauwming (approved):

PR #2826, opened 2021-09-13
Prevent assertion if Contact URI is not found when checking NAT address

sauwming:
For some use cases, application may put `*` as its Contact during registration and server may respond with no `Expires` header. This may trigger an assertion (or crash if assertion is disabled), so it may be desirable to handle this case, although this seems to be quite an uncommon scenario.

### DIFF
--- a/pjsip-apps/src/samples/pjsua2_demo.cpp
+++ b/pjsip-apps/src/samples/pjsua2_demo.cpp
@@ -162,8 +162,8 @@ void MyCall::onCallMediaState(OnCallMediaStateParam &prm)
     }
 
     // This will connect the wav file to the call audio media
-    if (wav_player)
-    	wav_player->startTransmit(aud_med);
+//    if (wav_player)
+//    	wav_player->startTransmit(aud_med);
 
     // And this will connect the call audio media to the sound device/speaker
     aud_med.startTransmit(play_dev_med);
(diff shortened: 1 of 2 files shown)


nanangizz:
Right, `acc->contact` generated by the lib should always contain a URI, I think that's why the verification uses assertion instead using normal check (using 'if'). So the problematic scenario may use `acc->cfg.force_contact` to set contact to `*`, haven't tried if it is possible though (i.e: no assertion or any other error raised). Also note that PJSIP module can only modify the SIP message, but not `acc->contact`.

On the other hand, the `force_contact` docs actually implicitly says it should be a URI:
```
Optional URI to be put as Contact for this account.
```
Unfortunately the lib currently does not verify if the `force_contact` contains a URI. So instead of replacing the assertion in this one place, perhaps we should add the URI verification for `force_contact`, as otherwise other places of the code may assume `force_contact` always has a URI, and crashing/assertion-raised as in the reported scenario?

sauwming:
Right, we put the checking in the app instead of inside the library:
https://github.com/pjsip/pjproject/blob/master/pjsip-apps/src/pjsua/pjsua_app_config.c#L830

And if I disable the checking, the contact `*` is indeed accepted by the library, and triggers this assertion in `acc_check_nat_addr()`.
Unfortunately, since we never put any checking in the library, putting the checking now may cause some backward incompatibility issue in user apps (even though the doc has specifically disallowed this), unless we provide another alternative for the users.

nanangizz:
Just a thought, allowing Contact without URI may introduce greater problem in the future, as mentioned before, some places in the lib may not be ready with non-URI Contact, e.g: in the deep of `pjsip_regc_update_contact()`, `pjsip_uri_cmp()` is called without URI verification. There is `pjsip_regc_unregister_all()` that uses '*' in Contact header, but it never actually stores the Contact header/setting anywhere, so the library practically never sees that Contact again.

And back to the original issue source, the reporter mentioned that he uses PJSIP module to modify the REGISTER message, I guess he could modify the Contact header from the module too, instead of setting it via `force_contact` (to be confirmed if it is done this way though).

sauwming:
Closing this since as the doc specifies, the Contact must indeed be a URI.

PR #2830, opened 2021-09-22
parser_get_and_unescape: If the spec allows %, don't unescape it

### LINKED ISSUE #2829: Percent characters (%) in Via parameters are unescaped
### Describe the bug

The `branch` parameter (and others) in the `Via` header are `token`s according to RFC 3261 but we unescape them anyway. Percent encoded hex pairs are decoded into their binary representations and subsequently printed incorrectly. For example, the log below is a request/response from Asterisk 16 (bundled with pjproject 2.10).

### Steps to reproduce

Parse a request that contains the `%` character followed by 2 hex digits which should be stored literally and see that they are decoded instead.

### PJSIP version

2.10 (Asterisk bundled)

### Context

Asterisk 16 from Git on Linux.

#### Request:

```none
INVITE sip:8675309@192.168.26.150 SIP/2.0
Via: SIP/2.0/UDP 192.168.26.1:62128;branch=z9hG4bKa524287%00%01%02%03%04%05%06%07%08%09%0A%0B%0C%0D%0E%0F%10%11%12%13%14%15%16%17%18%19%1A%1B%1C%1D%1E%1F%20%21%22%23%24%25%26%27%28%29%2A%2B%2C%2D%2E%2F%30%31%32%33%34%35%36%37%38%39%3A%3B%3C%3D%3E%3F%40%41%42%43%44%45%46%47%48%49%4A%4B%4C%4D%4E%4F%50%51%52%53%54%55%56%57%58%59%5A%5B%5C%5D%5E%5F%60%61%62%63%64%65%66%67%68%69%6A%6B%6C%6D%6E%6F%70%71%72%73%74%75%76%77%78%79%7A%7B%7C%7D%7E%7F%80%81%82%83%84%85%86%87%88%89%8A%8B%8C%8D%8E%8F%90%91%92%93%94%95%96%97%98%99%9A%9B%9C%9D%9E%9F%A0%A1%A2%A3%A4%A5%A6%A7%A8%A9%AA%AB%AC%AD%AE%AF%B0%B1%B2%B3%B4%B5%B6%B7%B8%B9%BA%BB%BC%BD%BE%BF%C0%C1%C2%C3%C4%C5%C6%C7%C8%C9%CA%CB%CC%CD%CE%CF%D0%D1%D2%D3%D4%D5%D6%D7%D8%D9%DA%DB%DC%DD%DE%DF%E0%E1%E2%E3%E4%E5%E6%E7%E8%E9%EA%EB%EC%ED%EE%EF%F0%F1%F2%F3%F4%F5%F6%F7%F8%F9%FA%FB%FC%FD%FE%FF;rport
Max-Forwards: 70
Contact: <sip:acct-206@192.168.26.1:62128;rinstance=7c73d8637ff8be32>
To: <sip:8675309@192.168.26.150>
From: "Gerrit"<sip:acct-206@192.168.26.150>;tag=d8edbb40
Call-ID: 84929ZGJkMjBiYTIwMWYzZGFiMDg0ZThiYmU4NWI2Zjg1NmQ
CSeq: 1 INVITE
Allow: SUBSCRIBE, NOTIFY, INVITE, ACK, CANCEL, BYE, REFER, INFO, OPTIONS, MESSAGE
Content-Type: application/sdp
Supported: replaces
User-Agent: Bria 4 release 4.8.1 stamp 84929
Content-Length: 211

v=0
o=- 13276799035054824 1 IN IP4 192.168.26.1
s=Bria 4 release 4.8.1 stamp 84929
c=IN IP4 192.168.26.1
t=0 0
m=audio 64192 RTP/AVP 9 0 101
a=rtpmap:101 telephone-event/8000
a=fmtp:101 0-15
a=sendrecv
```

#### Response:
```none
SIP/2.0 401 Unauthorized
Via: SIP/2.0/UDP 192.168.26.1:62128;rport=40850;received=127.0.0.1;branch=z9hG4bKa524287%00%01%02%03%04%05%06%07%08%09%0a%0b%0c%0d%0e%0f%10%11%12%13%14%15%16%17%18%19%1a%1b%1c%1d%1e%1f%20!%22%23%24%%26'%28%29*+%2c-.%2f0123456789%3a%3b%3c%3d%3e%3f%40ABCDEFGHIJKLMNOPQRSTUVWXYZ%5b%5c%5d%5e_`abcdefghijklmnopqrstuvwxyz%7b%7c%7d~%7f%80%81%82%83%84%85%86%87%88%89%8a%8b%8c%8d%8e%8f%90%91%92%93%94%95%96%97%98%99%9a%9b%9c%9d%9e%9f%a0%a1%a2%a3%a4%a5%a6%a7%a8%a9%aa%ab%ac%ad%ae%af%b0%b1%b2%b3%b4%b5%b6%b7%b8%b9%ba%bb%bc%bd%be%bf%c0%c1%c2%c3%c4%c5%c6%c7%c8%c9%ca%cb%cc%cd%ce%cf%d0%d1%d2%d3%d4%d5%d6%d7%d8%d9%da%db%dc%dd%de%df%e0%e1%e2%e3%e4%e5%e6%e7%e8%e9%ea%eb%ec%ed%ee%ef%f0%f1%f2%f3%f4%f5%f6%f7%f8%f9%fa%fb%fc%fd%fe%ff
Call-ID: 84929ZGJkMjBiYTIwMWYzZGFiMDg0ZThiYmU4NWI2Zjg1NmQ
From: "Gerrit" <sip:acct-206@192.168.26.150>;tag=d8edbb40
To: <sip:8675309@192.168.26.150>;tag=z9hG4bKa524287%00%01%02%03%04%05%06%07%08%09%0a%0b%0c%0d%0e%0f%10%11%12%13%14%15%16%17%18%19%1a%1b%1c%1d%1e%1f%20!%22%23%24%%26'%28%29*+%2c-.%2f0123456789%3a%3b%3c%3d%3e%3f%40ABCDEFGHIJKLMNOPQRSTUVWXYZ%5b%5c%5d%5e_`abcdefghijklmnopqrstuvwxyz%7b%7c%7d~%7f%80%81%82%83%84%85%86%87%88%89%8a%8b%8c%8d%8e%8f%90%91%92%93%94%95%96%97%98%99%9a%9b%9c%9d%9e%9f%a0%a1%a2%a3%a4%a5%a6%a7%a8%a9%aa%ab%ac%ad%ae%af%b0%b1%b2%b3%b4%b5%b6%b7%b8%b9%ba%bb%bc%bd%be%bf%c0%c1%c2%c3%c4%c5%c6%c7%c8%c9%ca%cb%cc%cd%ce%cf%d0%d1%d2%d3%d4%d5%d6%d7%d8%d9%da%db%dc%dd%de%df%e0%e1%e2%e3%e4%e5%e6%e7%e8%e9%ea%eb%ec%ed%ee%ef%f0%f1%f2%f3%f4%f5%f6%f7%f8%f9%fa%fb%fc%fd%fe%ff
CSeq: 1 INVITE
WWW-Authenticate: Digest realm="asterisk",nonce="1632330882/7a9dd893a589fc0bfee733218f9d99af",opaque="3a3ba3ac56e6322c",algorithm=md5,qop="auth"
Server: Asterisk PBX GIT-16-97ce647afd
Content-Length:  0
```

seanbright:
Fixes #2829

### DIFF
--- a/pjsip/src/pjsip/sip_parser.c
+++ b/pjsip/src/pjsip/sip_parser.c
@@ -220,7 +220,9 @@ PJ_INLINE(void) parser_get_and_unescape(pj_scanner *scanner, pj_pool_t *pool,
 #else
     PJ_UNUSED_ARG(unesc_spec);
     pj_scan_get(scanner, spec, token);
-    *token = pj_str_unescape(pool, token);
+    if (!pj_cis_match(spec, '%')) {
+        *token = pj_str_unescape(pool, token);
+    }
 #endif
 }
 


sauwming (dismissed):

pjsip/src/pjsip/sip_parser.c
```diff
@@ -220,7 +220,9 @@ PJ_INLINE(void) parser_get_and_unescape(pj_scanner *scanner, pj_pool_t *pool,
 #else
     PJ_UNUSED_ARG(unesc_spec);
     pj_scan_get(scanner, spec, token);
-    *token = pj_str_unescape(pool, token);
+    if (!pj_cis_match(spec, '%')) {
+        *token = pj_str_unescape(pool, token);
+    }
```
nanangizz, line 225: From the CI, looks like this fails pjsip unit test?
This function is used in many places so changing it directly does sound quite risky.

An alternative approach is to modify `int_parse_via_param()` implementation: parse the param value based on the param name, for example, param `branch` should use simply `pj_scan_get()` instead of `parser_get_and_unescape()`, while other params (including `other-param`) can use `parser_get_and_unescape()`. Need to be careful with quoted string (which is currently already taken care by `parse_param_imp()`.

Or do the param name check in the `parse_param_imp()`, which already handles quoted string, perhaps by introducing a new option such as `PJSIP_PARSE_VIA_MODE`, so when that option is set, [this block](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_parser.c#L1243) will not escape the param value if the param name is `branch`.

seanbright:
That would solve the problem for the branch parameter, but it seems that everything that uses the token production will suffer the same issue (tokens are permitted to contain "raw" `%` characters) unless I am not understanding something? I may not be the best person to attempt to resolve this bug.

nanangizz:
> That would solve the problem for the branch parameter, but it seems that everything that uses the token production will suffer the same issue (tokens are permitted to contain "raw" `%` characters) unless I am not understanding something? I may not be the best person to attempt to resolve this bug.

Yes, 'token' can have '%' that is not escaping char, but 'escaped' also has '%' in the [PJSIP scanner spec](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_parser.c#L45), so the patch will introduce issues for escaped strings (i.e: they will no longer unescaped, while they actually should).

IMHO, to avoid same issues on other parts (other than Via branch), we need to recheck all specs which use 'token' and make sure they are not unescaped.

sauwming:
So perhaps the fix should be:
- Not to delete '%' from `TOKEN_ESC` spec, since token is allowed to have '%' character.
- In 'parser_get_and_unescape()`, check for the unescaped spec instead to know if it's allowed to have '%'.

Something like this?
[token-esc-fix.txt](https://github.com/pjsip/pjproject/files/7567949/token-esc-fix.txt)

nanangizz:
> * Not to delete '%' from `TOKEN_ESC` spec, since token is allowed to have '%' character.

Not sure about this, the `*_ESC` version seems to be always without '%' (to be used by `pj_scan_get_unescape()`):
```
    pj_cis_t pjsip_TOKEN_SPEC_ESC;	/**< Token without '%' character */
```
Btw, IMO it is strange that `TOKEN_SPEC` has `_ESC` version, as the `%` char in that spec is literal (not escaper).


> * In 'parser_get_and_unescape()`, check for the unescaped spec instead to know if it's allowed to have '%'.

And I doubt we should touch `parser_get_and_unescape()`, as the function name has 'unescape' keyword, it should always unescape. IMO it is the app that should check whether it needs to use `parser_get_and_unescape()` or `pj_scan_get()`. So the problem starts in `parse_param_imp()` which always unescapes.



> The above patch passes the pjsip test, which has to be modified since it expects the generic param to be unescaped (according to the RFC 3261, generic param is a token as well:
> 
> ```
> generic-param  =  token [ EQUAL gen-value ]
> gen-value      =  token / host / quoted-string
> ```

Agree.

sauwming:
Attached is the modified patch:
[token-esc-fix2.txt](https://github.com/pjsip/pjproject/files/7569261/token-esc-fix2.txt)

Not sure whether we should remove `TOKEN_SPEC_ESC` and `VIA_PARAM_SPEC_ESC`. Checking the library, it doesn't seem to be used anywhere else.

nanangizz:
> Attached is the modified patch: [token-esc-fix2.txt](https://github.com/pjsip/pjproject/files/7569261/token-esc-fix2.txt)

Escaped spec seems to be equal to regular spec without '%', so this line seems hackish :) 
```
if (pj_cis_match(esc_spec, '%')) {
```
Btw, the "escaped spec" seems to be used by `pj_scan_get_unescape()` only.
Haven't checked the code, but what if we just pass `NULL` for the escaped spec param in parsing Via-params?
 
> Not sure whether we should remove `TOKEN_SPEC_ESC` and `VIA_PARAM_SPEC_ESC`. Checking the library, it doesn't seem to be used anywhere else.

Yes, we can remove/comment those perhaps, IMO they sound strange.

sauwming:
I continue this in #2933 to make the code review process easier.

PR #2832, opened 2021-09-24
unix pthread create, attribute not destroy, cause memory leak

jimying:
In freebsd platform， valgrind check it has memory leak.
After thread has created, need destroy the thread attribute that pthread_attr_init initialized.  

pthread_attr_init/pthread_attr_destroy man page say:
When a thread attributes object is no longer required, it should be destroyed using the pthread_attr_destroy() function.  Destroying a thread attributes object has no effect on threads that were created using that object.


jimying:
later new replace

PR #2836, opened 2021-09-27
One way audio due to Opus falsely detecting codec ptime change

nanangizz:
Reported that in a more than one hour long audio call using Opus, audio sometimes becomes one way, and the log file shows a lot of discarded packets. The log also has line ```codec decode ptime change detected``` but nothing like ```Opus ptime change detected: 10 ms --> 20 ms```.

After investigation, it looks like the Opus wrapper does not handle possible error returned by `opus_packet_get_nb_samples()`, which may cause false codec ptime change.

Thanks to Marcus Froeschl for the report and the analysis.

### DIFF
--- a/pjmedia/src/pjmedia-codec/opus.c
+++ b/pjmedia/src/pjmedia-codec/opus.c
@@ -867,19 +867,31 @@ static pj_status_t  codec_parse( pjmedia_codec *codec,
 					   ((unsigned char*)pkt) + out_pos,
 					   sizeof(tmp_buf));
 	if (size < 0) {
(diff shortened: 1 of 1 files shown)


sauwming (approved):
Right, according to the doc, `opus_packet_get_nb_samples()` can return `OPUS_INVALID_PACKET` indeed. But we're still not sure if this will solve the issue?

nanangizz:
> Right, according to the doc, `opus_packet_get_nb_samples()` can return `OPUS_INVALID_PACKET` indeed. But we're still not sure if this will solve the issue?

yes, still not sure if that's the root cause.

Issue #2103, opened 2018-03-27
Green screen in the beginning of video call


pjsipbot:
_2018-03-27 03:36:15: @nanangizz_ _created the issue on trac ticket [2103](https://trac.pjsip.org/repos/ticket/2103)_

In YUV format, if the buffer is all zero, it will be displayed as green (as compared to RGBA format, where it will be black). Unfortunately currently buffer is initialized with zero. As most video codecs uses I420 format, perhaps it is better to have special initialization for I420 so the screen will be black.

The allocation and initialization of frame buffer is done in `pjmedia/src/pjmedia/vid_port.c`, in function `pjmedia_vid_port_create()`.
___
_2018-08-20 06:30:03: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2018-08-20 06:30:03: @nanangizz_ _set **resolution** to **fixed**_

___
_2018-08-20 06:30:03: @nanangizz_ _commented_

In [r5862](../commit/f654c5aa0814c088df705492555c546f276c2791):
Close #2103: Initiate video port frame buffer with black for I420 & YV12 formats.

___
_2018-08-23 05:45:19: @mingteluu_ _commented_

In [r5866](../commit/34de9fcee00a48f34f17f8b55e6a2899468c911c):
Re #2103: Move the black screen initialization of the renderer to pjmedia_vid_port_start() instead

___
_2018-08-23 05:48:06: @mingteluu_ _commented_

In [r5867](../commit/657df2b09e3953b6bfacd0df594415d8343e5cb5):
Re #2103: Darwin's capture device is passive, thus the video port's clock will fetch the frames much earlier than when the device is ready, getting zero frames and resulting in green screen on the remote side.

___
_2018-08-28 03:00:17: @mingteluu_ _commented_

In [r5868](../commit/7db2028ec2314a1bdad62bcec173991ed682278e):
Re #2103:
A couple of improvements:
- Now initialize both video port's renderer AND capture buffer
- Support RGB format as well.

Known limitation: YUV currently only supports I420 and YV12

Issue #2811, opened 2021-09-01
Android Crash when load native lib pjsip, pjsua2 on android 7


bienpx224:
### Describe the bug

pjsip 2.11 built with bcg729, opus,. and everything do well but in andoird 7 ( api : 24) i got crash with error : Failed to load native lib pjsua2.
Failed to load native library pjsua2
    java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "__sendto_chk" referenced by "/data/app/com.sipsdk.demo-2/lib/x86/libpjsua2.so"...

### Steps to reproduce

run project pjsua2 

### PJSIP version

2.11

### Context

pjsip 2.11 built with bcg729, opus,. and everything do well but in andoird 7 ( api : 24) i got crash with error : Failed to load native lib pjsua2.

### Log, call stack, etc

```shell
Failed to load native library pjsua2
    java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "__sendto_chk" referenced by "/data/app/com.sipsdk.demo-2/lib/x86/libpjsua2.so"...
```

PR #2864, opened 2021-10-27
Set black color for I420 & YV12 formats if video frame buffer is empty.

klnvch:
Problem: At the begging of a video call the first few frames have buffer filled with zeroes, the next few frames have buffer with values just slightly above zero and only after that expected camera output is rendered.
For YUV format empty frame buffer causes green screen. This enhancement is closely related to https://github.com/pjsip/pjproject/issues/2103 . 
Solution: If all pixels are in range [0-64, 0-64, 0-64] for YUV format (RGB [0,67-199,0]), then frame buffer is filled with black color.

### DIFF
--- a/pjmedia/src/pjmedia/vid_port.c
+++ b/pjmedia/src/pjmedia/vid_port.c
@@ -1346,6 +1346,35 @@ static pj_status_t vid_pasv_port_put_frame(struct pjmedia_port *this_port,
         if (status != PJ_SUCCESS)
             return status;
 
+        /* Initialize buffer with black color if it is empty */
+        {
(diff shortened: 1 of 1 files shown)


klnvch:
@sauwming I agree with you that it is strange to reduce YUV color space. I assumed that some YUV values is not possible to get from real camera, but I need to check it. Probably another solution will be skipping zero buffers on a transmitter side, that after coding-decoding starts to have some minor values. (I am not sure if it is the reason here).

sauwming:
This should be fixed by #2811. Let us know if there's still any issue.

Issue #2870, opened 2021-10-31
avi play crash


jimying:
### Describe the bug

Use aviplay sample to play an avi file,  when reach to end , it will crash.

### Steps to reproduce

1. Play a short avi file  (I use [panda.avi](http://a.jimying.fun/video/panda.avi))
2. Wait file reach the end  (before rewinding)

### PJSIP version

master (2.11-dev)

### Context

I test it on  Mac, Windows

### Log, call stack, etc

```shell
Mac lldb log
x86_64-apple-darwin18.7.0 jimying$ lldb ./aviplay -- ~/Downloads/panda.avi
(lldb) target create "./aviplay"
Current executable set to './aviplay' (x86_64).
(lldb) settings set -- target.run-args  "/Users/jimying/Downloads/panda.avi"
(lldb) r
Process 87322 launched: '/Users/jimying/work/pjproject/pjsip-apps/bin/samples/x86_64-apple-darwin18.7.0/aviplay' (x86_64)
21:34:09.172         os_core_unix.c !pjlib 2.11-dev for POSIX initialized
2021-10-31 21:34:09.200669+0800 aviplay[87322:3271180]   no saved enable hardware sample rate converter preference found
21:34:09.263           darwin_dev.m  Darwin video initialized with 1 devices:
21:34:09.263           darwin_dev.m   0: [Capturer] AVF - Webcam C170
21:34:09.284              sdl_dev.c  SDL 2.0 initialized
21:34:09.284         colorbar_dev.c  Colorbar video src initialized with 2 device(s):
21:34:09.284         colorbar_dev.c   0: Colorbar generator
21:34:09.284         colorbar_dev.c   1: Colorbar-active
21:34:09.285           avi_player.c  The AVI file has 1 streams.
21:34:09.285           avi_player.c  AVI file player '/Users/jimying/Downloads/panda.avi' created with 1 media ports
21:34:09.286             vid_port.c  Opening device SDL renderer [SDL] for render: format=I420, size=768x432 @23:1 fps
2021-10-31 21:34:09.357742+0800 aviplay[87322:3271169] MessageTracer: load_domain_whitelist_search_tree:73: Search tree file's format version number (0) is not supported
2021-10-31 21:34:09.357789+0800 aviplay[87322:3271169] MessageTracer: Falling back to default whitelist
21:34:09.358             vid_port.c  Device SDL renderer [SDL] opened: format=I420, size=768x432 @23:1 fps
[...]

PR #2871, opened 2021-10-31
fix avi play crash

jimying:
as title
ref to issue #2870 
I have test it okay on Mac & Windows

### DIFF
--- a/pjmedia/src/pjmedia/avi_player.c
+++ b/pjmedia/src/pjmedia/avi_player.c
@@ -897,12 +897,15 @@ static pj_status_t avi_get_frame(pjmedia_port *this_port,
 
 on_error2:
     if (status == AVI_EOF) {
+        fport->eof = PJ_TRUE;
+        if (size_read > 0)
+        {
         size_to_read -= size_read;
         pj_bzero((char *)frame->buf + frame->size - size_to_read,
                  size_to_read);
-        fport->eof = PJ_TRUE;
-
         return PJ_SUCCESS;
+        }
+        return PJ_EEOF;
     }
 
     return status;


silentindark:
@sauwming @nanangizz  Could you please review this change?

sauwming:
I cannot play the provided sample avi. If you can convert it to `h264`, it would be great.
If not, can you tell us more about the value of `size_read` when the issue happened and if it is negative, why it could happen?

jimying:
> I cannot play the provided sample avi. If you can convert it to `h264`, it would be great.
> If not, can you tell us more about the value of `size_read` when the issue happened and if it is negative, why it could happen?


sorry  for delay.  please visit http://a.jimying.fun/video/
panda.h264.avi : is h264 format avi
panda1.mp4 is the raw mp4 file

jimying:
> I cannot play the provided sample avi. If you can convert it to `h264`, it would be great. If not, can you tell us more about the value of `size_read` when the issue happened and if it is negative, why it could happen?

when crash, size_read = 0, I modify to debug, log as follow: 
```
20:47:07.915           avi_player.c  ------------> size_read = 0
20:47:07.921                 ffmpeg  mpeg4: header damaged
20:47:07.921    ffmpeg_vid_codecs.c  ffmpeg err -1094995529: Invalid data found when processing input
20:47:07.921              aviplay.c  codec_get_frame() error: Unknown error 220087
20:47:07.961           avi_player.c  File port /Users/jimying/Downloads/panda.avi EOF
20:47:07.962           avi_player.c  File port /Users/jimying/Downloads/panda.avi rewinding..
Segmentation fault: 11
```

sauwming:
When I tried to play `panda.h264.avi` using OpenH264, it would fail much earlier during initialization, which I suppose due to the damaged header(?):
```
11:19:35.221         [OPENH264_LOG]  [L1] [OpenH264] this = 0x0x6000019a71b0, Error:Invalid bitrate settings in total configure, bitrate= 0
11:19:35.221         [OPENH264_LOG]  [L1] [OpenH264] this = 0x0x6000019a71b0, Error:WelsInitEncoderExt(), ParamValidationExt failed return 16.
11:19:35.221         [OPENH264_LOG]  [L1] [OpenH264] this = 0x0x6000019a71b0, Error:CWelsH264SVCEncoder::Initialize(), WelsInitEncoderExt failed.
```

Thanks for the additional info regarding the value of `size_read`. I still don't understand thou, how a value of zero (and not negative as I thought earlier) can lead to a crash. The following code should be fine even when `size_read == 0`:
```
        size_to_read -= size_read;
         pj_bzero((char *)frame->buf + frame->size - size_to_read,
                  size_to_read);
```
So perhaps you can provide us with the stack trace of the crash?

jimying:
@sauwming 
I have no environment for now， I will supply the crash stack later

or 
I build pjsip video support use ffmpeg h264, not openH264.  You can build again use ffmpeg h264, and trace again ^_^

sauwming:
Okay, I could reproduce it here and it is actually our ffmpeg codec wrapper "bug". https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c#L1832:
```
    /* ffmpeg warns:
     * - input buffer padding, at least FF_INPUT_BUFFER_PADDING_SIZE
     * - null terminated
     * Normally, encoded buffer is allocated more than needed, so lets just
     * bzero the input buffer end/pad, hope it will be just fine.
     */
#if LIBAVCODEC_VER_AT_LEAST(56,35)
    pj_bzero(avpacket.data+avpacket.size, AV_INPUT_BUFFER_PADDING_SIZE);
```

So when the input frame size is already at the maximum, which is the case of AVI_EOF in `avi_player.c`, the `pj_bzero()` above will cause buffer overwrite. In my test, it will overwrite the vid_port's stream to `NULL` and cause the crash in `pjmedia_vid_dev_stream_put_frame(strm, ...)` since `strm` is `NULL`, exactly as you reported in #2870.

So the fix you proposed is not quite correct, although it does prevent the issue from happening (returning `PJ_EEOF` will prevent the call to `pjmedia_vid_codec_decode()` in `aviplay`:
```
    status = pjmedia_port_get_frame(port_data->src_port, &enc_frame);
    if (status != PJ_SUCCESS) goto on_error;
    
    status = pjmedia_vid_codec_decode(codec, 1, &enc_frame,
                                      (unsigned)frame->size, frame);
```
which of course prevents the problematic ffmpeg decode above from being invoked.

Will update you again when I figure out the proper fix for this.

sauwming:
My proposed fix is in PR #2940.

jimying:
@sauwming  Yes，you find the final reason.  
But i think that if read video file EOF,  pjmedia_port_get_frame should return an error code (PJ_EOF)  and no need continue to decode

sauwming:
We should still try to decode whatever has been read since it's possible the decode can still be successful, for example in the case that the file is slightly corrupted at the end.

What we can do is to skip the decode if the frame is completely empty. I just updated the PR.

jimying:
@sauwming  I have see the pr,  see no problem

PR #2874, opened 2021-11-02
fix clock thread lock not release

jimying:
clock thread,  some case lock not release  when join thread


silentindark:
@sauwming Could you please review this change?

sauwming (approved):

nanangizz (approved):

PR #2877, opened 2021-11-04
fix getaddrinfo call on windows

jurica:
When trying to connect to a FQDN on Windows with pjsip 2.11 the connection cannot be established with a PJ_ERESOLVE error. Debugging the issue showed, that the getaddrinfo result is not checked properly.
This patch solved the issue for me and might be integrated into master.


jurica:
Nevermind I just saw that it was fixed in the meanwhile.

PR #2878, opened 2021-11-05
Change interval calculation for sending keep-alive message

trengginas:
Currently keep-alive interval is calculated using frame timestamp. 
If somehow there's a timestamp jump or rollover, the calculation would be incorrect. This patch will use time of day (`pj_gettimeofday()`) instead to calculate the interval.
This ticket will also add `ka_interval` setting to `pjmedia_stream_ka_config` to enable users specify the keep-alive interval instead of specifying it using `PJMEDIA_STREAM_KA_INTERVAL`.

### DIFF
--- a/pjmedia/include/pjmedia/stream_common.h
+++ b/pjmedia/include/pjmedia/stream_common.h
@@ -58,8 +58,7 @@ typedef struct pjmedia_stream_rtp_sess_info
 #if defined(PJMEDIA_STREAM_ENABLE_KA) && PJMEDIA_STREAM_ENABLE_KA!=0
 
 /**
- * Structure of configuration settings for stream keepalive after it
- * is created.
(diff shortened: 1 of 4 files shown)


pjmedia/include/pjmedia/stream_common.h
```diff
@@ -79,6 +78,14 @@ typedef struct pjmedia_stream_ka_config
      */
     unsigned			    start_interval;
 
+    /**
+     * The keepalive sending interval, after #start_count number keepalive 
+     * was sent.
+     * 
+     * Defautt: PJMEDIA_STREAM_KA_INTERVAL (seconds)
```
andreas-wehrmann, line 85: Typo here: Defautt -> Default

nanangizz (approved):

sauwming (approved):

PR #2880, opened 2021-11-08
Handle multiple handshake fail on TLS connection

trengginas:
When accepting a new connection as a server, there's a scenario where multiple read event was received.
This will lead to multiple handshake fail, and on multi worker thread settings might lead to crash.
```
02-11 08:50:41,030 ssl0x7f9810051390 !Certificate chain loaded from buffer
02-11 08:50:41,032 ssl0x7f9810051390  CA certificates loaded from buffer (cnt=9)
02-11 08:50:41,032 ssl0x7f9810051390  Private key loaded from buffer
02-11 08:50:41,033 ssl0x7f9810051390  CA certificates loaded from buffer
02-11 08:50:41,033 SSL  SSL_ERROR_SSL (Handshake): Level: 0 err: <336130315> <SSL routines-ssl3_get_record-wrong version number> len: 0 peer: 
02-11 08:50:41,033 ssl0x7f9810051390  Handshake failed in accepting 89.248.165.50:63040: wrong version number
02-11 08:50:41,033 SSL !BIO error, SSL_ERROR_SYSCALL (Handshake): errno: <11> <Resource temporarily unavailable> len: 0
02-11 08:50:41,033 ssl0x7f9810051390  Handshake failed in accepting : DH lib
```

This is due to this code:
```
	    if (ssock->parent->param.grp_lock) {
		pj_grp_lock_dec_ref(ssock->parent->param.grp_lock);
		ssock->parent = NULL;
	    }
```
This patch will add check if `ssock->parent` has been reset to avoid the crash.

### DIFF
--- a/pjlib/src/pj/ssl_sock_imp_common.c
+++ b/pjlib/src/pj/ssl_sock_imp_common.c
@@ -257,14 +257,14 @@ static pj_bool_t on_handshake_complete(pj_ssl_sock_t *ssock,
     if (ssock->is_server) {
 	pj_bool_t ret = PJ_TRUE;
 
+	if (!ssock->parent) {
+	    return PJ_FALSE;
+	}
+
 	if (status != PJ_SUCCESS) {
 	    /* Handshake failed in accepting, destroy our self silently. */
 
 	    char buf[PJ_INET6_ADDRSTRLEN+10];
-
-	    if (!ssock->parent) {
-		return PJ_FALSE;
-	    }
 
 	    PJ_PERROR(3,(ssock->pool->obj_name, status,
 			 "Handshake failed in accepting %s",
--- a/pjlib/src/pj/ssl_sock_imp_common.c
+++ b/pjlib/src/pj/ssl_sock_imp_common.c
@@ -262,6 +262,10 @@ static pj_bool_t on_handshake_complete(pj_ssl_sock_t *ssock,
 
 	    char buf[PJ_INET6_ADDRSTRLEN+10];
 
+	    if (!ssock->parent) {
+		return PJ_FALSE;
+	    }
+
 	    PJ_PERROR(3,(ssock->pool->obj_name, status,
 			 "Handshake failed in accepting %s",
 			 pj_sockaddr_print(&ssock->rem_addr, buf,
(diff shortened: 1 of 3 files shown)


pjlib/src/pj/ssl_sock_imp_common.c
```diff
@@ -339,7 +343,7 @@ static pj_bool_t on_handshake_complete(pj_ssl_sock_t *ssock,
 	/* Decrement ref count of parent and reset parent (we don't need it
 	 * anymore, right?).
 	 */
-	if (ssock->parent->param.grp_lock) {
+	if (ssock->parent && ssock->parent->param.grp_lock) {
```
sauwming, line 346: If `ssock->parent` is NULL it will also cause issues when calling the callaback `on_accept_complete/complete2()` above.

sauwming:
If there can be more than one read errors during state `SSL_STATE_HANDSHAKING`, I believe checking `ssock->parent` may not be enough since we can't know where exactly the race will happen. This can potentially cause issues such as:
- the callback is called more than once, possibly with `ssock->parent` parameter already set to NULL.
- and more critically, the group lock's ref can be decreased more than once.

trengginas:
> If there can be more than one read errors during state `SSL_STATE_HANDSHAKING`, I believe checking `ssock->parent` may not be enough since we can't know where exactly the race will happen. This can potentially cause issues such as:
> 
> * the callback is called more than once, possibly with `ssock->parent` parameter already set to NULL.
> * and more critically, the group lock's ref can be decreased more than once.

One alternative is to make use of the `write_mutex` and check if there's error on the previous handshake.  This will also avoid calling the callback and decreasing the group lock

sauwming:
Our current OpenSSL implementation of `ssl_do_handshake()` looks like this (some code are intentionally omitted for clarity):
```
    /* Perform SSL handshake */
    err = SSL_do_handshake(ossock->ossl_ssl);

    status = flush_circ_buf_output(ssock, &ssock->handshake_op_key, 0, 0);

    if (err < 0) {
	/* Handshake fails */
	status = STATUS_FROM_SSL_ERR2("Handshake", ssock, err, err2, 0);
```

The log:
```02-11 08:50:41,033 SSL  SSL_ERROR_SSL (Handshake): Level: 0 err: <336130315> <SSL routines-ssl3_get_record-wrong version number> len: 0 peer:```
must have come from the above code: `STATUS_FROM_SSL_ERR2("Handshake", ssock, err, err2, 0)`

But since we also call `flush_circ_buf_output()`, this will trigger another error:
```02-11 08:50:41,033 SSL !BIO error, SSL_ERROR_SYSCALL (Handshake): errno: <11> <Resource temporarily unavailable> len: 0```

So there can be problems if there are multiple threads doing handshake at the same time and return error. Unfortunately, there doesn't seem to be any easy fix and modifying the handshake flow seems risky.

So I will close this and migrate it to a new issue.

PR #2882, opened 2021-11-12
Fix on_clock_tick function  will cause crash

rolandyue:
/*if not got_frame,frame.buf is null,and pjmedia_port_put_frame not check it.will cause crash when convert*/
	   //status = pjmedia_port_put_frame(sink->port, &frame);


sauwming:
The `vid_conf` source has changed due to #2911. Please test whether it fixes the issue and reopen/resubmit if the problem persists.

PR #2886, opened 2021-11-16
Check for decoding delay in video stream

sauwming:
For video, the jitter buffer has a much larger capacity (compared to audio) to accommodate a large video frame that can be split into many chunks. However, typically only the keyframes are large in size while the rest are relatively small. Thus the jitter buffer can store video frames that are much longer than max latency/delay specified in jb_max setting (in one local testing, it can easily keep 6-7 seconds of video for a default setting of 500 ms jb_max). Because of this, video rendering latency can potentially be larger than expected/desired.

Currently, the video stream checks if the jitter buffer is full to decide if it's time to decode. While this works for audio, as explained above, this is not enough for video. So in this PR, we propose to check the timestamp difference as well in order to know if it's time to decode.

A simple way to test this issue is by adding a sleep/delay in the local rendering process.

### DIFF
--- a/pjmedia/src/pjmedia/vid_stream.c
+++ b/pjmedia/src/pjmedia/vid_stream.c
@@ -760,7 +760,7 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
     unsigned payloadlen;
     pjmedia_rtp_status seq_st;
     pj_status_t status;
-    unsigned ts_diff;
+    long ts_diff;
     pj_bool_t pkt_discarded = PJ_FALSE;
 
     /* Check for errors */
--- a/pjmedia/src/pjmedia/vid_stream.c
+++ b/pjmedia/src/pjmedia/vid_stream.c
@@ -137,6 +137,7 @@ struct pjmedia_vid_stream
     pjmedia_ratio	     dec_max_fps;   /**< Max fps of decoding dir.   */
     pjmedia_frame            dec_frame;	    /**< Current decoded frame.     */
     unsigned		     dec_delay_cnt; /**< Decoding delay (in frames).*/
+    unsigned		     dec_max_delay; /**< Decoding max delay (in ts).*/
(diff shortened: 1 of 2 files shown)


pjmedia/src/pjmedia/vid_stream.c
```diff
@@ -1902,16 +1916,20 @@ PJ_DEF(pj_status_t) pjmedia_vid_stream_create(
...
-    if (info->jb_max >= frm_ptime)
+    if (info->jb_max >= frm_ptime) {
 	jb_max	    = info->jb_max * chunks_per_frm / frm_ptime;
-    else
+	stream->dec_max_delay = info->codec_info.clock_rate * info->jb_max /
+				1000;
+    } else {
 	jb_max	    = 500 * chunks_per_frm / frm_ptime;
+	stream->dec_max_delay = info->codec_info.clock_rate * 500 / 1000;
+    }
```
nanangizz, line 1932: Looks like audio also uses 500ms as default jb_max, perhaps we need to change the pjsua docs too (which specifies 360ms):
```
    /**
     * Set maximum delay that can be accomodated by the jitter buffer msec.
     *
     * Default: -1 (to use default stream settings, currently 360 msec)
     */
    int			jb_max;
```

pjmedia/src/pjmedia/vid_stream.c
```diff
@@ -759,6 +760,7 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
     unsigned payloadlen;
     pjmedia_rtp_status seq_st;
     pj_status_t status;
+    unsigned ts_diff;
```
nanangizz, line 763: What if it is signed (and perhaps uses `long`), to handle ts rollover better?

nanangizz (approved):
Looks okay, just minor below.

trengginas (approved):

PR #2891, opened 2021-11-18
Disable assertion when checking for an invalid transport

trengginas:
During testing on OpenSIPIt'02, when testing with SIPVicious an assertion is raised [here](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_transport.c#L361) as it sends invalid transport 
name on the message.

```c
    /* Get transport type from name. */
    for (i=0; i<PJ_ARRAY_SIZE(transport_names); ++i) {
	if (pj_stricmp(name, &transport_names[i].name) == 0) {
	    return transport_names[i].type;
	}
    }

    pj_assert(!"Invalid transport name");
    return PJSIP_TRANSPORT_UNSPECIFIED;
```
This ticket will comment out such assertion and just return instead.

Note:
The assertion will not be raised when assertion is disabled (e.g: release version).

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -288,7 +288,9 @@ static struct transport_names_t *get_tpname(pjsip_transport_type_e type)
 	if (transport_names[i].type == type)
 	    return &transport_names[i];
     }
-    pj_assert(!"Invalid transport type!");
+
+    /* Just return intstead of raising assertion. */
+    //pj_assert(!"Invalid transport type!");
     return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsip/sip_transport.c
```diff
@@ -358,7 +360,8 @@ PJ_DEF(pjsip_transport_type_e) pjsip_transport_get_type_from_name(const pj_str_t
 	}
     }
 
-    pj_assert(!"Invalid transport name");
+    /* Just return intstead of raising assertion. */
+    //pj_assert(!"Invalid transport name");
```
nanangizz, line 364: - Typo "intstead".
- IMO only modify this function and leave the others (don't fix it if it ain't broke :). Note that the transport name string may be parsed from incoming SIP message (which is not controlled by programmer), while in other functions, transport type/flag params are set programmatically (under programmer's control), so assertion should be the correct way already.

sauwming (approved):

nanangizz (approved):

PR #2899, opened 2021-11-25
Deadly overflow causing hang after running 106 days

BinWu-859:
If elapsed in elapsed_msec() reach 9223372036854776 which stands for around 106 days, after the calculation, the elapsed will overflow to -9223372036854775.
And this will cause troubles in timer calculation(refer to pj_gettickcount(), comparing monotonic tick with 0), the timer of upper-caller in pjsip will be postponed for 106 days.
Our software end users have to reboot the device every 3 months.

### DIFF
--- a/pjlib/src/pj/os_timestamp_common.c
+++ b/pjlib/src/pj/os_timestamp_common.c
@@ -65,16 +65,19 @@ static pj_highprec_t elapsed_msec( const pj_timestamp *start,
     freq += ts_freq.u32.lo;
 #endif
 
-    /* Avoid division by zero. */
-    if (freq == 0) freq = 1;
-
     /* Get elapsed time in cycles. */
     elapsed = get_elapsed(start, stop);
 
     /* usec = elapsed * MSEC / freq */
     pj_highprec_div(freq, MSEC);
-    pj_highprec_div(elapsed, freq);
-
+     
+    /* Avoid division by zero. */
+    if (freq == 0) {
+      /* Regard freq = 1 */
+      pj_highprec_mul(elapsed, MSEC);
+    } else {
+      pj_highprec_div(elapsed, freq);
+    }
     return elapsed;
 }
 
--- a/pjlib/src/pj/os_timestamp_common.c
+++ b/pjlib/src/pj/os_timestamp_common.c
@@ -72,7 +72,7 @@ static pj_highprec_t elapsed_msec( const pj_timestamp *start,
     elapsed = get_elapsed(start, stop);
 
     /* usec = elapsed * MSEC / freq */
-    pj_highprec_mul(elapsed, MSEC);
+    pj_highprec_div(freq, MSEC);
     pj_highprec_div(elapsed, freq);
 
     return elapsed;


silentindark:
@sauwming @trengginas @nanangizz What you think about this fix? Maybe add this patch to master branch?

pjlib/src/pj/os_timestamp_common.c
```diff
@@ -72,7 +72,7 @@ static pj_highprec_t elapsed_msec( const pj_timestamp *start,
     elapsed = get_elapsed(start, stop);
 
     /* usec = elapsed * MSEC / freq */
-    pj_highprec_mul(elapsed, MSEC);
+    pj_highprec_div(freq, MSEC);
```
sauwming, line 75: There's a possibility that freq is 1 (see above), so better to add some check here.

BinWu-859: Yes, it will have problem if freq is less than MSEC
If so , keep the original calculation.

sauwming (approved):

pjlib/src/pj/os_timestamp_common.c
nanangizz: This line may degrade the accuracy significantly, especially when high-prec type is integer, e.g (extreme example):
```
freq = 1999
freq/MSEC = 1999/1000 = 1

elapsed = 1999
return value = elapsed/freq = 1999ms (while it should be 1000ms!)
```
Also note that there are also `elapsed_usec() & pj_elapsed_nanosec()`, the same method will degrade accuracy even further.

Here is an idea:
```
    /* Original formula: usec = elapsed * MSEC / freq
     * To avoid overflow when the elapsed number is high, the calculation is
     * done in two steps: seconds and sub-second parts.
     */
    if (elapsed <= freq) { // ideally the condition is based on high prec type
	pj_highprec_mul(elapsed, MSEC);
	pj_highprec_div(elapsed, freq);
    } else {
	pj_highprec_t mod = elapsed;

	/* Seconds part */
	pj_highprec_div(elapsed, freq);
	elapsed = pj_highprec_mul(elapsed, MSEC);

	/* Sub-second part */
	pj_highprec_mod(mod, freq);
	pj_highprec_mul(mod, MSEC);
	pj_highprec_div(mod, freq);

	/* Combine the parts */
	elapsed += mod;
    }
```
The cons is more calculation steps needed.

sauwming: Looking at the implementation of `pj_get_timestamp_freq()`, I believe `freq` should be either 0, or at least 1,000,000:

Posix implementation:
```
	machine_speed_mhz = get_machine_speed_mhz();
	if (machine_speed_mhz > 0) {
	    machine_speed.u64 = machine_speed_mhz * 1000000.0;
	}
```
Windows:
```
    CpuHz = CpuMhz;
    CpuHz = CpuHz * 1000000;
```

nanangizz: Yes, AFAIK it is basically the CPU clock. For `elapsed_msec()` perhaps the approach is acceptable, but perhaps not for `elapsed_usec()`  or even `elapsed_nanosec()`, note that they also suffer from the same issue, so I guess we need to update/fix them too.

BinWu-859: So far `elapsed_xxx()` are mainly used to calculate short intervals, which will not trigger the overflow.
Only `pj_gettickcount()` who takes 0 as `start` will have the risk.

nanangizz: Ah okay, that makes sense actually, didn't realize that the main problem is `pj_gettickcount()`.

nanangizz (approved):

sauwming (approved):

PR #2900, opened 2021-11-25
Support loop media transport in PJSUA2

wosrediinanatour:
This PJSUA feature is useful with third-party media libraries and thus useful in PJSUA2.
This commit provides two PJSUA account settings in PJSUA2.

### DIFF
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -1016,6 +1016,24 @@ struct AccountMediaConfig : public PersistentObject
      */
     RtcpFbConfig	rtcpFbConfig;
 
+    /**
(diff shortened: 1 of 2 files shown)


wosrediinanatour:
7 failing checks seem to be a "SSL problem" and "slept for 243 ms instead of 200 ms (outside 20% err window)" - so seems to be unrelated.

sauwming (approved):

nanangizz (approved):

nanangizz:
> 7 failing checks seem to be a "SSL problem" and "slept for 243 ms instead of 200 ms (outside 20% err window)" - so seems to be unrelated.

Yes, seems unrelated indeed, will check later.
Thank you for the patch.

Issue #2901, opened 2021-11-25
50 kB leaked per call when ICE transport is re-used


SamVanBelle:
### Describe the bug

Dear PJSIP support,

We developed an IVR based on PJSIP, and followed the recommendation to re-use media-transports (to avoid closing & opening sockets all the time).
This works fine for standard UDP transport, but causes a huge memory leak per call for ICE transport (wrapped in SRTP).
Each call leaks at least 49880 bytes (the size of the pj_ice_sess struct), because the pool on which it is allocated, is only freed by closing the transport.
The pool is released in ice_on_destroy, which is protected by a grp_lock comming from pj_ice_strans.

Can you please solve this leak?

Best regards,
Sam

Sam Van Belle
www.xqting.be

### Steps to reproduce

Make a call with an SAVP media descriptor, created by pjmedia_transport_media_create on an ICE/SRTP transport.
After the call stop the transport with pjmedia_transport_media_stop, but don't close it.

### PJSIP version

2.11

### Context

Happens on both Windows & Linux.

### Log, call stack, etc

```shell
- SipStackTest.exe!pj_ice_sess_create(pj_stun_config * stun_cfg, const char * name, pj_ice_sess_role role, unsigned int comp_cnt, const pj_ice_sess_cb * cb, const pj_str_t * local_ufrag, const pj_str_t * local_passwd, pj_grp_lock_t * grp_lock, pj_ice_sess * * p_ice) Line 356	C
- SipStackTest.exe!pj_ice_strans_init_ice(pj_ice_strans * ice_st, pj_ice_sess_role role, const pj_str_t * local_ufrag, const pj_str_t * local_passwd) Line 1283	C
- SipStackTest.exe!transport_media_create(pjmedia_transport * tp, pj_pool_t * sdp_pool, unsigned int options, const pjmedia_sdp_session * rem_sdp, unsigned int media_index) Line 1869	C
- SipStackTest.exe!pjmedia_transport_media_create(pjmedia_transport * tp, pj_pool_t * sdp_pool, unsigned int options, const pjmedia_sdp_session * rem_sdp, unsigned int media_index) Line 917	C
- SipStackTest.exe!transport_media_create(pjmedia_transport * tp, pj_pool_t * sdp_pool, unsigned int options, const pjmedia_sdp_session * sdp_remote, unsigned int media_index) Line 1542	C
[...]

PR #2903, opened 2021-11-30
Avoid memory bloat on ICE restart

nanangizz:
On ICE restart, ICE session will be recreated. However the ICE session's pool will not be released immediately, it will only be released when the ICE stream transport is destroyed, this is due to group lock.

Similar issue also occurs on ICE transport reuse scenario (#2901).

### DIFF
--- a/pjnath/src/pjnath/ice_strans.c
+++ b/pjnath/src/pjnath/ice_strans.c
@@ -209,6 +209,7 @@ struct pj_ice_strans
     pj_ice_strans_state	     state;	/**< Session state.		*/
     pj_ice_sess		    *ice;	/**< ICE session.		*/
     pj_ice_sess		    *ice_prev;	/**< Previous ICE session.	*/
+    pj_grp_lock_handler	     ice_prev_hndlr; /**< Handler of prev ICE	*/
     pj_time_val		     start_time;/**< Time when ICE was started	*/
 
     unsigned		     comp_cnt;	/**< Number of components.	*/
@@ -986,6 +987,12 @@ static void ice_st_on_destroy(void *obj)
 {
     pj_ice_strans *ice_st = (pj_ice_strans*)obj;
 
+    /* Destroy any previous ICE session */
+    if (ice_st->ice_prev) {
+	(*ice_st->ice_prev_hndlr)(ice_st->ice_prev);
+	ice_st->ice_prev = NULL;
+    }
+
     PJ_LOG(4,(ice_st->obj_name, "ICE stream transport %p destroyed", obj));
 
     /* Done */
@@ -1019,6 +1026,7 @@ static void destroy_ice_st(pj_ice_strans *ice_st)
     if (ice_st->ice) {
 	ice_st->ice_prev = ice_st->ice;
 	ice_st->ice = NULL;
+	pj_ice_sess_detach_grp_lock(ice_st->ice_prev, &ice_st->ice_prev_hndlr);
 	pj_ice_sess_destroy(ice_st->ice_prev);
     }
 
@@ -1286,7 +1294,7 @@ PJ_DEF(pj_status_t) pj_ice_strans_init_ice(pj_ice_strans *ice_st,
      * (due to group lock).
      */
     if (ice_st->ice_prev) {
-	pj_pool_safe_release(&ice_st->ice_prev->pool);
+	(*ice_st->ice_prev_hndlr)(ice_st->ice_prev);
 	ice_st->ice_prev = NULL;
     }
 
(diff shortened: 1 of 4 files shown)


sauwming:
I am not very fond of this approach:
1. It releases a pool that belongs to another object. The pool was not created by `ice_strans` so `ice_strans` shouldn't be the one destroying it.
2. Won't this cause double destroy? (i.e. the group lock's handler will still call `ice_on_destroy()` and release the pool again?)

nanangizz:
> I am not very fond of this approach:
> 
> 1. It releases a pool that belongs to another object. The pool was not created by `ice_strans` so `ice_strans` shouldn't be the one destroying it.

The ICE session, who created the pool, has been destroyed (other objects should see it as destroyed/NULL), so virtually the pool belongs to group lock handler. Not sure if adding API to access group lock handler is a good approach too (note that it will be ICE strans that accessing/querying/initiating the handler, not ICE session itself).

Introducing & invoking `ice_sess_pool_release(ice_sess /*==NULL*/)` does not seem make sense either.

The main idea is to avoid immediate ICE session's pool release, too risky, e.g: race between STUN socket receive callback vs ICE session destroy. This is one of the objective of group lock integration, i.e: synchronizing pool release.

So any idea?

> 2. Won't this cause double destroy? (i.e. the group lock's handler will still call `ice_on_destroy()` and release the pool again?)

Yes, the handler will still be called, but safe release should avoid double release.

sauwming:
> The main idea is to avoid immediate ICE session's pool release, too risky, e.g: race between STUN socket receive callback vs ICE session destroy. This is one of the objective of group lock integration, i.e: synchronizing pool release.
> 
> So any idea?

Perhaps we can use the current idea (i.e. storing the previous ICE session) and add 2 APIs: ice_sess_stop() and ice_sess_destroy2( bool immediate/force)

> > 2. Won't this cause double destroy? (i.e. the group lock's handler will still call `ice_on_destroy()` and release the pool again?)
> Yes, the handler will still be called, but safe release should avoid double release.

But if the released pool memory has been reused for another object, I believe it will still crash.

nanangizz:
> Perhaps we can use the current idea (i.e. storing the previous ICE session) and add 2 APIs: ice_sess_stop() and ice_sess_destroy2( bool immediate/force)

Still not sure about `destroy2(immediate/force)`, ICE session is supposed not to need (or even can't have) immediate/force destroy.

> But if the released pool memory has been reused for another object, I believe it will still crash.

Ah, you're right.

Perhaps `pj_ice_sess_detach_grp_lock()` which will delete ICE sess handler & return the handler address & member, and later the handler will be invoked by ICE strans?

pjnath/src/pjnath/ice_strans.c
```diff
@@ -1016,8 +1024,10 @@ static void destroy_ice_st(pj_ice_strans *ice_st)
 
     /* Destroy ICE if we have ICE */
     if (ice_st->ice) {
-	pj_ice_sess_destroy(ice_st->ice);
+	ice_st->ice_prev = ice_st->ice;
 	ice_st->ice = NULL;
+	pj_ice_sess_detach_grp_lock(ice_st->ice_prev, &ice_st->ice_prev_hndlr);
+	pj_ice_sess_destroy(ice_st->ice_prev);
     }
```
sauwming, line 1031: I think this block can be left unchanged.

sauwming (commented):
Also make sure to test with multiple ICE restart scenario (such as in the case of reinvite without SDP) and check if there's no memory leak nor double destroy.

pjnath/src/pjnath/ice_strans.c
nanangizz: Setting `ice_st->ice` to `NULL` before calling `pj_ice_sess_destroy()` is important to avoid stray crash (IIRC, ICE strans uses `ice_st->ice != NULL` checks in many parts)?

sauwming: ah, I phrased my comment wrongly.
what I meant was that we don't seem to need to detach the group lock here and can safely rely on the group lock handler's mechanism.

nanangizz:
> Also make sure to test with multiple ICE restart scenario (such as in the case of reinvite without SDP) and check if there's no memory leak nor double destroy.

Tested reinvite without SDP multiple times from caller & callee (ICE restarts are confirmed from log), looks okay.

pjnath/src/pjnath/ice_strans.c
nanangizz: ah okay, this code is piggybacking `ice_prev` clean-up [here](https://github.com/pjsip/pjproject/pull/2903/files#diff-45e0dd75b3d3df135b497d029a81f78fae95e827c76a7e34c9b0ce9a51c4b0f3R990), it works, but it may not look clean.

sauwming (approved):

trengginas (approved):

PR #2904, opened 2021-12-01
Add documentation on how port_range is used

trengginas:
This ticket will add some documentation on how `port_range` is used by providing example.
This way user will have a better understanding of what to expect.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -3001,6 +3001,13 @@ typedef struct pjsua_transport_config
      * port number specified in \a port. Note that this setting is only
      * applicable when the start port number is non zero.
      *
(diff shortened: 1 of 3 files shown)


nanangizz (approved):

sauwming:
To avoid confusion as to why RTCP port can be out of [`port`, `port+port_range`], better move the media example to `pjsua_acc_cfg.rtp_cfg` instead.

This way, it's more understandable that since the name is `rtp_cfg`, it only applies to the RTP port that must always be within [`port`, `port+port_range`], while RTCP, since it's selected as `RTP+1`, can be out of range.

sauwming (approved):

wosrediinanatour:
Maybe it would be good to mention the algorithm performed in Function create_rtp_rtcp_sock(), e.g the RTP_RETRY=100. That gave me some headache in the past.

PR #2905, opened 2021-12-01
Make keyframe interval configurable in Apple VideoToolbox codecs

nanangizz:
Here is the new config:

```
/**
 * Maximum interval between keyframes for Apple VideoToolbox codecs,
 * in second.
 *
 * Default: 5 (seconds)
 */
#ifndef PJMEDIA_CODEC_VID_TOOLBOX_MAX_KEYFRAME_INTERVAL
#   define PJMEDIA_CODEC_VID_TOOLBOX_MAX_KEYFRAME_INTERVAL	5
#endif
```


sauwming:
I propose to include the keyword max (the property itself is called `kVTCompressionPropertyKey_MaxKeyFrameInterval`). This is because there's also minimum interval, which we may also need in the future (openh264 already has this).

sauwming (approved):
Don't forget to update the PR desc as well

trengginas (approved):

PR #2906, opened 2021-12-02
Fix issues in updating media dir to "none" in the middle of a call

nanangizz:
Re #2705 (setting media direction in call initialization and update), some issues have been reported:
1. In this code :
    ```
   pjsua_call_setting opt;
   pjsua_call_setting_default(&opt);
   opt.flag = PJSUA_CALL_SET_MEDIA_DIR | PJSUA_CALL_REINIT_MEDIA;
   opt.media_dir[0] = PJMEDIA_DIR_NONE;
   pjsua_call_update2(call_id, &opt, NULL);
   ```
   - without `PJSUA_CALL_REINIT_MEDIA` flag, media direction is unaffected.
   - with `PJSUA_CALL_REINIT_MEDIA` flag, assertion is raised:
     ```
     ../src/pjmedia/endpoint.c:931: pj_status_t pjmedia_endpt_create_base_sdp(
          pjmedia_endpt *, pj_pool_t *, const pj_str_t *, const pj_sockaddr *, pjmedia_sdp_session **):
          assertion "!"Invalid address family"" failed
     ```
2. Missing SWIG interface for vector/array `CallSetting::mediaDir`.

Thanks to Christoph Gritschenberger for the reports.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -2472,9 +2472,6 @@ pj_status_t pjsua_media_channel_init(pjsua_call_id call_id,
 	    call_med->def_dir = PJMEDIA_DIR_ENCODING_DECODING;
 	}
 	call_med->dir = call_med->def_dir;
-	if (call_med->dir == PJMEDIA_DIR_NONE) {
-	    enabled = PJ_FALSE;
-	}
 
 	if (enabled) {
 	    call_med->enable_rtcp_mux = acc->cfg.enable_rtcp_mux;
--- a/pjmedia/src/pjmedia/endpoint.c
+++ b/pjmedia/src/pjmedia/endpoint.c
@@ -42,6 +42,7 @@ static const pj_str_t STR_SDP_NAME = { "pjmedia", 7 };
 static const pj_str_t STR_SENDRECV = { "sendrecv", 8 };
 static const pj_str_t STR_SENDONLY = { "sendonly", 8 };
 static const pj_str_t STR_RECVONLY = { "recvonly", 8 };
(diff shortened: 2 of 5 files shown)


pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -2472,9 +2472,6 @@ pj_status_t pjsua_media_channel_init(pjsua_call_id call_id,
 	    call_med->def_dir = PJMEDIA_DIR_ENCODING_DECODING;
 	}
 	call_med->dir = call_med->def_dir;
-	if (call_med->dir == PJMEDIA_DIR_NONE) {
-	    enabled = PJ_FALSE;
-	}
```
sauwming, line 2477: I suppose this is the modification that fixes the first issue (i.e. the media dir is unaffected)?

If I remember correctly, the reason I included this in the original PR was for the initial offer/answer, to avoid media initialisation if the media direction is set to NONE. So please help test for this case as well to make sure that it works fine, such as:
```
    // video-only call
    call_opt.flag |= PJSUA_CALL_SET_MEDIA_DIR;
    call_opt.media_dir[0] = PJMEDIA_DIR_NONE;
    call_opt.media_dir[1] = PJMEDIA_DIR_DECODING;
```

nanangizz: There are two ways of disabling media: by setting port to 0 or by setting direction to inactive, the first will result in a totally disabled media, and the latter the media will still have RTCP traffics (and perhaps RTP keep alive) so media transports are still need to be initialized. As currently we only have one media direction (i.e: PJMEDIA_DIR_NONE), I guess it'd be better to use the latter way?

sauwming (approved):

trengginas (approved):

PR #2910, opened 2021-12-07
Remove epoll back-off timeout

tinarm:
Hi!

Does anyone remember the reason for adding this sleep to the worker thread(s)? My prio is not to get this merged if it has implications, I'm mostly interested why it came to place. :)

I have for some time been trying to understand why there are a fair amount of discards in the jitter buffers even though the network looks fine (i.e. a steady RTP cadence). What I discovered is that my worker threads (8 to 16 depending on hardware) are spending lots of time sleeping on the job. The result is that the incoming socket buffers are filling up because there are not enough workers to dispatch packets up the stack in a timely manner, and when the workers polls the events they get multiple packets which will overflow the jitter buffers (set to handle 40 ms delay).

If the sleep is removed according to this PR the discards drops by a factor of about 10 to 15, even when there are hundreds of simultaneous streams.

BR
Martin

### DIFF
--- a/pjlib/src/pj/ioqueue_epoll.c
+++ b/pjlib/src/pj/ioqueue_epoll.c
@@ -815,13 +815,6 @@ PJ_DEF(int) pj_ioqueue_poll( pj_ioqueue_t *ioqueue, const pj_time_val *timeout)
 	                            "ioqueue", 0);
     }
 
-    /* Special case:
-     * When epoll returns > 0 but no descriptors are actually set!
-     */
-    if (count > 0 && !event_cnt && msec > 0) {
-	pj_thread_sleep(msec);
-    }
-
     TRACE_((THIS_FILE, "     poll: count=%d events=%d processed=%d",
 		       count, event_cnt, processed_cnt));
 


tinarm:
Here's an example showing the number of active (i.e. not sleeping) worker threads (8 available on this server):

sip98_workers_full

sauwming:
I believe it's the same with the other implementation, such as select:
https://github.com/pjsip/pjproject/blob/master/pjlib/src/pj/ioqueue_select.c#L959
```	
        TRACE__((THIS_FILE, "     poll: no fd is set"));
        if (timeout)
            pj_thread_sleep(PJ_TIME_VAL_MSEC(*timeout));
        return 0;
```
i.e. if no descriptor is set, `ioqueue_poll()` will wait until timeout.

To overcome your issue, you can set a smaller timeout value perhaps?

tinarm:
It's true that select has a similar construct, but I don't think the semantics are exactly the same.

The timeout in epoll came about in the following commit which only mention it as a temporary fix: 
https://github.com/pjsip/pjproject/commit/09413ca1df78ba19a40db23d1af899da4141885f

I'm interested to hear if anyone remember the cause for the temporary fix.

In my setup I added a separate setting for the timeout and made it configurable via a new ioqueue function which works great for me.
I could suggest that patch as a PR since that wouldn't change any defaults, but give clients the possibility to change the timeout, or to remove it completely.

sauwming:
I think the answer is the same as the one provided by Andreas Wehrmann in the mailing list.
```
while(true) {
    pj_ioqueue_poll(timeout);
}
```
Without the sleep, the above polling may be busy looping when count > 0 but no fd is set.

tinarm:
Sorry, I've missed that response. I think I was expecting to get a notification back from the list if it got an answer. ;)
Thanks!

PR #2911, opened 2021-12-08
Fix video issues upon format change

sauwming:
Typically, the (decoding) format change flow is as follows:
1. Vid codec publishes fmt_changed event (sync)
2. vid_stream decoder receives fmt_changed event, republish it async.
3. pjsua_media fmt_changed event (from vid_stream), update vid_conf vid_stream decoding port.
4. vid_port renderer fmt_changed event (from vid_stream), restart renderer, republish async.
5. pjsua_media fmt_changed event (from vid_port), update vid_conf renderer port.

The log shows one example of this flow:
```
10:36:25.981           openh264.cpp !Frame size changed to 720x480
10:36:25.994      vstdec0x120197028 ! Decoding format changed: 720x480 I420<- 15/1(~15)fps
10:36:25.994             vid_conf.c  Port 1 (vstdec0x120197028): updated frame rate 22 -> 15
10:36:25.994             vid_conf.c  Port 1 (vstdec0x120197028): updated frame size 1456x1456 -> 720x480
10:36:25.994       and_opengl_dev.c  Stopping android opengl stream
10:36:25.994       and_opengl_dev.c  change format 720x480
10:36:25.994       and_opengl_dev.c  Starting android opengl stream
10:36:25.995             vid_conf.c !got frame port 0: 720x480
10:36:26.065             vid_conf.c !got frame port 0: 720x480
10:36:26.140             vid_conf.c !got frame port 0: 720x480
10:36:26.195             vid_conf.c  Port 0 (OpenGL renderer): updated frame size 1456x1456 -> 720x480
```

As described in the flow and shown in the log, due to the mostly asynchronous nature of the format change event propagation, there can be a delay between each step. For step 4 and 5,  the delay means the renderer port to have been updated, while `vid_conf` hasn't received the event. This format difference in vid_conf causes video artefacts to appear in the beginning of format change event.

Important note:
The patch in this PR is to check for this format difference during rendering in the video conference. Still, due to the lack of synchronisation, a race may still occur (i.e. the renderer port is being changed at the same time with the checking), but the patch should minimise the possibility of such event happening.

Also included in this PR:
buffer initialization with black color (so blank frame doesn't appear as a green screen to the user)

### DIFF
--- a/pjmedia/src/pjmedia-videodev/util.c
+++ b/pjmedia/src/pjmedia-videodev/util.c
@@ -164,8 +164,8 @@ pjmedia_vid_dev_conv_create_converter(pjmedia_vid_dev_conv *conv,
     
     conv->wxh = conv->dst_size.w * conv->dst_size.h;
     conv->src_frame_size = dst_size.w * dst_size.h * vfi->bpp / 8;
-    conv->conv_frame_size = conv->rot_size.w * conv->rot_size.h;
-    conv->conv_frame_size *= vfi->bpp / 8;
+    conv->conv_frame_size = conv->rot_size.w * conv->rot_size.h *
+			    vfi->bpp / 8;
     conv->conv_buf = pj_pool_alloc(pool, conv->src_frame_size);
     
     pjmedia_vid_dev_conv_set_rotation(conv, PJMEDIA_ORIENT_NATURAL);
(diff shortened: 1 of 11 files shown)


pjmedia/src/pjmedia-videodev/util.c
```diff
@@ -164,8 +164,8 @@ pjmedia_vid_dev_conv_create_converter(pjmedia_vid_dev_conv *conv,
     
     conv->wxh = conv->dst_size.w * conv->dst_size.h;
     conv->src_frame_size = dst_size.w * dst_size.h * vfi->bpp / 8;
-    conv->conv_frame_size = conv->rot_size.w * conv->rot_size.h *
-			    vfi->bpp / 8;
+    conv->conv_frame_size = conv->rot_size.w * conv->rot_size.h;
+    conv->conv_frame_size *= vfi->bpp / 8;
```
nanangizz, line 168: this change should not be reverted, note that for IYUV/I420, bpp==12, so conv_frame_size will be multiplied to 1 instead of 1.5.

nanangizz (approved):

PR #2915, opened 2021-12-09
Do not preserve ownership when installing libs

martin-flaska:
This is fixing 'host-user-contaminated' Yocto QA warning:

https://www.yoctoproject.org/docs/current/ref-manual/ref-manual.html#ref-tasks-install

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,13 +122,13 @@ cmp_wav:
 install:
 	mkdir -p $(DESTDIR)$(libdir)/
 	if [ "$(PJ_EXCLUDE_PJSUA2)x" = "x" ] ; then \
-	    cp -af $(APP_LIBXX_FILES) $(DESTDIR)$(libdir)/; \
+	    cp -af --no-preserve=ownership $(APP_LIBXX_FILES) $(DESTDIR)$(libdir)/; \
 	else \
-	    cp -af $(APP_LIB_FILES) $(DESTDIR)$(libdir)/; \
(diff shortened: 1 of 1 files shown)


sauwming:
Unfortunately I think the option is not available on all platforms.
When I tried it here on Mac, it gave the error `illegal option`.

martin-flaska:
@sauwming I replaced `cp` with recommended [install](https://www.gnu.org/software/coreutils/manual/html_node/install-invocation.html#install-invocation) command for installation process. is this command also available on Mac?

sauwming:
Yes, it's available on Mac. From a quick check, it also seems to be available on most platforms that I check. The problem is I'm not 100% sure to be the first to approve this. Will need to wait for the other's vote of confidence or if there's any objection.

martin-flaska:
Thank you @sauwming for your answer. No problem if it is posponed. I noticed more QA warnings related to the Yocto build. Some of them are already fixed since 2.11.1 version was released. I have just some quick workaround solutions for now, if I find some good solution later, I will share.

nanangizz:
Looks okay to me (not Makefile expert).

Btw, will it be better to make it a bit more configurable, e.g: add `AC_CHECK_TOOL(INSTALL, ..)` in the `aconfigure.ac` and `cc-*.mak.in`, so user can add custom options or even use an alternative install tool?

martin-flaska:
I found some issues with installing sym links which I need to solve. Closing...

PR #2920, opened 2021-12-14
sip_inv:  Additional multipart support (#2919)

### LINKED ISSUE #2919: sip_inv:inv_check_sdp_in_incoming_msg doesn't handle multipart bodies in tx_data
### Describe the bug

sip_inv.c:inv_check_sdp_in_incoming_msg() deals with multipart message bodies in rdata correctly.  In the case where early
media is involved though, the existing sdp has to be retrieved from the last tdata sent in this transaction.  This, however, always assumed that the sdp sent is in a non-multipart body.  While there's a function to retrieve the sdp from multipart and
non-multpart rdata bodies, no similar function for tdata exists.

PR forthcoming.

### Steps to reproduce

Run an early-media scenario where the original request contained a multipart body.

### PJSIP version

2.11.1

### Context

See above.

### Log, call stack, etc

```shell
Offending code from sip_inv:inv_check_sdp_in_incoming_msg:

	    /* Retrieve original SDP offer from INVITE request */
	    reoffer_sdp = (const pjmedia_sdp_session*) 
			  tsx->last_tx->msg->body->data;

	    /* Feed the original offer to negotiator */
	    status = pjmedia_sdp_neg_modify_local_offer2(inv->pool_prov, 
							 inv->neg,
                                                         inv->sdp_neg_flags,
						         reoffer_sdp);
```

gtjoseph:
FYI...  Although not strictly needed,  I created both pjsip_tdata_get_sdp_info and pjsip_tdata_get_sdp_info2 to mirror the existing  pjsip_rdata_get_sdp_info and pjsip_rdata_get_sdp_info2 functions.  I can eliminate  pjsip_tdata_get_sdp_info2 if that's better.  Also double check code formatting.   I lost my "pjproject" Eclipse code formatter and had to recreate it from scratch. :)

----------------
sip_inv.c:inv_check_sdp_in_incoming_msg() deals with multipart
message bodies in rdata correctly. In the case where early media is
involved though, the existing sdp has to be retrieved from the last
tdata sent in this transaction. This, however, always assumes that
the sdp sent is in a non-multipart body. While there's a function
to retrieve the sdp from multipart and non-multpart rdata bodies,
no similar function for tdata exists.  So...

* The existing pjsip_rdata_get_sdp_info2 was refactored to
  find the sdp in any body, multipart or non-multipart, and
  from either an rdata or tdata.  The new function is
  pjsip_get_sdp_info.  This new function detects whether the
  pjsip_msg->body->data is the text representation of the sdp
  from an rdata or an existing pjmedia_sdp_session object
  from a tdata, or whether pjsip_msg->body is a multipart
  body containing either of the two sdp formats.

* The exsting pjsip_rdata_get_sdp_info and pjsip_rdata_get_sdp_info2
  functions are now wrappers that get the body and Content-Type
  header from the rdata and call pjsip_get_sdp_info.

* Two new wrappers named pjsip_tdata_get_sdp_info and
  pjsip_tdata_get_sdp_info2 have been created that get the body
  from the tdata and call pjsip_get_sdp_info.

* For testing purposes, a new field "create_multipart" was added
  to the pjsip_inv_session structure to allow forcing all bodies
  created to be multipart.

* The internal create_sdp_body function now takes an option to
  indicate that a multipart body should be created.

* inv_offer_answer_test.c was updated to use the create_multipart
  field and additional tests were added.

Fixes #2919

### DIFF
--- a/pjsip/include/pjsip-ua/sip_inv.h
+++ b/pjsip/include/pjsip-ua/sip_inv.h
@@ -463,11 +463,11 @@ struct pjsip_inv_session
 
 
 /**
- * This structure represents SDP information in a pjsip_rx_data. Application
- * retrieve this information by calling #pjsip_rdata_get_sdp_info(). This
+ * This structure represents SDP information in a pjsip_(rx|tx)_data. Application
+ * retrieve this information by calling #pjsip_get_sdp_info(). This
  * mechanism supports multipart message body.
  */
-typedef struct pjsip_rdata_sdp_info
+typedef struct pjsip_sdp_info
 {
     /**
      * Pointer and length of the text body in the incoming message. If
@@ -487,7 +487,15 @@ typedef struct pjsip_rdata_sdp_info
      */
     pjmedia_sdp_session *sdp;
 
-} pjsip_rdata_sdp_info;
+} pjsip_sdp_info;
+
+/**
+ * For backwards compatibility and completeness,
+ * pjsip_rdata_sdp_info and pjsip_tdata_sdp_info
+ * are typedef'd to pjsip_sdp_info.
+ */
+typedef pjsip_sdp_info pjsip_rdata_sdp_info;
+typedef pjsip_sdp_info pjsip_tdata_sdp_info;
 
 
 /**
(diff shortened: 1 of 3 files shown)


pjsip/include/pjsip-ua/sip_inv.h
```diff
@@ -459,15 +459,19 @@ struct pjsip_inv_session
     pj_atomic_t		*ref_cnt;		    /**< Reference counter. */
     pj_bool_t            updated_sdp_answer;        /**< SDP answer just been
 							 updated?	    */
+    pj_bool_t            create_multipart;          /**< Create multipart
+                                                         bodies.  Should be
+                                                         used for testing
+                                                         only.              */
```
nanangizz, line 465: Not sure about this setting because:
- Potential behavioral change affecting app or upper level API.  For example, currently pjsua seems to always assume that Invite module will always create non-multipart body, see how this [pjsua_process_msg_data()](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_core.c#L2994) implementation creates multipart body, it does not check whether the existing body, which is generated by Invite module, is a multipart (as specified [here](https://datatracker.ietf.org/doc/html/rfc5621#section-4.3), nested multipart is not recommended). Some apps may assume the same.
- Even if the ```pjsip_inv_session``` structure is public (as opposed to transparent), as far as I know, modifying/writing the field directly should be avoided, it should be done via a function (or an option/flag), which does not seem to be provided here.

Also, as currently this setting seems to be used by a test app only, so perhaps better to handle multipart body generation in the test app if possible?

gtjoseph: Hmmm.   That flag really was there for internal testing only as the comment states.  I wanted the existing unit tests in inv_offer_answer_test to be able to exercise the  multipart logic in sip_inv.  Having the test do the multipart handling would defeat the purpose.  I can just remove the extra tests and the field if you prefer.

nanangizz: I mean the app should generate the multipart body (instead of handling the whole multipart things) and leave sip_inv to generate non-multipart body as is (so the new field `create_multipart` can be avoided), something like this: [2920_diff.txt](https://github.com/pjsip/pjproject/files/7724906/2920_diff.txt).
.

gtjoseph: Ah, gotcha.  Let me try that and see if it still exercises the code paths.

gtjoseph: OK, I removed the create_multipart option and applied the diff to inv_offer_answer_test.  Seems to work fine.  Anything else I need to do?

nanangizz (approved):
Looks okay to me. Thanks!

pjsip/include/pjsip-ua/sip_inv.h
```diff
@@ -487,7 +491,15 @@ typedef struct pjsip_rdata_sdp_info
      */
     pjmedia_sdp_session *sdp;
 
-} pjsip_rdata_sdp_info;
+} pjsip_sdp_info;
+
+/**
+ * For backwards compatibility and completemess,
```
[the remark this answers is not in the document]
gtjoseph: > completemess

That was a Freudian slip :)

pjsip/include/pjsip-ua/sip_inv.h
```diff
@@ -1057,6 +1069,44 @@ PJ_DECL(pj_status_t) pjsip_create_sdp_body(pj_pool_t *pool,
...
+ * SIP body as the first part.
+ *
+ * @param pool		Pool to allocate memory.
+ * @param sdp		SDP session to be put in the SIP message body.
+ * @param p_body	Pointer to receive SIP message body containing
+ *			the SDP session.
+ *
+ * @return		PJ_SUCCESS on success.
+ */
+PJ_DEF(pj_status_t) pjsip_create_multipart_sdp_body( pj_pool_t *pool,
```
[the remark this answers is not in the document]
gtjoseph: oops.  Cut and paste error.

sauwming (approved):

PR #2921, opened 2021-12-15
Support Flow-Timer on rfc5626

trengginas:
This ticket will add support to `Flow-Timer` header field specified in [rfc5626](https://datatracker.ietf.org/doc/html/rfc5626#section-11.1).

```
When a successful registration response contains the Flow-Timer
   header field, the value of this header field is the number of seconds
   the server is prepared to wait without seeing keep-alives before it
   could consider the corresponding flow dead.  Note that the server
   would wait for an amount of time larger than the Flow-Timer in order
   to have a grace period to account for transport delay.  The UA MUST
   send keep-alives at least as often as this number of seconds.  If the
   UA uses the server-recommended keep-alive frequency it SHOULD send
   its keep-alives so that the interval between each keep-alive is
   randomly distributed between 80% and 100% of the server-provided
   time.  For example, if the server suggests 120 seconds, the UA would
   send each keep-alive with a different frequency between 95 and 120
   seconds.
```

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -2130,8 +2130,12 @@ static void keep_alive_timer_cb(pj_timer_heap_t *th, pj_timer_entry *te)
 	goto on_return;
 
     /* Reschedule next timer */
-    delay.sec = acc->rfc5626_flowtmr?acc->rfc5626_flowtmr:
-				     acc->cfg.ka_interval;
+    if (acc->rfc5626_flowtmr) {
+        unsigned tmr_exp = (pj_rand() % 20) + 80;
+        delay.sec = acc->rfc5626_flowtmr * tmr_exp / 100;
+    } else {
+        delay.sec = acc->cfg.ka_interval;
+    }
     delay.msec = 0;
     status = pjsip_endpt_schedule_timer(pjsua_var.endpt, te, &delay);
     if (status == PJ_SUCCESS) {
@@ -2171,9 +2175,8 @@ static void update_keep_alive(pjsua_acc *acc, pj_bool_t start,
 	      pjsip_msg_find_hdr_by_name(param->rdata->msg_info.msg,
 					 &STR_FLOW_TIMER, hsr);
 
-	if (hsr) {
-	    unsigned tmr_exp = (pj_rand() % 20) + 80;
-	    acc->rfc5626_flowtmr = pj_strtoul(&hsr->hvalue) * tmr_exp / 100;
+	if (hsr) {	    
+	    acc->rfc5626_flowtmr = pj_strtoul(&hsr->hvalue);
 	}
 
 	/* Only do keep-alive if:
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -2162,8 +2163,21 @@ static void update_keep_alive(pjsua_acc *acc, pj_bool_t start,
...
+	static const pj_str_t STR_FLOW_TIMER  = { "Flow-Timer", 10 };
+
+	hsr = (pjsip_generic_string_hdr*)
+	      pjsip_msg_find_hdr_by_name(param->rdata->msg_info.msg,
+					 &STR_FLOW_TIMER, hsr);
+
+	if (hsr) {
+	    unsigned tmr_exp = (pj_rand() % 20) + 80;
+	    acc->rfc5626_flowtmr = pj_strtoul(&hsr->hvalue) * tmr_exp / 100;
+	}
```
sauwming, line 2177: From the RFC:
```
If the
   UA uses the server-recommended keep-alive frequency it SHOULD send
   its keep-alives so that the interval between each keep-alive is
   randomly distributed between 80% and 100% of the server-provided
   time.  For example, if the server suggests 120 seconds, the UA would
   send each keep-alive with a different frequency between 95 and 120
   seconds.
```

The randomisation applies when sending each keep alive, so it shouldn't be here, but when scheduling the timer below.

This is probably not very major, but also note that the random number selection should be between a lower bound and an upper bound, instead of percentage based (percentage based calculation may cause certain numbers to never be picked).
```
The UA MUST select a random number between a fixed or
   configurable upper bound and a lower bound, where the lower bound is
   20% less then the upper bound.
```

trengginas: The random number recommendation is when `Flow-timer` is not provided on the response.
```
When a Flow-Timer header field is not provided in the most recent
   success registration response, the proper selection of keep-alive
   frequency is primarily a trade-off between battery usage and
   availability.  The UA MUST select a random number between a fixed or
   configurable upper bound and a lower bound, where the lower bound is
   20% less then the upper bound.
```
In this case we already have ```ka_interval``` settings, and it will change the current timer calculation.

sauwming: True, but I think the spirit is the same.
```If the
   UA uses the server-recommended keep-alive frequency it SHOULD send
   its keep-alives so that the interval between each keep-alive is
   randomly distributed between 80% and 100% of the server-provided
   time.  For example, if the server suggests 120 seconds, the UA would
   send each keep-alive with a different frequency between 95 and 120
   seconds.
```
Now let's say the server suggests 1000 seconds (the RFC recommends 840 sec when battery power is a concern in another section, and here I picked a nice round number for easier calculation). Now the interval should be randomly distributed between 800/799 to 1000 seconds. With percentage based calculation, it would be 810, 820, 830, ..., and so on, and will never pick 801, 802, ..., 811, 812, etc. In other words, there will be numbers that have no distribution at all.

Again, this is a minor issue and probably won't matter much in practice, but since the calculation is as simple, I believe we should adopt this instead?

Note: the current percentage-based approach is actually not wrong, but it needs to be floating point in order to work, i.e. it needs to choose a random floating point number between 80% to 100%.

sauwming: I was saying that the percentage approach is actually not wrong, but still, I won't recommend it since it makes the calculation much more complicated.

Instead of choosing an (integer) random number between 80 to 100, why not choose an integer between lower and upper bound as the RFC suggested? The calculation is the same and as simple, i.e.:
```
unsigned lower_bound = (unsigned)timer * 0.8f;
delay = pj_rand() % (timer - lower_bound) + lower_bound;
```

sauwming (approved):

nanangizz (approved):

PR #2924, opened 2021-12-16
Prevent OOB read for RTCP XR block

trengginas:
When receiving RTCP XR block, the packet will be parsed by using cast
``` c
switch (rb_hdr->bt) {
case BT_RR_TIME:
    rb_rr_time = (pjmedia_rtcp_xr_rb_rr_time*) rb_hdr;
    break;
case BT_DLRR:
    rb_dlrr = (pjmedia_rtcp_xr_rb_dlrr*) rb_hdr;
    break;
case BT_STATS:
    rb_stats = (pjmedia_rtcp_xr_rb_stats*) rb_hdr;
    break;
case BT_VOIP_METRICS:
    rb_voip_mtc = (pjmedia_rtcp_xr_rb_voip_mtc*) rb_hdr;
    break;
default:
    break;
}
```
Without checking the length, this can lead to out of bound read.
This ticket will prevent such issue by checking the packet length.

### DIFF
--- a/pjmedia/src/pjmedia/rtcp_xr.c
+++ b/pjmedia/src/pjmedia/rtcp_xr.c
@@ -417,7 +417,7 @@ void pjmedia_rtcp_xr_rx_rtcp_xr( pjmedia_rtcp_xr_session *sess,
     const pjmedia_rtcp_xr_rb_voip_mtc *rb_voip_mtc = NULL;
     const pjmedia_rtcp_xr_rb_header   *rb_hdr = (pjmedia_rtcp_xr_rb_header*) 
 						rtcp_xr->buf;
-    unsigned pkt_len, rb_len;
+    unsigned pkt_len, rb_len, ck_len;
 
     if (rtcp_xr->common.pt != RTCP_XR)
 	return;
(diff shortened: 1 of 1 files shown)


sauwming (commented):
This looks correct, but not easy to understand (I had to think really hard to follow the algorithm).

Why not compare it against the end packet buffer instead?
i.e.
```
pj_int32_t *pkt_end = (pj_int32_t*)pkt + pkt_len + 1;

if ((pj_int32_t *)rb_hdr + sizeof(something)/4 > pkt_end) break;
```
Or
```
if ((char *)rb_hdr + sizeof(something) > (char *)pkt + size) break;
```

pjmedia/src/pjmedia/rtcp_xr.c
```diff
@@ -427,32 +427,49 @@ void pjmedia_rtcp_xr_rx_rtcp_xr( pjmedia_rtcp_xr_session *sess,
...
     {	
+	unsigned bl_len;
 	rb_len = pj_ntohs((pj_uint16_t)rb_hdr->length);
 
 	/* Just skip any block with length == 0 (no report content) */
 	if (rb_len) {
 	    switch (rb_hdr->bt) {
 		case BT_RR_TIME:
-		    rb_rr_time = (pjmedia_rtcp_xr_rb_rr_time*) rb_hdr;
+		    bl_len = sizeof(*rb_rr_time) / 4;
```
the-storm, line 442: I think the fix looks good. My concern here what happens if in the future a new type is added that's not divisible by 4? 
Will a similar check be sufficient to prevent an OOB? 
imagine a custom type with struct `pjmedia_rtcp_xr_rb_some_custom_type` that has sizeof() => 3 

a check that's copy and paste from this check wont work. This concern also applies if any of these struct changes in size and doesn't become divisible by 4 anymore.

I think getting the end of the buffer, and check that the `buffer + sizeof(*rb_rr_time)` is within the bounds of the buffer should be more general and work hopefully right regardless of the struct size

Does this sound reasonable?

pjmedia/src/pjmedia/rtcp_xr.c
```diff
@@ -427,32 +427,49 @@ void pjmedia_rtcp_xr_rx_rtcp_xr( pjmedia_rtcp_xr_session *sess,
     if ((pkt_len + 1) > (size / 4))
 	return;
 
+    ck_len = pkt_len - 1;
+
     /* Parse report rpt_types */
     while ((pj_int32_t*)rb_hdr < (pj_int32_t*)pkt + pkt_len)
     {	
+	unsigned bl_len;
```
the-storm, line 435: nit: maybe you can declare this outside the loop?

pjmedia/src/pjmedia/rtcp_xr.c
```diff
@@ -427,32 +427,49 @@ void pjmedia_rtcp_xr_rx_rtcp_xr( pjmedia_rtcp_xr_session *sess,
     if ((pkt_len + 1) > (size / 4))
 	return;
 
+    ck_len = pkt_len - 1;
```
the-storm, line 430: why this is `pkt_len -1` and not `pkt_len` ?

the-storm (commented):
I am curious if this should have been a private PR since it is a security issue and wether it should have a GH security advisory attached to it with a CVE.

the-storm (commented):
This looks good to me from security :) 
I will defer approval from someone from the team who is more aware of the codebase :) 

Thanks for fixing this @trengginas

the-storm:
> I am curious if this should have been a private PR since it is a security issue and wether it should have a GH security advisory attached to it with a CVE.

@trengginas any comments with regard to this?

PR #2926, opened 2021-12-21
NAT64: handle special case where server put FQDN (resolved to IPv6 only) in SDP answer

nanangizz:
Handle special case in NAT64 scenario where for some reason, the server puts IPv6 (literal or FQDN) in SDP answer while indicating "IP4" in its address type in SDP c= line. Perhaps the server knows that local endpoint is IPv6 from SIP signaling, so it replaces/puts IPv6 address in its SDP. While "IP4" in SDP c= line is just to match the SDP offer.

### DIFF
--- a/pjmedia/src/pjmedia/stream_info.c
+++ b/pjmedia/src/pjmedia/stream_info.c
@@ -445,6 +445,14 @@ PJ_DEF(pj_status_t) pjmedia_stream_info_from_sdp(
     /* Set remote address: */
     status = pj_sockaddr_init(rem_af, &si->rem_addr, &rem_conn->addr,
 			      rem_m->desc.port);
+    if (status == PJ_ERESOLVE && rem_af == pj_AF_INET()) {
+	/* Handle special case in NAT64 scenario where for some reason, server
+	 * puts IPv6 (literal or FQDN) in SDP answer while indicating "IP4"
+	 * in its address type, let's retry resolving using AF_INET6.
+	 */
+	status = pj_sockaddr_init(pj_AF_INET6(), &si->rem_addr,
+				  &rem_conn->addr, rem_m->desc.port);
+    }
     if (status != PJ_SUCCESS) {
 	/* Invalid IP address. */
 	return PJMEDIA_EINVALIDIP;


sauwming:
From the RFC 4566:
```
c=<nettype> <addrtype> <connection-address>

When the <addrtype> is IP4 and IP6, the connection address is defined
   as follows:

   o  If the session is multicast, the connection address will be an IP
      multicast group address.  If the session is not multicast, then
      the connection address contains the unicast IP address of the
      expected data source or data relay or data sink as determined by
      additional attribute fields.  It is not expected that unicast
      addresses will be given in a session description that is
      communicated by a multicast announcement, though this is not
      prohibited.
```

I'm not even sure if FQDN is allowed in the SDP. If we want to accommodate this due to a certain server behavior, perhaps better add a compile-time setting (with `disabled` as the default since it seems to be against the RFC).

nanangizz:
Actually I have checked it, in the [SDP grammar](https://datatracker.ietf.org/doc/html/rfc4566#section-9) section, FQDN seems to be allowed:
```
   connection-field =    [%x63 "=" nettype SP addrtype SP
                         connection-address CRLF]

   ; sub-rules of 'c='
   connection-address =  multicast-address / unicast-address

   ; generic sub-rules: addressing
   unicast-address =     IP4-address / IP6-address / FQDN / extn-addr
```

sauwming:
Ah right, I missed that.

Still, the connection address seems to be invalid if checked with the addrtype. Since in normal (non-NAT64) scenario, this should be correctly handled as error, so a setting should still be needed?

nanangizz:
True, it is already correctly handled as error, but it is perhaps not so bad when treated as non-error (if the resolution succeeds). For normal-IPv4/non-NAT64 scenario, if it fails in IPv6 resolution attempt, it just behaves the same as usual, if it does not fail, the media may still get benefit from the "ratching" feature in UDP transport (auto switch target).

On the other hand, adding setting is useful to avoid any unexpected behavioral change, e.g: some app may want the call to fail faster. IMHO it should be fine either way, configurable or not.

sauwming (approved):

trengginas (approved):

PR #2929, opened 2021-12-24
Avoid ICE agents seeing different valid ICE candidate pairs

nanangizz:
Sometimes, when ICE nego has completed, ICE agents may see different valid pairs.

After investigation, this may be caused by:
1. In nominating stage, currently any unchecked candidates (whose state is `PJ_ICE_SESS_CHECK_STATE_FROZEN/WAITING`) may be scheduled for check (with nomination flag!) too, I think it should check/nominate the valid check only.
2. After nego has completed, any late incoming check may cause local agent to update its valid check.
3. After nego has completed, any late check response (late STUN binding completion) may cause local agent to update its valid check.

Note that the ICE session will use the candidate from the valid check for sending data, not the nominated check, assuming the nominated check must be the valid check, also it allows app to send data earlier before nomination is completed.

Currently the available patch is to handle cause 2 & 3 only. Will add patch later for cause 1 (after verification and discussion).


sauwming:
Looks okay. Will the additional patch be in this PR or a different one?

nanangizz:
> Looks okay. Will the additional patch be in this PR or a different one?

In this PR too. Haven't started creating the patch yet actually, need feedback if the analysis of cause 1 is correct & the proposed solution (i.e: should check/nominate the valid check only) is acceptable, and if there is any potential problem with the behavior change.

sauwming:
After re-reading the RFC again, I'm not sure that the patch is correct.
https://datatracker.ietf.org/doc/html/rfc5245#section-8.1.1.2

The late checks (case 2 and 3) seem expected in aggressive nomination:
```
   With aggressive nomination, the controlling agent includes the USE-
   CANDIDATE attribute in every check it sends.  Once the first check
   for a component succeeds, it will be added to the valid list and have
   its nominated flag set.  When all components have a nominated pair in
   the valid list, media can begin to flow using the highest priority
   nominated pair.  However, because the agent included the USE-
   CANDIDATE attribute in all of its checks, another check may yet
   complete, causing another valid pair to have its nominated flag set.
   ICE always selects the highest-priority nominated candidate pair from
   the valid list as the one used for media.  Consequently, the selected
   pair may actually change briefly as ICE checks complete, resulting in
   a set of transient selections until it stabilizes.
```

Then in the section below:
```
Once there is at least one nominated pair in the valid list for
      every component of at least one media stream and the state of the
      check list is Running:

      *  The agent MUST change the state of processing for its check
         list for that media stream to Completed.

      *  The agent MUST continue to respond to any checks it may still
         receive for that media stream, and MUST perform triggered
         checks if required by the processing of Section 7.2.

      *  The agent MUST continue retransmitting any In-Progress checks
         for that check list.
```

And:
```
 Once the state of each check list is Completed:

      *  The agent sets the state of ICE processing overall to
         Completed.

      *  If an agent is controlling, it examines the highest-priority
         nominated candidate pair for each component of each media
         stream.  If any of those candidate pairs differ from the
         default candidate pairs in the most recent offer/answer
         exchange, the controlling agent MUST generate an updated offer
         as described in Section 9.  If the controlling agent is using
         an aggressive nomination algorithm, this may result in several
         updated offers as the pairs selected for media change.  An
         agent MAY delay sending the offer for a brief interval (one
         second is RECOMMENDED) in order to allow the selected pairs to
         stabilize.
```

So it seems that we actually need to accommodate those late checks and send (another) updated offer instead.

And perhaps this is for case 1?
```
o  If there is at least one nominated pair in the valid list for a
      media stream and the state of the check list is Running:

      *  The agent MUST remove all Waiting and Frozen pairs in the check
         list and triggered check queue for the same component as the
         nominated pairs for that media stream.
```

sauwming:
Closing this and moving it to issue #3272 since the issue is rather complicated.

sauwming:
After I reread the RFC and our implementation, more carefully, I think the patch is correct.

https://github.com/pjsip/pjproject/blob/master/pjnath/src/pjnath/ice_session.c#L1460
Now I realize that our ICE is for one media stream only, so the late checks I quoted above do not apply. As soon as we have a nominated pair in every component, the state of check list for that media stream is Completed, and since there is only one media, the state of ICE processing overall is instantly set to Completed.

After ICE overall is Completed, any late checks should be discarded/ignored as proposed in the patch.

I don't quite understand about concern no. 1 though, if I check:
https://github.com/pjsip/pjproject/blob/master/pjnath/src/pjnath/ice_session.c#L1702
it seems correct.

```
If there is at least one nominated pair in the valid list:
     *
     *    - The agent MUST remove all Waiting and Frozen pairs in the check
     *      list for the same component as the nominated pairs for that
     *      media stream
```
So in aggressive nomination, if there's a valid pair, which means that it must also be a nominated pair, there shouldn't be any more Waiting and Frozen pairs.

So is there a bug somewhere with our handling for this?

nanangizz:
In point no 1, it happens in **nominating** stage, so there is no nominated pair yet. IIRC the problematic codes are about here:
https://github.com/pjsip/pjproject/blob/e210c69104a871cdcaa9dc9747a684a4fa7e8b20/pjnath/src/pjnath/ice_session.c#L2368-L2369
and here:
https://github.com/pjsip/pjproject/blob/e210c69104a871cdcaa9dc9747a684a4fa7e8b20/pjnath/src/pjnath/ice_session.c#L2388-L2389

IIRC the problem is that WAITING/FROZEN pair will get checked **with** nomination flag set, while IMO only valid pairs can get the nomination flag set, and this is about regular nomination, not aggressive.

sauwming:
Correct me if I'm wrong, I believe for regular nomination, the flow is as follows:
- Either all checks have completed or until TIMER_START_NOMINATED_CHECK is triggered.
- `start_nominated_check()` is called
- https://github.com/pjsip/pjproject/blob/e210c69104a871cdcaa9dc9747a684a4fa7e8b20/pjnath/src/pjnath/ice_session.c#L2444 "For each component, set the check state of valid check with highest priority to Waiting (it should have Success state now)". This seems to be done to follow the RFC: `When the controlling agent selects the valid pair, it repeats the
   check that produced this valid pair (by enqueuing the pair that
   generated the check into the triggered check queue), this time with
   the USE-CANDIDATE attribute. `

So the WAITING pair is actually a successful valid pair that has been downgraded in order to perform a repeat check.

nanangizz:
Yes, that nominating pair check is shifted to WAITING so it will be checked soon with nominated flag, no problem with this pair. My concern is if there is any other pair that is also in WAITING (due to scheduled, or triggered check, etc) so that pair will be checked with nominated flag too. Note that all WAITING/FROZEN checks are only shifted to FAILED after that nominating check completes successfully, so when nomination is started, there can be more than one pair in WAITING/FROZEN state (for each component) that will be nominated, which eventually may cause the original issue. Not sure if this scenario is possible though.

sauwming:
I think you are right.

I believe for regular nomination, we should stop all checks (except those being nominated) when we proceed to nominating stage.
So the fixes should be:
- stop the periodic check timer
- add the condition in the patch to `if (ice->is_complete || (ice->is_nominating && !ice->opt.aggressive))`

What do you think?

nanangizz:
Not sure if I understand the fixes proposal correctly. Currently the nomination check is initiated by the periodic timer (shared by regular check & nomination check), we still need the timer to initiate nomination checks for all components. And in the current patch, if the condition is added to `on_request_complete()`, then the possibility of sending multiple nominations (for each component) will still happen?

My idea is rather straightforward (haven't checked the standard, mostly IIRC :): for regular nomination, only perform nomination check for the designated pair for each component (should be the highest prio in the valid checklist, perhaps is equal to `comp->valid_check` which should be updated by `update_comp_check()`). So in `start_periodic_check()` when nomination is started and it is regular nomination (not aggressive), only call `perform_check()` for described above pairs (i.e: valid pair of each comp).

sauwming:
> My idea is rather straightforward (haven't checked the standard, mostly IIRC :): for regular nomination, only perform nomination check for the designated pair for each component (should be the highest prio in the valid checklist, perhaps is equal to `comp->valid_check` which should be updated by `update_comp_check()`). So in `start_periodic_check()` when nomination is started and it is regular nomination (not aggressive), only call `perform_check()` for described above pairs (i.e: valid pair of each comp).

Ah okay, yes this should work.

I was just wondering why only in `start_nominated_check()` we suddenly use a timer to perform triggered checks? In other places, we always directly call `perform_check()`. So that's why I suggested to do the same here and stop the timer altogether since we don't need it anymore.

nanangizz:
AFAIK the timer (or `start_periodic_check()`) is the main way for connectivity check where regular interval between checks can be applied, it is currently also used for nominating in regular nomination (and of course aggressive nomination where all checks are nomination checks). While a direct call to `perform_check()` seems to happen only for triggered check (check initiated as a 'response' after received & responded an incoming check from remote) and role conflict.

nanangizz:
Will try to create the patch for point no 1.

sauwming (approved):
Only minor: indentation. I think the old commit still use tab style.

PR #2931, opened 2021-12-24
pjsua: Return earlier when handling of rx offer is async

mscdex:
This avoids having to make a useless call to `pjsua_media_channel_create_sdp()` and the like if we know we're responding to the offer asynchronously (especially with `pjsua_call_answer_with_sdp()`).

I'm not 100% sure if this has the potential to cause any issues (e.g. if answering the call at a later point *without* an SDP), but all of the pjsua tests pass before and after this change for me so :man_shrugging:

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_call.c
+++ b/pjsip/src/pjsua-lib/pjsua_call.c
@@ -5402,6 +5402,11 @@ static void pjsua_call_on_rx_offer(pjsip_inv_session *inv,
 	call->opt = opt;
     }
 
+    if (async) {
+	call->rx_reinv_async = async;
+	goto on_return;
+    }
+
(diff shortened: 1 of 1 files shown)


wosrediinanatour:
The 4 failed checks seems to be unrelated.

sauwming:
I'm not 100% certain, but I believe although `pjsua_media_channel_create_sdp()` can be safely skipped, `apply_call_setting()` shouldn't be, in case that the call setting is changed from the callback.
This is because `pjsua_call_answer_with_sdp()` will not call `apply_call_setting()`.

sauwming:
Also, one of the benefit of `pjsua_media_channel_create_sdp()` is that it will call `on_call_sdp_created()` so you can get the default PJSIP answer.

So IMO I prefer to leave it as it is now.

PR #2937, opened 2022-01-06
Fixed build failure on clang 13

sauwming:
Upgrading to the latest Clang 13 (the current exact version used: `(clang-1300.0.29.30)`), PJSIP will fail to build with the following errors:
```
./../speex/libspeex/bits.c:109:13: error: conflicting types for 'speex_bits_read_from'
EXPORT void speex_bits_read_from(SpeexBits *bits, char *chars, int len)
            ^
/opt/local/include/speex/speex_bits.h:80:6: note: previous declaration is here
void speex_bits_read_from(SpeexBits *bits, const char *bytes, int len);
     ^
../../speex/libspeex/bits.c:156:13: error: conflicting types for 'speex_bits_read_whole_bytes'
EXPORT void speex_bits_read_whole_bytes(SpeexBits *bits, char *chars, int nbytes)
            ^
/opt/local/include/speex/speex_bits.h:88:6: note: previous declaration is here
void speex_bits_read_whole_bytes(SpeexBits *bits, const char *bytes, int len);
     ^
2 errors generated.
```

Also in this PR: suppress some compilation warnings

### DIFF
--- a/pjlib/src/pj/ssl_sock_imp_common.c
+++ b/pjlib/src/pj/ssl_sock_imp_common.c
@@ -184,6 +184,8 @@ static pj_status_t circ_write(circ_buf_t *cb,
  *******************************************************************
  */
 
+#ifndef SSL_SOCK_IMP_USE_OWN_NETWORK
+
 /* Check IP address version. */
 static int get_ip_addr_ver(const pj_str_t *host)
 {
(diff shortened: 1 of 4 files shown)


sauwming:
False alarm. The conflict is with the built-in Speex installation.

Closing this PR.

PR #2953, opened 2022-01-21
Additional multipart improvements

gtjoseph:
Added the following APIs:
pjsip_multipart_find_part_by_header()
pjsip_multipart_find_part_by_header_str()
pjsip_multipart_find_part_by_cid_str()
pjsip_multipart_find_part_by_cid_uri()

### DIFF
--- a/pjsip/include/pjsip/sip_multipart.h
+++ b/pjsip/include/pjsip/sip_multipart.h
@@ -154,6 +154,89 @@ pjsip_multipart_find_part( const pjsip_msg_body *mp,
 			   const pjsip_multipart_part *start);
 
 /**
+ * Find a body inside multipart bodies which has a header matching the
+ * supplied one. Most useful for finding a part with a specific Content-ID.
+ *
+ * @param pool		Memory pool to use for temp space.
+ * @param mp		The multipart body.
+ * @param search_hdr	Header to search for.
+ * @param start		If specified, the search will begin at
+ * 			start->next part. Otherwise it will begin at
+ * 			the first part in the multipart bodies.
+ *
+ * @return		The first part which has a header matching the
+ * 			specified one, or NULL if not found.
+ */
+PJ_DECL(pjsip_multipart_part*)
+pjsip_multipart_find_part_by_header(pj_pool_t *pool,
+				    const pjsip_msg_body *mp,
+				    void *search_hdr,
+				    const pjsip_multipart_part *start);
+
+/**
+ * Find a body inside multipart bodies which has a header matching the
+ * supplied name and value. Most useful for finding a part with a specific
+ * Content-ID.
+ *
+ * @param pool		Memory pool to use for temp space.
+ * @param mp		The multipart body.
+ * @param hdr_name	Header name to search for.
+ * @param hdr_value	Header value search for.
+ * @param start		If specified, the search will begin at
+ * 			start->next part. Otherwise it will begin at
+ * 			the first part in the multipart bodies.
+ *
+ * @return		The first part which has a header matching the
+ * 			specified one, or NULL if not found.
+ */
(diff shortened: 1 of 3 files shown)


sauwming (approved):

nanangizz (approved):

nanangizz:
FYI, for unknown headers such as Content-ID, the header will be parsed as generic string header (i.e: `pjsip_generic_string_hdr`). App then will be able to compare the string value directly (`hvalue` field, which has type of `pj_str_t`) without printing it on first (but perhaps needs to trim the (copy of) `hvalue` first), this way is perhaps more efficient?

Also, if generic string is not suitable, app can create its own header structure and register the parser via `pjsip_register_hdr_parser()`.

gtjoseph:
> FYI, for unknown headers such as Content-ID, the header will be parsed as generic string header (i.e: `pjsip_generic_string_hdr`). App then will be able to compare the string value directly (`hvalue` field, which has type of `pj_str_t`) without printing it on first (but perhaps needs to trim the (copy of) `hvalue` first), this way is perhaps more efficient?
> 

Yeah, maybe I went a little overboard. :)  I started with just the string compare for Content-ID then thought about searching for other headers and decided to make it more generic (and more complicated).    I don't actually have a use case for searching by headers other than Content-ID at this point so let me think on this for a day or so.  

> Also, if generic string is not suitable, app can create its own header structure and register the parser via `pjsip_register_hdr_parser()`.

Technically, Content-ID is defined by as having the "name_addr" format but should be treated as a globally unique value.  Since that makes it "opaque-like" I think generic string is fine.

## HOLD FOR NOW

gtjoseph:
I just pushed up a new revision that has search by cid to just do the hvalue string matching instead of doing the print_on.  I also fixed a bug in the hdr search and added a test for it.

nanangizz (approved):

trengginas (approved):

gene-git:
FYI - After commit 551273f63969d19f6e6a52ed17a72dd720d51391 build  git master is failing for me on linux (gcc 11.1) with:

```
/usr/bin/ld: output/pjsip-test-x86_64-unknown-linux-gnu/multipart_test.o: in function `verify_cid_str':
multipart_test.c:(.text+0x25): undefined reference to `pjsip_multipart_find_part_by_cid_str'
/usr/bin/ld: output/pjsip-test-x86_64-unknown-linux-gnu/multipart_test.o: in function `verify2':
multipart_test.c:(.text+0x554): undefined reference to `pjsip_multipart_find_part_by_header'
/usr/bin/ld: multipart_test.c:(.text+0x5b7): undefined reference to `pjsip_multipart_find_part_by_cid_str'
```

sauwming:
I tried it here and it seemed to be fine. Try to clean your build first and then rebuild.

gene-git:
Thanks @sauwming - I assume was working for others given its been a few days.
Not sure what I'm doing wrong - and it was building fine - really :).  Here's what i do:

```
git pull origin
git clean -fddx
git checkout -f master

./configure ./aconfigure --prefix=/usr --enable-shared --enable-ext-sound  --disable-opencore-amr --with-external-yuv --with-external-speex  CFLAGS="-O2 -DNDEBUG"

make -s dep
make
```

(make clean not needed given git clean -fddx)

gtjoseph:
I just tried it as well with your command line and had no problem.   Do you have an earlier version of the shared libraries installed somewhere? Could the linker be picking them up instead of the ones just compiled?

gene-git:
Ah good point -thanks.

Yes - the package is installed on the build machine - ok I'll change the build machine to clean machine without pjproject installed.  I do this for a couple of other packages as well.

gene-git:
Thank you that works fine with package not installed on build machine (or clean chroot etc.). Should have realized that :)

PR #2954, opened 2022-01-24
OpenSSL server-side enhancements to support session reuse and SNI

sauwming:
In this PR, there are a few enhancements for server usage:
1. Support session reuse
We need to create only one SSL server context and maintain it if we want to reuse the SSL session created from that context.

2. Support Server Name Indication
Implement SNI callback which will check the server name passed by the client and fail the handshake upon mismatch (note that as per the [RFC3546](https://tools.ietf.org/html/rfc3546#section-3.1), server name must not be an IP address).

3. Add option to disable session tickets
SSL/TLS supports two mechanisms for resuming sessions: session ids and stateless session tickets. Here we provide an option to disable session tickets.

Client use should be unaffected by this PR.

### DIFF
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -57,9 +57,6 @@
 #include <openssl/opensslconf.h>
 #include <openssl/opensslv.h>
 
-/* A single instance of server SSL context. */
-static SSL_CTX *server_ctx = NULL;
-
 /* Specify whether server supports session reuse using session ID. */
 #define SERVER_SUPPORT_SESSION_REUSE 1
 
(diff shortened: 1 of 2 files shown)


pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -57,6 +57,26 @@
 #include <openssl/opensslconf.h>
 #include <openssl/opensslv.h>
 
+/* A single instance of server SSL context. */
+static SSL_CTX *server_ctx = NULL;
```
nanangizz, line 61: instead of static var or singleton, is it possible to move this to ssl sock structure (e.g: only set for the listener socket)? so server ssl socks from the same listener will share the same SSL_CTX.

sauwming: okay. will try to incorporate Peter's patch as well.

trengginas (approved):

nanangizz (approved):

PR #2957, opened 2022-01-27
Fix documentation and usage of txdata ref counter

wosrediinanatour:
The documentation is not according to `pjsip/src/pjsip/sip_util.c:513`.
The Statful Proxy should handle one error case, similar to rest of the code.

### DIFF
--- a/pjsip-apps/src/samples/stateful_proxy.c
+++ b/pjsip-apps/src/samples/stateful_proxy.c
@@ -346,6 +346,7 @@ static pj_bool_t proxy_on_rx_response( pjsip_rx_data *rdata )
     status = pjsip_endpt_send_response(global.endpt, &res_addr, tdata,
 				       NULL, NULL);
     if (status != PJ_SUCCESS) {
+	pjsip_tx_data_dec_ref(tdata);
 	app_perror("Error forwarding response", status);
 	return PJ_TRUE;
     }
--- a/pjsip/include/pjsip/sip_util.h
+++ b/pjsip/include/pjsip/sip_util.h
@@ -322,7 +322,7 @@ pjsip_endpt_create_request_from_hdr( pjsip_endpoint *endpt,
  * will construct all the Via, Record-Route, Call-ID, From, To, CSeq, and 
  * Call-ID headers from the request.
  *
- * Note: the txdata reference counter is set to ZERO!.
+ * Once a transmit data is created, the reference counter is initialized to 1.
  *
  * @param endpt	    The endpoint.
  * @param rdata	    The request receive data.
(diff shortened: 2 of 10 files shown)


wosrediinanatour:
Failed test seems to be unrelated: ```caller: 09:55:57.514     strm0x7fb07801d778  Error sending RTCP: SRTP parameters negotiation still in progress (PJMEDIA_SRTP_EKEYNOTREADY)
callee: ================== Timed-out! ================== [at t=018]
====== Caught error: callee : Timeout expecting pattern: "Incoming DTMF on call [0-9]+: 1" ====== ```

pjsip/include/pjsip/sip_util.h
```diff
@@ -322,7 +322,7 @@ pjsip_endpt_create_request_from_hdr( pjsip_endpoint *endpt,
  * will construct all the Via, Record-Route, Call-ID, From, To, CSeq, and 
  * Call-ID headers from the request.
  *
- * Note: the txdata reference counter is set to ZERO!.
+ * Note: the txdata reference counter is set to unity!
```
nanangizz, line 325: Minors:
- What if just use the same wordings as `pjsip_endpt_create_request*()`, i.e.: `Once a transmit data is created, the reference counter is initialized to 1.`
- Also while in this, perhaps better to also add the same info for `pjsip_endpt_create_ack()` & `pjsip_endpt_create_cancel()`.

pjsip-apps/src/samples/stateful_proxy.c
```diff
@@ -346,6 +346,7 @@ static pj_bool_t proxy_on_rx_response( pjsip_rx_data *rdata )
     status = pjsip_endpt_send_response(global.endpt, &res_addr, tdata,
 				       NULL, NULL);
     if (status != PJ_SUCCESS) {
+	pjsip_tx_data_dec_ref(tdata);
```
nanangizz, line 349: After searching for `pjsip_endpt_create_response` in the library code, looks like there are some other places need to be fixed as well.

We can fix those if you like, but, in case it has not been set, we will need you to allow us to apply changes to the fork as described [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

nanangizz (commented):
I think you're right, it seems that the `pjsip_endpt_create_response` implementation was changed [here](https://github.com/pjsip/pjproject/commit/0ca04b62f681e0b9dcb0f6050278fd15fc01b0b8#diff-8640b1757d0d85f5c2f6caa547c797768b9cbb02fce14522172bef085738254aR360), but the doc was not updated.

Re: CI error, yes, we can ignore it.

pjsip-apps/src/samples/stateful_proxy.c
wosrediinanatour: Yes, please those. Thank you! ("allow edits by maintainers" is selected)

wosrediinanatour: @nanangizz - I made changes you suggested regarding docu and added `pjsip_tx_data_dec_ref(tdata);` to `pjsip_endpt_send_response` in case of an error. Please check and feel free to improve/update/fix.

nanangizz:
Sorry for the long response. Just get a chance to get back to this and it looks like you've fixed the other places as well, thank you, very appreciate it.

nanangizz (approved):

sauwming (approved):

PR #2958, opened 2022-01-28
Avoid assertions when using PJ_HAS_THREADS 0

schlagert:
This fixes assertions occuring when using PJSIP without thread support. The assertions get triggered when `pj_thread_this` gets called.

Please be aware that in additon to this patch you will also need to exclude `PJ_LOG_HAS_THREAD_ID` and `PJ_LOG_HAS_THREAD_SWC `from your logging decor.

This patch sets the `grp lock owner` pointer to a constant invalid value. While this is ugly, the lock owner should not equal `NULL` while locked and a constant seemed the simplest solution to achieve this.

### DIFF
--- a/pjlib/src/pj/lock.c
+++ b/pjlib/src/pj/lock.c
@@ -260,17 +260,25 @@ PJ_DEF(void) pj_grp_lock_config_default(pj_grp_lock_config *cfg)
 static void grp_lock_set_owner_thread(pj_grp_lock_t *glock)
 {
     if (!glock->owner) {
+#if PJ_HAS_THREADS
 	glock->owner = pj_thread_this();
+#else
+        glock->owner = (pj_thread_t *) -1;
+#endif
 	glock->owner_cnt = 1;
     } else {
+#if PJ_HAS_THREADS
 	pj_assert(glock->owner == pj_thread_this());
+#endif
 	glock->owner_cnt++;
     }
 }
 
 static void grp_lock_unset_owner_thread(pj_grp_lock_t *glock)
 {
+#if PJ_HAS_THREADS
     pj_assert(glock->owner == pj_thread_this());
+#endif
     pj_assert(glock->owner_cnt > 0);
     if (--glock->owner_cnt <= 0) {
 	glock->owner = NULL;
--- a/pjlib/src/pj/log.c
+++ b/pjlib/src/pj/log.c
@@ -167,12 +167,28 @@ pj_status_t pj_log_init(void)
     }
 #endif
     g_last_thread = NULL;
+
+    /* Normalize log decor, e.g: unset thread flags when threading is
+     * disabled.
+     */
+    pj_log_set_decor(pj_log_get_decor());
+
     return PJ_SUCCESS;
 }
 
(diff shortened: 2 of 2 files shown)


nanangizz:
Thank you for the PR.

Also, good points regarding log decor flags `PJ_LOG_HAS_THREAD_ID` & `PJ_LOG_HAS_THREAD_SWC`. So just tried to create a patch for this as follow, would you mind to also add it to the PR (or perhaps allow maintainer to push to the PR).

```
diff --git a/pjlib/src/pj/log.c b/pjlib/src/pj/log.c
index cf7ac3789..cee37ca1b 100644
--- a/pjlib/src/pj/log.c
+++ b/pjlib/src/pj/log.c
@@ -167,12 +167,28 @@ pj_status_t pj_log_init(void)
     }
 #endif
     g_last_thread = NULL;
+
+    /* Normalize log decor, e.g: unset thread flags when threading is
+     * disabled.
+     */
+    pj_log_set_decor(pj_log_get_decor());
+
     return PJ_SUCCESS;
 }

 PJ_DEF(void) pj_log_set_decor(unsigned decor)
 {
     log_decor = decor;
+
+#if !PJ_HAS_THREADS
+    /* Unset thread related flags */
+    if (log_decor & PJ_LOG_HAS_THREAD_ID) {
+       log_decor &= ~(PJ_LOG_HAS_THREAD_ID);
+    }
+    if (log_decor & PJ_LOG_HAS_THREAD_SWC) {
+       log_decor &= ~(PJ_LOG_HAS_THREAD_SWC);
+    }
+#endif
 }

 PJ_DEF(unsigned) pj_log_get_decor(void)
```

sauwming:
And since we more commonly associate something invalid with -1 instead, won't it be better if this is also the same here? (here it will set the `glock->owner` to `0xffffffffffffffff` on 64-bit machine instead of `0x1`)

schlagert:
> And since we more commonly associate something invalid with -1 instead, won't it be better if we also do the same here?

Agree, this would be a better choice.

schlagert:
@nanangizz I would be fine with you do the change(s).

nanangizz:
> @nanangizz I would be fine with you do the change(s). Do I have to grant you access anywhere or will you do the additional changes on a branch owned by `pjsip`?

Please check [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), if it is possible, it would be practical. Otherwise, we may need to create a separate PR.

schlagert:
Unfortunately, that button does not show up :|

I'll apply the proposed changes myself to this PR, that's probably the least work for everyone.

schlagert:
Would you like everything in one single commit or should I keep them separate?

nanangizz:
> Would you like everything in one single commit or should I keep them separate?

IMO single commit is okay, I don't think there will be any roll/revert back on those :) And usually we squash the commits in merging to the master branch.

schlagert:
Here you go, I amended the commit.

nanangizz (approved):
Thank you.

sauwming (approved):

PR #2959, opened 2022-01-28
Add PJSUA2 callback for SIP INFO supporting a response

wosrediinanatour:
The call class has a new onIncomingInfo() callback to be able to
send a response easily.

### DIFF
--- a/pjsip/include/pjsua2/call.hpp
+++ b/pjsip/include/pjsua2/call.hpp
@@ -715,6 +715,27 @@ struct OnCallTsxStateParam
 };
 
 /**
+ * This structure contains parameters for Call::onIncomingInfo() callback.
+ */
+struct OnIncomingInfo
+{
+    /**
+     * The SIP INFO's body.
+     */
+   std::string body;
+
+   /**
+    * MIME content type
+    */
+   std::string  type;
+
+   /**
+    * MIME content sub type
+    */
+   std::string  subtype;
+};
+
+/**
  * This structure contains parameters for Call::onCallMediaState() callback.
  */
 struct OnCallMediaStateParam
@@ -1818,6 +1839,22 @@ public:
      */
     virtual void onCallTsxState(OnCallTsxStateParam &prm)
     { PJ_UNUSED_ARG(prm); }
+
+    /**
+     * Notify application about an incoming INFO.
+     *
+     * @param prm	Callback parameter.
+     *
+     * @return
+     *  - Status code >= PJSIP_SC_OK triggers a response.
+     *  - Status code PJSIP_SC_NULL for no action.
+     *    (e.g. to use build-in support of DTMF content in SIP INFO)
+     */
+    virtual pjsip_status_code onIncomingInfo(OnIncomingInfo &prm)
+    { 
+	PJ_UNUSED_ARG(prm); 
+	return PJSIP_SC_NULL;
+    }
     
     /**
      * Notify application when media state in the call has changed.
--- a/pjsip/src/pjsua2/endpoint.cpp
+++ b/pjsip/src/pjsua2/endpoint.cpp
@@ -1127,8 +1127,6 @@ void Endpoint::on_call_tsx_state(pjsua_call_id call_id,
                                  pjsip_transaction *tsx,
                                  pjsip_event *e)
 {
-    PJ_UNUSED_ARG(tsx);
-
     Call *call = Call::lookup(call_id);
     if (!call) {
 	return;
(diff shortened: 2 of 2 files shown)


dawon (approved):

wosrediinanatour:
I have rebased on current master (hope that tests are fixed now).

wosrediinanatour:
Rebased on latest master, fixed compilation (Keyword "auto" not supported)

nanangizz:
Could you also describe the scenarios where such callback is needed? Basically we want to know if the scenario is quite common and/or perhaps it is a burdensome to be implemented in the app, so there will be a "strong" reason to add it to the library.

Also it would be better if the approach is quite generic (may cover other scenario in the future as well, e.g: if app needs to add some header/body in the response).

Anyway, thank you for the patch.

wosrediinanatour:
@nanangizz : With SIP INFO you can carry "carry application level information along the SIP signaling path" (https://datatracker.ietf.org/doc/html/rfc2976), a functionality which is very useful in some apps (PJSUA supports DTMF already, but think of push-to-talk or some metadata, heartbeating, configuration....). Sending a SIP INFO is quite easy, i.e. similar to
```cpp
CallSendRequestParam call_parameter;                                   
call_parameter.method = "INFO";                                                   
call_parameter.txOption.contentType = "application/json";
call_parameter.txOption.msgBody = body;                                       
send_request(call_parameter);
```
Reasons, why I opt for including a callback:
* You have to use PJSUA functions in PJSUA2's overloaded `onCallTsxState()` - see codelet below:
   * In order to send responses of SIP INFO, you need a very good knowledge about PJSUA.
   * Beyond the scope of https://www.pjsip.org/docs/book-latest/html/call.html
   * It is error prone.
* Use "leaking" abstraction, i.e. use a pointer in `OnCallTsxStateParam` (see the how to access PJSUA's transaction `tsx` in the following).
* A body is not required: " A 200 OK response MUST be sent by a UAS for an INFO request with no message body if the INFO request was successfully received for  an existing call." (https://datatracker.ietf.org/doc/html/rfc2976)
* Possibility of adding headers could be easily achived by adding an attribute to `struct OnIncomingInfo`. API change would be compatible.

In an app, a code would look like the following:
```C
if (tsx->role==PJSIP_ROLE_UAS && 
        tsx->state==PJSIP_TSX_STATE_TRYING &&
        pjsip_method_cmp(&tsx->method, &pjsip_info_method)==0)
{
      auto* e = static_cast<pjsip_event*>(transaction.e.pjEvent);
      ... // ensure != nullptr
      pjsip_transaction* tsx = e->body.tsx_state.tsx;
      ... // ensure != nullptr
      pjsip_rx_data* rdata = e->body.tsx_state.src.rdata;
     ... // ensure != nullptr
      pjsip_tx_data* tdata = nullptr;

      if (pjsip_endpt_create_response(tsx->endpt, rdata, code, nullptr, &tdata) != PJ_SUCCESS)
      {
         // Failed to create response"
         // Is it save to just return from the callback?
         return;
      }
      if (pjsip_tsx_send_msg(tsx, tdata) != PJ_SUCCESS)
      {
         pjsip_tx_data_dec_ref(tdata);
         //Sending failed"
         // Is it save to just return from the callback?
         return;
      }
 }```

Instead, with `virtual pjsip_status_code onIncomingInfo(OnIncomingInfo &prm)`
above code might look like this:
```cpp
   pjsip_status_code onIncomingInfo(pj::OnIncomingInfo& prm) override
   {
      auto id = getId();
      ... // Post prm.body and id to APP.
      return PJSIP_SC_OK;
   }
```

wosrediinanatour:
PS: Another generic approach would be to have a method `answer(const OnCallTsxStateParam&, const SomeConfigurationForHeadersAndResponseCode)` for use in onCallTsxState. From my point of view I see onCallTsxState more for logging/debugging or developing on a low level PJSIP code.

wosrediinanatour:
@nanangizz : what is your opinion? Would you have a different approach? 
My main objective is to make receiving SIP INFO and transmitting the response much easier for an application. I would either rework it, or close the PR. If you have another idea about a much simplier approach (e.g. another class for transactions) please tell.

sauwming:
Sorry if you're not hearing from us earlier. Let me give my two cents.

First, it's unusual for pjsua2 to have its own callback (I believe this is the only one) without having the same corresponding callback in the pjsua level (i.e. `pjsua_callback.on_incoming_info()`).

Second, it may potentially clash with the library's handling of incoming INFO in pjsua. I know you already mention in the doc that app should return PJSIP_SC_NULL to use the built-in default handling by the library, but it's not obvious under what circumstances (i.e. what they should check?) they should do so (you only mention DTMF, but how about trickle ICE)? Just for some example, app will probably NOT want to handle the incoming INFO if it's of subtype "media_control+xml", or "dtmf-relay", or "trickle-ice-sdpfrag". So this would require an in-depth knowledge of PJSIP, which defeats the purpose of this callback that's supposed to make it easier.

So in short, the PR will need a lot of rework if it is to be integrated to PJSIP.

wosrediinanatour:
Thx for the reply.
The main reason for this PR is the many code lines needed for just getting the body of SIP INFO and reply with 200 OK. Do you have any idea for a simplier approach?

sauwming:
Since you already have it implemented, if it's working well for your use case, there's no need to change it.

And especially since you can put it in your own subclass implementation (i.e. in the child class of Call), which means that you don't need to modify the library code.
If you still want to integrate it to PJSIP, that would be great of course, but it would require many changes.

wosrediinanatour:
@sauwming: Ok.
One (I hope) last question in current context and out of curiosity: Have I missed something or isn't it possible for pure C++ (PJUSA2, and hence Python, Java, Android,...) to send a response for an incoming INVITE? Hence, without C functions like `pjsip_endpt_create_response` etc. it is not possible to reply to incoming requests like SIP INFO?
Would it make sense to have a new method in pj::Call?

PR #2960, opened 2022-01-28
Fix use after free

typetetris:
uiae


typetetris:
Oh my, now I have to bring all that stuff CLA an so on back to my employer, so it is all ok. Paperwork as reward for bringing that to your attention :)

typetetris:
And I will not grant this Bitrise thingy WRITE access to my repos, just so I can look, what the build failure for your CI was.

typetetris:
Maybe I should send you an E-Mail.

PR #2961, opened 2022-01-28
Add additional unit test options

gtjoseph:
All of the unit tests will now accept a '-n/--no-trap' option
that will prevent the test's main.c wrapper from intercepting the
SIGSEGV and SIGABRT signals.  This allows the OS to generate
a coredump which can be examined by gdb.

The pjsip unit test wrapper will now accept a '-t/--tests' option
that will take a list of tests to run.  If you're working on
a specific area, this makes it easier to just test that area
without having to comment out the defines in test.h. The default
is to run all tests.

Added *bench*.htm to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,5 @@ tests/pjsua/wavs/tmp*.wav
 tests/pjsua/tools/cmp_wav
 tests/pjsua/tools/cmp_wav.dSYM
 tests/pjsua/logs
+# benchmark output
+*bench*.htm
--- a/pjlib-util/src/pjlib-util-test/main.c
+++ b/pjlib-util/src/pjlib-util-test/main.c
@@ -51,7 +51,7 @@ static void print_stack(int sig)
     exit(1);
 }
 
-static void init_signals()
+static void init_signals(void)
 {
     signal(SIGSEGV, &print_stack);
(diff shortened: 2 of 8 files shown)


sauwming:
The Windows CI tests seem to be failing.

nanangizz:
Nice features, thanks!

Just a minor issue, from CI it looks like there is linking error `unresolved external symbol _strtok_r` on Windows platforms (or perhaps specifically MSVC). Based on [this](https://stackoverflow.com/questions/9021502/whats-the-difference-between-strtok-r-and-strtok-s-in-c), perhaps a simple fix is as follow:
```
#ifdef _MSC_VER
# define strtok_r strtok_s
#endif
```

gtjoseph:
The Windows tests still seem to be failing but the default one is failing because of a SSL/TLS issue and the others for a file not found.  Neither seems to be related to my pull request.

sauwming:
I have fixed the Windows CI tests, so please merge the current master.

gtjoseph:
> I have fixed the Windows CI tests, so please merge the current master.

Done.

nanangizz (approved):

trengginas (approved):

sauwming (approved):

PR #2962, opened 2022-01-31
Implement verify callback to TLS connection

trengginas:
This patch will allow application to specify it's verification to a new TLS connection.
Currently, application can do the verification via `pjsip_tp_state_callback` (e.g: shutdown the connection when verification failed). This will allow remote to send data during application verification. 
The callback will allow connection disconnection immediately if verification failed.

### DIFF
--- a/pjlib/include/pj/ssl_sock.h
+++ b/pjlib/include/pj/ssl_sock.h
@@ -671,7 +671,8 @@ typedef struct pj_ssl_sock_cb
     
     /**
      * This callback is called when certificate verification is being done.
-     * Certification info can be obtained from #pj_ssl_sock_info.
+     * Certification info can be obtained from #pj_ssl_sock_info. Currently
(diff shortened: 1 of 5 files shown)


pjlib/include/pj/ssl_sock.h
```diff
@@ -668,6 +668,20 @@ typedef struct pj_ssl_sock_cb
      */
     pj_bool_t (*on_connect_complete)(pj_ssl_sock_t *ssock,
 				     pj_status_t status);
+    
+    /**
+     * This callback is called when certificate verification is being done.
+     * Certification info can be obtained from #pj_ssl_sock_info.
```
sauwming, line 674: Add info that this callback is currently only implemented for OpenSSL backend.
Also below for transport_tls

nanangizz (approved):

sauwming (approved):

PR #2971, opened 2022-02-09
do not #define snprintf and vsnprintf on Visual Studio 2015 and newer

andreas-wehrmann:
Without this, compilation is broken when using std::snprintf or std::vsnprintf in C++

### DIFF
--- a/pjlib/include/pj/compat/string.h
+++ b/pjlib/include/pj/compat/string.h
@@ -46,8 +46,13 @@
 #if defined(_MSC_VER)
 #   define strcasecmp	_stricmp
 #   define strncasecmp	_strnicmp
+
+/* snprintf() and vsnprintf() are available since Visual Studio 2015 */
+#if _MSC_VER < 1900
 #   define snprintf	_snprintf
 #   define vsnprintf	_vsnprintf
+#endif
+
 #   define snwprintf	_snwprintf
 #   define wcsicmp	_wcsicmp
 #   define wcsnicmp	_wcsnicmp


sauwming:
We use VS 2015 in our development and our CI build also uses VS2019 and there's no problem so far, so I'm not sure why this patch is necessary?
https://github.com/pjsip/pjproject/blob/master/.github/workflows/ccpp.yml#L407

Note: the CI tests are failing right now but for an unrelated reason, and I'm still trying to fix it.

andreas-wehrmann:
Because it breaks C++ Code that uses snprintf and vsnprintf from the std namespace, because these defines will cause std::snprintf to become std::_snprintf which doesn't exist.

As a minimal example (which I should have provided in the first place),
try this (I don't have VS2015 here, but use VS2019):
```
#include <cstdio>
#include <pj/string.h>

int main(void)
{
	enum { bufsiz = 128 };
	char tmp[ bufsiz ];
	std::snprintf( tmp, (bufsiz - 1), "hello world" );
	return 0;
}
```

Furthermore, one should refrain from re-#defining things that are part of the standard library because
you may never now what might break or if it introduces subtle changes
in behaviour that might break other parts of the code.

In my case the problem wasn't even triggered within my part of the code, but inside a library that I use.

sauwming:
I have fixed the Windows CI tests, so please merge the current branch to run the tests again.

nanangizz:
> Furthermore, one should refrain from re-#defining things that are part of the standard library because you may never now what might break or if it introduces subtle changes in behaviour that might break other parts of the code.

Looks like the `snprintf()` is not part of C standard yet before C99, so older MS compilers (with C89, perhaps until VS2013?) provide the implementation with underscore prefix only, hence the `#define` macros are set. And agree, it should not be re-defined in the newer MS compilers.

nanangizz (approved):

sauwming (approved):

Issue #2964, opened 2022-02-03
Compiling for Android with NDK 23 and video support fails


goffioul:
### Describe the bug

When compiling PJSIP for Android with NDK 23 and video support (e.g. with libvpx), linking `libpjsua2.so` fails with unknown reference to `ANativeWindow_fromSurface`.

This comes from the detection failure for OpenGL ES 2 in configure script. The script has hard-coded `-lgcc`, but the library has been removed from NDK 23. This results in missing libraries when linking `libpjsua2.so`, in particular `-landroid`.

### Steps to reproduce

Try to compile PJSIP for android with NDK 23 and libvpx.

### PJSIP version

2.10

### Context

Android NDK 23

### Log, call stack, etc

```shell
N/A
```

PR #2974, opened 2022-02-11
Fixed Android build issues on NDK r23 and r24

sauwming:
As reported in #2964, Android SWIG build failed with undefined reference to `ANativeWindow_fromSurface`. The solution is to link with `-landroid`.

I tested with Android NDK r24b and the original report is for NDK r23.

[Internal]: If you're using an older NDK, please help test to make sure it doesn't break the build.


trengginas (dismissed):

sauwming:
Just added another commit to fix OpenGL ES 2 detection. Please help retest.

sauwming:
I just tested on my older laptop with NDK r21b and it seems okay.

trengginas (approved):

trengginas:
> I just tested on my older laptop with NDK r21b and it seems okay.

tested here on r21e and without issue

nanangizz:
tried using NDK r20, built okay.

nanangizz:
Tried using NDK r19, also build okay (`PJMEDIA_HAS_VIDEO` set and swig).

nanangizz (approved):

PR #2979, opened 2022-02-14
Release wav player and recorder pool to clean up when calling pjsua_destroy()

trengginas:
This patch will call `pjsua_player_destroy()`/`pjsua_recorder_destroy()` if app haven't prior to `pjsua_destroy()` / `pjsua_destroy2()`.
To make sure that the resources are released properly.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_aud.c
+++ b/pjsip/src/pjsua-lib/pjsua_aud.c
@@ -472,7 +472,6 @@ pj_status_t pjsua_aud_subsys_destroy()
 	if (pjsua_var.player[i].port) {
 	    PJ_LOG(2,(THIS_FILE, "Destructor for player id=%d is not called"));
 	    pjsua_player_destroy(i);
-	    pjsua_var.player[i].port = NULL;
 	}
     }
 
@@ -481,8 +480,7 @@ pj_status_t pjsua_aud_subsys_destroy()
 	if (pjsua_var.recorder[i].port) {
 	    PJ_LOG(2,(THIS_FILE, "Destructor for recorder id=%d "
 		      "is not called"));
-	    pjsua_recorder_destroy(i);	    
-	    pjsua_var.recorder[i].port = NULL;
+	    pjsua_recorder_destroy(i);
 	}
     }
 
--- a/pjsip/src/pjsua-lib/pjsua_aud.c
+++ b/pjsip/src/pjsua-lib/pjsua_aud.c
@@ -479,6 +479,9 @@ pj_status_t pjsua_aud_subsys_destroy()
 
     /* Destroy file players */
     for (i=0; i<PJ_ARRAY_SIZE(pjsua_var.player); ++i) {
+	if (pjsua_var.player[i].pool) {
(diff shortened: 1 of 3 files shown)


pjsip/src/pjsua-lib/pjsua_core.c
```diff
@@ -2033,6 +2033,22 @@ PJ_DEF(pj_status_t) pjsua_destroy2(unsigned flags)
...
+	    }
+        }
+
+	/* Destroy pool in the wav writer object */
+        for (i = 0; i < (int)PJ_ARRAY_SIZE(pjsua_var.recorder); ++i) {
+	    if (pjsua_var.recorder[i].pool != NULL) {
+		pj_pool_release(pjsua_var.recorder[i].pool);
+		pjsua_var.recorder[i].pool = NULL;
+	    }
+        }
```
sauwming, line 2051: Shouldn't this be in `pjsua_aud_subsys_destroy()`?
https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_aud.c#L480

nanangizz, line 2051: Pool will automatically be released when the factory is destroyed, so pjsua-lib does not need to release it manually here. Even better the auto release comes with warning messages in the log so app will be aware of it (and hopefully fix it).

Also, if resource leak is the concern, shouldn't it invoke `pjsua_player/recorder_destroy()` instead of `pj_pool_release()`, so any file descriptors will also be released?

pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -479,6 +479,9 @@ pj_status_t pjsua_aud_subsys_destroy()
 
     /* Destroy file players */
     for (i=0; i<PJ_ARRAY_SIZE(pjsua_var.player); ++i) {
+	if (pjsua_var.player[i].pool) {
+	    PJ_LOG(2,(THIS_FILE, "Pool for player id=%d is not released"));
+	}
 	if (pjsua_var.player[i].port) {
 	    pjmedia_port_destroy(pjsua_var.player[i].port);
```
nanangizz, line 486: What if `pjmedia_port_destroy()` is replaced with `pjsua_player_destroy()`?

pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -479,6 +479,9 @@ pj_status_t pjsua_aud_subsys_destroy()
 
     /* Destroy file players */
     for (i=0; i<PJ_ARRAY_SIZE(pjsua_var.player); ++i) {
+	if (pjsua_var.player[i].pool) {
+	    PJ_LOG(2,(THIS_FILE, "Pool for player id=%d is not released"));
+	}
```
nanangizz, line 484: Perhaps the warning message is better about "destructor not called" instead of "pool not released" (can be put in the same "if" block below)?

sauwming (approved):
`pjsua_var.player[i].port = NULL;` should have already been done inside `pjsua_player_destroy()`, so can be removed. Same for recorder.

Also don't forget to change the PR desc.

nanangizz (approved):

PR #2981, opened 2022-02-15
Pjsua2 Swift sample project bug fixes & feature updates.

emretufekci:
- NAT Settings
- Saving configurations and fetching it from local db.
- Scroll Bar


sauwming:
Thanks for the enhancements, but it's quite complicated to be integrated as a simple sample app. Nevertheless, we'll refer our users to this project if they need an example of a more full-featured pjsua2 swift app.

PR #2986, opened 2022-02-16
Notify the TU if the transport for a UAS tsx enters disconnect or error state

ajw3-msw:
RFC 3261 Section 17.2.4 (https://www.rfc-editor.org/rfc/rfc3261.html#section-17.2.4) suggests that, when handling Transport Errors as a Server:

> ... the server transaction SHOULD inform the TU that a failure has occurred, and SHOULD transition to the terminated state.

I appreciate the flag `PJSIP_TSX_UAS_CONTINUE_ON_TP_ERROR` prevents us from transitioning to the terminated state (and I can see how that is useful).

This PR proposes also informing the TU when such a failure occurs, so the TU can influence the response target as the RFC suggests.

### DIFF
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -1181,7 +1181,7 @@ static void tsx_timer_callback( pj_timer_heap_t *theap, pj_timer_entry *entry)
 	    tsx_update_transport(tsx, NULL);
 
 #if PJSIP_TSX_UAS_CONTINUE_ON_TP_ERROR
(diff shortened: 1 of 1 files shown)


silentindark:
@sauwming This patch will be also included in 2.12?

sauwming:
Will try if possible. The PR is currently failing PJSIP test.
Additionally, we also need @ajw3-msw to sign the CLA.

ajw3-msw:
I appreciate your quick response folks. I was delayed a bit because I wasn't able to sign the CLA trivially.

I'm going to close this particular PR - I'll follow up when I've resolved the CLA issue on my side.

Thanks!

Issue #2998, opened 2022-02-28
Option to randomise the start port in transport config


sauwming:
`pjsua_transport_config` currently has option to specify the start port and port range. The start port is currently fixed, but sometimes it may be desirable to randomise it, such as if multiple clients use the same rtp start port, and/or if the port reuse takes some time to release.

The suggestion is to add `pj_bool_t randomize_start_port` in `pjsua_transport_config`, and the implementation (for RTP) should be something like this:
```
 int start_port_offset = ((pj_rand() % (tt_var.cfg.media_cfg.rtp_port_range)) / 2) * 2;
 pjsua_var.acc[mcptt_var.account_id].next_rtp_port = tt_var.cfg.media_cfg.rtp_start_port + start_port_offset;
```

PR #3000, opened 2022-03-01
Add option to randomize RTP starting port

sauwming:
To implement #2998.

Also change the behavior of `pjsua_acc_modify()` to reset `acc->next_rtp_port` if `rtp_cfg.port` is changed.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -3011,6 +3011,16 @@ typedef struct pjsua_transport_config
     unsigned		port_range;
 
     /**
+     * Specify whether to randomly pick the starting port number from
+     * the range of [\a port, \a port + \a port_range]. This setting is
+     * used only if both port and port_range are non-zero, and only
+     * applicable for the port selection of UDP and loop media transport.
+     * 
+     * Default is PJ_FALSE.
+     */
+    pj_bool_t		randomize_port;
+
+    /**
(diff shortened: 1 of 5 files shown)


sauwming:
One thing I'm still not sure of is whether we do need to to reset `acc->next_rtp_port` in `pjsua_acc_modify()` if `rtp_cfg.port` is changed. I don't think the current behavior (which is to continue with the next rtp port as long as it's within the specified port and port range) is a bug, so let me know your opinions whether resetting or not resetting is the preferable behavior.

nanangizz (approved):
Minor: indentation, I believe it may look just fine on the IDE, but here perhaps it can be improved for better readibility (if possible)?

Just FYI, usually in MSVC2005, I use tab key to indent and the IDE will give 4 spaces for each tab key, but it seems to automatically replace spaces to tab char everytime the spaces reach 8 chars. So there will be mixed tabs & spaces for indents with not-multiplication-of-8-chars.

nanangizz:
> It looked just fine here, both on the PR webpage (screenshot attached -> this was taken before the latest commit of course) and on the IDE. If there's any particular part that still doesn't look right, let me know. Screenshot 2022-03-02 at 10 46 22 AM

Ah, it still looks different on firefox after the latest commit. But just also checked on chrome, and it looks fine indeed. So, I guess we can move on.

trengginas (approved):

PR #3001, opened 2022-03-01
wrong default value for jitter buffer max

liviu-rolo:
jitter buffer max is 500 ms, as seen in pjsua.h and pjmedia_stream_create (jb_max = 500 / stream->codec_param.info.frm_ptime;)


sauwming:
Right, the pjsua's doc has been fixed in #2886:
https://github.com/pjsip/pjproject/pull/2886/files#diff-5b56644fdd5ebd9a45018cc1f90f76f27e60cf1843cd6bcf2d0f02c8095fcb48R6845

sauwming (approved):

trengginas (approved):

Issue #638, opened 2008-09-19
Update jitter buffer to avoid/reduce such case of shrinking and underflow occuring one after another.


pjsipbot:
_2008-09-19 20:21:08: @nanangizz_ _created the issue on trac ticket [638](https://trac.pjsip.org/repos/ticket/638)_

Proposed solutions:
 1. Shrinking to be based on recent maximum burst level. This will make jitter buffer shrinking a bit less aggressive. Currently shrinking is done based on current burst level (presented by jb_prefetch).
 2. Reenable prefetch buffering. This should be able to be configured in run-time, e.g: setting initial-prefetch > 0 will make this feature enabled.

___
_2008-09-23 17:46:50: @nanangizz_ _edited the issue description_

___
_2008-09-23 17:56:27: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2008-09-23 17:56:27: @nanangizz_ _set **resolution** to **fixed**_

___
_2008-09-23 17:56:27: @nanangizz_ _commented_

In [r2310](../commit/19551d6ce7167b03c0f9074fb6d03f9b70d7c880):
 - Jitter buffer shrinking is done based on max_hist_level (recent maximum burst level).
 - Updated jitter buffer to maintain max_hist_level instead of resetting it each time current level updated.
 - Updated default jitter buffer size in stream.c from 360ms to 500ms to accomodate more burst level.

___
_2008-09-24 18:07:16: @nanangizz_ _changed **status** from **closed** to **reopened**_

___
_2008-09-24 18:07:16: @nanangizz_ _removed **resolution** (was fixed)_

___
_2008-09-24 18:25:20: @nanangizz_ _edited the issue description_

___
_2008-09-24 18:25:20: @nanangizz_ _changed **title** from **Update jitter buffer shrinking to be based on recent maximum burst level.** to **Update jitter buffer to avoid/reduce such case of shrinking and underflow occuring one after another.**_

___
_2008-09-24 18:28:14: @nanangizz_ _commented_

In [r2317](../commit/883d8de2a0fefd20e36c9bf52f6868aeaf9057cc): Reenable prefetch buffering via initial prefetch setting (set 0 to disable, otherwise to enable).

___
_2008-09-25 11:01:34: @bennylp_ _changed **status** from **reopened** to **closed**_

___
_2008-09-25 11:01:34: @bennylp_ _set **resolution** to **fixed**_

___
_2008-09-25 12:06:53: @nanangizz_ _commented_

[...]

PR #3006, opened 2022-03-04
wrong documentation about default jbInit

liviu-rolo:
"updated wrong documentation about default jb_init value in pjsua. The default is 0, not 150."

Duplicates
commit 2f2de4cc3ba0b186713447c3e993a09543253b56
Author: Benny Prijono <bennylp@teluu.com>
Date:   Tue Dec 3 05:22:10 2013 +0000


liviu-rolo:
As a related question, please, why is the JB prefetching disabled by default ?

sauwming (approved):

sauwming:
I believe it's because of #638.

trengginas (approved):

PR #3007, opened 2022-03-04
Wrong default values for JB prefetch

liviu-rolo:
wrong default values, pjsua and pjsua2:
* Jitter buffer minimum prefetch
* Jitter buffer maximum prefetch

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -6849,14 +6849,14 @@ struct pjsua_media_config
     /**
      * Jitter buffer minimum prefetch delay in msec.
      *
-     * Default: -1 (to use default stream settings, currently 60 msec)
+     * Default: -1 (to use default stream settings, currently codec ptime)
      */
     int			jb_min_pre;
     
     /**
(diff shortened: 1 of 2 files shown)


pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -6849,14 +6849,14 @@ struct pjsua_media_config
     /**
      * Jitter buffer minimum prefetch delay in msec.
      *
-     * Default: -1 (to use default stream settings, currently 60 msec)
+     * Default: -1 (to use default stream settings, currently codec ptime)
```
sauwming, line 6852: currently is 1, I believe.
https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream.c#L2694

liviu-rolo: Yes, because 1 = stream->codec_param.info.frm_ptime/stream->codec_param.info.frm_ptime.
But I preferred using time related terms (and units) for consistency with the other values and minimal impact. I agree that it is somehow confusing.

sauwming (approved):

trengginas (approved):

PR #3008, opened 2022-03-07
Deprecate PJ_ENABLE_EXTRA_CHECK macro

trengginas:
This ticket will deprecate `PJ_ENABLE_EXTRA_CHECK` macro. Initially, it is intended to reduce size on assertion checks.

### DIFF
--- a/pjlib/include/pj/assert.h
+++ b/pjlib/include/pj/assert.h
@@ -55,21 +55,29 @@
  * If the expression yields false, assertion will be triggered
  * and the current function will return with the specified return value.
  */
+// #if defined(PJ_ENABLE_EXTRA_CHECK) && PJ_ENABLE_EXTRA_CHECK != 0
 #define PJ_ASSERT_RETURN(expr,retval)    \
 	    do { \
 		if (!(expr)) { pj_assert(expr); return retval; } \
 	    } while (0)
+//#else
+//#   define PJ_ASSERT_RETURN(expr,retval)    pj_assert(expr)
+//#endif
 
 /**
  * @hideinitializer
  * If the expression yields false, assertion will be triggered
  * and @a exec_on_fail will be executed.
  */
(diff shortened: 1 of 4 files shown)


pjlib/include/pj/assert.h
```diff
@@ -52,42 +52,24 @@
...
- * If #PJ_ENABLE_EXTRA_CHECK is declared and the value is non-zero, then 
- * #PJ_ASSERT_RETURN macro will evaluate the expression in @a expr during
- * run-time. If the expression yields false, assertion will be triggered
+ * If the expression yields false, assertion will be triggered
  * and the current function will return with the specified return value.
- *
- * If #PJ_ENABLE_EXTRA_CHECK is not declared or is zero, then no run-time
- * checking will be performed. The macro simply evaluates to pj_assert(expr).
  */
-#if defined(PJ_ENABLE_EXTRA_CHECK) && PJ_ENABLE_EXTRA_CHECK != 0
```
sauwming, line 63: For deprecation, we normally comment it first instead of directly removing it.
```
/* PJ_ENABLE EXTRA_CHECK is deprecated in PJSIP 2.13. */
// #if defined(PJ_ENABLE_EXTRA_CHECK) && PJ_ENABLE_EXTRA_CHECK != 0
```

pjlib/include/pj/config.h
```diff
@@ -933,6 +916,20 @@
 #   define PJ_HAS_STRICMP_ALNUM	    0
 #endif
 
+/*
+ * Warn about obsolete macros.
+ *
+ * PJ_ENABLE_EXTRA_CHECK has been deprecated in 2.13.
+ */
+#if defined(PJ_ENABLE_EXTRA_CHECK)
```
sauwming, line 924: If user explicitly enable it, I think we don't need to give any warning to minimise any inconvenience (user may need to remove the setting and rebuild the entire library). The behavior doesn't change anyway.
So we should only give warning if PJ_ENABLE_EXTRA_CHECK==0.

sauwming (approved):

PR #3009, opened 2022-03-07
Use PJ_ASSERT_RETURN() on pjsip_auth_create_digest() and pjsua_init_tpselector()

trengginas:
This ticket will use `PJ_ASSERT_RETURN()` instead of `pj_assert()` on `pjsip_auth_create_digest()` and `pjsua_init_tpselector()`. 
In case assertion is disabled (e.g: release mode build) then the check will not be skipped.

### DIFF
--- a/pjsip/src/pjsip/sip_auth_client.c
+++ b/pjsip/src/pjsip/sip_auth_client.c
@@ -175,7 +175,7 @@ PJ_DEF(void) pjsip_auth_create_digest( pj_str_t *result,
     unsigned char digest[16];
     pj_md5_context pms;
 
-    PJ_ASSERT_RETURN(result->slen >= PJSIP_MD5STRLEN, );
+    pj_assert(result->slen >= PJSIP_MD5STRLEN);
 
     AUTH_TRACE_((THIS_FILE, "Begin creating digest"));
 
@@ -194,7 +194,7 @@ PJ_DEF(void) pjsip_auth_create_digest( pj_str_t *result,
 	digestNtoStr(digest, 16, ha1);
 
     } else if ((cred_info->data_type & PASSWD_MASK) == PJSIP_CRED_DATA_DIGEST) {
-	pj_assert(cred_info->data.slen == 32);
+	PJ_ASSERT_RETURN(cred_info->data.slen == 32, );
 	pj_memcpy( ha1, cred_info->data.ptr, cred_info->data.slen );
     } else {
 	pj_assert(!"Invalid data_type");
--- a/pjsip/src/pjsip/sip_auth_client.c
+++ b/pjsip/src/pjsip/sip_auth_client.c
@@ -160,15 +160,15 @@ static void digestNtoStr(const unsigned char digest[], int n, char *output)
  * Create response digest based on the parameters and store the
  * digest ASCII in 'result'.
  */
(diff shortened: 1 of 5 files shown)


pjsip/src/pjsip/sip_auth_client.c
```diff
@@ -175,7 +175,7 @@ PJ_DEF(void) pjsip_auth_create_digest( pj_str_t *result,
     unsigned char digest[16];
     pj_md5_context pms;
 
-    pj_assert(result->slen >= PJSIP_MD5STRLEN);
+    PJ_ASSERT_RETURN(result->slen >= PJSIP_MD5STRLEN, );
```
sauwming, line 178: The issue is at line 197, not here.

sauwming (approved):

pjsip/src/pjsip/sip_auth_client.c
```diff
@@ -194,7 +194,11 @@ PJ_DEF(void) pjsip_auth_create_digest( pj_str_t *result,
 	digestNtoStr(digest, 16, ha1);
 
     } else if ((cred_info->data_type & PASSWD_MASK) == PJSIP_CRED_DATA_DIGEST) {
-	PJ_ASSERT_RETURN(cred_info->data.slen == 32, );
+	if (cred_info->data.slen != 32) {
+	    pj_bzero(result->ptr, result->slen);
+	    result->slen = 0;
+	}
+	PJ_ASSERT_RETURN(cred_info->data.slen == 32, PJ_EINVAL);
```
nanangizz, line 201: minor, IMHO instead of using `PJ_ASSERT_RETURN()` outside the `if` block, perhaps it is easier to read to just use `pj_assert()` & `return EINVAL` inside the block? also a bit more efficient as the check is done once.

pjsip/src/pjsip/sip_auth_client.c
```diff
@@ -194,7 +194,11 @@ PJ_DEF(void) pjsip_auth_create_digest( pj_str_t *result,
...
 	pj_memcpy( ha1, cred_info->data.ptr, cred_info->data.slen );
     } else {
 	pj_assert(!"Invalid data_type");
```
nanangizz, line 204: some error handling too here?

pjsip/src/pjsip/sip_auth_client.c
```diff
@@ -291,7 +296,11 @@ PJ_DEF(void) pjsip_auth_create_digestSHA256(pj_str_t *result,
 
     } else if ((cred_info->data_type & PASSWD_MASK) == PJSIP_CRED_DATA_DIGEST)
     {
-	pj_assert(cred_info->data.slen == 32);
+	if (cred_info->data.slen != 32) {
+	    pj_bzero(result->ptr, result->slen);
+	    result->slen = 0;
+	}
+	PJ_ASSERT_RETURN(cred_info->data.slen == 32, PJ_EINVAL);
```
nanangizz, line 303: same as above.
also perhaps the len of `data/ha1` should be 64 or `PJSIP_SHA256STRLEN` instead of 32 (need to check the standard)?

nanangizz (approved):

risicle:
I'm having a hard time figuring out whether `pjsip_auth_create_digest` is part of the public API and therefore whether applying this patch would be API/ABI-breaking.

Issue #2992, opened 2022-02-22
PJSUA - call will not be hung up if hangup() is issued too early


andreas-wehrmann:
### Describe the bug

I just noticed a rather serious problem, unfortunately I currently don't have time today to dissect this further,
so I'll just describe what I've found together with the logs I gathered.

I confirmed this problem with current pjsip master @ 88b9126 and the PJSUA-APP
though I observed this problem originally with pjproject 2.11.

**What's the problem?**

Hanging up an outgoing call before "100 - Trying" is received can cause PJSUA to never hang up the call afterwards.
The application will be signalled that the call has been terminated when in fact the underlying dialog is still there and still occupies a call slot.
If this problem repeats often enough, all call slots will eventually be taken by by these... let's call them "ghost calls" and prevent any further calls.

**What's the scenario?**

- Make call to a device which always immediately accepts SIP Calls (always sends 200-OK after 100-Trying)
- Hang up the call immediately after initiating the call.

_Please note:_
In the example trace I attached to this issue,
the remote device sends a 180-Ringing also, but where I observed the problem,
the call was accepted immediately.

**What happens?**

- PJSUA sends INVITE
- User issues hangup
- PJSUA immediately reports to the application: call terminated (at this point, the call is gone as far as the app is concerned)
- PJSUA logs: "Delaying CANCEL since no provisional response is received yet"
- Peer responds with "100 - Trying" and immediately after sends "200 - OK"
- PJSUA receives "100 - Trying" and sends the CANCEL request (before it receives/handles the 200-OK).
- PJSUA receives the "200-OK" and sends an ACK, the dialog is established at this point
- Peer responds with "200-OK" to the CANCEL request
- PJSUA received the "200-OK" to the CANCEL request, but nothing happens, the call will stay active

Please note:
I have observed cases where the peer responds with an error code 4xx to the CANCEL request in a situation like this,
the behaviour is the same though: nothing happens, the call will stay up.

**What do I expect to happen?**

I a situation like this, where INVITE and CANCEL overlap,
I expect the call to be fully established shortly but then be immediately taken down with BYE;
I believe the BYE should be send regardless of whether a response to the CANCEL request has been received or not
(after the dialog has been established) though I am currently unsure about what the RFC says.

Please see the attached logs.
In the PJSUA APP I limited max-calls to just one.
When I make a call and immediately issue a hangup, the problem occurs.
[...]

PR #3010, opened 2022-03-08
Add compile time option to allow multiple Authorization header

trengginas:
On #2887, the library will only add the top most supported auth header when receiving a 401/407 response with multiple WWW-Authenticate/Proxy-Authenticate headers. This ticket will add compile time setting `PJSIP_AUTH_ALLOW_MULTIPLE_AUTH_HEADER` to allow multiple Authorization header to be added.

### DIFF
--- a/tests/pjsua/scripts-recvfrom/215_reg_good_multi_ok.py
+++ b/tests/pjsua/scripts-recvfrom/215_reg_good_multi_ok.py
@@ -31,10 +31,6 @@ req3 = sip.RecvfromTransaction("Registration retry with auth (allowed multiple a
 					 "realm=\"python1\"", "realm=\"python2\"", 
 					 "username=\"theuser1\"", "username=\"theuser2\"", 
 					 "nonce=\"1234\"", "nonce=\"6789\"", 
-					 "Authorization:", # [\\s\\S]+Authorization:"
-					 "realm=\"python1\"", # "realm=\"python2\"", 
-					 "username=\"theuser1\"", # "username=\"theuser2\"", 
-					 "nonce=\"1234\"", # "nonce=\"6789\"", 
 					 "response="],
 				expect="registration success",
 				pj_config="PJSIP_AUTH_ALLOW_MULTIPLE_AUTH_HEADER.*: 1"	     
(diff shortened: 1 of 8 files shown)


sauwming:
Need to modify the PJSIP test as well (see #2992,) so that if the setting is enabled, it will pass the test.

trengginas:
> Need to modify the PJSIP test as well (see #2992,) so that if the setting is enabled, it will pass the test.

It should be #2922? Anyway, does this mean the test should know the compile time settings?

sauwming:
Yes, you're right, sorry about the wrong PR no.

For certain settings, I think it should. For example:
https://github.com/pjsip/pjproject/blob/master/pjsip/src/test/msg_test.c#L1896
And for this case, if the user decides to allow multiple auth, we don't want PJSIP test to fail on their machine, right?

tests/pjsua/scripts-recvfrom/215_reg_good_multi_ok.py
```diff
@@ -14,16 +14,31 @@
...
+				pj_config="PJSIP_AUTH_ALLOW_MULTIPLE_AUTH_HEADER.*: 0"
 			  )
 
+req3 = sip.RecvfromTransaction("Registration retry with auth (allowed multiple auth)", 200,
+				include=["REGISTER sip", 
+					 "Authorization:[\\s\\S]+Authorization:", # Must have 2 Auth hdrs
+					 "realm=\"python1\"", "realm=\"python2\"", 
+					 "username=\"theuser1\"", "username=\"theuser2\"", 
+					 "nonce=\"1234\"", "nonce=\"6789\"", 
+					 "Authorization:", # [\\s\\S]+Authorization:"
```
sauwming, line 34: remove this block (until nonce below)

sauwming (commented):
I was actually expecting something simpler, i.e. as long as the test doesn't fail if the config is enabled, it should be fine (but admittedly, I didn't think too much on how to do it :) )

Should be okay, just minor below.

sauwming (approved):

nanangizz (approved):

PR #3011, opened 2022-03-08
minor output changes: keystroke_help

liviu-rolo:
commands description: some rogue period and capital letters


pjsip-apps/src/pjsua/pjsua_app_legacy.c
```diff
@@ -244,7 +244,7 @@ static void keystroke_help()
     puts("|  S  Send arbitrary REQUEST   | Cp  Codec priorities     |                   |");
     puts("+-----------------------------------------------------------------------------+");
 #if PJSUA_HAS_VIDEO
-    puts("| Video: \"vid help\" for more info                                             |");
+    puts("| Video: \"vid help\" for more info                                           |");
```
sauwming, line 247: the current code is already correct, there are two escape characters (for `\"`), so there will be two extra spaces at the end.

sauwming (approved):

nanangizz (approved):

PR #3017, opened 2022-03-10
allow pj to build for arm64 under clang when not using autotools

ddurham2:
honoring __aarch64__ (clang's macro) as being equivalent to _M_ARM64 (MSVC's macro) to properly detect arm64 when compiling under clang (not using autotools).  This is necessary for multiarch builds (e.g. Apple Silicon).

### DIFF
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -255,7 +255,7 @@
 #   define PJ_IS_BIG_ENDIAN	1
 
 #elif defined(ARM) || defined(_ARM_) ||  defined(__arm__) || \
-       defined(_M_ARM) || defined(_M_ARM64)
+       defined(_M_ARM) || defined(_M_ARM64) || defined(__aarch64__)
 #   define PJ_HAS_PENTIUM	0
     /*
      * ARM, bi-endian, so raise error if endianness is not configured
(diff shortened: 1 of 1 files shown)


sauwming (approved):

sauwming:
Right, it's been used in the third_party sources of `webrtc_aec3`, as well:
```
./third_party/webrtc_aec3/src/rtc_base/system/arch.h:#elif defined(_M_ARM64) || defined(__aarch64__)
./third_party/webrtc_aec3/src/third_party/pffft/src/pffft.c:#elif !defined(PFFFT_SIMD_DISABLE) && (defined(__arm__) || defined(__aarch64__) || defined(__arm64__) || defined(_M_ARM64))
./third_party/webrtc/src/webrtc/typedefs.h:#elif defined(__aarch64__) || defined(_M_ARM64)
```

ddurham2:
@sauwming, do I need to do anything about the failures above?  I appears to be platform-specific issues.  I'm not sure I know how to address those.

sauwming:
no, no, those are not failures. I am just pointing out that the same check is used in other places as well.

So nothing to do on your part for now while awaiting for other reviews.

trengginas (approved):

PR #3018, opened 2022-03-14
Always re-arm ioqueue for events on pj_ioqueue_poll() when epoll with EPOLLONESHOT is used

trengginas:
When epoll with EPOLLONESHOT is used, re-arming ioqueue is required regardless there is no events to be processed.
Otherwise, subsequent events will be ignored by `pj_ioqueue_poll()`.

Thank you to @der-ambi for the report and feedback.


sauwming:
pjlib-test will fail and stall here without the patch:
`04:30:47.788 ...ioqueue unregister stress test 2/3, unregister in cb (epoll-oneshot)`

sauwming (approved):

nanangizz (approved):

Issue #3012, opened 2022-03-10
Registration issues with IP change scenario IPv4<->IPv6.


sauwming:
There are a few issues with registration when calling `pjsua_handle_ip_change()` for a network change scenario to a different address type (please refer to the [IP change wiki](https://trac.pjsip.org/repos/wiki/IPAddressChange-2.7#NetworkchangetoadifferentIPaddresstype.IPv4toIPv6orIPv6toIPv4)):
1. Registration triggered by IP change does not use the newly set transport
Fixed in #3004.

2. IP change will fail if there is a registration in progress
Due to: https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_acc.c#L4143
Fixed in #3021 .

3. IP change will make a registration in progress fail (with 503).
The issue seems specific for IP change scenario, since ordinary registration race scenario was tested to work fine and should have been fixed by #2653.

4. Modifying account followed by calling `pjsua_handle_ip_change()` will trigger double registration attempts.
This shouldn't be a problem if 2 has been solved, but this may look inefficient (and annoying) from the user's perspective, since why send a registration that's going to fail anyway.
There are a couple of possible solutions:
a. Create a new API `pjsua_acc_modify2()` with option not to send registration.
b. Account modification is included in the IP change process. i.e. IP change param is added a new field `new_acc_cfg`. So user shouldn't need to call `pjsua_acc_modify()` and let `pjsua_handle_ip_change()` modify the account internally (while also avoiding reregistration of course).
Option b seems better.

sauwming:
1-3 completed. 4 optional if 2 has been solved.

PR #3021, opened 2022-03-15
Handle PJSIP_EBUSY error when sending Registration after calling pjsua_handle_ip_change()

trengginas:
Calling `pjsua_handle_ip_change()` might send Registration when updating account contact. It will fail (PJ_EBUSY) when there's already an outgoing Registration. This patch will handle such issue by prioritizing the Registration sent by `pjsua_handle_ip_change()`.


pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -2357,7 +2357,7 @@ static void handle_call_on_ip_change_cb(void *user_data)
  */
 static void regc_cb(struct pjsip_regc_cbparam *param)
```
sauwming, line 2358: I don't quite understand the purpose of the modification of `regc_cb()` and `schedule_reregistration()`. Is this to fix #3012 item (3)?

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -4147,24 +4149,43 @@ pj_status_t pjsua_acc_update_contact_on_ip_change(pjsua_acc *acc)
...
-    if ((status != PJ_SUCCESS) && (pjsua_var.ua_cfg.cb.on_ip_change_progress)
+    status = pjsua_acc_set_registration(acc->index, !need_unreg);    
+    if ((status != PJ_SUCCESS)
 	&& (acc->ip_change_op == PJSUA_IP_CHANGE_OP_ACC_UPDATE_CONTACT))
     {
-	/* If update contact fails, notification might already been triggered
-	 * from registration callback.
-	 */
-	pjsua_ip_change_op_info info;
+	if (status != PJ_EBUSY) {
```
sauwming, line 4156: I don't understand this either. The title of the PR is to handle if the registration returns busy, but why is the retry done if the status is not busy?

pjsip/src/pjsua-lib/pjsua_acc.c
trengginas: It's for a different issue indeed. Moved the changes to #3027

sauwming: Separating or combining are actually both okay. Most important is to describe it in the PR so users & reviewers know exactly what the PR contains.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -2567,16 +2567,15 @@ static void regc_cb(struct pjsip_regc_cbparam *param)
 		if ((status != PJ_SUCCESS) &&
 		    pjsua_var.ua_cfg.cb.on_ip_change_progress)
 		{
-		    pjsua_ip_change_op_info ip_chg_info;
+			pjsua_ip_change_op_info ip_chg_info;
```
sauwming, line 2570: revert these unnecessary white space changes, to avoid conflict with #3027 once merged.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -4147,24 +4147,43 @@ pj_status_t pjsua_acc_update_contact_on_ip_change(pjsua_acc *acc)
...
-	pjsua_ip_change_op_info info;
+	if (status == PJSIP_EBUSY) {
+	    /* Retry registration by cancelling the existing one. */
+	    if (acc->regc) {
+		pjsip_regc_destroy(acc->regc);
+		acc->regc = NULL;
+                acc->contact.slen = 0;
+                acc->reg_mapped_addr.slen = 0;
+                acc->rfc5626_status = OUTBOUND_UNKNOWN;
+                acc->rfc5626_flowtmr = 0;
```
sauwming, line 4162: not quite related with this PR. but perhaps this is time to put this in a separate helper function such as `destroy_regc()` since it has been used throughout pjsua_acc. And I notice a couple of missing of resetting `acc->rfc5626_flowtmr`, which shouldn't happen if we use the helper function.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -218,6 +218,18 @@ static void init_outbound_setting(pjsua_acc *acc)
     acc->rfc5626_status = OUTBOUND_WANTED;
 }
 
+/*
+ * Helper function to deinit registration info on account.
+ */
+static void regc_destroy(pjsua_acc *acc)
```
sauwming, line 224: since the name is regc_destroy(), put pjsip_regc_destroy() inside the function as well.

sauwming (approved):
Please do a final test:
* Make sure that the old registration is properly cancelled and reg callbacks are no longer called.
* Make sure that the new registration correctly uses the new details (for example, the newly set IPv6 transport for IPv4->IPv6 IP change scenario).

nanangizz:
if it is possible (or not so complex to implement), keep a note on the existing contacts from the destroyed regc so they can be unregistered (while registering the new contact). Perhaps this is actually the current behaviour of IP change handling, so this is just to maintain the behavior.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -4195,34 +4195,40 @@ pj_status_t pjsua_acc_update_contact_on_ip_change(pjsua_acc *acc)
...
-	pjsua_acc_end_ip_change(acc);
+        if (status == PJSIP_EBUSY) {
+            /* Retry the (un)registration. */
+            if (acc->regc) {
+                status = PJ_SUCCESS;
+                destroy_regc(acc, PJ_TRUE);
+                update_keep_alive(acc, PJ_FALSE, NULL);
+
+                if (need_unreg)
+                    status = pjsua_regc_init(acc->index);
```
nanangizz, line 4206: If the regc is destroyed in `destroy_regc()` (where the contact.slen seems to be reset), will the retry try to unregister the new contact instead of the old contact?

sauwming: handle this in the last commit

nanangizz (approved):

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -4196,15 +4198,24 @@ pj_status_t pjsua_acc_update_contact_on_ip_change(pjsua_acc *acc)
...
+                pj_str_t old_reg_contact = acc->reg_contact;
                 status = PJ_SUCCESS;
                 destroy_regc(acc, PJ_TRUE);
                 update_keep_alive(acc, PJ_FALSE, NULL);
 
-                if (need_unreg)
-                    status = pjsua_regc_init(acc->index);
-                
+                status = pjsua_regc_init(acc->index);
+                if (need_unreg || no_unreg)
```
trengginas, line 4214: Why `no_unreg` check is needed here? The check below should cover it?

sauwming: if `no_unreg`, the Contact update will be done in a single registration message, by removing the old Contact binding, and adding a new Contact binding.

`pjsua_regc_init()` will initialise the regc with the new contact.
So I need to modify it by adding the old Contact first, then adding the new Contact.

sauwming (approved):

Issue #2999, opened 2022-02-28
PJSUA2: Structs as aggreates


wosrediinanatour:
If you want to use user-defined headers you need following code:
```cpp
SipHeader h;
h.hName = "X-Header";
h.hValue = "User header";
SipHeaderVector headers;
headers.push_back(h);
```

In modern C++ [following](https://en.cppreference.com/w/cpp/language/aggregate_initialization) would be possible, if `SipHeader` was an aggregate:
```cpp
SipHeaderVector headers{{"X-Header", "User header"}};
// e.g.
SipHeaderVector headers{{"X-Header", "User header"}, {"X-Header", "User header2"}, {"X-Header", "User header3"}};
```
and since C++20 even:
```cpp
SipHeaderVector headers{{.hName="X-Header", .hValue="User header"}};
```

More general: In PJSUA2 structs are most often non-aggregates, e.g. methods `toPj()` and `fromPj()` are private. Sometimes there are others. In my opinion, moving them out (from methods to function) would make much cleaner/nicer code possible in the applications.

PS (*update*): Currently, an application can use the following to at least bypass one step:
```cpp
45  struct header : public pj::SipHeader
  1 {
  2    header(std::string&& name, std::string&& value)
  3    {
  4       SipHeader::hName = std::move(name);
  5       hValue = std::move(value);
  6    };
  7 };
```
but it is a bad hack, since the solution shouldn't be to add new constructors.

sauwming:
In some structs, fromPj() and toPJ() are actually public, so I believe they should be okay here. Besides, private member functions should be okay too, shouldn't it?

But the main obstacle IMO is the virtual member functions, especially for structs that inherit virtual base class `PersistentObject`. This can present major backward compatibility.

PR #3024, opened 2022-03-16
Make sure AccountMediaConfig params are set to false in constructor

sean-horton:
Hi,

After upgrading to pjsip 2.12 I noticed occasional call failures. In the case of call failure it was noticed the call sdp shows the client stating its IP address as 127.0.0.1. After a discussion with @nanangizz over email it was mentioned the client seemed to be using loop transport. I was able to verify the client was using loop transport, but only sometimes. Looking further, `AccountMediaConfig#setUseLoopMedTp(...)`  was recently added in #2900 

This proposed change makes sure the two parameters `useLoopMedTp` and `enableLoopback` are set to `false` in the `AccountMediaConfig` constructor. I am seeing expected behavior with this change.

This brings the question, should other fields in `AccountMediaConfig` have default values set in the constructor like `rtcpMuxEnabled`, `srtpSecureSignaling`, etc?

### DIFF
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -1046,7 +1046,10 @@ public:
      */
     AccountMediaConfig() : srtpUse(PJSUA_DEFAULT_USE_SRTP),
 			   ipv6Use(PJSUA_IPV6_DISABLED)
-    {}
+    {
+        useLoopMedTp = false;
+        enableLoopback = false;
+    }
 
     /**
      * Read this object from a container node.


sauwming:
Currently, the proper initialisations of `Account*Config` are done only via `AccountConfig`.
```
AccountConfig::AccountConfig()
{
    pjsua_acc_config acc_cfg;
    pjsua_acc_config_default(&acc_cfg);
    pjsua_media_config med_cfg;
    pjsua_media_config_default(&med_cfg);
    fromPj(acc_cfg, &med_cfg);
}
```

So this is okay:
```
AccountConfig acc_cfg;
AccountMediaConfig med_cfg = acc_cfg.mediaConfig;
```

While this is not:
```
AccountMediaConfig med_cfg;
AccountConfig acc_cfg;
acc_cfg.mediaConfig = med_cfg;
```

wosrediinanatour:
I think that revising the PJSUA2's types make sense for consistence, safety, initialization, and usability, e.g. https://github.com/pjsip/pjproject/issues/2999, https://github.com/pjsip/pjproject/issues/3039. As a first step, using aggregates would make sense, which should be possible for all C++ standards.  Using feature macros (https://en.cppreference.com/w/cpp/feature_test), their usage can be even simplified.

PR #3027, opened 2022-03-21
Don't schedule re-Registration on IP change case

trengginas:
The call to `pjsua_handle_ip_change()` might trigger Registration to update Contact.
On some cases when it fails, the library will schedule a retry. 
Currently, when this happen the the library will end the IP change progress ignoring the retry.
The reschedule will use `pjsua_acc_config::reg_retry_interval` (default: 5minutes) as the interval, which is not desirable for IP change case.
The patch will disable scheduling re-Registration on IP change and let application retry manually.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -4038,8 +4038,6 @@ static void schedule_reregistration(pjsua_acc *acc)
     acc->auto_rereg.timer.id = PJ_TRUE;
     if (pjsua_schedule_timer(&acc->auto_rereg.timer, &delay) != PJ_SUCCESS)
 	acc->auto_rereg.timer.id = PJ_FALSE;
-
-    return;
 }
 
 
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -33,7 +33,7 @@ enum
 
 
 static int get_ip_addr_ver(const pj_str_t *host);
-static pj_bool_t schedule_reregistration(pjsua_acc *acc);
+static void schedule_reregistration(pjsua_acc *acc);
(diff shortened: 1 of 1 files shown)


sauwming:
The idea is good but currently the default retry interval is 5 minutes, so I'm not sure if user wants to wait that long for the IP change to continue.

sauwming:
The idea is good but currently the default retry interval is 5 minutes, so I'm not sure if user wants to wait that long for the IP change to continue.

trengginas:
> The idea is good but currently the default retry interval is 5 minutes, so I'm not sure if user wants to wait that long for the IP change to continue.

The alternatives:
1. bypass the re-Register schedule and end the IP change
2. use a different interval specific for IP change

Prefer option 1, so that application can retry manually.

sauwming:
Option 1 looks okay for now, it's simpler.

Later, if there's any request, we can implement option 2, i.e.:
- Add `pjsua_ip_change_acc_cfg.reg_retry_interval` (default `PJSUA_IP_CHANGE_REG_RETRY_INTERVAL`, around 5-10 sec should bed fine IMO. If this is zero, no rereg retry and IP change immediately ended (option 1)).
- Also need `reg_retry_attempts` (default: 1(?), number of retries before IP change is ended with failure, only applicable if the above setting not zero).

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -4038,6 +4038,8 @@ static void schedule_reregistration(pjsua_acc *acc)
     acc->auto_rereg.timer.id = PJ_TRUE;
     if (pjsua_schedule_timer(&acc->auto_rereg.timer, &delay) != PJ_SUCCESS)
 	acc->auto_rereg.timer.id = PJ_FALSE;
+
+    return;
```
sauwming, line 4042: unnecessary return

sauwming (approved):
also change the PR desc.

nanangizz (approved):

Issue #3028, opened 2022-03-21
Add  kqueue  support for macos/BSD


jimying:
I want to add ioqueue framework for macos/ BSD  with kqueue.

BSD kqueue is simillar with linux epoll.

```bash
# how to enable kqueue
./configure --enable-kqueue
```

PR #3029, opened 2022-03-21
add kqueue support for macos/BSD

jimying:
as title,  please see  #3028 
Add new ioqueue framework by kqueue on macos/BSD

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -823,6 +823,7 @@ ac_user_opts='
 enable_option_checking
 enable_floating_point
 enable_epoll
+enable_kqueue
 enable_shared
 enable_pjsua2
 with_external_speex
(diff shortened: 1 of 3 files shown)


sauwming:
Looks good. I tested it here on MacOS and it passes pjlib test and simple call test.

One minor issue: The configure script will output `select` first before `kqueue`:
```
checking ioqueue backend... select()
kqueue
```
You can also remove `ac_linux_poll` since it's only used by `os-linux.mak`, which is no longer used.

jimying:
@sauwming update aconfigure.ac/aconfigure

sauwming (approved):

nanangizz (approved):

PR #3037, opened 2022-03-25
Fixed missing field in CodecParam

sauwming:
Add missing field in pjsua2 `CodecParam` for `pjmedia_codec_param.info.enc_ptime`. This missing field can cause undefined behavior when calling `Endpoint::codecSetParam()` (since the field can be any value), which can later trigger various assertions such as:
`../src/pjmedia/stream.c:1276: void rebuffer(pjmedia_stream *, pjmedia_frame *): assertion "stream->enc_buf_count + (frame->size >> 1) < stream->enc_buf_size" failed`
or
`'../src/pjmedia-codec/g722.c:560: pj_status_t g722_codec_encode(pjmedia_codec *, const struct pjmedia_frame *, unsigned int, struct pjmedia_frame *): assertion "(input->size >> 2) <= output_buf_len" failed'`

### DIFF
--- a/pjsip/include/pjsua2/media.hpp
+++ b/pjsip/include/pjsua2/media.hpp
@@ -2431,6 +2431,8 @@ struct CodecParamInfo
     unsigned	maxBps;			/**< Maximum bandwidth in bits/sec  */
     unsigned    maxRxFrameSize;		/**< Maximum frame size             */
     unsigned 	frameLen;		/**< Decoder frame ptime in msec.   */
+    unsigned 	encFrameLen;		/**< Encoder ptime, or zero if it's
+					     equal to decoder ptime.	    */
     unsigned  	pcmBitsPerSample;	/**< Bits/sample in the PCM side    */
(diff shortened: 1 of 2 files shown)


nanangizz (approved):
Btw, OOT, IIRC the `enc_ptime` parameter was about to be deprecated, because it is only used by iLBC codec to allow encoding & decoding uses different frame length (it has 20s & 30ms modes), however the [RFC](https://datatracker.ietf.org/doc/html/rfc3952#section-5) mandates to use the same mode in both sides.

sauwming:
Ah okay, didn't know that.

Currently it's still used in stream, such as:
```
    /* How many samples are needed */
    count = stream->codec_param.info.enc_ptime *
	    PJMEDIA_PIA_SRATE(&stream->port.info) / 1000;
```
which causes all sorts of problems. So this will need to be changed if we decide to deprecate the field in the future.

trengginas (approved):

PR #3038, opened 2022-03-25
Added conference port id checking in pjsua

sauwming:
`pjsua_conf_port_id` is of type signed integer:
```
/** Conference port identification */
typedef int pjsua_conf_port_id;
```
so that it can contain the value of `PJSUA_INVALID_ID` (-1).

But in `pjsua_conf` and `pjsua_vid_conf` APIs there's no checking whatsoever and we immediately pass it to the PJMEDIA audio/video conference APIs that actually expects the port id parameters to be of type `unsigned`. Thus, app that passes invalid id port can cause out-of-bound access.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_aud.c
+++ b/pjsip/src/pjsua-lib/pjsua_aud.c
@@ -853,6 +853,8 @@ PJ_DEF(pj_status_t) pjsua_conf_get_port_info( pjsua_conf_port_id id,
     unsigned i;
     pj_status_t status;
 
+    PJ_ASSERT_RETURN(id >= 0, PJ_EINVAL);
+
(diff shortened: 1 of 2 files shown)


trengginas (approved):

wosrediinanatour:
Wouldn't it possible to change the type, i.e. use unsigned int?

sauwming:
> Wouldn't it possible to change the type, i.e. use unsigned int?

Yes, it's possible, but the modification will have a greater impact. At the very least, application will get a compiler warning for this: `if (port_id == PJSUA_INVALID_ID)`, which can be an immediate failure for stricter warning options (such as `-Werror`).

nanangizz (approved):

Issue #2167, opened 2018-12-06
Perform RTP fuzzing


pjsipbot:
_2018-12-06 02:52:26: @nanangizz_ _created the issue on trac ticket [2167](https://trac.pjsip.org/repos/ticket/2167)_

As we have never done this, perhaps it is good idea to start trying it, e.g: simply sending random byte stream to RTP port, or use existing tool (e.g: [ohrwurm](https://packetstormsecurity.com/fuzzer/ohrwurm-0.1.tar.bz2)).

PR #3045, opened 2022-04-04
Add fuzz harness

Cossack9989:
Fuzzing towards msg/sdp/dns may make pjsip more secure.


sauwming:
Thank you for your contribution. Fuzzer is definitely an important tool that we have been wanting to include (since #2167).

We were recently introduced to Google's [oss-fuzz](https://github.com/google/oss-fuzz) and we thought that we'd use it due to its comprehensive coverage.

So this will be covered by #3297 instead.

PR #3059, opened 2022-04-11
Some callbacks not invoked in call replace in PJSUA2

nanangizz:
Currently in call replace scenario, the new PJSUA2 Call instance can only be created in `onCallReplaced()` callback, which cause application to miss some callbacks such as `onCallSdpCreated()` and `onCreateMediaTransport()`. This PR allows application to create the new Call instance earlier, i.e: in `onCallReplaceRequest()`, so it won't miss the `onCallSdpCreated()` & `onCreateMediaTransport()` callbacks.

Thanks to Johannes Westhuis for the report.

Note: this PR also bumps up log level of some warnings from 4 to 3 (when app does not create a new Call instance for call transfer & replace).

### DIFF
--- a/pjsip-apps/src/samples/pjsua2_demo.cpp
+++ b/pjsip-apps/src/samples/pjsua2_demo.cpp
@@ -72,7 +72,7 @@ public:
     
     virtual void onCallState(OnCallStateParam &prm);
     virtual void onCallTransferRequest(OnCallTransferRequestParam &prm);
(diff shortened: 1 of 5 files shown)


sauwming:
Better to change the sample pjsua2_demo to use the new callback onCallReplaceRequest() instead.

sauwming (approved):

trengginas (approved):

PR #3065, opened 2022-04-14
Handle crash when converting std::string with non UTF-8 character on Android

trengginas:
PJSUA2 (C++) API uses std::string to store string information. If the string contains non UTF-8 character, then on Android it will lead to a crash.
```
JNI WARNING: NewStringUTF input is not valid Modified UTF-8
```
The swig will use `NewStringUTF()` JNI call to convert the std::string, and it will crash if it contains non UTF-8 characters.

Reference:
- https://github.com/swig/swig/issues/1174

### DIFF
--- a/pjsip-apps/src/swig/and_string.i
+++ b/pjsip-apps/src/swig/and_string.i
@@ -1,0 +1,57 @@
+#if defined(SWIGJAVA) && defined(__ANDROID__)
+%typemap(out) std::string
+%{
+    jsize $1_len = (jsize)$1.length();
+    jbyteArray bytes = jenv->NewByteArray($1_len);
+    jenv->SetByteArrayRegion(bytes, 0, $1_len, (jbyte *) $1.c_str());
+    jclass strClass = jenv->FindClass("java/lang/String");
+    jmethodID ctorID = jenv->GetMethodID(strClass, "<init>", "([BLjava/lang/String;)V");
+    jstring encoding = jenv->NewStringUTF("UTF-8");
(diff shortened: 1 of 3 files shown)


nanangizz:
Btw, perhaps the patch should be applied for all platforms instead of only Android, because `NewStringUTF()` & `GetStringUTFChars()` that are using modified UTF-8 seems to happen on all platforms?

sauwming:
> Btw, perhaps the patch should be applied for all platforms instead of only Android, because `NewStringUTF()` & `GetStringUTFChars()` that are using modified UTF-8 seems to happen on all platforms?

When I tested it on Mac, it worked without any obvious issue. But after reading the doc again, it seems you're right, the string conversion issue happens on all platforms, it's just that it doesn't always abort. It only aborts if `CheckJNI` is available and enabled.
`Android also offers a mode called CheckJNI... CheckJNI—which is on by default for emulators—scans strings and aborts the VM if it receives invalid input.` (https://developer.android.com/training/articles/perf-jni).
```There are several ways to enable CheckJNI. If you're using the emulator, CheckJNI is on by default. You can also set the android:debuggable attribute in your application's manifest to turn on CheckJNI for your app.```

So yeah, the patch can apply to all platforms, but I believe the crash/abort issue is Android specific.

trengginas:
I think we should limit this for android, since there's no report for other platform.

nanangizz (approved):

sauwming (approved):

Issue #3069, opened 2022-04-14
Double free of ossock->ossl_ctx in case of errors.


johado:
### Describe the bug

In pjlib/src/pj/ssl_sock_ossl.c:init_ossl_ctx()
ossock->ossl_ctx = ctx = SSL_CTX_new(ssl_method);
is used, but in the error cases only 
 SSL_CTX_free(ctx); 
 is called, but ossock->ossl_ctx is not NULL:ed so in ssl_destroy() the context will be double freed.

### Steps to reproduce

Try to specify a cert path that does not exist.

### PJSIP version

2.12, 2.11.1

### Context

When using openssl with some settings that causes errors.

### Log, call stack, etc

```shell
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -1174,6 +1174,7 @@ static pj_status_t init_ossl_ctx(pj_ssl_sock_t *ssock)
                                 cert->CA_path.ptr));
                }
                SSL_CTX_free(ctx);
+               ossock->ossl_ctx = NULL;
                return status;
            } else {
                PJ_LOG(4,(ssock->pool->obj_name,
@@ -1204,6 +1205,7 @@ static pj_status_t init_ossl_ctx(pj_ssl_sock_t *ssock)
                             "Error loading certificate chain file '%s'",
                             cert->cert_file.ptr));
                SSL_CTX_free(ctx);
+               ossock->ossl_ctx = NULL;
                return status;
            } else {
                PJ_LOG(4,(ssock->pool->obj_name,
@@ -1225,6 +1227,7 @@ static pj_status_t init_ossl_ctx(pj_ssl_sock_t *ssock)
                             "Error adding private key from '%s'",
                             cert->privkey_file.ptr));
                SSL_CTX_free(ctx);
+               ossock->ossl_ctx = NULL;
                return status;
            } else {
                PJ_LOG(4,(ssock->pool->obj_name,
@@ -1274,6 +1277,7 @@ static pj_status_t init_ossl_ctx(pj_ssl_sock_t *ssock)
                        X509_free(xcert);
                        BIO_free(cbio);
                        SSL_CTX_free(ctx);
+                       ossock->ossl_ctx = NULL;
                        return status;
                    } else {
                        PJ_LOG(4,(ssock->pool->obj_name,
[...]

PR #3070, opened 2022-04-14
Fix double free of ossock->ossl_ctx in case of errors (#3069)

johado:
Set ossl_ctx pointer to NULL when freeing it to avoid later
double free in ssl_destroy()


pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1,4 +1,4 @@
-/* $Id$ */
+/* $Id$ */
```
sauwming, line 1: accidental modification?

sauwming (approved):

pjlib/src/pj/ssl_sock_ossl.c
johado: Yes, sorry about that. Have pushed an update.

nanangizz (approved):

Issue #3068, opened 2022-04-14
Build problem due to pthread_setspecific and pj_thread_register with picky compiler


johado:
### Describe the bug

Get compilation problem when building an application using pjsip.

` In function 'pj_thread_local_set',
|     inlined from 'suspend_logging' at ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/log.c:287:2,
|     inlined from 'pj_log' at ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/log.c:355:5:
| ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/os_core_unix.c:1192:12: warning: 'pthread_setspecific' expecting 1 byte in a region of size 0 [-Wstringop-overread]
| ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/os_core_unix.c: In function 'pj_log':
| /usr/include/pthread.h:1308:12: note: in a call to function 'pthread_setspecific' declared with attribute 'access (none, 2)'
| In function 'pj_thread_init',
|     inlined from 'pj_thread_init' at ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/os_core_unix.c:601:13,
|     inlined from 'pj_init' at ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/os_core_unix.c:185:13:
| ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/os_core_unix.c:611:12: warning: 'pj_thread_register' accessing 256 bytes in a region of size 56 [-Wstringop-overflow=]
| ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/os_core_unix.c: In function 'pj_init':
| ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/os_core_unix.c:611:12: note: referencing argument 2 of type 'long int *'
| ../../../pjproject/2.12-r0/pjproject-2.12/pjlib/build/../src/pj/os_core_unix.c:526:21: note: in a call to function 'pj_thread_register'`

### Steps to reproduce

Building our application that uses pjsip.

One issue is resolved by using a pj_thread_desc instead of pj_thread_t:
```

--- a/pjlib/src/pj/os_core_unix.c
+++ b/pjlib/src/pj/os_core_unix.c
[...]

PR #3071, opened 2022-04-14
Fix compilation problem due to pthread_setspecific (#3068)

johado:
This fixes compilation problems with a picky compiler using
-Wstringop-overread -Wstringop-overflow for functions
pj_thread_local_set and pj_thread_init

Change-Id: Ia14d6079b8274e5ccc6ba72cf579e5f8344f8fed


pjlib/src/pj/log.c
```diff
@@ -284,8 +284,8 @@ static void suspend_logging(int *saved_level)
 #if PJ_HAS_THREADS
     if (thread_suspended_tls_id != -1) 
     {
-	pj_thread_local_set(thread_suspended_tls_id, 
-			    (void*)(pj_ssize_t)PJ_TRUE);
+	static int suspend_log = PJ_TRUE;
```
sauwming, line 287: perhaps make the type `pj_ssize_t` so it's the same as the previous code?

sauwming (commented):
My compiler only accepts `-Wstringop-overflow` but doesn't have `-Wstringop-overread`. It doesn't produce any warning here though, but I think the patch looks correct, especially the use of `pj_thread_desc`.

pjlib/src/pj/log.c
johado: Don't think it should matter since it's a pointer being passed.
I guess it's related to this: https://sourceware.org/bugzilla/show_bug.cgi?id=27714
I checked how this was used and when fetching thread_suspended_tls_id, it only checks if the pointer is there, it doesn't really check the value.

sauwming: Reading the reference, it doesn't seem that everybody is happy about the latest change, so perhaps they'll go back to the old way?

There's another one below though, in `resume_logging()`.

sauwming (approved):

pjlib/src/pj/log.c
```diff
@@ -299,8 +300,9 @@ static void resume_logging(int *saved_level)
 #if PJ_HAS_THREADS
     if (thread_suspended_tls_id != -1) 
     {
-        pj_thread_local_set(thread_suspended_tls_id,
-                            (void*)(pj_size_t)PJ_FALSE);
+        static pj_ssize_t suspend_log = PJ_FALSE;
+        pj_thread_local_set(thread_suspended_tls_id, 
+                            (void*)&suspend_log);
```
nanangizz, line 305: As now the TLS values have been changed, `is_logging_suspended()` needs to be updated too, something like:
```
return *((pj_ssize_t*)pj_thread_local_get(thread_suspended_tls_id)) == PJ_TRUE;
```

sauwming: Good catch!
It can be NULL though, the first time `pj_log()` is called.

nanangizz (approved):

Issue #3067, opened 2022-04-14
undefined reference to 'check_srtp_roc' when PJSUA_MEDIA_HAS_PJMEDIA = 0


johado:
### Describe the bug

We're using gstreamer as media backend so we have #define PJSUA_MEDIA_HAS_PJMEDIA 0
With 2.12 it no longer compiles/links since the #if statements in src/pjsua-lib/pjsua_media.c removes the check_srtp_roc() function.

### Steps to reproduce

Add
#define PJSUA_MEDIA_HAS_PJMEDIA 0 
in config_site.h and try to compile.

### PJSIP version

2.12

### Context

Linux with #define PJSUA_MEDIA_HAS_PJMEDIA 0 in config_site.h

### Log, call stack, etc

```shell
/path/recipe-sysroot-native/usr/bin/arm-poky-linux-gnueabi/../../libexec/arm-poky-linux-gnueabi/gcc/arm-poky-linux-gnueabi/11.2.0/ld: /tmp/cc3CqFzc.ltrans3.ltrans.o: in function `pjsua_media_channel_update':
| /path/../../../pjproject/2.12-r0/pjproject-2.12/pjsip/build/../src/pjsua-lib/pjsua_media.c:3902: undefined reference to `check_srtp_roc'
| collect2: error: ld returned 1 exit status
```

johado:
Added https://github.com/pjsip/pjproject/pull/3072

sauwming:
Fixed by #3019.

PR #3019, opened 2022-03-14
third party media build error fix

emretufekci:

PR #3072, opened 2022-04-14
Fix possibly undefined reference to check_srtp_roc (#3067)

johado:
Move the
if PJSUA_MEDIA_HAS_PJMEDIA || PJSUA_THIRD_PARTY_STREAM_HAS_GET_INFO
so check_srtp_roc gets included even if PJSUA_MEDIA_HAS_PJMEDIA = 0

Change-Id: I974480a294c89124a528b6fd6d101b8095bcf036

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -3279,6 +3279,7 @@ pj_status_t pjsua_media_channel_deinit(pjsua_call_id call_id)
     return PJ_SUCCESS;
 }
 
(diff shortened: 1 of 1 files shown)


sauwming:
The fix in #3019 is actually the other way around, i.e. to prevent the calling of `check_srtp_roc()` if the conditions are true.

johado:
So is adding 
#define PJSUA_THIRD_PARTY_STREAM_HAS_GET_INFO 1
to site_config.h the correct solution? (Appears to compile and link...)

sauwming:
It depends. Does your third party media support retrieving the media stream info? If yes, you can enable it, otherwise you can leave it disabled. Either way, it should build and run fine without error.

I will close this issue, but feel free to reopen it and let us know if this is not the case.

Issue #3077, opened 2022-04-19
ioqueue_create performance


jimying:
### Describe the bug

```c
pj_ioqueue_create( pj_pool_t *pool, pj_size_t max_fd, pj_ioqueue_t **p_ioqueue)
```
When use epoll/kqueue, max_fd may be a large number , eg. 30000

In this case,  pj_ioqueue_create() will take a long time (several seconds, 5 seconds or more) !

The reason is that:  it'll alloc max_fd times memory, which is so slow. 

It can be optimized to apply for contiguous memory only once.

### Steps to reproduce

1.  build pjproject with ioqueue use epoll/kqueue (--enable-epoll / --enable-kqueue)
2. write a simple test, set max_fd to a large number

The main code as follow:
```c
pj_pool_t *pool = pj_pool_create(pf, "tst", 1000, 1000, NULL);
pj_ioqueue_t *ioq;

// set max_fd to  a large number,  here is 30000,  will be slow
pj_ioqueue_create(pool, 30000,  &ioq);
```
3. run test, which  take a long time

### PJSIP version

2.12-dev

### Context

In platform: linux / macos / freebsd

### Log, call stack, etc

```shell
None
```

jimying:
I give a pull request #3078 

I had tested it on linux-x86_64/ macos-x86_64/freebsd-amd64

sauwming:
Actually `pj_pool` is designed to achieve a better performance than regular memory allocation, it's just here it's configured suboptimally.
`pj_pool_create(..., 1000, 1000, ...)` means that the initial size is 1 KB and the increment is also 1 KB, which is definitely insufficient if app already knows that it wants to handle 30K max_fd (not only the initial size is not enough, the pool has to be reallocated so many times which is a very expensive operation).

So perhaps you can try larger numbers, a minimum of `30K * (sizeof(pj_ioqueue_t) + sizeof(pj_mutex_t)) + some allowances` for the initial size and a reasonable increment size.

As for the fix in the PR, if the above works, then a solution is for ioqueue to create its own internal pool, i.e.
```
// Make sure that pool_size can hold max_fd memory requirement
pool_size = ...
ioq->pool = pj_pool_create(pool.factory, , ...)
```
And later in ioq destroy: `pj_pool_release(ioq->pool)`.

Alternatively, leave it to the app to configure the pool with more optimal setting.

jimying:
@sauwming 
Configure the app pool initial size to an estimated value to:

[...]

PR #3078, opened 2022-04-19
Improve performance for ioqueue_create (alloc memory)

jimying:
When use epoll/kqueue, pj_ioqueue_create() 's param max_fd set to a large number (eg. 30000),
 it will take a long time to init.

The reason is that: 
It'll alloc max_fd  times memory (indeed 2 * max_fd,  alloc ioqueue_key_t and recursive_mutex), which is so slow.

It can be optimized to apply for contiguous memory only once.

Detail please see issue #3077

### DIFF
--- a/pjlib/src/pj/ioqueue_epoll.c
+++ b/pjlib/src/pj/ioqueue_epoll.c
@@ -30,6 +30,7 @@
 #include <pj/log.h>
 #include <pj/list.h>
 #include <pj/pool.h>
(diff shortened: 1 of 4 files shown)


jimying:
Configure the app pool initial size to an estimated value,

> 30K * (sizeof(pj_ioqueue_t) + sizeof(pj_mutex_t)) + some allowances

It work fine.   Detail please see issue #3077

No need to change ioqueueu main code.   This pull request is discard

PR #3080, opened 2022-04-20
Fail sending periodic RTP transmission on silence

trengginas:
PR #2989 will re-enable RTP transmission on silence. Further test showed that the RTP transmission is not done on some codec due to this check:
```
else if (frame->type == PJMEDIA_FRAME_TYPE_AUDIO &&
	       frame->buf == NULL &&
	       stream->port.info.fmt.id == PJMEDIA_FORMAT_L16 &&
	       (stream->dir & PJMEDIA_DIR_ENCODING) &&
	       stream->enc_samples_per_pkt < PJ_ARRAY_SIZE(zero_frame))
```

Codec with 48000 clockrate, 1 channel and 20ms ptime, the `enc_samples_per_pkt` is 960. The same array size of `zero_frame`.
Hence failing the check and RTP transmission will be skipped. 
This issue should be noticeable when using opus codec which has the above settings.

### DIFF
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -1443,7 +1443,7 @@ static pj_status_t put_frame_imp( pjmedia_port *port,
 	       frame->buf == NULL &&
 	       stream->port.info.fmt.id == PJMEDIA_FORMAT_L16 &&
 	       (stream->dir & PJMEDIA_DIR_ENCODING) &&
-	       stream->enc_samples_per_pkt < PJ_ARRAY_SIZE(zero_frame))
+	       stream->enc_samples_per_pkt <= PJ_ARRAY_SIZE(zero_frame))
     {
 	pjmedia_frame silence_frame;
 
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -175,6 +175,7 @@ struct pjmedia_stream
     void		    *out_rtcp_pkt;  /**< Outgoing RTCP packet.	    */
     unsigned		     out_rtcp_pkt_size;
 					    /**< Outgoing RTCP packet size. */
+    pj_int16_t		    *zero_frame;    /**< Zero frame buffer.	    */
 
     /* RFC 2833 DTMF transmission queue: */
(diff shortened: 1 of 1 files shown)


sauwming:
> Codec with 48000 clockrate, 1 channel and 20ms ptime, the `enc_samples_per_pkt` is 960. The same array size of `zero_frame`.

Just realised that the transmission will also be skipped if the Opus setting's channel count and ptime are larger than those.

There are two approaches to fix it:
1. Static allocation
```
#if PJMEDIA_HAS_OPUS
    static pj_int16_t zero_frame[2 * 120 * 48000 / 1000];
#else
    static pj_int16_t zero_frame[2 * 30 * 16000 / 1000];
#endif
```
Con:
It will always allocate ~20 KB of memory (whereas the previous allocation was only ~2 KB).
Pro:
The memory required will never exceed the preallocation and doesn't depend on the number of calls.

2. Dynamic allocation
Check if `strm->enc_samples_per_pkt` is greater than the array size of `zero_frame` and allocate dynamically.
Pro:
Only allocate when needed.
Con:
The memory required depends on the number of calls. On the worst case, it can be N * 20 KB, but most of the time (if using default Opus setting), there will be no additional memory required.

What do you guys think?

trengginas:
Vote for second option. It should also accommodate new codec with similar settings, if any in the future.
BTW, how is this calculated:
`static pj_int16_t zero_frame[2 * 120 * 48000 / 1000];` 
I thought it is max channel * max ptime * clock rate.
For opus the ptime is fixed at 20

sauwming:
OK, you can try to accommodate the second option in this PR.

I just read the comment in https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream.c#L2533
that Opus ptime can be up to 120 ms, didn't check the RFC though.

From where did you get the impression that Opus ptime is fixed at 20?

trengginas:
I read it from https://www.gl.com/voice-codecs.html

sauwming:
That limit is set for their own Opus implementation.

sauwming (approved):
Okay, test it with Opus with --ptime greater than 20 and --channel-count 2 to make sure it works as expected.

nanangizz (approved):

Issue #3084, opened 2022-04-25
Add support for embedded headers received in Contact header of SIP 3xx


di-shi:
**Is your feature request related to a problem? Please describe.**
All embedded headers received in SIP 3xx Contact header should present (in non-embedded form) in the redirected INVITE.

**Describe the solution you'd like**
In pjsip_inv_process_redirect, after the target URI (containing embedded header parameters) is set, the embedded headers (in non-embedded form) can be inserted into the buffered SIP message structure, then sent in the redirected INVITE. The header parameters in the target URI must be removed in order to prevent them from being inserted into ACK message. 

**Describe alternatives you've considered**
It may also work to handle the embedded headers in other places, such as when the redirect INVITE message string is built.

PR #3085, opened 2022-04-25
Add support for embedded headers received in Contact header of SIP 3xx

di-shi:
Insert SIP 3xx Contact header URI header parameters (embedded headers) into redirected SIP INVITE message, then remove the header parameters from the URI to prevent the embedded headers from being inserted twice.


di-shi:
For https://github.com/pjsip/pjproject/issues/3084

bennylp:
Thank you for looking into this topic. I think if we are to implement this, we'd better follow the guidelines in RFC 3261 section 8.1.3.4 and 19.1.5 regarding forming headers from URI.

fenichelar:
@bennylp I believe this is already compliant with RFC 3261. Can you please elaborate on what specifically needs to be changed?

bennylp:
Since the URI containing the headers is taken from the 3xx response from the server, it's possible that a rouge or malfunctioning server sends invalid headers such as From, Via, Call-ID, Contact, etc., thus we need to follow the details from the sections that I mentioned.

PR #3086, opened 2022-04-26
Update media event poll timeout

nanangizz:
Currently it is 500ms (perhaps for power/battery saving reason?). This may cause issues, e.g: a newly added socket may need to wait for ~500ms to start polled.

The new timeout, 10ms, is quite far from 500ms, but considering PJSUA is also using 10ms for polling SIP events (code [here](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_core.c#L779)), it should not introduce any issue.

### DIFF
--- a/pjmedia/src/pjmedia/endpoint.c
+++ b/pjmedia/src/pjmedia/endpoint.c
@@ -345,7 +345,7 @@ static int PJ_THREAD_FUNC worker_proc(void *arg)
     pjmedia_endpt *endpt = (pjmedia_endpt*) arg;
 
     while (!endpt->quit_flag) {
-	pj_time_val timeout = { 0, 500 };
+	pj_time_val timeout = { 0, 10 };
 	pj_ioqueue_poll(endpt->ioqueue, &timeout);
     }
 


sauwming (approved):

jimying:
I think it maybe better change in file ioqueue_select.c
```c

959	TRACE__((THIS_FILE, "     poll: no fd is set"));
960       if (timeout)
961            pj_thread_sleep(PJ_TIME_VAL_MSEC(*timeout)); 
      // For pj_thread_sleep(),  When timeout is big than a value
      // set to  smaller value

```

jimying:
Only when use 'Select' ioqueue framwork has this issue, epoll/kqueue framework no this issue

trengginas (approved):

nanangizz:
> Another solution, could be take effect faster, Ioqueue can work fine (add new socket effect no delay) no matter what the pj_ioqueue_poll() timeout is set.

Yes, interesting, it should work, but it does not seem simple to implement, which implicitly also means more risks (e.g: potentially introducing new bugs).

Was thinking to use semaphore or event which seems simpler, but then we don't have the API for waiting with timeout. Note that a worker thread may not poll for ioqueue events only, e.g: PJSUA worker thread also polls timer events, so it cannot wait the semaphore forever.
 
> The disadvantage of this solution is that some code needs to be wrote/modified and tested. But i think it can be as an enhancement point for ioqueue framework.

Also, maintaining two fds may have some drawbacks, e.g:
- Pipe fd may not work on Windows, while using socket may need a bit more effort to maintain as network change may cause socket to be in a bad state (may need to be recreated and re-setup).
- Some ioqueue has max fd number (either by config or OS limit), the max has to be reduced by 2, not much but may not be preferable for some app.

So I guess it is not an ideal solution for now.

jimying:
@nanangizz 
yes, you are right. But if add/remove socket frequently, such as server side process, 
wakeup mechanism will good for real-time data send/recv  (less delay)

nanangizz:
> @nanangizz yes, you are right. But if add/remove socket frequently, such as server side process, wakeup mechanism will good for real-time data send/recv (less delay).
> 
> eg. pjturn-srv maybe has delay for add new sock data relay

Right, just added this ioqueue enhancement to internal note.
Thank you for the feedback.

PR #3088, opened 2022-04-27
Add ioqueue wakeup mechanism to avoid new socket delay

jimying:
This is an enhancement for ioqueue framework.
For detail reason, please see discussion in #3086 

Describe:
When poll_wait(),  if no any event, which will wait till timeout.
During this waiting period, If add a new socket to ioqueue, 
it will has delay to catch new socket's read/write events

Solution:
Socketpair create  a 'pipe', register pipe's read fd to ioqueue first.
When a new socket register to ioqueue,  write 1 byte to 'pipe',
which will wakeup ioqueue (exit poll_wait)  immediately
 to get pipe's read event.
And then ioqueue can start watching new socket's events quickly,
which make it less delay.
(set poll_wait()'s timeout  to a big value, will be more obvious)

### DIFF
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -682,6 +682,21 @@
 #   define PJ_IOQUEUE_MAX_HANDLES	(64)
 #endif
 
+/**
+ *
+ * When add new socket to ioqueue, poll maybe blocked waiting,
+ * new socket's events can't be real-time caught.
+ *
+ * Wakeup mechanism will make ioqueue exit poll blocked waiting state
+ * immediately, to poll sockets event again, which make it less delay for newly
+ * added socket.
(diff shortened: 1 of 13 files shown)


jimying:
The code is already written,   tested on linux. 
Set the macro PJ_IOQUEUE_HAS_WAKEUP = 1 to enable this feature

jimying:
Do more in this patch:
1.  ioqueue_select
     when no sockets to be polled,  maybe sleep long time. Limit it to 10 ms.
2.  ioqueue_epoll
    - Remove  pj_thread_sleep() in pj_ioqueue_poll()
    - Remove some debug code

jimying:
I will find time to merge with latest master code, and open a new PR laster

PR #3092, opened 2022-04-28
Fix AVI player: compile warning and unchecked file_read() return value

nanangizz:
Warning message:
`pjmedia\src\pjmedia\avi_player.c(357) : warning C4018: '<' : signed/unsigned mismatch`

### DIFF
--- a/pjmedia/src/pjmedia/avi_player.c
+++ b/pjmedia/src/pjmedia/avi_player.c
@@ -334,7 +334,7 @@ pjmedia_avi_player_create_streams(pj_pool_t *pool,
     /* Skip any JUNK or LIST INFO until we get MOVI tag */
     do {
         pjmedia_avi_subchunk ch;
-        int read = 0;
+        pj_uint32_t read = 0;
         pj_off_t size_to_read;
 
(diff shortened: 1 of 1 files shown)


sauwming (approved):

pjmedia/src/pjmedia/avi_player.c
```diff
@@ -334,7 +334,7 @@ pjmedia_avi_player_create_streams(pj_pool_t *pool,
     /* Skip any JUNK or LIST INFO until we get MOVI tag */
     do {
         pjmedia_avi_subchunk ch;
-        int read = 0;
+        pj_uint32_t read = 0;
```
sauwming, line 337: I have a second thought. How about we just leave the data type here and cast the comparison below instead?
`read` is to be passed to `file_read()` which takes `pj_ssize_t`, so `int` is correct (or if we want to be perfect, it can be changed to `pj_ssize_t`).

It is the comparison that generates the warning, so changing it to `ch.len < (pj_uint32_t)read` should be safe since `read` is either 0 or 4.

trengginas (approved):

PR #3095, opened 2022-05-02
Use SSL socket for HTTPS connection

chgabriel79:
When using HTTPS, a SSL connection to the server has to be used. Use SSL or regular socket depending on http:// or https:// scheme in URL.


sauwming:
It would be great if you can include/add the unit test as well here:
https://github.com/pjsip/pjproject/blob/master/pjlib-util/src/pjlib-util-test/http_client.c

chgabriel79:
I will add the unit tests once the non-ssl build works.

sauwming:
Closing due to no further update. Please reopen/recreate if necessary.

PR #3100, opened 2022-05-10
Add --disable-libuuid to autoconf

jimying:
libuuid is autodetected and linked by default on linux platform. 
People may not want use libuuid.
--disable-libuuid will  force disable libuuid.


jimying:
Some cases may no need uuid,  so no need force link libuuid.

sauwming (approved):

nanangizz (approved):

PR #3110, opened 2022-05-18
Add run time settings to enable/disable RTCP-XR in account config

nanangizz:
Currently RTCP-XR can be enabled/disabled via compile time settings only, i.e: `PJMEDIA_HAS_RTCP_XR` & `PJMEDIA_STREAM_ENABLE_XR`. This PR adds run time settings:
- PJSUA: `pjsua_acc_config.enable_rtcp_xr`
- PJSUA2: `AccountMediaConfig.rtcpXrEnabled`

Default is `PJMEDIA_STREAM_ENABLE_XR`.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -4329,14 +4329,12 @@ typedef struct pjsua_acc_config
      */
     pjmedia_rtcp_fb_setting rtcp_fb_cfg;
 
-#if defined(PJMEDIA_HAS_RTCP_XR) && (PJMEDIA_HAS_RTCP_XR != 0)
     /**
      * Enable RTCP Extended Report (RTCP XR).
      *
      * Default: PJMEDIA_STREAM_ENABLE_XR
      */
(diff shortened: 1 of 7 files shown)


pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -4329,6 +4329,15 @@ typedef struct pjsua_acc_config
      */
     pjmedia_rtcp_fb_setting rtcp_fb_cfg;
 
+#if defined(PJMEDIA_HAS_RTCP_XR) && (PJMEDIA_HAS_RTCP_XR != 0)
```
sauwming, line 4332: why in pjsua is this protected by #if macro but not in pjsua2?

looking at the other settings (such as video, srtp), it seems that typically settings do not need macro protection

nanangizz: Ah right, added that because there is macro protection of `PJMEDIA_STREAM_ENABLE_KA` few lines up.
Will update the patch.

sauwming (approved):

trengginas (approved):

PR #3119, opened 2022-05-23
version bump: 2.12.1-dev 

liviu-rolo:
synchronized with the released 2.12.1


sauwming:
This is intentional since the release is done from a separate branch.

PR #3120, opened 2022-05-24
Add OpenSSL version check for SSL_SESSION_is_resumable()

sauwming:
The function is only available in OpenSSL 1.1.1 (instead of 1.1.0).
https://www.openssl.org/docs/man1.1.1/man3/SSL_SESSION_is_resumable.html

Thanks to @formater for the info.


trengginas (approved):

ffloimair:
is this still being worked on? No updates for more than a month.

PR #3121, opened 2022-05-24
Patch to improve epoll implementation

sauwming:
It's been reported that there is a performance issue with epoll backend (increased delay of packet received during heavy load, and CPU load not dropping back to normal state afterward).

The PR is an **EXPERIMENTAL(!)** attempt to address the issue, as well as to improve the current epoll implementation:
- Modify the fix of #2747 and #3018 since the rearming was done in `remove_from_set()`, which doesn't seem like the proper place to do so. It is suspected that the early rearming may cause other polling threads to be notified of the same events before they're even processed, potentially causing performance inefficiency. So the rearming is now moved to `add_to_set()` and performed after dispatch instead.
- Modify the sleep period in the special case when `epoll_wait()` returns > 0 but we don't process it. Previously, the sleep period is always using the timeout period provided by the user, disregarding how much time we have spent in `epoll_wait()`, which means that total duration spent inside `pj_ioqueue_poll()` can potentially be up to double of the `timeout` parameter passed by the user (example: `pj_ioqueue_poll(timeout=500 msec) -> os_epoll_wait() returns > 0 after 400 msec -> event_cnt = 0 -> pj_thread_sleep(500)`). Also add some more explanation about this special case.
- Currently socket unregistration will fail if `os_epoll_ctl(EPOLL_CTL_DEL, ...)` fails. This may cause the socket to be in the interest list forever, potentially causing performance issue and failure to properly cleanup and return to the initial state.
- Add compile-time setting to enable/disable exclusive/oneshot feature, which can be useful for app that only use a single thread to poll ioq and therefore, doesn't need the feature.

### DIFF
--- a/pjlib/src/pj/ioqueue_epoll.c
+++ b/pjlib/src/pj/ioqueue_epoll.c
@@ -119,16 +119,17 @@ static void scan_closing_keys(pj_ioqueue_t *ioqueue);
 
 #  include <openssl/opensslv.h>
 #  if OPENSSL_VERSION_NUMBER < 0x10100000L
-#    define DONT_USE_EXCL_ONESHOT
+#    undef PJ_IOQUEUE_EPOLL_ENABLE_EXCLUSIVE_ONESHOT
+#    define PJ_IOQUEUE_EPOLL_ENABLE_EXCLUSIVE_ONESHOT	0
 #  endif
 
 #endif
 
 /* Use EPOLLEXCLUSIVE or EPOLLONESHOT to signal one thread only at a time. */
-#if defined(EPOLLEXCLUSIVE) && !defined(DONT_USE_EXCL_ONESHOT)
+#if defined(EPOLLEXCLUSIVE) && PJ_IOQUEUE_EPOLL_ENABLE_EXCLUSIVE_ONESHOT
 #  define USE_EPOLLEXCLUSIVE	1
 #  define USE_EPOLLONESHOT	0
-#elif defined(EPOLLONESHOT) && !defined(DONT_USE_EXCL_ONESHOT)
+#elif defined(EPOLLONESHOT) && PJ_IOQUEUE_EPOLL_ENABLE_EXCLUSIVE_ONESHOT
 #  define USE_EPOLLEXCLUSIVE	0
 #  define USE_EPOLLONESHOT	1
 #else
@@ -854,15 +855,17 @@ PJ_DEF(int) pj_ioqueue_poll( pj_ioqueue_t *ioqueue, const pj_time_val *timeout)
      *   never specifically asked for them.
      */
     if (count > 0 && !event_cnt && msec > 0) {
-	int delay = msec - pj_elapsed_usec(&t1, &t2)/1000;
+#if !USE_EPOLLEXCLUSIVE && !USE_EPOLLONESHOT
         /* We need to sleep in order to avoid busy polling, such
          * as in the case of the thread that doesn't process
          * the event as explained above.
          * Note that the sleep period should be reduced by
          * the amount of time already used for epoll_wait().
          */
+	int delay = msec - pj_elapsed_usec(&t1, &t2)/1000;
         if (delay > 0)
 	    pj_thread_sleep(delay);
+#endif
     }
 
     TRACE_((THIS_FILE, "     poll: count=%d events=%d processed=%d",
--- a/pjlib/src/pj/ioqueue_epoll.c
+++ b/pjlib/src/pj/ioqueue_epoll.c
@@ -578,21 +578,28 @@ static void ioqueue_remove_from_set( pj_ioqueue_t *ioqueue,
                                      pj_ioqueue_key_t *key, 
                                      enum ioqueue_event_type event_type)
 {
-#if USE_EPOLLONESHOT
-    /* For EPOLLONESHOT, always rearm ioqueue for events. */
-    PJ_UNUSED_ARG(event_type);
-    pj_uint32_t ev = EPOLLIN | EPOLLERR;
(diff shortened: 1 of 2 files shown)


jimying:
I recommand add a macro to global enable/disable  epoll exclusive/oneshot feature,
in some case (single thread poll event) no need the feature.

in pjlib/include/config.h, add global macro
```
#ifndef PJ_IOQUEUE_ENABLE_EPOLL_EXCLUSIVE_ONESHOT
#define PJ_IOQUEUE_ENABLE_EPOLL_EXCLUSIVE_ONESHOT    1
#endif
```

nanangizz:
> @sauwming
> 
>     * I recommand add a macro to global enable/disable  epoll exclusive/oneshot feature,
>       which can be Customized by user.
>       If an app use single thread to poll events,  people can disable this needless feature  when build pjsip lib.
> 
> 
> in pjlib/include/config.h, add macro:
> 
> ```
> #ifndef PJ_IOQUEUE_ENABLE_EPOLL_EXCLUSIVE_ONESHOT
> #define PJ_IOQUEUE_ENABLE_EPOLL_EXCLUSIVE_ONESHOT    1
> #endif
> ```

IMO good idea.

> 
> Also need modify ioqueue_epoll.c, when PJ_IOQUEUE_ENABLE_EPOLL_EXCLUSIVE_ONESHOT = 0, set USE_EPOLLEXCLUSIVE and USE_EPOLLONESHOT to 0
> 
>     * pj_thread_sleep() in pj_ioqueue_poll() problem
>       Only when multi-thread poll events will get event_cnt=0 and trigger sleep
>       **process assume**:
>       If one thread polling got event_cnt > 0, normal process events,
>       simultaneously  another thread polling  got event_cnt=0,  it'll sleep
> 
> 
> eg. Three threads poll events and sleep value is 500ms (maybe more bigger value for test, 5000ms), finally maybe only one thread really work (the other two are sleeping)
> 
> ```
> pj_thread_sleep()  will reduce concurrency,  that's not we want.
> ```
> 
> So sleep time should limit to a small value, eg. 5ms OR when count> 0&& event_cnt ==0, Not call pj_thread_sleep() in pj_ioqueue_poll() , pj_ioqueue_poll() return an error code, upper layer deal with the result (sleep or not)
> 
> OR Not call pj_thread_sleep(), just ignore it
> 
> PS: This problem also appears in other ioqueue frameworks(select/kqueue.....), **should also add the same logic code**.

Current behavior is that `pj_ioqueue_poll()` will sleep only when there is false-event (the events are already handled by other poll() in other threads, represented in the code as `count>0 && event_cnt==0`). Theoritically false-event is only possible when there are multiple pollings and oneshot/exclusive is disabled (note that the sleep is added long before oneshot/exclusive is added). So perhaps we can skip the sleep if oneshot/exclusive is enabled?

nanangizz (approved):

pjlib/src/pj/ioqueue_epoll.c
```diff
@@ -578,20 +586,11 @@ static void ioqueue_remove_from_set( pj_ioqueue_t *ioqueue,
                                      pj_ioqueue_key_t *key, 
                                      enum ioqueue_event_type event_type)
 {
-#if USE_EPOLLONESHOT
```
trengginas, line 581: Now `ioqueue_remove_from_set()` will only executed when event_type is `WRITEABLE_EVENT`.
The method itself is called from `ioqueue_dispatch_read_event(), ioqueue_dispatch_exception_event() and ioqueue_dispatch_write_event()`.
On read and exception event, then there's no rearm? And this patch will call `ioqueue_add_to_set()` only when there's something wrong on the dispatch method.

sauwming: > Now `ioqueue_remove_from_set()` will only executed when event_type is `WRITEABLE_EVENT`.

Correct, because `remove_from_set()` is supposed to remove the event, instead of adding/rearming it, which can cause premature or unexpected rearming.

> The method itself is called from `ioqueue_dispatch_read_event(), ioqueue_dispatch_exception_event() and ioqueue_dispatch_write_event()`.
> On read and exception event, then there's no rearm? And this patch will call `ioqueue_add_to_set()` only when there's something wrong on the dispatch method.

For success case of dispatch, the rearm will be done when `ioqueue_add_to_set()` is called.
For example: `ioqueue_dispatch_read_event()` -> `on_read_complete()` -> `pj_ioqueue_recv/recvfrom()` -> `ioqueue_add_to_set()`.

trengginas: I think there's a possibility that `ioqueue_add_to_set()` won't be called after the dispatch event.
e.g: on `pj_ioqueue_recv()` [ref](https://github.com/pjsip/pjproject/blob/master/pjlib/src/pj/ioqueue_common_abs.c#L767)
There's a possibility the method returns without calling `ioqueue_add_to_set()`. [ref](https://github.com/pjsip/pjproject/blob/master/pjlib/src/pj/ioqueue_common_abs.c#L736).
I think we should re-arm on that scenario as well.

sauwming: Good catch!

On normal scenario, it should be okay, since immediate read doesn't need event notification, and normally app will continue reading until no further data and the read status is pending/async (such as in [sip_transport_udp](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_transport_udp.c#L169) or pjmedia/transport_udp).

But with oneshot, it can be quite fragile indeed since it's difficult to track every flow and any error may cause the rearm not to happen. So to be safe and more robust, I modify it as in the next commit.

trengginas (approved):

Issue #3115, opened 2022-05-20
Add clang-format file


jimying:
It could help contributors and avoid useless code style reviews.

jimying:
According to  [Coding Convention](https://www.pjsip.org/docs/latest/pjlib/docs/html/pjlib_coding_convention_page.htm),

It can be expanded on this basis

```
---
BasedOnStyle: LLVM

IndentWidth: 4
TabWidth:   8   
UseTab: AlignWithSpaces
BreakBeforeBraces: Linux
ColumnLimit: 0
IndentCaseLabels: false

# Disable single line
AllowShortFunctionsOnASingleLine: None
AllowShortIfStatementsOnASingleLine: Never
AllowShortBlocksOnASingleLine : false
AllowShortCaseLabelsOnASingleLine: false
AllowShortLoopsOnASingleLine: false 
AllowShortEnumsOnASingleLine: false

# Function params
AllowAllParametersOfDeclarationOnNextLine: false
BinPackParameters: false
```

sauwming:
We use editorconfig in #3210.

PR #3131, opened 2022-05-31
Add .clang-format file

jimying:
Use clang-format  to help format code.
It could help contributors and avoid useless code style reviews.

According to [Coding Convention](https://www.pjsip.org/docs/latest/pjlib/docs/html/pjlib_coding_convention_page.htm).

This only contains basic functions, It can be expanded on this basis, add other features.

#3115

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,23 @@
+---
+BasedOnStyle: LLVM
+
+IndentWidth: 4
+TabWidth:   8
+UseTab: AlignWithSpaces
+BreakBeforeBraces: Linux
+ColumnLimit: 80
+IndentCaseLabels: false
+
+# Disable single line
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortBlocksOnASingleLine : false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false 
+AllowShortEnumsOnASingleLine: false
+
+# Function params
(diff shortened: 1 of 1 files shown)


sauwming:
Having a coding convention setting file is a good idea, but after playing around with `clang-format`, it doesn't seem to suit our need.

> This only contains basic functions, It can be expanded on this basis, add other features.

You've pretty much summarised what we also want, which is that if we don't specify a setting, it should be ignored and accepted as is. So for example, if we only want to start with basic indentation convention (such as tab, indent width, and column width) and add more later, we should be able to do that. But with clang-format, if you don't specify a style a default will be chosen, so **all** style settings will be taken into account during formatting. This triggers a lot of warning for our existing codebase (`warning: code should be clang-formatted [-Wclang-format-violations]`) for something as simple as trailing whitespace, which is undesirable.

In general, if we are to use a code formatter, it should be clean (or produce only minimal warnings) when we run `code-formatter any_PJSIP_codefile.c`. So this will help contributors fix their code style since they know that any new warnings must be produced by their modifications.

jimying:
@sauwming   

I use vscode edit code, it can format the selected code by mourse .

Prepare:
1. Confirm clang-format command in the PATH
2. Put .clang-format style file in the root of  your project
3.  vscode settting:  "C_Cpp.clang_format_style": "file"

Now you can use mourse select some code, and click right, in the menu click "format the selected document"

sauwming:
Ah, that makes sense.

Unfortunately the .clang-format will fail the rest of PJSIP code :)

PR #3132, opened 2022-05-31
improved status checking in audiodev.c

jurica:
I wanted to contribute this improvement to audiodev.c. The problem here was that, if devices where removed pjmedia caused crashes in the application. Debug builds triggered the removed assertion and release builds caused crashes later on as device id's of devices that where removed where referenced.

As I'm quite unfamiliar with pjprojects codebase itself so I'm not sure if my changes can be adopted without changes, but I hope at least you can implement some improvements here based on my suggestion. At least for our application we use that patch for a few weeks already and are not experiencing any issues with it, but it successfully prevents the crashes we experienced. 

As far as I have seen videodev.c implements the same pattern and most likely causes the same issues, but I can't tell for sure as I haven't tested that part.

### DIFF
--- a/pjmedia/src/pjmedia/audiodev.c
+++ b/pjmedia/src/pjmedia/audiodev.c
@@ -352,8 +352,9 @@ static pj_status_t make_global_index(unsigned drv_idx,
     PJ_ASSERT_RETURN(aud_subsys.drv[drv_idx].f, PJ_EBUG);
 
     /* Check that device index is valid */
-    PJ_ASSERT_RETURN(*id>=0 && *id<(int)aud_subsys.drv[drv_idx].dev_cnt, 
-		     PJ_EBUG);
+    if (!(*id>=0 && *id<(int)aud_subsys.drv[drv_idx].dev_cnt)) {
+        return PJMEDIA_AUD_INVALID_DEV;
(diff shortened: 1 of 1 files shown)


sauwming:
Can you explain what the scenario is here? Is there a race between the device removal and application trying to access the device?

In case that you don't know, we have a FAQ here that touches a similar subject:
https://trac.pjsip.org/repos/wiki/FAQ#snd-hot-plug
although there it discusses about the addition of a new device, instead of a removal, but the concept is the same.

jurica:
Hi @sauwming, I wasn't aware of that FAQ you linked, but that's pretty much what we do. We listen to OS device change events and then refresh PJMEDIA devices, but without calling `pjsua_set_no_snd_dev()` before.
As the FAQ states, I guess calling `pjsua_set_no_snd_dev()` before the refresh would prevent the issue. When debugging this issue I didn't think of that, so I looked for the cause and worked out the change I submitted.
I will try calling `pjsua_set_no_snd_dev()`before refreshing the devices, but it might take some time until I can provide feedback wether it works or not.

sauwming:
Closing due to no further update. Please reopen/recreate whenever you have managed to test it.

Issue #2967, opened 2022-02-07
PRACK SDP offer does not processed by pjsip UAS


vasilevalex:
### Describe the bug

According to RFC 3262, section 5 (https://datatracker.ietf.org/doc/html/rfc3262#section-5) "If the UAC receives a reliable provisional response with an answer, it MAY generate an additional offer in the PRACK. If the UAS receives a PRACK with an offer, it MUST place the answer in the 2xx to the PRACK"
Pjproject mentions in the code that PRACK can contain only SDP answer (https://github.com/pjsip/pjproject/blob/551273f63969d19f6e6a52ed17a72dd720d51391/pjsip/src/pjsip-ua/sip_inv.c#L52) and works according to that assumption.

### Steps to reproduce

1. Send INVITE with SDP offer to pjproject UAS
2. Get 183 with SDP answer
3. Send PRACK with new SDP offer
4. pjproject UAS must respond 200 OK with SDP answer, but it responds with 200 OK without SDP
5. The call is cancelled, as SDP offer have not received the answer.

### PJSIP version

2.11

### Context

pjproject assumes that PRACK can't contain SDP offer, but it can. This behavior should be fixed to do SDP renegotiation when receiving new offer in PRACK.

### Log, call stack, etc

```shell
Example SIP-trace. INVITE is coming to pjsip UAS.

2022/01/31 13:24:45.608352 125.10.10.12:32489 -> 185.15.10.21:5060
INVITE sip:+034551989898@spid123.ims.mnc001.mcc238.3gppnetwork.org SIP/2.0
Max-Forwards: 65
Via: SIP/2.0/TCP 125.10.10.12:5060;branch=z9hG4bKccj55nj9v1i5ki511jdepk54r
To: <sip:+4551989898@62.35.18.42;user=phone>
From: <sip:+4522336611@MSS.OPERATOR.TEL;user=phone>;tag=h7g4Esbg_p65550t1643635485m399035c1159391761s1_3040834397-1104810396
Call-ID: p65550t1643635485m399035c1159391761s2
CSeq: 1 INVITE
Contact: <sip:sgc_c@125.10.10.12;transport=udp>;+g.3gpp.icsi-ref="urn%3Aurn-7%3A3gpp-service.ims.icsi.mmtel"
Record-Route: <sip:125.10.10.12;transport=udp;lr>
[...]

PR #3137, opened 2022-06-01
Don't terminate call on failed SDP negotiation during EARLY state

### LINKED ISSUE #3135: Wrong order of replies to UPDATE request with unacceptable SDP offer
### Describe the bug

pjsip is UAS and receiving UPDATE request before finishing original INVITE transaction. If SDP negotiation for this UPDATE is failed, pjsip first terminates INVITE transaction and then sends reply 488 "Not acceptable" for the UPDATE.
During `inv_negotiate_sdp()` there is execution of callback `pjsua_call_on_media_update` which can terminate the dialog. So if SDP negotiation failed, pjsua first sends reply to original INVITE, terminating current dialog (415 "Unsupported Media Type"), and only after that it is sending reply 488/UPDATE. I think, that correct behavior must be opposite, 488/UPDATE followed by 415/INVITE.

### Steps to reproduce

Scripts to reproduce the issue are attached
[uac-upd-sdp-decline_xml.txt](https://github.com/pjsip/pjproject/files/8812373/uac-upd-sdp-decline_xml.txt)
[uac-upd-sdp-decline_py.txt](https://github.com/pjsip/pjproject/files/8812374/uac-upd-sdp-decline_py.txt)

### PJSIP version

master

### Context

$ mv uac-upd-sdp-decline_xml.txt ./pjproject/tests/pjsua/scripts-sipp/uac-upd-sdp-decline.xml
$ mv uac-upd-sdp-decline_py.txt ./pjproject/tests/pjsua/scripts-sipp/uac-upd-sdp-decline.py
$ cd pjproject/tests/pjsua
$ python run.py  mod_sipp.py scripts-sipp/uac-upd-sdp-decline.xml

### Log, call stack, etc

```shell
pjsua0: 10:25:39.921           pjsua_call.c  .....Call 0: received updated media offer
pjsua0: 10:25:39.921      inv0x7fe7a0003b68  .....SDP negotiation done: No suitable codec for remote offer (PJMEDIA_SDPNEG_NOANSCODEC)
pjsua0: 10:25:39.921           pjsua_call.c  ......SDP negotiation has failed: No suitable codec for remote offer (PJMEDIA_SDPNEG_NOANSCODEC) [status=220049]
pjsua0: 10:25:39.921           pjsua_core.c  .........TX 410 bytes Response msg 415/INVITE/cseq=1 (tdta0x7fe7a0023af8) to UDP 127.0.0.1:50070:
pjsua0: SIP/2.0 415 Unsupported Media Type
pjsua0: Via: SIP/2.0/UDP 127.0.0.1:50070;received=127.0.0.1;branch=z9hG4bK-1244-1-0
pjsua0: Call-ID: 1-1244@127.0.0.1
pjsua0: From: "sipp" <sip:sipp@127.0.0.1>;tag=1244SIPpTag091
pjsua0: To: "service" <sip:service@127.0.0.1>;tag=bda40eb6-b9d9-4ca2-b3ed-836201a5d063
pjsua0: CSeq: 1 INVITE
pjsua0: Allow: PRACK, INVITE, ACK, BYE, CANCEL, UPDATE, INFO, SUBSCRIBE, NOTIFY, REFER, MESSAGE, OPTIONS
pjsua0: Content-Length:  0
pjsua0:
pjsua0:
pjsua0: --end msg--
pjsua0: ================== send call answer 200 ================== [at t=000]
pjsua0: ================== expect state.*CONFIRMED ================== [at t=000]
pjsua0: 10:25:39.921          pjsua_media.c  ............Call 0: deinitializing media..
pjsua0: 10:25:39.921          pjsua_media.c  .............
pjsua0:   [DISCONNCTD] To: "sipp" <sip:sipp@127.0.0.1>;tag=1244SIPpTag091
pjsua0: Call time: 00h:00m:00s, 1st res in 35 ms, conn in 0ms
pjsua0: #0 audio deactivated
pjsua0: 10:25:39.921          pjsua_media.c  ..............Media stream call00:0 is destroyed
pjsua0: 10:25:39.921      udp0x7fe7a000dbf0  .............UDP media transport stopped
pjsua0: 10:25:39.921      udp0x7fe7a000dbf0  .............UDP media transport destroyed
pjsua0: 10:25:39.921            pjsua_app.c  ............Call 0 is DISCONNECTED [reason=415 (Unsupported Media Type)]
pjsua0: 10:25:39.921           pjsua_core.c  .......TX 309 bytes Response msg 488/UPDATE/cseq=3 (tdta0x7fe7a0020ab8) to UDP 127.0.0.1:50070:
pjsua0: SIP/2.0 488 Not Acceptable Here
pjsua0: Via: SIP/2.0/UDP 127.0.0.1:50070;received=127.0.0.1;branch=z9hG4bK-1244-1-5
pjsua0: Call-ID: 1-1244@127.0.0.1
pjsua0: From: "sipp" <sip:sipp@127.0.0.1>;tag=1244SIPpTag091
pjsua0: To: "service" <sip:service@127.0.0.1>;tag=bda40eb6-b9d9-4ca2-b3ed-836201a5d063
pjsua0: CSeq: 3 UPDATE
pjsua0: Content-Length:  0
pjsua0:
pjsua0:
pjsua0: --end msg--
```

vasilevalex:
Fixes #3135

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_call.c
+++ b/pjsip/src/pjsua-lib/pjsua_call.c
@@ -5204,6 +5204,7 @@ static void pjsua_call_on_media_update(pjsip_inv_session *inv,
 	 * UAS dialog and if this is not a re-INVITE
 	 */
 	if (inv->state != PJSIP_INV_STATE_NULL &&
+	    inv->state != PJSIP_INV_STATE_EARLY &&
 	    inv->state != PJSIP_INV_STATE_CONFIRMED)
 	{
 	    call_disconnect(inv, PJSIP_SC_UNSUPPORTED_MEDIA_TYPE);
--- a/pjsip/src/pjsua-lib/pjsua_call.c
+++ b/pjsip/src/pjsua-lib/pjsua_call.c
@@ -5203,9 +5203,10 @@ static void pjsua_call_on_media_update(pjsip_inv_session *inv,
 	/* Disconnect call if we're not in the middle of initializing an
 	 * UAS dialog and if this is not a re-INVITE
 	 */
-	if (inv->state != PJSIP_INV_STATE_NULL &&
+	if ((inv->state != PJSIP_INV_STATE_NULL &&
 	    inv->state != PJSIP_INV_STATE_EARLY &&
-	    inv->state != PJSIP_INV_STATE_CONFIRMED)
+	    inv->state != PJSIP_INV_STATE_CONFIRMED) ||
+	    (inv->state == PJSIP_INV_STATE_EARLY && call->med_cnt == 0))
 	{
 	    call_disconnect(inv, PJSIP_SC_UNSUPPORTED_MEDIA_TYPE);
 	}
--- a/tests/pjsua/scripts-sipp/uac-183-wrong-codec.py
+++ b/tests/pjsua/scripts-sipp/uac-183-wrong-codec.py
@@ -1,0 +1,9 @@
+# $Id$
+#
+import inc_const as const
+
+PJSUA = ["--null-audio --max-calls=1 --no-tcp --auto-answer=183"]
+
+PJSUA_EXPECTS = [[0, "Response msg 488/INVITE", ""]]
+
+PJSUA_CLI_EXPECTS = [[0, "Response msg 488/INVITE", ""]]
--- a/tests/pjsua/scripts-sipp/uac-183-wrong-codec.xml
+++ b/tests/pjsua/scripts-sipp/uac-183-wrong-codec.xml
@@ -1,0 +1,85 @@
+<?xml version="1.0" encoding="ISO-8859-1" ?>
+<!DOCTYPE scenario SYSTEM "sipp.dtd">
+
+<!-- This program is free software; you can redistribute it and/or      -->
+<!-- modify it under the terms of the GNU General Public License as     -->
+<!-- published by the Free Software Foundation; either version 2 of the -->
+<!-- License, or (at your option) any later version.                    -->
+<!--                                                                    -->
+<!-- This program is distributed in the hope that it will be useful,    -->
+<!-- but WITHOUT ANY WARRANTY; without even the implied warranty of     -->
+<!-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the      -->
+<!-- GNU General Public License for more details.                       -->
+<!--                                                                    -->
+<!-- You should have received a copy of the GNU General Public License  -->
+<!-- along with this program; if not, write to the                      -->
+<!-- Free Software Foundation, Inc.,                                    -->
+<!-- 59 Temple Place, Suite 330, Boston, MA  02111-1307 USA             -->
+<!--                                                                    -->
+<!-- Correct call cancellation on wrong SDP offer PR #3137              -->
+
+<scenario name="UAC sending INVITE with unsupported codec in SDP offer">
+  <send retrans="500">
+    <![CDATA[
+
+      INVITE sip:[service]@[remote_ip]:[remote_port] SIP/2.0
+      Via: SIP/2.0/[transport] [local_ip]:[local_port];branch=[branch]
+      From: sipp <sip:sipp@[local_ip]:[local_port]>;tag=[pid]SIPpTag09[call_number]
+      To: [service] <sip:[service]@[remote_ip]:[remote_port]>
+      Call-ID: [call_id]
+      CSeq: [cseq] INVITE
+      Contact: sip:sipp@[local_ip]:[local_port]
+      Max-Forwards: 70
+      Allow: INVITE, ACK, CANCEL, BYE, PRACK, UPDATE, REFER, MESSAGE
+      Supported: 100rel, precondition
+      Content-Type: application/sdp
+      Content-Length: [len]
+
+      v=0
+      o=user1 53655765 2353687637 IN IP[local_ip_type] [local_ip]
+      s=-
+      c=IN IP[local_ip_type] [local_ip]
+      t=0 0
+      m=audio [auto_media_port] RTP/AVP 109 105
+      a=rtpmap:109 EVS/16000
+      a=fmtp:109 br=5.9-24.4; bw=nb-wb; max-red=220
+      a=rtpmap:105 telephone-event/16000
+      a=fmtp:105 0-15
+      a=sendrecv
+
+    ]]>
+  </send>
+
+  <recv response="100" optional="true">
+  </recv>
+
+  <recv response="183" optional="true">
+  </recv>
+
+  <recv response="488" rtd="true" crlf="true">
+  </recv>
+
+  <send>
+    <![CDATA[
+
+      ACK sip:[service]@[remote_ip]:[remote_port] SIP/2.0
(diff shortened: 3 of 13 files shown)


sauwming (approved):

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -5204,6 +5204,7 @@ static void pjsua_call_on_media_update(pjsip_inv_session *inv,
 	 * UAS dialog and if this is not a re-INVITE
 	 */
 	if (inv->state != PJSIP_INV_STATE_NULL &&
+	    inv->state != PJSIP_INV_STATE_EARLY &&
```
nanangizz, line 5207: This patch seems to cause a behavior change, i.e: if the SDP offer/answer nego from INVITE transaction (not UPDATE) fails, the call will be no longer disconnected, or even the call will never get disconnected on any SDP nego failure.

As the reported issue seems to be SDP nego failure of UPDATE while processing the initial INVITE, the fix should affect that specific scenario only. Unfortunately the callback does not provide `tsx` or `rdata` info, so we cannot check if it is SDP offer/answer from UPDATE, but perhaps we can use `call->med_cnt` info for checking whether there has been a successful SDP nego before (from INVITE+18x)?

vasilevalex: Hi @nanangizz , this issue was found, trying to fix #2967, so yes it should affect not only UPDATE, but also PRACK SDP offer/answer negotiation.
I tried, but I really can't imagine the scenario, when the call in EARLY state must be disconnected. Can you please describe the scenario? Then I can prepare the sipp test, that is failing and fix it.

nanangizz: AFAIK, the current behavior of PJSUA lib is as follow:

For UAC:
1. send INVITE with SDP offer
2. recv 183 with SDP answer
3. when UAC sees failure in SDP nego, it will disconnect call (by sending CANCEL)

For UAS:
1. receive INVITE with SDP offer
2. generate SDP answer (e.g: for 183 or 200 response)
3. the generated answer will be negotiated with the offer, if it fails, call will be disconnected

The proposed patch seems to change this behavior?

vasilevalex: I prepared UAC and UAS tests with 3 different unacceptable SDP:
1) audio codec that is not supported by pjsua 
2) media type video, instead of audio
3) SDP version=1 (must be v=0)
Don't be confused by test names, they show the role of sipp (uac/uas)

First of all, I can confirm, that this patch changes behavior in one case (pjsua is UAC and receives 183 with SDP media type=video) uas-answer-183-wrong-media.xml
It can be fixed for example with this not too elegant code:
```
 if ((inv->state != PJSIP_INV_STATE_NULL &&
    inv->state != PJSIP_INV_STATE_EARLY &&
    inv->state != PJSIP_INV_STATE_CONFIRMED) ||
    (inv->state == PJSIP_INV_STATE_EARLY && call->med_cnt == 0))
```
Interesting case is number 3. Pjsua receives completely wrong SDP. It does not work correctly in both UAC/UAS scenarios even without any patches, but using current source code.
When pjsua is UAC (uas-answer-183-wrong-version.xml) it detects an error `Error parsing/validating SDP body: Invalid SDP version line`, but never disconnects the call.
When pjsua is UAS (uac-183-wrong-version.xml) it answers `400 Bad SDP`, receives ACK, but continues retransmitting "400 Bad SDP". Call is not disconnected too.
Function `inv_on_state_early` never checks return status of `inv_check_sdp_in_incoming_msg`.
Tests are attached.
[uas-answer-183-wrong-version.xml.txt](https://github.com/pjsip/pjproject/files/9828542/uas-answer-183-wrong-version.xml.txt)
[uas-answer-183-wrong-version.py.txt](https://github.com/pjsip/pjproject/files/9828543/uas-answer-183-wrong-version.py.txt)
[uas-answer-183-wrong-media.xml.txt](https://github.com/pjsip/pjproject/files/9828544/uas-answer-183-wrong-media.xml.txt)
[uas-answer-183-wrong-media.py.txt](https://github.com/pjsip/pjproject/files/9828545/uas-answer-183-wrong-media.py.txt)
[uas-answer-183-wrong-codec.xml.txt](https://github.com/pjsip/pjproject/files/9828546/uas-answer-183-wrong-codec.xml.txt)
[uas-answer-183-wrong-codec.py.txt](https://github.com/pjsip/pjproject/files/9828547/uas-answer-183-wrong-codec.py.txt)
[uac-183-wrong-version.xml.txt](https://github.com/pjsip/pjproject/files/9828548/uac-183-wrong-version.xml.txt)
[uac-183-wrong-version.py.txt](https://github.com/pjsip/pjproject/files/9828549/uac-183-wrong-version.py.txt)
[uac-183-wrong-media.xml.txt](https://github.com/pjsip/pjproject/files/9828550/uac-183-wrong-media.xml.txt)
[uac-183-wrong-media.py.txt](https://github.com/pjsip/pjproject/files/9828551/uac-183-wrong-media.py.txt)
[uac-183-wrong-codec.xml.txt](https://github.com/pjsip/pjproject/files/9828552/uac-183-wrong-codec.xml.txt)
[uac-183-wrong-codec.py.txt](https://github.com/pjsip/pjproject/files/9828553/uac-183-wrong-codec.py.txt)

nanangizz: Thanks for the pjsua/sipp tests! So the behavior change is confirmed and we need to fix it. And regarding case no 3, perhaps we should create a separate issue/PR? Good catch btw.

sauwming: > ```
>  if ((inv->state != PJSIP_INV_STATE_NULL &&
>     inv->state != PJSIP_INV_STATE_EARLY &&
>     inv->state != PJSIP_INV_STATE_CONFIRMED) ||
>     (inv->state == PJSIP_INV_STATE_EARLY && call->med_cnt == 0))
> ```

I'm wondering if checking `if (call->med_cnt == 0)` only actually is sufficient here? since we should only disconnect the call if there's no media, otherwise we should revert the previous media.

vasilevalex: @sauwming I'm completely newbie in pjproject code, so I'm not 100% sure, that `(call->med_cnt == 0)` covers all the cases, when call should be disconnected in EARLY state. It's worth to make tests with incompatible SDP for late negotiation for example. But they should also have `call->med_cnt == 0`
According to RFC 6337 section 2.2 (https://datatracker.ietf.org/doc/html/rfc6337#section-2.2) only PRACK and UPDATE can send SDP offer in EARLY state, but `call->med_cnt == 1` in these cases.
Actually, all these tests with wrong SDP in 183 are emulating evil UAS, because normal response must be 488 or something similar. There can be other unpredictable cases, which break RFCs.
@nanangizz I'll create one more issue for 183 reply with broken SDP (e.g. v=1)

nanangizz: Also, adding some comments (and/or add link to this PR) about the new added checks is usually useful for future ref.

nanangizz (approved):

sauwming (approved):

PR #3145, opened 2022-06-13
Random crash due to invoking SRTP APIs on custom media transport instance

nanangizz:
There were reports that `pjmedia_transport_srtp_get_setting()` inside `check_srtp_roc()` ([code](https://github.com/pjsip/pjproject/blob/8daf5a9216f846cd2a2866e48b17810915ba3c92/pjsip/src/pjsua-lib/pjsua_media.c#L3405) was added in #2846) is called on custom media transport instance instead of SRTP transport instance:
```
    pjmedia_transport_srtp_get_setting(call_med->tp, &setting);
```
When custom media transport is not used and SRTP is not disabled in compile-time, `call_med->tp` is normally an SRTP media transport, so the function call above is executed fine. But when custom media transport is used, it may introduce problems as there is no check whether the `call_med->tp` is really an SRTP instance.

Actually the `check_srtp_roc()` has a check whether SRTP is being used:
```
    srtp_info = (pjmedia_srtp_info *) pjmedia_transport_info_get_spc_info(
	            &tpinfo, PJMEDIA_TRANSPORT_TYPE_SRTP);
    /* We are not using SRTP. */
    if (!srtp_info)
    	return;
```
But when the custom media transport is stacked on top of the PJSUA-precreated SRTP+UDP/ICE transports (i.e: custom transport adapter -> SRTP -> ICE/UDP), it will pass the above check.

This PR will make sure that `pjmedia_transport_srtp_get_setting()` is called on SRTP transport. For this purpose, the media transport instance info is now added to media transport specific info structure.

### DIFF
--- a/pjmedia/include/pjmedia/transport.h
+++ b/pjmedia/include/pjmedia/transport.h
@@ -539,6 +539,11 @@ typedef struct pjmedia_transport_specific_info
      */
     char		     buffer[PJMEDIA_TRANSPORT_SPECIFIC_INFO_MAXSIZE];
 
+    /**
+     * The media transport instance.
+     */
+    pjmedia_transport	    *tp;
+
 } pjmedia_transport_specific_info;
 
 
@@ -714,6 +719,29 @@ PJ_INLINE(void*) pjmedia_transport_info_get_spc_info(
     for (i = 0; i < info->specific_info_cnt; ++i) {
 	if (info->spc_info[i].type == type)
 	    return (void*)info->spc_info[i].buffer;
+    }
+    return NULL;
+}
+
+
+/**
+ * Utility API to get the transport instance from the specified media
+ * transport info.
+ *
+ * @param info	    Media transport info.
+ * @param type	    Media transport type.
+ *
+ * @return	    The media transport instance, or NULL if
+ * 		    the transport type is not found.
+ */
+PJ_INLINE(pjmedia_transport*) pjmedia_transport_info_get_transport(
+						pjmedia_transport_info *info,
+						pjmedia_transport_type type)
+{
+    unsigned i;
+    for (i = 0; i < info->specific_info_cnt; ++i) {
+	if (info->spc_info[i].type == type)
+	    return info->spc_info[i].tp;
     }
     return NULL;
 }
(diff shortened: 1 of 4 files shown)


sauwming (approved):

mcpttdev:
> /* We are not using SRTP. */

I was confused by this as well initially but as you say, it is always created but might not be used.

First test with the patch looks fine.

nanangizz:
> Hey, I stumbled upon this and just wanted to add that I find this comment somewhat misleading:
> 
> https://github.com/pjsip/pjproject/blob/772d6906134376102d1b4bf4067b4e7a7d449873/pjsip/src/pjsua-lib/pjsua_media.c#L3343
> 
> In my app I implemented the on_stream_created callback and replace the transport given by PJ with my own in there. However, I now wanted to implement a logic to **NOT** do this when SRTP is involved and added the check that is done in check_srtp_roc() which I found to not work because it always returned an srtp_settings structure. After investigating a little bit I found that the SRTP adapter is created **always** when SRTP support is compiled in; even when it is not being used (i.e. because SRTP use has been disabled in the corresponding account).

Hmm right, "being used" can mean SRTP is active as in actively used for encrypting/decrypting RTP packets, so just changed the comment. Let us know if it may still cause confusion.

> 
> My current workaround is now to check the "use" parameter of the returned srtp_settings structure and _assume_ SRTP is used if the values does not equal _PJMEDIA_SRTP_DISABLED_ which in my case works.

Actually `pjmedia_srtp_info.use` does not necessarily mean SRTP is active too. As you may already know, SRTP nego may end up SRTP being disabled too (e.g: when local is in optional-mode and remote is in disabled-mode). As you may already know, there is `pjmedia_srtp_info.active` field, but it will only be set to true after SRTP nego is succesful.

andreas-wehrmann:
> Actually `pjmedia_srtp_info.use` does not necessarily mean SRTP is active too. As you may already know, SRTP nego may end up SRTP being disabled too (e.g: when local is in optional-mode and remote is in disabled-mode). As you may already know, there is `pjmedia_srtp_info.active` field, but it will only be set to true after SRTP nego is succesful.

Right, I'm  aware of this; like I said it works in my specific use case.
I have three local accounts (UDP, TCP with SRTP disabled explicitly) and one TLS account with srtp_use set to OPTIONAL.
Incoming calls via TLS are automatically assigned to the local account, so checking for the _"use"_ param should be okay even if SRTP is not activated later on.

trengginas (approved):

PR #3146, opened 2022-06-14
Option to use SIP UPDATE for refreshing calls in IP change

nanangizz:
- Added new setting `pjsua_ip_change_acc_cfg.reinv_use_update`.
- It will check if remote supports SIP UPDATE first.
- If SIP UPDATE fails, retry using re-INVITE.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -2696,6 +2696,15 @@ typedef struct pjsua_ip_change_acc_cfg
      */
     unsigned		reinvite_flags;
     
+    /**
+     * For refreshing the call, use SIP UPDATE when remote supports it.
+     * If remote does not support UPDATE method (i.e: published in its Allow
+     * header), use SIP re-INVITE.
+     *
+     * Default: PJ_FALSE (using re-INVITE).
+     */
+    pj_bool_t		reinv_use_update;
+
 } pjsua_ip_change_acc_cfg;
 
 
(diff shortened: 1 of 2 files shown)


trengginas (approved):

sauwming (approved):
Add docs that explains that reinvite_flags will be used regardless whether it is re-INVITE or UPDATE that is sent.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -2696,6 +2696,17 @@ typedef struct pjsua_ip_change_acc_cfg
...
+    /**
+     * For refreshing the call, use SIP UPDATE, instead of re-INVITE, if
+     * remote supports it (by publishing it in Allow header). If remote
+     * does not support UPDATE method or somehow the UPDATE attempt fails,
+     * it will fallback to using re-INVITE. The \a reinvite_flags will be
+     * used regardless whether it is re-INVITE or UPDATE that is sent.
+     *
+     * Default: PJ_FALSE (using re-INVITE).
+     */
+    pj_bool_t		reinv_use_update;
```
kevinkouketsu, line 2708: I guess it is missing changes on pjsua2 account.cpp, on `AccountConfig::toPj` and `
PJ_DEF(void) pjsua_acc_config_default(pjsua_acc_config *cfg)`

nanangizz: You're right! So just added them, except for `pjsua_acc_config_default()`, looks like when the default value is PJ_FALSE, zero, or NULL, it can rely on the `pj_bzero()` up above, e.g: `mwi_enabled`, `proxy_cnt`, `contact_params`.

Issue #3150, opened 2022-06-18
ioqueue_epoll  exclusive/oneshot detect wrong on linux:centos7


jimying:
### Describe the bug

When build pjsip on centos7 with option --enable-epoll,  it detects that use epoll-oneshot.

But  I use [is_epollexclusive_available.cc](https://github.com/grpc/grpc/blob/87acbadba1fff7b4c4add5b8ff9ed37a3d61c3fe/src/core/lib/iomgr/is_epollexclusive_available.cc) to check,
the result is that it supports epoll-exclusive.

REASON:
```
Centos7/REHL7 's kernel (3.10.xxxx ) has pached the EPOLLEXCLUSIVE feature 
(which newly added since Linux 4.5, glibc 2.24 about year 2016)

ioqueue_epoll check  EPOLLEXCLUSIVE macro which not define in <sys/epoll.h> on centos7 (use glibc 2.17)
```

### Steps to reproduce

1. configure --enable-epoll
2. run any sample that use ioqueue, see the log print about epoll info

### PJSIP version

2.12-dev

### Context

test on centos7.9 x86_64

### Log, call stack, etc

none

PR #3151, opened 2022-06-18
Run-time detect epoll exclusive/oneshot support

jimying:
Why use Run-time detection:
1.  Build on some linux version (centos7),  the epoll  detection is wrong 
    #3150 
2. Build environment maybe is old version linux (not support exclusive),  but it  run on higher version linux (support exclusive). 
     Eg. 
Build the icedemo on centos6 (only support epoll-oneshot), 
copy and run  icedemo binary on centos6/centos7/centos8, 
Run-time detection will make it auto choose use exclusive or oneshot,
centos6 is oneshot, centos7/centos8 is exclusive.
**If no Run-time detection,   centos6/centos7/centos8 will all use oneshot.**

PS: Thanks to  [is_epollexclusive_available.cc](https://github.com/grpc/grpc/blob/87acbadba1fff7b4c4add5b8ff9ed37a3d61c3fe/src/core/lib/iomgr/is_epollexclusive_available.cc), the main logic is copy from it.

### DIFF
--- a/pjlib/src/pj/ioqueue_epoll.c
+++ b/pjlib/src/pj/ioqueue_epoll.c
@@ -38,6 +38,7 @@
 #include <pj/rand.h>
 
 #include <sys/epoll.h>
+#include <sys/eventfd.h>
 #include <errno.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


sauwming:
Thanks for the patch. I tested it here on Ubuntu that doesn't have `EPOLLEXCLUSIVE` defined in the header and it can now use exclusive.

I pushed a minor commit to fix the indentation. I hope that's okay with you.

sauwming (approved):

nanangizz (approved):

PR #3156, opened 2022-06-21
Set thread display name

jimying:
This is an enhancement
when  pj_thread_create()  to setup  thread_name, also should set  thread display name
So that we can  see thread name use top command on linux/freebsd: 
```
top -H -p pid
```
also gdb debug cmd:
```
like this:
(gdb) info threads
  Id   Target Id         Frame 
  3   Thread 0x7ffff1521700 (LWP 24614) "worker_2" 0x00007ffff73e6
  2   Thread 0x7ffff1d22700 (LWP 24613) "worker_1" 0x00007ffff73e5
  1   Thread 0x7ffff2523700 (LWP 24612) "main" 0x00007ffff73e5eb3
```

win32 platform also supported now.

This is very useful for debugging.

#3157

### DIFF
--- a/pjlib/src/pj/os_core_unix.c
+++ b/pjlib/src/pj/os_core_unix.c
@@ -47,6 +47,10 @@
 #include <errno.h>	    // errno
 
 #include <pthread.h>
+#if defined(PJ_HAS_PTHREAD_NP_H) && PJ_HAS_PTHREAD_NP_H != 0
+#  include <pthread_np.h>
+#endif
+#include <pj/config.h>
 
 #define THIS_FILE   "os_core_unix.c"
 
@@ -616,6 +620,37 @@ pj_status_t pj_thread_init(void)
 }
 
 #if PJ_HAS_THREADS
+
+/*
+ * Set current thread display name
+ * This can be useful for debugging, as the name is displayed in the thread status
+ */
+static void set_thread_display_name(const char *name)
+{
+#if (defined(PJ_LINUX) && PJ_LINUX != 0) ||                                    \
+    (defined(PJ_ANDROID) && PJ_ANDROID != 0)
+    char xname[16];
+    // On linux, thread display name length is restricted to 16 (include '\0')
+    if (pj_ansi_strlen(name) >= 16) {
+	pj_ansi_snprintf(xname, 16, "%s", name);
+	name = xname;
+    }
+#endif
+
+#if defined(PJ_HAS_PTHREAD_SETNAME_NP) && PJ_HAS_PTHREAD_SETNAME_NP != 0
+#   if defined(PJ_DARWINOS) && PJ_DARWINOS != 0
+    pthread_setname_np(name);
+#   else
+    pthread_setname_np(pthread_self(), name);
+#   endif
+#elif defined(PJ_HAS_PTHREAD_SET_NAME_NP) && PJ_HAS_PTHREAD_SET_NAME_NP != 0
+    pthread_set_name_np(pthread_self(), name);
+#else
+#   warning "OS not support set thread display name"
+    PJ_UNUSED_ARG(name);
+#endif
+}
+
 /*
  * thread_main()
  *
@@ -644,6 +679,8 @@ static void *thread_main(void *param)
     }
 
     PJ_LOG(6,(rec->obj_name, "Thread started"));
+
+    set_thread_display_name(rec->obj_name);
 
     /* Call user's entry! */
     result = (void*)(long)(*rec->proc)(rec->arg);
--- a/aconfigure
+++ b/aconfigure
@@ -5714,6 +5714,10 @@ case $target in
 	printf "%s\n" "#define PJ_LINUX 1" >>confdefs.h
 
 	;;
+    *bsd*)
+	printf "%s\n" "#define PJ_BSD 1" >>confdefs.h
+
+	;;
     *rtems*)
 	printf "%s\n" "#define PJ_RTEMS 1" >>confdefs.h
 
@@ -5725,9 +5729,6 @@ case $target in
     *)
 	;;
 esac
-
-
-
 
 # Check whether --enable-floating-point was given.
 if test ${enable_floating_point+y}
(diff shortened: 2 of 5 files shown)


pjlib/src/pj/os_core_unix.c
```diff
@@ -616,6 +620,37 @@ pj_status_t pj_thread_init(void)
...
+#if defined(PJ_HAS_PTHREAD_SETNAME_NP) && PJ_HAS_PTHREAD_SETNAME_NP != 0
+#   if defined(PJ_DARWINOS) && PJ_DARWINOS != 0
+    pthread_setname_np(name);
+#   else
+    pthread_setname_np(pthread_self(), name);
+#   endif
+#elif defined(PJ_HAS_PTHREAD_SET_NAME_NP) && PJ_HAS_PTHREAD_SET_NAME_NP != 0
+    pthread_set_name_np(pthread_self(), name);
+#else
+#   warning "OS not support set thread dispaly name"
```
kevinkouketsu, line 650: typo

pjlib/src/pj/os_core_unix.c
```diff
@@ -616,6 +620,37 @@ pj_status_t pj_thread_init(void)
 }
 
 #if PJ_HAS_THREADS
+
+/*
+ * Set current thread display name
+ * This can be useful for debugging, as the name is displayed in thread status
+ * You can list a process's threads status on linux/freebsd
+ * use cmd: "top -H -p pid"
```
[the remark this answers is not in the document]
jimying: Already fix it,  thanks

jimying:
Win32 has supported （According to the doc  [ how-to-set-a-thread-name-in-native-code]）(https://docs.microsoft.com/en-us/visualstudio/debugger/how-to-set-a-thread-name-in-native-code?view=vs-2022)

I have tested  linux/freebsd,  hope someone can help test win32 (I have no environment).

PS: Thanks [andreas-wehrmann](https://github.com/andreas-wehrmann) recommand to add win32: SetThreadDescription() support.

andreas-wehrmann:
About win32:
I tested and found the SetThreadDescription part to be not working as expected (MSVS2019 Debugger is showing gibberish in the thread name column). The setting the name by throwing an exception part seems to work.

SetThreadDescription always expects a Unicode string, so you need to modify the initial part of set_thread_display_name():

```
static void set_thread_display_name(const char *name)
{
    wchar_t thread_wname[PJ_MAX_OBJ_NAME];

    /* Set thread name by SetThreadDescription (if support) */
    FnSetThreadDescription fn = (FnSetThreadDescription)GetProcAddress(
	GetModuleHandle(PJ_T("Kernel32.dll")), "SetThreadDescription");
    PJ_LOG(5, (THIS_FILE, "SetThreadDescription:%p, name:%s", fn, name));

    if (fn) {
        pj_ansi_to_unicode(name, 
                           pj_ansi_strlen(name),
                           thread_wname, PJ_MAX_OBJ_NAME);
	fn(GetCurrentThread(), thread_wname);
	return;
    }

    /* Set thread name by throwing an exception */
#if defined(__MINGW32__) || defined(__CYGWIN__)
#pragma message("Warning: Not support exception")
    return;
#else
    // The debugger needs to be around to catch the name in the exception.
    // If there isn't a debugger, we are needlessly throwing an exception.
    if (!IsDebuggerPresent()) {
	return;
    }

    const DWORD MS_VC_EXCEPTION = 0x406D1388;
    THREADNAME_INFO info;
    info.dwType = 0x1000;
    info.szName = name;
    info.dwThreadID = (DWORD)-1;
    info.dwFlags = 0;
#pragma warning(push)
#pragma warning(disable : 6320 6322)
    __try {
	RaiseException(MS_VC_EXCEPTION, 0, sizeof(info) / sizeof(ULONG_PTR),
		       (ULONG_PTR *)&info);
    } __except (EXCEPTION_EXECUTE_HANDLER) {
    }
#pragma warning(pop)
#endif
}
```

With the modification above you need to add another include at the beginning of the file:
`#include <pj/unicode.h>`


With the modification above, it works:
thread_names

Note: I tested this with Visual Studio 2019 in Debug/Release modes for Win32 and x64 build.

jimying:
@andreas-wehrmann 
PJ_NATIVE_STRING_IS_UNICODE is detected by  VC macro UNICODE
```c
// os_win32.h
#ifdef UNICODE
#   define PJ_NATIVE_STRING_IS_UNICODE    1
#else
#   define PJ_NATIVE_STRING_IS_UNICODE    0
#endif
```
could you set/unset  macro UNICODE and test again?

andreas-wehrmann:
So, I'm back again.
I switched all projects to Unicode and tested again, seems to work fine
because the thread names are always defined and passed as ANSI strings.

I did notice another issue when switching to Unicode though, but that is unrelated to this PR.
I'll open a PR/issue when depending on when I find the time for it...

sauwming (approved):

pjlib/src/pj/os_core_unix.c
```diff
@@ -616,6 +620,37 @@ pj_status_t pj_thread_init(void)
...
+#if defined(PJ_HAS_PTHREAD_SETNAME_NP) && PJ_HAS_PTHREAD_SETNAME_NP != 0
+#   if defined(PJ_DARWINOS) && PJ_DARWINOS != 0
+    pthread_setname_np(name);
+#   else
+    pthread_setname_np(pthread_self(), name);
+#   endif
+#elif defined(PJ_HAS_PTHREAD_SET_NAME_NP) && PJ_HAS_PTHREAD_SET_NAME_NP != 0
+    pthread_set_name_np(pthread_self(), name);
+#else
+#warning "OS not support set thread dispaly name"
```
andreas-wehrmann, line 649: Typo here

nanangizz (approved):

silentindark:
As wrote @andreas-wehrmann that here is a typo
@jimying Could you please update commit
"OS not support set thread **dispaly** name" -> "OS not support set thread **display** name"

Issue #3155, opened 2022-06-21
ioqueue_epoll  add udp fd to epoll with flag EPOLLONESHOT,sometimes will not have read event


wahjrzd:
### Describe the bug

epoll with multithread

file : ioqueue_common_abs.c 
function：ioqueue_dispatch_read_event

pj_status_t rc;

    /* Try lock the key. */
    rc = pj_ioqueue_trylock_key(h);
    if (rc != PJ_SUCCESS) {
	return PJ_FALSE;
    }
if rc != PJ_SUCCESS, when we return ,wo need do this:
  rc = pj_ioqueue_trylock_key(h);
    if (rc != PJ_SUCCESS) {
          ioqueue_add_to_set(ioqueue, h, READABLE_EVENT);
	return PJ_FALSE;
    }
next loop, wo can continue handle read events

### Steps to reproduce

short time send a large number sip message

### PJSIP version

2.11.1

### Context

null

### Log, call stack, etc

```shell
null
```

jimying:
#3018 fixed it and already merge to master branch
Get the lastest code and test again.

wahjrzd:
> #3018 should fixed it and already merged into master branch Get the lastest code and test again.

udp socket  test success,but when i use it as tcpserver ,wo connect it ,and then close connect ,then application will crash
bt like this
#2  0x00007f03b45ab1a6 in __assert_fail_base () from /lib64/libc.so.6
#3  0x00007f03b45ab252 in __assert_fail () from /lib64/libc.so.6
#4  0x00000000005fe3e3 in grp_lock_acquire ()
#5  0x00000000005f88f1 in pj_ioqueue_poll ()
#6  0x00000000005db047 in pjsip_endpt_handle_events2 ()
#7  0x00000000004b7aea in SipUAC::EventRun (this=0x1839fa0) at SipUAC.cpp:583

jimying:
Should rebuild with debug flags
steps:
1. Rebuild  pjsip lib with debug flag
```
export CFLAGS="-g"
./configure
make dep;make
```
2. Rebuild your app(c++)  with CPPFLAGS = "-g"

3. gdb again, when crash, supply the stack info



wahjrzd ***@***.***> 于2022年6月23日周四 11:12写道：

wahjrzd:
Program terminated with signal 11, Segmentation fault.
#0&nbsp; 0x0000000000611f98 in pj_pool_release (pool=0x0) at ../include/pj/pool_i.h:102
[...]

PR #3158, opened 2022-06-24
Fix epoll-oneshot may cause crash (use released key)

jimying:
For detail , Please see Issue #3155 , thanks wahjrzd's feedback

ioqueue_add_to_set() should locked by pj_grp_lock_add_ref_dbg/pj_grp_lock_dec_ref_dbg in this case.

### DIFF
--- a/pjlib/src/pj/ioqueue_epoll.c
+++ b/pjlib/src/pj/ioqueue_epoll.c
@@ -831,6 +831,14 @@ PJ_DEF(int) pj_ioqueue_poll( pj_ioqueue_t *ioqueue, const pj_time_val *timeout)
 	    }
 	    continue;
 	}
+
+	if (ioqueue->use_epolloneshot) {
+	    /* We are not processing this event, but we still need to rearm
+	     * to receive future events.
+	     */
+	    if (!IS_CLOSING(h))
+		ioqueue_add_to_set(ioqueue, h, 0);
+	}
     }
     for (i=0; i<event_cnt; ++i) {
 	if (queue[i].key->grp_lock)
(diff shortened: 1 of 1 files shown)


pjlib/src/pj/ioqueue_epoll.c
```diff
@@ -819,6 +819,24 @@ PJ_DEF(int) pj_ioqueue_poll( pj_ioqueue_t *ioqueue, const pj_time_val *timeout)
...
+	 * a new event mask.
+	 */
+	pj_ioqueue_lock_key(queue[i].key);
+	if (!IS_CLOSING(queue[i].key))
+	    ioqueue_add_to_set(ioqueue, queue[i].key, 0);
+	pj_ioqueue_unlock_key(queue[i].key);
+#endif
+
 #if PJ_IOQUEUE_HAS_SAFE_UNREG
 	decrement_counter(queue[i].key);
```
jimying, line 841: line: 823   key may be released, and use it later

pjlib/src/pj/ioqueue_epoll.c
```diff
@@ -819,6 +819,24 @@ PJ_DEF(int) pj_ioqueue_poll( pj_ioqueue_t *ioqueue, const pj_time_val *timeout)
 	    }
 	}
 
+#if USE_EPOLLONESHOT
```
sauwming, line 822: Thanks for the patch and also to @wahjrzd for the report.
But there may be a scenario when the epoll fails to be rearmed, i.e. as written in the first case of the note, when we do not wish to process a reported event (event_cnt < count). That's why initially the rearm was done outside the `event_cnt` loop (which unfortunately, caused the reported bug).

I haven't had any idea on how to fix it though.

sauwming: I had a chance to look at this again, and I think the fix looks good. But we also need to rearm the epoll in the scenario that we're not processing the event as mentioned above. Here's the additional patch (to be applied on top of this one):
[epoll_patch.txt](https://github.com/pjsip/pjproject/files/9036411/epoll_patch.txt)

jimying: Already merge the patch ,  exclude closing fd.

sauwming:
I've merged #3151, so please merge with the latest master. Thanks.

jimying:
> I've merged #3151, so please merge with the latest master. Thanks.

Already merged

sauwming (approved):

nanangizz (approved):

PR #3165, opened 2022-06-28
Update VAD docs/help

nanangizz:
The pjsua app help & PJSUA/PJSUA2 docs assume that VAD is always enabled for all codecs by default, while actually some codecs (e.g: Opus & SILK) have VAD disabled by default.

Thank you Liviu Andron for the report.

### DIFF
--- a/pjsip-apps/src/pjsua/pjsua_app_config.c
+++ b/pjsip-apps/src/pjsua/pjsua_app_config.c
@@ -147,8 +147,8 @@ static void usage(void)
     puts  ("  --auto-rec          Automatically record conversation");
     puts  ("  --quality=N         Specify media quality (0-10, default="
     				  xstr(PJSUA_DEFAULT_CODEC_QUALITY) ")");
-    puts  ("  --ptime=MSEC        Override codec ptime to MSEC (default=specific)");
-    puts  ("  --no-vad            Disable VAD/silence detector (default=specific)");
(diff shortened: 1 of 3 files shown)


pjsip-apps/src/pjsua/pjsua_app_config.c
```diff
@@ -148,7 +148,7 @@ static void usage(void)
     puts  ("  --quality=N         Specify media quality (0-10, default="
     				  xstr(PJSUA_DEFAULT_CODEC_QUALITY) ")");
     puts  ("  --ptime=MSEC        Override codec ptime to MSEC (default=specific)");
-    puts  ("  --no-vad            Disable VAD/silence detector (default=vad enabled)");
+    puts  ("  --no-vad            Disable VAD/silence detector (default=specific)");
```
sauwming, line 151: if there is room to put "codec specific", better put the complete description, IMO (also for --ptime above).

sauwming (approved):

trengginas (approved):

pjsip-apps/src/pjsua/pjsua_app_config.c
nanangizz: Actually the VAD setting in PJSUA affects both codec & sound device (PJSUA will add/remove `PJMEDIA_AUD_DEV_CAP_VAD` flag). But as VAD in audio device seems to only applicable to non-PCM mode (e.g: Symbian APS direct), so more or less it is codec related too.

Issue #3160, opened 2022-06-26
Add ioqueue wakeup mechanism


jimying:
**Why add this feature?**

Suppose ioqueue poll timeout is 5 seconds

Cases:
  1. For ioqueue-select framework. When register a new socket to ioqueue, poll
 maybe blocked waiting, the new socket's events can't be real-time caught.
 Because the new socket has not added to select's read set.
 It'll be delay added to select read set until  exit waiting state and poll again.
 
 3. For ioqueue all frameworks. when exit app, poll maybe blocked waiting, it'll block until timeout.

 5. For ioqueue all frameworks.  Add a new timer, which's timeout value is
 small. If the ioqueue poll blocked waiting, the timer will
 be trigged more later than expected.

**Describe the solution you'd like**

Wakeup mechanism can solve above cases' problem, it will send
event to ioqueue and make it exit blocked waiting state immediately. <br>

Wakeup mechanism how to work:
When create ioqueue,  
use socketpair() to create  two sockets(one for read, the other for write),
and register the read socket to ioqueue.
When need wakeup the ioqueue,   write 1 byte to the write-side socket 
which will tigger the read-side socket to read.

** related **/
#3086 , if has wakup mechanism,  no need set poll timeout to a small value.

<br>

If you agree it, i'll find a time to create a new PR for this feature.

sauwming:
Thanks for the suggestion, as well as very quickly (!) coming up with the implementation in #3086. We had a similar dream from a long time ago in #1125, which we eventually close since we didn't have time to implement.

The socket pair idea seems to be working fine, although I'm not sure if it's justifiable to use double the amount of the socket resources (i.e. `2*n`) just for the purpose of immediate wakeup. But since it's the only solution we have right now, we really appreciate it and hope you can keep it available for everyone so they can test it.

For now, I think no action is needed from your side. Let's wait first to hear from others if they require the wakeup feature and whether the socket pair mechanism is sufficient for their requirement.

Thanks again.

jimying:
@sauwming 
I have implement this feature, please help review it.

PR #3172, opened 2022-07-06
Add ioqueue wakeup mechanism

jimying:
Why add this feature, please go to #3160 for detail.

Set macro PJ_IOQUEUE_HAS_WAKEUP =1 to enable feature.

If enable wakeup-mechanism, the ioqueue poll timeout can be a large value ( eg. 9 seconds),
which can reduce cpu usage to 0% in idle state.

I have already added this feature support in pjsip and pjmedia modules,  so can test it use pjsua app.

PS:
This feature is supported in ioqueue frameworks select/epoll/kqueue,  
other ioqueue frameworks uwp/winnt/symbian not support now.

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -6220,6 +6220,32 @@ printf "%s\n" "no" >&6; }
 fi
 rm -f core conftest.err conftest.$ac_objext conftest.beam conftest.$ac_ext
 
+{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking if socketpair() is available" >&5
+printf %s "checking if socketpair() is available... " >&6; }
+cat confdefs.h - <<_ACEOF >conftest.$ac_ext
+/* end confdefs.h.  */
+#include <sys/types.h>
+				     #include <sys/socket.h>
+int
+main (void)
+{
+socketpair(0, 0, 0, 0);
+  ;
+  return 0;
+}
+_ACEOF
+if ac_fn_c_try_compile "$LINENO"
+then :
+  printf "%s\n" "#define PJ_SOCK_HAS_SOCKETPAIR 1" >>confdefs.h
+
+		   { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: yes" >&5
+printf "%s\n" "yes" >&6; }
+else $as_nop
+  { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: no" >&5
+printf "%s\n" "no" >&6; }
+fi
+rm -f core conftest.err conftest.$ac_objext conftest.beam conftest.$ac_ext
+
 { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking if sockaddr_in has sin_len member" >&5
 printf %s "checking if sockaddr_in has sin_len member... " >&6; }
 cat confdefs.h - <<_ACEOF >conftest.$ac_ext
--- a/aconfigure.ac
+++ b/aconfigure.ac
@@ -344,6 +344,15 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <sys/types.h>
 		   AC_MSG_RESULT(yes)],
 		  [AC_MSG_RESULT(no)])
 
+dnl # Determine if socketpair() is available
+AC_MSG_CHECKING([if socketpair() is available])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <sys/types.h>
+				     #include <sys/socket.h>]],
+		    		  [socketpair(0, 0, 0, 0);])],
+		  [AC_DEFINE(PJ_SOCK_HAS_SOCKETPAIR,1)
+		   AC_MSG_RESULT(yes)],
+		  [AC_MSG_RESULT(no)])
+
 dnl # Determine if sockaddr_in has sin_len member
 AC_MSG_CHECKING([if sockaddr_in has sin_len member])
 AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <sys/types.h>
--- a/pjlib/include/pj/compat/os_auto.h.in
+++ b/pjlib/include/pj/compat/os_auto.h.in
@@ -91,6 +91,7 @@
 #undef PJ_SOCK_HAS_INET_PTON
 #undef PJ_SOCK_HAS_INET_NTOP
 #undef PJ_SOCK_HAS_GETADDRINFO
+#undef PJ_SOCK_HAS_SOCKETPAIR
 
 /* On these OSes, semaphore feature depends on semaphore.h */
 #if defined(PJ_HAS_SEMAPHORE_H) && PJ_HAS_SEMAPHORE_H!=0
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -682,6 +682,15 @@
 #   define PJ_IOQUEUE_MAX_HANDLES	(64)
 #endif
 
+/**
+ * Enable ioqueue wakeup mechanism
+ *
+ * Default: 0
+ */
+
+#ifndef PJ_IOQUEUE_HAS_WAKEUP
+#   define PJ_IOQUEUE_HAS_WAKEUP 	0
+#endif
 
 /**
  * If PJ_IOQUEUE_HAS_SAFE_UNREG macro is defined, then ioqueue will do more
--- a/pjlib/include/pj/ioqueue.h
+++ b/pjlib/include/pj/ioqueue.h
@@ -362,6 +362,17 @@ PJ_DECL(pj_status_t) pj_ioqueue_create( pj_pool_t *pool,
  */
 PJ_DECL(pj_status_t) pj_ioqueue_destroy( pj_ioqueue_t *ioque );
 
+#if PJ_IOQUEUE_HAS_WAKEUP
+/**
+ * Wakeup the I/O queue.
+ *
+ * @param ioque	        The I/O Queue to be wakeuped
+ *
+ * @return              PJ_SUCCESS if success.
+ */
+PJ_DECL(pj_status_t) pj_ioqueue_wakeup( pj_ioqueue_t *ioque );
+#endif
+
 /**
  * Set the lock object to be used by the I/O Queue. This function can only
  * be called right after the I/O queue is created, before any handle is
--- a/pjlib/include/pj/sock.h
+++ b/pjlib/include/pj/sock.h
@@ -1527,6 +1527,25 @@ PJ_DECL(pj_status_t) pj_sock_shutdown( pj_sock_t sockfd,
 PJ_DECL(char *) pj_addr_str_print( const pj_str_t *host_str, int port, 
 				   char *buf, int size, unsigned flag);
 
+/**
+ * Create socket pair
+ * @param family    Specifies a communication domain; this selects the
+ *		    protocol family which will be used for communication.
+ * @param type	    The socket has the indicated type, which specifies the 
+ *		    communication semantics.
+ * @param protocol  Specifies  a  particular  protocol  to  be used with the
+ *		    socket.  Normally only a single protocol exists to support 
+ *		    a particular socket  type  within  a given protocol family, 
+ *		    in which a case protocol can be specified as 0.
+ * @param sv	    The new sockets vector
+ *
+ * @return	    Zero on success.
+ *
+ */
+PJ_DECL(pj_status_t) pj_sock_socketpair(int family,
+				    int type,
+				    int protocol,
+				    pj_sock_t sv[2]);
 
 /**
  * @}
--- a/pjlib/include/pj/timer.h
+++ b/pjlib/include/pj/timer.h
@@ -183,6 +183,16 @@ PJ_DECL(pj_status_t) pj_timer_heap_create( pj_pool_t *pool,
  */
 PJ_DECL(void) pj_timer_heap_destroy( pj_timer_heap_t *ht );
 
+#if PJ_IOQUEUE_HAS_WAKEUP
+/**
+ * Bind the timer heap to an ioqueue
+ *
+ * @param ht        The timer heap.
+ * @param ioq       The ioqueue to bind.
+ *
+ */
+PJ_DECL(void) pj_timer_heap_bind(pj_timer_heap_t *ht, pj_ioqueue_t *ioq);
+#endif
 
 /**
  * Set lock object to be used by the timer heap. By default, the timer heap
--- a/pjlib/src/pj/config.c
+++ b/pjlib/src/pj/config.c
@@ -74,6 +74,7 @@ PJ_DEF(void) pj_dump_config(void)
     PJ_LOG(3, (id, " ioqueue type              : %s", pj_ioqueue_name()));
     PJ_LOG(3, (id, " PJ_IOQUEUE_MAX_HANDLES    : %d", PJ_IOQUEUE_MAX_HANDLES));
     PJ_LOG(3, (id, " PJ_IOQUEUE_HAS_SAFE_UNREG : %d", PJ_IOQUEUE_HAS_SAFE_UNREG));
+    PJ_LOG(3, (id, " PJ_IOQUEUE_HAS_WAKEUP     : %d", PJ_IOQUEUE_HAS_WAKEUP));
     PJ_LOG(3, (id, " PJ_HAS_THREADS            : %d", PJ_HAS_THREADS));
     PJ_LOG(3, (id, " PJ_LOG_USE_STACK_BUFFER   : %d", PJ_LOG_USE_STACK_BUFFER));
     PJ_LOG(3, (id, " PJ_HAS_SEMAPHORE          : %d", PJ_HAS_SEMAPHORE));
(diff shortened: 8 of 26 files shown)


jimying:
I have added wakeup-mechanism in pjsua-lib and pjmedia , poll timeout set to 9 seconds.
I test on macos, everything works fine. 
Idle CPU usage  has been reduced a lot.

jimying:
Fix a bug: kqueue timeout convert ms to ns error.
Also merge here.

bennylp:
Thanks for the PR. I've come across it and I think I have some general comments. I'll split them as two separate comments.

**`pj_sock_socketpair()` comments**

1. It needs unit-testing in pjlib-test, testing IPv4/IPv6 vs TCP/UDP combinations, and maybe add AF_UNIX as well where available if the API claims to emulate Unix's socketpair 
2. In the implementation (`sock_bsd.c`), instead of `#if defined(WIN32)..`, better use `PJ_SOCK_HAS_SOCKETPAIR` macro and detect this in `configure` script (similar to `PJ_SOCK_HAS_INET_PTON` detection).
3. some codes (such as pjlib-test) should be modified to make use of this new feature. Note that pjlib-test/util.c already has `app_socketpair() function`, this could be replaced by the new API.
4. adding these will make the changes pretty big hence it's better to put them as separate PR.

bennylp:
For the **wakeup mechanism**:
1. I'm not sure why the timer heap needs to be aware of the ioqueue wake-up mechanism? Sorry if I missed something, I haven't read the patch in very detail, but I find this a bit weird intuitively. And IMO this coupling is not acceptable (at least currently. We also have other problems where we think having the timer heap polls the ioqueue could solve the problem, but we tried to avoid it as well).
2. ioqueue should not treat the wake-up sockets differently than usual sockets. For all it cares, they are just usual sockets registered to it. It's interaction should be limited to creating/destroying them, calling the wakeup during register/unregister. It should not need to do special processing for it during poll() for example.
3. I tend to prefer having the feature as run-time configurable nowadays, for flexibility and to reduce the complexity of the `ifdefs` in the source code. I propose to have this feature run-time configurable as well.

P.S. I'm also currently working on the ioqueue (especially epoll), maybe there will be PRs coming from me as well.

bennylp:
(accidentally closed the PR, reopening..)

jimying:
@bennylp 


1. agree:  pj_sock_socketpair()   use PJ_SOCK_HAS_SOCKETPAIR   replace PJ_WIN32/PJ_WIN64
2. agree:  run-time enable/disable wakeup feature  replace build by macro PJ_IOQUEUE_HAS_WAKEUP
3.  why wakeup ioqueue in timer?
>1. I'm not sure why the timer heap needs to be aware of the ioqueue wake-up mechanism? Sorry if I missed something, I 
haven't read the patch in very detail, but I find this a bit weird intuitively. And IMO this coupling is not acceptable (at least currently. We also have other problems where we think having the timer heap polls the ioqueue could solve the problem, 
but we tried to avoid it as well).

If ioqueue poll is in blocking state, eg. timeout is 5 seconds,
currently add a timer with small delay 1 second. If no wakeup ioqueue,
this timer will be triggered 4 seconds later as expected

4. why wakeup ioqueue in select framework?
>2. ioqueue should not treat the wake-up sockets differently than usual sockets. For all it cares, they are just usual sockets registered to it. It's interaction should be limited to creating/destroying them, calling the wakeup during register/unregister. It should not need to do special processing for it during poll() for example.

When register a new socket to ioqueue, poll maybe blocked waiting, 
the new socket's events can't be real-time caught.
Because the new socket has not added to select's read set.
It'll be delay added to select read set until exit waiting state and poll again.

The epoll/kqueue framework no this problem.

bennylp:
> 3. why wakeup ioqueue in timer?
>
> If ioqueue poll is in blocking state, eg. timeout is 5 seconds, 
> Add a timer with small delay 1 second. If no wakeup ioqueue,
> this timer will be triggered 4 seconds later as expected
> 
> So when schedule a timer, if the timer is smallest in the heap,
> should wakeup ioqueue to poll again.

Sorry I still don't understand this, can you explain more? In any case, coupling the timer heap with the ioqueue is not a good idea. But I'm curious about the use case why do you think this is necessary.

bennylp:
> 7\. why wakeup ioqueue in select framework when register socket?
> 
> 
> > 2. ioqueue should not treat the wake-up sockets differently than usual sockets. For all it cares, they are just usual sockets registered to it. It's interaction should be limited to creating/destroying them, calling the wakeup during register/unregister. It should not need to do special processing for it during poll() for example.
> 
> When register a new socket to ioqueue, poll maybe blocked waiting, the new socket's events can't be real-time caught. Because the new socket has not added to select's read set. It'll be delay added to select read set until exit waiting state and poll again. (The epoll/kqueue framework no this problem.)
> 
> I also wakeup ioqueue now when unregister socket from select ioqueue, i think it is optional, can remove the logic.

Yes I understood why the wakeup mechanism is needed.

What I meant was the `pj_ioqueue_poll()` should not need to be changed.

bennylp:
Reading the patch in more detail, I have some more comments regarding the wakeup mechanism.

Since this feature only applies to `select_ioqueue`, I propose making this functionality internal to `select_ioqueue`, and just enable it by default. In more detail, I propose the following:
1. rename `PJ_IOQUEUE_HAS_WAKEUP` to `PJ_SELECT_IOQUEUE_HAS_WAKEUP`. This controls whether wakeup mechanism should be enabled at runtime in `ioqueue_select`.
2. the code itself is added in `ioeueue_select.c` without `#ifdef..` (we already have too many #ifdefs)
3. therefore (hopefully) there is no need to have modifications in other ioqueue backends.

Thanks. What others think about this?

bennylp (changes requested):
Proposing general modifications as mentioned in PR comments.

jimying:
> > 3. why wakeup ioqueue in timer?
> > 
> > If ioqueue poll is in blocking state, eg. timeout is 5 seconds,
> > Add a timer with small delay 1 second. If no wakeup ioqueue,
> > this timer will be triggered 4 seconds later as expected
> > So when schedule a timer, if the timer is smallest in the heap,
> > should wakeup ioqueue to poll again.
> 
> Sorry I still don't understand this, can you explain more? In any case, coupling the timer heap with the ioqueue is not a good idea. But I'm curious about the use case why do you think this is necessary.


Eg. pjsip_endpt_handle_events2() logic， every time call handle_events2() steps:
step1:  pj_timer_heap_poll(),  expired timers triggered, and get the smallest timeout (t1: 5seonds) in timer heap
step2: compare t1 with ioqueue_poll timeout (t2: 9 seconds), get the smaller value t3
step3: ioqueue_poll() use the timeout t3(5 seconds) to poll

if  ioqueue_poll() is in blocing state (no any socket events),  t3
now add a new timer with delay timeout  1 seconds  to timer heap
what will happen:  
after 5 seconds,  ioqueue_poll() timeout exit blocking state,  and call pjsip_endpt_handle_events2() again,
step1 will trigger expired timers,
this timer is trigged more delay( ~5seconds form add to timer heap) than excepted (1 second).

If add timer to timer heap, check that this timer (1 seond) is the smallest,   wakeup ioqueue,
call pjsip_endpt_handle_events2() again,   step3 poll timeout t3 will set to 1 seond

jimying:
> > 7. why wakeup ioqueue in select framework when register socket?
> > > 2. ioqueue should not treat the wake-up sockets differently than usual sockets. For all it cares, they are just usual sockets registered to it. It's interaction should be limited to creating/destroying them, calling the wakeup during register/unregister. It should not need to do special processing for it during poll() for example.
> > 
> > 
> > When register a new socket to ioqueue, poll maybe blocked waiting, the new socket's events can't be real-time caught. Because the new socket has not added to select's read set. It'll be delay added to select read set until exit waiting state and poll again. (The epoll/kqueue framework no this problem.)
> > I also wakeup ioqueue now when unregister socket from select ioqueue, i think it is optional, can remove the logic.
> 
> Yes I understood why the wakeup mechanism is needed.
> 
> What I meant was the `pj_ioqueue_poll()` should not need to be changed.

you mean this?
```c
#if PJ_IOQUEUE_HAS_WAKEUP && PJ_HAS_THREADS
    /*
     * Make sure that wakeupfd always be added to rfdset
     * When multi-thread polling,  wakeupfd may has been removed
     * (ioqueue_remove_from_set) by other thread
     */
    if (ioqueue->wakeup_fd[0] != PJ_INVALID_SOCKET) {
	PJ_FD_SET(ioqueue->wakeup_fd[0], &rfdset);
    }
#endif
```

Suppose no above logic code，logic simulate use pjsua:
app:pjsua  start 4 threads to poll  select ioqueue,  and poll timeout is 9 seonds.
when exit app, will wakeup ioqueue:
```c
#if PJ_IOQUEUE_HAS_WAKEUP
    /* Wakeup ioqueue, avoid block long time */
    for (i = 0; i < (int)pjsua_var.ua_cfg.thread_cnt; ++i) {
	pj_ioqueue_t *ioq = pjsip_endpt_get_ioqueue(pjsua_var.endpt);
	pj_ioqueue_wakeup(ioq);
    }
#endif
```
It will notify ioqueue 4 times,   if  one thread ioqueue_poll() exit blocking state and start processing wakefd read events,
it will call ioqueue_remove_from_set()  remove wakefd from read fdset first
If at this time, other three threads  ioqueue_poll() timeout and poll again,  wakefd will not in select read fdset.

The 1st thread will exit,  and other three threads will  after 9 seconds timeout ,and select wakefd again.
App exit will block a long time.

epoll / kqueue, no this problem, because  socket  read event is always listened.

jimying:
>I tend to prefer having the feature as run-time configurable nowadays, for flexibility and to reduce the complexity of the ifdefs in the source code. I propose to have this feature run-time configurable as well.

Need change a lot of code. 
ioqueue.h:  pj_ioqueue_create() need set wakeup flag
pjsip: pjsip_endpt_create()  need set wakeup flag
pjmedia: pjmedia_endpt_create() need set wakeup flag
pjsua app:  need add command line argument: --ioqueue-wakeup to enable  ioqueue wakeup for testing

Compatibility also need to be considered, Maybe this stage should not do it.

jimying:
I have done some changes in pjlib-test, so that it can test ioqueue when enable wakeup.

bennylp:
Thanks for the patch. Re: new `pj_sock_socketpair()`, could you please create separate PR, as that would be easier to accept. I also see that you haven't addressed my previous comments re: socketpair above (the ones numbered 1-4), I assume you will be addressing them in the new PR. I will code-review in detail in that PR. 

In the meantime, I have some more comments (re: socketpair only. I will comment about wake-up separately):

1. socketpair API should work with INET6 as well.
2. our style is to use `goto on_error` for error conditions since that makes it easier to follow. It would be great if you could adjust the coding style in the patch so they are consistent with existing code base.
3. the indentation is a bit off with the parameters of pj_sock_socketpair() function (I apologize for our weird indentation setting, but it is what it is unfortunately).
4. also I will tell you a "secret" convention in our code :). For code comments, we use `/*.. */` style comment. When we use `//..` style comment, we use it when we want to disable some code (sometimes instead of deleting broken codes, we prefer to disable it so that we don't forget about that in the future, and this is the only case where we use the `//..` style comment, except in "less official" places such as sample apps when sometimes we forgot or too lazy to follow this convention :).

jimying:
@bennylp  I will create an new PR for pj_sock_socketpair()
linux man page for socketpair say:
```
NOTES
       On Linux, the only supported domain for this call is AF_UNIX (or synonymously, AF_LOCAL).  
(Most implementations have the same restriction.)
```
So when PJ_SOCK_HAS_SOCKETPAIR = 0,  we simulate unix-like socketpair(),  do  i need support INET6?

bennylp:
Re: wakeup mechanism.

In general, I propose to split the wakeup feature into two stages:

1. Add wakeup mechanism to solve problems with long waiting time when registering socket to ioqueue and when quitting application because ioqueue is blocked in `pj_ioqueue_poll()`
2. More general case to make PJSIP uses less CPU for polling by increasing the polling time, or maybe even setting the polling time to infinity (thus using zero CPU)

Stage 1 above is a lot simpler as it only deals with ioqueue and pjsua-lib (when exiting), so it should be easier to accept.

Once we have it in place, we could move to stage 2, which is arguably more "fun" indeed! For this stage though, it would be great if we could measure the benefits of the proposed approach w.r.t power usage, for example.

Some more detailed comments below.

> wakeu mechanism logic:
> If add timer to timer heap, check that this timer (1 seond) is the smallest, wakeup ioqueue,
> call pjsip_endpt_handle_events2() again, step3 poll timeout t3 will set to 1 seond, timer tiggered ~ as expected

Okay understood. My general intuition is that this *"more intelligent ioqueue polling time/timer schedule"* should be implemented above both ioqueue and the timer. Just as we have *"intelligent ioqueue polling timeout"* calculation in `pjsip_endpt_handle_events2()`, maybe we could modify `pjsip_endpt_schedule_timer()` to take into account ioqueue blocking time there as well. I understand that will would require implementing similar mechanism in `pjnath` as well, but IMO that's the right approach. (P.S. with my proposal above, this would belong to stage 2)

> you mean this?
> 
> ```c
> #if PJ_IOQUEUE_HAS_WAKEUP && PJ_HAS_THREADS
>     /*
>      * Make sure that wakeupfd always be added to rfdset
>      * When multi-thread polling,  wakeupfd may has been removed
>      * (ioqueue_remove_from_set) by other thread
>      */
>     if (ioqueue->wakeup_fd[0] != PJ_INVALID_SOCKET) {
> 	PJ_FD_SET(ioqueue->wakeup_fd[0], &rfdset);
>     }
> #endif
> ```

Yes.

> It will notify ioqueue 4 times, if one thread ioqueue_poll() exit blocking state and start processing wakefd read event,
> it will call ioqueue_remove_from_set() remove wakefd from read fdset first
> If at this time, other three threads ioqueue_poll() timeout and poll again, wakefd will not in select read fdset.
> The 1st thread will exit, and other three threads will after 9 seconds timeout ,and select wakefd again.
> App exit will block a long time.

Usually the polling loop is protected by `!quit` flag, something like this:
```
  while (!quit) {
    pj_ioqueue_poll(..);
  }
```
so I think this shouldn't be a problem.

> > 1. rename `PJ_IOQUEUE_HAS_WAKEUP` to `PJ_SELECT_IOQUEUE_HAS_WAKEUP`
> 
> epoll/kqueue also need wakeup mechanism,reason:

Okay understood.


> > I tend to prefer having the feature as run-time configurable nowadays, for flexibility and to reduce the complexity of the ifdefs in the source code. I propose to have this feature run-time configurable as well.
> 
> Need change a lot of code.

I'm thinking once we have this feature, we should just enable it by default. In the (near) future, maybe we should have new API to create ioqueue with additional options. Such option is also useful to enable/disable EPOLLEXCLUSIVE/EPOLLONESHOT at run-time, which is useful at least for testing purposes. I'm working on this.

Thanks for working on this issue.

jimying:
@bennylp 
1.  New  PR for api pj_sock_socketpair()  #3179 
2.  select problem
```c++
#if PJ_IOQUEUE_HAS_WAKEUP && PJ_HAS_THREADS
    /*
     * Make sure that wakeupfd always be added to rfdset
     * When multi-thread polling,  wakeupfd may has been removed
     * (ioqueue_remove_from_set) by other thread
     */
    if (ioqueue->wakeup_fd[0] != PJ_INVALID_SOCKET) {
	PJ_FD_SET(ioqueue->wakeup_fd[0], &rfdset);
    }
#endif
```

>Usually the polling loop is protected by !quit flag, something like this:
  while (!quit) {
    pj_ioqueue_poll(..);
  }
so I think this shouldn't be a problem.

This block code should not remove, I have tested it. If  comment out the block code,  4 threads app will block long time.

jimying:
> Okay understood. My general intuition is that this "more intelligent ioqueue polling time/timer schedule" should be implemented above both ioqueue and the timer. Just as we have "intelligent ioqueue polling timeout" calculation in pjsip_endpt_handle_events2(), maybe we could modify pjsip_endpt_schedule_timer() to take into account ioqueue blocking time there as well. I understand that will would require implementing similar mechanism in pjnath as well, but IMO that's the right approach. (P.S. with my proposal above, this would belong to stage 2)

Agree with timer and ioqueue should decoupling

I think that can add a new callback member in  struct pj_timer_entry,  when pj_timer_heap_schedule(),
if the timer is the smallest in the heap,  call the callback.

jimying:
@bennylp 

For timer imer and ioqueue  decoupling, Need change a lot of code (Any modules use timer need do  changes)
 
Current coupling method (pj_timer_heap_bind())，all code changes  limited to timer.c,  ohter modules no need change any code. ^_^

To my point,  coupling  method maybe better.

bennylp:
Sorry for the long delay. And thanks for working on this topic. Now we know in more detail what it takes, if we want "zero poll" using this approach. 

But there are other approaches, for example: 1) by making each timer and ioqueue "zero poll" in the first place (i.e. each has its own thread). I think this is quite promising, as it mimics what some platforms are doing (e.g. the late Symbian) 2) by creating another object that bundles timer and ioqueue, and then make this object "zero poll". But I think this is more complicated.

So this would require some more discussions. I'm closing this for now.

PR #3173, opened 2022-07-07
do not print outside the buffer

BeardedSkunk:
else stack-corruption might occur

### DIFF
--- a/pjsip/src/pjsip/sip_auth_msg.c
+++ b/pjsip/src/pjsip/sip_auth_msg.c
@@ -156,7 +156,10 @@ static int pjsip_authorization_hdr_print( pjsip_authorization_hdr *hdr,
 	return -1;
 
     buf += printed;
-    *buf = '\0';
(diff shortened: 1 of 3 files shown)


silentindark:
Could you @BeardedSkunk to sign the CLA, then pjproject team can review it.

BeardedSkunk:
> Could you @BeardedSkunk to sign the CLA, then pjproject team can review it.

I did but the workflow still says pending :( when I click on details it says I've already signed.

sauwming:
Thank you very much for finding this issue.

Unfortunately, we cannot skip the NULL terminator since it will make the printed result useless (and potentially dangerous), especially if the result is later passed to APIs such as `pjsip_parse_*()` or `pjsua_verify_url()` that expects the input to be NULL terminated.

For `sip_tel_uri`, it seems to be safe since `endbuf` has been decremented by one, so I apply the same approach to the others.
I have created a new PR #3175 with the modified fix.

PR #3175, opened 2022-07-08
Fix potential buffer overrun when printing SIP header/message

sauwming:
As reported in #3173, when printing a SIP header/message, the NULL terminator character may go beyond the provided buffer.

This PR also fixes several buffer assignment that doesn't check against the end buffer. The solution is to use `copy_advance_char_check()` consistently.

Special thanks to @BeardedSkunk for the original report.

### DIFF
--- a/pjsip/src/pjsip/sip_uri.c
+++ b/pjsip/src/pjsip/sip_uri.c
@@ -256,10 +256,11 @@ static pj_ssize_t pjsip_url_print(  pjsip_uri_context_e context,
 {
     int printed;
     char *startbuf = buf;
-    char *endbuf = buf+size;
+    char *endbuf = buf+size-1; // Need to minus one for NULL terminator
     const pj_str_t *scheme;
     const pjsip_parser_const_t *pc = pjsip_parser_const();
 
+    if (endbuf < startbuf) return -1;
     *buf = '\0';
 
     /* Print scheme ("sip:" or "sips:") */
@@ -558,7 +559,7 @@ static pj_ssize_t pjsip_name_addr_print(pjsip_uri_context_e context,
 {
     int printed;
     char *startbuf = buf;
-    char *endbuf = buf + size;
+    char *endbuf = buf + size - 1; // Need to minus one for NULL terminator
     pjsip_uri *uri;
 
     uri = (pjsip_uri*) pjsip_uri_get_uri(name->uri);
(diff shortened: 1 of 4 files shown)


pjsip/src/pjsip/sip_uri.c
```diff
@@ -687,7 +688,7 @@ static pj_ssize_t other_uri_print(pjsip_uri_context_e context,
 
     /* Print scheme. */
     copy_advance(buf, uri->scheme);
-    *buf++ = ':';
+    copy_advance_char_check(buf, ':');
```
nanangizz, line 691: rather OOT, but while on this area, perhaps this functions should also add NULL termination (and length minus one) too?

nanangizz (approved):

trengginas (approved):

pjsip/src/pjsip/sip_uri.c
```diff
@@ -256,10 +256,11 @@ static pj_ssize_t pjsip_url_print(  pjsip_uri_context_e context,
 {
     int printed;
     char *startbuf = buf;
-    char *endbuf = buf+size;
+    char *endbuf = buf+size-1; // Need to minus one for NULL terminator
     const pj_str_t *scheme;
     const pjsip_parser_const_t *pc = pjsip_parser_const();
 
+    if (endbuf < startbuf) return -1;
```
andreas-wehrmann, line 263: You're checking here for the special case of size == 0, which is fine, but you're not doing this check in all the other places where you do
`char *endbuf = buf+size-1;`
like for example here:
https://github.com/pjsip/pjproject/blob/4813f17bff2e95780403276d4cfd88785c187a0e/pjsip/src/pjsip/sip_auth_msg.c#L128

https://github.com/pjsip/pjproject/blob/4813f17bff2e95780403276d4cfd88785c187a0e/pjsip/src/pjsip/sip_auth_msg.c#L294

https://github.com/pjsip/pjproject/blob/4813f17bff2e95780403276d4cfd88785c187a0e/pjsip/src/pjsip/sip_tel_uri.c#L185

https://github.com/pjsip/pjproject/blob/4813f17bff2e95780403276d4cfd88785c187a0e/pjsip/src/pjsip/sip_uri.c#L562

Do you not need to add this check in these places as well?

sauwming: I thought of that, and the reason I didn't put it in other places is because it will be checked in `copy_advance()`.

PR #3179, opened 2022-07-11
Add new api pj_sock_socketpair()

jimying:
Add ioqueue wakeup mechanism (#3172 ) need use socketpair(), implement it as separate PR here.

### DIFF
--- a/pjlib/include/pj/sock.h
+++ b/pjlib/include/pj/sock.h
@@ -1527,6 +1527,27 @@ PJ_DECL(pj_status_t) pj_sock_shutdown( pj_sock_t sockfd,
 PJ_DECL(char *) pj_addr_str_print( const pj_str_t *host_str, int port, 
 				   char *buf, int size, unsigned flag);
 
+/**
+ * Create socket pair
+ * @param family    Specifies a communication domain; this selects the
+ *		    protocol family which will be used for communication.
+ *		    On Unix, support AF_UNIX, AF_INET and AF_INET6,
+ *		    On Win32, not support AF_UNIX.
+ * @param type	    The socket has the indicated type, which specifies the 
+ *		    communication semantics.
+ * @param protocol  Specifies  a  particular  protocol  to  be used with the
+ *		    socket.  Normally only a single protocol exists to support 
+ *		    a particular socket  type  within  a given protocol family, 
+ *		    in which a case protocol can be specified as 0.
+ * @param sv	    The new sockets vector
+ *
+ * @return	    Zero on success.
+ *
+ */
+PJ_DECL(pj_status_t) pj_sock_socketpair(int family,
+					int type,
+					int protocol,
+					pj_sock_t sv[2]);
 
 /**
  * @}
--- a/aconfigure
+++ b/aconfigure
@@ -6224,6 +6224,32 @@ printf "%s\n" "no" >&6; }
 fi
 rm -f core conftest.err conftest.$ac_objext conftest.beam conftest.$ac_ext
 
+{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking if socketpair() is available" >&5
+printf %s "checking if socketpair() is available... " >&6; }
+cat confdefs.h - <<_ACEOF >conftest.$ac_ext
+/* end confdefs.h.  */
+#include <sys/types.h>
+				     #include <sys/socket.h>
+int
+main (void)
+{
+socketpair(0, 0, 0, 0);
+  ;
+  return 0;
+}
+_ACEOF
+if ac_fn_c_try_compile "$LINENO"
+then :
+  printf "%s\n" "#define PJ_SOCK_HAS_SOCKETPAIR 1" >>confdefs.h
+
+		   { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: yes" >&5
+printf "%s\n" "yes" >&6; }
+else $as_nop
+  { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: no" >&5
+printf "%s\n" "no" >&6; }
+fi
+rm -f core conftest.err conftest.$ac_objext conftest.beam conftest.$ac_ext
+
 { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking if sockaddr_in has sin_len member" >&5
 printf %s "checking if sockaddr_in has sin_len member... " >&6; }
 cat confdefs.h - <<_ACEOF >conftest.$ac_ext
--- a/aconfigure.ac
+++ b/aconfigure.ac
@@ -344,6 +344,15 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <sys/types.h>
 		   AC_MSG_RESULT(yes)],
 		  [AC_MSG_RESULT(no)])
 
+dnl # Determine if socketpair() is available
+AC_MSG_CHECKING([if socketpair() is available])
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <sys/types.h>
+				     #include <sys/socket.h>]],
+		    		  [socketpair(0, 0, 0, 0);])],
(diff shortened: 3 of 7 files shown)


pjlib/build/Makefile
```diff
@@ -52,7 +52,7 @@ export TEST_OBJS += activesock.o atomic.o echo_clt.o errno.o exception.o \
 		    select.o sleep.o sock.o sock_perf.o ssl_sock.o \
 		    string.o test.o thread.o timer.o timestamp.o \
 		    udp_echo_srv_sync.o udp_echo_srv_ioqueue.o \
-		    util.o
+		    util.o socketpair.o
```
bennylp, line 55: I would suggest putting the test in existing `sock.c` to simplify changes.

pjlib/include/pj/sock.h
```diff
@@ -1527,6 +1527,25 @@ PJ_DECL(pj_status_t) pj_sock_shutdown( pj_sock_t sockfd,
...
+ *		    in which a case protocol can be specified as 0.
+ * @param sv	    The new sockets vector
+ *
+ * @return	    Zero on success.
+ *
+ */
+PJ_DECL(pj_status_t) pj_sock_socketpair(int family,
+				    int type,
+				    int protocol,
+				    pj_sock_t sv[2]);
```
bennylp, line 1548: Is the indentation of 2nd, 3rd, 4th arguments wrong, or is it just my browser?

pjlib/src/pj/sock_bsd.c
```diff
@@ -905,4 +905,127 @@ PJ_DEF(pj_status_t) pj_sock_accept( pj_sock_t serverfd,
 }
 #endif	/* PJ_HAS_TCP */
 
+/*
```
bennylp, line 908: First of all, I would suggest putting the whole implementation in `sock_common.c` instead of `sock_bsd.c`, so that it can be used even when the platform doesn't use `sock_bsd.c`.

pjlib/src/pj/sock_bsd.c
```diff
@@ -905,4 +905,127 @@ PJ_DEF(pj_status_t) pj_sock_accept( pj_sock_t serverfd,
...
+
+#if PJ_HAS_TCP
+    PJ_ASSERT_RETURN(type == pj_SOCK_DGRAM() || type == pj_SOCK_STREAM(),
+		     PJ_EINVAL);
+#else
+    PJ_ASSERT_RETURN(type == pj_SOCK_DGRAM(), PJ_EINVAL);
+#endif
+
+    PJ_ASSERT_RETURN(family == pj_AF_UNIX() || family == pj_AF_INET() ||
+			 family == pj_AF_INET6(), PJ_EINVAL);
```
bennylp, line 929: I'd say it's better to reject pj_AF_UNIX() altogether, rather than silently changing the type to pj_AF_INET() below.

pjlib/src/pj/sock_bsd.c
```diff
@@ -905,4 +905,127 @@ PJ_DEF(pj_status_t) pj_sock_accept( pj_sock_t serverfd,
...
+	goto on_error;
+    status = pj_sock_connect(cfd, &sa, salen);
+    if (status != PJ_SUCCESS)
+	goto on_error;
+
+    if (type == pj_SOCK_DGRAM()) {
+	status = pj_sock_getsockname(cfd, &sa, &salen);
+	if (status != PJ_SUCCESS)
+	    goto on_error;
+	status = pj_sock_connect(lfd, &sa, salen);
```
bennylp, line 971: I think it's simpler to connect cfd to lfd, so we don't have to call getsockname() on cfd. It's also more intuitive (I'm thinking "c" in "cfd" as client).

pjlib/src/pj/sock_bsd.c
```diff
@@ -905,4 +905,127 @@ PJ_DEF(pj_status_t) pj_sock_accept( pj_sock_t serverfd,
 }
 #endif	/* PJ_HAS_TCP */
 
+/*
+ * Simulate unix-like socketpair()
+ * Use a pair of IPv4/IPv6 local sockets (listening on 127.0.0.1/::1)
+ */
+static int socketpair_s(int family, int type, int protocol, pj_sock_t sv[2])
+{
+    int status;
```
bennylp, line 914: s/int/pj_status_t/

pjlib/src/pj/sock_bsd.c
```diff
@@ -905,4 +905,127 @@ PJ_DEF(pj_status_t) pj_sock_accept( pj_sock_t serverfd,
...
+    if (cfd != PJ_INVALID_SOCKET)
+	pj_sock_close(cfd);
+    return status;
+}
+
+#if defined(PJ_SOCK_HAS_SOCKETPAIR) && PJ_SOCK_HAS_SOCKETPAIR != 0
+PJ_DEF(pj_status_t) pj_sock_socketpair(int family,
+				   int type,
+				   int protocol,
+				   pj_sock_t sv[2])
```
bennylp, line 1003: Arguments indentation again

pjlib/src/pj/sock_bsd.c
```diff
@@ -905,4 +905,127 @@ PJ_DEF(pj_status_t) pj_sock_accept( pj_sock_t serverfd,
 }
 #endif	/* PJ_HAS_TCP */
 
+/*
+ * Simulate unix-like socketpair()
+ * Use a pair of IPv4/IPv6 local sockets (listening on 127.0.0.1/::1)
+ */
+static int socketpair_s(int family, int type, int protocol, pj_sock_t sv[2])
```
bennylp, line 912: Proposing s/socketpair_s/socketpair_imp/ to make it clearer

bennylp (changes requested):

pjlib/include/pj/sock.h
jimying: Could you share code formatting rule?

pjlib/src/pj/sock_bsd.c
jimying: socketpair is two-way connection. so udp both need cfd connect lfd  and lfd connect cfd.

pjlib/include/pj/sock.h
bennylp: I don't we have one IIRC. But as you can see in the function declarations/defs, usually arguments are aligned vertically

pjlib/build/Makefile
bennylp: Any reason why this is not done?

pjlib/src/pj/sock_bsd.c
jimying: For compatibility with unix-like socketpair(), must support AF_UNIX，so  when family is pj_AF_UNIX(), change it to pj_AF_INET()

pjlib/build/Makefile
jimying: sorry forgot it. I have changed it.

pjlib/src/pj/sock_bsd.c
bennylp: Perhaps it would be clearer if we write down the spec of the function:

- On Linux/Unix, `af` values of AF_UNIX, AF_INET, and AF_INET6 are supported
- On Windows/others that don't have AF_UNIX, only `af` values AF_INET and AF_INET6 are supported
- In all cases, the correct type of socket must be returned according to `af` and `type`

jimying: I have removed  forcely convert AF_UNIX to AF_INET logic

jimying:
sync with lastest master branch

bennylp (approved):

bennylp:
There is still one more issue, but I think we can worry about it later.

If you see `pj_sock_socket()` implementation, after the `socket()` is called, some socket options are applied. Ideally we should do the same to the sockets created by the socket pair (for AF_INET4/6). But since the socket pair is expected only to be used internally, I guess it should be find without it.

sauwming (approved):

PR #3181, opened 2022-07-13
Add option to use default settings when switching sound device

trengginas:
The sound device settings used in ```pjsua_snd_set_setting()/pjsua_snd_get_setting()``` are global settings and not device specific.
It will be applied to any devices, even when application has changed the sound device to be used.
This might raise in an unwanted issue:
- First sound device opened is the main speaker. The settings will be retrieve and stored
- User switch the sound device to a headphone. The volume might changes because it is using the volume settings from the main speaker.

This patch will allow to switch the sound device and use the default settings and ignore the previously stored settings.
When the option is used, the current sound device settings will be reset and reinitialize based on the sound device used.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -7790,7 +7790,8 @@ PJ_DECL(pj_bool_t) pjsua_snd_is_active(void);
  * 
  * Note that in case the setting is kept for future use, it will be applied
  * to any devices, even when application has changed the sound device to be
- * used.
+ * used. To reset the setting, application can call #pjsua_set_snd_dev2()
+ * with \a use_default_settings set to PJ_TRUE.
  *
  * Note also that the echo cancellation setting should be set with 
  * #pjsua_set_ec() API instead.
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -7195,10 +7195,12 @@ typedef struct pjsua_snd_dev_param
     unsigned		mode;
 
     /*
-     * When opening the sound device, the settings set from
-     * #pjsua_snd_set_setting() will be applied to any sound device.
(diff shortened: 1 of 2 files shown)


pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -7194,6 +7194,20 @@ typedef struct pjsua_snd_dev_param
      */
     unsigned		mode;
 
+    /*
+     * When opening the sound device, the settings set from
+     * #pjsua_snd_set_setting() will be applied to any sound device.
```
sauwming, line 7199: Modify the doc to explain that the setting is not just from pjsua_snd_set_setting(), but also taken from the first sound device opened. As in the original report, the latter is more problematic because app is not aware since it never calls any API to set the sound dev settings.

pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -1805,8 +1813,7 @@ static pj_status_t update_initial_aud_param()
     pjmedia_aud_param param;
     pj_status_t status;
 
-    PJ_ASSERT_RETURN(pjsua_var.snd_port != NULL, PJ_EBUG);
-
+    PJ_ASSERT_RETURN(pjsua_var.snd_port != NULL, PJ_EBUG);
```
sauwming, line 1816: unnecessary modification

sauwming (commented):
There seems to be a behavioral change. Previously:
```
pjsua_snd_set_setting(, keep=PJ_TRUE);
pjsua_set_snd_dev();
```
will apply the setting, but now it no longer works.

Also need to add doc in `pjsua_snd_set_setting()`:
```
 * Note that in case the setting is kept for future use, it will be applied
 * to any devices, even when application has changed the sound device to be
 * used.
```
`To reset the setting, application can call #pjsua_set_snd_dev2() with \a use_default_settings set to PJ_TRUE.`

sauwming (approved):

nanangizz (approved):

PR #3183, opened 2022-07-13
Lock ordering issue in video conference bridge

nanangizz:
### Sample scenario
- Video conference (VC) worker thread: `put/get_frame()` publishes media event (synchronously), currently `put/get_frame()` is called while holding the VC mutex and app receiving the event invokes a SIP dialog API which tries to acquire the SIP dialog mutex.
- Other thread, app receives a SIP dialog callback (note: such callback is usually invoked while holding dialog mutex), it invokes a VC API which tries to acquire VC mutex.

The scenario involves non-uniform lock ordering between VC and SIP dialog mutexes which can cause deadlock.

### Proposed solution
The proposed solution is to avoid holding VC mutex while calling `put/get_frame()` in VC worker thread. To allow this, a couple major modifications are implemented in this PR:
1. VC internal states (such as port connection, rendering states) must not be changed when `put/get_frame()` is called, so some VC API operations (such as connect/disconnect ports, remove port, update port) must be queued or done asynchronously, the real operation will be performed in the VC worker thread context.
2. As remove port operation will be done asynchronously, media ports (especially video ports for now) need to have a reference counter (or group lock) to avoid premature destroy.

### Other fixes
This PR also contains a few minor other fixes such as:
- in `pj_grp_lock_create_w_handler()`: memory pool used for adding handler, updated to using group lock's memory pool itself, previously using app pool, was getting a premature destroy issue because of this (i.e: VC port added using PJSUA vidwin pool, the vidwin may get destroyed before the VC port, group lock crashed when iterating handlers).
- in PJMEDIA stream, somehow transport is accessed before the attach happens and it is still NULL, or it is accessed after set to NULL (in destructor), so put NULL transport check in some RTP/RTCP sending.
- in VC, fill buffer with black is performed when buffer is initially allocated and sink buffer after sink format changed, was every time render state is updated.
- add/modify logs in a few places including in app.

### DIFF
--- a/pjmedia/include/pjmedia/port.h
+++ b/pjmedia/include/pjmedia/port.h
@@ -30,6 +30,7 @@
 #include <pjmedia/frame.h>
 #include <pjmedia/signatures.h>
 #include <pj/assert.h>
+#include <pj/lock.h>
 #include <pj/os.h>
 
 
@@ -387,6 +388,14 @@ typedef struct pjmedia_port
     } port_data;
 
     /**
+     * Group lock.
+     *
+     * This is optional, but if this port is registered to the audio/video
+     * conference bridge, the bridge will create one if the port has none.
+     */
+    pj_grp_lock_t	*grp_lock;
+
+    /**
      * Get clock source.
      * This should only be called by #pjmedia_port_get_clock_src().
      */
@@ -492,7 +501,10 @@ PJ_DECL(pj_status_t) pjmedia_port_put_frame( pjmedia_port *port,
 					     pjmedia_frame *frame );
 
 /**
- * Destroy port (and subsequent downstream ports)
+ * Destroy port (and subsequent downstream ports).
+ *
+ * Note that if the port has group lock, instead of destroying the port
+ * immediately, this function will just decrease the reference counter.
  *
  * @param port	    The media port.
  *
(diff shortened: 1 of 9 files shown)


pjmedia/src/pjmedia/port.c
```diff
@@ -124,17 +124,94 @@ PJ_DEF(pj_status_t) pjmedia_port_put_frame( pjmedia_port *port,
...
+    pj_grp_lock_t *grp_lock = glock;
+    pj_status_t status;
+
+    PJ_ASSERT_RETURN(port && pool, PJ_EINVAL);
+    PJ_ASSERT_RETURN(port->grp_lock == NULL, PJ_EEXISTS);
+
+    /* We need to be caution on ports that do not have the on_destroy()!
+     * It can be uninitialized yet or it really does not have one.
+     * If it doesn't have one, then we'd expect a possible premature destroy!
+     */
```
sauwming, line 167: Better print some logs so we can know during log analysis which port doesn't have on_destroy().

pjmedia/include/pjmedia/port.h
```diff
@@ -501,6 +513,48 @@ PJ_DECL(pj_status_t) pjmedia_port_put_frame( pjmedia_port *port,
...
+ * lock by adding the port's destructor to the group lock handler list, and
+ * increment the reference counter.
+ *
+ * This function should only be called by a media port implementation and
+ * after port's on_destroy() function has been assigned.
+ *
+ * @param port		    The pjmedia port to be initialized.
+ * @param pool		    The pool, this can be a temporary pool as
+ *			    group lock will create and use its internal pool.
+ * @param grp_lock	    The group lock, or create a new one if NULL.
```
sauwming, line 528: glock

sauwming (commented):
Also, can you merge with the latest master so I can do some local testing?

nanangizz:
> Also, can you merge with the latest master so I can do some local testing?

Done, hopefully I didn't do it wrong, the traces above look a bit strange :)

pjmedia/src/pjmedia/vid_stream.c
```diff
@@ -235,6 +235,10 @@ static void on_rx_rtcp( void *data,
                         void *pkt,
                         pj_ssize_t bytes_read);
 
+static void on_destroy(void *arg);
+
+static void channel_on_destroy(void *arg) { PJ_UNUSED_ARG(arg); }
```
sauwming, line 240: `../src/pjmedia/vid_stream.c:240:13: warning: unused function 'channel_on_destroy' [-Wunused-function]
static void channel_on_destroy(void *arg) { PJ_UNUSED_ARG(arg); }
`

sauwming (approved):

bennylp (approved):

PR #3184, opened 2022-07-14
Add UPnP support for SIP UDP and media UDP transports

sauwming:
Steps to add UPnP support:
- Download `libupnp` from https://github.com/pupnp/pupnp
- Build `libupnp` by following the [instructions](https://github.com/pupnp/pupnp#10-build-instructions). Check the build note below for some recommendations.
- Configure PJSIP and make sure there is the following output:
```checking UPnP usability... yes```
- Enable UPnP in your app by setting `pjsua_config.enable_upnp` or `UaConfig::enableUpnp`.

**Build recommendation for`libupnp`** :
By default,`libupnp` send its requests using blocking sockets. This may cause long timeout delay (in our test, it can take more than 1 minute) if your router/IGD (Internet Gateway Device) suddenly goes offline or in the network disconnection scenario. To overcome this, enable non-blocking option for the sockets by configuring `libupnp` with:
```./configure --disable-blocking_tcp_connections [--enable-debug]```
Next, even though the socket is now non-blocking, the default timeout set by `libupnp` is also quite long (30 secs). Unfortunately, the timeout is not configurable, so you may need to modify it directly in the following files:
- `upnp/src/inc/httpreadwrite.h:` change `HTTP_DEFAULT_TIMEOUT` (default 30 secs)
- `upnp/src/inc/upnpapi.h`: change `UPNP_TIMEOUT` (default 30 secs)

Known limitation:
- `libupnp` initialisation function `UpnpInit2()` expects an interface name as its parameter, however currently there's no PJSIP API to get the interface name. `pj_enum_ip_interface()` will only enumerate the interface addresses. The parameter is optional though, so passing `NULL` and let `libupnp` find the first suitable interface for you should probably work in most circumstances.

Tested with `libupnp` 1.14.x.

Still TODO:
- Windows VS support

### CODE REVIEW

File: pjlib-util/src/pjlib-util/upnp.c [hunk 1]
```diff
@@ -0,0 +1,907 @@
...
+    upnp_err = UpnpInitLog();
+    if (upnp_err != UPNP_E_SUCCESS) {
+	PJ_LOG(4, (THIS_FILE, "Failed to initialize UPnP log: %s",
+			      UpnpGetErrorMessage(upnp_err)));
+    }
+#endif
+
+    pj_bzero(&upnp_mgr, sizeof(upnp_mgr));
+    upnp_err = UpnpInit2(param->if_name, (unsigned short)param->port);
+    if (upnp_err != UPNP_E_SUCCESS) {
```
bennylp, line 579: Q: has this been tested with pjsua restart scenario?

bennylp (changes requested):

pjlib-util/src/pjlib-util/upnp.c
sauwming: Yes, this was the scenario that got my most attention because if pjsua is restarted when the network is down/changes, `libupnp` does create a lot of problems.

By default, `libupnp` creates its sockets in blocking mode, so a failed connection can take a long time before it times out. The solution for this, as described in the PR description above, is to disable blocking TCP connections when configuring libupnp. But then, even in non-blocking mode, `libupnp` sets a notoriously long timeout (30 secs) and the timeout is non-configurable so we need to modify the source (details in the desc above) to adjust this (I set it to 2 or 3 secs at most).

Still, the situation is not ideal because `UpnpFinish()` is a blocking function that will wait until every connection completes and all worker threads end, so it typically takes at least 4 secs to deinit:
```
10:21:48.904                 upnp.c  .UPnP deinitializing...
10:21:52.900                 upnp.c !Port mapping deletion action complete error: -207 (UPNP_E_TIMEDOUT)
10:21:52.900                 upnp.c !.UPnP deinitialized
```

But then, I imagine that UPnP should not be used a lot in dynamic/mobile network environments, so it's probably acceptable. I may be wrong here though.

bennylp (approved):

nanangizz (approved):

Issue #2089, opened 2018-02-12
Support receiving Opus packets with various frame lengths


pjsipbot:
_2018-02-12 06:17:41: @mingteluu_ _created the issue on trac ticket [2089](https://trac.pjsip.org/repos/ticket/2089)_

From the [RFC 7587](https://tools.ietf.org/html/rfc7587#section-4.2):
```
The Opus encoder can output encoded frames representing 2.5, 5, 10,
   20, 40, or 60 ms of speech or audio data.  Further, an arbitrary
   number of frames can be combined into a packet, up to a maximum
   packet duration representing 120 ms of speech or audio data.
```

For example, for a packet duration of 40ms, we can receive 1 frame of 40ms, 2x20ms, 4x10ms, etc. However, currently pjmedia only expects constant frame length to be specified during stream creation.

Thank you to Marcus Froeschl for the suggestion and patch testing.

___
_2018-02-12 06:18:27: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2018-02-12 06:18:27: @mingteluu_ _set **resolution** to **fixed**_

___
_2018-02-12 06:18:27: @mingteluu_ _commented_

In [r5734](../commit/c7d20d5111621560fc7e6b431dc25b738ac362d6):
Fixed #2089: Support receiving Opus packets with various frame lengths

___
_2018-05-30 08:41:27: @nanangizz_ _commented_

In [r5799](../commit/ba59059944fd6246188f5b838cdf43be0dcd0fcf):
Re #2089:
 - Set default max Opus RX frame length, i.e: param.info.max_rx_frame_size, to 1275. This is to avoid issue of truncated frame in jitter buffer when incoming frame length gets increased (after remote Opus encoder changes Opus params on the fly, e.g: ptime, VBR/CBR, sampling rate). Thanks Marcus Froeschl for the report, the investigation, and the solution.
 - Fixed pjmedia_codec_opus_set_default_param() to initiate codec param with current default settings.
 - Fixed some compile warnings.

___
_2018-07-23 07:39:38: @mingteluu_ _commented_

Replying to [comment:2 nanang]:
> In [r5799](../commit/ba59059944fd6246188f5b838cdf43be0dcd0fcf):
> ```
> > Re #2089:
[...]

PR #3189, opened 2022-07-21
Add ability in stream to modify Opus codec parameters on the fly

sauwming:
As described in [RFC 6176](https://datatracker.ietf.org/doc/html/rfc6716):
```
The Opus codec includes a number of control parameters that can be
   changed dynamically during regular operation of the codec, without
   interrupting the audio stream from the encoder to the decoder. These
   parameters only affect the encoder since any impact they have on the
   bitstream is signaled in-band such that a decoder can decode any Opus
   stream without any out-of-band signaling.
```

Typically, the adjusted parameters are the bitrate, bandwidth, and frame duration, in order to adapt to the changes in network condition. For example, in a slower network, sender can increase the frame duration and reduce the bitrate&bandwidth.
```
Opus can encode frames of 2.5, 5, 10, 20, 40, or 60 ms.  It can also
   combine multiple frames into packets of up to 120 ms.  For real-time
   applications, sending fewer packets per second reduces the bitrate,
   since it reduces the overhead from IP, UDP, and RTP headers.
```

For the receiving direction, most of the hard work has been done in #2089 that handles the frame duration change. While for bitrate and bandwidth changes, as the above RFC text says, they will be internally taken care of by the Opus decoder itself.

But for the outgoing direction, currently there's no mechanism for PJSIP to modify the codec parameters, so this PR will add `pjmedia_stream_modify_codec_param()`. Applications monitoring network conditions can thus use this API to modify the current codec's settings, in particular when using Opus codec.

### DIFF
--- a/pjmedia/include/pjmedia/stream.h
+++ b/pjmedia/include/pjmedia/stream.h
@@ -299,6 +299,25 @@ PJ_DECL(pj_status_t) pjmedia_stream_start(pjmedia_stream *stream);
 
 
 /**
+ * Modify the stream's codec parameter after the codec is opened.
+ * Note that not all codec parameters can be modified during run-time.
+ * When the parameter cannot be changed, this function will return
+ * non-PJ_SUCCESS, and the original parameters will not be changed.
+ *
+ * Currently, only Opus codec supports changing key codec parameters
+ * such as bitrate and bandwidth, while other codecs may only be able to
+ * modify minor settings such as VAD.
+ *
+ * @param stream	The media stream.
+ * @param param		The new codec parameter.
+ *
+ * @return		PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t)
+pjmedia_stream_modify_codec_param(pjmedia_stream *stream,
+			  	  const pjmedia_codec_param *param);
+
+/**
  * Get the stream info.
  *
  * @param stream	The media stream.
--- a/pjmedia/include/pjmedia/stream.h
+++ b/pjmedia/include/pjmedia/stream.h
@@ -301,12 +301,9 @@ PJ_DECL(pj_status_t) pjmedia_stream_start(pjmedia_stream *stream);
 /**
  * Modify the stream's codec parameter after the codec is opened.
  * Note that not all codec parameters can be modified during run-time.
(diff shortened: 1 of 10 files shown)


bennylp (approved):

pjmedia/include/pjmedia/stream.h
```diff
@@ -298,6 +298,25 @@ PJ_DECL(pjmedia_transport*) pjmedia_stream_get_transport(pjmedia_stream *st);
 PJ_DECL(pj_status_t) pjmedia_stream_start(pjmedia_stream *stream);
 
 
+/**
+ * Modify the stream's codec parameter after the codec is opened.
+ * Note that not all codec parameters can be modified during run-time.
+ * When the parameter cannot be changed, this function will return
+ * non-PJ_SUCCESS, and the original parameters will not be changed.
```
nanangizz, line 305: From a quick check, non-Opus codecs will just ignore parameter changes other than VAD/PLC and still return `PJ_SUCCESS`.

pjsip/include/pjsua2/call.hpp
```diff
@@ -562,6 +562,11 @@ struct CallInfo
  */
 struct StreamInfo
 {
+    /**
+     * Media stream instance, either audio or video.
+     */
+    MediaStream		stream;
+
```
nanangizz, line 569: how about apps using SWIG such as Android, there seem to be no API to modify the codec param?

sauwming: Yes, because there's currently no API in the pjsua level as well (currently, the PR only provides access to the media stream and user will call PJMEDIA API directly).

So do we want to add pjsua&pjsua2 API as well? Perhaps something like `pjsua_call_aud_stream_modify_codec_param(call_id, med_idx, codec_param)`?

nanangizz: Yes.

Btw, IMO direct access to pjmedia stream in pjsua/pjsua2 is better avoided for now, as there seem to be no urgent need of it for now, even it potentially introduces problems, e.g: SDP renego may recreate streams and there can be race cond or access-after-free issues when app accessing the stream without lock/ref-count protection.

nanangizz (approved):

PR #3193, opened 2022-07-26
Update third party media sample implementation

sauwming:
There has been a lot of additional new pjsua APIs ever since the sample was last created/updated, so we need to add these otherwise the sample will fail to build with the undefined references errors.

### DIFF
--- a/pjsip-apps/src/3rdparty_media_sample/alt_pjsua_aud.c
+++ b/pjsip-apps/src/3rdparty_media_sample/alt_pjsua_aud.c
@@ -327,6 +327,16 @@ PJ_DEF(pjsua_conf_port_id) pjsua_call_get_conf_port(pjsua_call_id call_id)
     return PJSUA_INVALID_ID;
 }
 
+/* Modify audio stream's codec parameters. */
+PJ_DEF(pj_status_t)
+pjsua_call_aud_stream_modify_codec_param(pjsua_call_id call_id,
+                                         int med_idx,
+			  	  	 const pjmedia_codec_param *param)
+{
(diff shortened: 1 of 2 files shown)


nanangizz (approved):

trengginas (approved):

wosrediinanatour:
Note that for PJSUA2, it would even be more.

sauwming:
> Note that for PJSUA2, it would even be more.

Not sure what you mean, but PJSUA2 is just a wrapper for PJSUA and there should be no extra step needed in PJSUA2 to incorporate 3rd party media. I tested here by trying to build `libpjsua2.so` for Android using `alt_pjsua_aud&vid.c` and it built successfully.

wosrediinanatour:
> > Note that for PJSUA2, it would even be more.
> 
> Not sure what you mean, but PJSUA2 is just a wrapper for PJSUA and there should be no extra step needed in PJSUA2 to incorporate 3rd party media. I tested here by trying to build `libpjsua2.so` for Android using `alt_pjsua_aud&vid.c` and it built successfully.

Example: `pjsua_ext_snd_dev_get_conf_port` is used in `external-pjproject/pjsip/src/pjsua2/media.cpp` only (latest release of PJSUA2). I see that you have already included it, which was my intend.

Another note (which I have not veryfied with latest version/commit): To prevent out-dated `alt_pjsua_*.c`, it would be good, if the `3rdparty_media_sample` was build automatically.

sauwming:
Unfortunately 3rd media sample can't be automatically built together with PJSIP with default config since it needs `PJSUA_MEDIA_HAS_PJMEDIA` to be disabled.

PR #3195, opened 2022-07-29
Move UPnP to PJNATH

sauwming:
As a NAT traversal helper, UPnP (added in #3184)  should be better put inside pjnath, instead of pjlib-util.


bennylp (commented):
For documentation purpose:
- Rename PJLIB_UTIL_UPNP to PJNATH_UPNP in upnp.h, and remove "ingroup PJ_PROTOCOLS"
- Add PJNATH_UPNP entry in pjnath/docs/doc_mainpage.h (under "The library provides the following main component groups")

bennylp (approved):
Nice :)

nanangizz (approved):

AmarOk1412:
:confetti_ball: 

So adding upnp candidates from libupnp will be supported? We use it to inject our candidates inside the ICE components.
We also support NATPMP as a library which is pretty similar as some routers doesn't support one or the other (and on mobiles the support differs)

Finally, last question, will libpcp be planned?

AmarOk1412:
Also we have some complex mechanisms on our side, so I guess we will not use PJNATH_UPNP for a long time.
upnp can be slow depending on the version/implementation in the router (there is a lot of problems also, with limited pool, etc that we saw), so I guess it will take time before we switch anyway.

For example, we pre-open some ports because some routers take > 400ms to open a port (which is long)

sauwming:
No, currently UPnP is only used for simple UDP media and SIP transports.

And right now we don't have plan to support PCP or PMP.

PR #3196, opened 2022-07-29
Fix tx/rx directions for last RTCP update

mscdex:


nanangizz:
It is actually not a bug. Please check the docs of "rx/tx" in RTCP stats:
https://github.com/pjsip/pjproject/blob/74569f926a2f553006d4bbac70c7371dfb1eb400/pjmedia/include/pjmedia/rtcp.h#L210-L211
RTCP receiver reports from the remote represents our encoder/outgoing stream stats (how many packets sent by local are received/lost/etc seen by remote), that's why it is assigned to "tx". Similar explanation for the "rx" direction in RTCP stats.

Also note some lines above it, they are assigned to "tx" on purpose, for example:
https://github.com/pjsip/pjproject/blob/74569f926a2f553006d4bbac70c7371dfb1eb400/pjmedia/src/pjmedia/rtcp.c#L556

PR #3199, opened 2022-08-03
Fix SWIG C# build script to consider OS other than iOS and Android

sauwming:
Currently, SWIG C# build script only considers iOS and Android, but it is possible to generate PJSIP C# binding on other platforms (such as Windows) even though it's not officially supported yet. So we need to update the build script in order to avoid unexpected behaviour when invoked on other platforms.


bennylp (approved):
As a side note. prefixing paths with `./` (i.e. to become `rm -rf ./$(LIBPJSUA2_DIR)/*`) would also help (in the sense that if bad thing happens, at least only current directory is wiped out :D)

bennylp (approved):

nanangizz (approved):

PR #3200, opened 2022-08-04
Fix configure error on android NDK r25

trengginas:
Using android NDK r25, the `configure-android` command will fail :
```
clang: error: no such file or directory: '.../android-ndk-r25/sources/cxx-stl/llvm-libc++/libs/arm64-v8a/libc++_shared.so'
...
aconfigure: error: C compiler cannot create executables
```

On NDK r25, `libc++_shared.so` is located:
```
./toolchains/llvm/prebuilt/darwin-x86_64/sysroot/usr/lib/aarch64-linux-android/libc++_shared.so
```

On NDK r22, it is located:
```
./sources/cxx-stl/llvm-libc++/libs/arm64-v8a/libc++_shared.so
./toolchains/llvm/prebuilt/darwin-x86_64/sysroot/usr/lib/aarch64-linux-android/libc++_shared.so
```

Currently the `configure-android` will point to  `${STDCPP_TC}/libs/${TARGET_ABI}` resulting error on NDK r25.
This patch will fix this by using the correct path.


bennylp (approved):

aconfigure.ac
```diff
@@ -155,10 +159,12 @@ dnl
 dnl Legacy macros
 dnl
 AC_SUBST(ac_target_arch)
+AC_SUBST(ac_stl_cpp_lib)
 case $target in
     *android*)
 	AC_DEFINE(PJ_ANDROID,1)
 	ac_target_arch=$TARGET_ABI
+	ac_stl_cpp_lib=$ANDROID_LIBS
```
nanangizz, line 167: This may introduce problems as theoritically `ANDROID_LIBS` contains all the libs, instead of just the STL.
E.g: as in the makefile the libs will be copied to the app package, the APK size may get bloated due to those libs, while in this context it only needs the STL.

nanangizz (approved):

sauwming (approved):

PR #3202, opened 2022-08-06
Documentation fixes for docs.pjsip.org v2.0 project

bennylp:
Contents:
- fix Doxygen comments 
       - note that *breathe* may not display struct members if they are not documented, that's why I strive to add them even when it's not very informative
- remove `doc/pjsip-book` (the correct one is in pjproject_docs repository)
- replace Trac ticket URL with GitHub issues URL (note: Trac wiki URL will be replaced with docs.pjsip.org URL once the docs site is finished)
- add `clean-doc` target in root Makefile
- add PJSUA2 "hello world" sample (from docs page) to samples
- small fixes in sample apps to improve robustness

Note: common things that I came across:

- No documentation. See note in item 1 in contents above. Just write something.
- Typo perhaps: starting comment with plain `/*` rather than `/**`. Or `/*` comment after a struct field, which can be easily converted to `/**<`. Note if this is not a typo, then I feel it's a waste that your writing doesn't get into the doc.
- Requesting for explicit reference that doesn't exist, e.g. in `#b` below:
  ```
  @param a   See param #b
  @param b   This is param b.
  ```
  Note that things like parameters, struct members (e.g. callback) cannot be referenced.   This is not fatal, but it generates doxygen warning
- Duplicate section name, e.g.:
   ```
  \section G.729
   \subsection param  Codec param
   ```
   The `\subsection param` is used pretty much by all codec headers causing many duplicates.
- In pjsua2 headers, there are many parameter documentations that don't match the parameter name, e.g `prm.txOption` below.
  ```
  /**
   @param dest          The destination
   @param prm.txOption  Optional message component.
   */
  void xfer(const string &dest, const CallOpParam &prm);
  ```
  The generated documentation I think is okay, it's just that this generates a lot of doxygen warnings. Currently this is not fixed.
- Also doxygen group is important for the website.


pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -1158,7 +1158,7 @@ typedef struct pjsua_callback
 
     /**
      * Notify application when an audio media session is about to be created
-     * (as opposed to #on_stream_created() and #on_stream_created2() which are
+     * (as opposed to #on_stream_created() and on_stream_created2() which are
```
sauwming, line 1161: remove #

sauwming (approved):

pjsip/include/pjsua-lib/pjsua.h
bennylp: done

nanangizz (approved):

PR #3209, opened 2022-08-09
Remove useless timeb.h

jimying:
sys/timeb.h  ftime(), it is not used anymore.  I think we can remove it.

ftime() only called by os_time_bsd.c, and os_time_bsd.o only apear in os_win32.mak which looks wrong (should be os_time_win32.o?).


nanangizz:
After investigation, initially (in the very beginning of PJLIB development) it was named `os_time_ansi.c` and was used by linux and win32 platforms (btw, win32 has `sys/timeb.h` and `ftime()` is declared in there, so I don't think it is wrong). Later it was renamed to `os_time_bsd.c` [here](https://github.com/pjsip/pjproject/commits/e67d99a58a1b8f6d74582252dfa751b6f9a3dfd6/pjlib/src/pj/os_time_ansi.c?browsing_rename_history=true&new_path=pjlib/src/pj/os_time_bsd.c&original_branch=master). And few months later [here](https://github.com/pjsip/pjproject/commit/e67d99a58a1b8f6d74582252dfa751b6f9a3dfd6#diff-a2c1295fc98e9ea484fd8edbfde35a300825d2caceec1e8174984fe23450084d,) `os_time_unix.c` providing `gettimeofday()` was introduced, and linux platforms switched to using this. And later win32 platforms switched to using `os_time_win32.c`.

So yes, you're right, `ftime()`/`os_time_bsd.c` is no longer used by current build configs. However, as PJLIB tries to be as portable as possible and the existence of `os_time_bsd.c` may still be able to provide an alternative for platform porters (e.g: for embedded systems using an older OS or a limited system lib) and it does not seem to cause any issues (or does it?), we are not sure if removing it is a good idea (or at least an urgent matter) for now.

PR #3210, opened 2022-08-09
Expand tab to spaces

trengginas:
Currently, the source code was written with these formatting setting:
```
Indentation: 4
Tab size: 8
```
Some editors might not be able to correctly format when displaying/editing the code, resulting miss-alignment.
This PR will replace the `Tab` character with spaces to address such issue.
This PR will also remove svn keyword `$Id$` string which is no longer used.

Notes:
- To ignore the whitespaces-only change from this PR, use `git blame -w` to find the commit where the code was actually writtern.


wosrediinanatour:
+1 for .editorconfig, which also contains codec, end_of_line etc.

bennylp:
> +1 for .editorconfig, which also contains charset, end_of_line, trim_trailing_whitespace, insert_final_newline etc.

They're not available on all platforms/plugins though (e.g. eclipse)

wosrediinanatour:
> > +1 for .editorconfig, which also contains charset, end_of_line, trim_trailing_whitespace, insert_final_newline etc.
> 
> They're not available on all platforms/plugins though (e.g. eclipse)

yes, but at least it is a documentation you can refer to and the e.g. Makefile or Python code styles could be part of it (in case of C and C++, .clang-format file would make it for some people even better, since you could define the coding style more or less completely, you could format files by clang-format, and, hence, contributing would be easier.)

trengginas:
> 1. Have we confirmed that the feature to ignore certain revisions in GitHub blame works? I think this is a must.

The alternatives would be:
- adding `.git-blame-ignore-revs` files which include the ignored commit. 
- use `git blame -w` which will ignore whitespace changes. (no need to add `.git-blame-ignore-revs` file)

I tested both and they worked.
> 2. IMO the PR also needs to be more clear about what the official indentation style from now on

If we want to continue using the current style, it should be using space with size 4.

> 3. Also add instruction in the PR about how to ignore whitespace while seeing PRs/diffs

Added on the PR description.

> 4. Maybe also add an .editorconfig file, as a best effort solution for future edits (and add instructions about required plugins)

Let me do some tests for this

.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+# properties
+[*.{h,c,hpp,cpp,m,java}]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
```
sauwming, line 8: I believe this config won't work with VS Code? (see internal discussion email)

Also, perhaps add note in the PR what editors you have tested with and which work and which doesn't, in terms of viewing and adding/editing code, so users know what to expect.

trengginas: This should work after the tab is expanded to spaces. It's not going to work if spaces is still use in the code.

sauwming (commented):
Looks okay to me. But let's wait close to 2.13 release so we can merge with the latest master first.

sauwming (approved):
For 2.13, I think we can integrate this.

Remember to merge the master first and then expand the indentation of the latest merge.

Then after this PR is merged, create another PR to ignore this commit from blame.

nanangizz (approved):

trengginas:
Move to #3292

PR #3217, opened 2022-08-14
Add two VS2015 projects for CSharp: SWIG binding builder & sample app

nanangizz:
### Using `swig_csharp_pjsua2.vcxproj` for generating SWIG binding files

1. As we are generating a DLL, set the PJPROJECT solution config to "Debug-Dynamic" or "Release-Dynamic".
2. Add any third-party dependency in the library search path in the project setting, e.g: libh264 library path.
3. Make sure `swig.exe` is in the system path.
4. Build. This should generate `.cs` files and `pjsua2.dll` in `pjsip-apps/src/swig/csharp/output/`.

### Using the sample app `SimplePjsua2CS`

1. Copy the SWIG generated `.cs` files (currently ~271 files) to folder `pjsip-apps/src/swig/csharp/sample/SimplePjsua2CS/pjsua2/`. 
2. Add reference to `pjsua2.dll` in the `SimplePjsua2CS` project.
3. Run.

### Troubleshooting

- Error `TypeInitializationException` in running and there is build warning ```There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "pjsua2"```.
   Usually this can be solved by setting platform target in `SimplePjsua2CS` properties (in `Build > General > Platform target`) to match to the CPU config used to build the PJPROJECT.


sauwming (approved):

trengginas (approved):

aparsaeeyan:
I followed the steps you said.
 It created all the cs files and also created some other files named pjsua2.exp, pjsua2.lib, pjsua2_wrap.cpp but did not create the pjsua2.dll file. What should I do?
Also, I get another 185 error in Visual Studio after building the solution. But even with these errors, it creates cs files.errors

aparsaeeyan:
Can you upload and post pjsua2.dll separately? No matter what I do it doesn't give me the pjsua2.dll file.

maradv:
@aparsaeeyan 
We could resolve the linking errors by excluding the ignored libraries. However, we are unaware of any valid reason for excluding 'msvcrt.lib'.

PR #3222, opened 2022-08-22
Fix assertion when trying to set QOS type "signalling"

andreas-wehrmann:
Commit b1490d57d8c8004cc8427d591931c6a4691d7bf4 introduced a new enum member for _pj_qos_type_ called _PJ_QOS_TYPE_SIGNALLING_.
However, the assertion check in [pj_qos_get_params()](https://github.com/pjsip/pjproject/blob/4f280161f11f9d2d1a5dd0e223a462301c77f6d4/pjlib/src/pj/sock_qos_common.c#L46) was not updated to account for the new length of the QOS mapping array.
I noticed this, because I ported an app to Windows which would always abort() in Debug mode when starting up...

### DIFF
--- a/pjlib/src/pj/sock_qos_common.c
+++ b/pjlib/src/pj/sock_qos_common.c
@@ -43,7 +43,7 @@ static const pj_qos_params qos_map[] =
 PJ_DEF(pj_status_t) pj_qos_get_params(pj_qos_type type, 
 				      pj_qos_params *p_param)
 {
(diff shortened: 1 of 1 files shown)


sauwming (approved):

andreas-wehrmann:
I usually don't do this, but would it be possible to accelerate the merge of this request?
I kind of f*cked creating this PR up by creating it from master, which basically prevents me from creating more PRs
at least two of which I want to submit to this project.

nanangizz (approved):

PR #3225, opened 2022-08-23
Build Python Wheels via Makefile

florian-sattler:
Currently via make it's only possible to install python bindings.
This PR adds the possibility to build a python [wheel](https://pythonwheels.com/) instead.

Prerequisite: `wheel` (run `pip install wheel` to install).

Command:
```bash
cd pjsip-apps/src/swig/python
make wheel
```

To install PJSIP's wheel dist package, from the above directory, run:
`pip install dist/*.whl`

### DIFF
--- a/pjsip-apps/src/swig/python/Makefile
+++ b/pjsip-apps/src/swig/python/Makefile
@@ -33,6 +33,9 @@ SWIG_FLAGS += -w312 $(USE_THREADS)
 
 all: $(PYTHON_SO)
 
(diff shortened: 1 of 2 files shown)


florian-sattler:
CLA was signed shortly after opening the PR. Why does it not register?

sauwming:
Tested here and worked fine on MacOS. I edited the description (changed `make python wheel` to `make wheel` and add installation instruction, let me know if anything's missing).

sauwming (approved):

nanangizz (approved):

PR #3226, opened 2022-08-25
fix sdp miss check '=' when parse connection and bandwith info

jimying:
- Fix sdp miss check '=' when parse connection and bandwith info
- Enhance sdp error log print
   use pjmedia_strerror() to get error string message
- Fix format error

### DIFF
--- a/pjmedia/include/pjmedia/sdp.h
+++ b/pjmedia/include/pjmedia/sdp.h
@@ -495,7 +495,8 @@ typedef struct pjmedia_sdp_media pjmedia_sdp_media;
  * @return	    the length printed, or -1 if the buffer is too
  *		    short.
  */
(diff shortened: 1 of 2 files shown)


jimying:
A doubtful point:
https://github.com/pjsip/pjproject/blob/797088ed133c98492519b7d042b75735f6f9388c/pjmedia/src/pjmedia/sdp.c#L1457

if mean: if exception error parse sdp and assert when debug.
This line is error? should change to:
```
pj_assert(ctx.last_error == PJ_SUCCESS);
```

jimying:
Can use pj_register_strerror() to register pjmedia error
pj_strerror can catch pjmedia error string

```c
pj_register_strerror(PJMEDIA_ERRNO_START, PJ_ERRNO_SPACE_SIZE, 
				  &pjmedia_strerror)
```

PR #3230, opened 2022-08-31
Forcing lock codec to always use INVITE method

nanangizz:
Reported that some endpoints may not respond to UPDATE properly while indicating UPDATE support (by specifying UPDATE in its SIP Allow header). So when UPDATE is sent for locking codec, it is responded with 491 and call is disconnected.

This PR introduce compile-time `PJSUA_LOCK_CODEC_DONT_USE_UPDATE` to force lock codec feature to always use INVITE method. This will also affect ICE completion update in updating default address in SDP.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -4944,6 +4944,24 @@ PJ_DECL(pj_status_t) pjsua_acc_set_transport(pjsua_acc_id acc_id,
 #   define PJSUA_MAX_VID_WINS	    16
 #endif
 
+
+/**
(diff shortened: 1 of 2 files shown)


sauwming (commented):
Since we have `pjsua_acc_cfg.lock_codec` setting and the difference is not that much, I'm wondering why don't we make this a runtime setting instead?

trengginas (approved):

nanangizz:
The issue seems to rarely happen and not sure if it is worth adding new run-time settings (at some point, the buggy endpoint should be fixed and we don't need the run-time settings anymore?).

Issue #1899, opened 2015-12-03
Create stream for inactive media to allow keep-alive and RTCP traffics


pjsipbot:
_2015-12-03 11:31:45: @nanangizz_ _created the issue on trac ticket [1899](https://trac.pjsip.org/repos/ticket/1899)_

Currently when media is inactive by SDP a=inactive line, not by SDP media port zero, stream will not be re-created (after destroyed), while transport is maintained (not destroyed), this causes no keep-alive RTP nor RTCP packet could be sent. [RFC3264 5.1](https://tools.ietf.org/html/rfc3264#section-5.1) describes that "RTCP is still sent and received for sendonly, recvonly, and inactive streams".

This ticket will change pjsua behaviour to re-create stream even when SDP direction attribute is inactive (as long as media port is not zero).

Thank you Marcus Froeschl for the report.
___
_2015-12-03 11:33:26: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2015-12-03 11:33:26: @nanangizz_ _set **resolution** to **fixed**_

___
_2015-12-03 11:33:26: @nanangizz_ _commented_

In [r5205](../commit/051099f3ab7172cadc7776f391827b6725292c9d):
Fix #1899: Updated pjsua behaviour to re-create stream even when SDP direction attribute is inactive, as long as media port is not zero.

PR #3237, opened 2022-09-15
Update media state when SDP contain "a=inactive" attribute

trengginas:
SDP may specify `a=inactive` with media port set to non zero.

```
v=0
o=- 3872230549 3872230550 IN IP4 192.168.1.121
s=pjmedia
b=AS:84
t=0 0
a=X-nat:0
m=audio 4002 RTP/AVP 0 120
c=IN IP4 192.168.1.121
b=TIAS:64000
a=rtcp:4003 IN IP4 192.168.1.121
a=rtpmap:0 PCMU/8000
a=rtpmap:120 telephone-event/8000
a=fmtp:120 0-16
a=ssrc:2705443 cname:39b32d12074d4dc8
a=inactive
```
Consider the scenario:
1. Receive initial INVITE with a=inactive.
1. User does local Hold, sends a=sendonly and receives a=inactive.
1. a=inactive is received from remote side for Hold.

For the side which receives the 'a=inactive':
On 1 & 3, the media state to `PJSUA_CALL_MEDIA_ACTIVE`.
On 2, the media state to `PJSUA_CALL_MEDIA_LOCAL_HOLD`.

This patch will update the media state when a=inactive is received, also when it is sent (e.g: by specifying media dir using `PJSUA_CALL_SET_MEDIA_DIR`).

The update media state:
- `PJSUA_CALL_MEDIA_LOCAL_HOLD` 
   Either the user holds the call, or sets the media direction that requests the remote party to stop sending media (i.e. sendonly or inactive).
- `PJSUA_CALL_MEDIA_REMOTE_HOLD`
   Remote doesn't want us to send media (recvonly or inactive) and we don't set media dir that locally holds the media.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -3983,8 +3983,7 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 	    }
 
 	    /* Check if no media is active */
-	    if ((local_sdp->media[mi]->desc.port == 0) || 
-	        (si->dir == PJMEDIA_DIR_NONE))	    
+            if (local_sdp->media[mi]->desc.port == 0)
 	    {
 
 		/* Update call media state and direction */
@@ -3999,12 +3998,27 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 		call_med->dir = si->dir;
 
 		/* Call media state */
-		if (call->local_hold)
+                if (call->local_hold) {
 		    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
-		else if (call_med->dir == PJMEDIA_DIR_DECODING)
+		} else if (call_med->dir == PJMEDIA_DIR_DECODING) {
 		    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
-		else
+		} else if (call_med->dir == PJMEDIA_DIR_NONE) {
+		    pj_bool_t remote_hold = PJ_FALSE;
+
+		    if (pjmedia_sdp_media_find_attr2(remote_sdp->media[mi], 
+						     "inactive", NULL) ||
+			pjmedia_sdp_media_find_attr2(remote_sdp->media[mi], 
+						     "sendonly", NULL)) 
+                    {
+                        remote_hold = PJ_TRUE;
+                    }
+		    
+		    call_med->state = remote_hold? 
+						PJSUA_CALL_MEDIA_REMOTE_HOLD:
+					        PJSUA_CALL_MEDIA_NONE;
+                } else {
 		    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
+		}
 
 		if (call->inv->following_fork) {
 		    unsigned options = (call_med->enable_rtcp_mux?
@@ -4231,12 +4245,27 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 		call_med->dir = si->dir;
 
 		/* Call media state */
-		if (call->local_hold)
+                if (call->local_hold) {
 		    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
-		else if (call_med->dir == PJMEDIA_DIR_DECODING)
+		} else if (call_med->dir == PJMEDIA_DIR_DECODING) {
 		    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
-		else
+		} else if (call_med->dir == PJMEDIA_DIR_NONE) {
+		    pj_bool_t remote_hold = PJ_FALSE;
+
+		    if (pjmedia_sdp_media_find_attr2(remote_sdp->media[mi], 
+						     "inactive", NULL) ||
+			pjmedia_sdp_media_find_attr2(remote_sdp->media[mi], 
+						     "sendonly", NULL)) 
+                    {
+                        remote_hold = PJ_TRUE;
+                    }
+		    
+		    call_med->state = remote_hold? 
+						PJSUA_CALL_MEDIA_REMOTE_HOLD:
+					        PJSUA_CALL_MEDIA_NONE;
+                } else {
 		    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
+		}
 
 		/* Start/restart media transport */
 		status = pjmedia_transport_media_start(call_med->tp,
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -4229,7 +4229,7 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 	    }
 
 	    /* Check if no media is active */
-	    if (si->dir == PJMEDIA_DIR_NONE) {
+	    if (local_sdp->media[mi]->desc.port == 0) {
 
 		/* Update call media state and direction */
 		call_med->state = PJSUA_CALL_MEDIA_NONE;
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -3983,7 +3983,8 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 	    }
 
 	    /* Check if no media is active */
-            if (local_sdp->media[mi]->desc.port == 0)
+	    if ((local_sdp->media[mi]->desc.port == 0) || 
+	        (si->dir == PJMEDIA_DIR_NONE))	    
 	    {
 
 		/* Update call media state and direction */
@@ -3998,27 +3999,12 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 		call_med->dir = si->dir;
 
 		/* Call media state */
-                if (call->local_hold) {
+		if (call->local_hold)
 		    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
-		} else if (call_med->dir == PJMEDIA_DIR_DECODING) {
+		else if (call_med->dir == PJMEDIA_DIR_DECODING)
 		    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
-		} else if (call_med->dir == PJMEDIA_DIR_NONE) {
-		    pj_bool_t remote_hold = PJ_FALSE;
-
-		    if (pjmedia_sdp_media_find_attr2(remote_sdp->media[mi], 
-						     "inactive", NULL) ||
-			pjmedia_sdp_media_find_attr2(remote_sdp->media[mi], 
-						     "sendonly", NULL)) 
-                    {
-                        remote_hold = PJ_TRUE;
-                    }
-		    
-		    call_med->state = remote_hold? 
-						PJSUA_CALL_MEDIA_REMOTE_HOLD:
-					        PJSUA_CALL_MEDIA_NONE;
-                } else {
+		else
 		    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
-		}
 
 		if (call->inv->following_fork) {
 		    unsigned options = (call_med->enable_rtcp_mux?
@@ -4245,27 +4231,12 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 		call_med->dir = si->dir;
 
 		/* Call media state */
-                if (call->local_hold) {
+		if (call->local_hold)
 		    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
-		} else if (call_med->dir == PJMEDIA_DIR_DECODING) {
+		else if (call_med->dir == PJMEDIA_DIR_DECODING)
 		    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
-		} else if (call_med->dir == PJMEDIA_DIR_NONE) {
-		    pj_bool_t remote_hold = PJ_FALSE;
-
-		    if (pjmedia_sdp_media_find_attr2(remote_sdp->media[mi], 
-						     "inactive", NULL) ||
-			pjmedia_sdp_media_find_attr2(remote_sdp->media[mi], 
-						     "sendonly", NULL)) 
-                    {
-                        remote_hold = PJ_TRUE;
-                    }
-		    
-		    call_med->state = remote_hold? 
-						PJSUA_CALL_MEDIA_REMOTE_HOLD:
-					        PJSUA_CALL_MEDIA_NONE;
-                } else {
+		else
 		    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
-		}
 
 		/* Start/restart media transport */
 		status = pjmedia_transport_media_start(call_med->tp,
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -3983,9 +3983,7 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 	    }
 
 	    /* Check if no media is active */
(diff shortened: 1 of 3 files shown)


pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3983,7 +3983,9 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 	    }
 
 	    /* Check if no media is active */
-	    if (local_sdp->media[mi]->desc.port == 0) {
+	    if ((local_sdp->media[mi]->desc.port == 0) || 
+	        (si->dir == PJMEDIA_DIR_NONE))	    
+	    {
```
nanangizz, line 3988: This seems to introduce rather major behavioral change. Note that port zero and a=inactive in SDP are not really the same. When the port is zero, media is completely disabled (no media transport is created), otherwise it still needs to have media transport, e.g: RTP keep alive may still active, RTCP too.

trengginas: In the above case (`a=inactive`), when `on_call_media_state()` is called the media state will be set to `PJSUA_CALL_MEDIA_ACTIVE`. So, app needs to check for the call media dir in addition to the media state.
e.g: `media dir = PJMEDIA_DIR_NONE` indicates `a=inactive`.

nanangizz: > In the above case (`a=inactive`), when `on_call_media_state()` is called the media state will be set to `PJSUA_CALL_MEDIA_ACTIVE`. So, app needs to check for the call media dir in addition to the media state. e.g: `media dir = PJMEDIA_DIR_NONE` indicates `a=inactive`.

Is this the current behavior or the new behavior after the patch?

Anyway, my concern is that media transports should remain active (for keep-alive & RTCP) on `a=inactive` scenario, not destroyed as when SDP port is zero.

trengginas: I'm pointing out the current behavior (call has `a=inactive` scenario).
Users usually only checks for media state when call is connected to perform conf connect.
In the case of `a=inactive`, the conf connect is not as expected although the media is active.
Checking for media dir will also be required besides the media state check.

nanangizz: Okay. So as mentioned before, my concern on this PR patch, will the media transport now destroyed when `a=inactive`?

trengginas: With this patch the media transport (RTP/RTCP) will still be created, not destroyed however it will not be started. 
For `keep-alive` it will not change the behavior. It is part of the stream, and in this case the stream is not created.

Another alternative would be to set the media state to PJSUA_CALL_MEDIA_NONE:
```
if (call->local_hold)
    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
else if (call_med->dir == PJMEDIA_DIR_DECODING)
    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
else if (call_med->dir == PJMEDIA_DIR_NONE)
    call_med->state = PJSUA_CALL_MEDIA_NONE;
else
    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
```

nanangizz: The above alternative looks much safer, i.e: only affects the call media state value, right? If it does, I'll vote for this approach :)

trengginas: Yes, it should only affects the call media state.

sauwming: What about the double hold scenario?
Do we want the call media state to be `PJSUA_CALL_MEDIA_LOCAL_HOLD` or `PJSUA_CALL_MEDIA_NONE`?

trengginas: I think we should preserve the existing state (keep it `PJSUA_CALL_MEDIA_LOCAL_HOLD`).
It would prevent app breaking which uses that state.

trengginas: Adding to the stream creation. The stream will be created if media changes is detected. e.g: on the case of call hold, media changes from `sendrecv` -> `sendonly`. The stream will be recreated. However on initial call if the SDP contains `inactive`, then stream will not be created since no media change occurs ([ref](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_media.c#L4069)). 
I agree that the original patch will introduce major behavior change. So it would be safe to discard this solution.

alternative 1:
```
if (call->local_hold)
    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
else if (call_med->dir == PJMEDIA_DIR_DECODING)
    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
else if (call_med->dir == PJMEDIA_DIR_NONE)
    call_med->state = PJSUA_CALL_MEDIA_NONE;
else
    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
```

alternative 2:
```
if (call->local_hold)
    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
else if (call_med->dir == PJMEDIA_DIR_DECODING)
    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
else if (call_med->dir== PJMEDIA_DIR_NONE)
    /* If remote sends `a=inactive`, we'll treat it as a remote hold */
    call_med->state =  is_answerer? PJSUA_CALL_MEDIA_REMOTE_HOLD : PJSUA_CALL_MEDIA_NONE;
    /* ==== end of new lines ====*/
else
    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
```

About the purpose patch.
scenario 1: (call hold - no issue)
caller sends `sendonly`
callee sends `recvonly`

| Patch | Caller | Callee |
| ----------- | ----------- | ----------- |
| No patch     |  PJSUA_CALL_MEDIA_LOCAL_HOLD   | PJSUA_CALL_REMOTE_HOLD |
| alt 1     |  PJSUA_CALL_MEDIA_LOCAL_HOLD   | PJSUA_CALL_REMOTE_HOLD |
| alt 2     |  PJSUA_CALL_MEDIA_LOCAL_HOLD   | PJSUA_CALL_REMOTE_HOLD |

scenario 2: 
caller sends `sendrecv`
callee sends `inactive`

| Patch | Caller | Callee |
| ----------- | ----------- | ----------- |
| No patch     |  PJSUA_CALL_MEDIA_ACTIVE  | PJSUA_CALL_MEDIA_ACTIVE  |
| alt 1     |  PJSUA_CALL_MEDIA_NONE | PJSUA_CALL_MEDIA_NONE |
| alt 2     |  PJSUA_CALL_MEDIA_NONE   | PJSUA_CALL_MEDIA_NONE ? |

scenario 3: 
caller sends `sendonly`
callee sends `inactive`

| Patch | Caller | Callee |
| ----------- | ----------- | ----------- |
| No patch     |  PJSUA_CALL_MEDIA_LOCAL_HOLD   | PJSUA_CALL_MEDIA_ACTIVE  |
| alt 1     |  PJSUA_CALL_MEDIA_LOCAL_HOLD | PJSUA_CALL_MEDIA_NONE |
| alt 2     |  PJSUA_CALL_MEDIA_LOCAL_HOLD | PJSUA_CALL_REMOTE_HOLD |

For alt 2, we need to check if the INVITE message contains `sendonly` or `sendrecv`.

trengginas: For alternative 2:
If we set media state based on remote SDP has `a=inactive`, then we should also handle if local is the one that sent `a=inactive`. e.g: should the offerer set to PJSUA_CALL_MEDIA_LOCAL_HOLD?
Because of this, I would prefer the alt 1.

nanangizz: The idea of alt 2 is based on the assumption of that some implementations (not PJSUA, such as in the original report) send `inactive` (instead of `sendonly`) when initiating call hold. That's why the alt 2 only checks if we are the answerer. It is just an attempt to handle such implementations. It of course can be improved to handle other scenario.

Note that PJSUA by default (note that app may override the SDP via PJSUA callback) will never send `inactive` for initiating call hold (unless remote initates call hold too).

sauwming: Btw, PJSUA now actually has the API/capability to set media direction (without having to modify the SDP), i.e. using `PJSUA_CALL_SET_MEDIA_DIR` flag, for example:
```
    call_opt.media_dir[0] = PJMEDIA_DIR_NONE;
    call_opt.flag = PJSUA_CALL_SET_MEDIA_DIR;
    pjsua_call_reinvite2(...,&call_opt,...);
```
So we probably need to take this into consideration as well.

trengginas: Right, the `PJSUA_CALL_SET_MEDIA_DIR` needs to be handled as well. e.g: for call hold, the `local_hold` field will only be set when we call `pjsua_call_set_hold()/pjsua_call_set_hold2()`. We can set the local_hold field on pjsua_call_make_call()/pjsua_call_reinvite()/pjsua_call_update() when `PJSUA_CALL_SET_MEDIA_DIR + PJMEDIA_DIR_ENCODING` is set.

Back to alternative 2. If we are going to set `PJSUA_CALL_MEDIA_REMOTE_HOLD` when remote sent `a=inactive`, then 
it would be expected that local would be set to `PJSUA_CALL_MEDIA_LOCAL_HOLD`. This means the `PJSUA_CALL_MEDIA_NONE` will no longer be used. Since the `a=inactive` can be issued by the offerer/answerer.
```
if (call->local_hold)
    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
else if (call_med->dir == PJMEDIA_DIR_DECODING)
    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
else if (call_med->dir== PJMEDIA_DIR_NONE)
    /* If remote sends `a=inactive`, we'll treat it as a remote hold */
    call_med->state =  is_answerer? PJSUA_CALL_MEDIA_REMOTE_HOLD : PJSUA_CALL_MEDIA_LOCAL_HOLD;
    /* ==== end of new lines ====*/
else
    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
```
`PJSUA_CALL_MEDIA_NONE` is not used anymore and might break app checking for such state.

For the original report, the issue with call hold is caused by the original patch
```
if ((local_sdp->media[mi]->desc.port == 0) || 
	        (si->dir == PJMEDIA_DIR_NONE))	    
{
		/* Update call media state and direction */
		call_med->state = PJSUA_CALL_MEDIA_NONE;
		call_med->dir = PJMEDIA_DIR_NONE;
}
```
This will break the current call hold scenario. e.g: app expects `PJSUA_CALL_LOCAL_HOLD`, but `PJSUA_CALL_MEDIA_NONE`.
It didn't expect to receive new state e.g: `PJSUA_CALL_REMOTE_HOLD` when receiving `a=inactive`.

sauwming:
Apparently `PJSUA_CALL_SET_MEDIA_DIR` can make things complicated, so let's try to simplify it first.

From [RFC 3264 section 8.4](https://www.rfc-editor.org/rfc/rfc3264#section-8.4):
```
   If a party in a call wants to put the other party "on hold", i.e.,
   request that it temporarily stops sending one or more unicast media
   streams, a party offers the other an updated SDP.

   If the stream to be placed on hold was previously a sendrecv media
   stream, it is placed on hold by marking it as sendonly.  If the
   stream to be placed on hold was previously a recvonly media stream,
   it is placed on hold by marking it inactive.
...
   Typically, when a user "presses" hold, the agent will generate an
   offer with all streams in the SDP indicating a direction of sendonly,
   and it will also locally mute, so that no media is sent to the far
   end, and no media is played out.
```
From my understanding, there can be two types of hold:
- media hold, that affects a particular media stream, or
- call hold, which *typically* applies to **all** streams.

So we don't need to mark the entire call to be on hold when one media direction is set as `sendonly`. Also, to keep things simple, let's not worry about the hold/unhold mechanism with `PJSUA_CALL_SET_MEDIA_DIR` for now, unless you notice that something is broken.

IMO, we can go back to the initial patch, which is already difficult enough to get right. If we take into account `PJSUA_CALL_SET_MEDIA_DIR`, which is stored in `call_med->def_dir`, I think the patch should be like this:
```
		/* Call media state */
		if (call->local_hold ||
		    ((call_med->dir & PJMEDIA_DIR_DECODING) == 0 &&
		     (call_med->def_dir & PJMEDIA_DIR_DECODING) == 0))
		{
		    /* Local hold: Either the user holds the call, or sets
		     * the media direction that requests the remote party to
		     * stop sending media (i.e. sendonly or inactive).
		     */
		    call_med->state = PJSUA_CALL_MEDIA_LOCAL_HOLD;
		} else if ((call_med->dir & PJMEDIA_DIR_ENCODING) == 0 &&
		           (call_med->def_dir & PJMEDIA_DIR_DECODING) != 0)
		{
		    /* Remote hold: Remote doesn't want us to send media
		     * (recvonly or inactive) and we don't set media dir that
		     * locally holds the media.
		     */
		    call_med->state = PJSUA_CALL_MEDIA_REMOTE_HOLD;
/* This is no longer necessary and can be safely removed.
		} else if (call_med->dir == PJMEDIA_DIR_NONE) {
		    call_med->state = PJSUA_CALL_MEDIA_NONE;
***********************************************************/
		} else {
		    call_med->state = PJSUA_CALL_MEDIA_ACTIVE;
		}
```

There are 3 scenarios reported by the user:
1. Initial INVITE with `a=inactive`.
2. User does local Hold, sends `a=sendonly` and receives` a=inactive`.
3. `a=inactive` is received from remote side for Hold.

For 1 & 3, the media state will change from `PJSUA_CALL_MEDIA_ACTIVE` (without the patch) to `PJSUA_CALL_MEDIA_REMOTE_HOLD` (with the patch).
For 2, the media state will be unchanged with and w/o patch: `PJSUA_CALL_MEDIA_LOCAL_HOLD`.

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3983,9 +3983,7 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
 	    }
 
 	    /* Check if no media is active */
-	    if ((local_sdp->media[mi]->desc.port == 0) || 
-	        (si->dir == PJMEDIA_DIR_NONE))	    
-	    {
+	    if (local_sdp->media[mi]->desc.port == 0) {
```
trengginas, line 3986: This change was introduce in https://github.com/pjsip/pjproject/issues/1899. However it is implemented for audio only. I think we should handle the same way for video.

trengginas:
[1075e8b](https://github.com/pjsip/pjproject/pull/3237/commits/1075e8ba622eafcb06ffa2d798c8b59c7434335f) implements the suggestion by @sauwming. I agree with his points and made the changes simple.
Have tested here locally and works as expected.
Note:
- The media state will updated correctly when setting media dir using `PJSUA_CALL_SET_MEDIA_DIR`, e.g: setting to `PJMEDIA_DIR_ENCODING` will set the media state to `PJSUA_CALL_MEDIA_REMOTE_HOLD`. However, app is expected to change the media  dir manually and mixing it with `pjsua_call_update()`/`pjsua_call_reinvite()` to unhold (using `PJSUA_CALL_UNHOLD`) will not work.

pjsip/src/pjsua-lib/pjsua_media.c
sauwming: Good point. What do you think, @nanangizz ?

nanangizz: Agree, I think the video should be treated the same way (in `port=0` vs `a=inactive` scenarios).

sauwming (approved):
Looks okay to me.

Test the video part to ensure everything is okay, update the PR title and desc, and inform the reporter.

nanangizz (approved):

PR #3241, opened 2022-09-16
Doc fixes, default value of Media Quality is 8

SebaLedesma:


SebaLedesma:
I was analyzing how the --quality param works, specially for opus, and detected this small inconsistency.
Please note that the pjsua user manual (https://www.pjsip.org/pjsua.htm) defines the default quality as 6 but in the source code it's actually 8.

The addittional/future goal it's to add "audio quality N" command to the CLI and so allow to change the media quality in runtime by using the new pjmedia_stream_modify_codec_param added in PR 3189 (https://github.com/pjsip/pjproject/pull/3189).

sauwming (approved):

nanangizz (approved):
Alternatively, perhaps just remove the number, so it won't cause inconsistency in the future if `PJSUA_DEFAULT_CODEC_QUALITY` value is changed (again) and we forget to also update it here.

sauwming:
The new additional command seems to be a separate issue, so better put it in a separate PR.

SebaLedesma:
Yes, that was my intention (to create an indepedent PR). Let me see how to fix it.

trengginas (approved):

PR #3242, opened 2022-09-18
Support ffmpeg 5.1

trengginas:
FFMPEG 5.1 has some changes which will break current build.
The changes mostly related to removing some APIs that were already deprecated:
1. `AVCodec::encode2()/decode()` removed [here](https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/20f972701806be20a77f808db332d9489343bb78).
2. `av_codec_next()/avcodec_register_all()` removed [here](https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/4b2be3f8d45db8fa795233b938391b23f6c518ea )
3. `avcodec_encode_video2()/avcodec_decode_video2()` removed [here](https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/7c1f347b184b6738abdc22fdcda40baa9f932522)

This ticket will fix the issue by using the updated method. 
Tested using FFMPEG 4.2 and FFMPEG 5.1

### DIFF
--- a/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
+++ b/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
@@ -1654,7 +1654,7 @@ static pj_status_t ffmpeg_codec_encode_whole(pjmedia_vid_codec *codec,
         if (pkt) {
             while (err >= 0) {
                 err = avcodec_receive_packet(ff->enc_ctx, pkt);
-                if (err == AVERROR_EOF) {
+                if (err == AVERROR(EAGAIN) || err == AVERROR_EOF) {
                     err = out_size;
                     break;
                 }
--- a/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
+++ b/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
@@ -928,7 +928,7 @@ PJ_DEF(pj_status_t) pjmedia_codec_ffmpeg_vid_init(pjmedia_vid_codec_mgr *mgr,
 
     /* Enum FFMPEG codecs */
     for (c=av_codec_next(NULL); c; c=av_codec_next(c)) {
-        init_codec(c, AV_HAS_ENCODE(c), AV_HAS_DECODE(c));
+        init_codec(c, AVCODEC_HAS_ENCODE(c), AVCODEC_HAS_DECODE(c));
     }
 
 #endif
--- a/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
+++ b/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
@@ -59,17 +59,15 @@
 #endif
 
 #if LIBAVCODEC_VER_AT_LEAST(53,61)
-#  if LIBAVCODEC_VER_AT_LEAST(59,25)
-#      define AVCODEC_HAS_ENCODE(c)	(is_encoder)
-#  elif LIBAVCODEC_VER_AT_LEAST(54,59)
+#  if LIBAVCODEC_VER_AT_LEAST(54,59)
    /* Not sure when AVCodec::encode is obsoleted/removed. */
-#      define AVCODEC_HAS_ENCODE(c)	(c->encode2)
+#    define AVCODEC_HAS_ENCODE(c)	(c->encode2)
 #  else
    /* Not sure when AVCodec::encode2 is introduced. It appears in 
     * libavcodec 53.61 where some codecs actually still use AVCodec::encode
     * (e.g: H263, H264).
     */
-#      define AVCODEC_HAS_ENCODE(c)	(c->encode || c->encode2)
+#    define AVCODEC_HAS_ENCODE(c)	(c->encode || c->encode2)
 #  endif
 #  define AV_OPT_SET(obj,name,val,opt)	(av_opt_set(obj,name,val,opt)==0)
 #  define AV_OPT_SET_INT(obj,name,val)	(av_opt_set_int(obj,name,val,0)==0)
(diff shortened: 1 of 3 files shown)


pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
```diff
@@ -1861,7 +1911,18 @@ static pj_status_t ffmpeg_codec_decode_whole(pjmedia_vid_codec *codec,
     avpacket.flags = 0;
 #endif
 
-#if LIBAVCODEC_VER_AT_LEAST(52,72)
+#if LIBAVCODEC_VER_AT_LEAST(58,137)
+    err = avcodec_send_packet(ff->dec_ctx, &avpacket);
+    if (err >= 0) {
+        err = avcodec_receive_frame(ff->dec_ctx, &avframe);
+        if (err == AVERROR(EAGAIN) || err == AVERROR_EOF)
```
sauwming, line 1918: according to https://ffmpeg.org/doxygen/3.4/group__lavc__decoding.html#ga11e6542c4e66d3028668788a1a74217c:
"[AVERROR(EAGAIN)](https://ffmpeg.org/doxygen/3.4/group__lavu__error.html#gae4bb6f165973d09584e0ec0f335f69ca): output is not available in this state - user must try to send new input AVERROR_EOF: the decoder has been fully flushed, and there will be no more output frames"

so is it correct that got_picture is true for this case?

pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
```diff
@@ -732,6 +734,140 @@ static int find_codec_idx_by_fmt_id(pjmedia_format_id fmt_id)
     return -1;
 }
 
+static void init_codec(const AVCodec *c, pj_bool_t is_encoder)
```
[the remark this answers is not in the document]
trengginas: It is used in `AVCODEC_HAS_DECODE()/AVCODEC_HAS_ENCODE()` macro. The old behavior is to iterate the codec and check if it supports encode2()/decode(). Now we create the encoder/decoder based on the Id.

pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
trengginas: I think you are right

sauwming (commented):
First, I just want to confirm whether AVERROR(EAGAIN) is positive or negative?
(Based on the macro definition in https://ffmpeg.org/doxygen/4.0/group__lavu__error.html#gae4bb6f165973d09584e0ec0f335f69ca, it seems it should be positive).

If it's positive, for encoding, I think the previous version was already correct, based on https://ffmpeg.org/doxygen/4.0/group__lavc__encdec.html
"Receive output in a loop. Periodically call one of the avcodec_receive_*() functions and process their output:
- For decoding, call [avcodec_receive_frame()](https://ffmpeg.org/doxygen/4.0/group__lavc__decoding.html#ga11e6542c4e66d3028668788a1a74217c). On success, it will return an [AVFrame](https://ffmpeg.org/doxygen/4.0/structAVFrame.html) containing uncompressed audio or video data.
- For encoding, call [avcodec_receive_packet()](https://ffmpeg.org/doxygen/4.0/group__lavc__decoding.html#ga5b8eff59cf259747cf0b31563e38ded6). On success, it will return an [AVPacket](https://ffmpeg.org/doxygen/4.0/structAVPacket.html) with a compressed frame. Repeat this call until it returns [AVERROR(EAGAIN)](https://ffmpeg.org/doxygen/4.0/group__lavu__error.html#gae4bb6f165973d09584e0ec0f335f69ca) or an error. The [AVERROR(EAGAIN)](https://ffmpeg.org/doxygen/4.0/group__lavu__error.html#gae4bb6f165973d09584e0ec0f335f69ca) return value means that new input data is required to return new output. In this case, continue with sending input. For each input frame/packet, the codec will typically return 1 output frame/packet, but it can also be 0 or more than 1."

Then based on the doc, it seems that AVERROR_EOF can only happen during flushing mode (i.e. if we send NULL to the [avcodec_send_packet()](https://ffmpeg.org/doxygen/4.0/group__lavc__decoding.html#ga58bc4bf1e0ac59e27362597e467efff3) (decoding) or [avcodec_send_frame()](https://ffmpeg.org/doxygen/4.0/group__lavc__decoding.html#ga9395cb802a5febf1f00df31497779169) (encoding) functions).
So the AVERROR_EOF check can probably be safely removed, but leaving it should be okay and harmless.

trengginas:
> 

It should be positive, but printing the value to log. It prints `-11`.
However, the check AVERROR(EAGAIN) for encoding should already be correct. 
It breaks the loop when there's no other output. If there's no check, the encoding will be detected as error.
As for AVERROR_EOF check, I'm going to leave it.

sauwming (approved):
Make sure to do a video call retest and check if everything is okay.

pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
```diff
@@ -63,7 +59,9 @@
 #endif
 
 #if LIBAVCODEC_VER_AT_LEAST(53,61)
-#  if LIBAVCODEC_VER_AT_LEAST(54,59)
+#  if LIBAVCODEC_VER_AT_LEAST(59,25)
+#      define AVCODEC_HAS_ENCODE(c)	(is_encoder)
```
nanangizz, line 63: Minor thing, this macro looks rather ugly as `is_encoder` is actually a local var :)
If only init_codec() uses the macros, what if move this AVCODEC_HAS_ENCODE/DECODE #def right-before the function and #undef later (below the function)?

nanangizz (approved):

PR #3246, opened 2022-09-19
Open/Create files on Windows with FILE_SHARE_DELETE flag

andreas-wehrmann:
This PR adds the FILE_SHARE_DELETE flag to the CreateFile() call when opening/creating files on Windows.
Setting this flag allows i.e. for using temporary files with the WAV player/recorder.

My concrete use case looks like this:
- User wants to play a WAV file
- WAV file has to be downloaded via HTTP
- HTTP download will create a temporary file (with the flags FILE_SHARE_DELETE and FILE_FLAG_DELETE_ON_CLOSE)
 (This ensures the file will be removed when the last handle to the file is closed).
- When download is finished, the file handle is kept open
- Create WAV player, set EOF callback, play WAV file
- When EOF hits: Destroy WAV player and close file handle -> file is removed

Of course there could be other use cases like text-to-speech, where the file is not downloaded but generated locally.
One must make sure, that temporary/generated files are always cleaned up.
Using it like I described above makes this easy.

I don't think there is any harm in adding this flag in general, is there?
The behaviour is now closer to what it's like on Linux where the directory entry
of a file can be moved around or deleted altogether while it's still in use.

### DIFF
--- a/pjlib/src/pj/file_io_win32.c
+++ b/pjlib/src/pj/file_io_win32.c
@@ -119,7 +119,7 @@ PJ_DEF(pj_status_t) pj_file_open( pj_pool_t *pool,
         return PJ_EINVAL;
     }
 
-    dwShareMode = FILE_SHARE_READ | FILE_SHARE_WRITE;
+    dwShareMode = FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE;
     
     dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL;
 


sauwming:
Hmm.... I find some long discussions about this on the net and it seems to be more complicated than I expected.

From what I read (which I haven't tested it myself), if you specify the flag, then: `Subsequent open operations on the object will succeed only if delete access is requested.`

So from my understanding, before the patch:
2 PJSIP instances can call `pj_file_open()` to the same file.

But after the patch:
PJSIP with the patch will successfully open, but if there's an older version of PJSIP trying to subsequently open the same file, then it will fail?

If this is true, then it's definitely a major downside.

andreas-wehrmann:
So I have tried different combinations now, the simplest of which
is an unpatched PJ against a patched version with FILE_SHARE_DELETE.
For testing I used the pjsua app with the "--play-file" option.
What I found is, that it doesn't seem to matter in which order I start the applications,
they're always able to open and play the file simultaneously.

And that is what I would expect from the MSDN description.
FILE_SHARE_DELETE only _enables_ subsequent operations to _request_ delete access (which pj_file_open() doesn't do).
So this should be fine.

What doesn't work is, when you open/create a file where you request delete access
(like in my use case, where I create a new file with the file attribute FILE_FLAG_DELETE_ON_CLOSE)
and then open the file again without FILE_SHARE_DELETE.

If I don't overlook something, the only thing that FILE_SHARE_DELETE does here is to widen the operations allowed on the file.
From what I understand, FILE_SHARE_DELETE is only mandatory if an actual delete/rename operation is involved
but not when you're simply doing I/O.

sauwming (approved):
Quite frankly, I'm not 100% sure about this, considering that the same proposal was rejected for some other projects, such as https://github.com/golang/go/issues/32088

But in PJSIP, perhaps the risk is low anyway to introduce such flag, and the benefit is higher, so I'm leaning towards "okay". Let's see what others think about this.

trengginas (approved):

PR #3259, opened 2022-09-29
ioqueue poll logic optimization

jimying:
ioqueue poll logic optimization:  Reduce the number of cycles

 - one event readable/writeable/error flag,  can be record & distinguish by event_type
   multiple event changes (eg. readable and writeable) can record and process (later) at once.
 - remove the loop that call pj_grp_lock_add_ref_dbg   (move to first step when mark event as changed)

### DIFF
--- a/pjlib/src/pj/ioqueue_epoll.c
+++ b/pjlib/src/pj/ioqueue_epoll.c
@@ -833,7 +833,8 @@ PJ_DEF(int) pj_ioqueue_poll( pj_ioqueue_t *ioqueue, const pj_time_val *timeout)
     struct epoll_event events[MAX_EVENTS];
     struct queue queue[MAX_EVENTS];
     pj_timestamp t1, t2;
-    
+    pj_ioqueue_key_t *h;
+
     PJ_CHECK_STACK();
 
     msec = timeout ? PJ_TIME_VAL_MSEC(*timeout) : 9000;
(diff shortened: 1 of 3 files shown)


sauwming:
Thanks a lot for the patch.

To be honest with you, it's unlikely to be accepted though. Not because it's not good or anything, on the contrary, it does look impressive. However, it is impossible to verify its correctness due to its size (for us, sometimes even a patch that only changes a few lines of code can take weeks-long discussion and countless iterations). And since it affects ioq which is a very critical component, even the tiniest mistake can be fatal.

If possible, I would suggest to break the patch down to the smallest part possible and hopefully, it can make the review process easier and less daunting. For example, start with one ioq backend first, then it seems that there are two separate parts, event checking and dispatch loop. Choose one and describe in detail what the optimisation does and how it can improve the performance.

sauwming:
I managed to go through the change in `ioq_select` and here are my comments:
- There seems to be a fundamental modification in the meaning of an event. Previously, one ioq key can have multiple events, but now only one. So `PJ_IOQUEUE_MAX_EVENTS_IN_SINGLE_POLL` and the return value of `pj_ioqueue_poll()` (`processed_cnt`) now also have different meanings.
- The optimisation doesn't appear to be much. Okay the event array is now smaller and the dispatch loop thus has less iterations, but speed/performance-wise, it should be relatively the same, no?

jimying:
> I managed to go through the change in `ioq_select` and here are my comments:
> 
> * There seems to be a fundamental modification in the meaning of an event. Previously, one ioq key can have multiple events, but now only one. So `PJ_IOQUEUE_MAX_EVENTS_IN_SINGLE_POLL` and the return value of `pj_ioqueue_poll()` (`processed_cnt`) now also have different meanings.
> * The optimisation doesn't appear to be much. Okay the event array is now smaller and the dispatch loop thus has less iterations, but speed/performance-wise, it should be relatively the same, no?


- process_cnt
  if one fd has read and write event,   old code will treat as two,  but new treat as one
- optmisation
   1.  smaller loop count as you say above
   2. remove  the loop to do: pj_grp_lock_add_ref_dbg()
   3. when all socks are busy,  can do more in one poll action

sauwming:
The behavioral change is a major issue, so we'd rather leave it as it is now.

sauwming:
The behavioral change is a major issue, so we'd rather leave it as it is now.

PR #3263, opened 2022-10-10
Fix the build with musl

orbea:
The `execinfo.h` header is GNU specific and is not available with musl.

This commit is based on a patch from Alpine Linux.

https://git.alpinelinux.org/aports/tree/main/pjproject/execinfo.patch?h=3.16-stable

Gentoo Issue: https://bugs.gentoo.org/867343


orbea:
Checking for `PJ_DARWINOS` may be redundant here? I don't have any way to test that.

jimying:
why not use autoconf to check the exists of execinfo.h,  and set a macro PJ_HAS_EXECINFO_H ?

(in aconfigure.ac/aconfigure  and os_auto.h.in)

orbea:
@jimying Your patch works on my end just fine and perhaps is more future proof than mine. If you would like to make a PR for that which closes mine that is fine with me. :)

sauwming (approved):

nanangizz (approved):

PR #3275, opened 2022-10-23
Rust bindings for pjsua

JadKHaddad:
A simple rust program that shows how to use pjsua in Rust.

1. Install Rust: https://www.rust-lang.org/tools/install
2. Install PJSIP:
```
./configure --enable-shared
make dep & make
sudo make install
```
3. Build package:
```
cd pjsip-apps/src/rust/pjsua
cargo build
```
4. (Optional) To run sample app:
From `{your_pjproject_dir}/pjsip-apps/src/rust/demo`:
```
cargo run
```


leasual:
@JadKHaddad Hi can i ask you a question? how to build for android and iOS? Thanks a lot.

JadKHaddad:
@leasual I never used pjsua on Android or IOS. But I would suggest, you read the guides if you haven't already.

leasual:
@JadKHaddad Yes, i already read the offical guides, maybe i question is not clear, i want to use rust to call pjsua to implement some business logic, then use cargo to build for android and iOS, cargo build is success, but when i depend on this lib to build app in xcode, it failed, the reason is Undefined symbol to point pjsua function. And i change link to use static lib in binding.rs, it is the same result, so do you have some advise? many thanks.

sauwming:
It would be great if you can provide a step-by-step instruction on how to use this.

I tried downloading Rust and then running `cargo build` but it didn't work (endianness not defined, which indicates that the compiler flags wasn't setup properly).

JadKHaddad:
In this example - for the sake of simplicity - no pjsip libraries would be generated. Thus pjsip has to be compiled using the known instructions (./configure --enable-shared; make dep; make; make install).  The program will create the bindings. To run the program (tested on debian), install pjsip and set LD_LIBRARY_PATH to where the libraries have been installed (usually /usr/local/lib). Then in the root directory of the example run (cargo run). A step by step guide can be found on 

https://github.com/JadKHaddad/THE-PJPROJECT

sauwming:
OK, I finally managed to get the sample app to run, but:
- I have to manually add `#define PJ_AUTOCONF 1` in `pjsip/include/pjsua-lib/pjsua.h` since I don't know how I can pass it properly to cargo build.
- There are about 6 warnings of `warning: the type ... does not permit being left uninitialized` (`pjsua_call_info`, `pjsua_config`, etc)

sauwming:
I would suggest a couple of changes:
- Add `PJ_AUTOCONF=1` in `build.rs`, i.e.:
```
        .header("../../../pjsip/include/pjsua.h")
	.clang_args(["-D", "PJ_AUTOCONF=1"])
```
Ideally, we should include the result of `pkg-config --cflags libpjproject` as described here: https://docs.pjsip.org/en/latest/get-started/posix/build_instructions.html#using-pjsip-libraries-in-your-applications. Or get the `CFLAGS` from `build.mak`, but I'm not sure if it's doable here. So at the very least, the macro `PJ_AUTOCONF` needs to be defined, otherwise the build will fail.
- Remove `Cargo.lock`, as explained here:
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html#cargotoml-vs-cargolock
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

sauwming (approved):
Amazing work for integrating pkg-config. Also, the warning is completely gone now.

JadKHaddad:
This could be a first step for creating a high level api in rust

Issue #2217, opened 2019-07-26
Omit deprecated IPv6 addresses from ICE candidates


pjsipbot:
_2019-07-26 08:43:16: @nanangizz_ _created the issue on trac ticket [2217](https://trac.pjsip.org/repos/ticket/2217)_

This ticket introduces `pj_enum_ip_interface2()` with additional option to omit deprecated IPv6 addresses, however currently it is only implemented for Linux. This ticket also modify behavior of STUN socket in enumerating local IPv6 addresses, it now skips deprecated IPv6 addresses.

Contribution from Sébastien Blin, thanks.
___
_2019-07-26 10:32:17: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2019-07-26 10:32:17: @nanangizz_ _set **owner** to **@nanangizz**_

___
_2019-07-26 10:32:17: @nanangizz_ _set **resolution** to **fixed**_

___
_2019-07-26 10:32:17: @nanangizz_ _commented_

In [r6045](../commit/3029598af6d01c0ce6b119d92942ba018e61949f):
Close #2217: Omit deprecated IPv6 addresses from ICE candidates.

PR #3276, opened 2022-10-24
Fix build warning and error for UWP

trengginas:
This PR will fix some build warning and error for UWP.

Build warning/error:
1. `warning C4005: 'LONG_MIN': macro redefinition` 
    due to changes in https://github.com/pjsip/pjproject/issues/2056
    patch : `os_winuwp.h`
1.  `error C2079: 'sin6_addr' uses undefined struct 'in6_addr'`
    due to changes in https://github.com/pjsip/pjproject/pull/2423
    patch : `socket.h`
1. `error C2057: expected constant expression` 
   due to changes in https://github.com/pjsip/pjproject/issues/2217
   patch : `ip_helper_generic.h`
1. `error LNK2019: unresolved external symbol GetModuleHandle referenced in function set_thread_display_name` 
   due to changes in https://github.com/pjsip/pjproject/pull/3156
   patch : `os_core_win32.c`
1. `error LNK2019: unresolved external symbol LocalFileTimeToFileTime referenced in function pj_time_local_to_gmt`
   due to changes in https://github.com/pjsip/pjproject/pull/2638
   patch : `pjsua_cli_uwp_comp.vcxproj` `VoipBackEnd.vcxproj`

### DIFF
--- a/pjlib/include/pj/compat/os_winuwp.h
+++ b/pjlib/include/pj/compat/os_winuwp.h
@@ -58,6 +58,7 @@
 #define PJ_HAS_SYS_TYPES_H	    0	/* Doesn't have sys/types.h */
 #define PJ_HAS_TIME_H		    1
 #define PJ_HAS_UNISTD_H		    0
+#define PJ_HAS_LIMITS_H             1
 
 #define PJ_HAS_MSWSOCK_H	    1
 #define PJ_HAS_WINSOCK_H	    0
--- a/pjlib/include/pj/compat/socket.h
+++ b/pjlib/include/pj/compat/socket.h
@@ -37,6 +37,9 @@
 #   include <ws2tcpip.h>
 #endif
 
+#if (defined(PJ_WIN32_UWP) && PJ_WIN32_UWP!=0)
+#   include <in6addr.h>
+#endif
 
 /*
  * IPv6 for Visual Studio's
--- a/pjlib/include/pj/sock.h
+++ b/pjlib/include/pj/sock.h
@@ -549,7 +549,7 @@ struct pj_sockaddr_in
  * recommended as it may cause build issues for anyone who uses
  * the macro. See #2311 for more details.
  */
-#if 0
+#if (defined(PJ_WIN32_UWP) && PJ_WIN32_UWP!=0)
 #undef s6_addr
 
 typedef union pj_in6_addr
--- a/pjlib/include/pj/sock.h
+++ b/pjlib/include/pj/sock.h
@@ -549,7 +549,7 @@ struct pj_sockaddr_in
  * recommended as it may cause build issues for anyone who uses
  * the macro. See #2311 for more details.
  */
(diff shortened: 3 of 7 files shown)


sauwming:
Nice one with all the detailed changes that caused the break, making it easier to trace.

For no. 2, is it possible to fix it without undefining `s6_addr`? Perhaps `in6_addr` is declared somewhere?

trengginas:
Right, this is unusual.
`in6_addr` is define on `in6addr.h`
```
#ifndef s6_addr
#pragma once

#pragma region Desktop Family or OneCore Family
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM)

//
// IPv6 Internet address (RFC 2553)
// This is an 'on-wire' format structure.
//
typedef struct in6_addr {
    union {
        UCHAR       Byte[16];
        USHORT      Word[8];
    } u;
} IN6_ADDR, *PIN6_ADDR, FAR *LPIN6_ADDR;

#define in_addr6 in6_addr
//
// Defines to match RFC 2553.
//
#define _S6_un      u
#define _S6_u8      Byte
#define s6_addr     _S6_un._S6_u8

//
// Defines for our implementation.
//
#define s6_bytes    u.Byte
#define s6_words    u.Word

#endif /* WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM) */
#pragma endregion

#endif
```
However `in6_addr` is not defined.

On the case `in_addr` is defined in `inaddr.h`
```
#ifndef s_addr
#pragma once

#pragma region Desktop Family or OneCore Family
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM)

//
// IPv4 Internet address
// This is an 'on-wire' format structure.
//
typedef struct in_addr {
        union {
                struct { UCHAR s_b1,s_b2,s_b3,s_b4; } S_un_b;
                struct { USHORT s_w1,s_w2; } S_un_w;
                ULONG S_addr;
        } S_un;
#define s_addr  S_un.S_addr /* can be used for most tcp & ip code */
#define s_host  S_un.S_un_b.s_b2    // host on imp
#define s_net   S_un.S_un_b.s_b1    // network
#define s_imp   S_un.S_un_w.s_w2    // imp
#define s_impno S_un.S_un_b.s_b4    // imp #
#define s_lh    S_un.S_un_b.s_b3    // logical host
} IN_ADDR, *PIN_ADDR, FAR *LPIN_ADDR;

#endif /* WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM) */
#pragma endregion

#endif
```
No issue for `in_addr`.

sauwming:
What if `in6addr.h` is directly included?

trengginas:
> What if `in6addr.h` is directly included?
👍 
That works

nanangizz (approved):

Issue #3271, opened 2022-10-19
pjsua crash when --jb-max-size=20 or more


SebaLedesma:
### Describe the bug

I'm trying to increase the jitter buffer, but when pjsua.conf contains:
--jb-max-size 20 
the application will crash when accepting a call.
Values less than 20 works fine.
The crash is due an ASSERT at pjmedia/src/pjmedia/jbuf.c, around line 672 the function:
pjmedia_jbuf_set_adaptive will trigger the assert as  min_prefetch <= max_prefetch returns false.
min_prefetch has the value of 1 while max_prefetch = 0.

The error originates at pjmedia/src/pjmedia/stream.c, at function pjmedia_stream_create
```
 if (info->jb_min_pre >= stream->codec_param.info.frm_ptime)
        jb_min_pre = info->jb_min_pre / stream->codec_param.info.frm_ptime;
    else
        //jb_min_pre = 60 / stream->codec_param.info.frm_ptime;
        jb_min_pre = 1;
```
This will be evaluated as false, so jb_min_pre  will receive 1 as value.

```
  if (info->jb_max_pre >= stream->codec_param.info.frm_ptime)
        jb_max_pre = info->jb_max_pre / stream->codec_param.info.frm_ptime;
    else {
        //jb_max_pre = 240 / stream->codec_param.info.frm_ptime;
        jb_max_pre = jb_max * 4 / 5;
    }
```
Also will be evaluated as false, but it will assing jb_max_pre = jb_max * 4 / 5;
As jb_max has the value of 1, then jb_max_pre will receive 0.

Suggested fix (I wll try to make a PR if everythings ok)
```
    if (info->jb_max_pre >= stream->codec_param.info.frm_ptime)
        jb_max_pre = info->jb_max_pre / stream->codec_param.info.frm_ptime;
    else {
        //jb_max_pre = 240 / stream->codec_param.info.frm_ptime;
        jb_max_pre = jb_max * 4 / 5;
        if  (jb_max_pre<jb_min_pre)
             jb_max_pre=jb_min_pre;
    }
```

### Steps to reproduce

Set/add into pjsua.conf 
--jb-max-size 20 

and then run
pjsua --config-file=pjsua.conf 

it will crash when a call it's accepted

### PJSIP version

2.12.1 

### Context

Tested on Linux but it should also fail on Windows, OSX, etc

### Log, call stack, etc

```shell
.
```

sauwming:
[...]

PR #3277, opened 2022-10-27
Avoid crash when info->jb_max equals frame ptime 

SebaLedesma:
See issue #3271


silentindark:
@SebaLedesma 
I think that should be one commit - 
"Avoid crash when info->jb_max equals frame ptime" the last one?

nanangizz:
Yes, we will do squash & merge, so only the last diff matters and it looks clean.

Few notes though:
- video stream may need to be updated too?
- perhaps need to fix/update the pjsua option `--jb-max-size` help string too (from `in frames` to `in msec`):
  https://github.com/pjsip/pjproject/blob/74569f926a2f553006d4bbac70c7371dfb1eb400/pjsip-apps/src/pjsua/pjsua_app_config.c#L170
  This may sound like a backward compatibility issue, but then the underlying code is already buggy (the setting won't work as expected), so I guess it is okay to change it. Moreover, frame time may vary a lot, e.g: for audio it can be ranging from 1ms to 30 ms (depends on the codec), while for video, the range is even broader (depends on frame rate), so I guess it is not very sensible to have JB max size setting in frame count.

SebaLedesma:
Sorry about the late, we are heading an exhibition so everyone is running here :)
1 - The logic is generic for audio/video stream, so the solution is valid for both. Say that, I have more experience with audio, and we are doing some testing/optimizations for opus encoding, when they are more polished I will create their respective PR.
2 - You are right, the command line help also needs to be updated.  I'll put on my agenda.

sauwming:
I can help fix this if you'd like. But since you're using master branch in your repo, I would like to ask your permission first if it's okay if I commit directly (I received a warning message just now when trying to resolve the code conflict).

sauwming:
I created a new PR #3551 to resume this.

Issue #3191, opened 2022-07-22
Feature Request: ability to retrieve ioqueue fd/handle


mscdex:
**Is your feature request related to a problem? Please describe.**
Currently there is no easy and efficient way to integrate pjsip's ioqueue into an existing polling mechanism (e.g. existing epoll instance that is already handling other file descriptors).

**Describe the solution you'd like**
An API that returns a `pj_oshandle_t` containing the ioqueue's fd/handle. In the case of epoll, it would return the `epfd`. I'm not picky as to where the API lives, whether it's `pjsip_endpt_get_ioqueue_handle(pjsip_endpoint *endpt)`, `pj_ioqueue_get_handle(pj_ioqueue_t *ioque)`, and/or something else.

**Describe alternatives you've considered**
Currently I have to copy a large chunk of struct definitions from the (epoll) ioqueue implementation (since those structures are not public) just so I can get at the `epfd` field. This is very unwieldy and not easily maintainable going forward.

nanangizz:
As you may already know, the PJLIB ioqueue only does the polling and the event dispatching for its PJLIB-registered file descriptors (e.g: via `pj_ioqueue_register_sock()`). And based on your description, you seem to want to share the PJLIB ioqueue with non-registered file descriptors. If that's correct, I am afraid I don't understand how it may work, for example, events for the non-registered descriptors will not be dispatched. Or if your app also does its own additional polling (to serve the non-registered descriptors), I am not sure if it won't cause issues or complexities, why not creating its own ioqueue instead. So please describe more detail how retrieving native ioqueue handle can be useful for app.

mscdex:
I have an existing epoll handle that monitors pjsip and other app-specific file descriptors on a single thread. I don't want to add file descriptors to pjsip's epoll handle, I just want to utilize epoll handle nesting to efficiently integrate pjsip into my existing epoll handle. Also using a pjsip ioqueue for the app-specific file descriptors isn't an option because of the way I'm using epoll vs. the way pjsip uses epoll.

Like I said, it currently works, but it's a hack since I have to copy all of the related internal struct definitions. That is obviously not very maintainable.

nanangizz:
Ah okay, so it is for nested epoll, thank you for the clarification.

PR #3279, opened 2022-11-01
Add ability to retrieve ioqueue OS handle

mscdex:
Fixes: https://github.com/pjsip/pjproject/issues/3191

### DIFF
--- a/pjlib/src/pjlib-test/ioq_udp.c
+++ b/pjlib/src/pjlib-test/ioq_udp.c
@@ -142,6 +142,8 @@ static int compliance_test(const pj_ioqueue_cfg *cfg)
     pj_str_t temp;
     pj_bool_t send_pending, recv_pending;
     pj_status_t rc;
+    pj_str_t ioque_name;
+    pj_str_t ioqueue_type;
 
     pj_set_os_error(PJ_SUCCESS);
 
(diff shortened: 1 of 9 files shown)


nanangizz (approved):

sauwming (approved):

pjlib/src/pjlib-test/ioq_udp.c
```diff
@@ -180,6 +182,19 @@ static int compliance_test(const pj_ioqueue_cfg *cfg)
...
+    ioque_name = pj_str((char*)pj_ioqueue_name());
+    if (pj_strncmp(&ioque_name, pj_cstr(&ioqueue_type, "epoll"), 5) == 0 ||
+        pj_strncmp(&ioque_name, pj_cstr(&ioqueue_type, "kqueue"), 6) == 0 ||
+        pj_strncmp(&ioque_name, pj_cstr(&ioqueue_type, "iocp"), 4) == 0) {
+      if (pj_ioqueue_get_os_handle(ioque) == NULL) {
+        PJ_LOG(1,(
+          THIS_FILE,
+          "...pj_ioqueue_os_handle() unexpectedly returned NULL"
+        ));
+        status=-21; goto on_error;
```
wosrediinanatour, line 194: Wouldn't it be better to have a constant than a magic number?

nanangizz: The unit tests use magic numbers for errors :)
As long as they are unique, so hopefully we can identify which test that fails, it should be fine.

PR #3290, opened 2022-11-17
Use environment LDFLAGS when building with resample as shared lib

trengginas:
When building the library with `--enable-resample-dll` option, environment LDFLAGS required to be used.

### DIFF
--- a/third_party/build/resample/Makefile
+++ b/third_party/build/resample/Makefile
@@ -20,6 +20,7 @@ export _LDFLAGS := $(CC_LDFLAGS) $(OS_LDFLAGS) $(M_LDFLAGS) $(HOST_LDFLAGS) \
 export RESAMPLE_SRCDIR = ../../resample/src
 export RESAMPLE_OBJS = resamplesubs.o
 export RESAMPLE_CFLAGS = $(_CFLAGS)
+export RESAMPLE_LDFLAGS = $(_LDFLAGS)
 
 export RESAMPLE_LIB := libresample-$(TARGET_NAME)$(LIBEXT)
 
@@ -44,7 +45,6 @@ ifneq ($(PJ_SHARED_LIBRARIES),)
 else
 ifneq ($(PJ_RESAMPLE_DLL),)
 export RESAMPLE_CFLAGS := -fPIC $(RESAMPLE_CFLAGS)
-export RESAMPLE_LDFLAGS := $(RESAMPLE_LDFLAGS)
 endif
 endif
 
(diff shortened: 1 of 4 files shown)


third_party/build/resample/Makefile
```diff
@@ -44,6 +44,7 @@ ifneq ($(PJ_SHARED_LIBRARIES),)
 else
 ifneq ($(PJ_RESAMPLE_DLL),)
 export RESAMPLE_CFLAGS := -fPIC $(RESAMPLE_CFLAGS)
+export RESAMPLE_LDFLAGS := $(RESAMPLE_LDFLAGS)
```
nanangizz, line 47: Instead of modifying the configure script, perhaps adding
```
export RESAMPLE_LDFLAGS := $(_LDFLAGS)
```
after https://github.com/pjsip/pjproject/blob/ae8f9796a45e95a945195d807c75868f0dd1607c/third_party/build/resample/Makefile#L22 is sufficient?

For any approach, we need to make sure it does not cause build failure to other configs, such as `--enable-shared` and `--enable-static`.

trengginas: If we use
```
export RESAMPLE_LDFLAGS := $(_LDFLAGS)
```
Then, all the library that is not needed on building libresample will be included. e.g: `-luuid -lm -lrt -lpthread  -lasound`

trengginas: From 
https://stackoverflow.com/questions/13984180/is-an-executable-built-differently-if-linked-against-a-library-thats-not-used

states that linking against a a static library not used should not have any effect. 
Using 
```
export RESAMPLE_LDFLAGS := $(_LDFLAGS)
```
should be sufficient.

trengginas: Tested here using `--enable-static`, `--enable-shared`, `--enable-resample-dll`, with 
```
export RESAMPLE_LDFLAGS := $(_LDFLAGS)
```
All succees

nanangizz: > From https://stackoverflow.com/questions/13984180/is-an-executable-built-differently-if-linked-against-a-library-thats-not-used
> 
> states that linking against a a static library not used should not have any effect. Using

How about linked against dynamic library, will the resample requires the unneeded libraries as its dependencies (such as `-luuid -lm -lrt -lpthread -lasound`)?

Btw, just a note, after a quick check on other third-party Makefile's, some of them include the `*_LDFLAGS` line, but few others not.

trengginas: I check the file size of `libresample.so` and there's no file size difference when using the patch/not. 
Check with `readelf` tools and also there's no difference.

nanangizz (approved):

sauwming (approved):

PR #3297, opened 2022-11-29
Fuzzing Support for oss-fuzz

0x34d:
Review.


0x34d:
Code is Working as expected https://github.com/google/oss-fuzz/pull/8934/checks

Makefile
```diff
@@ -124,6 +124,22 @@ pjsua-test: cmp_wav
 cmp_wav:
 	cd tests/pjsua/tools && make
 
+fuzz:
```
sauwming, line 127: You're not supposed to modify build.mak.in for non-global settings. Instead, you can put it in `tests/fuzz/Makefile`. Then, here,  you can do similar like above, i.e. call `cd tests/fuzz && make`.

sauwming (commented):
We would also appreciate if you can sign the CLA. Thanks a lot!

0x34d:
> We would also appreciate if you can sign the CLA. Thanks a lot!

Done.

0x34d:
Done.
You can review it now. and possibly merge.

FUZZING.MD
```diff
@@ -0,0 +1,33 @@
...
+```
+
+
+### Run
+
+```
+cd tests/fuzz
+mkdir cov-json
+unzip seed/fuzz-json_seed_corpus.zip
+./fuzz-json cov-json/ fuzz-xml_seed_corpus/
```
sauwming, line 31: should be: `fuzz-json_seed_corpus/`

tests/fuzz/Makefile
```diff
@@ -0,0 +1,37 @@
...
+all: $(TARGET)
+
+$(TARGET):
+	$(CC) $(PJ_CFLAGS) $(EXTFLAGS) -c $(JSON).c
+	$(CC) $(PJ_CFLAGS) $(EXTFLAGS) -c $(XML).c
+	$(CC) $(PJ_CFLAGS) $(EXTFLAGS) -c $(SDP).c
+	$(CC) $(PJ_CFLAGS) $(EXTFLAGS) -c $(STUN).c
+	$(CC) $(PJ_CFLAGS) $(EXTFLAGS) -c $(SIP).c
+
+	$(CXX) $(PJ_CFLAGS) -o $(JSON) $(JSON).o $(PJ_LDFLAGS) $(LIB_FUZZING_ENGINE) $(FUZZ_pjlib)
```
sauwming, line 30: I tested it here and it almost works. It should be:
`$(CXX) $(PJ_CFLAGS) -o $(JSON) $(JSON).o $(PJ_LDFLAGS) $(PJ_LDLIBS) $(LDFLAGS) $(LIB_FUZZING_ENGINE)`

Yes, that means that it needs to list all the libraries, instead of select ones, but the compiler should figure it out and link only with the required libraries (in my test here, the resulting binaries have exactly the same size).

The reason we need to use `$(PJ_LDLIBS`) is because it may also contain other libraries that's needed by pjsip libraries (such as certain frameworks).

FUZZING.MD
0x34d: Typo.

tests/fuzz/Makefile
0x34d: Like This:
```
	$(CXX) $(PJ_CFLAGS) -o $(JSON) $(JSON).o $(PJ_LDFLAGS) $(PJ_LDLIBS) $(LDFLAGS) $(LIB_FUZZING_ENGINE)
	$(CXX) $(PJ_CFLAGS) -o $(XML)  $(XML).o  $(PJ_LDFLAGS) $(PJ_LDLIBS) $(LDFLAGS) $(LIB_FUZZING_ENGINE)
	$(CXX) $(PJ_CFLAGS) -o $(SDP)  $(SDP).o  $(PJ_LDFLAGS) $(PJ_LDLIBS) $(LDFLAGS) $(LIB_FUZZING_ENGINE)
	$(CXX) $(PJ_CFLAGS) -o $(STUN) $(STUN).o $(PJ_LDFLAGS) $(PJ_LDLIBS) $(LDFLAGS) $(LIB_FUZZING_ENGINE)
	$(CXX) $(PJ_CFLAGS) -o $(SIP)  $(SIP).o  $(PJ_LDFLAGS) $(PJ_LDLIBS) $(LDFLAGS) $(LIB_FUZZING_ENGINE)
```

sauwming (approved):
LGTM. It builds and run successfully here:
```
...
INFO: seed corpus: files: 372 min: 10b max: 3950b total: 172161b rss: 90Mb
#374	INITED cov: 359 ft: 1567 corp: 312/149Kb exec/s: 0 rss: 99Mb
#635	REDUCE cov: 359 ft: 1567 corp: 312/149Kb lim: 3950 exec/s: 0 rss: 101Mb L: 109/3950 MS: 1 EraseBytes-
```

nanangizz:
Tried to list some other "parsers" that may also be added (later perhaps?):
 - RTCP/[pjmedia_rtcp_rx_rtcp()](https://github.com/pjsip/pjproject/blob/52f7c28622c793c7445a98f333f2a8e306a683b1/pjmedia/include/pjmedia/rtcp.h#L405) and RTCP-XR/[pjmedia_rtcp_xr_rx_rtcp_xr()](https://github.com/pjsip/pjproject/blob/52f7c28622c793c7445a98f333f2a8e306a683b1/pjmedia/include/pjmedia/rtcp_xr.h#L422)
 - DNS/[pj_dns_parse_packet()](https://github.com/pjsip/pjproject/blob/52f7c28622c793c7445a98f333f2a8e306a683b1/pjlib-util/include/pjlib-util/dns.h#L345)
 - RTP unpacketizers: pjmedia_codec_parse() for audio codecs, pjmedia_h263_unpacketize(), pjmedia_h264_unpacketize(), pjmedia_vpx_unpacketize()
 - [CLI](https://github.com/pjsip/pjproject/blob/master/pjlib-util/include/pjlib-util/cli.h) parser

In case we want to add them later, we should add the fuzz-targets to pjsip repo & update the `build.sh` in `google/oss-fuzz` (via PR), right?

nanangizz (approved):

sauwming:
> In case we want to add them later, we should add the fuzz-targets to pjsip repo & update the `build.sh` in `google/oss-fuzz` (via PR), right?

This is actually a good point. @0x34d , can you please help change the `build.sh` in `google/oss-fuzz` so that it will find the fuzzer programs, instead of hardcoding them?
I'm talking something like this: https://github.com/google/oss-fuzz/blob/master/projects/boringssl/build.sh#L35

This way, we just need to add more fuzzers later in our local repo and won't need to modify anything there.

0x34d:
Done: https://github.com/google/oss-fuzz/pull/8934/commits/1074f0988598f0abf8bcdc55b167539204d04d84

trengginas (approved):

PR #3301, opened 2022-12-02
Fixed encoding video issue with odd resolution

sauwming:
It is reported that running pjsip that sends video with a non-even resolution will cause a crash when hanging up the call.

```
    frame #0: `op_disconnect_ports(vid_conf=0x0000000104d1eda8, prm=0x00000001707d22c0) at vid_conf.c:913:16
    frame #1: `op_remove_port(vid_conf=0x0000000104d1eda8, prm=0x0000000104d1ebec) at vid_conf.c:619:9
    frame #2: `handle_op_queue(conf=0x0000000104d1eda8) at vid_conf.c:206:17
    frame #3: `on_clock_tick(now=0x0000000104d1eb58, user_data=0x0000000104d1eda8) at vid_conf.c:1009:9
```

Running the app under address sanitizer will reveal that there's a heap buffer overflow during the conversion:
```
SUMMARY: AddressSanitizer: heap-buffer-overflow scale_common.cc:497 in ScaleFilterCols_C
    #0 0x1006bed20 in ScaleFilterCols_C scale_common.cc:497
    #1 0x1006b1838 in ScalePlane scale.cc:1667
    #2 0x1002ea884 in pjmedia_converter_convert converter.c:185
    #3 0x1003d91a0 in convert_frame vid_port.c:1108
    #4 0x1003d5394 in vid_pasv_port_get_frame vid_port.c:1433
    #5 0x100325138 in pjmedia_port_get_frame port.c:99
    #6 0x1003f1394 in on_clock_tick vid_conf.c:1087
```
The memory buffer is allocated here:
```
0x00010ddbdee2 is located 46 bytes to the right of 120500-byte region [0x00010dda0800,0x00010ddbdeb4)
allocated by thread T9 here:
    #6 0x1003f03cc in pj_pool_zalloc pool.h:490
    #7 0x1003f3f10 in pjmedia_vid_conf_add_port vid_conf.c:492
```

~~The solution is to round up the resolution to the nearest even number during calculation~~ 
We need to reject these odd resolutions since throughout video processing, many components/operations (such as codec, buffer size allocation, or video conversion) assume these numbers to be even and divide the resolution by 2, hence causing the rounding down.

Note that it only affects planar video type, and decoding is unaffected since the decoding buffer is allocated based on the max possible resolution.

### DIFF
--- a/pjmedia/src/pjmedia/vid_codec.c
+++ b/pjmedia/src/pjmedia/vid_codec.c
@@ -680,7 +680,7 @@ PJ_DEF(pj_status_t) pjmedia_vid_codec_mgr_set_default_param(
         (param->enc_fmt.det.vid.size.h % 2 == 1)))
     {
         PJ_LOG(3, (THIS_FILE, "Video resolution must be even"));
-        return PJ_EINVALIDOP;
+        return PJ_EINVAL;
     }
 
     pj_mutex_lock(mgr->mutex);
--- a/pjmedia/src/pjmedia/vid_codec.c
+++ b/pjmedia/src/pjmedia/vid_codec.c
@@ -677,7 +677,7 @@ PJ_DEF(pj_status_t) pjmedia_vid_codec_mgr_set_default_param(
         return PJ_EINVAL;
 
     if (param && ((param->enc_fmt.det.vid.size.w % 2 == 1) ||
-        (param->enc_fmt.det.vid.size.w % 2 == 1)))
+        (param->enc_fmt.det.vid.size.h % 2 == 1)))
     {
         PJ_LOG(3, (THIS_FILE, "Video resolution must be even"));
         return PJ_EINVALIDOP;
--- a/pjmedia/src/pjmedia/vid_codec.c
+++ b/pjmedia/src/pjmedia/vid_codec.c
@@ -186,6 +186,7 @@ PJ_DEF(pj_status_t) pjmedia_vid_codec_mgr_register_factory(
                                     pjmedia_vid_codec_factory *factory)
 {
     pjmedia_vid_codec_info info[PJMEDIA_CODEC_MGR_MAX_CODECS];
+    pjmedia_vid_codec_param param;
     unsigned i, count;
     pj_status_t status;
 
@@ -199,6 +200,18 @@ PJ_DEF(pj_status_t) pjmedia_vid_codec_mgr_register_factory(
     status = factory->op->enum_info(factory, &count, info);
     if (status != PJ_SUCCESS)
         return status;
+
+    if ((factory->op->test_alloc)(factory, info) == PJ_SUCCESS) {
+        status = (factory->op->default_attr)(factory, info, &param);
+        if (status == PJ_SUCCESS) {
+            if ((param.enc_fmt.det.vid.size.w % 2 == 1) ||
+                (param.enc_fmt.det.vid.size.h % 2 == 1))
+            {
+                PJ_LOG(3, (THIS_FILE, "Video resolution must be even"));
+                return PJ_EINVAL;
+            }
+        }
+    }
 
     pj_mutex_lock(mgr->mutex);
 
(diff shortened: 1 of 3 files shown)


nanangizz:
Some components do not seem to be prepared for odd resolution (i.e: no checking or rounding up mechanism), for example:
https://github.com/pjsip/pjproject/blob/f5580e3faf5357dc925da32a32a33e9278553649/pjmedia/src/pjmedia-codec/openh264.cpp#L570-L571
https://github.com/pjsip/pjproject/blob/f5580e3faf5357dc925da32a32a33e9278553649/pjmedia/src/pjmedia-codec/openh264.cpp#L679-L681

Without checking & fixing them all, allowing odd resolution may introduce issues (bad video, assert, or crash), so perhaps the safest approach will be rejecting odd resolution set by app (in PJMEDIA and PJSUA APIs), and/or specify it explicitly in the wiki/docs?

pjmedia/src/pjmedia/vid_codec.c
```diff
@@ -676,6 +676,13 @@ PJ_DEF(pj_status_t) pjmedia_vid_codec_mgr_set_default_param(
     if (!pjmedia_vid_codec_info_to_id(info, (char*)&codec_id, sizeof(codec_id)))
         return PJ_EINVAL;
 
+    if (param && ((param->enc_fmt.det.vid.size.w % 2 == 1) ||
+        (param->enc_fmt.det.vid.size.h % 2 == 1)))
+    {
+        PJ_LOG(3, (THIS_FILE, "Video resolution must be even"));
+        return PJ_EINVALIDOP;
```
nanangizz, line 683: perhaps PJ_EINVAL is more appropriate here?

nanangizz:
How about adding a similar check in `pjmedia_vid_codec_mgr_register_factory()`?

sauwming:
> How about adding a similar check in `pjmedia_vid_codec_mgr_register_factory()`?

There's no param when registering to the factory?

I added the check in `get_default_param()` instead, in case that developer changes the `DEFAULT_WIDTH/HEIGHT` in the video codec source. But instead of returning error, I think it will be better if we automatically round it up.

If we want more checks, it can also be in `codec_open()` and `modify()`, but it probably means that we shouldn't make it inline as it is now.

nanangizz:
> > How about adding a similar check in `pjmedia_vid_codec_mgr_register_factory()`?
> 
> There's no param when registering to the factory?

Yes, but it can query the default param, e.g: via `factory->op->default_attr()`?

`pjmedia_vid_codec_mgr_register_factory()` is a public API, so developer can implement its own codec and register it to PJMEDIA, so it needs to be checked.


> I added the check in `get_default_param()` instead, in case that developer changes the `DEFAULT_WIDTH/HEIGHT` in the video codec source. 

Good idea.

> But instead of returning error, I think it will be better if we automatically round it up.

Automatically rounding it up sounds okay too (with docs).


> 
> If we want more checks, it can also be in `codec_open()` and `modify()`, but it probably means that we shouldn't make it inline as it is now.

Keeping them as inline funcs looks okay to me as the checks are not that complex?

pjmedia/src/pjmedia/vid_codec.c
```diff
@@ -200,6 +201,18 @@ PJ_DEF(pj_status_t) pjmedia_vid_codec_mgr_register_factory(
...
+        if (status == PJ_SUCCESS) {
+            if ((param.enc_fmt.det.vid.size.w % 2 == 1) ||
+                (param.enc_fmt.det.vid.size.h % 2 == 1))
+            {
+                PJ_LOG(3, (THIS_FILE, "Video resolution must be even"));
+                return PJ_EINVAL;
+            }
+        }
+    }
+
```
nanangizz, line 215: `info` is an array of codecs, so I think this block should be moved to inside the codec loop below?

nanangizz: IMO either approach is fine as long as the video processing components (including third-party converters/codecs) get even resolution.

nanangizz (approved):

trengginas (approved):

Issue #2706, opened 2021-05-05
PJSIP uses IP address instead of hostname when creating TLS transport to the caller to send BYE


lauriva:
**Describe the bug**
PJSIP uses IP address instead of hostname when creating TLS transport to the caller to send BYE.

**To Reproduce**

1. Start pjsua  .`/pjsua-x86_64-unknown-linux-gnu --use-tls  --tls-ca-file=... --tls-cert-file=... --tls-privkey-file=... --tls-verify-server --contact="sip:sip.example.com:5063;transport=tls" --local-port=5062 --use-srtp=1`
1. Call to pjsua
1. Answer the call in pjsua
1. Hangup the call in pjsua
1. PJSIP creates transport using IP address and thus certificate validation fails
`10:25:16.696     tlsc0x55d222827818  ....TLS transport 10.10.10.231:32837 is connecting to 10.10.10.131:5081...`

**Expected behavior**
Transport is created and certificate is validated using hostname
`11:20:19.813     tlsc0x55c17d93aee8  ....TLS transport 10.10.10.231:37387 is connecting to origin.example.com:5081...`

**Desktop/Smartphone (please complete the following information):**
- OS: Ubuntu 20.04
- PJSIP, version 2.11

**Additional context**
Works with PJSIP 2.10

**Logs/Screenshots**
Call trace from pjsua
```
10:24:04.011           pjsua_core.c  .RX 2395 bytes Request msg INVITE/cseq=35545930 (rdata0x7f8b90045f28) from TLS 10.10.10.131:49769:
INVITE sip:7777@sip.example.com:5063;transport=tls SIP/2.0
Via: SIP/2.0/TLS 10.10.10.131:5081;rport;branch=z9hG4bKQ0F05jDcvQmgp
Max-Forwards: 69
From: "FreeSWITCH" <sip:fs@origin.example.com>;tag=B54eyBvvr1ZmK
To: <sip:7777@sip.example.com:5063>
Call-ID: b6209832-2815-123a-aab5-26faa99b4e3d
CSeq: 35545930 INVITE
Contact: <sip:gw+sip.example.com@origin.example.com:5081;transport=tls;gw=sip.example.com>
User-Agent: FreeSWITCH-mod_sofia/1.10.5-release+git~20200818T185121Z~25569c1631~64bit
Allow: INVITE, ACK, BYE, CANCEL, OPTIONS, MESSAGE, INFO, UPDATE, REGISTER, REFER, NOTIFY
[...]

PR #3310, opened 2022-12-13
TLS uses remote IP address instead of hostname when sending BYE

nanangizz:
This is an attempt to fix issue #2706.

The issue seems to occur only when the transport used in dialog creation is a server socket, where remote address is an IP address not a hostname, so the `dlg->initial_dest` will be an IP address. Note that the `dlg->initial_dest` is currently used only for initializing `tdata->dest_info.name` which eventually used for verifying remote TLS certificate.

The idea here is to only set the `dlg->initial_dest` from transport info when the transport is a client socket, which TLS verification is usually performed. When it is a server socket, the `tdata->dest_info.name` will be initialized from request URI or route header (later by `pjsip_endpt_send_request_stateless()` or `pjsip_endpt_send_response()`.

### DIFF
--- a/pjsip/src/pjsip/sip_dialog.c
+++ b/pjsip/src/pjsip/sip_dialog.c
@@ -469,8 +469,8 @@ pj_status_t create_uas_dialog( pjsip_user_agent *ua,
     
     /* Save initial destination host from transport's info */
     if (rdata->tp_info.transport->dir == PJSIP_TP_DIR_OUTGOING) {
-	pj_strdup(dlg->pool, &dlg->initial_dest,
-		  &rdata->tp_info.transport->remote_name.host);
+        pj_strdup(dlg->pool, &dlg->initial_dest,
+                  &rdata->tp_info.transport->remote_name.host);
     }
 
     /* Init remote's contact from Contact header.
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsip/sip_dialog.c
```diff
@@ -468,9 +468,10 @@ pj_status_t create_uas_dialog( pjsip_user_agent *ua,
     pj_strdup(dlg->pool, &dlg->remote.info_str, &tmp);
     
     /* Save initial destination host from transport's info */
-    pj_strdup(dlg->pool, &dlg->initial_dest,
-              &rdata->tp_info.transport->remote_name.host);
-
+    if (rdata->tp_info.transport->dir == PJSIP_TP_DIR_OUTGOING) {
+	pj_strdup(dlg->pool, &dlg->initial_dest,
```
sauwming, line 472: minor: your editor seems to still use tab?

nanangizz: Nice catch! Just updated my editor settings & fixed the patch.

pjsip/src/pjsip/sip_dialog.c
```diff
@@ -1831,8 +1832,9 @@ static void dlg_update_routeset(pjsip_dialog *dlg, const pjsip_rx_data *rdata)
      * transaction as the initial transaction that establishes dialog.
      */
     if (dlg->role == PJSIP_ROLE_UAC) {
-        /* Save initial destination host from transport's info. */
-        if (!dlg->initial_dest.slen) {
+        /* Update initial destination host from transport's info. */
+        if (rdata->tp_info.transport->dir == PJSIP_TP_DIR_OUTGOING)
+	{
```
trengginas, line 1837: I think this is still Tab

sauwming (approved):

trengginas (approved):

PR #3319, opened 2022-12-21
Fixed race condition between DTLS stoppage and incoming packet processing

### LINKED ISSUE #3299: Crash in ssl_get_srtp_material() when invoking dtls_media_stop() in transport_srtp_dtls.c
### Describe the bug

If **_dtls_media_stop()_** is invoked simultaneously with **_ssl_flush_wbio()_** both methods would have a sync point for a little while while waiting on **_pjmedia_clock_stop()_**.

snippet from **_ssl_flush_wbio()_**:
```
    ds->nego_completed = PJ_TRUE;
    PJ_LOG(2,(ds->base.name, "DTLS-SRTP negotiation completed!"));

    /* Stop the retransmission clock. Note that the clock may not be stopped
     * if this function is called from clock thread context. We'll try again
     * later in socket context.
     */
    if (ds->clock)
    pjmedia_clock_stop(ds->clock);

    /* Get SRTP key material */
    status = ssl_get_srtp_material(ds);
```

snippet from **_dtls_media_stop()_**:
```
#if DTLS_DEBUG
    PJ_LOG(2,(ds->base.name, "dtls_media_stop()"));
#endif

    if (ds->clock)
	pjmedia_clock_stop(ds->clock);
    
    /* Reset DTLS state */
    ssl_destroy(ds);
```

If the scenario described above happens, after **_pjmedia_clock_stop()_** returns, there may be a very likely scenario of _**ssl_destroy()**_ being invoked (from within **_dtls_media_stop()_**) and right after it **_ssl_get_srtp_material()_** (from within **_ssl_flush_wbio()_**). This causes a crash when attempting to access the null pointer of _ds->ossl_ssl_.

### Steps to reproduce

Here are some possible steps that can lead to the crash:

1. **_dtls_media_stop()_** is invoked and waits on **_pjmedia_clock_stop()_**
2. ** _ssl_flush_wbio()_** is invoked and waits on **_pjmedia_clock_stop()_**
3. after  **_pjmedia_clock_stop()_** returns on both methods, **_ssl_destroy(ds)_** executes and right after it **_ssl_get_srtp_material(ds)_**

### PJSIP version

2.12.1

### Context

This was tested on Android and iOS devices.

### Log, call stack, etc

```shell
Given the description and the steps to reproduce described above I think logs are not relevant here.
```

sauwming:
To fix #3299.

`dtls_media_stop()` will destroy the SSL. If at the same time, there's an incoming data, it will cause a crash since it will access the SSL context.

The proposed solution is to stop the member transport first, in order to prevent the race. Note that the solution relies on the member transport guaranteeing that after it has been stopped:
- any incoming packet processing has completed, and
- no further incoming traffic should arrive


nanangizz (commented):
Can't really tell if this will be okay, e.g: what about the opposite direction (i.e: the keying/DTLS want to send something). Perhaps it is safer to add check (e.g: if member/keying is still valid), if possible, rather than apply call reordering?

sauwming:
> Can't really tell if this will be okay, e.g: what about the opposite direction (i.e: the keying/DTLS want to send something).
> Perhaps it is safer to add check (e.g: if member/keying is still valid), if possible, rather than apply call reordering?

Ah okay, that's an interesting scenario. Does this mean that we need locking? There's a proposal fix by the reporter in #3319 that uses locking before accessing the SSL, perhaps you can check that and see if it can better resolve the issue.

Alternatively, I think some sync between ssl_handshake() and dtls_media_stop() maybe be sufficient since if I check the code, there seems to be only 3 locations that can initiate outgoing transmission:
- from the clock, which will be stopped in `dtls_media_stop()`
- upon processing incoming data, which should have been stopped when stopping member transport.
- from `ssl_handshake()`.

So I think only the last can potentially cause issue. Let me know if I miss anything.

nanangizz:
Tracing DTLS code is not easy :) So I guess adding & using lock is a simpler & safer bet (as long as there is no stacking locks which may cause deadlock)? The DTLS should be used only in the beginning for negotiating the SRTP key, so it should not affect performance much (?).

sauwming:
The issue still happens and unfortunately the locking employed in #3321 doesn't help, because the race now happens between transport destruction, which will be destroying the lock, and incoming data, which will try to acquire the lock.

Note that `dtls_destroy()` will also destroy the pool shortly afterward: `pj_pool_safe_release(&ds->pool)`, so I believe the only way to prevent it is to make sure there's no incoming data by changing the ordering.

```
    frame #4: 0x00000001004497b0 pjsua-arm-apple-darwin22.2.0`pj_lock_destroy(lock=0x00000001059100a8) at transport_srtp_dtls.c:1838:9
    frame #5: 0x00000001004419f0 pjsua-arm-apple-darwin22.2.0`dtls_destroy(tp=0x00000001059100a8) at transport_srtp_dtls.c:1833:5
    frame #6: 0x0000000100436778 pjsua-arm-apple-darwin22.2.0`pjmedia_transport_close(tp=0x00000001059100a8) at transport.h:1033:16
    frame #7: 0x000000010044f34c pjsua-arm-apple-darwin22.2.0`transport_destroy(tp=0x0000000107c1c928) at transport_srtp.c:1468:9
    frame #8: 0x0000000100106144 pjsua-arm-apple-darwin22.2.0`pjmedia_transport_close(tp=0x0000000107c1c928) at transport.h:1033:16
    frame #9: 0x000000010011950c pjsua-arm-apple-darwin22.2.0`pjsua_media_channel_update(call_id=0, local_sdp=0x000000010aa852a8, remote_sdp=0x000000010584b9c8) at pjsua_media.c:4360:13
    frame #10: 0x0000000100095ba8 pjsua-arm-apple-darwin22.2.0`pjsua_call_on_media_update(inv=0x000000010584b6f0, status=0) at pjsua_call.c:5277:14
    frame #11: 0x0000000100198710 pjsua-arm-apple-darwin22.2.0`inv_negotiate_sdp(inv=0x000000010584b6f0) at sip_inv.c:2242:9
    frame #12: 0x000000010018a770 pjsua-arm-apple-darwin22.2.0`process_answer(inv=0x000000010584b6f0, st_code=200, tdata=0x00000001058dfda8, local_sdp=0x0000000000000000) at sip_inv.c:2581:22
    frame #13: 0x000000010018b3c4 pjsua-arm-apple-darwin22.2.0`pjsip_inv_answer(inv=0x000000010584b6f0, st_code=200, st_text=0x0000000000000000, local_sdp=0x0000000000000000, p_tdata=0x00000001702550a0) at sip_inv.c:2749:14
    frame #14: 0x00000001000b8e64 pjsua-arm-apple-darwin22.2.0`pjsua_call_answer2
```

```
    frame #1: 0x000000010033366c pjsua-arm-apple-darwin22.2.0`pj_lock_acquire(lock=0x00000001059100a8) at transport_srtp_dtls.c:808:13
    frame #2: 0x0000000100444c3c pjsua-arm-apple-darwin22.2.0`ssl_flush_wbio(ds=0x00000001059100a8, idx=0) at transport_srtp_dtls.c:803:9
    frame #3: 0x00000001004436d0 pjsua-arm-apple-darwin22.2.0`ssl_on_recv_packet(ds=0x00000001059100a8, idx=0, data=0x000000010790ff28, len=1061) at transport_srtp_dtls.c:1178:12
    frame #4: 0x000000010043e30c pjsua-arm-apple-darwin22.2.0`dtls_on_recv_rtp(tp=0x00000001059100a8, pkt=0x000000010790ff28, size=1061) at transport_srtp_dtls.c:1303:12
    frame #5: 0x0000000100447350 pjsua-arm-apple-darwin22.2.0`pjmedia_transport_send_rtp(tp=0x00000001059100a8, pkt=0x000000010790ff28, size=1061) at transport.h:869:12
    frame #6: 0x0000000100450d7c pjsua-arm-apple-darwin22.2.0`srtp_rtp_cb(param=0x00000001700b5720) at transport_srtp.c:1529:22
```

nanangizz:
IIRC for most places, the ordering between keying & transport itself is very crucial, so this approach may be risky, should do more manual checks & tests.

Btw, can we just switch the existing simple lock to group lock?

sauwming:
Ah yes, the group lock may help here. But it will be more complex.

For now, I propose to close the member transport first without changing the ordering (latest patch). This should be safe?

Issue #1688, opened 2013-07-10
Add support for different To and Target URI in outgoing call and sending IM (thanks Johan Lantz for the suggestion)


pjsipbot:
_2013-07-10 05:25:55: @mingteluu_ _created the issue on trac ticket [1688](https://trac.pjsip.org/repos/ticket/1688)_

___
_2013-07-10 05:27:10: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2013-07-10 05:27:10: @mingteluu_ _set **resolution** to **fixed**_

___
_2013-07-10 05:27:10: @mingteluu_ _commented_

In [r4555](../commit/77dadeffa36f2ddde79bf09851cabc1412458eec):
Closed #1688: Add support for different To and Target URI in outgoing call and sending IM

___
_2013-07-15 02:29:04: @mingteluu_ _commented_

In [r4561](../commit/6f4a16e32bfe49857e5d2f9496a9024769c54c31):
Re #1688: Add fix to check whether msg_data is null

___
_2013-09-24 06:01:45: @bennylp_ _changed **title** from **Add support for different To and Target URI in outgoing call and sending IM** to **Add support for different To and Target URI in outgoing call and sending IM (thanks Johan Lantz for the suggestion)**_

PR #3320, opened 2022-12-21
Support different From URI in outgoing calls

wosrediinanatour:
In outgoing calls, PJSUA(2) uses the account's ID as URI for From headers. This commit enables to overwrite the URI in the From headers.

The solution is similar to PR #1688.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -2403,6 +2403,13 @@ struct pjsua_msg_data
     pj_str_t    target_uri;
 
     /**
+     * Optional local URI (i.e. From header). If NULL, the account ID
+     * \a pjsua_acc_config.id is used for the From header. This field is
+     * currently used only by pjsua_call_make_call() and pjsua_im_send().
+     */
+    pj_str_t    local_uri;
+
+    /**
      * Additional message headers as linked list. Application can add
      * headers to the list by creating the header, either from the heap/pool
      * or from temporary local variable, and add the header using
--- a/pjsip/include/pjsua2/siptypes.hpp
+++ b/pjsip/include/pjsua2/siptypes.hpp
@@ -838,6 +838,13 @@ struct SipTxOption
     string                  targetUri;
 
     /**
+     * Optional local URI (i.e. From header). If empty (""), the
+     * \a AccountConfig::idUri is used for the From header. At the moment this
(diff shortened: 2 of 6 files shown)


pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -2402,6 +2402,12 @@ struct pjsua_msg_data
      */
     pj_str_t    target_uri;
 
+    /**
+     * Optional local URI (i.e. From header). If NULL, the
+     * the contact is used for the From header.
```
sauwming, line 2407: Need to add doc which APIs will this apply to, i.e.: `This field is currently only used by ....`?
Is it only applicable to pjsua_call_make_call() only?

sauwming (commented):
Also need to duplicate the new field in `pjsua_msg_data_clone()`

wosrediinanatour:
Incorporated feedback by @sauwming. Added support in pjsua_im_send(). Please take a look.

pjsip/include/pjsua-lib/pjsua.h
wosrediinanatour: It was. Latest change adds `pjsua_im_send()`.

sauwming (approved):

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -2402,6 +2402,13 @@ struct pjsua_msg_data
      */
     pj_str_t    target_uri;
 
+    /**
+     * Optional local URI (i.e. From header). If NULL, the
+     * contact is used for the From header. This field is currently
```
nanangizz, line 2407: instead of `the contact`, shouldn't it be `the account ID`?

pjsip/include/pjsua2/siptypes.hpp
```diff
@@ -837,6 +837,13 @@ struct SipTxOption
      */
     string                  targetUri;
 
+    /**
+     * Optional local URI (i.e. From header). If empty (""), the
+     * the contact is used for the From header. At the moment this
```
nanangizz, line 842: similar to above.

pjsip/include/pjsua-lib/pjsua.h
wosrediinanatour: Good point: the account ID defines the contact header and the from header. But the from header is not defined by the contact header.

nanangizz (approved):

PR #3322, opened 2022-12-28
Increment origin version when there's a change in local SDP answer

trengginas:
Related to #2646.

Consider the scenario:
1. TX INITIAL INVITE. (multiple codecs offer)
2. RX 200 OK for INVITE (single codec answer) 
3. TX INVITE (Session Refresh) 
4. TX INVITE (Session Refresh)

When sending re-INVITE for SIP session refresh, the origin version will always be incremented although there's no SDP change (INVITE [4] and onwards). 

This ticket will move the SDP check to `pjmedia_sdp_neg_negotiate()` and change the origin version there.

### DIFF
--- a/pjmedia/src/pjmedia/sdp_neg.c
+++ b/pjmedia/src/pjmedia/sdp_neg.c
@@ -41,7 +41,11 @@ struct pjmedia_sdp_neg
                         *active_local_sdp,  /**< Currently active local SDP. */
                         *active_remote_sdp, /**< Currently active remote's.  */
                         *neg_local_sdp,     /**< Temporary local SDP.        */
-                        *neg_remote_sdp;    /**< Temporary remote SDP.       */
+                        *neg_remote_sdp,    /**< Temporary remote SDP.       */
+                        *prev_sent;         /**< Previously sent SDP based on the nego or
+                                                 changes on the local SDP. 
+                                                 Note that application might change the
+                                                 actual SDP sent using PJSIP module. */
 };
 
 static const char *state_str[] = 
@@ -118,6 +122,7 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_create_w_local_offer( pj_pool_t *pool,
     neg->answer_with_multiple_codecs = PJMEDIA_SDP_NEG_ANSWER_MULTIPLE_CODECS;
     neg->initial_sdp = pjmedia_sdp_session_clone(pool, local);
     neg->neg_local_sdp = pjmedia_sdp_session_clone(pool, local);
+    neg->prev_sent = neg->initial_sdp;
 
     *p_neg = neg;
     return PJ_SUCCESS;
@@ -412,7 +417,7 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_modify_local_offer2(
 #if PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION
     new_offer->origin.version = old_offer->origin.version;
 
-    if (pjmedia_sdp_session_cmp(new_offer, neg->initial_sdp, 0) != PJ_SUCCESS)
+    if (pjmedia_sdp_session_cmp(new_offer, neg->prev_sent, 0) != PJ_SUCCESS)
     {
         ++new_offer->origin.version;
     }    
@@ -423,6 +428,7 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_modify_local_offer2(
     neg->initial_sdp_tmp = neg->initial_sdp;
     neg->initial_sdp = new_offer;
     neg->neg_local_sdp = pjmedia_sdp_session_clone(pool, new_offer);
+    neg->prev_sent = new_offer;
 
     return PJ_SUCCESS;
 }
@@ -464,6 +470,10 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_send_local_offer( pj_pool_t *pool,
          * In this case set the neg_local_sdp as the offer.
          */
         *offer = neg->neg_local_sdp;
+    }
+
+    if (pjmedia_sdp_session_cmp(neg->prev_sent, *offer, 0) != PJ_SUCCESS) {
+        neg->prev_sent = neg->neg_local_sdp;
     }
 
     return PJ_SUCCESS;
@@ -1588,6 +1598,7 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_negotiate( pj_pool_t *pool,
 #endif
 
         }
+        neg->prev_sent = neg->neg_local_sdp;
     } else {
         pjmedia_sdp_session *answer = NULL;
 
@@ -1619,6 +1630,7 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_negotiate( pj_pool_t *pool,
             neg->active_local_sdp = answer;
             neg->active_remote_sdp = neg->neg_remote_sdp;
         }
+        neg->prev_sent = neg->active_local_sdp;
     }
 
     /* State is DONE regardless */
--- a/pjmedia/src/pjmedia/sdp_neg.c
+++ b/pjmedia/src/pjmedia/sdp_neg.c
@@ -42,10 +42,10 @@ struct pjmedia_sdp_neg
                         *active_remote_sdp, /**< Currently active remote's.  */
                         *neg_local_sdp,     /**< Temporary local SDP.        */
                         *neg_remote_sdp,    /**< Temporary remote SDP.       */
-                        *prev_sent;         /**< Previously sent SDP based on the nego or
-                                                 changes on the local SDP. 
-                                                 Note that application might change the
-                                                 actual SDP sent using PJSIP module. */
+                        *last_sent;         /**< Last sent local SDP.
+                                                 Note that application might
+                                                 change the actual SDP sent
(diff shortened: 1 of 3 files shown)


nanangizz (approved):

sauwming:
The patch looks okay and it seems to solve the above issue as well as the issue in #2646. And it also makes sense that if the active local SDP as a result of the negotiation is different from the SDP offer that we sent, we increment the version.

But that's only if `PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION` is set to 1.

So what if `PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION` is 0, which is currently the default? With the patch, now it will increment the version every time after we get an answer to our offer, which may break the the session timer refresh scenario above where the server rejects the request due to unchanged SDP with increased version.

I don't know what should be the solution for this. My idea is to make a special exception here and remove the macro checking, i.e. we always have to compare the SDP. Of course the exception will make this different from the other places that have the macro check in place, but I think it should be okay since the SDP comparison is necessary here, not optional.

Another impact worth noting is that with the patch, it's possible that the SDP will be incremented twice, i.e.
```
TX offer -> RX answer -> version++
reinvite -> on_create_offer() -> sdp_neg_modify_local_offer() -> version++
```

Checking the RFC, it should be okay since in states that:
```
<sess-version> is a version number for this session description.  Its
      usage is up to the creating tool, so long as <sess-version> is
      increased when a modification is made to the session data.
```
So the sess-version usage is up to our implementation with a condition that the version can only become larger, but still, this will mean that there will be a behavioural change, albeit minor (?).

Any thoughts or other ideas?

nanangizz:
> So what if `PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION` is 0, which is currently the default? With the patch, now it will increment the version every time after we get an answer to our offer, which may break the the session timer refresh scenario above where the server rejects the request due to unchanged SDP with increased version.

The default is set that way perhaps because we tried to maintain the behavior (previously the version will always be increased regardless)?

Btw, just realizing one thing that is perhaps important: the comparison is always against the initial SDP, while the initial SDP does not seem to be equal to the previously-sent SDP (from a quick check, initial SDP is updated only in `modify_local_offer()`, while it may need to be done in send_local() and sdp_nego() too). Not really sure whether initial SDP is meant to be previously-sent SDP though.

Tried to list some APIs involving local SDP and when/what previously-sent SDP (`prev-sent`) should be set to:
```
create_w_local_offer() -> init prev_sent
send_local_offer()     -> compare prev_sent vs active_local, set prev_sent=active_local, return active_local
modify_local_offer()   -> compare prev_sent vs new offer, set prev_sent=new offer, return new offer
sdp_nego()             -> if local is answerer, set prev_sent=active_local (we send SDP answer from the nego result, i.e: active_local, note that app may modify the SDP via PJSIP module, just put warnings somewhere perhaps)
                       -> optional, if local is offerer, set prev_sent=neg_local (app may modify SDP after send_local_offer() or via PJSIP module, so neg_local should be the SDP that is really sent)
set_local_answer()     -> the provided sdp is not really sent, it is used for nego, the nego result will be sent
```

sauwming:
Keeping track of the previously sent SDP sounds like a better idea.

So the solution should be:
- Use `initial_sdp` as prev-sent and explicitly state it in the doc.
- Deprecate the macro `PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION`. Otherwise this will still create the issue of always incrementing the version while bypassing the sdp comparison with prev-sent.

trengginas:
Another reason we use the macro was to avoid the SDP comparison.

nanangizz:
The section specifying about SDP version related to session modification seems to be (only?) this:

https://www.rfc-editor.org/rfc/rfc3264#section-8
```
   When issuing an offer that modifies the session,
   the "o=" line of the new SDP MUST be identical to that in the
   previous SDP, except that the version in the origin field MUST
   increment by one from the previous SDP.  If the version in the origin
   line does not increment, the SDP MUST be identical to the SDP with
   that version number.
```
As it does not explicitly prohibit incrementing the SDP version when the session is unmodified, IMO always incrementing the version does not violate the standard (but some may interpret it differently, e.g: the reported issue this PR wants to fix).

Also, the rfc section above explicitly specifies to increment the version **by one**, so incrementing it twice may cause issues in the future.

This thread briefly discussed this: https://lists.cs.columbia.edu/pipermail/sip-implementors/2014-July/029740.html

Just additional note, we don't really check the remote SDP version ourselves, PJSUA-LIB does compare the SDP content (and other settings such as local media settings) to restart the media instead of relying on remote SDP version.

sauwming:
Good find. So it's confirmed that we cannot use the current patch since it may increment the version more than once.

Unfortunately, always incrementing the version is not an option either, for the session refresh scenario as reported above. Here's according to [RFC 4028](https://www.rfc-editor.org/rfc/rfc4028#section-7.4):
```
It is RECOMMENDED that the UPDATE request not contain an
   offer [[4](https://www.rfc-editor.org/rfc/rfc4028#ref-4)], but a re-INVITE SHOULD contain one, even if the details of
   the session have not changed.  In that case, the offer MUST indicate
   that it has not changed.  In the case of SDP, this is accomplished by
   including the same value for the origin field as did previous SDP
   messages to its peer.  The same is true for an answer exchanged as a
   result of a session refresh request; if it has not changed, that MUST
   be indicated.
```

So the conclusion so far is:
- Use `initial_sdp` as prev-sent and explicitly state it in the doc.
- We must always compare the SDP with the prev-sent (i.e. not use the macro `PJMEDIA_SDP_NEG_COMPARE_BEFORE_INC_VERSION`).

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -409,20 +414,15 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_modify_local_offer2(
...
-    }    
-#else
-    new_offer->origin.version = old_offer->origin.version + 1;
-#endif
-    
+    }
     neg->initial_sdp_tmp = neg->initial_sdp;
     neg->initial_sdp = new_offer;
     neg->neg_local_sdp = pjmedia_sdp_session_clone(pool, new_offer);
+    neg->prev_sent = neg->neg_local_sdp;
```
nanangizz, line 425: Just a note, the SDP nego module will use a flip-flop pools setup. The pool used for cloning the SDP is usually a provisional pool, which means it may be released/reset after SDP nego if the nego fails. When the nego succeeds, it will be an active pool (will not be reset until a next successful SDP nego). That's why maintaining the `initial_sdp` looks a bit weird. Didn't checked the code detail, but please make sure it is safe in such flip-flop pool setup :)

sauwming: So is this handled already? Because it seems that there's no reversion for `last_sent` here:
https://github.com/pjsip/pjproject/blob/caabba58a0c27229998a510b7aa85832623493ae/pjmedia/src/pjmedia/sdp_neg.c#L1631
Shouldn't it be the same like `initial_sdp`?

And btw, what is `initial_sdp` now used for? i.e. what's the difference between `initial_sdp` and `last_sent` now?

nanangizz: Actually reversion may not be needed for `last_sent` as whatever has been sent to remote cannot be reversed/cancelled? Also the provisional pool (`pool_prov`) will not be reset after a failed nego, so it looks safe already in the flip-flop pool setup.

Re: `initial_sdp`, it seems to be used for storing initial local capability which is used by `pjmedia_sdp_neg_set_local_answer()`
https://github.com/pjsip/pjproject/blob/c43af43627baeb2caa43977f6783081906bd1908/pjmedia/include/pjmedia/sdp_neg.h#L651-L657

sauwming: But in `sip_inv.c` `inv_negotiate_sdp()`, there's something like this:
```
    status = pjmedia_sdp_neg_negotiate(inv->pool_prov, inv->neg, 0);
...
    if (inv->state != PJSIP_INV_STATE_DISCONNECTED) {

        /* Swap the flip-flop pool when SDP negotiation success. */
        if (status == PJ_SUCCESS) {
            swap_pool(&inv->pool_prov, &inv->pool_active);
        }

        /* Reset the provisional pool regardless SDP negotiation result. */
        pj_pool_reset(inv->pool_prov);
```
So if `last_sent` is allocated using pool_prov, it may have been reset?

nanangizz: Ah you're right, missed that! Thought the outer `if` is SDP nego result check.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1594,6 +1596,13 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_negotiate( pj_pool_t *pool,
             /* Only update active SDPs when negotiation is successfull */
             neg->active_local_sdp = active;
             neg->active_remote_sdp = neg->neg_remote_sdp;
+
+            /* Keep the pool used for allocating the active SDPs */
+            neg->pool_active = pool;
+        } else {
+            /* SDP nego failed, retain the last_sdp. */
+            neg->last_sent = pjmedia_sdp_session_clone(neg->pool_active,
```
sauwming, line 1604: this will crash on first negotiation failure since pool_active is NULL.

PR #3331, opened 2023-01-15
SRTP not able to restart after receiving old packets, SRTP being bypassed when it should not 

kevinkouketsu:
This PR addresses two issues:

- When the configuration PJMEDIA_SRTP_CHECK_RTP_SEQ_ON_RESTART is enabled, old RTP packets arriving can cause the SRTP transport to stop and start again. However, during this process, the stop method empties the currently used TX/RX policies, causing the transport to not start correctly and leaving it in an invalid state.
- If the policies become invalid (issue above), the bypass_srtp flag is set to PJ_TRUE. This poses a security risk if the SRTP use is set to MANDATORY.

To resolve the first issue, the PR uses the temporary variables (rx_policy_neg and tx_policy_neg) which are still valid at this point. Personally speaking, the reason for the restart feature is unknown. Is it really needed? I am asking that because the SRTP library it self can handle correctly old packets.

The second issue is resolved by setting the bypass_srtp flag to PJ_FALSE if the use is MANDATORY and to PJ_TRUE otherwise.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp.c
+++ b/pjmedia/src/pjmedia/transport_srtp.c
@@ -892,7 +892,10 @@ PJ_DEF(pj_status_t) pjmedia_transport_srtp_start(
 
     /* If all options points to 'NULL' method, just bypass SRTP */
     if (cr_tx_idx == 0 && cr_rx_idx == 0 && au_tx_idx == 0 && au_rx_idx == 0) {
-        srtp->bypass_srtp = PJ_TRUE;
+        srtp->bypass_srtp = srtp->setting.use != PJMEDIA_SRTP_MANDATORY ? PJ_FALSE : PJ_TRUE;
+
+        PJ_LOG(4, (THIS_FILE, "SRTP bypass flag has been %s",
+            srtp->bypass_srtp == PJ_TRUE ? "enabled" : "disabled"));
         goto on_return;
     }
 
--- a/pjmedia/src/pjmedia/transport_srtp.c
+++ b/pjmedia/src/pjmedia/transport_srtp.c
@@ -1530,8 +1530,8 @@ static void srtp_rtp_cb(pjmedia_tp_cb_param *param)
         pjmedia_srtp_crypto tx, rx;
         pj_status_t status;
 
-        tx = srtp->tx_policy;
-        rx = srtp->rx_policy;
+        tx = srtp->tx_policy_neg;
+        rx = srtp->rx_policy_neg;
 
         /* Stop SRTP first, otherwise srtp_start() will maintain current
          * roll-over counter.
--- a/pjmedia/src/pjmedia/transport_srtp.c
+++ b/pjmedia/src/pjmedia/transport_srtp.c
@@ -892,7 +892,7 @@ PJ_DEF(pj_status_t) pjmedia_transport_srtp_start(
 
     /* If all options points to 'NULL' method, just bypass SRTP */
     if (cr_tx_idx == 0 && cr_rx_idx == 0 && au_tx_idx == 0 && au_rx_idx == 0) {
-        srtp->bypass_srtp = srtp->setting.use != PJMEDIA_SRTP_MANDATORY ? PJ_FALSE : PJ_TRUE;
+        srtp->bypass_srtp = srtp->setting.use != PJMEDIA_SRTP_MANDATORY ? PJ_TRUE : PJ_FALSE;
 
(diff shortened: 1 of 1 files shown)


kevinkouketsu:
If someone would like to test this PR, it is really pretty simple. You can use the tcgui project (to be simpler, https://github.com/tum-lkn/tcgui), and set the parameters to somethind around of 25% of packet loss, 25% of reordering, 25% of duplicated. With those settings, the PJsip will behave like I have described.

pjmedia/src/pjmedia/transport_srtp.c
```diff
@@ -892,7 +892,10 @@ PJ_DEF(pj_status_t) pjmedia_transport_srtp_start(
 
     /* If all options points to 'NULL' method, just bypass SRTP */
     if (cr_tx_idx == 0 && cr_rx_idx == 0 && au_tx_idx == 0 && au_rx_idx == 0) {
-        srtp->bypass_srtp = PJ_TRUE;
+        srtp->bypass_srtp = srtp->setting.use != PJMEDIA_SRTP_MANDATORY ? PJ_FALSE : PJ_TRUE;
```
sauwming, line 895: shouldn't this be the other way around (if mandatory, we can't bypass?)

kevinkouketsu: Yes, you are right. I don't know why I've committed this way. 

I'm going to fix this

kevinkouketsu: Fixed

pjmedia/src/pjmedia/transport_srtp.c
```diff
@@ -892,7 +892,10 @@ PJ_DEF(pj_status_t) pjmedia_transport_srtp_start(
 
     /* If all options points to 'NULL' method, just bypass SRTP */
     if (cr_tx_idx == 0 && cr_rx_idx == 0 && au_tx_idx == 0 && au_rx_idx == 0) {
-        srtp->bypass_srtp = PJ_TRUE;
+        srtp->bypass_srtp = srtp->setting.use != PJMEDIA_SRTP_MANDATORY ? PJ_TRUE : PJ_FALSE;
```
trengginas, line 895: Can you please adjust the formatting to fit within a maximum of 80 characters per line?

sauwming:
I tried to reproduce the issue with `tc` as described but didn't encounter any issue with my SRTP call.

And I still haven't understood how the issue can appear.

> This PR addresses two issues:
> * When the configuration PJMEDIA_SRTP_CHECK_RTP_SEQ_ON_RESTART is enabled, old RTP packets arriving can cause the SRTP transport to stop and start again. However, during this process, the stop method empties the currently used TX/RX policies, causing the transport to not start correctly and leaving it in an invalid state.

It's true that stop method will bzero the policies, that's why it was recently fixed by @trengginas in https://github.com/pjsip/pjproject/commit/d2acb9af4e27b5ba75d658690406cec9c274c5cc to store the policy first before stopping and (re)starting.

So perhaps your fix is supposed to be applied to the version prior to the commit? Or is there any scenario that we missed with the above commit (perhaps a failed restart?)?

@trengginas, if you managed to reproduce this, and/or confirm the patch in this PR, let us know.

kevinkouketsu:
> I tried to reproduce the issue with `tc` as described but didn't encounter any issue with my SRTP call.
> 
> And I still haven't understood how the issue can appear.
> 
> > This PR addresses two issues:
> > 
> > * When the configuration PJMEDIA_SRTP_CHECK_RTP_SEQ_ON_RESTART is enabled, old RTP packets arriving can cause the SRTP transport to stop and start again. However, during this process, the stop method empties the currently used TX/RX policies, causing the transport to not start correctly and leaving it in an invalid state.
> 
> It's true that stop method will bzero the policies, that's why it was recently fixed by @trengginas in [d2acb9a](https://github.com/pjsip/pjproject/commit/d2acb9af4e27b5ba75d658690406cec9c274c5cc) to store the policy first before stopping and (re)starting.
> 
> So perhaps your fix is supposed to be applied to the version prior to the commit? Or is there any scenario that we missed with the above commit (perhaps a failed restart?)?
> 
> @trengginas, if you managed to reproduce this, and/or confirm the patch in this PR, let us know.

Apologizes, I am going to try to see if my own codebase isn't with the same code in SRTP.

trengginas:
I followed the reproducing steps, however cannot reproduce it here.

sauwming:
Closing this as it should have been fixed in the latest master. Please reopen if there's new update.

Issue #3306, opened 2022-12-06
SIPS uris are being forward to TLS transport regardless of 'transport=' parameter


j-m-pinto:
### Describe the bug

After looking into #1740 I wonder if the specification was correctly interpreted. From the spec quote in that issue:
```
As specified in [RFC3261 8.1.2](http://tools.ietf.org/html/rfc3261#section-8.1.2):
 
   Independent of which URI is used as input to the procedures of [4], if the
   Request-URI specifies a SIPS resource, the UAC MUST follow the
   procedures of [4] as if the input URI were a SIPS URI.

   Local policy MAY specify an alternate set of destinations to attempt.
   If the Request-URI contains a SIPS URI, any alternate destinations
   MUST be contacted with TLS.
```
However if we look into reference [4] of RFC3261, section [4.1](https://www.rfc-editor.org/rfc/rfc3263#section-4.1) Selecting a Transport Protocol, we can read the following:
```
First, the client selects a transport protocol.

   If the URI specifies a transport protocol in the transport parameter,
   that transport protocol SHOULD be used.

   Otherwise, if no transport protocol is specified, but the TARGET is a
   numeric IP address, the client SHOULD use UDP for a SIP URI, and TCP
   for a SIPS URI.  Similarly, if no transport protocol is specified,
   and the TARGET is not numeric, but an explicit port is provided, the
   client SHOULD use UDP for a SIP URI, and TCP for a SIPS URI.  This is
   because UDP is the only mandatory transport in [RFC 2543](https://www.rfc-editor.org/rfc/rfc2543) [[6](https://www.rfc-editor.org/rfc/rfc3263#ref-6)], and thus
   the only one guaranteed to be interoperable for a SIP URI.  It was
   also specified as the default transport in [RFC 2543](https://www.rfc-editor.org/rfc/rfc2543) when no transport
   was present in the SIP URI.  However, another transport, such as TCP,
   MAY be used if the guidelines of SIP mandate it for this particular
   request.  That is the case, for example, for requests that exceed the
   path MTU.
```

We can see the transport parameter is the first criteria for transport selection.

The way PJSIP is currently implemented, if we have a _sips_ with a custom transport (for example websockets 'transport=ws'), it will always forward the request to the TLS transport instead of any potencial custom transport implemented to handle the request properly (in this example a custom transport to deal with web sockets).

[...]

Issue #1740, opened 2014-02-25
When request URI uses sips scheme, TLS must always be used (thanks Peter Koletzki for the report)


pjsipbot:
_2014-02-25 05:59:56: @nanangizz_ _created the issue on trac ticket [1740](https://trac.pjsip.org/repos/ticket/1740)_

As specified in [RFC3261 8.1.2](http://tools.ietf.org/html/rfc3261#section-8.1.2):
```
   Independent
   of which URI is used as input to the procedures of [4], if the
   Request-URI specifies a SIPS resource, the UAC MUST follow the
   procedures of [4] as if the input URI were a SIPS URI.

   Local policy MAY specify an alternate set of destinations to attempt.
   If the Request-URI contains a SIPS URI, any alternate destinations
   MUST be contacted with TLS.
```
___
_2014-02-25 07:19:33: @nanangizz_ _changed **milestone** from **release-2.2** to **release-2.3**_

___
_2014-02-27 06:16:38: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2014-02-27 06:16:38: @nanangizz_ _set **resolution** to **fixed**_

___
_2014-02-27 06:16:38: @nanangizz_ _commented_

In [r4770](../commit/1c9dd69383473d0413d2010ecfb7ce6334b23f7d):
Fixed #1740: TLS will be used whenever request URI uses "sips" scheme regardless the target-URI scheme/transport. This behavior is configurable via 'PJSIP_DONT_SWITCH_TO_TLS' in compile-time, or 'pjsip_cfg()->disable_tls_switch' in run-time.

Issue #1735, opened 2014-02-17
Terminate the call if 200/OK response to INVITE to SIPS target is not secure


pjsipbot:
_2014-02-17 05:06:05: @bennylp_ _created the issue on trac ticket [1735](https://trac.pjsip.org/repos/ticket/1735)_

The standard says that a SIPS dialog should be secure end to end. However some proxies or UASes disobey this requirement and give back plain SIP URI in the Record-Route or Contact header. We should terminate such call.

This ticket will implement !Contact/Record-Route header checks in incoming initial/subsequent INVITE/UPDATE request or 2xx response to INVITE/UPDATE. Contact URI should use "`sips`" scheme and Record-Route URI should use either "`sips`" scheme or "`transport=tls`" param. As UAC, if it receives 200/OK response to initial/subsequent INVITE/UPDATE with insecure !Contact/Record-Route, it will terminate the session. As UAS, if it receives initial/subsequent INVITE/UPDATE request with insecure !Contact/Route headers, it will send response with SIP status code 480 and Warning header: 381 "SIPS required".

This feature can be disabled by setting `pjsip_cfg()->endpt.disable_secure_dlg_check` to `PJ_TRUE` (default is `PJ_FALSE`).

Here are some texts from the standard:
 - All hops must use TLS (should we check if all Record-Route URIs use "sips" scheme or "transport=tls" param? Currently we'll just check the top Record-Route header)
   - http://tools.ietf.org/html/rfc5630#section-4:
     ```
This document specifies that SIPS means that the SIP resource
designated by the target SIPS URI is to be contacted securely, using
TLS on each hop between the UAC and the remote UAS (as opposed to
only to the proxy responsible for the target domain of the Request-
URI).
     ```
 - Contact header must use "sips" scheme:
   - in request
     http://tools.ietf.org/html/rfc5630#section-5.1.1:
     ```
As mandated by [RFC3261], Section 8.1.1.8, in a request, "if the
Request-URI or top Route header field value contains a SIPS URI, the
Contact header field MUST contain a SIPS URI as well".
     ```
   - in target refresh
     http://tools.ietf.org/html/rfc5630#section-5.1.1.2:
     ```
When a target refresh occurs within a dialog (e.g., re-INVITE
request, UPDATE request), the UAC MUST include a Contact header field
with a SIPS URI if the original request used a SIPS Request-URI.
     ```
   - in response
     http://tools.ietf.org/html/rfc3261#section-12.1.1:
     ```
[...]

PR #3332, opened 2023-01-17
Allow transport other than TLS to be used for SIPS scheme

sauwming:
As reported in #3306 (SIPS URIs are being forwarded to TLS transport regardless of 'transport=' parameter).

In #1735 and #1740, we use TLS for SIPS URI based on the RFCs such as RFC 3261 and 5630:
```
Local policy MAY specify an alternate set of destinations to attempt.
   If the Request-URI contains a SIPS URI, any alternate destinations
   MUST be contacted with TLS.
```
and
```
This document specifies that SIPS means that the SIP resource
   designated by the target SIPS URI is to be contacted securely, using
   TLS on each hop between the UAC and the remote UAS (as opposed to
   only to the proxy responsible for the target domain of the Request-
   URI).  It is outside of the scope of this document to specify what
   happens when a SIPS URI identifies a UAS resource that "maps" outside
   the SIP network, for example, to other networks such as the Public
   Switched Telephone Network (PSTN).
```

However, note that the RFC makes a specific note that this applies only in a SIP network, so TLS is actually not mandatory, such as described in RFC 7118:
```
The "sips" scheme in a SIP URI dictates that the entire request path
   to the target be secure.  If such a path includes a WebSocket
   connection, it MUST be a secure WebSocket connection.
```

Thus this PR is to make the following modifications:
- The implementation in `sip_util.c` `pjsip_get_dest_info()` is actually already correct. We only require the transport to be secure for SIPS scheme, however the documentation and log message can be misleading as it seems to suggest that the secure transport has to be only TLS. So we fix the doc and log here. The setting and macro names `disable_tls_switch` and `PJSIP_DONT_SWITCH_TO_TLS` however, will remain unchanged for backward compatibility.
- Fixing the contact generation in pjsua acc that is hardcoded to TLS.
- Remove requirement for the transport to be reliable as there doesn't seem to be any RFC recommendation that mandates it.

Note: this PR also modifies the behaviour of API `pjsip_transport_get_type_from_flag()` to return the first transport type that contains the specified flags, instead of the one that matches the exact flag.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -3434,7 +3434,7 @@ pj_status_t pjsua_acc_get_uac_addr(pjsua_acc_id acc_id,
     if (PJSIP_URI_SCHEME_IS_SIPS(sip_uri)) {
         unsigned flag, tp_flag;
 
-        tp_flag = (PJSIP_TRANSPORT_SECURE | PJSIP_TRANSPORT_RELIABLE);
+        tp_flag = PJSIP_TRANSPORT_SECURE;
         flag  = pjsip_transport_get_flag_from_type(tp_type);
         if ((flag & tp_flag) != tp_flag) {
             tp_type = pjsip_transport_get_type_from_flag(tp_flag);
@@ -3817,7 +3817,7 @@ PJ_DEF(pj_status_t) pjsua_acc_create_uas_contact( pj_pool_t *pool,
     if (PJSIP_URI_SCHEME_IS_SIPS(sip_uri)) {
         unsigned flag, tp_flag;
 
-        tp_flag = (PJSIP_TRANSPORT_SECURE | PJSIP_TRANSPORT_RELIABLE);
+        tp_flag = PJSIP_TRANSPORT_SECURE;
         flag  = pjsip_transport_get_flag_from_type(tp_type);
         if ((flag & tp_flag) != tp_flag) {
             tp_type = pjsip_transport_get_type_from_flag(tp_flag);
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -3425,12 +3425,21 @@ pj_status_t pjsua_acc_get_uac_addr(pjsua_acc_id acc_id,
     }
 
     /* Get transport type of the URI */
-    if (PJSIP_URI_SCHEME_IS_SIPS(sip_uri))
-        tp_type = PJSIP_TRANSPORT_TLS;
-    else if (sip_uri->transport_param.slen == 0) {
+    if (sip_uri->transport_param.slen == 0) {
         tp_type = PJSIP_TRANSPORT_UDP;
     } else
         tp_type = pjsip_transport_get_type_from_name(&sip_uri->transport_param);
+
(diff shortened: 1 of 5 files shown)


pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -3425,13 +3425,22 @@ pj_status_t pjsua_acc_get_uac_addr(pjsua_acc_id acc_id,
...
+    if (sip_uri->transport_param.slen == 0) {
         tp_type = PJSIP_TRANSPORT_UDP;
     } else
         tp_type = pjsip_transport_get_type_from_name(&sip_uri->transport_param);
 
+    /* If the URI uses sips scheme, make sure we use secure transport. */
+    if (PJSIP_URI_SCHEME_IS_SIPS(sip_uri)) {
+        unsigned flag, tp_flag;
+
+        tp_flag = (PJSIP_TRANSPORT_SECURE | PJSIP_TRANSPORT_RELIABLE);
```
nanangizz, line 3437: why `RELIABLE` is also added?

sauwming: It's the same requirement as in `pjsip_get_dest_info()`:
https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_util.c#L829

The code doesn't provide the reason and I can't find any recommendation in the RFC, but I believe this is because in practice, most secure transport protocols are also designed to be reliable, because the security mechanisms are useless if the data is not delivered successfully.

nanangizz: Was just wondering if reliability is required too by SIPS by standard, because I am not sure if websocket is reliable (so if someone implemented it in the future, there will be `RELIABLE` flag in it, or perhaps whether SIPS using DTLS is allowed). Or perhaps it is because currently we only has TLS for SIP secure transport which is reliable?

sauwming: So do you think we should remove the reliability requirement or just leave it as is?

nanangizz: Well, that's my question actually :)

From a quick search, it looks like SIPS over DTLS was considered:
https://www.ietf.org/archive/id/draft-jennings-sip-dtls-05.html
```
Requests with a SIPS URI can be sent over DTLS as well as TLS. 
```
The draft never becomes RFC though.

Anyway, from the draft above and other SIPs related RFCs which do not seem to explicitly mention about reliable/unreliable transports, it looks like reliability is not mandatory in SIPS, so IMO it should be removed. But not sure about the practicality, especially for PJSIP which currently only has TLS.

sauwming: Okay, I think I found out the reason why RELIABLE flag was included. I believe it's because the function `pjsip_transport_get_type_from_flag()` now searches for exact match for the specified flags. So I modified it to find transport that contains the flags instead.

Note that this will cause a behavioural change. Within the library itself, the function is currently only used in `pjsip_get_dest_info()` in `sip_util.c`, so there's no issue. While for apps that directly call the function, it should have no problem as well, except perhaps for apps that significantly modify SIP transports.

nanangizz (approved):

trengginas (approved):

PR #3336, opened 2023-01-21
Extend algorithm to find account for incoming calls in PJSUA(2)

wosrediinanatour:
With issue #1688, it is possible to set different request and To-header URIs for outgoing SIP INVITEs. For incoming INVITEs, where both URIs are SIP URIs, the algorithm that finds the corresponding account considers the To header only:
If the username of the To-header is not equal to the account ID, the first account of same transport/hostname or the default account is used.

This algorithm is now extended to use the request URI's username as lowest weight for the score. If the To header's username does not match any account ID (local contact), the request URI's username may help.

Possible application: Similar to using a telephone URI in the To-header, it might be a SIP URI containing a telephone number. In case of telephone URI, PJSUA already uses the request URI.

Ref: RFC 3261, 8.2.2.1 To and Request-URI
   The To header field identifies the original recipient of the request
   designated by the user identified in the From field.  The original
   recipient may or may not be the UAS processing the request, due to
   call forwarding or other proxy operations.  A UAS MAY apply any
   policy it wishes to determine whether to accept requests when the To
   header field is not the identity of the UAS.  However, it is
   RECOMMENDED that a UAS accept requests even if they do not recognize
   the URI scheme (for example, a tel: URI) in the To header field, or
   if the To header field does not address a known or current user of
   this UAS.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -3192,11 +3192,22 @@ PJ_DEF(pjsua_acc_id) pjsua_acc_find_for_outgoing(const pj_str_t *url)
 /*
  * This is an internal function to find the most appropriate account to be
  * used to handle incoming calls.
+ *
+ * If the To header contains a SIP(S) URI, then it will be used for
+ * finding a appropriate account, else the request URI.
+ * The function selects an account by weighted score.
+ *
+ * The matching priority order is: transport type (matched or not set),
+ * domain part, user part, and the user part of the request URI (if it is
+ * a SIP URI). Note that the transport type has higher priority as 
+ * unmatched transport type may cause failure in sending response.
  */
 PJ_DEF(pjsua_acc_id) pjsua_acc_find_for_incoming(pjsip_rx_data *rdata)
 {
     pjsip_uri *uri;
+    pjsip_uri *request_uri;
     pjsip_sip_uri *sip_uri;
+    pjsip_sip_uri *request_sip_uri = NULL;
     pjsua_acc_id id = PJSUA_INVALID_ID;
     int max_score;
     unsigned i;
@@ -3209,10 +3220,10 @@ PJ_DEF(pjsua_acc_id) pjsua_acc_find_for_incoming(pjsip_rx_data *rdata)
     }
 
     uri = rdata->msg_info.to->uri;
+    request_uri = rdata->msg_info.msg->line.req.uri;
 
     PJSUA_LOCK();
 
-    /* Use Req URI if To URI is not SIP */
     if (!PJSIP_URI_SCHEME_IS_SIP(uri) &&
         !PJSIP_URI_SCHEME_IS_SIPS(uri))
     {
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -3208,28 +3210,30 @@ PJ_DEF(pjsua_acc_id) pjsua_acc_find_for_incoming(pjsip_rx_data *rdata)
...
     }
 
-    uri = rdata->msg_info.to->uri;
+    to_header_uri = rdata->msg_info.to->uri;
+    request_uri = rdata->msg_info.msg->line.req.uri;
 
     PJSUA_LOCK();
 
     /* Use Req URI if To URI is not SIP */
-    if (!PJSIP_URI_SCHEME_IS_SIP(uri) &&
```
wosrediinanatour, line 3216: This and the following if-statements that include PJSIP_URI_SCHEME_IS_SIP(S) could be improved. I don't know why 
```cpp
if (rdata->msg_info.msg->type == PJSIP_REQUEST_MSG)
```
exists, so I haven't touched it.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -3208,28 +3210,30 @@ PJ_DEF(pjsua_acc_id) pjsua_acc_find_for_incoming(pjsip_rx_data *rdata)
...
-        !PJSIP_URI_SCHEME_IS_SIPS(uri))
+    if (!PJSIP_URI_SCHEME_IS_SIP(to_header_uri) &&
+        !PJSIP_URI_SCHEME_IS_SIPS(to_header_uri))
     {
         if (rdata->msg_info.msg->type == PJSIP_REQUEST_MSG)
-            uri = rdata->msg_info.msg->line.req.uri;
+            to_header_uri = rdata->msg_info.msg->line.req.uri;
```
nanangizz, line 3223: The named var `to_header_uri` may actually contain request URI instead of To header URI, may it cause confusion later? Why not leave the `to_header_uri` as originally `uri`?

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -3208,28 +3210,30 @@ PJ_DEF(pjsua_acc_id) pjsua_acc_find_for_incoming(pjsip_rx_data *rdata)
...
-        !PJSIP_URI_SCHEME_IS_SIPS(uri)) 
+    if (!PJSIP_URI_SCHEME_IS_SIP(to_header_uri) &&
+        !PJSIP_URI_SCHEME_IS_SIPS(to_header_uri))
     {
         goto on_return;
     }
 
-    sip_uri = (pjsip_sip_uri*)pjsip_uri_get_uri(uri);
+    to_header_sip_uri = (pjsip_sip_uri*)pjsip_uri_get_uri(to_header_uri);
+    request_sip_uri = (pjsip_sip_uri*)pjsip_uri_get_uri(request_uri);
```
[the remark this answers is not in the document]
wosrediinanatour: I now rewrote even this part. Please take a look at the new commit.

pjsip/src/pjsua-lib/pjsua_acc.c
wosrediinanatour: I now rewrote even this part. Please take a look at the new commit.

pjsip/src/pjsua-lib/pjsua_acc.c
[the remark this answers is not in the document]
wosrediinanatour: Now I understand what you mean. Will fix it soon!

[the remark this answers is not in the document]
wosrediinanatour: Fixed in the last commit!

pjsip/src/pjsua-lib/pjsua_acc.c
wosrediinanatour: You are right. Fixed in the last commit!

wosrediinanatour:
I have incorporated your, @nanangizz, comments. This means I reverted some of my changes and now it is closer to the original code.

wosrediinanatour:
Updated it again: now it should ignore the request URI's user part in the algorithm, if it is a Tel URI.

nanangizz (approved):

nanangizz:
Looks good to me, thanks.

sauwming (approved):
looks okay to me, too. except the log doesn't seem to be necessary (and not exactly clear for users what the score is for). Either remove it or make it more clear, e.g.:
- should be level 6, since most of the time users do not need this info
- the text: "Acc selected for incoming call: #%u, score: %d"

wosrediinanatour:
@sauwming: 
I approached PJPROJECT from the PJSUA2 layer & book. Hence, for me, it was not clear, how an account is found without looking a the code. And even then, having a log was really useful. Due to the "bit mask", it is easy to infer why the algorithm chooses an account.

I have updated the commit incorporating your comment - see https://github.com/pjsip/pjproject/compare/4f6cc25dfa5cdcaf87c563c1cab33b55ed43a6ae..a239423760118aeeab91d63252d1a63d630ef860.

sauwming (approved):

nanangizz (approved):

PR #3343, opened 2023-01-27
Fix SIP-module documentation regarding return values

wosrediinanatour:
Constants PJ_SUCCESS and PJ_FALSE are both zero. If a callback would return PJ_FALSE, the message would still be transmitted.


sauwming (approved):

nanangizz (approved):
Nice catch!

PR #3344, opened 2023-01-27
Hangup not yet confirmed outgoing calls in IP change

nanangizz:
There was PR #2737 for avoiding call disconnection on IP change, but its scope is limited to a confirmed call. Also this PR #2683 for maintaining INVITE transactions in the UAS/callee side. While for UAC/caller in the connecting state, it is actually a bit more complex, an attempt to handle it has been done in PR #3288 (still unsolved).

Some of the challenges are:
- as the INVITE transaction is not completed yet, by standard we cannot send another INVITE to update remote about our new address, we can only use UPDATE and not all endpoints support UPDATE,
- if the initial INVITE contains SDP offer, and the SDP negotiation is not completed yet, also by standard, the UPDATE cannot contain another SDP offer, we need to renegotiate the SDP to update the media addresses,
- updating routing info for the INVITE transaction so remote/server can send its response to the correct latest address, but is not a simple task as remote/server may already send the response before our UPDATE is sent.

So for now, let's just hang up not confirmed outgoing call in IP change scenario. Note that the remote may see a dangling call as there may be race condition, e.g: remote has sent some response but not received by local/caller.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -4281,6 +4281,31 @@ pj_status_t pjsua_acc_handle_call_on_ip_change(pjsua_acc *acc)
                                                              status,
                                                              &info);
                 }
+            }
+            else if (call_info.role==PJSIP_ROLE_UAC &&
+                     call_info.state < PJSIP_INV_STATE_CONFIRMED)
+            {
+                /* Hangup not yet confirmed outgoing calls */
+                acc->ip_change_op = PJSUA_IP_CHANGE_OP_ACC_HANGUP_CALLS;
+                PJ_LOG(3, (THIS_FILE, "Unconfirmed outgoing call to %.*s: "
+                           "hangup triggered by IP change",
+                           (int)call_info.remote_info.slen,
+                           call_info.remote_info.ptr));
+
+                status = pjsua_call_hangup(i, PJSIP_SC_GONE, NULL, NULL);
+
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -4269,6 +4269,31 @@ pj_status_t pjsua_acc_handle_call_on_ip_change(pjsua_acc *acc)
...
+
+                    pjsua_var.ua_cfg.cb.on_ip_change_progress(
+                                                             acc->ip_change_op,
+                                                             status,
+                                                             &info);
+                }
+            }
+            else if (call_info.role==PJSIP_ROLE_UAC &&
+                     call_info.state < PJSIP_INV_STATE_CONFIRMED)
+            {
```
sauwming, line 4287: why not combine it with the above?

nanangizz: - For readibility, because they are conceptually different. The above is intentional hangup via configured settings, the latter is forced hangup due to specific conditions.
- Also the actions can also be different, for example with the current patch, `pjsua_call_hangup()` which will send CANCEL and CANCEL may not be useful if remote does not receive the INVITE or remote have sent its final answer (either success/2xx or failure/3xx-6xx), so perhaps terminating the call forcefully somehow may be better.

sauwming (approved):

trengginas (approved):

PR #3347, opened 2023-01-31
fix some code format error

jimying:
as title
The patch #3292 expanded tab to spaces, but there still some format error.

This is just a beginning!
Now only [src/pjmedia-codec] is searched, other modules need to continue.
Hope other can participate, it's boring :)


sauwming:
Thank you for your help. It's a tedious task indeed! Let us know when you're done so we can merge this (you don't have to complete the whole project of course) .

jimying:
I give up manual search. 
Too many places need to be modified, it is a wise choice to use the automation tool clang-format.

Issue #3278, opened 2022-10-31
~5x RSS memory increase with 9b37e587bc901d65ef5ca7b4ddeb5b9ac2c5267f (2.12 and newer)


mscdex:
### Describe the bug

9b37e587bc901d65ef5ca7b4ddeb5b9ac2c5267f (determined via git bisect) results in a ~5x RSS memory increase with a basic, non-SRTP, non-ICE, non-TURN, audio-only pjsua-based app immediately on startup/init. I briefly looked at the changes and didn't see any obvious causes. valgrind seems to report that the extra memory is all still reachable.

I also tried the latest commit on the master branch (currently 74569f926a2f553006d4bbac70c7371dfb1eb400) and the memory issue still existed.

If I revert (only) the problematic commit on either master or 2.12/2.12.1, the increased RSS memory usage issue goes away.

### Steps to reproduce

1. Create and start a very basic pjsua app -- no calls needed
2. Observe (immediate) RSS usage in a tool such as htop

### PJSIP version

multiple

### Context

- Tested on Linux 5.4.0
- `config_site.h`:

  ```
  #define PJ_SCANNER_USE_BITWISE 0
  #undef PJ_OS_HAS_CHECK_STACK
  #define PJ_OS_HAS_CHECK_STACK  0
  #define PJ_IOQUEUE_MAX_HANDLES 8192
  #define PJSIP_MAX_TSX_COUNT    ((640*1024)-1)
  #define PJSIP_MAX_DIALOG_COUNT ((640*1024)-1)
  #define PJ_DEBUG     0
  #define PJSIP_SAFE_MODULE    0
  #define PJ_HAS_STRICMP_ALNUM   0
  #define PJSUA_MAX_CALLS    4096
  #define PJMEDIA_HAS_SPEEX_AEC 0
  #define PJSUA_MAX_PLAYERS     4096
  #define PJSUA_MAX_RECORDERS     4096
  #define PJSUA_MAX_CONF_PORTS    (PJSUA_MAX_CALLS+2*PJSUA_MAX_PLAYERS)
  #define PJMEDIA_ADVERTISE_RTCP      0
  #define PJSIP_MAX_URL_SIZE (384)
  #define PJ_LOG_MAX_LEVEL   5
  #define PJMEDIA_HAS_SRTP 0
  #define PJMEDIA_HAS_VIDEO 0
  #define PJ_ENABLE_EXTRA_CHECK 1
[...]

PR #3349, opened 2023-01-31
Optimize memory usage for SRTP ROC checking functionality

sauwming:
To fix https://github.com/pjsip/pjproject/issues/3278.

To implement #2846 (maintenance of SRTP ROC), we add two fields `pjmedia_stream_info` and `pjmedia_stream_info` into `pjsua_call_media`. 

The size of each structure itself is "only" 2k, but for large number of max calls, this can significantly increase memory usage.
For example, if max calls is 4K:
4K * 16 (PJSUA_MAX_CALL_MEDIA) * 2 (media+prov_media) * 2 (audio&video stream info) * 2K (sizeof(struct)) = 512 MB.

Since not all the stream info fields are needed, we only keep the necessary ones.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua_internal.h
+++ b/pjsip/include/pjsua-lib/pjsua_internal.h
@@ -80,10 +80,8 @@ struct pjsua_call_media
                                          bit 1     : timestamp flag         */
 
     pjmedia_type         prev_type;            /**< Previous media type     */
-    pj_sockaddr          prev_aud_local_addr;  /**< Prev aud local address  */
-    pj_sockaddr          prev_aud_rem_addr;    /**< Prev aud remote address */
-    pj_sockaddr          prev_vid_local_addr;  /**< Prev vid local address  */
-    pj_sockaddr          prev_vid_rem_addr;    /**< Prev vid remote address */
+    pj_sockaddr          prev_local_addr;      /**< Prev media local address*/
+    pj_sockaddr          prev_rem_addr;        /**< Prev media remote addr  */
     pj_bool_t            prev_srtp_use;        /**< Prev SRTP use           */
     pjmedia_srtp_info    prev_srtp_info;       /**< Prev SRTP transport info*/
     pj_bool_t            prev_ice_use;         /**< Prev ICE use            */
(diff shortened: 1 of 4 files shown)


pjsip/include/pjsua-lib/pjsua_internal.h
```diff
@@ -79,12 +79,14 @@ struct pjsua_call_media
...
-    pjmedia_stream_info         prev_aud_si;   /**< Prev audio stream info  */
-    pjmedia_vid_stream_info     prev_vid_si;   /**< Prev video stream info  */
-    pj_bool_t                   prev_srtp_use; /**< Prev SRTP use           */
-    pjmedia_srtp_info           prev_srtp_info;/**< Prev SRTP transport info*/
-    pj_bool_t                   prev_ice_use;  /**< Prev ICE use            */
+    pjmedia_type         prev_type;            /**< Previous media type     */
+    pj_sockaddr          prev_aud_local_addr;  /**< Prev aud local address  */
+    pj_sockaddr          prev_aud_rem_addr;    /**< Prev aud remote address */
+    pj_sockaddr          prev_vid_local_addr;  /**< Prev vid local address  */
+    pj_sockaddr          prev_vid_rem_addr;    /**< Prev vid remote address */
```
nanangizz, line 86: Can `aud` & `vid` in local & remote addresses be merged or perhaps use union for a little bit more optimization?

sauwming: I'm not sure how to merge them since they're used at the same time?
https://github.com/pjsip/pjproject/pull/2846/files#diff-29def4cfd6b2e0bb7bc454130942b8d60a375d2fa5ac8aece3bef6639c569e83R3346

nanangizz: `pjsua_call_media` is supposed to be one media type (i.e: either audio or video) at a time and cannot be both?
So, the pointed code is supposed to be `if` + `else (if)` instead of two `if`s perhaps?

nanangizz (approved):

trengginas (approved):

sauwming:
On a side note, how about reducing the macro PJMEDIA_MAX_SDP_MEDIA from the current default of 16 (perhaps a separate PR?) The number of medias per call rarely goes beyond 2 (audio&video) so reducing the default to 4, or even 2, seems to be beneficial.

Issue #1616, opened 2013-02-13
Group lock and other foundation in PJLIB for fixing synchronization issues


pjsipbot:
_2013-02-13 07:58:29: @bennylp_ _created the issue on trac ticket [1616](https://trac.pjsip.org/repos/ticket/1616)_

This ticket contains works to be/that have been done in PJLIB to lay foundation for fixing various synchronization issues in upper layers such as PJNATH (see #1617). The details of the work are as follow.

-*Group Lock**

Implement [Group Lock](https://trac.pjsip.org/repos/wiki/Group_Lock), a new synchronization object for dealing with deadlock and session management. Please see the link for more info.

-*More robust timer heap**

 - Integration of group lock to the timer heap. A new API is introduced: `pj_timer_heap_schedule_w_grp_lock()`, which adds reference counter to the group lock when timer is scheduled, and automatically decrements it after the timer expires or cancelled, to prevent the group lock from being destroyed when a timer callback is about to be executed. This API also sets the timer "id" automatically and atomically.
 - New convenience API: `pj_timer_heap_cancel_if_active()` to cancel a timer without having to firstly check if it is active or not.

-*Changes is ioqueue and active socket**

 - New API `pj_ioqueue_register_sock2()` that takes group lock argument. The group lock would be incremented before callback is called to prevent it from being destroyed while callback is about to run.
 - Added `grp_lock` in `pj_activesock_cfg`, to specify which group lock to use by the ioqueue key.

-*Miscellaneous**

 - New `PJ_EGONE` error code. It is used when `pj_grp_lock_dec_ref()` or `pj_grp_lock_release()` causes the group lock to be destroyed.
 - New `PJ_RACE_ME(x)` macro which can be inserted in various strategic locations. During testing, the macro can be replaced with `pj_thread_sleep(x)` in order to trigger race condition to happen.
 - Event object (similar to Windows event) implementation for `os_core_unix.c` based on Pthread condition variable. This is used by the pjnath-test testing program to synchronize the tests.

___
_2013-02-13 08:00:45: @bennylp_ _edited the issue description_

___
_2013-02-21 11:18:38: @bennylp_ _changed **status** from **new** to **closed**_

___
_2013-02-21 11:18:38: @bennylp_ _set **resolution** to **fixed**_

___
_2013-02-21 11:18:38: @bennylp_ _commented_

In [r4359](../commit/ac932bb52288695e3d20bb33b0bd9eec21260201):
[...]

PR #3358, opened 2023-02-03
Fixed ioq winnt build failure

### LINKED ISSUE #2576: windows ioqueue use iocp (ioqueue_winnt.c) build fail
**Describe the bug**
use visual studio 2015, open project,  select ioqueue_winnt.c replace ioqueue_select.c,   
build fail, no function ' pj_ioqueue_register_sock2'

iocp file (ioqueue_winnt.c) not sync update, please check it.

### LINKED ISSUE #3291: ioqueue_winnt.c is outdated and can't be linked
### Describe the bug

ioqueue_winnt.c is outdated and can't be linked because of lack new API like pj_ioqueue_register_sock2. By default Windows builds use ioqueue_select.c but this implementation has limits. ioqueue_winnt.c has no limitations but just outdated.

### Steps to reproduce

enable ioqueue_winnt.c compilation and build

### PJSIP version

2.11.1

### Context

Windows builds

### Log, call stack, etc

```shell
1>pjlib-util-i386-Win32-vc14-Release-Static.lib(resolver.obj) : error LNK2001: unresolved external symbol _pj_ioqueue_register_sock2
```

sauwming:
****IMPORTANT**:
Although ioq-winnt (IOCP) backend can now build successfully and pass the unit tests, it lacks the group lock functionality so users should only use it for experimental testing purposes and should NOT use it for actual development and production.

To fix #3291 and close #2576.

### DIFF
--- a/pjlib/src/pj/ioqueue_winnt.c
+++ b/pjlib/src/pj/ioqueue_winnt.c
@@ -614,7 +614,8 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock2(pj_pool_t *pool,
     /* Group lock */
     rec->grp_lock = grp_lock;
     if (rec->grp_lock) {
-        pj_grp_lock_add_ref_dbg(rec->grp_lock, "ioqueue", 0);
+        PJ_TODO(INTEGRATE_GROUP_LOCK);
+        // pj_grp_lock_add_ref_dbg(rec->grp_lock, "ioqueue", 0);
     }
 
     *key = rec;
--- a/pjlib/src/pj/ioqueue_winnt.c
+++ b/pjlib/src/pj/ioqueue_winnt.c
@@ -614,6 +614,9 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock2(pj_pool_t *pool,
     /* Group lock */
     rec->grp_lock = grp_lock;
     if (rec->grp_lock) {
+        /* IOCP backend doesn't have group lock functionality, so
+         * you should not use it other than for experimental purposes.
+         */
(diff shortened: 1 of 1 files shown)


pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -586,6 +611,12 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock( pj_pool_t *pool,
         return PJ_RETURN_OS_ERROR(GetLastError());
     }
 
+    /* Group lock */
+    rec->grp_lock = grp_lock;
+    if (rec->grp_lock) {
+        pj_grp_lock_add_ref_dbg(rec->grp_lock, "ioqueue", 0);
```
nanangizz, line 617: - there should be the `pj_grp_lock_dec_ref()` pair?
- as in other ioqueue, the group lock `add/dec_ref()` is invoked surrounding event callbacks to prevent premature destroy/unregistration by other thread while the app is inside the callback?

sauwming: Ah, right. Integrating group lock will be a much more complicated task.

Just a bit of background, the group lock introduction was done in #1616 dated Feb 2013 for release 2.1 and was integrated into ioq select and even (the then-experimental) epoll backend. Ioq-winnt (IOCP) was notably missing and not sure whether it was simply overlooked or if there were some implementation challenges during the integration. But the obvious thing is that IOCP hasn't been working for more than a decade and there didn't seem to be much interest (?) either.

So for now I suggest the PR is just to fix the build failure. If more users start using this backend again, we may consider revisiting the group lock integration.

nanangizz (approved):

PR #3360, opened 2023-02-03
FIx Android Mediacodec h264 SPS/PPS decoder config

### LINKED ISSUE #2831: Not fully correct video AMediaCodec decoder configuration
### Describe the bug

All of the video codec decoders in AMediaCodec implementation have *I420* as a `and_media_codec_data.prm->dec_fmt.id` is set in `add_codec`.
As a result SPS/PPS data won't be set to the AMediaCodec format in the `configure_decoder` function as `and_media_data->prm->dec_fmt.id == PJMEDIA_FORMAT_H264` will be always false

### Steps to reproduce

Build for Android and enable AMediaCodec
Start or accept an H264 video call with SPS/PPS data

### PJSIP version

2.11

### Context

Any Android device
PjSip 2.11+ build with AMediaCodec enabled 

### Log, call stack, etc

```shell
Not required
```

trengginas:
This is to fix #2831, which will incorrectly configure SPS/PPS on Android Mediacodec (H264 decoder).

### DIFF
--- a/pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
+++ b/pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
@@ -419,7 +419,9 @@ static pj_status_t configure_decoder(and_media_codec_data *and_media_data) {
     AMediaFormat_setInt32(vid_fmt, AND_MEDIA_KEY_ENCODER, 0);
     AMediaFormat_setInt32(vid_fmt, AND_MEDIA_KEY_PRIORITY, 0);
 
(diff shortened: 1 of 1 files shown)


sauwming (commented):
Just to clarify, this is different than the patch created in #2831, right?

trengginas:
> Just to clarify, this is different than the patch created in #2831, right?

Yes that is correct. Don't feel right using `enc_fmt` on configuring decoder.

sauwming (approved):

nanangizz (approved):

PR #3365, opened 2023-02-06
Use clang-format  to format code

jimying:
There are some code format errors.
I tried to fix it manually but gave up. (#3347)
Now i plan to  use clang-format to fomat code.

The first commit "clang-format prepare" which do some prepares before format code:
1.  use 'clang-format off/on'  to keep current code format
2.  add  .clang-format and format.sh,   use format.sh tool to auto format code
     eg.  format pjlib dir only:   sh format.sh  pjlib
     eg.  format all dirs:   sh format.sh all

If  you accept, i will use format.sh to format all code,  and  commit changes as final version

PS: clang-format version must >=13


jimying:
@sauwming  @nanangizz @trengginas 

Preparations for clang-format are complete.
Next step is to format all code, run:
```bash
sh format.sh all
```
You can try and test it^_^

jimying:
I think clang-format is well-done except some comments' alignment. 
Temporarily closed

PR #3384, opened 2023-02-14
ffmpeg_codec: ffmpeg_codec_open() should use pjmedia_vid_codec_param_clone() copy codec param

jimying:
Currently use pj_memcpy() copy param:
https://github.com/pjsip/pjproject/blob/642179a0e5fcf02aa80b11c556a8cb68dc1199f7/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c#L1418

It's not memory safe in the attribute enc_fmtp and dec_fmtp 
```
typedef struct pjmedia_codec_fmtp
{
    pj_uint8_t      cnt;            /**< Number of parameters.          */
    struct param {
        pj_str_t    name;           /**< Parameter name.                */
        pj_str_t    val;            /**< Parameter value.               */
    } param [PJMEDIA_CODEC_MAX_FMTP_CNT]; /**< The parameters.          */
} pjmedia_codec_fmtp;
```
(The pj_str name/val memory maybe free but use later, pjmedia_vid_codec_param_clone()  use strdup clone to avoid it )

So  use pjmedia_vid_codec_param_clone( ) to replace   (To keep same with other codec OPenH264, VPX,vid_toolbox ..)

### DIFF
--- a/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
+++ b/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
@@ -163,7 +163,7 @@ typedef struct ffmpeg_codec_desc ffmpeg_codec_desc;
 typedef struct ffmpeg_private
 {
     const ffmpeg_codec_desc         *desc;
-    pjmedia_vid_codec_param          param;     /**< Codec param            */
+    pjmedia_vid_codec_param         *param;     /**< Codec param            */
     pj_pool_t                       *pool;      /**< Pool for each instance */
 
     /* Format info and apply format param */
@@ -196,7 +196,7 @@ typedef struct ffmpeg_private
                                                 /**< Expected output format of 
                                                      ffmpeg decoder         */
 
-    void                            *data;      /**< Codec specific data    */              
+    void                            *data;      /**< Codec specific data    */
 } ffmpeg_private;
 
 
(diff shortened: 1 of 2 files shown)


sauwming:
Thank you for your numerous contributions.

But sometimes it's difficult for us to know whether your PR is ready to be merged or if there are still more commits to come. So I'd suggest to create a draft PR instead and mark it as 'ready for review' once it's done. Alternatively, you can just write in the PR comments, or commit message, if the PR is done, or still in progress. Thanks again.

jimying:
@sauwming   sorry for confusing you.  This PR has been finished

sauwming (approved):

pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
```diff
@@ -1415,7 +1416,7 @@ static pj_status_t ffmpeg_codec_open( pjmedia_vid_codec *codec,
     PJ_ASSERT_RETURN(codec && attr, PJ_EINVAL);
     ff = (ffmpeg_private*)codec->codec_data;
 
-    pj_memcpy(&ff->param, attr, sizeof(*attr));
+    ff->param = pjmedia_vid_codec_param_clone(ff->pool, attr);
 
     /* Normalize encoding MTU in codec param */
     if (attr->enc_mtu > PJMEDIA_MAX_VID_PAYLOAD_SIZE)
```
jimying, line 1422: old code logic  has a problem:  if mtu changed,  not sync the modify to ff->param.

sauwming (approved):

trengginas (approved):

PR #3386, opened 2023-02-16
Formulate headers in new request upon receiving 3xx

### LINKED ISSUE #3084: Add support for embedded headers received in Contact header of SIP 3xx

sauwming:
When receiving 3xx response, we need to formulate headers in the new request according to [RFC 3261 section 8.1.3.4](https://www.rfc-editor.org/rfc/rfc3261#section-8.1.3.4) and [19.1.5](https://www.rfc-editor.org/rfc/rfc3261#section-19.1.5):

If you wish to implement your own custom function to handle the headers, you can do the following:
```
/* Optional: Declare this if you want to call the default handler */
void pjsip_inv_process_hparam(pjsip_inv_session *sess, 
                              const pj_str_t *hname, 
                              const pj_str_t *hvalue, 
                              pjsip_tx_data *tdata);

/* Declare this to get access to the external function handler pointer */
extern void (*pjsip_inv_process_hparam_ptr)(pjsip_inv_session *sess, 
                                     const pj_str_t *hname, 
                                     const pj_str_t *hvalue, 
                                     pjsip_tx_data *tdata);

/* Your custom processing function */
void custom_process_hparam(pjsip_inv_session *sess, 
                              const pj_str_t *hname, 
                              const pj_str_t *hvalue, 
                              pjsip_tx_data *tdata)
{
    /* Perform the custom handling here */
    /* .... */

    /* Call the default handler */
    pjsip_inv_process_hparam(sess, hname, hvalue, tdata);
}

int main() {
    /* Assign the function pointer to your own handler */
    pjsip_inv_process_hparam_ptr = &custom_process_hparam;
}
```

Notes (any default behaviour mentioned below can be overwritten by implementing the custom function above):
* In section 8.1.3.4, it is specified:
```
   It uses the "header" parameters to create header field values for the
   new request, overwriting header field values associated with the
   redirected request in accordance with the guidelines in Section 19.1.5.

Note that in some instances, header fields that have been
   communicated in the contact address may instead append to existing
   request header fields in the original redirected request.  As a
   general rule, if the header field can accept a comma-separated list
   of values, then the new header field value MAY be appended to any
   existing values in the original redirected request.  If the header
   field does not accept multiple values, the value in the original
   redirected request MAY be overwritten by the header field value
   communicated in the contact address.
```
By default, we only append `Call-Info` header as exemplified in 8.1.3.4. Other headers will be overwritten, if it already exists.
* In 19.1.5:
```
If the URI
   contains a method parameter, its value MUST be used as the method of
   the request.  The method parameter MUST NOT be placed in the
   Request-URI.
```
In the patch, we will terminate the session if the method is not INVITE.
* In 19.1.5:
```
   An implementation SHOULD NOT honor these obviously dangerous header
   fields: From, Call-ID, CSeq, Via, and Record-Route.

   An implementation SHOULD NOT honor any requested Route header field
   values in order to not be used as an unwitting agent in malicious
   attacks.

   An implementation SHOULD NOT honor requests to include header fields
   that may cause it to falsely advertise its location or capabilities.
   These include: Accept, Accept-Encoding, Accept-Language, Allow,
   Contact (in its dialog usage), Organization, Supported, and User-
   Agent.
```
By default, we will reject the headers.
* In 19.1.5:
```
   An implementation SHOULD verify the accuracy of any requested
   descriptive header fields, including: Content-Disposition, Content-
   Encoding, Content-Language, Content-Length, Content-Type, Date,
   Mime-Version, and Timestamp.
```
By default, we will include the above headers, except `Content-Type` or `Content-Length` since they will be automatically populated with the correct values by PJSIP. Application wishing to verify the headers need to implement the custom handler above.

To close #3084.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -83,6 +83,26 @@ static const pjsip_method pjsip_update_method =
 #define POOL_INIT_SIZE  256
 #define POOL_INC_SIZE   256
 
+/* Default action function to determine whether the header param with
+ * name hname and value hvalue should be included when formulating
+ * a new request based on the redirected request.
+ *
+ * Parameter hdr is the header in the original request that has the same
+ * hname.
+ * Parameter hdr and pool is typically needed only when you want to append
+ * the header to existing one.
+ */
+int pjsip_inv_get_hparam_action(pjsip_inv_session *sess, const pj_str_t *hname,
+                                const pj_str_t *hvalue, pjsip_hdr *hdr,
+                                pj_pool_t *pool);
+
+
+/* Application can override this if it wants to implement custom action. */
+int (*pjsip_inv_get_hparam_action_ptr)(pjsip_inv_session *, const pj_str_t *,
+                                       const pj_str_t *, pjsip_hdr *,
+                                       pj_pool_t *) =
+                                       &pjsip_inv_get_hparam_action;
+
 /*
  * Static prototypes.
  */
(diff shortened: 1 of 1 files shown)


nanangizz:
> In this patch, we choose to always overwrite the header, if exists, since there is no straightforward way to determine if a header value can be appended.

Perhaps if a header is 'descendant' of `pjsip_generic_array_hdr` (same `vptr` field), we can safely assume that it can be appended?

> Here we will ignore the above headers since there is no straightforward way to verify its accuracy.

The content info seem too important to ignore (the content may not be usable without them), if we cannot verify, perhaps passing it through is better than ignoring it? Except `Content-Length` perhaps.

An optimization idea, instead of redefining the header name string, perhaps we can use `pjsip_hdr_names[]` somehow?

sauwming:
> An optimization idea, instead of redefining the header name string, perhaps we can use `pjsip_hdr_names[]` somehow?

This is done.

> Perhaps if a header is 'descendant' of `pjsip_generic_array_hdr` (same `vptr` field), we can safely assume that it can be appended?

`generic_array_hdr_vptr` is not accessible, unless you want to expose it in the header file, or make it non-static like `pjsip_hrd_names`.

> > Here we will ignore the above headers since there is no straightforward way to verify its accuracy.
> The content info seem too important to ignore (the content may not be usable without them), if we cannot verify, perhaps passing it through is better than ignoring it? Except `Content-Length` perhaps.

How about `Content-Type`?

nanangizz:
> `generic_array_hdr_vptr` is not accessible, unless you want to expose it in the header file, or make it non-static like `pjsip_hrd_names`.

Perhaps adding a helper function like `pjsip_hdr_is_array()`? But looking it again, yes, this approach does not seem to be straightforward to implement as the header to check/compare is actually not header (yet), it is a `name=value` param pair? So perhaps the easiest way is to create a list of array header types, if we want to handle it.

> > > The content info seem too important to ignore (the content may not be usable without them), if we cannot verify, perhaps passing it through is better than ignoring it? Except `Content-Length` perhaps.
> 
> How about `Content-Type`?

Not sure, aren't there too many content-types? Or is it either multipart or sdp?

sauwming:
> Perhaps adding a helper function like `pjsip_hdr_is_array()`? But looking it again, yes, this approach does not seem to be straightforward to implement as the header to check/compare is actually not header (yet), it is a `name=value` param pair? So perhaps the easiest way is to create a list of array header types, if we want to handle it.

I have created a draft implementation to append the header if it's a generic array hdr (patch attached).
[redirect-append.txt](https://github.com/pjsip/pjproject/files/10754046/redirect-append.txt)
But upon further checking, generic array is currently used by Accept, Allow, Require, Supported, which we will reject anyway.

> > > > The content info seem too important to ignore (the content may not be usable without them), if we cannot verify, perhaps passing it through is better than ignoring it? Except `Content-Length` perhaps.
> > How about `Content-Type`?
> Not sure, aren't there too many content-types? Or is it either multipart or sdp?

The concern is not passing it through, but if it will overwrite existing headers. If Content-Type is overwritten with incorrect type, the session will fail.
The other headers should be fine to be passed through/overwritten though, since it's not critical.

So in the latest commit, I allow everything except type&length.

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -3066,6 +3068,107 @@ PJ_DEF(pj_status_t) pjsip_inv_process_redirect( pjsip_inv_session *inv,
...
+                                                  PJSIP_H_OTHER, hdr);
+                    }
+                    hdr = pjsip_generic_string_hdr_create(tdata->pool,
+                                                          &param->name,
+                                                          &param->value);
+                    pjsip_msg_add_hdr(tdata->msg, (pjsip_hdr*)hdr);
+                    param = param->next;
+                }
+
+                /* Remove method param */
```
bennylp, line 3145: Mention section 19.1.5 as reference

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -3066,6 +3068,107 @@ PJ_DEF(pj_status_t) pjsip_inv_process_redirect( pjsip_inv_session *inv,
...
+            req_uri = tdata->msg->line.req.uri;
+            if (PJSIP_URI_SCHEME_IS_SIP(req_uri) ||
+                PJSIP_URI_SCHEME_IS_SIPS(req_uri))
+            {
+                pjsip_sip_uri *uri;
+                pjsip_param *param;
+
+                uri = (pjsip_sip_uri *)pjsip_uri_get_uri(req_uri);
+                param = uri->header_param.next;
+                while (param != &uri->header_param) {
```
bennylp, line 3080: Mention section 8.1.3.4 and 19.1.5 as reference

bennylp:
What about putting the logic to reject, append, or replace header in a function that can be overridden by app?

```
// this function needs to be global so app can call it from its handler
int pjsip_inv_get_hparam_action(pjsip_inv_session *sess, const pj_str_t *hname)
{
   ... return 1, 2, 3 to ignore, append, or replace the header
}
// app can override it
int (*pjsip_inv_get_hparam_action_ptr)(...) = &pjsip_inv_get_hparam_action; 
```

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -3066,6 +3068,107 @@ PJ_DEF(pj_status_t) pjsip_inv_process_redirect( pjsip_inv_session *inv,
...
+                    }
+
+                    /* If the header already exists, we choose to
+                     * overwrite it.
+                     */
+                    hdr = pjsip_msg_find_hdr_by_name(tdata->msg, &param->name,
+                                                     NULL);
+                    if (hdr) {
+                        pjsip_msg_find_remove_hdr(tdata->msg,
+                                                  PJSIP_H_OTHER, hdr);
```
bennylp, line 3136: I'm not sure what's the correct processing for Authorization and Proxy-Authorization

sauwming:
Attached is the test log of the latest version (that includes method param, rejected headers, appended header, new additional header, and overwriting existing one).

[redirect-log.txt](https://github.com/pjsip/pjproject/files/10763637/redirect-log.txt)
(I remove the SDPs for more clarity)

bennylp (approved):
Suggestion to use PJ_ARRAY_SIZE

nanangizz (approved):

wosrediinanatour:
Would be cool to have callbacks in PJSUA and PJSUA2.

sauwming:
> Would be cool to have callbacks in PJSUA and PJSUA2.

It's not exactly a callback, but we provide a mechanism to provide your own custom handler. I've just updated the PR desc.

PR #3391, opened 2023-02-20
make the timer more precise

jimying:
When timer heap has more than 1 timers,
one timer callback may take long time, when check if the next timer expired,
the 'now' value is old (PJ_TIME_VAL_LTE(min_time_node, now) )  
which may cause timer is actually expired but not triggered.
So need update 'now' in the loop.

### DIFF
--- a/pjlib/src/pj/timer.c
+++ b/pjlib/src/pj/timer.c
@@ -76,8 +76,8 @@ typedef struct pj_timer_entry_dup
     /**
      * The duplicate copy.
      */
-    pj_timer_entry  dup;                
-    
+    pj_timer_entry  dup;
+
     /**
      * Pointer of the original timer entry.
      */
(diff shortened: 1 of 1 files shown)


pjlib/src/pj/timer.c
```diff
@@ -923,10 +923,13 @@ PJ_DEF(unsigned) pj_timer_heap_poll( pj_timer_heap_t *ht,
             slot = ht->timer_ids[GET_FIELD(ht->head_list.next, _timer_id)];
 #endif
             min_time_node = ht->heap[slot]->_timer_value;
+            /* Update now */
+            pj_gettickcount(&now);
         }
     }
     if (ht->cur_size && next_delay) {
         *next_delay = ht->heap[0]->_timer_value;
+        pj_gettickcount(&now);
```
sauwming, line 932: One small optimization that I can suggest is to add `if (count > 0)` here, since there shouldn't be much time difference if no callback is called.

nanangizz (approved):

sauwming (approved):

PR #3394, opened 2023-02-21
Allow dialog establishment when remote does not provide To tag

### LINKED ISSUE #1096: Assertion in sending UPDATE after receiving 200/INVITE response without to-tag
_2010-06-16 16:57:51: @nanangizz_ _created the issue on trac ticket [1096](https://trac.pjsip.org/repos/ticket/1096)_

nanangizz:
To fix #1096.

The assertion occurs in the following line because the dialog is not established:
https://github.com/pjsip/pjproject/blob/66c101f3d361761993a2a8a59082cb5a30d28e0d/pjsip/src/pjsip-ua/sip_inv.c#L3453-L3455

The standard (rfc3261 [12.1.2 UAC Behavior](https://www.rfc-editor.org/rfc/rfc3261#section-12.1.2)) specifies that it should be tolerated:
```
   A UAC MUST be prepared to receive a response without a tag in the To
   field, in which case the tag is considered to have a value of null.

      This is to maintain backwards compatibility with [RFC 2543](https://www.rfc-editor.org/rfc/rfc2543), which
      did not mandate To tags.
```

### DIFF
--- a/pjsip/src/pjsip/sip_dialog.c
+++ b/pjsip/src/pjsip/sip_dialog.c
@@ -1944,8 +1944,7 @@ void pjsip_dlg_on_rx_response( pjsip_dialog *dlg, pjsip_rx_data *rdata )
      */
     if ((dlg->state == PJSIP_DIALOG_STATE_NULL &&
          pjsip_method_creates_dialog(&rdata->msg_info.cseq->method) &&
-         (res_code > 100 && res_code < 300) &&
-         rdata->msg_info.to->tag.slen)
+         (res_code > 100 && res_code < 300))
          ||
         (dlg->role==PJSIP_ROLE_UAC &&
          !dlg->uac_has_2xx &&
--- a/tests/pjsua/scripts-sipp/uas-answer-200-without-to-tag.py
+++ b/tests/pjsua/scripts-sipp/uas-answer-200-without-to-tag.py
@@ -1,0 +1,6 @@
+#
+import inc_const as const
+
+PJSUA = ["--null-audio --max-calls=1 --no-tcp $SIPP_URI"]
+
+PJSUA_EXPECTS = [[0, const.STATE_CONFIRMED, "U"]]
--- a/tests/pjsua/scripts-sipp/uas-answer-200-without-to-tag.xml
+++ b/tests/pjsua/scripts-sipp/uas-answer-200-without-to-tag.xml
@@ -1,0 +1,86 @@
+<?xml version="1.0" encoding="ISO-8859-1" ?>
+<!DOCTYPE scenario SYSTEM "sipp.dtd">
+
+<!-- This program is free software; you can redistribute it and/or      -->
+<!-- modify it under the terms of the GNU General Public License as     -->
+<!-- published by the Free Software Foundation; either version 2 of the -->
+<!-- License, or (at your option) any later version.                    -->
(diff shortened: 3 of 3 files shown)


sauwming:
How about here?
https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_dialog.c#L716

nanangizz:
> How about here? https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_dialog.c#L716

Honestly I don't know whether it needs to be changed too or not. SIP dialog seems to be introduced in rfc3261 (searched for 'dialog' in rfc2543 found none), so perhaps we should just leave it as is?

sauwming:
What I found in the older RFC https://www.rfc-editor.org/rfc/rfc2543#section-6.37 regarding forking is this:
```
The "tag" parameter serves as a general mechanism to distinguish
   multiple instances of a user identified by a single SIP URL. As
   proxies can fork requests, the same request can reach multiple
   instances of a user (mobile and home phones, for example). As each
   can respond, there needs to be a means to distinguish the responses
   from each at the caller. The situation also arises with multicast
   requests. The tag in the To header field serves to distinguish
   responses at the UAC. It MUST be placed in the To field of the
   response by each instance when there is a possibility that the
   request was forked at an intermediate proxy. The "tag" MUST be added
   by UAS, registrars and redirect servers, but MUST NOT be inserted
   into responses forwarded upstream by proxies. The "tag" is added for
   all definitive responses for all methods, and MAY be added for
   informational responses from a UAS or redirect server. All subsequent
   transactions between two entities MUST include the "tag" parameter,
   as described in [Section 11](https://www.rfc-editor.org/rfc/rfc2543#section-11).
```
So, yes, the To tag must be present in the case of forking. But I'm just not sure why we should assert instead of just returning error.

nanangizz:
> So, yes, the To tag must be present in the case of forking. 
```
   A UAC MUST be prepared to receive a response without a tag in the To
   field, in which case the tag is considered to have a value of null.
```
It looks like no To tag is basically acceptable, i.e: equal to having tag with null value (whatever it means), I believe it is applicable for forking scenario too.

But then, the API is part of SIP dialog API (using `pjsip_dlg` prefix), while the concept of *dialog* does not seem to present in the older RFC, and in the newer RFC, To tag is mandatory. So we can say that the check for To tag in the code is proper.

> But I'm just not sure why we should assert instead of just returning error.

Yes, the library does not seem to verify the To tag existence before calling it (start [here](https://github.com/pjsip/pjproject/blob/66c101f3d361761993a2a8a59082cb5a30d28e0d/pjsip/src/pjsip/sip_ua_layer.c#L895)).

A similar check in `create_uas_dialog()` [here](https://github.com/pjsip/pjproject/blob/66c101f3d361761993a2a8a59082cb5a30d28e0d/pjsip/src/pjsip/sip_dialog.c#L378-L381
), the comment mentions that the application should check/verify it before.
But `pjsua_call_on_incoming()` does not seem to do so. So it is either the app or the `pjsua_dlg_fork()` that needs to be updated.

sauwming (approved):

trengginas (approved):

nanangizz:
For now, I guess let's just leave the other issues (normal check instead of assertion).

PR #3396, opened 2023-02-21
Use better initial values for pool initialization

sauwming:
This is similar to the issue reported in #3077 , where suboptimal pool allocation can cause delay that can stretch up to several seconds due to the expensive reallocation processes.

Here's the typical result of memory dump with default config:
```
    pept0x13600c400:   322776 of   328096 (98%) used
            med-ept:    43064 of    50176 (85%) used
       darwin video:    19712 of    20480 (96%) used
          sdl video:     7564 of     9024 (83%) used
         cbar video:     5096 of     5632 (90%) used
```

Here we also fix memory that are allocated too large (for codec we are apparently okay with allocating 4KB, which we don't need, but for other components, such as media and video!, we start too conservatively with only 0.5kb):
```
              speex:      312 of     4096 (7%) used
                gsm:      312 of     4096 (7%) used
               g711:      312 of     4096 (7%) used
               g722:      312 of     1024 (30%) used
                l16:      312 of     4096 (7%) used
```

### DIFF
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -47,14 +47,14 @@
  * Initial memory block for media endpoint.
  */
 #ifndef PJMEDIA_POOL_LEN_ENDPT
-#   define PJMEDIA_POOL_LEN_ENDPT               4096
+#   define PJMEDIA_POOL_LEN_ENDPT               8192
 #endif
 
 /**
  * Memory increment for media endpoint.
  */
 #ifndef PJMEDIA_POOL_INC_ENDPT
-#   define PJMEDIA_POOL_INC_ENDPT               8192
+#   define PJMEDIA_POOL_INC_ENDPT               4096
 #endif
 
 /**
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -47,14 +47,14 @@
  * Initial memory block for media endpoint.
  */
 #ifndef PJMEDIA_POOL_LEN_ENDPT
-#   define PJMEDIA_POOL_LEN_ENDPT               512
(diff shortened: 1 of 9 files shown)


nanangizz (approved):

bennylp:
IMO pool increment size should be limited to 4000, because:
- for system with constrained memory, after running for a while, the memory may become fragmented and it will be hard(er) to find large contiguous block,
- the size (i.e. 4000 + pool block header size < 4KB) matches the common page size of most processors (i.e. 4KB, see [here](https://devblogs.microsoft.com/oldnewthing/20210510-00/?p=105200)), and there may be performance benefit when the block fits in a single L1 page.
- I think it's big enough, and if app needs bigger, pool will allocate multiple of this size automatically anyway.

So for pool with large memory usage, I propose setting bigger initial size and limit the increment to 4000. Also for more general case, I think increment size should be <= initial size.

pjmedia/include/pjmedia/config.h
```diff
@@ -47,14 +47,14 @@
...
-#   define PJMEDIA_POOL_LEN_ENDPT               512
+#   define PJMEDIA_POOL_LEN_ENDPT               8192
 #endif
 
 /**
  * Memory increment for media endpoint.
  */
 #ifndef PJMEDIA_POOL_INC_ENDPT
-#   define PJMEDIA_POOL_INC_ENDPT               512
+#   define PJMEDIA_POOL_INC_ENDPT               4096
```
bennylp, line 57: I propose 4000 instead of 4096, just to be on the conservative side, in case the OS adds additional bytes to the block, to make sure the whole things fits in 4KB page. (similarly for the 8192 and 16384, trim them by ~100 bytes maybe)

trengginas (approved):

PR #3402, opened 2023-02-23
Misuse PJ_DECL in some function define

jimying:
as title
use PJ_DEF to replace


sauwming (approved):
Thanks for finding them.

nanangizz (approved):

Issue #3164, opened 2022-06-27
PJSUA2 - Example Pythons Script Unsolicited NOTIFY caused core dump


slipalong:
### Describe the bug

import pjsua2 as pj
import time
import threading

class Account(pj.Account):
    def onRegState(self, prm):
        print("***OnRegState: " + prm.reason)

    def onMwiInfo(self, prm):
        print("OnMwiState: " + prm.reason)

    def onBuddyState(self, prm):
        print("OnBuddyState: " + prm.reason)

    def onIncommingSubscribe(self, prm):
        print("OnSubscribeState: " + prm.reason)

def initalise_sip_stack():

    ep_cfg = pj.EpConfig()
    ep = pj.Endpoint()
    ep.libCreate()
    ep.libInit(ep_cfg)

    sipTpConfig = pj.TransportConfig()
    sipTpConfig.port = 5060
    ep.transportCreate(pj.PJSIP_TRANSPORT_UDP, sipTpConfig)
    ep.libStart()
    
    acfg = pj.AccountConfig()
    acfg.idUri = "sip:1234@192.168.1.10"
    acfg.regConfig.registrarUri = "sip:192.168.1.10"
    creds = pj.AuthCredInfo("digest", "ccmsipline", "iann", 0, "12345")
    acfg.sipConfig.authCreds.append(creds)
    
    acc = Account()
    acc.create(acfg)

    return ep

if __name__ == "__main__":

   ep = initalise_sip_stack()

   time.sleep(100)

   ep.libDestroy()

### Steps to reproduce

Copy and paste the example python script in to main.py and run it with a valid account on a Cisco CUCM 12.5. After the sip registers the CUCM sends an up solicited NOTIFY to update VM MWI (I think). The code count seem to be able to fins the account in an internal list to map the Notify to one of the 3 accounts

 pjsua_acc.c  ..No available account to handle Request msg NOTIFY/cseq=101 (rdata0x7f76a4009158)
python3: ../src/pjsua-lib/pjsua_acc.c:623: pjsua_acc_get_user_data: Assertion `acc_id>=0 && acc_id<(int)(sizeof(pjsua_var.acc)/sizeof(pjsua_var.acc[0]))' failed.
Aborted (core dumped)

### PJSIP version

2.12 dev

### Context

Ununtu linux 

### Log, call stack, etc

```shell
--end msg--
[...]

Issue #982, opened 2009-10-26
Support for SIP Message Summary/Message Waiting Indication (MWI, RFC 3842)


pjsipbot:
_2009-10-26 11:12:13: @bennylp_ _created the issue on trac ticket [982](https://trac.pjsip.org/repos/ticket/982)_

Add support for Message Summary/Message Waiting Indication (MWI, RFC 3842). Two types of MWI events are supported: MWI with subscription and unsolicited MWI (unsolicited means NOTIFY is sent without subscription).

Specification:
 - implement pjsip evsub package for MWI
 - client and server support
 - unsolicited MWI support 
 - support in PJSUA-LIB
 - NOTIFY message body is unprocessed; they will be given to app unparsed.

New files in pjsip-simple:
 - mwi.h
 - mwi.c

Support in PJSUA-LIB:
 - "mwi_enabled" field in pjsua account config (default is disabled)
 - "on_mwi_info" callback
 - "enable_unsolicited_mwi" in pjsua_config (default is enabled)
 - if MWI subscription is terminated for some reason, PJSUA-LIB will renew periodically (PJSUA_PRES_TIMER timer)

PJSUA application:
 - new "--mwi" option
 - print MWI info on the callback

To implement MWI support in your application:
 1. implement `on_mwi_info()` callback
 1. at this point, any unsolicited incoming NOTIFY requests containing MWI will cause `on_mwi_info()` callback to be called, unless `enable_unsolicited_mwi` is explicitly disabled in `pjsua_config`.
 1. if you'd like to use MWI subscriptions, set `mwi_enabled` to TRUE in `pjsua_acc_config`

___
_2009-10-26 11:22:03: @bennylp_ _edited the issue description_

___
_2009-10-26 11:22:39: @bennylp_ _changed **status** from **new** to **closed**_

___
_2009-10-26 11:22:39: @bennylp_ _set **resolution** to **fixed**_

___
_2009-10-26 11:22:39: @bennylp_ _commented_

Implemented in [r2968](../commit/4dd961b2fb6afe7a6a90deca8fd2e9adb65efaa0):
 - PJSIP-SIMPLE:
    - implement MWI
 - PJSUA-LIB:
    - added "mwi_enabled" flag in account config
    - added "on_mwi_info" callback
 - pjsua app:
    - added "--mwi" option to enable MWI on account
    - added simple callback to log the NOTIFY message
 - other:
    - added SIPp scenario files to simulate UAS side
 - build:
[...]

PR #3411, opened 2023-02-28
Handle incoming MWI when there is no available account

trengginas:
This is related to #3164.

When an incoming MWI is received and no account is available, the `on_mwi_info()` callback will still be called and 
can lead to assertion on PJSUA2.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_pres.c
+++ b/pjsip/src/pjsua-lib/pjsua_pres.c
@@ -2296,6 +2296,7 @@ static pj_bool_t unsolicited_mwi_on_rx_request(pjsip_rx_data *rdata)
     pj_str_t EVENT_HDR  = { "Event", 5 };
     pj_str_t MWI = { "message-summary", 15 };
     pjsip_event_hdr *eh;
+    pjsua_acc_id acc_id;
 
     if (pjsip_method_cmp(&msg->line.req.method, pjsip_get_notify_method())!=0) 
     {
@@ -2314,6 +2315,21 @@ static pj_bool_t unsolicited_mwi_on_rx_request(pjsip_rx_data *rdata)
         return PJ_FALSE;
     }
 
+    /* Find which account for the incoming request. */
+    acc_id = pjsua_acc_find_for_incoming(rdata);
+    if (acc_id == PJSUA_INVALID_ID) {
+        PJ_LOG(2, (THIS_FILE,
+                   "Unable to process incoming message %s "
+                   "due to no available account",
+                   pjsip_rx_data_get_info(rdata)));
+
+        pjsip_endpt_respond_stateless(pjsua_var.endpt, rdata,
+                                      PJSIP_SC_TEMPORARILY_UNAVAILABLE, NULL,
+                                      NULL, NULL);
+
+        return PJ_TRUE;
+    }
+
     PJ_LOG(4,(THIS_FILE, "Got unsolicited NOTIFY from %s:%d..",
               rdata->pkt_info.src_name, rdata->pkt_info.src_port));
     pj_log_push_indent();
@@ -2325,7 +2341,6 @@ static pj_bool_t unsolicited_mwi_on_rx_request(pjsip_rx_data *rdata)
 
     /* Call callback */
     if (pjsua_var.ua_cfg.cb.on_mwi_info) {
-        pjsua_acc_id acc_id;
         pjsua_mwi_info mwi_info;
 
         acc_id = pjsua_acc_find_for_incoming(rdata);


pjsip/src/pjsua-lib/pjsua_pres.c
```diff
@@ -2314,6 +2315,21 @@ static pj_bool_t unsolicited_mwi_on_rx_request(pjsip_rx_data *rdata)
...
 
+    /* Find which account for the incoming request. */
+    acc_id = pjsua_acc_find_for_incoming(rdata);
+    if (acc_id == PJSUA_INVALID_ID) {
+        PJ_LOG(2, (THIS_FILE,
+                   "Unable to process incoming message %s "
+                   "due to no available account",
+                   pjsip_rx_data_get_info(rdata)));
+
+        pjsip_endpt_respond_stateless(pjsua_var.endpt, rdata,
```
sauwming, line 2326: temporarily unavailable seems too vague and suggests a temporary error.
Perhaps 481 or 404?

trengginas: > temporarily unavailable seems too vague and suggests a temporary error. Perhaps 481 or 404?

The same error (PJSIP_SC_TEMPORARILY_UNAVAILABLE) was returned on https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_pres.c#L839. We should change that as well

sauwming: The 481 response is `481 Subscription does not exist`, so it cannot be used for https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsua-lib/pjsua_pres.c#L839 .

While here's the recommendation of the RFC https://www.rfc-editor.org/rfc/rfc3265#section-3.2:
```
If any non-SUBSCRIBE mechanisms are defined to create subscriptions,
   it is the responsibility of the parties defining those mechanisms to
   ensure that correlation of a NOTIFY message to the corresponding
   subscription is possible.  Designers of such mechanisms are also
   warned to make a distinction between sending a NOTIFY message to a
   subscriber who is aware of the subscription, and sending a NOTIFY
   message to an unsuspecting node.  The latter behavior is invalid, and
   MUST receive a "481 Subscription does not exist" response (unless
   some other 400- or 500-class error code is more applicable), as
   described in [section 3.2.4](https://www.rfc-editor.org/rfc/rfc3265#section-3.2.4).  In other words, knowledge of a
   subscription must exist in both the subscriber and the notifier to be
   valid, even if installed via a non-SUBSCRIBE mechanism.
```

sauwming: We need the status text "Subscription does not exist", otherwise the default will be "Call/Transaction Does Not Exist" which is not quite appropriate here.

sauwming (approved):

pjsip/src/pjsua-lib/pjsua_pres.c
nanangizz: As this is a part of the feature called "unsolicited MWI", IMO it is rather ambiguous to respond with "Subscription does not exist" because that's exactly the purpose at the first place. Moreover 4xx is more for UAC-failure, I guess we better use UAS-failure/5xx response code as the error is on UAS, for example `PJSIP_SC_INTERNAL_SERVER_ERROR` or `PJSIP_SC_SERVICE_UNAVAILABLE`?

sauwming: Fair point, I considered that too. That's why I included 404 as alternative and quoted the entire paragraph of the RFC so we can get a more complete understanding.

According to my interpretation, the unsolicited MWI falls under "the non-SUBSCRIBE mechanisms defined to create subscriptions". The fact that there's no SUBSCRIBE message exchanged doesn't mean that there's no subscription, as highlighted again in the closing of the paragraph `subscription must exist in both the subscriber and the notifier to be valid, even if installed via a non-SUBSCRIBE mechanism.`

So here IMO, we rely on the existence of the account for a subscription to exist and be valid.

And for the error code, even though it recommends 481, it doesn't strictly prohibit any other code, so I'm okay with that. But I don't agree that this is a UAS failure since it's also likely that the UAC sends the NOTIFY to the wrong address/account.

nanangizz: > According to my interpretation, the unsolicited MWI falls under "the non-SUBSCRIBE mechanisms defined to create subscriptions". The fact that there's no SUBSCRIBE message exchanged doesn't mean that there's no subscription, as highlighted again in the closing of the paragraph `subscription must exist in both the subscriber and the notifier to be valid, even if installed via a non-SUBSCRIBE mechanism.`

This "unsolicited" is really no subscription (not just no SUBSCRIBE, yes I am aware of this too, call transfer is an example). Please check #982 where this features is implemented, it mentions that "unsolicited means NOTIFY is sent without subscription". Also as shown in the code, `mwi_info.evsub` for the callback is set to `NULL` because it does not use subscription (yes, violating the standard).

sauwming: Well, I guess we disagree on this one. To me, there is a subscriber, there is a notifier, and there is a subscription, albeit implicit.

But since the theoretical concept seems to be irrelevant here, I think we'll just move on.

The point is what error code we should send to the remote. I believe the error message should reflect exactly what happened, which is that the "Account does not exist/not found", and not that there's some random internal error. So any error response that conveys that message should be fine. Some alternatives:
- 481 with status "Account does not exist"
- 404 "Account not found"
- or perhaps 604 "Does Not Exist Anywhere"
``` 
   The server has authoritative information that the user indicated in
   the Request-URI does not exist anywhere.
```

nanangizz: Changing the message to represent the actual problem (no account to handle) sounds good to me.

nanangizz (approved):

PR #3412, opened 2023-02-28
Return error when calling pjsua_call_get_stream_info()/pjsua_call_get_stream_stat() on an inactive stream

### LINKED ISSUE #2756: PJSUA was crash on `pjmedia_vid_stream_get_info`
### Describe the bug

During direct call from A to B without video (i.e. SDP contain "m=video 0") PJSUA crashes with:

Traceback since invoke `pjsua_call_get_stream_info`

```
#1  0x00007f4e54ff2535 in __GI_abort () at abort.c:79
#2  0x00007f4e54ff240f in __assert_fail_base (fmt=0x7f4e55154ee0 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x7f4e544be19a "stream && info", file=0x7f4e544c15c0 "../src/pjmedia/vid_stream.c", line=2258,
    function=<optimized out>) at assert.c:92
#3  0x00007f4e55000102 in __GI___assert_fail (assertion=0x7f4e544be19a "stream && info", file=0x7f4e544c15c0 "../src/pjmedia/vid_stream.c", line=2258, function=0x7f4e544c19f0 "pjmedia_vid_stream_get_info") at assert.c:101
#4  0x00007f4e544a320c in pjmedia_vid_stream_get_info () from /lib/x86_64-linux-gnu/libpjmedia.so.2
#5  0x00007f4e546ce69b in pjsua_call_get_stream_info () from /lib/x86_64-linux-gnu/libpjsua.so.2
```

### Steps to reproduce

1. Create PJSUA device B (register)
2. Receive incoming call with video SDP on B (SDP - "m=video 0 31" )
3. Answer SDP without video 
4. Get call stream information (pjsua_call_get_stream_info)
5. Abort on assert on "stream & info" in vid_stream.c

### PJSIP version

2.11

### Context

- Linux
- PJSUA_VIDEO = 1

### Log, call stack, etc

```shell
N/A
```

trengginas:
To fix #2756.

When the stream is inactive, assertion will be raised in pjmedia_stream_get_info()/pjmedia_vid_stream_get_info() since the stream 
is set to NULL.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_aud.c
+++ b/pjsip/src/pjsua-lib/pjsua_aud.c
@@ -155,9 +155,8 @@ PJ_DEF(pj_status_t) pjsua_call_get_stream_info( pjsua_call_id call_id,
 
     call_med = &call->media[med_idx];
 
-    if (call_med->state == PJSUA_CALL_MEDIA_NONE ||
-        call_med->state == PJSUA_CALL_MEDIA_ERROR ||
-        call_med->dir == PJMEDIA_DIR_NONE)
+    if ((call_med->type == PJMEDIA_TYPE_AUDIO && !call_med->strm.a.stream) ||
+        (call_med->type == PJMEDIA_TYPE_VIDEO && !call_med->strm.v.stream))
     {
         PJSUA_UNLOCK();
         return PJ_EINVAL;
@@ -211,9 +210,8 @@ PJ_DEF(pj_status_t) pjsua_call_get_stream_stat( pjsua_call_id call_id,
 
     call_med = &call->media[med_idx];
 
-    if (call_med->state == PJSUA_CALL_MEDIA_NONE ||
-        call_med->state == PJSUA_CALL_MEDIA_ERROR ||
-        call_med->dir == PJMEDIA_DIR_NONE)
+    if ((call_med->type == PJMEDIA_TYPE_AUDIO && !call_med->strm.a.stream) ||
+        (call_med->type == PJMEDIA_TYPE_VIDEO && !call_med->strm.v.stream))
     {
         PJSUA_UNLOCK();
         return PJ_EINVAL;
--- a/pjsip/src/pjsua-lib/pjsua_aud.c
+++ b/pjsip/src/pjsua-lib/pjsua_aud.c
@@ -138,7 +138,7 @@ PJ_DEF(pj_status_t) pjsua_call_get_stream_info( pjsua_call_id call_id,
 {
     pjsua_call *call;
     pjsua_call_media *call_med;
-    pj_status_t status;
+    pj_status_t status = PJ_EINVAL;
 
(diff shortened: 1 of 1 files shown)


sauwming:
Okay, this is an interesting case. At first, I thought that it is the application's responsibility to make sure that media index is valid, and that the assertion is already correct.

But since we return normal error of PJ_EINVAL `if (med_idx >= call->med_cnt)`, then the API behaviour is rather inconsistent. So, let's say there are 1 active audio, and 1 inactive video, then the API will give the correct info for med_idx 0, return PJ_EINVAL for med_idx > 1, but assert if med_idx == 1. That's rather strange, IMO.

So the patch looks good to me. I wish there is a better return value than PJ_EINVAL, but I can't find a suitable one.

Note: We need to apply the same fix to `pjsua_call_get_stream_stat()` as well.

pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -154,6 +154,15 @@ PJ_DEF(pj_status_t) pjsua_call_get_stream_info( pjsua_call_id call_id,
     }
 
     call_med = &call->media[med_idx];
+
+    if (call_med->state == PJSUA_CALL_MEDIA_NONE ||
+        call_med->state == PJSUA_CALL_MEDIA_ERROR ||
+        call_med->dir == PJMEDIA_DIR_NONE)
```
sauwming, line 160: For double hold scenario, the media dir will be NONE, so we should remove this condition.

Alternatively, if you want to be safe, just check directly if the stream is not NULL, i.e.:
```
if ((call_med->type == PJMEDIA_TYPE_AUDIO && !call_med->strm.a.stream) ||
     (call_med->type == PJMEDIA_TYPE_VIDEO && !call_med->strm.v.stream))
```

sauwming (approved):

pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -155,9 +155,8 @@ PJ_DEF(pj_status_t) pjsua_call_get_stream_info( pjsua_call_id call_id,
 
     call_med = &call->media[med_idx];
 
-    if (call_med->state == PJSUA_CALL_MEDIA_NONE ||
-        call_med->state == PJSUA_CALL_MEDIA_ERROR ||
-        call_med->dir == PJMEDIA_DIR_NONE)
+    if ((call_med->type == PJMEDIA_TYPE_AUDIO && !call_med->strm.a.stream) ||
+        (call_med->type == PJMEDIA_TYPE_VIDEO && !call_med->strm.v.stream))
```
wosrediinanatour, line 159: Just minor command for this and the updated if-statement below:
Since there are already variables `status`, you could set the status and make a "goto on_return", i.e. a return section that takes care of `PJSUA_UNLOCK();`.

nanangizz (approved):

PR #3417, opened 2023-03-02
Add function for printing SDP attributes

wosrediinanatour:
It adds a function `pjmedia_sdp_attr_print()` similar to `pjmedia_sdp_media_print()`.

### DIFF
--- a/pjmedia/include/pjmedia/sdp.h
+++ b/pjmedia/include/pjmedia/sdp.h
@@ -491,7 +491,7 @@ typedef struct pjmedia_sdp_media pjmedia_sdp_media;
  * @param buf       The buffer.
  * @param size      The buffer length.
  *
- * @return          the length printed, or -1 if the buffer is too
+ * @return          The length printed, or -1 if the buffer is too
  *                  short.
  */
 PJ_DECL(int) pjmedia_sdp_media_print(const pjmedia_sdp_media *media, char *buf, pj_size_t size);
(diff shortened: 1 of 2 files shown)


pjmedia/include/pjmedia/sdp.h
```diff
@@ -508,6 +508,18 @@ PJ_DECL(pjmedia_sdp_media*)
 pjmedia_sdp_media_clone( pj_pool_t *pool, 
                          const pjmedia_sdp_media *rhs);
 
+/**
+ * Print attribute to a buffer.
+ *
+ * @param media     The attribute.
```
nanangizz, line 514: should be `attr` instead of `media`?

pjmedia/include/pjmedia/sdp.h
```diff
@@ -489,12 +489,12 @@ typedef struct pjmedia_sdp_media pjmedia_sdp_media;
  *
  * @param media     The media description.
  * @param buf       The buffer.
- * @param size      The buffer length.
+ * @param legth     The buffer length.
```
nanangizz, line 492: typo: `legth`

nanangizz (commented):
Thanks for the PR.
Could you also describe a sample scenario that this function is needed by app?

pjmedia/include/pjmedia/sdp.h
sauwming: I'm not sure why you need to change the parameter name here.

IMO, it's better to leave existing API's parameter names untouched, unless there's a reason for the change.

wosrediinanatour:
> Could you also describe a sample scenario that this function is needed by app?

There are already exposed functions for printing SDP (`pjmedia_sdp_print()`) and the SDP's media description (`pjmedia_sdp_media_print()`). For C++ wrappers that wrap the types `pjmedia_sdp_session`, `pjmedia_sdp_media`, and `pjmedia_sdp_attr`, it is handy to have consistent functionality to print the content (even if it is for logging "only") into a string (buffer).

sauwming (approved):

trengginas (approved):

nanangizz (approved):

PR #3424, opened 2023-03-07
Add socket option params to transport config in PJSUA2

nanangizz:
- Add type `SockOptParams` to PJSUA2 which wraps `pj_sockopt_params`.
- Add field `sockOptParams` to PJSUA2 `TransportConfig`.
- Update PJSUA2 JSON `WriteString()` to write as long as C++ string length, otherwise it may not write full string if there is any null character in the middle.

### DIFF
--- a/pjsip/include/pjsua2/siptypes.hpp
+++ b/pjsip/include/pjsua2/siptypes.hpp
@@ -308,6 +308,72 @@ public:
 
 
 /**
+ * Socket option parameters, to be specified in TransportConfig.
+ */
+struct SockOptParams : public PersistentObject
+{
+    /**
+     * Socket option type.
+     */
+    struct SockOpt {
+        /**
+         * The level at which the option is defined.
+         */
+        int                 level;
+
+        /**
+         * Option name.
+         */
+        int                 optName;
+
+        /**
+         * Pointer to the buffer in which the option is specified.
+         */
+        void               *optVal;
+
+        /**
+         * Buffer size of the buffer pointed by optVal.
+         */
+        int                 optLen;
+
+        /**
+         * Internal buffer for optval.
+         */
+        string optValBuf_;
+    };
+
+    /**
+     * Array of socket options.
+     */
+    std::vector<SockOpt>    sockOpts;
+
+public:
+    /** Default constructor initialises with default values */
+    SockOptParams();
+
+    /** Convert to pjsip */
+    pj_sockopt_params toPj() const;
+
+    /** Convert from pjsip */
+    void fromPj(const pj_sockopt_params &prm);
+
+    /**
+     * Read this object from a container node.
+     *
+     * @param node              Container to read values from.
+     */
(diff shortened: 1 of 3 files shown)


pjsip/include/pjsua2/siptypes.hpp
```diff
@@ -307,6 +307,72 @@ struct TlsConfig : public PersistentObject
 };
 
 
+/**
+ * Socket option parameters, to be specified in TransportConfig.
+ */
+struct SockOptParams : public PersistentObject
+{
+    /**
```
sauwming, line 315: if I remember correctly, this kind of nested structure needs to be flattened. I don't remember the exact reason, but maybe it's SWIG?

Also, just curious, can this work in Java?

nanangizz: Haven't tested with SWIG, will do it later.

pjsip/include/pjsua2/siptypes.hpp
```diff
@@ -307,6 +307,72 @@ struct TlsConfig : public PersistentObject
...
+
+        /**
+         * Pointer to the buffer in which the option is specified.
+         */
+        void               *optVal;
+
+        /**
+         * Buffer size of the buffer pointed by optVal.
+         */
+        int                 optLen;
```
wosrediinanatour, line 337: Why don't you use `size_t`?

pjsip/include/pjsua2/siptypes.hpp
```diff
@@ -307,6 +307,72 @@ struct TlsConfig : public PersistentObject
...
+    struct SockOpt {
+        /**
+         * The level at which the option is defined.
+         */
+        int                 level;
+
+        /**
+         * Option name.
+         */
+        int                 optName;
```
wosrediinanatour, line 327: How can a name be `int`?

pjsip/include/pjsua2/siptypes.hpp
```diff
@@ -307,6 +307,72 @@ struct TlsConfig : public PersistentObject
...
+struct SockOptParams : public PersistentObject
+{
+    /**
+     * Socket option type.
+     */
+    struct SockOpt {
+        /**
+         * The level at which the option is defined.
+         */
+        int                 level;
```
wosrediinanatour, line 322: What about using `size_t` or `unsigned int`?
Even much better: have an enum (class).

pjsip/include/pjsua2/siptypes.hpp
```diff
@@ -307,6 +307,72 @@ struct TlsConfig : public PersistentObject
...
+
+        /**
+         * Buffer size of the buffer pointed by optVal.
+         */
+        int                 optLen;
+
+        /**
+         * Internal buffer for optval.
+         */
+        string optValBuf_;
```
wosrediinanatour, line 342: `std::string` is consistent with other usage of `std`.

sauwming:
Feel free to reopen if you wish to work on this again.

Issue #3425, opened 2023-03-07
To fix gcc/clang CFLAGS -Wextra warning reports


jimying:
There are some warning reports use -Wextra in CFLAGS,  maybe find time to fix them.

PS:
It's better use new gcc/clang version. I test it use gcc12 on linux.

```bash
# How to dump warnings into file /tmp/warn.log  (suppress the  warning "missing-field-initializers")
export CFLAGS="-Wextra -Wno-missing-field-initializers"
./configure .....
make dep
# Better to use one job for better log order
make  -j  1   2>/tmp/warn.log
```

Update: 
Now i find it's better one module a log file (ignore third_party):
```bash
export JOBS=2
export MAKE=make
export CFLAGS="-Wextra -Wno-missing-field-initializers"
./configure ....
make dep 

$MAKE -C pjlib/build -j  $JOBS 2>/tmp/warn-pjlib.log
$MAKE -C pjlib-util/build -j  $JOBS 2>/tmp/warn-pjlibutil.log
$MAKE -C pjnath/build -j  $JOBS 2>/tmp/warn-pjnath.log
$MAKE -C third_party/build -j  $JOBS
$MAKE -C pjmedia/build -j  $JOBS 2>/tmp/warn-pjmedia.log
$MAKE -C pjsip/build -j $JOBS 2>/tmp/warn-pjsip.log
$MAKE -C pjsip-apps/build -j $JOBS 2>/tmp/warn-apps.log
```

PR #3427, opened 2023-03-08
Fix silencedet: code statement fall through (in switch-case)

jimying:
"case STATE_SILENCE"  no break

The logic seem to be error.  
 Maybe i am wrong, please check it.

PS:
#3425 help find this bug


jimying:
The logic is no problem

PR #3428, opened 2023-03-08
Fix build with CFLAGS -Wextra warning reports

jimying:
as title 
Please see #3425 for detail

This include all changes
 (
  also include PR #3426's changes, please merge it first, which is a bug.
  I think it better  use a new sperate PR
)

### DIFF
--- a/pjlib-util/include/pjlib-util/sha1.h
+++ b/pjlib-util/include/pjlib-util/sha1.h
@@ -58,7 +58,7 @@ PJ_DECL(void) pj_sha1_init(pj_sha1_context *ctx);
  */
 PJ_DECL(void) pj_sha1_update(pj_sha1_context *ctx, 
                              const pj_uint8_t *data, 
-                             const pj_size_t nbytes);
+                             pj_size_t nbytes);
 
 /** Finish the message and return the digest. 
  *  @param ctx          SHA1 context.
--- a/pjlib/src/pj/os_core_darwin.m
+++ b/pjlib/src/pj/os_core_darwin.m
@@ -23,6 +23,7 @@
 PJ_DEF(int) pj_run_app(pj_main_func_ptr main_func, int argc, char *argv[],
                        unsigned flags)
 {
+    PJ_UNUSED_ARG(flags);
     return (*main_func)(argc, argv);
 }
 
@@ -76,6 +77,7 @@ static void* main_thread(void *data)
 PJ_DEF(int) pj_run_app(pj_main_func_ptr main_func, int argc, char *argv[],
                        unsigned flags)
 {
+    PJ_UNUSED_ARG(flags);
     pthread_t thread;
     run_app_t param;
     NSAutoreleasePool *pool;
--- a/pjlib/src/pj/os_core_darwin.m
+++ b/pjlib/src/pj/os_core_darwin.m
@@ -77,11 +77,11 @@ static void* main_thread(void *data)
 PJ_DEF(int) pj_run_app(pj_main_func_ptr main_func, int argc, char *argv[],
                        unsigned flags)
 {
-    PJ_UNUSED_ARG(flags);
     pthread_t thread;
     run_app_t param;
     NSAutoreleasePool *pool;
-    
+    PJ_UNUSED_ARG(flags);
+
     pool = [[NSAutoreleasePool alloc] init];
     [NSApplication sharedApplication];
     [DeadThread enterMultiThreadedMode];
--- a/pjmedia/src/pjmedia/converter_libyuv.c
+++ b/pjmedia/src/pjmedia/converter_libyuv.c
@@ -76,7 +76,7 @@ typedef enum conv_func_type
     SCALE_PLANAR
 } conv_func_type;
 
-typedef void* gen_conv_func;
+typedef int (*gen_conv_func)();
 
 typedef int (*conv_pack_to_pack_method)(const uint8* src, int src_stride,
                                         uint8* dst, int dst_stride,
--- a/pjmedia/src/pjmedia/converter_libyuv.c
+++ b/pjmedia/src/pjmedia/converter_libyuv.c
@@ -76,7 +76,7 @@ typedef enum conv_func_type
     SCALE_PLANAR
 } conv_func_type;
 
-typedef void (*gen_conv_func)();
+typedef void* gen_conv_func;
 
 typedef int (*conv_pack_to_pack_method)(const uint8* src, int src_stride,
                                         uint8* dst, int dst_stride,
--- a/pjsip/src/pjsip/sip_parser.c
+++ b/pjsip/src/pjsip/sip_parser.c
@@ -1905,8 +1905,10 @@ static void int_parse_contact_param( pjsip_contact_hdr *hdr,
                 hdr->expires--;
             if (hdr->expires > PJSIP_MAX_EXPIRES)
                 hdr->expires = PJSIP_MAX_EXPIRES;
-            // if (hdr->expires < PJSIP_MIN_EXPIRES)
-            //     hdr->expires = PJSIP_MIN_EXPIRES;
+#if PJSIP_MIN_EXPIRES > 0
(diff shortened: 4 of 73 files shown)


pjlib-util/include/pjlib-util/sha1.h
```diff
@@ -58,7 +58,7 @@ PJ_DECL(void) pj_sha1_init(pj_sha1_context *ctx);
  */
 PJ_DECL(void) pj_sha1_update(pj_sha1_context *ctx, 
                              const pj_uint8_t *data, 
-                             const pj_size_t nbytes);
+                             pj_size_t nbytes);
```
sauwming, line 61: Thank you for the patch. The rest seems okay. I do not understand this, though. Why is this necessary? `nbytes/len` doesn't seem to be overwritten within the function, and if it is, then I'd rather change the implementation and leave the API declaration untouched.

jimying: Passing an unsigned int/long value does not require const(^_^) .  pj_md5_update() is simillar, no const

sauwming: Here's what I found from Googling:
Declaring a function with const int parameter simply indicates that the value of the integer parameter will not be modified inside the function. It is a common practice in C to declare function parameters as const if the function does not intend to modify the parameter value.

So while it's true that it's not strictly necessary, I don't think it's worth changing as well, especially considering it is an API. Besides, when I tried to compile it here with both `gcc` and `clang` with `-Wextra`, I didn't receive any warning, so I am curious as to what warning did you get there?

jimying: I think const mean not modify original memory value, here is just pass a int value， not able to afect original value.

Better to keep same with other xx_update() (md5,hmac ...)

jimying:
Current change is  use epoll, no video, no ssl/srtp, I want to do some checks.
So marked it as draft.

jimying:
More checks when enable video,ssl. Task completed,  please review it.

pjlib/src/pj/os_core_darwin.m
```diff
@@ -76,6 +77,7 @@ + (void)emptyThreadMethod:(id)obj { ; }
 PJ_DEF(int) pj_run_app(pj_main_func_ptr main_func, int argc, char *argv[],
                        unsigned flags)
 {
+    PJ_UNUSED_ARG(flags);
```
sauwming, line 80: PJ_UNUSED_ARG() is typically after the variable declarations.

pjmedia/src/pjmedia/converter_libyuv.c
```diff
@@ -76,7 +76,7 @@ typedef enum conv_func_type
     SCALE_PLANAR
 } conv_func_type;
 
-typedef void (*gen_conv_func)();
+typedef void* gen_conv_func;
```
sauwming, line 79: this doesn't seem right.
void (*gen_conv_func)(), is a function pointer declaration, which specifies that gen_conv_func is a pointer to a function that takes an unspecified number of arguments and returns nothing (i.e., void). The syntax (*gen_conv_func) is used to dereference the pointer and access the function itself.

void* gen_conv_func, is a declaration of a void pointer, which specifies that gen_conv_func is a pointer to an unspecified type of data. Void pointers can point to any type of data, but they do not have any associated type information, so you cannot directly dereference them.

jimying: I use gcc12 on linux test, so many warnings as follow:
```
../src/pjmedia/converter_libyuv.c:163:48: warning: cast between incompatible function types from ‘int (*)(const uint8 *, int,  const uint8 *, int,  const uint8 *, int,  uint8 *, int,  int,  int)’ {aka ‘int (*)(const unsigned char *, int,  const unsigned char *, int,  const unsigned char *, int,  unsigned char *, int,  int,  int)’} to ‘void (*)()’ [-Wcast-function-type]
  163 |         GET_PJ_FORMAT(dst),CONV_PLANAR_TO_PACK,(gen_conv_func)&method
      |                                                ^   
../src/pjmedia/converter_libyuv.c:187:6: note: in expansion of macro ‘MAP_CONV_PLANAR_TO_PACK’
  187 |     {MAP_CONV_PLANAR_TO_PACK(I420,UYVY,I420ToUYVY)},
      |      ^~~~~~~~~~~~~~~~~~~~~~~
../src/pjmedia/converter_libyuv.c:165:50: warning: cast between incompatible function types from ‘int (*)(const uint8 *, int,  const uint8 *, int,  const uint8 *, int,  uint8 *, int,  uint8 *, int,  uint8 *, int,  int,  int)’ {aka ‘int (*)(const unsigned char *, int,  const unsigned char *, int,  const unsigned char *, int,  unsigned char *, int,  unsigned char *, int,  unsigned char *, int,  int,  int)’} to ‘void (*)()’ [-Wcast-function-type]
  165 |         GET_PJ_FORMAT(dst),CONV_PLANAR_TO_PLANAR,(gen_conv_func)&method
      |                                                  ^   
../src/pjmedia/converter_libyuv.c:188:6: note: in expansion of macro ‘MAP_CONV_PLANAR_TO_PLANAR’
  188 |     {MAP_CONV_PLANAR_TO_PLANAR(I420,I422,I420ToI422)},
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~
../src/pjmedia/converter_libyuv.c:169:41: warning: cast between incompatible function types from ‘int (*)(const uint8 *, int,  const uint8 *, int,  const uint8 *, int,  int,  int,  uint8 *, int,  uint8 *, int,  uint8 *, int,  int,  int,  enum FilterMode)’ {aka ‘int (*)(const unsigned char *, int,  const unsigned char *, int,  const unsigned char *, int,  int,  int,  unsigned char *, int,  unsigned char *, int,  unsigned char *, int,  int,  int,  enum FilterMode)’} to ‘void (*)()’ [-Wcast-function-type]
  169 |         GET_PJ_FORMAT(fmt),SCALE_PLANAR,(gen_conv_func)&method

```

after change, warnings disappear

But i use clang17 check it,  both not change and change  no the warnings

jimying: @sauwming You'd better install lastest gcc12 or clang16/17 on linux to test it

jimying: @sauwming  clang  CFLAGS   add "-Wcast-function-type" can find out the warning

```bash
# This is the sugguest flags that compatible with both gcc and clang
export CFLAGS= \
"-Wextra -Wno-missing-field-initializers -Wno-unused-but-set-variable 
-Wsingle-bit-bitfield-constant-conversion -Wcast-function-type"
```

sauwming: you can simply change it to `typedef int (*gen_conv_func)();` to get rid of the warning

pjsip/src/pjsip/sip_parser.c
```diff
@@ -1905,8 +1905,8 @@ static void int_parse_contact_param( pjsip_contact_hdr *hdr,
                 hdr->expires--;
             if (hdr->expires > PJSIP_MAX_EXPIRES)
                 hdr->expires = PJSIP_MAX_EXPIRES;
-            if (hdr->expires < PJSIP_MIN_EXPIRES)
-                hdr->expires = PJSIP_MIN_EXPIRES;
+            // if (hdr->expires < PJSIP_MIN_EXPIRES)
+            //     hdr->expires = PJSIP_MIN_EXPIRES;
```
nanangizz, line 1909: The check may be needed when `PJSIP_MIN_EXPIRES` is not zero (not sure if it is possible though)? So what if instead of commenting it out, change it to something like
```
#if PJSIP_MIN_EXPIRES != 0
   if (hdr->expires < PJSIP_MIN_EXPIRES)
       hdr->expires = PJSIP_MIN_EXPIRES;
#endif
```

nanangizz (approved):

sauwming (approved):

PR #3439, opened 2023-03-14
Use PJ_ARRAY_SIZE() replace sizeof()/sizeof()

jimying:
as title


sauwming (approved):
Thanks for the PR. The patch itself looks okay, but just FYI, we typically avoid making changes unless it triggers a build warning. The reason is to make git blame easier to read/track. Even changing tabs to spaces last time, we have to consider for a while.

nanangizz (approved):

PR #3442, opened 2023-03-15
Check if interface address returned by pj_getipinterface() is link-local address

sauwming:
It's been reported that for IPv6, `pj_getipinterface()` can return success and give link-local address even though the device has no routable IPv6 address. This will cause the address returned not to be of much use since it can't be used to send/receive data to the external host.

`pj_getipinterface()` relies on `connect()` and `getsockname()` to get the interface address, i.e.:
```
    status = pj_sock_socket(af, pj_SOCK_DGRAM(), 0, &fd);
    status = pj_sock_connect(fd, &dst_addr, pj_sockaddr_get_len(&dst_addr));
    status = pj_sock_getsockname(fd, itf_addr, &len);
```
 
As it turns out, since the socket created is a connection-less socket, yes, it is possible `connect()` to return 0/success for an IPv6 datagram socket even if there is no routable IPv6 address available on any of the local network interfaces. The failure will happen later when you attempt to send or receive data using the socket.

When you use `connect()` on an IPv6 datagram socket, the operating system selects a source IP address to use for the connection, based on the routing table and the local network interfaces. If there is no routable IPv6 address available on any of the local network interfaces, the operating system may still be able to select a source IP address for the connection, such as a link-local address or a site-local address.

However, if the selected source address is not routable to the destination, any subsequent attempt to send or receive data using the socket will fail. This is because the IPv6 routing infrastructure will not be able to deliver the packets to the destination.

Therefore, while it's technically possible for connect() to return 0 for an IPv6 datagram socket even if there is no routable IPv6 address available, it's generally not useful or practical to establish a connection in this case.

### DIFF
--- a/pjlib/src/pj/sock_common.c
+++ b/pjlib/src/pj/sock_common.c
@@ -1069,11 +1069,16 @@ PJ_DEF(pj_status_t) pj_getipinterface(int af,
     }
 
 #if defined(PJ_HAS_IPV6) && PJ_HAS_IPV6
-    /* Check that the address is not a link-local address */
+    /* Check if the local address is link-local but the destination
+     * is not link-local.
+     */
     if (af == pj_AF_INET6() &&
-        IN6_IS_ADDR_LINKLOCAL(&itf_addr->ipv6.sin6_addr))
+        IN6_IS_ADDR_LINKLOCAL(&itf_addr->ipv6.sin6_addr) &&
+        (!IN6_IS_ADDR_LINKLOCAL(&dst_addr.ipv6.sin6_addr)))
     {
-        TRACE_((THIS_FILE, "Interface address is a link-local address"));
+        TRACE_((THIS_FILE, "Local interface address is link-local and "
+                           "the destination host is external"));
+
         return PJ_ENOTFOUND;
     }
 #endif
--- a/pjlib/src/pj/sock_common.c
+++ b/pjlib/src/pj/sock_common.c
@@ -1068,6 +1068,16 @@ PJ_DEF(pj_status_t) pj_getipinterface(int af,
         return PJ_ENOTFOUND;
     }
 
+#if defined(PJ_HAS_IPV6) && PJ_HAS_IPV6
+    /* Check that the address is not a link-local address */
+    if (af == pj_AF_INET6() &&
+        IN6_IS_ADDR_LINKLOCAL(&itf_addr->ipv6.sin6_addr))
+    {
+        TRACE_((THIS_FILE, "Interface address is a link-local address"));
+        return PJ_ENOTFOUND;
+    }
+#endif
+
     if (p_dst_addr)
         *p_dst_addr = dst_addr;
 


nanangizz:
- If this may introduce behavioral change, perhaps this needs to be mentioned in the docs too?
- From a quick searching around, `pj_gethostip()` seems to expect link-local addrs too? Not sure if it is expected from `pj_getipinterface()` though (via `pj_getdefaultipinterface()`).
- Link-local may be useful for sending/receiving too, so what if the address is link-local, then the function performs a simple sending packet test (with non blocking socket or a configurable timeout, not sure if it can identify routing error)?

Btw, there seem to be link-local for IP4, e.g: 169.*?

sauwming:
> * If this may introduce behavioral change, perhaps this needs to be mentioned in the docs too?

The doc says: `Get the interface IP address to send data to the specified destination`, and the PR is attempting to fix the implementation to match the doc's specification (because currently it's possible to get an address that can't be used to send data), so I don't think the doc needs to be changed?

> * From a quick searching around, `pj_gethostip()` seems to expect link-local addrs too? Not sure if it is expected from `pj_getipinterface()` though (via `pj_getdefaultipinterface()`).

Good point. I guess not all link-local addresses should be rejected.

> * Link-local may be useful for sending/receiving too, so what if the address is link-local, then the function performs a simple sending packet test (with non blocking socket or a configurable timeout, not sure if it can identify routing error)?

Is it considered okay to send some rogue/random data to a host, just for testing?

Another alternative that was suggested is to check if the host is link-local. Something like this:
```
    hints.ai_family = AF_UNSPEC;
    hints.ai_socktype = SOCK_DGRAM;
    hints.ai_flags = AI_ADDRCONFIG;

    ret = getaddrinfo(dst, port, &hints, &res);
    // Check if the local address is link-local and the destination is not link-local
    if (IN6_IS_ADDR_LINKLOCAL(&itf_addr->ipv6.sin6_addr) && !IN6_IS_ADDR_LINKLOCAL((struct in6_addr *)res->ai_addr)) {
        return ...;
```
But it will need to call `getaddrinfo()`, so which approach do you think is better?

> 
> Btw, there seem to be link-local for IP4, e.g: 169.*?

Ah, right. I have removed it from the PR desc. I found that IPv4 and IPv6 addresses work rather differently though, so I'm not sure if this issue can also happen on IPv4 and I can't find any source to confirm this. But it shouldn't matter if we use either of the above approach.

sauwming:
Regarding the doc change, yet another alternative is to change the doc to: `Get the interface IP address to connect to the specified destination.` This way, the implementation does not need to change although it means that the responsibility falls to the application to check if the address can be used to send data or not.

nanangizz:
> Is it considered okay to send some rogue/random data to a host, just for testing?

Ah right, not okay I guess.


> Another alternative that was suggested is to check if the host is link-local. Something like this:
> 
> ```
>     hints.ai_family = AF_UNSPEC;
>     hints.ai_socktype = SOCK_DGRAM;
>     hints.ai_flags = AI_ADDRCONFIG;
> 
>     ret = getaddrinfo(dst, port, &hints, &res);
>     // Check if the local address is link-local and the destination is not link-local
>     if (IN6_IS_ADDR_LINKLOCAL(&itf_addr->ipv6.sin6_addr) && !IN6_IS_ADDR_LINKLOCAL((struct in6_addr *)res->ai_addr)) {
>         return ...;
> ```
> 
> But it will need to call `getaddrinfo()`, so which approach do you think is better?

This alternative (i.e: checking if the destination being a link-local address) sounds much better IMO, it is reasonable to return link-local only if the destination is also link-local. Also, this way the docs may not need to be updated.

Btw, the `getaddrinfo()` may be called only if the interface addr is a link-local?

nanangizz (approved):

sauwming:
After testing, it turns out if `allow_resolve` is true, pj_sockaddr_init() will internally call `pj_getaddrinfo()`, so I believe I just need to check whether the destination address is link local. So just updated the patch.

sauwming:
Just for additional note, if this doesn't work, we can revisit the data sending idea. Data sending can be feasible (and not intrusive to the host) if we use ICMP (Internet Control Message Protocol), which is used for `ping`.

Something like this:
```
int sockfd = socket(AF_INET, SOCK_RAW, IPPROTO_ICMP);
// send the ICMP packet to the destination address
int bytes_sent = sendto(sockfd, packet, PACKET_SIZE, 0, (struct sockaddr*)&dest_addr, sizeof(dest_addr));
if (bytes_sent < 0) {
        perror("sendto() error");
```

The problem is on some platforms (Apple), raw socket needs privilege access (`ping: Lacking privilege for raw socket`), but it seems to work fine on Linux.

nanangizz:
> The problem is on some platforms (Apple), raw socket needs privilege access (`ping: Lacking privilege for raw socket`), but it seems to work fine on Linux.

Another catch it that ping may be blocked/ignored by some paranoid firewalls (not sure what will be returned by sendto()) while the usual TCP/UDP traffics is being allowed, so ping failure may not always represent unreachability?

andreas-wehrmann:
> Another catch is that ping may be blocked/ignored by some paranoid firewalls (not sure what will be returned by sendto()) while the usual TCP/UDP traffics is being allowed, so ping failure may not always represent unreachability?

This is definitely the case. I myself have deployments behind firewalls where everything except "related" traffic is blocked.

trengginas (approved):

PR #3390, opened 2023-02-17
Support Coverity Scan

bennylp:
Work in progress.

See [tests/coverity-scan/README.md](https://github.com/pjsip/pjproject/blob/coverity01/tests/coverity-scan/README.md) for instructions

https://scan.coverity.com/projects/pjsip
Initial outstanding: 416, high impact: 67, fixed: 155
Host: fresh Ubuntu Docker (installed with `run.sh`)

PR #3443, opened 2023-03-15
Fix scan-build (Clang Static Analyzer) build fail

jimying:
I want to use scan-build to build & analysis source

```bash
export CC=clang
export CXX=clang++

./configure ....
make dep

# only build pjlib for test
scan-build make -C pjlib/build
```

which will be fail.

I found that the fail reason is that:
      scan-build auto use the clang tool ccc-analyzer replace cc/cxx to generate *.o  without "-c" option

Two methods to solve it:
1.  export CC_OUT="-c -o "
2.  this commit

I think the 2nd is better

PS:
scan-build find out some problems/bugs, need find time to fix them

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -4858,7 +4858,7 @@ if test "$LIBEXT" = ""; then LIBEXT='a'; fi
 
 if test "$LIBEXT2" = ""; then LIBEXT2=""; fi
 
-if test "$CC_OUT" = ""; then CC_OUT="-o "; fi
+if test "$CC_OUT" = ""; then CC_OUT="-c -o "; fi
 
 if test "$CC_INC" = ""; then CC_INC="-I"; fi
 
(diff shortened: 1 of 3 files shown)


sauwming:
This looks okay to me, but if anyone foresees that it can potentially break the build in some platforms (such as mingw or VC?), let us know.

Here's the consequence of the change:
- `make dep` will no longer use the `-c` option (in `rules.mak`: `dep="$(CC) -M $(DEPCFLAGS) $$F"; `).
  From what I check, it seems to be okay to build dependency without `-c` option.
- `cc-vc.mak` only sets: `export CC_OUT := /Fo`. Do we still use this?

sauwming (approved):

nanangizz (approved):
just tested using mingw-w64/msys2, looks okay.

jimying:
Next step is to supress the report errors , sorry i will not going on.
Please make a plan to fix it. ^_^

sauwming ***@***.***> 于2023年3月16日周四 14:07写道：

sauwming:
Currently we already have our hands full fixing the hundreds reported by another static analyser Coverity (#3390). There should be a lot of overlaps, so hopefully after fixing one, there should not be much left.

PR #3447, opened 2023-03-18
zero initialize the message buffer

oliverepper:
While using pjmedia I encounter a few error messages like this:

`10:45:55.540            pjsua_aud.c  .Unable to open sound device: Core audio errorP�` [status=450851]`

where there is some garbage behind the error message. I'm not 100% sure if this is the way to do it. Normally I'd use memset.

### DIFF
--- a/pjmedia/src/pjmedia-audiodev/errno.c
+++ b/pjmedia/src/pjmedia-audiodev/errno.c
@@ -81,11 +81,11 @@ PJ_DEF(pj_str_t) pjmedia_audiodev_strerror(pj_status_t statcode,
         statcode <= PJMEDIA_AUDIODEV_COREAUDIO_ERRNO_END)
     {
         int ca_err = PJMEDIA_AUDIODEV_COREAUDIO_ERRNO_START - statcode;
+        pj_str_t msg = pj_str("Core audio error");
 
         PJ_UNUSED_ARG(ca_err);
         // TODO: create more helpful error messages
         errstr.ptr = buf;
-        pj_str_t msg = pj_str("Core audio error");
         pj_strncpy_with_null(&errstr, &msg, bufsize);
         return errstr;
     } else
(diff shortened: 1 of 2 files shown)


oliverepper:
Checking my own code I use `char buffer[SIZE] = {0};` which might even end up in the bss section? But I am not sure if that works on every implementation.

jimying:
https://github.com/pjsip/pjproject/blob/651df5b50129b7c5a5feec8336dda4468d53d2b0/pjmedia/src/pjmedia-audiodev/errno.c#L88

Better to replace this line with:
```c++
pj_str_t msg = pj_str("Core audio erro");
pj_strncpy_with_null(&errstr, &msg, bufsize);
```

to keep same logic with `PJMEDIA_AUDIO_DEV_HAS_PORTAUDIO `

sauwming:
Please modify the patch as per @jimying 's suggestion, and we'll integrate it. Thanks.

pjmedia/src/pjmedia-audiodev/errno.c
```diff
@@ -85,7 +85,8 @@ PJ_DEF(pj_str_t) pjmedia_audiodev_strerror(pj_status_t statcode,
         PJ_UNUSED_ARG(ca_err);
         // TODO: create more helpful error messages
         errstr.ptr = buf;
-        pj_strcpy2(&errstr, "Core audio error");
+        pj_str_t msg = pj_str("Core audio error");
```
sauwming, line 88: only minor, but this declaration needs to be moved above, in order not to trigger any warning/failure with the old compiler standard

oliverepper: Not sure I understand. Do you mean `pj_str_t msg = pj_str("Core audio error");` should go below `pj_str_t errstr;` in line 74?

sauwming: No, it should go to L84 only

sauwming (approved):

nanangizz (approved):

trengginas (approved):

nanangizz:
@oliverepper, it looks like this is your first contribution to PJSIP, thank you. Could you please sign this contribution agreement [form](https://cla-assistant.io/pjsip/pjproject?pullRequest=3447)?

sauwming:
Ups... I didn't check about the form status. I thought it's automatic upon submitting a PR.

PR #3449, opened 2023-03-20
Fix asan (stack-use-after-scope) error on ICE transport_get_info()

trengginas:
On `transport_get_info()` from transport_ice.c :
``` c
    addr = NULL;
    if (pj_sockaddr_has_addr(&cand.addr)) {
        addr = &cand.addr;
    } else if (pj_ice_strans_has_sess(tp_ice->ice_st)) {
        unsigned i, cnt = PJ_ICE_ST_MAX_CAND;
        pj_ice_sess_cand cands[PJ_ICE_ST_MAX_CAND];
        pj_ice_strans_enum_cands(tp_ice->ice_st, 1, &cnt, cands);
        for (i = 0; i < cnt && !addr; ++i) {
            if (pj_sockaddr_has_addr(&cands[i].addr))
                addr = &cands[i].addr;
        }
    }
    if (addr) {
        pj_sockaddr_cp(&info->sock_info.rtp_addr_name, addr);
    } else {
        pj_sockaddr_init(PJ_AF_INET, &info->sock_info.rtp_addr_name, NULL, 9);
    }
```
The `addr` might point to a locally defined var (`cands`), and raised ASAN error when used.

### DIFF
--- a/pjmedia/src/pjmedia/transport_ice.c
+++ b/pjmedia/src/pjmedia/transport_ice.c
@@ -2258,6 +2258,8 @@ static pj_status_t transport_get_info(pjmedia_transport *tp,
     struct transport_ice *tp_ice = (struct transport_ice*)tp;
     pj_ice_sess_cand cand;
     pj_sockaddr_t *addr;
+    unsigned cnt = PJ_ICE_ST_MAX_CAND;
+    pj_ice_sess_cand cands[PJ_ICE_ST_MAX_CAND];
     pj_status_t status;
 
     pj_bzero(&info->sock_info, sizeof(info->sock_info));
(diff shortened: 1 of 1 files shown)


nanangizz (approved):

sauwming:
Correct, and this has also been fixed in `coverity01` branch:
https://github.com/pjsip/pjproject/blob/coverity01/pjmedia/src/pjmedia/transport_ice.c#L2255

sauwming (approved):

trengginas:
> Correct, and this has also been fixed in `coverity01` branch: https://github.com/pjsip/pjproject/blob/coverity01/pjmedia/src/pjmedia/transport_ice.c#L2255

Right, then we should just use the fix from the `coverity01` branch

sauwming:
#3390 has been merged

PR #3454, opened 2023-03-21
Avoid cancellation of retransmit timer after receiving provisional response of non-INVITE

ODCA117:
Currently, retransmit timer is cancelled for provisional response of non-INVITE request when using reliable transport. The RFC does not state that the timer should be reset.

RFC 3261 section [17.1.2.1](https://www.rfc-editor.org/rfc/rfc3261#section-17.1.2.1):
```
Non-INVITE transactions do not make use of ACK.  They are simple
   request-response interactions.  For unreliable transports, requests
   are retransmitted at an interval which starts at T1 and doubles until
   it hits T2.  If a provisional response is received, retransmissions
   continue for unreliable transports, but at an interval of T2.  The
   server transaction retransmits the last response it sent, which can
   be a provisional or final response, only when a retransmission of the
   request is received.  This is why request retransmissions need to
   continue even after a provisional response; they are to ensure
   reliable delivery of the final response.

   Unlike an INVITE transaction, a non-INVITE transaction has no special
   handling for the 2xx response.  The result is that only a single 2xx
   response to a non-INVITE is ever delivered to a UAC.
```

### DIFF
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -2665,11 +2665,6 @@ static pj_status_t tsx_on_state_calling( pjsip_transaction *tsx,
             }
 
         } else {
-            /* Cancel retransmit timer (for non-INVITE transaction, the
-             * retransmit timer will be rescheduled at T2.
-             */
-            tsx_cancel_timer(tsx, &tsx->retransmit_timer);
-
             /* For provisional response, only cancel retransmit when this
              * is an INVITE transaction. For non-INVITE, section 17.1.2.1
              * of RFC 3261 says that:
@@ -2678,6 +2673,8 @@ static pj_status_t tsx_on_state_calling( pjsip_transaction *tsx,
              */
             if (tsx->method.id == PJSIP_INVITE_METHOD) {
 
+                tsx_cancel_timer(tsx, &tsx->retransmit_timer);
+
                 /* Cancel timeout timer */
                 lock_timer(tsx);
                 tsx_cancel_timer(tsx, &tsx->timeout_timer);
@@ -2685,6 +2682,7 @@ static pj_status_t tsx_on_state_calling( pjsip_transaction *tsx,
 
             } else {
                 if (!tsx->is_reliable) {
+                    tsx_cancel_timer(tsx, &tsx->retransmit_timer);
                     tsx_schedule_timer(tsx, &tsx->retransmit_timer,
                                        &t2_timer_val, RETRANSMIT_TIMER);
                 }


sauwming:
How about section 17.1.2.1?
```
Non-INVITE transactions do not make use of ACK.  They are simple
   request-response interactions.  For unreliable transports, requests
   are retransmitted at an interval which starts at T1 and doubles until
   it hits T2.  If a provisional response is received, retransmissions
   continue for unreliable transports, but at an interval of T2.  The
   server transaction retransmits the last response it sent, which can
   be a provisional or final response, only when a retransmission of the
   request is received.  This is why request retransmissions need to
   continue even after a provisional response; they are to ensure
   reliable delivery of the final response.

   Unlike an INVITE transaction, a non-INVITE transaction has no special
   handling for the 2xx response.  The result is that only a single 2xx
   response to a non-INVITE is ever delivered to a UAC.
```

ODCA117:
The reason is due to a mandantory test done throuhg an ETSI test plan found here: [https://www.etsi.org/deliver/etsi_ts/102000_102099/10202702/04.01.01_60/ts_10202702v040101p.pdf](https://www.etsi.org/deliver/etsi_ts/102000_102099/10202702/04.01.01_60/ts_10202702v040101p.pdf )
The following test fails: SIP_RG_RT _TI_005
What the tester wanted was to have the T1 timer fire once after the provisional response was received, and after that have T2. The request will should be sent at time 0s, 0.5s, 4s, 4s, 4s, .... instead of 0s, 4s, 4s, 4s. I agree that the original behavior is more "correct" but I did not find anything about resetting T1 after provisional response is received.

sauwming:
Thanks for providing the document, it's very helpful.
Since SIP_RG_RT _TI_001-004 is for unreliable transport, I believe the test case 005 should be for reliable transport.

So yes, it looks like you are right. For reliable transport, we incorrectly cancel the retransmission timer.

Please find the attached patch to fix this:
[sip_tsx_timer_prov.txt](https://github.com/pjsip/pjproject/files/11037307/sip_tsx_timer_prov.txt)

sauwming (dismissed):

nanangizz:
AFAIK we only initiate retransmission on unreliable transport:
https://github.com/pjsip/pjproject/blob/74b0df1b047a4a8d55454f4f940f14ea7d978380/pjsip/src/pjsip/sip_transaction.c#L2581

The standard only mentions that the timer E must be started on an unreliable transport:
```
   If an unreliable transport is in use, the client transaction MUST set timer
   E to fire in T1 seconds.
```

sauwming:
To add what @nanangizz says is that we need further clarification about the test case `SIP_RG_RT _TI_005`, whether this happens for reliable or unreliable transport.

If it's unreliable transport, then the behaviour is already correct:
`If a provisional response is received, retransmissions
   continue for unreliable transports, but at an interval of T2.`

So the request timing of 0s, prov response, 4s, 4s, 4s is the correct result.

ODCA117:
The test is done using UDP, so unreliable transport.

sauwming:
Okay, so then my proposed won't make a difference.

The current behaviour seems already correct.

What you observe is not cancellation of timer, but rescheduling, i.e. removing the existing timer and scheduling it at T2, which is what's recommended by the RFC.

ODCA117:
Thank you for taking the time to investigate this. Will take it to the tester.

PR #3456, opened 2023-03-21
Terminate transaction on 200OK after BYE

ODCA117:
Transaction should be terminated after sending or reception of 200 OK response on BYE request. This to ensure following requests respond with: 481 Call/Transaction Does Not Exist

### DIFF
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -3093,10 +3093,14 @@ static pj_status_t tsx_on_state_proceeding_uac(pjsip_transaction *tsx,
         tsx_cancel_timer( tsx, &tsx->timeout_timer );
         unlock_timer(tsx);
 
-        /* For INVITE, the state moves to Terminated state (because ACK is
-         * handled in TU). For non-INVITE, state moves to Completed.
+        /* For INVITE and BYE, the state moves to Terminated state.
+         * For INVITE it is because ACK is handled in TU.
+         * For BYE it is because ACK should not be sent as response on 200 OK.
+         * For other methods, state moves to Completed.
          */
-        if (tsx->method.id == PJSIP_INVITE_METHOD) {
+        if (tsx->method.id == PJSIP_INVITE_METHOD ||
+            tsx->method.id == PJSIP_BYE_METHOD)
+        {
             tsx_set_state( tsx, PJSIP_TSX_STATE_TERMINATED, 
                            PJSIP_EVENT_RX_MSG, event->body.rx_msg.rdata, 0 );
             //return PJSIP_ETSXDESTROYED;


sauwming:
This seems to make sense, with the only reference I can find in the RFC about this is:
```
Alice confirms receipt of
   the BYE with a 200 (OK) response, which terminates the session and
   the BYE transaction.
```

But it seems when I test it, the current behaviour is already okay, i.e. after receiving 200/OK for the BYE, it will reject the subsequent requests with 481.
I tested with `sipp` where the request is immediately sent after OK, i.e.:
```
  <recv request="BYE">
  </recv>

  <send>
    <![CDATA[
      SIP/2.0 200 OK
      ...
    ]]>
  </send>

  <send>
    <![CDATA[
      INVITE sip:[service]@[remote_ip]:[remote_port] SIP/2.0
      ...
```
So I wonder how you found the issue?

ODCA117:
The test that failed for us had the following message sequence (omitting provisional messages and ACK)
```
INVITE sip:7002@172.17.15.112 SIP/2.0
Via: SIP/2.0/UDP 172.17.15.113:5060;rport;branch=z9hG4bKPjgZghS8OWTJmmgHpTrcjy8kP6YWNWk1-M
Max-Forwards: 70
From: "7003" <sip:7003@172.17.15.113>;tag=r9Ggj-I9o2B5tk8uO1K65LTUICdBETwk
To: sip:7002@172.17.15.112
Contact: "7003" <sip:7003@172.17.15.113:5060;ob>
Call-ID: NZXpjavbhYsHYIWK.GnARx6cCkStdJXX
CSeq: 4661 INVITE
Allow: PRACK, INVITE, ACK, BYE, CANCEL, UPDATE, INFO, SUBSCRIBE, NOTIFY, REFER, MESSAGE, OPTIONS
Supported: replaces, 100rel, timer, norefersub, answermode
Session-Expires: 1800
Min-SE: 90
Content-Type: application/sdp
Content-Length:   606
```

```
SIP/2.0 200 OK
Via: SIP/2.0/UDP 172.17.15.113:5060;rport;branch=z9hG4bKPjgZghS8OWTJmmgHpTrcjy8kP6YWNWk1-M
From: "7003" <sip:7003@172.17.15.113>;tag=r9Ggj-I9o2B5tk8uO1K65LTUICdBETwk
To: sip:7002@172.17.15.112;tag=256753688
Call-ID: NZXpjavbhYsHYIWK.GnARx6cCkStdJXX
CSeq: 4661 INVITE
Contact: <sip:7002@172.17.15.112>
Content-Type: application/sdp
Content-Length: 135
```

```
BYE sip:7002@172.17.15.112 SIP/2.0
Via: SIP/2.0/UDP 172.17.15.113:5060;rport;branch=z9hG4bKPj02GiQags1mJqIDOp-zbbwvpZ1CLDi61H
Max-Forwards: 70
From: "7003" <sip:7003@172.17.15.113>;tag=r9Ggj-I9o2B5tk8uO1K65LTUICdBETwk
To: sip:7002@172.17.15.112;tag=256753688
Call-ID: NZXpjavbhYsHYIWK.GnARx6cCkStdJXX
CSeq: 4662 BYE
Content-Length:  0
```

```
SIP/2.0 200 OK
Via: SIP/2.0/UDP 172.17.15.113:5060;rport;branch=z9hG4bKPj02GiQags1mJqIDOp-zbbwvpZ1CLDi61H
From: "7003" <sip:7003@172.17.15.113>;tag=r9Ggj-I9o2B5tk8uO1K65LTUICdBETwk
To: sip:7002@172.17.15.112;tag=256753688
Call-ID: NZXpjavbhYsHYIWK.GnARx6cCkStdJXX
CSeq: 4662 BYE
Content-Length: 0
```

```
BYE sip:7003@172.17.15.113:5060;ob SIP/2.0
Via: SIP/2.0/UDP 172.17.15.112:5060;branch=z9hG4bK3c2b780-6ec3898d
Max-Forwards: 70
From: sip:7002@172.17.15.112;tag=256753688
To: "7003" <sip:7003@172.17.15.113>;tag=r9Ggj-I9o2B5tk8uO1K65LTUICdBETwk
Call-ID: NZXpjavbhYsHYIWK.GnARx6cCkStdJXX
CSeq: 1 BYE
Content-Length: 0
```

```
SIP/2.0 200 OK
Via: SIP/2.0/UDP 172.17.15.112:5060;received=172.17.15.112;branch=z9hG4bK3c2b780-6ec3898d
Call-ID: NZXpjavbhYsHYIWK.GnARx6cCkStdJXX
From: <sip:7002@172.17.15.112>;tag=256753688
To: "7003" <sip:7003@172.17.15.113>;tag=r9Ggj-I9o2B5tk8uO1K65LTUICdBETwk
CSeq: 1 BYE
Content-Length:  0
```

sauwming:
Thanks for the clarification.

Please find the attached patch to fix some minor fixes in yours:
- we use max column of 80
- we use spaces instead of tabs
- comments for non-INVITE was accidentally removed

[sip_tsx_timer_prov.txt](https://github.com/pjsip/pjproject/files/11037366/sip_tsx_timer_prov.txt)

sauwming:
Sorry I provided the wrong file.
[sip_tsx_bye.txt](https://github.com/pjsip/pjproject/files/11038143/sip_tsx_bye.txt)

sauwming (dismissed):

ODCA117:
No problem. I easily done with all the different PR

pjsip/src/pjsip/sip_transaction.c
```diff
@@ -3093,10 +3093,14 @@ static pj_status_t tsx_on_state_proceeding_uac(pjsip_transaction *tsx,
...
+         * For BYE it is because ACK should not be sent as response on 200 OK.
+         * For other methods, state moves to Completed.
          */
-        if (tsx->method.id == PJSIP_INVITE_METHOD) {
+        if (tsx->method.id == PJSIP_INVITE_METHOD ||
+            tsx->method.id == PJSIP_BYE_METHOD)
+        {
             tsx_set_state( tsx, PJSIP_TSX_STATE_TERMINATED, 
                            PJSIP_EVENT_RX_MSG, event->body.rx_msg.rdata, 0 );
             //return PJSIP_ETSXDESTROYED;
```
nanangizz, line 3106: According to the standard, non-INVITE tsx must go through the Completed state, I think the current flow is already correct.

https://www.rfc-editor.org/rfc/rfc3261#section-17.1.2:
```
   If a
   final response (status codes 200-699) is received while in the
   "Proceeding" state, the response MUST be passed to the TU, and the
   client transaction MUST transition to the "Completed" state.

   Once the client transaction enters the "Completed" state, it MUST set
   Timer K to fire in T4 seconds for unreliable transports, and zero
   seconds for reliable transports.  The "Completed" state exists to
   buffer any additional response retransmissions that may be received
   (which is why the client transaction remains there only for
   unreliable transports).
```

This patch works for the original "problem" reported in the PR because the dialog is destroyed immediately when BYE tsx is terminated immediately. But, terminating BYE tsx immediately (for unreliable transport) seems to violate the standard. Just curious whether the original "problem" is really an issue?

ODCA117: The problem comes from a failed test case in the following test plan: [https://www.etsi.org/deliver/etsi_ts/102000_102099/10202702/04.01.01_60/ts_10202702v040101p.pdf](https://www.etsi.org/deliver/etsi_ts/102000_102099/10202702/04.01.01_60/ts_10202702v040101p.pdf) Test: SIP_CC_OE_CR_V_008.
It could be argued that the test is too strict and that in reality it should not cause any problems. But we still encountered problems and need a way to fix this.

sauwming: How about the snippet from the RFC that I quoted above:
```
Alice confirms receipt of
   the BYE with a 200 (OK) response, which terminates the session and
   the BYE transaction.
```

nanangizz: > How about the snippet from the RFC that I quoted above:
> 
> ```
> Alice confirms receipt of
>    the BYE with a 200 (OK) response, which terminates the session and
>    the BYE transaction.
> ```

It looks like a sample scenario (which indeed can happen on a reliable transport) rather than rules. There are some other sample scenarios that the BYE transaction is not terminated immediately, but going through Completed state and waiting for timer K [here](https://www.rfc-editor.org/rfc/rfc5407.html).

sauwming:
After re-reading the RFCs, I think @nanangizz is correct.

The test case states that:
```
Ensure that the IUT, once a dialog has been established, having sent a BYE request,
on receipt of a Success (200 OK) response considers the session and the dialog terminated.
```
Unfortunately, this doesn't seem to be correct. The session is considered to be terminated, but the dialog is not.

[RFC 3261, section 15.1.2](https://www.rfc-editor.org/rfc/rfc3261#section-15.1.2)
```
A UAS core receiving a BYE request for an existing dialog MUST follow
   the procedures of Section 12.2.2 to process the request.  Once done,
   the UAS SHOULD terminate the session (and therefore stop sending and
   listening for media).  The only case where it can elect not to are
   multicast sessions, where participation is possible even if the other
   participant in the dialog has terminated its involvement in the
   session.  Whether or not it ends its participation on the session,
   the UAS core MUST generate a 2xx response to the BYE, and MUST pass
   that to the server transaction for transmission.

   The UAS MUST still respond to any pending requests received for that
   dialog.  It is RECOMMENDED that a 487 (Request Terminated) response
   be generated to those pending requests.
```

Note that the above is for UAS, but it just shows that dialog can still exist after 2xx.

For UAC, RFC 3261 doesn't clearly specify what happens after receipt of 2xx, but [RFC 5407](https://www.rfc-editor.org/rfc/rfc5407.html) states:
```
Mortal (Mort):  The caller and callee enter the Mortal state by
      sending or receiving a BYE.  The UA MUST NOT send any new requests
      within the dialog because there is no dialog.  (Here, the new
      requests do not include ACK for 2xx and BYE for 401 or 407, as
      further explained in Appendix D below.)  In the Mortal state, BYE
      can be accepted, and the other messages in the INVITE dialog usage
      are responded to with an error.
```

[Appendix D](https://www.rfc-editor.org/rfc/rfc5407.html#appendix-D).  Notes on the Request in the Mortal State
```
In the Mortal state, only a BYE can be accepted, and the other
   messages in the INVITE dialog usage are responded to with an error.
```
Then check any of the figure, such as in [3.1.2](https://www.rfc-editor.org/rfc/rfc5407.html#section-3.1.2). We can see that upon receipt of 200 for the BYE, the client is still in mortal state.

In conclusion, the BYE is correctly accepted here.

**
Now, what to do if you want a different behaviour? Note that the RFC doesn't mandate that the BYE MUST be accepted. So I believe it should be okay if you want to respond the BYE with `487 (Request Terminated)`, or `481 (Call Does not Exist)` instead.

In PJSIP, this is done in `sip_inv.c` `inv_on_state_disconnected`:
```
        /*
         * Respond BYE with 200/OK
         */
        if (tsx->method.id == PJSIP_BYE_METHOD) {
            inv_respond_incoming_bye( inv, tsx, rdata, e );
```

Thus, my suggestion is to first, highlight this point to the tester that it is okay and in fact recommended, to accept the BYE with 200/OK, during the scenario.

But if it's not possible, then I would suggest to change the behaviour in `sip_inv` instead, by replying the BYE with a different response code.

ODCA117:
Thank you for taking the time to investigate this and for suggesting a potential fix. 
We will talk to the tester and see what can be done.

PR #3457, opened 2023-03-21
Consider session terminated on 408 and 481 response

ODCA117:
Client must consider session terminated as soon as BYE is passed to the transaction client. On reseption of 408 or 481, the UAC MUST consider the session and the dialog terminated.

### DIFF
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -3220,6 +3220,21 @@ static pj_status_t tsx_on_state_proceeding_uac(pjsip_transaction *tsx,
             status = tsx_send_msg( tsx, ack_tdata);
             if (status != PJ_SUCCESS)
                 return status;
+
+        } else if ((tsx->method.id == PJSIP_BYE_METHOD) &&
+                   (tsx->status_code == PJSIP_SC_REQUEST_TIMEOUT ||
+                    tsx->status_code == PJSIP_SC_CALL_TSX_DOES_NOT_EXIST))
+        {
+            /* RFC 3261 15.1.1:
+             * If the response for the BYE is a 481 (Call/Transaction Does
+             * Not Exist) or a 408 (Request Timeout) or no response at all
+             * is received for the BYE (that is, a timeout is returned by
+             * the client transaction), the UAC MUST consider the
+             * session and the dialog terminated.
+             */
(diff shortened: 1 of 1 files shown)


sauwming:
This failed pjsua test.

sauwming:
If you have the test case for this one, please let us know.

ODCA117:
Yes we do have the test case:[https://www.etsi.org/deliver/etsi_ts/102000_102099/10202702/04.01.01_60/ts_10202702v040101p.pdf](https://www.etsi.org/deliver/etsi_ts/102000_102099/10202702/04.01.01_60/ts_10202702v040101p.pdf)
Two testa are failing: SIP_CC_OE_CR_V_008 and SIP_CC_OE_CR_V_009
The error we have is the following.

The pjsip based application is the UAC
```
UAC UAS
INVITE ->
<- TRYING
<- RINGING
<- OK
ACK ->
BYE -> 
<- OK
<- BYE
OK -> 
```
The error is that first the UAC (pjsip based application) sends a BYE and receives an OK. Then the UAS (test software) sends a BYE on which UAC reply OK. According to the test, it should send 481 Call Leg/Transaction Does Not Exist.

ODCA117:
More details is that the second BYE sent by the test software have `CSeq: 1 BYE`
While the pjsip sends a BYE with `CSeq: 4662 BYE`

ODCA117:
yeah sorry about that. They are similar but not the same.

```
UAC UAS
INVITE ->
<- TRYING
<- RINGING
<- OK
ACK ->
BYE -> 
<- 481 Call Leg/Transaction Does Not Exist
<- BYE
OK -> 
```
Compared to the other where a new BYE is sent from the test software, a 481 is sent and after that a BYE. According to the test it should be terminated. The actual test is this: SIP_CC_OE_CR_V_009

sauwming:
The failed test is because the patch also affects other method as well (such as UPDATE during early state). Since the issue is only for BYE, I think it's better to add check for the method.

Please check the modified patch:
[sip_tsx_bye_408_481.txt](https://github.com/pjsip/pjproject/files/11040921/sip_tsx_bye_408_481.txt)

ODCA117:
Applied the modified patch

sauwming:
Btw, would you be able to immediately confirm whether any of our proposed patches work? Or do you need to wait some time, for example you have to resubmit to them and wait for the test results back?

sauwming (approved):

PR #3458, opened 2023-03-21
Handle 2nd INVITE when in early state

ODCA117:
Ensures 2nd INVITE is handled if in state early.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -5052,6 +5052,42 @@ static void inv_on_state_early( pjsip_inv_session *inv, pjsip_event *e)
             pj_assert(!"Unexpected INVITE tsx state");
             break;
         }
+    } else if (tsx->method.id == PJSIP_INVITE_METHOD &&
+               tsx->role == PJSIP_ROLE_UAS &&
+               tsx->state == PJSIP_TSX_STATE_TRYING)
+    {
+        /* Handle incoming second INVITE. See RFC 3261 section 14.2. */
+        pjsip_rx_data *rdata = e->body.tsx_state.src.rdata;
+        pjsip_tx_data *tdata;
+        pj_status_t status;
+        int code;
+        const pj_str_t *reason;
+
+        if (inv->invite_tsx->role == PJSIP_ROLE_UAC)
+            code = PJSIP_SC_REQUEST_PENDING;
+        else
+            code = PJSIP_SC_INTERNAL_SERVER_ERROR;
+
+        reason = pjsip_get_status_text(code);
+
+        /* Can not receive second INVITE while another one is pending. */
+        status = pjsip_dlg_create_response( inv->dlg, rdata, code,
+                                            reason, &tdata);
+        if (status != PJ_SUCCESS)
+            return;
+
+        if (code == PJSIP_SC_INTERNAL_SERVER_ERROR) {
+            /* MUST include Retry-After header with random value
+             * between 0-10.
+             */
+            pjsip_retry_after_hdr *ra_hdr;
+            int val = (pj_rand() % 10);
+
+            ra_hdr = pjsip_retry_after_hdr_create(tdata->pool, val);
+            pjsip_msg_add_hdr(tdata->msg, (pjsip_hdr*)ra_hdr);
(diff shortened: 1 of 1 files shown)


sauwming:
As a reference, here's [RFC 3261 section 14.2](https://www.rfc-editor.org/rfc/rfc3261#section-14.2):
```
   A UAS that receives a second INVITE before it sends the final
   response to a first INVITE with a lower CSeq sequence number on the
   same dialog MUST return a 500 (Server Internal Error) response to the
   second INVITE and MUST include a Retry-After header field with a
   randomly chosen value of between 0 and 10 seconds.

   A UAS that receives an INVITE on a dialog while an INVITE it had sent
   on that dialog is in progress MUST return a 491 (Request Pending)
   response to the received INVITE.
```

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -5052,7 +5052,48 @@ static void inv_on_state_early( pjsip_inv_session *inv, pjsip_event *e)
...
+         * Handle incoming second INVITE. See RFC 3261 section 14.2.
+         */
+        pjsip_rx_data *rdata = e->body.tsx_state.src.rdata;
+        pjsip_tx_data *tdata;
+        pj_status_t status;
+
+        /* Check if we have INVITE pending. */
+        if (inv->invite_tsx && inv->invite_tsx!=tsx) {
+        int code;
+        pj_str_t *reason;
```
sauwming, line 5068: `const pj_str_t`

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -5052,7 +5052,48 @@ static void inv_on_state_early( pjsip_inv_session *inv, pjsip_event *e)
...
+                            reason, &tdata);
+        if (status != PJ_SUCCESS)
+            return;
+
+        if (code == 500) {
+            /* MUST include Retry-After header with random value
+                * between 0-10.
+                */
+            pjsip_retry_after_hdr *ra_hdr;
+            int val = (pj_rand() % 10);
```
sauwming, line 5089: since it's between 0-10, it should be `(pj_rand() % 9) + 1`

wosrediinanatour:
Please use enum instead of magic numbers, e.g. 500 or 491.

ODCA117:
Fixed the comments

sauwming:
Please find the attached patch below. Mostly, just to fix the indentation, but also one build failure of missing semicolon.

I have tested it here and it seems to work as expected.
[sip_inv_early.txt](https://github.com/pjsip/pjproject/files/11037521/sip_inv_early.txt)

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -5052,6 +5052,42 @@ static void inv_on_state_early( pjsip_inv_session *inv, pjsip_event *e)
...
+                                            reason, &tdata);
+        if (status != PJ_SUCCESS)
+            return;
+
+        if (code == PJSIP_SC_INTERNAL_SERVER_ERROR) {
+            /* MUST include Retry-After header with random value
+             * between 0-10.
+             */
+            pjsip_retry_after_hdr *ra_hdr;
+            int val = (pj_rand() % 9) + 1;
```
nanangizz, line 5084: There is a similar handling in `inv_on_state_confirmed()`, and it use:
```
int val = (pj_rand() % 10);
```
Yes it is unclear whether `between 0-10` should be inclusive, but I guess we better apply the same formula in early and confirmed?

ODCA117: So changing back to  `int val = (pj_rand() % 10);`?

sauwming: It looks like `between` in the RFC means inclusive such as `any response with a status code between 100 and 199 is referred to as a "1xx response"`.

The correct one should be `pj_rand() % 11`. But I think `% 10` is safer in case that there's any UA that rejects a value of 10.

So yes, please revert it to `pj_rand() % 10`. Thank you.

ODCA117: Done

nanangizz (approved):

PR #3459, opened 2023-03-21
Only add "received" param if necessary

ODCA117:
"received" param in VIA header is only added if "sent_by" differs from the address packet was receive from.

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -2143,10 +2143,15 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
 
         /* For request: */
         if (rdata->msg_info.msg->type == PJSIP_REQUEST_MSG) {
-            /* always add received parameter to the via. */
-            pj_strdup2(rdata->tp_info.pool, 
-                       &rdata->msg_info.via->recvd_param, 
-                       rdata->pkt_info.src_name);
+            /* always add received parameter to the via if "sent_by" differs from
+             * actual received IP */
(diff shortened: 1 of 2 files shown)


sauwming:
But it shouldn't cause any harm either, right, if there's always a received param?

ODCA117:
I Agree, but again a test failed due to this and I could not find anything in the RFC that could motivate having it there:

[18.2.1](https://www.rfc-editor.org/rfc/rfc3261.html#section-18.2.1) Receiving Requests
When the server transport receives a request over any transport, it
MUST examine the value of the "sent-by" parameter in the top Via
header field value. If the host portion of the "sent-by" parameter
contains a domain name, or if it contains an IP address that differs
from the packet source address, the server MUST add a "received"
parameter to that Via header field value. This parameter MUST
contain the source address from which the packet was received. This
is to assist the server transport layer in sending the response,
since it must be sent to the source IP address from which the request
came.

Test failing for us is [https://www.etsi.org/deliver/etsi_ts/102000_102099/10202702/04.01.01_60/ts_10202702v040101p.pdf](https://www.etsi.org/deliver/etsi_ts/102000_102099/10202702/04.01.01_60/ts_10202702v040101p.pdf) test SIP_CC_TE_CE_V_014
It states that the reply need to have VIA header copy from the INVITE.

sauwming:
I think the test is too strict. I don't think there's anything wrong with having a `received` param in the Via.

Again, just appeal to the tester, if possible.
Otherwise, since there's nothing fundamentally wrong within the library itself, I suggest to implement this in your app, as a module. Basically, you check the outgoing message, compare the received param, if it's the same, remove it.

ODCA117:
Thank you for the response.
I do agree that it is a strict test and that pjsip behave good. Will reject the PR
We will take a look into creating our own module. I appreciate the input!

PR #3460, opened 2023-03-21
Check require on all transactions inside dialog

ODCA117:
Ensures that "Require" field is checked for all transactions inside the dialog.

### DIFF
--- a/pjsip/include/pjsip-ua/sip_inv.h
+++ b/pjsip/include/pjsip-ua/sip_inv.h
@@ -659,6 +659,18 @@ PJ_DECL(pj_status_t) pjsip_inv_verify_request3( pjsip_rx_data *rdata,
                                                 pjsip_endpoint *endpt,
                                                 pjsip_tx_data **tdata);
 
+/**
+ * Verify incoming non-INVITE request.
+ * Patched to make sure all requests gets checked against Require header.
+ * The same check is done for INVITE requests.
+ */
+PJ_DEF(pj_status_t)
+pjsip_non_inv_verify_request(pjsip_rx_data* rdata,
+                             pj_pool_t* tmp_pool,
+                             unsigned* options,
+                             pjsip_dialog* dlg,
+                             pjsip_endpoint* endpt,
+                             pjsip_tx_data** p_tdata);
 
 /**
  * Create UAS invite session for the specified dialog in dlg. Application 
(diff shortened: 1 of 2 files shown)


sauwming:
Unfortunately, this patch has some problems:
- Missing implementation of the verification function.
- If it's to check incoming request, why is it in `on_rx_response`?
- If it's for non-INVITE, it shouldn't be in `sip_inv`.

My suggestion is, if this is dialog related, you can consider putting it in `sip_dialog`. But please provide the RFC recommendation as well. If the RFC doesn't mandate this, then perhaps it's better if the implementation is done in application level instead.

ODCA117:
I can see that this may be moved to the sip_dialog instead, or maybe to the application layer. 
The test failed for us had the following sequence.
UAC          UAS
INVITE -> 
             <- TRYING
             <- OK
ACK      -> 
(Call Active for some time)
BYE       ->   // This message is the first to contain a Require header which is bad.
             <- OK

According to the test we had, it should have rejected the BYE with 420 Bad Extension

ODCA117:
For the RFC behavior I can only find [https://www.rfc-editor.org/rfc/rfc3261#section-8.2.2.3](https://www.rfc-editor.org/rfc/rfc3261#section-8.2.2.3) where it states the following:
```
  Note that Require and Proxy-Require MUST NOT be used in a SIP CANCEL
  request, or in an ACK request sent for a non-2xx response
```

sauwming:
And if you include the complete paragraph, it will say:
```
Note that Require and Proxy-Require MUST NOT be used in a SIP CANCEL
   request, or in an ACK request sent for a non-2xx response.  These
   header fields MUST be ignored if they are present in these requests.
```

which is also in line with what it states above in [8.2.2](https://www.rfc-editor.org/rfc/rfc3261#section-8.2.2) 
```
Header Inspection

   If a UAS does not understand a header field in a request (that is,
   the header field is not defined in this specification or in any
   supported extension), the server MUST ignore that header field and
   continue processing the message.  A UAS SHOULD ignore any malformed
   header fields that are not necessary for processing requests
```

So, I believe rejecting the BYE with 420 is too harsh. The `Require` header bears no significance anyway for BYE, so we should just ignore it and proceed with the termination of the session.

Perhaps you can appeal to the tester? ;)

ODCA117:
Thank you for the input. 
I will take a look into fixing this in our application.

PR #3461, opened 2023-03-21
All OPTIONS should be accepted

ODCA117:
All OPTIONS requests should be accepted. They are then handled in different module.


ODCA117:
We are using the OPTIONS handler in PJSUA-LIB. What is the mest way to change a priority of a module created by PJSUA-LIB?

sauwming:
You shouldn't modify the library. Instead, you can simply copy paste it to your application and change it however you want.

sauwming:
why should you choose the same name? I haven't checked, but just use a different one.

PR #3467, opened 2023-03-24
Fix ipjsystest build error

trengginas:
This PR will fix the build error for `ipjsystest` on current xCode (14.2)
Changes:
- Fix path to input.8.wav 
- Update library/arch used 
- Update the app permission
- Error : `Application windows are expected to have a root view controller at the end of application launch`

### DIFF
--- a/pjsip-apps/src/ipjsystest/ipjsystest-Info.plist
+++ b/pjsip-apps/src/ipjsystest/ipjsystest-Info.plist
@@ -28,8 +28,6 @@
 	<string>MainWindow</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
-	<key>NSCameraUsageDescription</key>
-	<string>Camera permission required</string>
 	<key>NSMicrophoneUsageDescription</key>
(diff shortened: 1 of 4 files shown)


pjsip-apps/src/ipjsystest/ipjsystest-Info.plist
```diff
@@ -26,9 +26,16 @@
 	<true/>
 	<key>NSMainNibFile</key>
 	<string>MainWindow</string>
+	<key>LSRequiresIPhoneOS</key>
+	<true/>
+	<key>NSCameraUsageDescription</key>
```
sauwming, line 31: why would ipjsystest require camera?

sauwming:
Based on the original report, it seems that his error is caused by missing declaration of `PJ_AUTOCONF` or incorrect configuration. So what build errors do this PR fix? Perhaps you can add it in the description, so people encountering similar errors can search for this patch.

sauwming (approved):

nanangizz (approved):

PR #3468, opened 2023-03-27
Add option to maintain payload type to codec mapping to make sure one payload type is not used by different codec

trengginas:
The [rfc](https://www.ietf.org/rfc/rfc3264.html#section-8.3.2) states:
```
if A generates an offer
   with G.711 assigned to dynamic payload type number 46, payload type
   number 46 MUST refer to G.711 from that point forward in any offers
   or answers for that media stream within the session.  However, it is
   acceptable for multiple payload type numbers to be mapped to the same
   codec, so that an updated offer could also use payload type number 72
   for G.711.
```

Currently there's no check of the if the payload type on the SDP is already used by other/different codec.
This ticket will add the option to maintain the payload type used on the session, so application 
can validate the SDP and use a new one in the case on used pt.


pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1545,6 +1571,106 @@ static pj_status_t create_answer( pj_pool_t *pool,
...
+            pj_strdup(neg->pool_neg, &neg->sess_med_pt_map[i].med_type, med_type);
+            neg->sess_med_pt_map_cnt++;
+        }
+        return &neg->sess_med_pt_map[i];
+    }
+    return NULL;
+}
+
+static void update_media_pt_map(const pjmedia_sdp_session* sess,
+                                pjmedia_sdp_neg* neg)
```
nanangizz, line 1632: minor/naming: 
- this function seems to work on the session level, so perhaps the `media` can be confusing?

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1545,6 +1571,106 @@ static pj_status_t create_answer( pj_pool_t *pool,
...
+            if (pt_exists(med_pt_map, pt, &codec) != 3) {
+                med_pt_map->med_pt_map[med_pt_map->med_pt_map_cnt].pt = pt;
+                pj_strdup(neg->pool_neg,
+                          &med_pt_map->med_pt_map[med_pt_map->med_pt_map_cnt++].codec,
+                          &codec);
+            }
+        }
+    }
+}
+#endif
```
nanangizz, line 1673: minor: for readibility, leave one or two empty lines between functions?

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1545,6 +1571,106 @@ static pj_status_t create_answer( pj_pool_t *pool,
...
+            pt_len = sess_media->desc.fmt[j].slen + 1;
+            codec.ptr = attr->value.ptr + pt_len;
+            codec.slen = attr->value.slen - pt_len;
+
+            if (pt_exists(med_pt_map, pt, &codec) != 3) {
+                med_pt_map->med_pt_map[med_pt_map->med_pt_map_cnt].pt = pt;
+                pj_strdup(neg->pool_neg,
+                          &med_pt_map->med_pt_map[med_pt_map->med_pt_map_cnt++].codec,
+                          &codec);
+            }
```
nanangizz, line 1669: - If there is a reused PT for different codecs, that PT will be mapped to two codecs, is this correct?
- Better put some comments about the behavior for future ref, also better to explain why.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1793,3 +1929,128 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_fmt_match(pj_pool_t *pool,
...
+                if (p_exists == 1 || p_exists == 2) {
+                    /* pt is assigned to another codec, find existing pt used by the
+                     * codec or a new one.
+                     */
+                    unsigned new_pt;
+                    pj_str_t new_pt_str;
+
+                    new_pt = find_valid_pt(neg->pool_active, &used_pt_map, &codec);
+                    if (new_pt == 0)
+                        return PJ_ENOTFOUND;
```
nanangizz, line 2018: - Print log too? Error code can be confusing in debugging.
- Upon this failure, some codecs may have been rewritten while others leaved as is, is there a chance that a media SDP contains two codecs with the same PT and how it is handled?
- Better add comment/docs here/somewhere about the behavior if there is no more valid PT.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1793,3 +1929,128 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_fmt_match(pj_pool_t *pool,
...
+            i++;
+        }
+    }
+    if (pt_check > 127 || i == max_pt_cnt) {
+        /* No more available PT */
+        return 0;
+    }
+
+    used_pt->med_pt_map[used_pt->med_pt_map_cnt].pt = pt_check;
+    pj_strdup(pool, &used_pt->med_pt_map[used_pt->med_pt_map_cnt++].codec, codec);
```
nanangizz, line 1968: minor: this function is not just finding PT (=IMHO the map is read-only), it may assign a PT & update the map too, perhaps better rename it to something like `assign`?

pjmedia/src/pjmedia/sdp_neg.c
trengginas: when there's a failure in finding a new pt, then the previous changes (media format) will be reverted.

pjmedia/src/pjmedia/sdp_neg.c
trengginas: Yes, that is correct. This is to handle if remote might send a conflicting PT/codec.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1545,6 +1576,118 @@ static pj_status_t create_answer( pj_pool_t *pool,
...
+}
+
+#if PJMEDIA_SDP_NEG_MAINTAIN_SESSION_PT
+
+static media_pt_map* get_media_pt_map(pjmedia_sdp_neg* neg, pj_str_t* med_type)
+{
+    unsigned i = 0;
+    unsigned max_cnt = PJ_ARRAY_SIZE(neg->med_pt_map);
+    for (; i < neg->med_pt_map_cnt && i < max_cnt; ++i) {
+        if (pj_strcmp(&neg->med_pt_map[i].med_type, med_type) == 0) {
```
sauwming, line 1622: the mapping is per stream, not based on media type.

pjmedia/include/pjmedia/sdp_neg.h
```diff
@@ -809,6 +809,25 @@ PJ_DECL(pj_status_t) pjmedia_sdp_neg_fmt_match( pj_pool_t *pool,
...
+ * is set to 0.
+ *
+ * @param neg           The SDP negotiator instance.
+ * @param local         The local SDP session.
+ * 
+ * @return              PJ_SUCCESS when the SDP session is valid. However, if there is a
+ *                      conflict with the payload type, the function will return an error
+ *                      to indicate the failure to modify it.
+ */
+PJ_DECL(pj_status_t) pjmedia_sdp_neg_validate_pt(pjmedia_sdp_neg* neg,
```
sauwming, line 828: do we really need a new API?

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -183,6 +211,9 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_create_w_remote_offer(pj_pool_t *pool,
      * SDPs pool after a successful SDP nego.
      */
     neg->pool_active = pool;
+#if PJMEDIA_SDP_NEG_MAINTAIN_SESSION_PT
+    neg->pool = pool;
```
sauwming, line 215: be careful that the pool is a flip-flop one.

pjmedia/src/pjmedia/sdp_neg.c
nanangizz: considered that too, but what if the media type of a stream is changed?

pjmedia/src/pjmedia/sdp_neg.c
nanangizz: Then how we decide for that shared/reused PT, e.g: used for which codec?

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -35,6 +54,11 @@ struct pjmedia_sdp_neg
     pj_bool_t             answer_with_multiple_codecs;
     pj_bool_t             has_remote_answer;
     pj_bool_t             answer_was_remote;
+#if PJMEDIA_SDP_NEG_MAINTAIN_SESSION_PT
+    unsigned              med_pt_map_cnt;
+    media_pt_map          med_pt_map[PJMEDIA_MAX_SDP_MEDIA];
```
nanangizz, line 59: As the map is per media type, and currently we only "define" PTs for audio & video, perhaps `PJMEDIA_MAX_SDP_MEDIA` is not ideal for the array size as it does not really correspond to max media type number, note that currently the default is 16. IMO 2-3 is reasonable. To be safe, better put check somewhere that the map only applies to audio & video media.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1793,3 +1946,167 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_fmt_match(pj_pool_t *pool,
...
+                codec.ptr = attr->value.ptr + pt_len;
+                codec.slen = attr->value.slen - pt_len;
+
+                ori_attr_list[attr_cnt] = attr;
+
+                p_exists = pt_exists(&used_pt_map, pt, &codec);
+                if (p_exists == 1 || p_exists == 2) {
+                    /* pt is assigned to another codec, find existing pt used by the
+                     * codec or a new one.
+                     */
```
nanangizz, line 2040: There are two cases in the check, but only one is described in the comment, better put description for both cases.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1793,3 +1946,167 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_fmt_match(pj_pool_t *pool,
...
+    const pj_str_t telephone_event = pj_str("telephone-event");
+
+    if (pj_strncmp(codec, &telephone_event, telephone_event.slen) == 0)
+        pt_check = PJMEDIA_RTP_PT_TELEPHONE_EVENTS;
+    else
+        pt_check = START_DYNAMIC_PT;
+    
+    for (; i < used_pt_num && i < max_pt_cnt; ++i) {
+        if (pj_strcmp(codec, &used_pt->map[i].codec) == 0) {
+            return used_pt->map[i].pt;
```
nanangizz, line 1965: Ops, still the new name does not seem to represent the function action, i.e: it looks like the function lookups PT for a given codec, and assign if not found. Well, perhaps better put description what the function really does :)

pjmedia/src/pjmedia/sdp_neg.c
sauwming: hmm... the RFC doesn't really say, perhaps we can consider resetting the mapping?

But it should be quite a rare scenario, so leaving it as is should also be okay?

A more common scenario is removing and reusing the media slot. So, I guess the mapping needs to be reset upon removal.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1793,3 +1946,167 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_fmt_match(pj_pool_t *pool,
...
+        pjmedia_sdp_attr* ori_attr_list[PJMEDIA_MAX_SDP_ATTR];
+        pj_str_t ori_fmt[PJMEDIA_MAX_SDP_FMT];
+        pjmedia_sdp_media* sess_media = sdp_sess->media[i];
+        media_pt_map used_pt_map;
+        media_pt_map* med_pt_map = get_media_pt_map(neg, &sdp_sess->media[i]->desc.media);
+
+        if (!med_pt_map) {
+            PJ_LOG(4, (THIS_FILE, "Unable to find media pt map for [%.*]",
+                       sdp_sess->media[i]->desc.media.slen, 
+                       sdp_sess->media[i]->desc.media.ptr));
```
nanangizz, line 2010: To help troubleshooting, for example reading this line in a log file in the future when we may already forget this feature exists, there will be questions like "who & why find pt map?", "what happen if it is not found", so perhaps better add something like "..., PTs will not be validated for this media" for initial hints on reading logs?

pjmedia/src/pjmedia/sdp_neg.c
nanangizz: Media type change is allowed in rfc3264 8.3.3, but yes I believe it should be rare.
But we've also seen media/stream removed which is actually illegal.

pjmedia/src/pjmedia/sdp_neg.c
trengginas: Currently, pick the first PT assigned to the codec.

pjmedia/src/pjmedia/sdp_neg.c
trengginas: The call to `pjmedia_sdp_neg_create_w_remote_offer()` with temporary pool is called to verify the incoming request `pjsip_inv_verify_request3()`. So it's not used long term, it should be safe

pjmedia/include/pjmedia/sdp_neg.h
trengginas: The PT mapping is stored in `pjmedia_sdp_neg`, which is opaque. We need the API to access the mapping.

pjmedia/src/pjmedia/sdp_neg.c
trengginas: Consider the scenario:
Offer Media : Audio[1] Video[2] Audio[3]
ReOffer Media : Audio[4] Audio[5]
In this case we need to reset the mapping for Video[2] and Audio[3]. So Audio[4] will use mapping on Audio[1] and Audio[5] will use a new mapping. Or the map will be reset entirely if there's a media removal

sauwming: As usual, let's go back to the RFC:
```
The media type (audio, video, etc.) for a stream MAY be changed.  It
   is RECOMMENDED that the media type be changed (as opposed to adding a
   new stream), when the same logical data is being conveyed, but just
   in a different media format.  This is particularly useful for
   changing between voiceband fax and fax in a single stream, which are
   both separate media types.  To do this, the offerer creates a new
   media description, with a new media type, in place of the description
   in the previous SDP which is to be changed.

   The corresponding media stream in the answer is formulated as
   described in Section 6.  Assuming the stream is acceptable, the
   answerer SHOULD begin sending with the new media type and formats as
   soon as it receives the offer. The offerer MUST be prepared to
   receive media with both the old and new types until the answer is
   received, and media with the new type is received and reaches the top
   of the playout buffer.
```
The RFC only mentions media type change if `same logical data is conveyed`, such as voiceband fax and fax. Thus, drastic media change, such as audio to video or vice versa, although not illegal, should never happen in practice, especially due to requirement such as `The offerer MUST be prepared to receive media with both the old and new types`, which means that it has to maintain both media types during the change.

And even if it does happen, the mapping should stay the same, IMO, since there's chance that it may change back again (med_type1->type2->type1)

In conclusion, IMO, we don't need any special handling for media type change.

Removal of media stream, however, will require the map to be reset, since the RFC says:
```
in the
   case of RTP, the mapping from a particular dynamic payload type
   number to a particular codec within that media stream MUST NOT change
   for the duration of a session
```
If media stream is removed, it no longer exists, so the mapping should also no longer exist.

The PR turns out to be quite a major and complex task, so I will continue the discussion about this in the main PR page.

sauwming:
The PR turns out to be quite a major and complex task, since maintaining mapping requires handling many non-trivial scenario.

So the question we are facing now is whether:
1. we are confident to proceed with this without introducing bugs or backward regression, or
2. should we adopt a simpler approach or workaround instead, for now? (for example, by not reverting PT to pjmedia defaults, although haven't checked how difficult this is)

Any ideas or thoughts?

trengginas:
> The PR turns out to be quite a major and complex task, since maintaining mapping requires handling many non-trivial scenario.
> 
> So the question we are facing now is whether:
> 
> 1. we are confident to proceed with this without introducing bugs or backward regression, or
> 2. should we adopt a simpler approach or workaround instead, for now? (for example, by not reverting PT to pjmedia defaults, although haven't checked how difficult this is)
> 
> Any ideas or thoughts?

Prefer the second option:
- Add option to send active SDP when sending re-INVITE/UPDATE. We can send active local SDP (pjmedia_sdp_neg_get_active_local()) instead of using pjmedia defaults.

sauwming:
After further research, I found that @nanangizz is correct. The mapping is for one side:
```
there must be a separate mapping table for each endpoint. But once you have established a
mapping between a particular codec and a payload number for an endpoint, you must not
associate a different codec with that payload number *for that endpoint*.
```
https://lists.cs.columbia.edu/pipermail/sip-implementors/2002-October/003811.html

Although I'm not sure if this can make the implementation less complicated than it currently is.

There's also the problem that one-side mapping won't fix the reported original problem, and may force asymmetric PT to be used.

nanangizz:
Sending active local SDP may only be done for session refresh scenario only (perhaps it is already). While most other re-INVITE scenarios, using active local SDP may not be possible (or complicated), e.g:
- IIRC when receives INVITE without SDP, we should send offer with all capabilities.
- re-INVITE is done to update some codec params (e.g: change mode/bitrate/etc).
- re-INVITE with reset media for example in IP change scenario.

IMHO the current patch should be close to working? To reduce risks, add python/sipp tests which cover most/all scenarios? We just need to agree whether the map should be per stream or per media type (and so when to reset). Basically I am okay with either approach :)

sauwming:
Okay, so let's clarify the spec:
* Mapping application: as per the RFC, i.e. per media stream. Reset upon media stream removal
* Scope of mapping (which one?):
  a. for our endpoint only
      pro: simpler? cons: - may force remote to use asymmetric PT, - still causes issue with remote that interprets mapping should be for both endpoints (i.e. it doesn't fix the original reporter's issue)
  b. for both endpoints
      con: - increase in mapping size, more likely to run out of PT number

There's also an issue with the additional API proposed, but let's discuss one by one first, starting from the spec.

pjmedia/src/pjmedia/sdp_neg.c
trengginas: If we are using per stream mapping, then this should not be changed

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1623,6 +1955,11 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_negotiate( pj_pool_t *pool,
                                neg->neg_local_sdp, neg->neg_remote_sdp,
                                &answer);
         if (status == PJ_SUCCESS) {
+
+#if !PJMEDIA_SDP_NEG_ANSWER_SYMMETRIC_PT
+            validate_pt(pool, neg, answer);
```
sauwming, line 1960: Should we prioritise answering with symmetric PT, or making sure that the mapping is valid?

IMO, we should try to answer with symmetric PT, but if it causes the mapping to be invalid, then we have no choice but use asymmetric PT.

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1545,6 +1573,304 @@ static pj_status_t create_answer( pj_pool_t *pool,
...
+                med_pt_map->map[med_pt_map->map_cnt].pt = pt;
+                pj_strdup(pool, &med_pt_map->map[med_pt_map->map_cnt++].codec, &codec);
+            }
+        }
+    }
+    neg->med_pt_map_cnt = med_cnt;
+}
+
+/*
+ * The pt map is maintain using temporary/active pool. This is to avoid memory usage
```
sauwming, line 1684: This is a good feature, but it seems that all the codec string duplication can be quite inefficient.

Since the number of codec strings should be quite constant (especially if we only record our endpoint), i.e. it will only include all the codecs we support and won't grow beyond that, perhaps we can just store the codec strings using the main pool.

sauwming (commented):
Nice, this is a good progress.

sauwming:
Optional features:
1. Populate the codec string list at the beginning
Use the APIs `pjmedia_codec_mgr_enum_codecs()` and `pjmedia_codec_info_to_id()`.
3. Sort the codec list
If the codec list is sorted, you can do any search in `O(log n)`.
4. Move the codec list outside SDP neg
Since the codec list is basically identical, perhaps the list can be made static, or owned by another component (who? codec manager?)

IMO, the above features are optional, since they only offer benefit if there are many codecs supported and/or there are a large number of calls. Let me know if you think any/all of them should be implemented now.

Necessary improvement:
- Utilise the **_index_** of the codec string, by changing the mapping to `pj_int8_t map[DYNAMIC_PT_SIZE]`
This should make searching for any pt and any codec much more efficient. PT search will be O(1) since we don't need to call `pt_exists()`, if `map[payload number] == valid_string_idx`, the mapping between the PT number and the codec in that idx exists.

trengginas:
Basically I agree with the optional feature, since it will more efficient.
If the codec string is maintain in codec manager or anything else than SDP neg, then it should be passed to SDP neg to be able to use it. e.g: add API to supply the codec string to the SDP neg.
This will also affect users that uses SDP neg independently. If no codec string is supplied, then there should not be any behavioral change.

sauwming:
Alright, please go ahead with adding the optional features.
Yes, a new API is needed to retrieve the codec strings.

pjmedia/src/pjmedia/vid_codec.c
```diff
@@ -218,6 +222,8 @@ PJ_DEF(pj_status_t) pjmedia_vid_codec_mgr_register_factory(
         pjmedia_vid_codec_info_to_id( &info[i],
                                   mgr->codec_desc[mgr->codec_cnt+i].id,
                                   sizeof(pjmedia_codec_id));
+        pj_strdup2_with_null(mgr->pool, &mgr->codec_list[mgr->codec_cnt+i],
```
sauwming, line 225: No need to duplicate? Just point to `mgr->codec_desc[mgr->codec_cnt+i].id`.
This way, - we don't need pool, and - later during factory unregistration, we can also erase the codec string from the list.

pjmedia/src/pjmedia/vid_codec.c
```diff
@@ -755,6 +761,19 @@ PJ_DEF(pj_status_t) pjmedia_vid_codec_mgr_set_default_param(
     return PJ_SUCCESS;
 }
 
+PJ_DEF(pj_int8_t) pjmedia_vid_codec_mgr_get_codec_ids(pjmedia_vid_codec_mgr* mgr,
+                                                      pj_int8_t codec_cnt,
+                                                      pj_str_t codec_ids[])
+{
+    if (!mgr) mgr = def_vid_codec_mgr;
+    PJ_ASSERT_RETURN(mgr, PJ_EINVAL);
+
```
sauwming, line 770: need to lock `mgr->mutex`?

pjmedia/src/pjmedia/vid_codec.c
sauwming: I just realise that if we don't duplicate the string, the list obtained by sdp_neg can point to an invalid string if the codec has been unregistered. Unregistering a codec should be quite an uncommon use case, but still, I guess we have no choice but use the pool here.

pjmedia/include/pjmedia/codec.h
```diff
@@ -722,6 +722,9 @@ typedef struct pjmedia_codec_mgr
     /** Array of codec descriptor. */
     struct pjmedia_codec_desc    codec_desc[PJMEDIA_CODEC_MGR_MAX_CODECS];
 
+    /** Array of codec id. */
+    pj_str_t                     codec_list[PJMEDIA_CODEC_MGR_MAX_CODECS];
```
sauwming, line 726: since codec desc can be removed, make sure to handle the case that the codec_desc and codec_list have different number of elements

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -84,6 +120,100 @@ static pj_status_t custom_fmt_match( pj_pool_t *pool,
...
+
+    neg->pool = pool;
+    neg->codec_list_cnt = 0;
+
+    /* Get the audio/video codec, and sort them. */
+    get_audio_codec_id(neg);
+#if defined(PJMEDIA_HAS_VIDEO) && (PJMEDIA_HAS_VIDEO != 0)
+    get_video_codec_id(neg);
+#endif
+    sort_codec_id(neg->codec_list, neg->codec_list_cnt);
```
sauwming, line 189: separating the audio/video codec strings into different array should be better, so that we don't need to re-sort them.

pjmedia/src/pjmedia/codec.c
```diff
@@ -191,6 +249,8 @@ PJ_DEF(pj_status_t) pjmedia_codec_mgr_register_factory( pjmedia_codec_mgr *mgr,
     /* Re-sort codec based on priorities */
     sort_codecs(mgr);
 
+    sort_codec_id(mgr->codec_list, mgr->codec_list_cnt);
```
sauwming, line 252: just keep the array always sorted during insertion/deletion, so that we don't need to sort it over and over.

sauwming:
Since the basic features seem to be mostly done, when you're back, can you provide one or two test cases to show the behaviour? (i.e. the results before the patch, compared to with the patch (default setting) and with the patch (if `PJMEDIA_SDP_NEG_MAINTAIN_REMOTE_PT_MAP` is enabled)

trengginas:
Test configuration : (Below scenario will use the same test configuration)
- caller : disable all codec besides iLBC
- callee : enable all codec
- callee will send re-INVITE sith `aud_cnt` = 2 and `PJSUA_CALL_REINIT_MEDIA`

Scenario 1 :
- No patch
- `PJMEDIA_SDP_NEG_ANSWER_SYMMETRIC_PT` 1

Scenario 2 :
- with patch
- `PJMEDIA_SDP_NEG_ANSWER_SYMMETRIC_PT` 0
- `PJMEDIA_SDP_NEG_MAINTAIN_REMOTE_PT_MAP` 0

Scenario 3 :
- with patch
- `PJMEDIA_SDP_NEG_ANSWER_SYMMETRIC_PT` 0
- `PJMEDIA_SDP_NEG_MAINTAIN_REMOTE_PT_MAP` 1

[scenario3_caller.log](https://github.com/pjsip/pjproject/files/11306484/scenario3_caller.log)
[scenario2_caller.log](https://github.com/pjsip/pjproject/files/11306485/scenario2_caller.log)
[scenario1_caller.log](https://github.com/pjsip/pjproject/files/11306486/scenario1_caller.log)

sauwming:
The logs showed multiple PTs being used, such as:
```
a=rtpmap:98 speex/8000
a=rtpmap:98 speex/32000
a=rtpmap:96 iLBC/8000
a=fmtp:96 mode=30
a=rtpmap:98 opus/48000/2
a=fmtp:98 useinbandfec=1
```
so I tried to fix it in the latest commit.

In my test here, it seems to work quite well:
[log-original.txt](https://github.com/pjsip/pjproject/files/11349845/log-original.txt)
[log-remote_map_disabled.txt](https://github.com/pjsip/pjproject/files/11349846/log-remote_map_disabled.txt)
[log-remote_map_enabled.txt](https://github.com/pjsip/pjproject/files/11349847/log-remote_map_enabled.txt)

Also, please help check if all the locations where I call `assign_pt_and_update_map()` are already correct.

sauwming (approved):

nanangizz (approved):

PR #3473, opened 2023-03-30
Fix PJSUA2 documentation

wosrediinanatour:
This commit fixes copy&past errors from PJSUA, uses enums instead of magic numbers, and some minors.

### DIFF
--- a/pjsip/include/pjsua2/call.hpp
+++ b/pjsip/include/pjsua2/call.hpp
@@ -1881,7 +1881,7 @@ public:
 
     /**
      * Notify application when an audio media session is about to be created
-     * (as opposed to onCallSdpCreated(), which is called *after* the session
+     * (as opposed to onStreamCreated(), which is called *after* the session
      * has been created). The application may change
      * some stream info parameter values, i.e: jbInit, jbMinPre, jbMaxPre,
      * jbMax, useKa, rtcpSdesByeDisabled, jbDiscardAlgo (audio),
--- a/pjsip/include/pjsua2/call.hpp
+++ b/pjsip/include/pjsua2/call.hpp
@@ -788,11 +788,11 @@ struct OnStreamCreatedParam
     
     /**
      * Specify if PJSUA2 should take ownership of the port returned in
-     * the pPort parameter below. If set to PJ_TRUE,
+     * the pPort parameter below. If set to true,
      * pjmedia_port_destroy() will be called on the port when it is
      * no longer needed.
      *
-     * Default: PJ_FALSE
+     * Default: false
      */
     bool        destroyPort;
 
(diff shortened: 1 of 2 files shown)


pjsip/include/pjsua2/call.hpp
```diff
@@ -1051,8 +1051,8 @@ struct OnCallRxReinviteParam
     
     /**
      * Status code to be returned for answering the offer. On input,
-     * it contains status code 200. Currently, valid values are only
-     * 200 and 488.
+     * it contains status code PJSIP_SC_OK. Currently, valid values are only
+     * PJSIP_SC_OK and PJSIP_SC_NOT_ACCEPTABLE_HERE.
```
sauwming, line 1055: Perhaps it's just me, having worked with SIP for so long, the number sticks more to me than the text. So, I'd suggest not to remove the number, i.e.
`PJSIP_SC_OK (200)` and `PJSIP_SC_NOT_ACCEPTABLE_HERE (488)`.

Also for other instances above and below.

pjsip/include/pjsua2/call.hpp
```diff
@@ -2014,27 +2014,25 @@ class Call
      *
      * @param prm       Callback parameter.
      */
-    virtual void onCallReplaced(OnCallReplacedParam &prm)
-    { PJ_UNUSED_ARG(prm); }
+  virtual void onCallReplaced(OnCallReplacedParam &prm) { PJ_UNUSED_ARG(prm); }
```
sauwming, line 2017: why does this have to be squeezed into one line?

pjsip/include/pjsua2/call.hpp
```diff
@@ -1512,7 +1512,7 @@ class Call
      * which means for incoming call, this function can be called as soon as
      * call is received as long as incoming call contains SDP, and for outgoing
      * call, this function can be called only after SDP is received (normally in
-     * 200/OK response to INVITE). As a general case, application should call
+     * PJSIP_SC_OK/OK response to INVITE). As a general case, application should call
```
sauwming, line 1515: for this case, 200/OK is already correct. no modification needed.

pjsip/include/pjsua2/call.hpp
wosrediinanatour: This was done by accident.

wosrediinanatour:
Updated commit due to your feedback. Please take a look.

pjsip/include/pjsua2/call.hpp
```diff
@@ -1882,8 +1881,8 @@ class Call
 
     /**
      * Notify application when an audio media session is about to be created
-     * (as opposed to on_stream_created() and on_stream_created2() which are
-     * called *after* the session has been created). The application may change
+     * (as opposed to onCallSdpCreated(), which is called *after* the session
```
sauwming, line 1884: should be `onStreamCreated()`

sauwming (approved):

nanangizz (approved):

PR #3475, opened 2023-03-31
Add Reason/Warning header in disconnecting call

nanangizz:
This PR will add Reason header to BYE/CANCEL requests (rfc3326) or Warning header to error responses (rfc3261 20.43). SIP response message already has a status/reason text, but sometimes additional info can be useful for debugging/troubleshooting.

Thanks Taisto Qvist for the suggestions.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -197,27 +197,30 @@ static pj_status_t add_reason_warning_hdr(pjsip_tx_data *tdata,
                                           const char *str)
 {
     pj_str_t reason;
+
+    PJ_ASSERT_RETURN(tdata && str, PJ_EINVAL);
     pj_cstr(&reason, str);
+
     if (tdata->msg->type == PJSIP_REQUEST_MSG) {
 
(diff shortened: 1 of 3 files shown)


pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -181,6 +188,69 @@ struct tsx_inv_data
...
+
+        /* Add Warning header in response messages */
+        pjsip_warning_hdr *hdr;
+        pjsip_dialog *dlg = pjsip_tdata_get_dlg(tdata);
+
+        PJ_ASSERT_RETURN(dlg, PJ_EINVAL);
+
+        // Let's always use code 390 for now.
+        //PJ_ASSERT_RETURN(code >= 300 && code < 400, PJ_EINVAL);
+        code = 390;
```
sauwming, line 238: Since code and warn-code are actually different, perhaps better to change this to:
```
        // Let's always use code 390 for now.
        // TODO: unsigned warn_code = map_code_to_warn_code(code);
        unsigned warn_code = 390;
```

sauwming (approved):

trengginas (approved):

PR #3478, opened 2023-04-03
Fixed max hostname default value and usage

### LINKED ISSUE #3450: PJ_HOST_NAME should be clarified or use 253 +1
### Describe the bug

Hi,

Does PJ_HOST_NAME include null char ?

If so it should be push to 254, if not you should allocate array of PJ_HOST_NAME +1

### Steps to reproduce

See souce

### PJSIP version

alll

### Context

no context source

### Log, call stack, etc

```shell
Source only
```

sauwming:
To fix #3450.
Continuation of #3438.

It's not obvious whether a valid hostname length is strictly < `PJ_MAX_HOSTNAME` or <= `PJ_MAX_HOSTNAME`.

In some parts, such as https://github.com/pjsip/pjproject/blob/5ac9104514499d648f68991ef796368c51b4dfec/pjlib/src/pj/addr_resolv_sock.c#L38
we seem to assume the first one to be true, however, in other parts, we seem to assume the latter:
https://github.com/pjsip/pjproject/blob/5ac9104514499d648f68991ef796368c51b4dfec/pjlib-util/src/pjlib-util/resolver.c#L806

So here we clarify the specification in the doc and make the usage consistent.

### DIFF
--- a/pjlib-util/src/pjlib-util/resolver.c
+++ b/pjlib-util/src/pjlib-util/resolver.c
@@ -803,7 +803,7 @@ static void init_res_key(struct res_key *key, int type, const pj_str_t *name)
     key->qtype = (pj_uint16_t)type;
 
     len = name->slen;
-    if (len > PJ_MAX_HOSTNAME) len = PJ_MAX_HOSTNAME;
+    if (len >= PJ_MAX_HOSTNAME) len = PJ_MAX_HOSTNAME - 1;
 
     /* Copy key, in lowercase */
     for (i=0; i<len; ++i) {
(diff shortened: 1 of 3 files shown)


nanangizz (approved):
From [here](https://devblogs.microsoft.com/oldnewthing/20120412-00/?p=7873), the limit should be 253, without NULL termination char. So IMO clarifying this in the doc (i.e: 253 without NULL termination or 254 with NULL termination) is indeed better.

While for `init_res_key()`, IMO it should be fine to be not always NULL terminated, as the usage is for hash table key which the key is actually octets instead of string, i.e: the whole `res_key` struct (instead of just the array of char field). But I believe this patch should not cause any issue for the hash.

trengginas (approved):

PR #3487, opened 2023-04-05
Fixed coverity warnings

sauwming:
* Fixed `warning C4002: too many actual parameters for macro 'PJ_PRINT_FUNC_DECOR'` for non-GNU and non-clang compilers, such as MSVC.
* Fixed reason setting in `sip_inv` (#3475), since variable status has been overwritten.


nanangizz (commented):
Is this PR supposed to fix those two only or also any other coverity warnings in the future?

sauwming:
Just these two (unless a new defect pops up before it's merged, then I'll probably combine it).

I want to change the PR title to be more specific, but I can't find one that's brief and can accommodate both warnings.

nanangizz (approved):

trengginas (approved):

PR #3496, opened 2023-04-06
Fixed build failure if PJ_JNI_HAS_JNI_ONLOAD is disabled

### LINKED ISSUE #3495: pj_jni_jvm undeclared if PJ_JNI_HAS_JNI_ONLOAD = 0
### Describe the bug

If ```PJ_JNI_HAS_JNI_ONLOAD != 0```, ```pj_jni_jvm``` is declared and then initialized in the ```JNI_OnLoad``` function.
[os_core_unix.c#L63](https://github.com/pjsip/pjproject/blob/master/pjlib/src/pj/os_core_unix.c#L63)

In our case we force ```PJ_JNI_HAS_JNI_ONLOAD``` to 0 at compile time.

At this point in [os_core_unix.c#L318](https://github.com/pjsip/pjproject/blob/master/pjlib/src/pj/os_core_unix.c#L318), the check for ```PJ_ANDROID``` is not enough because the functions will use the ```pj_jni_jvm``` variable that can be not declared.

### Steps to reproduce

Prepare the compile system for andorid.
Compile with the ```-DPJ_JNI_HAS_JNI_ONLOAD=0``` flag.

### PJSIP version

2.12.1, 2.13

### Context

```config_site.h```
```#define PJ_CONFIG_ANDROID 1```

Compile with the ```-DPJ_JNI_HAS_JNI_ONLOAD=0``` flag.

### Log, call stack, etc

```shell
os_core_unix.c:323:11: error: use of undeclared identifier 'pj_jni_jvm'
os_core_unix.c:323:31: error: use of undeclared identifier 'pj_jni_jvm'
os_core_unix.c:326:15: error: use of undeclared identifier 'pj_jni_jvm'
os_core_unix.c:326:48: error: use of undeclared identifier 'pj_jni_jvm'
os_core_unix.c:340:11: error: use of undeclared identifier 'pj_jni_jvm'
os_core_unix.c:340:44: error: use of undeclared identifier 'pj_jni_jvm'
```

sauwming:
To fix #3495.

This PR will also add a new API `pj_jni_set_jvm()` which you need to manually call if you disable `PJ_JNI_HAS_JNI_ONLOAD`. This is necessary in order to ensure that components which use JNI, such as `guid_android`, `Android JNI` audio or `Oboe`, and `android_dev` video device, to still work properly.

The API `pj_jni_attach/detach_jvm()` as well as `PJ_JNI_HAS_JNI_ONLOAD` compile-time setting will also be made public as well.

### DIFF
--- a/pjlib/include/pj/os.h
+++ b/pjlib/include/pj/os.h
@@ -387,6 +387,28 @@ PJ_DECL(pj_status_t) pj_thread_get_stack_info(pj_thread_t *thread,
 
 /* **************************************************************************/
 /**
+ * @defgroup PJ_ANDROID_OS Android OS Specific
+ * @ingroup PJ_OS
+ * @{
+ * Functionalities specific to Android OS.
+ *
+ */
+
+/**
+ * Set the Java Virtual Machine environment variable for JNI purposes.
+ * Note that applications typically do not need to call this function unless
+ * PJ_JNI_HAS_JNI_ONLOAD is disabled.
+ *
+ * @param jvm           The Java Virtual Machine environment.
+ */
+PJ_DECL(void) pj_android_jni_set_jvm(void *jvm);
+
+/**
+ * @}
+ */
+
+/* **************************************************************************/
+/**
  * @defgroup PJ_SYMBIAN_OS Symbian OS Specific
  * @ingroup PJ_OS
  * @{
--- a/pjlib/include/pj/os.h
+++ b/pjlib/include/pj/os.h
@@ -387,21 +387,42 @@ PJ_DECL(pj_status_t) pj_thread_get_stack_info(pj_thread_t *thread,
 
 /* **************************************************************************/
 /**
- * @defgroup PJ_ANDROID_OS Android OS Specific
+ * @defgroup PJ_JNI Java Native Interface specific
  * @ingroup PJ_OS
  * @{
- * Functionalities specific to Android OS.
+ * Functionalities specific to JNI.
+ * Currently only implemented on Android OS, but may be extended to other
+ * platforms in the future.
  *
  */
 
 /**
- * Set the Java Virtual Machine environment variable for JNI purposes.
+ * Set the Java Virtual Machine environment variable.
  * Note that applications typically do not need to call this function unless
  * PJ_JNI_HAS_JNI_ONLOAD is disabled.
  *
  * @param jvm           The Java Virtual Machine environment.
  */
-PJ_DECL(void) pj_android_jni_set_jvm(void *jvm);
+PJ_DECL(void) pj_jni_set_jvm(void *jvm);
+
+/**
+ * Attach the current thread to a Java Virtual Machine.
+ *
+ * @param jni_env       Output parameter to store the JNI interface pointer.
+ *
+ * @return              PJ_TRUE if the attachment is successful,
+ *                      PJ_FALSE if otherwise.
+ */
+PJ_DECL(pj_bool_t) pj_jni_attach_jvm(void **jni_env);
+
+/**
+ * Detach the current thread from a Java Virtual Machine.
+ *
+ * @param attached      Specify whether the current thread is attached
+ *                      to a JVM.
+ */
+PJ_DECL(void) pj_jni_detach_jvm(pj_bool_t attached);
+
 
 /**
  * @}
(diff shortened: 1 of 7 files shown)


pjlib/src/pj/os_core_unix.c
```diff
@@ -315,7 +315,7 @@ PJ_DEF(pj_bool_t) pj_thread_is_registered(void)
  * A cool solution would be to port (if possible) the code from the
  * android os regarding set_sched groups.
  */
-#if PJ_ANDROID
+#if defined(PJ_JNI_HAS_JNI_ONLOAD) && PJ_JNI_HAS_JNI_ONLOAD != 0
```
nanangizz, line 318: - The function `set_android_thread_priority()` defined below should be Android specific, also in `pj_thread_set_prio()` below, it is called under `PJ_ANDROID` protection, so it needs `PJ_ANDROID` protection. But then it also calls `pj_jni_attach_jvm()` which needs `PJ_JNI_HAS_JNI_ONLOAD` protection. So perhaps it needs both protections?
- Perhaps we need to make sure any other parts in the lib invoking `pj_jni_attach_jvm()` (if any, the `PJ_DEF` and the naming indicates it is a public API) are protected by `PJ_JNI_HAS_JNI_ONLOAD` too, instead of `PJ_ANDROID`. Or alternatively, define a dummy/empty function version when `PJ_JNI_HAS_JNI_ONLOAD` is set to zero.

sauwming: Done. I have provided alternative/fallback implementation for thread priority setting and guid. Now it can build and link successfully.

The following components have to be specifically disabled (refer to the updated PR desc):
Android JNI, Oboe, and android_dev
The question is should we automatically disable them (by adding `PJ_JNI_HAS_JNI_ONLOAD` protection) or leave it to the user to disable them?
I prefer the latter option so that user is aware that those components won't be available.

nanangizz: So it seems that `PJ_JNI_HAS_JNI_ONLOAD` is currently used by Android features only (while it sounds like a generic JNI thing), perhaps that's why it is set to `PJ_ANDROID`.

IMO the only reason `PJ_JNI_HAS_JNI_ONLOAD` not set is that there is already `JNI_OnLoad` implemented somewhere else where PJSIP is integrated to it as a single dynamic library. In this case, perhaps PJSIP should provide a public API to set its JavaVM/`pj_jni_jvm` as alternative to getting JavaVM via `JNI_OnLoad` (note that there can only be one `JNI_OnLoad` in a library loaded from Java).

pjlib/include/pj/os.h
```diff
@@ -381,6 +381,28 @@ PJ_DECL(pj_status_t) pj_thread_get_stack_info(pj_thread_t *thread,
...
+ */
+
+/**
+ * Set the Java Virtual Machine environment variable for JNI purposes.
+ * Note that applications typically do not need to call this function unless
+ * PJ_JNI_HAS_JNI_ONLOAD is disabled.
+ *
+ * @param jvm           The Java Virtual Machine environment.
+ */
+PJ_DECL(void) pj_android_jni_set_jvm(void *jvm);
```
nanangizz, line 404: Why use `pj_android_jni` prefix instead of `pj_jni`, as basically it is generic JNI (?), so it will be in the same group as `pj_jni_attach/detach_jvm()`.

pjlib/src/pj/os_core_unix.c
```diff
@@ -315,18 +314,26 @@ PJ_DEF(pj_bool_t) pj_thread_is_registered(void)
...
-#if PJ_ANDROID
+#if defined(PJ_ANDROID) && PJ_ANDROID != 0
 
 #include <jni.h>
 #include <sys/resource.h>
 #include <sys/system_properties.h>
 
-PJ_DEF(pj_bool_t) pj_jni_attach_jvm(JNIEnv **jni_env)
+/* If you disable PJ_JNI_HAS_JNI_ONLOAD, set the JVM using this function. */
+PJ_DEF(void) pj_android_jni_set_jvm(void *jvm)
```
nanangizz, line 324: Should this be `pj_jni_set_jvm()`?

nanangizz (approved):
LGTM, just a minor below. Also please make sure that the original issue, i.e: build failure when `PJ_JNI_HAS_JNI_ONLOAD==0`, is fixed.

trengginas (approved):

PR #3500, opened 2023-04-10
Change socket assertion check to allow socket handle value 0

trengginas:
Currently there is an assertion to check socket handle validity:
```c
PJ_ASSERT_RETURN(sock!=0 && sock!=PJ_INVALID_SOCKET, PJ_EINVAL);
```
However socket handle itself might have 0 value: (-1 indicates error)
- https://man7.org/linux/man-pages/man2/socket.2.html

### DIFF
--- a/pjlib/src/pj/sock_qos_symbian.cpp
+++ b/pjlib/src/pj/sock_qos_symbian.cpp
@@ -21,7 +21,7 @@
 PJ_DEF(pj_status_t) pj_sock_set_qos_params(pj_sock_t sock,
                                            pj_qos_params *param)
 {
-    PJ_ASSERT_RETURN(sock>=0 && sock!=PJ_INVALID_SOCKET, PJ_EINVAL);
+    PJ_ASSERT_RETURN(sock!=0 && sock!=PJ_INVALID_SOCKET, PJ_EINVAL);
     
     CPjSocket *pjsock = (CPjSocket*)sock;
     RSocket & rsock = pjsock->Socket();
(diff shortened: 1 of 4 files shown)


pjlib/src/pj/sock_qos_symbian.cpp
```diff
@@ -21,7 +21,7 @@
 PJ_DEF(pj_status_t) pj_sock_set_qos_params(pj_sock_t sock,
                                            pj_qos_params *param)
 {
-    PJ_ASSERT_RETURN(sock!=0 && sock!=PJ_INVALID_SOCKET, PJ_EINVAL);
+    PJ_ASSERT_RETURN(sock>=0 && sock!=PJ_INVALID_SOCKET, PJ_EINVAL);
```
nanangizz, line 24: - Better not change anything here in Symbian as it is no longer supported/maintained platform?
- Zero/NULL sock check in Symbian seems to be correct already as `pj_sock_t` is an object (instead of a simple integer file descriptor like in *nix/Win platforms).

sauwming:
`
In Unix-like systems, file descriptors 0, 1, and 2 are reserved for standard input (stdin), standard output (stdout), and standard error (stderr), respectively. However, it is important to note that the operating system can potentially close or reassign standard input, output, or error file descriptors, and in such a case, socket() could return 0 as the socket descriptor. While this is not a common occurrence, it is possible.
`

There are also instances in the library where we don't close the socket if it's zero, for example in `sip_transport_udp.c`:
```
        /* Close socket. */
        if (tp->sock && tp->sock != PJ_INVALID_SOCKET) {
            pj_sock_close(tp->sock);
            tp->sock = PJ_INVALID_SOCKET;
        }
```
Should we fix this too?

sauwming (approved):

nanangizz (approved):

Issue #3515, opened 2023-04-21
Do not run cron workflow (Coverity Scan) in forks


jimying:
### Describe the bug

I find that daily received an email 
"[jimying/pjproject] Run failed: Coverity Scan (Daily) - master"

### Steps to reproduce

-

### PJSIP version

2.13-dev

### Context

-

### Log, call stack, etc

-

PR #3516, opened 2023-04-21
Do not run cron workflow( Coverity Scan) in forks

jimying:
as title  #3515 

[doc, Example: Only run job for specific repository](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository)


sauwming (commented):
I understand your concern for Coverity, but why CodeQL as well? As far as I know, it won't cause any harm? And it will help you if you inadvertently create a bug that can be found by CodeQL.

nanangizz (approved):

PR #3523, opened 2023-04-24
Fix PJ_CC_NAME/VER incorrec when use clang compiler to build

jimying:
as title


sauwming:
Without the patch:
`PJ_CC_NAME/VER_(1,2,3)    : gcc-4.2.1`

With the patch:
`PJ_CC_NAME/VER_(1,2,3)    : clang-15.0.7`

sauwming (approved):

nanangizz (approved):

PR #3530, opened 2023-04-26
Fixed PIDF XML processing if nodes are namespaced

### LINKED ISSUE #1150: Fail to process PIDF XML document if the nodes are namespaced (thanks Johan Lantz for the report)
_2010-10-21 11:21:51: @bennylp_ _created the issue on trac ticket [1150](https://trac.pjsip.org/repos/ticket/1150)_

The presence module fails to process PIDF XML document if the XML nodes contains XML namespaces. For example:

```
<ep:presence xmlns:ep="urn:ietf:params:xml:ns:pidf" 
             xmlns:op="urn:oma:xml:prs:pidf:oma-pres" 
             xmlns:rt="urn:ietf:params:xml:ns:pidf:rpid" 
             xmlns:gp="urn:ietf:params:xml:ns:pidf:geopriv10" 
             xmlns:pdm="urn:ietf:params:xml:ns:pidf:data-model" 
             xmlns:rs="urn:ietf:params:xml:ns:pidf:status:rpid-status" 
             entity="tel:+34630188891">
  <ep:tuple id="a123">
    <ep:status>
      <ep:basic>open</ep:basic>
    </ep:status>
  </ep:tuple>
</ep:presence>
```

___
_2010-10-21 11:23:12: @bennylp_ _changed **milestone** from **release-1.8.10** to **Known-Issues-and-Ideas**_

___
_2010-10-21 11:23:12: @bennylp_ _commented_

Moving to known-issues (should have been created there in the first place)

sauwming:
To close #1150.

Note that the proposed fix here utilizes the same method as the one used in: https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip-simple/pidf.c#L348
```
PJ_DEF(pjpidf_pres*) pjpidf_parse(pj_pool_t *pool, char *text, int len)
{
    pjpidf_pres *pres = pj_xml_parse(pool, text, len);
    if (pres && pres->name.slen >= 8) {
        pj_str_t name;

        name.ptr = pres->name.ptr + (pres->name.slen - 8);
        name.slen = 8;

        if (pj_stricmp(&name, &PRESENCE) == 0)
            return pres;
    }
```
i.e. it will do the string comparison from the back and assumes that everything in front is the namespace, or prefix.

A proper solution would be to parse the namespace itself, but since the feature is never requested, let me know if you think it would be better for us to leave the current master unchanged and mark the issue as `not planned/won't fix`.

### DIFF
--- a/pjlib-util/src/pjlib-util/xml.c
+++ b/pjlib-util/src/pjlib-util/xml.c
@@ -358,6 +358,15 @@ PJ_DEF(void) pj_xml_add_attr( pj_xml_node *node, pj_xml_attr *attr )
     pj_list_push_back(&node->attr_head, attr);
 }
 
+static int node_name_cmp(pj_str_t *node_name, pj_str_t *name)
+{
+    pj_str_t node;
+
+    node.ptr = node_name->ptr + (node_name->slen - name->slen);
+    node.slen = name->slen;
+    return pj_stricmp(&node, name);
+}
+
 PJ_DEF(pj_xml_node*) pj_xml_find_node(const pj_xml_node *parent, 
                                       const pj_str_t *name)
 {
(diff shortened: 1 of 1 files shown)


sauwming:
No follow up, closing this.

PR #3536, opened 2023-04-27
update pjsua_media.c is_media_changed

jhting:
update pjsua_media.c is_media_changed Compare codec info bug:old_si->tx_event_pt != new_si->rx_event_pt


wosrediinanatour:
Why is this needed?

sauwming (approved):

sauwming:
> Why is this needed?

There seems to be a typo in our media change detection logic, which is needed to determine whether during a reinvite for example, if we need to tear down media and reinit it, or just leave the media unchanged.

nanangizz (approved):

PR #3538, opened 2023-04-27
Fixed incorrect RTCP stat calculation for G722

### LINKED ISSUE #1198: Wrong RTCP loss period calculation with multiple frames per RTP packet
_2011-02-17 14:51:01: @bennylp_ _created the issue on trac ticket [1198](https://trac.pjsip.org/repos/ticket/1198)_

The calculation in `rtcp.c:332`:

```
   period = count * sess->pkt_size * 1000 / sess->clock_rate;
   period *= 1000;
```

where

```
  sess->pkt_size = settings->samples_per_frame;
```

Related to this is the hardcodec G.722 samples_per_frame value in `stream.c:2221` :

```
#if defined(PJMEDIA_HANDLE_G722_MPEG_BUG) && (PJMEDIA_HANDLE_G722_MPEG_BUG!=0)
	/* Special case for G.722 */
	if (info->fmt.pt == PJMEDIA_RTP_PT_G722) {
	    rtcp_setting.clock_rate = 8000;
	    rtcp_setting.samples_per_frame = 160;
	}
#endif
```

sauwming:
To fix #1198

From the description in #1198, it's not obvious whether the issue occurs in general, or only specific for G722.
The title `Wrong RTCP loss period calculation with multiple frames per RTP packet` seems to suggest this is a general problem, however the calculation in `stream.c` seems to be already correct:
```
    afd->frame_time_usec = stream->codec_param.info.frm_ptime *
                           stream->codec_param.setting.frm_per_pkt * 1000;
...
    rtcp_setting.samples_per_frame = PJMEDIA_AFD_SPF(afd);
```
i.e. it already takes into account `frm_per_pkt`. It's true that the name is incorrect, it should be samples_per_packet, as it seems to be the case with `stream->enc_samples_per_pkt = PJMEDIA_AFD_SPF(afd);`.

But despite the misleading name,  the calculation looks right when I tested it here with multiple frames per pkt. For `ptime=100` and `clock_rate=48KHz`, here are the numbers:
```
frm_per_pkt = 5
sess_pkt_size = 4800
period/count = sess->pkt_size * 1000 / sess->clock_rate =100
```

If it's only for G722 though, it shouldn't be specific for multiple frames per packet, since the `clock rate` and `samples_per_frame` forced modification should also affect single frame per packet.

Conclusion:
- there's no issue with the general codec
- since it only affects G722 and not critical, let me know if it's better to merge the PR (quite low risk anyway) or we should mark the issue as `won't fix`.

### DIFF
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -2746,10 +2746,15 @@ PJ_DEF(pj_status_t) pjmedia_stream_create( pjmedia_endpt *endpt,
 
 #if defined(PJMEDIA_HANDLE_G722_MPEG_BUG) && (PJMEDIA_HANDLE_G722_MPEG_BUG!=0)
         /* Special case for G.722 */
+        /* Update: We don't need to do this as this causes miscalculation
+         * inside RTCP.
+         */
+        /*
         if (info->fmt.pt == PJMEDIA_RTP_PT_G722) {
             rtcp_setting.clock_rate = 8000;
             rtcp_setting.samples_per_frame = 160;
         }
+        */
 #endif
 
         pjmedia_rtcp_init2(&stream->rtcp, &rtcp_setting);


nanangizz:
I believe it is generic for all codecs. Note that the loss duration is calculated for both directions. For outgoing direction we can control the packet size because the stream will follow the `frm_per_pkt` setting, so the calculation is okay here. But for incoming direction, we cannot really control the `frm_per_pkt` (IIRC receiver can only signal/control the max `frm_per_pkt`, then the sender may send single frame or multiple frames per packet and IIRC even it may switch between them on the fly).

> If it's only for G722 though, it shouldn't be specific for multiple frames per packet, since the `clock rate` and `samples_per_frame` forced modification should also affect single frame per packet.

IMO for single frame per packet, the duration calculation should be correct (but not correct for multiple frames per packet), because `rtcp_setting.samples_per_frame` is only defined for single frame per packet:
https://github.com/pjsip/pjproject/blob/aa99bc23ae354177f474d79bac515cfb93baad62/pjmedia/src/pjmedia/stream.c#L2751

sauwming:
So the fix should be:
- Add setting `pjmedia_rtcp_session_setting.dec_samples_per_frame`
- Provide mechanism to update it, such as in the scenario of ptime change
- For G722: multiply samples_per_frame by `stream->codec_param.setting.frm_per_pkt`?

Are these correct?

nanangizz:
>     * Add setting `pjmedia_rtcp_session_setting.dec_samples_per_frame`
>     * Provide mechanism to update it, such as in the scenario of ptime change

IMO yes, should work in most scenarios where ptime is not changed frequently.

Another approach is perhaps by simply substracting current RTP ts from last/prev received RTP ts. This is simpler as we don't need to detect sender ptime & update the RTCP states, but it may not work well if DTX is enabled.


>     * For G722: multiply samples_per_frame by `stream->codec_param.setting.frm_per_pkt`?

Yes (for RTP sending direction, while incoming dir can follow the above mechanism).

PR #3539, opened 2023-04-27
Fixed memory leak while decoding video packet with ffmpeg

kala13x:
In function `ffmpeg_codec_decode_whole()` AVFrame `avframe` is used to receive a frame from the decoder codec context. The frame data is copied in the output `pjmedia_frame` but `avframe` is never unreferenced. This is causing serious memory leaks and a freeze of the whole system. Using `ffmpeg` version `N-109430-g7af947c0c0 Copyright (c) 2000-2022`.

### DIFF
--- a/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
+++ b/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
@@ -1848,6 +1848,18 @@ static pj_status_t check_decode_result(pjmedia_vid_codec *codec,
 }
 
 /*
+ * Unreference AVFrame.
+ */
+static pj_status_t ffmpeg_frame_unref(AVFrame *frame)
+{
+#ifdef PJMEDIA_USE_OLD_FFMPEG
+    (void)frame;
+#else
+    av_frame_unref(&frame);
+#endif
+}
+
+/*
  * Decode frame.
  */
 static pj_status_t ffmpeg_codec_decode_whole(pjmedia_vid_codec *codec,
@@ -1957,12 +1969,16 @@ static pj_status_t ffmpeg_codec_decode_whole(pjmedia_vid_codec *codec,
          */
         status = check_decode_result(codec, &input->timestamp,
                                      avframe.key_frame);
-        if (status != PJ_SUCCESS)
+        if (status != PJ_SUCCESS) {
+            ffmpeg_frame_unref(&avframe);
             return status;
+        }
 
         /* Check provided buffer size */
-        if (vafp->framebytes > output_buf_len)
+        if (vafp->framebytes > output_buf_len) {
+            ffmpeg_frame_unref(&avframe);
             return PJ_ETOOSMALL;
+        }
 
         /* Get the decoded data */
         for (i = 0; i < ff->dec_vfi->plane_cnt; ++i) {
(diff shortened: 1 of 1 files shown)


sauwming (commented):
How about in these two places where it can return earlier?
```
        status = check_decode_result(codec, &input->timestamp,
                                     avframe.key_frame);
        if (status != PJ_SUCCESS)
            return status;

        /* Check provided buffer size */
        if (vafp->framebytes > output_buf_len)
            return PJ_ETOOSMALL;
```

And what if `(got_picture == FALSE)`? Do we still need to unref the frame?

jimying:
av_frame_unref() is add in lavc 55.45.101 
please see ffmpeg doc: https://github.com/FFmpeg/FFmpeg/blob/4e57a0f35a56a4524648530ae6fd9c7f7770d8ca/doc/APIchanges#L1981

do we need check libavcodec  version ?
```c
#if LIBAVCODEC_VER_AT_LEAST(55,46)
av_frame_unref(&avframe);
#endif
```

kala13x:
At first, I didn't pay attention, but now I looked and you are right, all return cases must be handled, sorry for that.

It will also be good to reset the frame with the old version

`
#ifdef PJMEDIA_USE_OLD_FFMPEG
    avcodec_get_frame_defaults(&avframe);
#else
    pj_bzero(&avframe, sizeof(avframe));
    av_frame_unref(&avframe);
#endif
`

It is generally a good practice to unreference or release any resources, such as frames when they are no longer needed. Even if `got_picture` is `FALSE` `av_frame_unref()` will check if the frame is referenced.

I can make another pull request with fixed all of this if required

sauwming:
You don't need to create a separate PR, you can just fix this one instead if possible, but if you want to make another all, it's also fine.
- Checking the sample https://ffmpeg.org/doxygen/trunk/demuxing__decoding_8c_source.html, it seems that we only need to call `av_frame_unref()` if `avcodec_receive_frame()` returns >= 0. In other words,  `av_frame_unref()` is not needed when `got_picture == FALSE`.
- I'm not sure why we need to call `av_frame_unref()` during initialisation, after bzero-ing avframe (https://github.com/pjsip/pjproject/blame/master/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c#L1884). Is this actually necessary or do we need to remove this? Perhaps @trengginas can explain why this is needed?

kala13x:
Just updated the source, unreferencing the frames only if `got_picture` and there is an appropriate version of the ffmpeg. Created a new function `ffmpeg_frame_unref` where I guarded `av_frame_unref` with ifdefs to keep the code readable.

As for the dereferencing frame at the beginning of the function, according to this doc, it resets the frame fields.
https://ffmpeg.org/doxygen/3.3/group__lavu__frame.html#ga0a2b687f9c1c5ed0089b01fd61227108

Then I checked the source of this function and it calls `get_frame_defaults` at the end
https://ffmpeg.org/doxygen/3.3/frame_8c_source.html#l00498

And [get_frame_defaults](https://ffmpeg.org/doxygen/3.3/frame_8c.html#a86b63f32a4d199c8daa17084f5c7ac94) itself sets values that are different from 0, so I think `av_frame_unref` is still required after calling `bzero` this is why I leaved this part unchanged.

sauwming (approved):

nanangizz (approved):

Issue #1907, opened 2016-03-10
Remove pjmedia* circular dependency


pjsipbot:
_2016-03-10 03:45:14: @mingteluu_ _created the issue on trac ticket [1907](https://trac.pjsip.org/repos/ticket/1907)_

Currently `pjmedia` library will depend on `pjmedia-audiodev` and `pjmedia-videodev`, and vice versa. This may cause problem, for example when building shared libraries with linker flags `--as-needed`.
___
_2016-03-10 05:02:09: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2016-03-10 05:02:09: @mingteluu_ _set **resolution** to **fixed**_

___
_2016-03-10 05:02:09: @mingteluu_ _commented_

In [r5255](../commit/b1f406ce4fa55399955f17cedaec974b83c7b34c):
Fixed #1907: Remove pjmedia* circular dependency

___
_2016-03-11 12:44:01: @mingteluu_ _commented_

In [r5257](../commit/f3790cfd24bf8a656dc69304fd3e446bee399302):
Re #1907: Fixed build error: undeclared variables PJMEDIA_VID_DEV_MAX_DRIVERS and PJMEDIA_VID_DEV_MAX_DEVS when PJMEDIA_HAS_VIDEO is disabled.

___
_2016-03-14 07:59:40: @nanangizz_ _commented_

In [r5260](../commit/df1b7f092a8b3c88409468159b9c16ed0b85f81a):
Re #1907: Updated VS2005 build settings, added the new files audio/videodev.h/c.

___
_2016-03-18 07:42:11: @trengginas_ _commented_

In [r5267](../commit/3704a6368a7ddbf51dd5a43012916e6630c7b903):
Re #1907: Updated VS2015 build settings, added the new files audio/videodev.h/c.

PR #3544, opened 2023-04-28
Fix minor error: duplicate link pjmedia lib

jimying:
APP_LDLIBS include PJMEDIA_LDLIB twice.
(which affects 'libs' in pkgconfig file libpjproject.pc )


sauwming:
I believe this was due to the circular dependency last time, which should been resolved in #1907.

sauwming (approved):

nanangizz (approved):

PR #3546, opened 2023-05-01
Fixed segmentation fault in pjmedia_vid_stream_destroy

kala13x:
In the function `pjmedia_vid_stream_destroy`, the variable `stream->grp_lock` can be `NULL` when `send_rtcp` is called and this is causing a segmentation fault, because `send_rtcp` tries to acquire the lock with it. Even in the `pjmedia_vid_stream_destroy` function, there is a comment below

```C
    /* This function may be called when stream is partly initialized,
     * i.e: group lock may not be created yet.
     */
````


sauwming:
Thanks for the fix. It's due to my recent PR #3528 , so it should affect the audio stream as well that uses `stream->jb_mutex`.

kala13x:
You are right same is happening for the audio. Just pushed the fix for it also.

sauwming (approved):

nanangizz (approved):

Issue #1786, opened 2014-09-15
Add support to specify video format and video window when opening video preview


pjsipbot:
_2014-09-15 09:59:09: @mingteluu_ _created the issue on trac ticket [1786](https://trac.pjsip.org/repos/ticket/1786)_

App has no way of specifying desired format and output window when opening preview.

Solution:
In Pjsua: Add format and wnd fields in pjsua_vid_dev_preview_param.

___
_2014-09-15 09:59:48: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2014-09-15 09:59:48: @mingteluu_ _set **resolution** to **fixed**_

___
_2014-09-15 09:59:48: @mingteluu_ _commented_

In [r4921](../commit/c1a554e478b59c9eaa7bbc84d69f9fd22639d5e8):
Fixed #1786: Add support to specify video format when opening video preview

___
_2015-03-17 09:58:57: @mingteluu_ _edited the issue description_

___
_2015-03-17 09:58:57: @mingteluu_ _changed **title** from **Add support to specify video format when opening video preview** to **Add support to specify video format and video window when opening video preview**_

___
_2015-03-17 09:59:44: @mingteluu_ _commented_

In [r4995](../commit/90ef3ff840b3a9ee94bd85a47798387152113c8a):
Re #1786: Add support to specify output window as well for video preview param.

PR #3547, opened 2023-05-03
Enable specifying format video when opening video preview using PJSUA2 API

trengginas:
The ticket in #1786 add the capability to specify format video and video window when opening video preview.
However, this currently doesn't work on PJSUA2 API since `VideoPreviewOpParam` is using `MediaFormat` field.
This PR will change the `MediaFormat` to `MediaFormatVideo` and enable the capability.

Note: minor API break: `MediaFormat` field in `VideoPreviewOpParam` no longer exists.

### DIFF
--- a/pjsip/include/pjsua2/media.hpp
+++ b/pjsip/include/pjsua2/media.hpp
@@ -1938,9 +1938,9 @@ struct VideoPreviewOpParam {
     unsigned                windowFlags;
 
     /**
-     * Media format. If left unitialized, this parameter will not be used.
+     * Media format video. If left unitialized, this parameter will not be used.
      */
-    MediaFormat             format;
+    MediaFormatVideo        format;
 
     /**
      * Optional output window to be used to display the video preview.
--- a/pjsip/include/pjsua2/media.hpp
+++ b/pjsip/include/pjsua2/media.hpp
@@ -1938,7 +1938,10 @@ struct VideoPreviewOpParam {
     unsigned                windowFlags;
 
     /**
-     * Media format video. If left unitialized, this parameter will not be used.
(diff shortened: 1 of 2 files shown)


pjsip/include/pjsua2/media.hpp
```diff
@@ -1938,9 +1938,9 @@ struct VideoPreviewOpParam {
     unsigned                windowFlags;
 
     /**
-     * Media format. If left unitialized, this parameter will not be used.
+     * Media format video. If left unitialized, this parameter will not be used.
```
nanangizz, line 1941: - s/unitialized/uninitialized
- rather OOT, while here, perhaps update the docs for more detail info, e.g:
   - if uninitialized, capture device will be opened using PJMEDIA wrapper default format, e.g: on Android it is ..., on iOS it is .
   - if the preview is already opened, this format setting will be ignored (or is it?).

pjsip/include/pjsua2/media.hpp
```diff
@@ -1938,7 +1938,10 @@ struct VideoPreviewOpParam {
     unsigned                windowFlags;
 
     /**
-     * Media format video. If left unitialized, this parameter will not be used.
+     * Media format video. If left uninitialized, this parameter will not be used and 
+     * the capture device will be opened using PJMEDIA wrapper default format, 
+     * e.g: on Android it is PJMEDIA_FORMAT_I420, on iOS it is PJMEDIA_FORMAT_BGRA.
```
nanangizz, line 1943: Actually I was thinking about the default resolution and frame rate info here rather than pixel format, as an initial hint about the video quality :)

nanangizz (approved):
Just a minor note below.

sauwming (approved):

PR #3549, opened 2023-05-03
don't assert when attempting to set PJ_DNS_RESOLVER_MAX_NS nameservers

andreas-wehrmann:
In Debug builds the function https://github.com/pjsip/pjproject/blob/e210c69104a871cdcaa9dc9747a684a4fa7e8b20/pjlib-util/src/pjlib-util/resolver.c#L535 asserts when trying to set the maximum number of nameservers allowed.

### DIFF
--- a/pjlib-util/src/pjlib-util/resolver.c
+++ b/pjlib-util/src/pjlib-util/resolver.c
@@ -542,7 +542,7 @@ PJ_DEF(pj_status_t) pj_dns_resolver_set_ns( pj_dns_resolver *resolver,
     pj_status_t status;
 
     PJ_ASSERT_RETURN(resolver && count && servers, PJ_EINVAL);
-    PJ_ASSERT_RETURN(count < PJ_DNS_RESOLVER_MAX_NS, PJ_EINVAL);
+    PJ_ASSERT_RETURN(count <= PJ_DNS_RESOLVER_MAX_NS, PJ_EINVAL);
 
     pj_grp_lock_acquire(resolver->grp_lock);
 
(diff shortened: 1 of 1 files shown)


sauwming (approved):

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -542,7 +542,7 @@ PJ_DEF(pj_status_t) pj_dns_resolver_set_ns( pj_dns_resolver *resolver,
     pj_status_t status;
 
     PJ_ASSERT_RETURN(resolver && count && servers, PJ_EINVAL);
-    PJ_ASSERT_RETURN(count < PJ_DNS_RESOLVER_MAX_NS, PJ_EINVAL);
+    PJ_ASSERT_RETURN(count <= PJ_DNS_RESOLVER_MAX_NS, PJ_EINVAL);
```
nanangizz, line 545: While here, perhaps also:
- update the error code to `PJ_ETOOMANY`?
- remote the redundant `count` check below?

sauwming: Yes, on release mode, PJ_ASSERT_RETURN will become regular error checking, so the condition `if (count > PJ_DNS_RESOLVER_MAX_NS)` will never be true and can be safely removed.

andreas-wehrmann: Done and I also changed the return value to `PJ_ETOOMANY` which makes more sense because it is more specific here.

nanangizz (approved):

Issue #3550, opened 2023-05-04
Add close-on-exec flag


jimying:
### Describe the bug

This is an new feature:

In unix(linux/bsd), there has close-on-exec flag for file descriptor,
which is very useful when start child process by exec functions.
If a file descriptor has close-on-exec flag,
 child process will auto close file descriptors that parent process opened ( not inherit) when start.

Now pjlib has no way to set this flag for ioqueue/socket/openfile.

Points that need to be modified：
1.  ioqueue
epoll_create()/kqueue() file descriptor,  use fcntl() to set flag
```c
fcntl(fd, F_SETFD, FD_CLOEXEC);
```

2. socket/socketpair/activesock
set the arg 'type'  with value SOCK_CLOEXEC
```c
socket(AF_INET, SOCK_STREAM | SOCK_CLOEXEC, 0)
socketpair(AF_LOCAL, SOCK_STREAM | SOCK_CLOEXEC, 0, sv)
```
3.  open file
By set  fopen() arg 'mode'  with value  'e'
```c
fopen(path, 'w+e')
fopen(path, 're')
fopen(path, 'ae')
```

Can we add this new feature？

This is just a suggestion， to make pjlib more general useful.

### Steps to reproduce

- none

### PJSIP version

2.13-dev

### Context

- none

### Log, call stack, etc

```shell
- none
```

sauwming:
I have read the manual of the flag.

I still don't understand though, how this could be useful for us.

jimying:
Suppose a small server:  
a module use pjsip/pjmedia/pjnath/pjlib, which open some sockets and files; 
another module in some cases will use fork&exec/system to create sub processes to run external  cmd.
If no close-on-exec,  the sub process will inherit the file descriptors that opened
This may lead to security issues：
1. file decriptors count is too many,  the left available count to the sub process is too few
2. sub process may has long life cycle, the sockets file handle will long time not release 
(udp/tcp ports are occupied ), which may cause main process can't alloc ports resources.


So we'd better add close-on-exec flag

sauwming:
I don't have any strong objection, my only concern is the platform (and version) specific availability.
`SOCK_CLOEXEC` only since Linux 2.6.27
`e (since glibc 2.7)
              Open the file with the O_CLOEXEC flag.`

What do you think, @nanangizz ?

nanangizz:
[...]

PR #3553, opened 2023-05-05
Add close-on-exec flag

jimying:
Detail please see #3550

### DIFF
--- a/pjlib-util/src/pjlib-util/dns_server.c
+++ b/pjlib-util/src/pjlib-util/dns_server.c
@@ -76,6 +76,7 @@ PJ_DEF(pj_status_t) pj_dns_server_create( pj_pool_factory *pf,
     pj_dns_server *srv;
     pj_sockaddr sock_addr;
     pj_activesock_cb sock_cb;
+    pj_activesock_cfg sock_cfg;
     pj_status_t status;
 
     PJ_ASSERT_RETURN(pf && ioqueue && p_srv && flags==0, PJ_EINVAL);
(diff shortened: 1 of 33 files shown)


jimying:
pjlib changes:
- add new api pj_set_cloexec_flag()
   use fcntl() to set file handle close-on-exec flag
- pj_file_open()  
   add new mode type PJ_O_CLOEXEC
- pj_sock_socket() 
  add new socket type PJ_SOCK_CLOEXEC/pj_SOCK_SOCKEXEC()
  Note: 
   linux and bsd platform has the define SOCK_SOCKEXEC,  but not on macos
   so on macos pj_SOCK_CLOEXEC() == 0
   on macos, can use fcntl() to set close-on-exec flag  (pj_set_cloexec_flag())
   So don't get confused by following code (which is to compatible with macos ):
   ```c
    status = pj_sock_socket(af, pj_SOCK_DGRAM() | pj_SOCK_CLOEXEC(), 0, &sock);
    if (status != PJ_SUCCESS)
        goto on_error;
    if (pj_SOCK_CLOEXEC() == 0)
        pj_set_cloexec_flag((int)sock);
   ```
 - add sock_cloexec option in  pj_activesock_cfg
 - add sock_cloexec option in  pj_ssl_sock_param

pjlib-util, pjsip, pjmedia, pjnath changes:
- default set socket and file handle close-on-exec flag

sauwming (approved):
If I check Asterisk code here: https://github.com/asterisk/asterisk/blob/master/main/asterisk.c#L2107,
they will manually set all fd with `FD_CLOEXEC` flag before calling `execvp()`, so IMO, enabling this flag will be useful for server apps.

Only a couple of minor warnings:
```
../src/pj/sock_bsd.c:555:9: warning: unused variable 'type0' [-Wunused-variable]
    int type0 = type;
```
```
../src/pj/sock_common.c:1256:9: warning: unused variable 'type0' [-Wunused-variable]
    int type0 = type;
```
if built on platforms that don't have the flag, such as Android.

nanangizz (approved):

PR #3555, opened 2023-05-08
Support for codec with non-integer ptime

### LINKED ISSUE #605: Support for codec with non-integer ptime (e.g: MELP with ptime 22.5ms).
_2008-09-01 13:01:40: @nanangizz_ _created the issue on trac ticket [605](https://trac.pjsip.org/repos/ticket/605)_

sauwming:
To fix #605

As [RFC 6716](https://www.rfc-editor.org/rfc/rfc6716) specifies:
`Opus can encode frames of 2.5, 5, 10, 20, 40, or 60 ms`

However, it's been reported that if we receive incoming audio packet with non-integer ptime, it may cause PJSIP to experience audio issue and even lead to crash.

Log from the reporter:
```
2022-11-03 10:00:51.141        opus.c        media Opus ptime change detected: 20 ms --> 2 ms
2022-11-03 10:00:51.141  strm15A55F24        media codec decode ptime change detected: 20 -> 2
2022-11-03 10:00:51.142        opus.c         wmme Recover failed!
2022-11-03 10:00:51.159        opus.c        media No frames retrieved (num_frames = 0)
2022-11-03 10:00:51.159  strm15A55F24        media Codec parse() error: Codec internal creation error (PJMEDIA_CODEC_EFAILED)
```

When I tested it here, I could get exactly the same log as above and the audio issue was obviously expected since we did not support fractional ptime, but the crash very rarely happened, and only happened once with the stack trace:
```
==97056==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x000110e45ca0 at pc 0x0001048b56f4 bp 0x00016dfb3170 sp 0x00016dfb2930
WRITE of size 3840 at 0x000110e45ca0 thread T26
    #0 0x1048b56f0 in __asan_memset+0xf0 (libclang_rt.asan_osx_dynamic.dylib:arm64+0x3d6f0) (BuildId: ee0fd54f5b5b37fab0dff0277a63fa1d32000000200000000100000000000b00)
    #1 0x102ba9190 in pj_bzero string.h:767
    #2 0x102bbaf48 in pjmedia_zero_samples frame.h:268
    #3 0x102ba4ff0 in get_frame stream.c:779
    #4 0x102b3af10 in pjmedia_port_get_frame port.c:99
    #5 0x102ae7d9c in read_port conference.c:1621
```

After the PR, the log will be as follows:
```
14:11:06.145                 opus.c !Opus ptime change detected: 20/1 ms --> 5/2 ms
14:11:06.145        strm0x10c8b1928  codec decode ptime change detected: 20/1 -> 5/2
```

### DIFF
--- a/pjmedia/include/pjmedia-codec/opus.h
+++ b/pjmedia/include/pjmedia-codec/opus.h
@@ -98,7 +98,8 @@ typedef struct pjmedia_codec_opus_config
 {
     unsigned   sample_rate; /**< Sample rate in Hz.                     */
     unsigned   channel_cnt; /**< Number of channels.                    */
-    unsigned   frm_ptime;   /**< Frame time in msec.                    */
+    unsigned   frm_ptime;   /**< Frame ptime in msec.                   */
+    unsigned   frm_ptime_denum;/**< Frame ptime denumerator, can be zero*/
     unsigned   bit_rate;    /**< Encoder bit rate in bps.               */
     unsigned   packet_loss; /**< Encoder's expected packet loss pct.    */
     unsigned   complexity;  /**< Encoder complexity, 0-10(10 is highest)*/
--- a/pjmedia/include/pjmedia/codec.h
+++ b/pjmedia/include/pjmedia/codec.h
@@ -270,6 +270,9 @@ typedef struct pjmedia_codec_param
     /**
      * The "info" part of codec param describes the capability of the codec,
      * and is recommended not to be modified unless necessary.
+     * Note that application must be ready to handle cases when ptime
+     * denumerators are zero, since most codecs that only support integer
+     * ptime will leave these fields untouched.
      */
     struct {
        unsigned    clock_rate;          /**< Sampling rate in Hz            */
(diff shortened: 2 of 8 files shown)


nanangizz:
Just an idea, another approach is allowing ptime to be defined in micro seconds unit (e.g: `ptime_usec`) rather than adding ptime-denumerator. An advantage is we don't need to 'predefine' the possible denumerators, like in Opus ptime detection it is hardcoded to 2 in the patch.

sauwming:
That's a good idea. It's much simpler, calculation wise.
So if it's > 1000, it means the ptime is in usec, correct?

The only drawback that I can think is that it may potentially break apps that are not ready to accept ptime info that's very large?

nanangizz:
Looks like I described it ambiguously. Actually just like `ptime_denum`, it uses a separate field/var such as `ptime_usec`. So for the new apps, instead of setting `ptime`, it sets the `ptime_usec`. The lib perhaps will check the `ptime` first, if it is not zero, use it, otherwise check & use `ptime_usec`.

sauwming:
Hmmm... if it's also a separate field, I don't see much advantage. We don't actually need to predefine the denumerator to a certain number such as 2. If you prefer, we can always set the denum to 1000 and we will basically get the same thing (i.e. the ptime enumerator will be in usec). So instead of 5/2 msec, it will be represented as 2500/1000 msec instead.

I actually kind of like the idea of `allowing ptime to be defined in micro seconds unit`, so we don't need additional field. The only problem is the spec change and potential backward incompatibility. The potential incompatibility should be minor to none though, since `pjmedia_codec_param.info` is typically for information purpose only.

nanangizz:
Is there a possibility that some non-realtime app does really need ptime 1000ms or higher?  :D

sauwming:
Looks like the answer is yes.
`For non-real-time applications, such as file-based media delivery or offline transcoding, larger ptime values may be used to reduce the number of packets required for transmission and improve the efficiency of the encoding process. In these scenarios, ptime values of several seconds or more may be used.`
So yeah, we probably shouldn't be using only one field.

nanangizz (approved):

PR #3560, opened 2023-05-12
activesock: check remaining buffer size to avoid buffer overflow

jimying:
For tcp, if app error set the `remainder`  > `r->max_size`,
 it will lead to buffer overflow error when call pj_ioqueue_recv().

### DIFF
--- a/pjlib/src/pj/activesock.c
+++ b/pjlib/src/pj/activesock.c
@@ -607,6 +607,16 @@ static void ioqueue_on_read_complete(pj_ioqueue_key_t *key,
             }
         }
 
+        /* Check remaining buffer size to avoid buffer overflow.
+         * For tcp, if app error set the 'remainder' (r->size) > max_size,
+         * it will lead to buffer overflow error when call pj_ioqueue_recv().
+         */
+        if (asock->stream_oriented && r->size > r->max_size) {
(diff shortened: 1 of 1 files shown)


sauwming (commented):
Perhaps adding:
`PJ_ASSERT_ON_FAIL(!asock->stream_oriented || remainder <= r->size, {PJ_LOG(3, ("", "App bug! Invalid remainder length from activesock on_data_read().")); remainder=0;});`
after each call to `(*asock->cb.on_data_read)()` should be better.

Reason:
We can use stricter checking, against the data length`r->size`, instead of `r->max_size`.

jimying:
> Perhaps adding: `PJ_ASSERT_ON_FAIL(!asock->stream_oriented || remainder <= r->size, {PJ_LOG(3, ("", "App bug! Invalid remainder length from activesock on_data_read().")); remainder=0;});` after each call to `(*asock->cb.on_data_read)()` should be better.
> 
> Reason: We can use stricter checking, against the data length`r->size`, instead of `r->max_size`.

You are right. use `r->size`, instead of `r->max_size`  is correct for tcp.

sauwming (approved):

nanangizz (approved):

PR #3561, opened 2023-05-12
Fix frame copy for stereo mode in WebRTC-AEC3

nanangizz:
Reported that when opening sound device in stereo/multichannel with AEC using WebRTC-AEC2, assertion is raised:
```
pj_assert("STACK OVERFLOW!! " && (usage <= thread->stk_size - 128));
```

Thanks to Adam Wientzek for the report.


sauwming (approved):
It's AEC 3, btw :D

I remember the first AEC doesn't work for stereo, so does AEC3 work for stereo?

nanangizz:
> It's AEC 3, btw :D
> 
> I remember the first AEC doesn't work for stereo, so does AEC3 work for stereo?

Not really sure, IIRC there was no echo, but tested only on that one PC with built-in mic+speaker.

PR #3565, opened 2023-05-15
Update SDP o= and t= lines to use 64 bit unsigned integer

nanangizz:
Reported that some implementations generate long numbers (more than 32 bit) for SDP id & version in SDP o= line. While our SDP defines those fields that as `pj_uint32_t`. The RFC 8866 seems to specify those as generic number, while in the older RFCs refer/recommend to using NTP timestamp (which is 64 bit, but it may wrap in year 2036).

So this PR updates the definitions of those field to using `pj_uint64_t`, if the system supports it. To avoid code complexity due to extensive use of `PJ_HAS_INT64`, this PR introduce a new type: generic unsigned integer, which is `pj_uint64_t` or `pj_uint32_t` (if the system does not have 64 bit integer). This PR update SDP t= line to using the newly defined generic integer too.

### Before & after the patch

When receiving SDP o= line:
```
o=alice 16719434913131229871 16719434913131229872 IN IP4 host.anywhere.com
```
Before the patch (in master branch), parsing SDP will fail with error code `PJMEDIA_SDP_EINORIGIN` because `pj_strtoul3()`, used for converting SDP id from string to long, returns `PJ_ETOOBIG`. After the patch, the SDP is parsed successfully & printed correctly.

### DIFF
--- a/pjlib/include/pj/string.h
+++ b/pjlib/include/pj/string.h
@@ -723,7 +723,7 @@ PJ_DECL(pj_status_t) pj_strtoul3(const pj_str_t *str, unsigned long *value,
  * been processed.
  *
  * @param str       The input string.
- * @param value     Pointer to a unsigned integer to receive the value.
+ * @param value     Pointer to an unsigned integer to receive the value.
  *                  The value will be a 64 bit unsigned integer if the system
  *                  supports it, otherwise a 32 bit unsigned integer.
  * @param base      Number base to use.
--- a/pjlib/include/pj/string.h
+++ b/pjlib/include/pj/string.h
@@ -718,6 +718,31 @@ PJ_DECL(pj_status_t) pj_strtoul3(const pj_str_t *str, unsigned long *value,
                                  unsigned base);
 
 /**
+ * Convert string to generic unsigned integer. The conversion will stop as
+ * soon as non-digit character is found or all the characters have
+ * been processed.
+ *
+ * @param str       The input string.
+ * @param value     Pointer to a unsigned integer to receive the value.
+ *                  The value will be a 64 bit unsigned integer if the system
+ *                  supports it, otherwise a 32 bit unsigned integer.
+ * @param base      Number base to use.
+ *
+ * @return PJ_SUCCESS if successful.  Otherwise:
+ *         PJ_ETOOBIG if the value was an impossibly long positive number.
+ *         In this case, *value will be set to ULLONG_MAX (for 64 bit) or
+ *         ULONG_MAX (for 32 bit).
+ *         \n
+ *         PJ_EINVAL if the input string was NULL, the value pointer was NULL
+ *         or the input string could not be parsed at all such as starting
+ *         with a character outside the base character range.  In this case,
+ *         *value will be left untouched.
+ */
+PJ_DECL(pj_status_t) pj_strtoul4(const pj_str_t *str, pj_uint_t *value,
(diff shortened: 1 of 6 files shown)


pjlib/include/pj/string.h
```diff
@@ -717,6 +717,31 @@ PJ_DECL(unsigned long) pj_strtoul2(const pj_str_t *str, pj_str_t *endptr,
 PJ_DECL(pj_status_t) pj_strtoul3(const pj_str_t *str, unsigned long *value,
                                  unsigned base);
 
+/**
+ * Convert string to generic unsigned integer. The conversion will stop as
+ * soon as non-digit character is found or all the characters have
+ * been processed.
+ *
+ * @param str       The input string.
+ * @param value     Pointer to a unsigned integer to receive the value.
```
sauwming, line 726: Minor: an

pjlib/src/pj/string.c
```diff
@@ -437,6 +437,72 @@ PJ_DEF(pj_status_t) pj_strtoul3(const pj_str_t *str, unsigned long *value,
     return PJ_SUCCESS;
 }
 
+PJ_DEF(pj_status_t) pj_strtoul4(const pj_str_t *str, pj_uint_t *value,
+                                unsigned base)
```
sauwming, line 441: can strtoul3() call this, to minimise duplication?

pjlib/src/pj/string.c
```diff
@@ -514,6 +580,43 @@ PJ_DEF(int) pj_utoa_pad( unsigned long val, char *buf, int min_dig, int pad)
     return len;
 }
 
+PJ_DEF(int) pj_utoa2(pj_uint_t val, char *buf)
+{
+    return pj_utoa_pad2(val, buf, 0, 0);
+}
+
+PJ_DEF(int) pj_utoa_pad2(pj_uint_t val, char *buf, int min_dig, int pad)
```
sauwming, line 588: utoa_pad() can also call this?

sauwming (commented):
Just curious, what happens if PJSIP without patch calls/receive calls from the one with patch? Will everything work fine?

nanangizz:
> Just curious, what happens if PJSIP without patch calls/receive calls from the one with patch? Will everything work fine?

Nothing changed in the `o=` line generation, year 2023 is still inside 32 bit range and no zero padding/prefix added. SDP id & version are generated by `pjmedia_endpt_create_base_sdp()` using `pj_gettimeofday()`.
```
patched (SDP offerer):
o=- 3893158240 3893158240 IN IP4 192.168.1.2

not patched (SDP answer):
o=- 3893158241 3893158242 IN IP4 192.168.1.2
```

pjlib/src/pj/string.c
nanangizz: Perhaps yes, however as they are actually different (long can be 32 bit or more), it should be safe if we assume that:
- if a system does not have 64 bit, long will always be 32 bit,
- `long` is always in the range of 32-64 bit (the spec seems to only specify minimum 32 bit), if some system's `long` is more than 64 bit, it is vice versa (strtoul4() should call strtoul3) :)

nanangizz:
> Oh okay, then what if we use some random 64-bit numbers instead of`gettimeofday()`? How is the behavior?

Just added a sample scenario with before & after the patch to the PR desc.

sauwming (approved):

PR #3566, opened 2023-05-15
Fixed PJSIP Developer's Guide link in sip_module doc

sauwming:
It is reported that the link is broken:
https://docs.pjsip.org/en/latest/api/generated/pjsip/group/group__PJSIP__MOD.html

which should point to:
https://docs.pjsip.org/en/latest/api/pjsip/guide.html

Another alternative is to remove the link altogether so we don't need to update the link in case that the doc is migrated in the future.


nanangizz (approved):
Vote for keeping the devs guide link.

PR #3568, opened 2023-05-16
Improve the handling of consecutive ioq accept errors

sauwming:
It is reported that listener can get into a bad state, but there was no log of `Received %d consecutive errors: %d for the accept() operation, stopping further ioqueue accepts.`
Note that the issue happens very rarely, so reproduction may not be straightforward.

There are a couple of potential issues identified:
- The accept errors may not always be identical.
- Application may not always be notified via the callback since currently `on_accept_complete2()` is only called `if status == PJ_STATUS_FROM_OS(OSERR_EWOULDBLOCK)`

There are a few alternatives to fix it:
a. Remove the requirement of errors tracked to be identical (the fix in this PR)
Con: `PJ_ACTIVESOCK_MAX_CONSECUTIVE_ACCEPT_ERROR` spec change
b. Introduce another setting `PJ_ACTIVESOCK_MAX_CONSECUTIVE_ACCEPT_ERROR2` (any better name?) to track non-identical errors.
Con: Too many settings
c. Also track non-identical errors but we don't need additional setting. Instead, we limit the non-identical errors to `c * PJ_ACTIVESOCK_MAX_CONSECUTIVE_ACCEPT_ERROR` (for example, double or triple the value).
Con: Slightly more complicated spec, and `c` is constant.

### DIFF
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -681,13 +681,13 @@
 #endif
 
 /**
- * Maximum consecutive identical error for accept() operation before
+ * Maximum consecutive errors for accept() operation before
  * activesock stops calling the next ioqueue accept.
  *
- * Default: 50
+ * Default: 100
  */
 #ifndef PJ_ACTIVESOCK_MAX_CONSECUTIVE_ACCEPT_ERROR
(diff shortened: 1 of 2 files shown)


sauwming:
Closing this for now, waiting for further test results.

Issue #2352, opened 2020-03-22
Interface for direct media manipuation with PJSUA2


malarinv:
I am looking for a way to manipulate media by capture and playback of audio frames using callbacks with PJSUA2 on python3.7 or higher.

 I looked for it in the Faqs but it seem, it requires working with implementing pjmedia_struct [1] which isn't possible for me as the rest of my pipeline is written in python and there is no high level interface for doing it using pjsua2. I am looking for a high level interface in pjsua2 

There is an old [mailing list thread](http://lists.pjsip.org/pipermail/pjsip_lists.pjsip.org/2013-March/036303.html)  that provides a patch to do. It still works albeit with many bugs. It uses the legacy pjsua-python bindings. The bugs cases unpredictable frame capture failures on some calls and race conditions when handling multiple audio calls, and unreliable in general.

It is a requirement for dynamically responding based on spoken audio.

trengginas:
you should try using ```on_aud_prev_play_frame()```/```on_aud_prev_rec_frame()``` from ```pjsua_media_config```. Note that you may need to add the python interface yourself

malarinv:
@trengginas. Thanks for your response. But the `on_aud_prev_play_frame`  interface doesn't give me any way to distinguish the frame associated with each call during a multiple call scenario i believe. Atleast the `pjmedia_port_get_frame` interface gives me the mixed audio frames of all the stream which can be tracked per call. 

Anyways i'll try to port the media frame callback interface provided by [Tomáš Valenta's](http://lists.pjsip.org/pipermail/pjsip_lists.pjsip.org/2013-March/036303.html) patch to pjsua2.

nicolaipre:
@malarinv Did you manage to patch it to pjsua2? I am looking for the same, although I have been looking for an implementation for the legacy pjsua. 

I'm not sure if it can be helpful to you or not, but I found a different implementation I ported to Python 3: https://github.com/nicolaipre/python3-pjsip-memory-buffer. There are still some bugs in it though. 

As I am not experienced enough in low level programming, I made a new issue/feature request regarding this as well since I am sick of working on this problem and getting nowhere...

malarinv:
@nicolaipre i'haven't yet started working on pjsua2, but i had managed to patch the legacy pjsua python interface to support audio frame callbacks in python3. you can find and example at https://github.com/malarinv/pjproject/blob/py37/pjsip-apps/src/python/samples/audio_cb.py
still yet to rebase against master.

[...]

PR #3569, opened 2023-05-17
Add PJSUA2 interface to get and put audio frames

sauwming:
To implement #2352.

Application can extend the class `AudioMediaPort` to get the callbacks `onFrameRequested()` and `onFrameReceived()`.

An example usage for C++ and Python can be found in `pjsua2_demo.cpp` and `test.py`.

### DIFF
--- a/pjsip/include/pjsua2/media.hpp
+++ b/pjsip/include/pjsua2/media.hpp
@@ -477,8 +477,8 @@ public:
     /**
      * This callback is called to request a frame from this port. On input,
      * frame.size indicates the capacity of the frame buffer and frame.buf
-     * will initially be an empty string. Application can then set the frame
-     * type, fill the string buffer, and set the string's length.
+     * will initially be an empty vector. Application can then set the frame
+     * type and fill the vector.
      *
      * @param frame       The frame.
      */
@@ -487,8 +487,8 @@ public:
 
     /**
      * This callback is called when this port receives a frame. The frame
-     * content will be provided in frame.buf, and the frame size can be
-     * found in either frame.size or the string buffer's length (both
+     * content will be provided in frame.buf vector, and the frame size
+     * can be found in either frame.size or the vector's size (both
      * have the same value).
      *
      * @param frame       The frame.
--- a/pjsip/include/pjsua2/media.hpp
+++ b/pjsip/include/pjsua2/media.hpp
@@ -479,7 +479,7 @@ public:
      *
      * @param frame       The frame.
      */
-    virtual void getFrame(MediaFrame &frame)
+    virtual void onFrameRequested(MediaFrame &frame)
     { PJ_UNUSED_ARG(frame); }
 
     /**
@@ -487,7 +487,7 @@ public:
      *
      * @param frame       The frame.
      */
-    virtual void putFrame(MediaFrame &frame)
+    virtual void onFrameReceived(MediaFrame &frame)
     { PJ_UNUSED_ARG(frame); }
 
 private:
--- a/pjsip/include/pjsua2/media.hpp
+++ b/pjsip/include/pjsua2/media.hpp
@@ -427,6 +427,75 @@ typedef std::vector<AudioMedia*> AudioMediaVector;
 typedef std::vector<AudioMedia> AudioMediaVector2;
 
 /**
+ * This structure describes a media frame.
+ */
+struct MediaFrame
+{
+    pjmedia_frame_type   type;      /**< Frame type.                        */
+    string               buf;       /**< Frame buffer content.              */
+    unsigned             size;      /**< Frame size in bytes.               */
+
+public:
+    /**
+     * Default constructor
+     */
+    MediaFrame()
+    : type(PJMEDIA_FRAME_TYPE_NONE),
+      size(0)
+    {}
+};
+
+/**
+ * Audio Media Port.
+ */
+class AudioMediaPort : public AudioMedia
+{
+public:
+    /**
+     * Constructor.
+     */
+    AudioMediaPort();
+
+    /**
+     * Destructor. This will unregister the audio media port from the
+     * conference bridge.
+     */
+    virtual ~AudioMediaPort();
+
+    /**
+     * Create an audio media port and register it to the conference bridge.
+     *
+     * @param name      The port name.
+     * @param fmt       The audio format.
+     */
+    void createPort(const string &name, MediaFormatAudio &fmt)
+                    PJSUA2_THROW(Error);
+
+    /*
+     * Callbacks
+     */
+    /**
+     * Get a frame from the port.
+     *
+     * @param frame       The frame.
+     */
+    virtual void getFrame(MediaFrame &frame)
+    { PJ_UNUSED_ARG(frame); }
+
+    /**
+     * Put a frame to the port.
+     *
+     * @param frame       The frame.
+     */
+    virtual void putFrame(MediaFrame &frame)
+    { PJ_UNUSED_ARG(frame); }
+
+private:
+    pj_pool_t *pool;
+    pjmedia_port port;
+};
+
+/**
  * This structure contains additional info about AudioMediaPlayer.
  */
 struct AudioMediaPlayerInfo
--- a/pjsip/include/pjsua2/media.hpp
+++ b/pjsip/include/pjsua2/media.hpp
@@ -432,7 +432,7 @@ typedef std::vector<AudioMedia> AudioMediaVector2;
 struct MediaFrame
 {
     pjmedia_frame_type   type;      /**< Frame type.                        */
-    string               buf;       /**< Frame buffer content.              */
+    ByteVector           buf;       /**< Frame buffer content.              */
     unsigned             size;      /**< Frame size in bytes.               */
 
 public:
--- a/pjsip/include/pjsua2/media.hpp
+++ b/pjsip/include/pjsua2/media.hpp
@@ -475,7 +475,10 @@ public:
      * Callbacks
      */
     /**
-     * Get a frame from the port.
+     * This callback is called to request a frame from this port. On input,
+     * frame.size indicates the capacity of the frame buffer and frame.buf
+     * will initially be an empty string. Application can then set the frame
+     * type, fill the string buffer, and set the string's length.
      *
      * @param frame       The frame.
      */
@@ -483,7 +486,10 @@ public:
     { PJ_UNUSED_ARG(frame); }
 
     /**
-     * Put a frame to the port.
+     * This callback is called when this port receives a frame. The frame
+     * content will be provided in frame.buf, and the frame size can be
+     * found in either frame.size or the string buffer's length (both
+     * have the same value).
      *
      * @param frame       The frame.
      */
--- a/pjsip/include/pjsua2/types.hpp
+++ b/pjsip/include/pjsua2/types.hpp
@@ -52,6 +52,9 @@ typedef std::vector<std::string> StringVector;
 /** Array of integers */
 typedef std::vector<int> IntVector;
 
+/** Array of bytes */
+typedef std::vector<unsigned char> ByteVector;
+
 /** Map string to string */
 typedef std::map<std::string, std::string> StringToStringMap;
 
--- a/pjsip/src/pjsua2/media.cpp
+++ b/pjsip/src/pjsua2/media.cpp
@@ -265,6 +265,78 @@ AudioMedia* AudioMedia::typecastFromMedia(Media *media)
 
 ///////////////////////////////////////////////////////////////////////////////
 
+AudioMediaPort::AudioMediaPort()
+: pool(NULL)
+{
+}
+
+AudioMediaPort::~AudioMediaPort()
+{
+    if (pool) {
+        PJSUA2_CATCH_IGNORE( unregisterMediaPort() );
+        pj_pool_release(pool);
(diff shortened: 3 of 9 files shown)


nanangizz (commented):
So PJSUA2 eventually goes low level :) Should be okay if there are demands though.

Btw, why not just integrate this to `AudioMedia`?

pjsip/include/pjsua2/media.hpp
```diff
@@ -426,6 +426,75 @@ typedef std::vector<AudioMedia*> AudioMediaVector;
...
 typedef std::vector<AudioMedia> AudioMediaVector2;
 
+/**
+ * This structure describes a media frame.
+ */
+struct MediaFrame
+{
+    pjmedia_frame_type   type;      /**< Frame type.                        */
+    string               buf;       /**< Frame buffer content.              */
+    unsigned             size;      /**< Frame size in bytes.               */
```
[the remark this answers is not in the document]
sauwming: For efficiency and robustness.
- Efficiency.
I can pass empty string and `frame.size=capacity`. Only when the user wants to fill the frame will the buffer be allocated, and also as needed (i.e. `size` can be < `capacity`). This is particularly useful if user is only interested in `putFrame()` but not `getFrame()` for example.
- Robustness
User doesn't need to set string size to zero if they don't want to pass a frame or leave callback implementation as the default.

pjsip/include/pjsua2/media.hpp
```diff
@@ -426,6 +426,75 @@ typedef std::vector<AudioMedia*> AudioMediaVector;
...
+    virtual void getFrame(MediaFrame &frame)
+    { PJ_UNUSED_ARG(frame); }
+
+    /**
+     * Put a frame to the port.
+     *
+     * @param frame       The frame.
+     */
+    virtual void putFrame(MediaFrame &frame)
+    { PJ_UNUSED_ARG(frame); }
```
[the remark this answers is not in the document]
sauwming: Good idea. Will change.

pjsip/src/pjsua2/media.cpp
```diff
@@ -265,6 +265,78 @@ AudioMedia* AudioMedia::typecastFromMedia(Media *media)
...
+static pj_status_t get_frame(pjmedia_port *port, pjmedia_frame *frame)
+{
+    AudioMediaPort *mport = (AudioMediaPort *) port->port_data.pdata;
+    MediaFrame frame_;
+
+    frame_.size = frame->size;
+    mport->getFrame(frame_);
+    frame->type = frame_.type;
+    frame->size = PJ_MIN(frame_.buf.size(), frame_.size);
+    pj_memcpy(frame->buf, frame_.buf.c_str(), frame->size);
```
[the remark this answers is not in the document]
sauwming: According to the doc:
https://cplusplus.com/reference/string/string/c_str/
it should be safe, since the returned array must `include the same sequence of characters that make up the value of the string object`, so there won't be truncation.

```
    const char byteArray[] = {0x48, 0x65, 0x6C, 0x6C, 0x6F, 0x00, 0x77, 0x6F, 0x72, 0x6C, 0x64};
    std::string byteString(byteArray, sizeof(byteArray));

    std::cout << "Byte String: " << byteString << std::endl;
    std::cout << "Length: " << byteString.length() << std::endl;

    std::cout << "Byte String: " << byteString.c_str() << std::endl;

    char buf[sizeof(byteArray)];
    pj_memcpy(buf, byteString.c_str(), byteString.length());
    std::cout << "c_str copy, buf == byteArray? " << (pj_memcmp(buf, byteArray, sizeof(byteArray))? "different": "same") << std::endl;

    byteString.copy(buf, byteString.length());
    std::cout << "string copy, buf == byteArray? " << (pj_memcmp(buf, byteArray, sizeof(byteArray))? "different": "same") << std::endl;

    std::copy(byteString.begin(), byteString.end(), buf);
    std::cout << "std copy, buf == byteArray? " << (pj_memcmp(buf, byteArray, sizeof(byteArray))? "different": "same") << std::endl;
```

Output:
```
Byte String: Helloworld
Length: 11
Byte String: Hello
c_str copy, buf == byteArray? same
string copy, buf == byteArray? same
std copy, buf == byteArray? same
```

After comparing the three options, now I prefer std::string copy
https://cplusplus.com/reference/string/string/copy/
`Copies a substring of the current value of the string object into the array pointed by s. This substring contains the len characters that start at position pos. The function does not append a null character at the end of the copied content.`

What do you think?

pjsip/include/pjsua2/media.hpp
```diff
@@ -426,6 +426,75 @@ typedef std::vector<AudioMedia*> AudioMediaVector;
...
+     */
+    virtual ~AudioMediaPort();
+
+    /**
+     * Create an audio media port and register it to the conference bridge.
+     *
+     * @param name      The port name.
+     * @param fmt       The audio format.
+     */
+    void createPort(const string &name, MediaFormatAudio &fmt)
```
wosrediinanatour, line 471: Why is this not part of the constructor?

sauwming (commented):
> So PJSUA2 eventually goes low level :) Should be okay if there are demands though.

Yes, this seems to be quite popular within PJSIP community with the most thumb-ups, several patches, and at least four Github projects that I know of offering the feature.

> Btw, why not just integrate this to `AudioMedia`?

I thought of that too, but the doc deterred me:
```
 * Audio Media. This is a lite wrapper class for audio conference bridge port,
 * i.e: this class only maintains one data member, conference slot ID, and 
 * the methods are simply proxies for conference bridge operations.
```
If I add more data member (i.e. `pjmedia_port`) and the callbacks, the spec will change.
Also, it may cause conflicts with existing derived classes, such as `ToneGenerator`, `AudioMediaPlayer/Recorder`, etc. (i.e. there will be confusion whether the id belongs to the port or the derived media (tonegen, etc), and there can also be two create() functions such as `createPort()` and `createToneGen()`).

pjsip/include/pjsua2/media.hpp
```diff
@@ -426,6 +426,75 @@ typedef std::vector<AudioMedia*> AudioMediaVector;
...
+     */
+    virtual void getFrame(MediaFrame &frame)
+    { PJ_UNUSED_ARG(frame); }
+
+    /**
+     * Put a frame to the port.
+     *
+     * @param frame       The frame.
+     */
+    virtual void putFrame(MediaFrame &frame)
```
wosrediinanatour, line 490: Why is a non-const reference needed?

pjsip/include/pjsua2/media.hpp
sauwming: I want to make it inline with `AudioMediaPlayer/Recorder` and `ToneGenerator` which all have their own `create...()`

pjsip/include/pjsua2/media.hpp
```diff
@@ -426,6 +426,75 @@ typedef std::vector<AudioMedia*> AudioMediaVector;
...
+    MediaFrame()
+    : type(PJMEDIA_FRAME_TYPE_NONE),
+      size(0)
+    {}
+};
+
+/**
+ * Audio Media Port.
+ */
+class AudioMediaPort : public AudioMedia
```
wosrediinanatour, line 451: How about adding a method for unregistering the media port?

sauwming: The parent class already has that method (it's protected though, but since it's mostly for internal purpose it should be fine).

pjsip/include/pjsua2/media.hpp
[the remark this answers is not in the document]
wosrediinanatour: Then the name size is misleading. What about capacity?

pjsip/include/pjsua2/media.hpp
sauwming: There's no restriction that the frame can't be modified. For example, if you want to process it before passing it further to a downstream port or to another component.

wosrediinanatour: I see....with onFrameReceived/onFrameRequested it is now clear.

pjsip/src/pjsua2/media.cpp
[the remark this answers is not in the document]
nanangizz: Sounds good if the spec says the content won't be truncated. Also, we need to be careful as C++ specs may be changed (slightly) among versions.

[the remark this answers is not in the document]
sauwming: In the latest commit, change the buffer to use `std::vector` as suggested.
Also add a usage example in Python.

Note that threading must be enabled. Should we leave this as default? Not sure why anyone would want it disabled nowadays.

pjsip/include/pjsua2/types.hpp
```diff
@@ -52,6 +52,9 @@ typedef std::vector<std::string> StringVector;
 /** Array of integers */
 typedef std::vector<int> IntVector;
 
+/** Array of bytes */
+typedef std::vector<unsigned char> ByteVector;
```
nanangizz, line 56: More like a note rather than a suggestion.

IMO bit number is quite urgent as our media components moslty work with 8 or 16 bit audio sample, so was about to propose replacing term `byte` to `octet`, as byte may be more/less than 8 bits (see [here](https://stackoverflow.com/questions/18576822/sizeof-and-when-a-byte-is-larger-than-8-bits#comment27335390_18576901)), while `octet` is always 8 bits. But not sure if it is still relevant on current hardwares.

Also, was considering `unsigned char` vs `pj_uint8_t` for a similar reason, but SWIG macros may not recognize `pj_uint8_t` and some media parts in our library seem to use `unsigned char` too.

nanangizz (approved):

pjsip/include/pjsua2/types.hpp
sauwming: Noted. After experimenting with various data types, I now know how difficult it is to get the bit number correct, and each target language may behave differently and present different challenges.

bobvanderlinden:
Thank you! That should cover the use-case (and custom patch) for me. I'll give it a try in the near-future :+1:

semwal0808:
hi @sauwming  i have been use this audiomediaport in android code .... to provide it audio from a gstreamer pipeline so that we can get provide a customized audio in the call. but does,nt seem to working  and other thing we are native c++ for gstreamer code  while we are building pjsua2 as usaual for android did that can cause any issues?

sauwming:
You need to elaborate more:
- what exactly is not working
- what changes do you need from the current mechanism

C++ should work better (as compared to Android/Java, which has to invoke SWIG binding).

semwal0808:
hi @sauwming  the thing is that wwe are not getting voice att all in gstreamer and there is few instance in which we are getting it but it is all noice 
here is the code of onFrameRequested @Override
public void onFrameRequested(MediaFrame frame) {
    // Ensure the frame type is audio
     Log.i("ASHISH", "ON FRAME REQUESTED");

    // Get the current buffer size from the MediaFrame
    // Create a short array to hold audio data
    short [] buffer = new short[4096];

    // Call the native method to fill the buffer with audio data
    nativeGetAudioData(buffer);

    // Create a ByteVector using the short array
    ByteVector byteVector = new ByteVector(buffer);

    // Set the ByteVector buffer to the MediaFrame
    frame.setBuf(byteVector);
}

here is the code for jni 
static void Java_com_example_myport_nativeGetAudioData(JNIEnv *env, jobject thiz, jshortArray byte_array) {
    CustomData *data = GET_CUSTOM_DATA(env, thiz, custom_data_field_id);

    if (data == NULL) {
        GST_DEBUG("Custom data is NULL, returning early.");
        return;
    }

    // Ensure byte_array is not NULL before getting length
    if (byte_array == NULL) {
        GST_DEBUG("byte_array is NULL, cannot proceed.");
        return;
    }

    GstBuffer* gstBuffer = (GstBuffer *)g_queue_peek_head(data->queue);
    GST_DEBUG("size of queue: %d", data->queue->length);

    if (gstBuffer == NULL) {
        GST_DEBUG("No valid buffer in the queue, returning early.");
        return;
    }

    GstMapInfo mapInfo;
    if (gst_buffer_map(gstBuffer, &mapInfo, GST_MAP_READ)) {
        jsize byte_array_length = (*env)->GetArrayLength(env, byte_array);

        // Log buffer size details
        GST_DEBUG("byte_array_length: %d, mapInfo.size: %zu", byte_array_length, mapInfo.size);

        jsize copy_size = (mapInfo.size < byte_array_length * sizeof(jshort)) ? mapInfo.size : byte_array_length * sizeof(jshort);
        GST_DEBUG("copy_size: %d", copy_size);

        (*env)->SetShortArrayRegion(env, byte_array, 0, copy_size / sizeof(jshort), (jshort *)mapInfo.data);

        gst_buffer_unmap(gstBuffer, &mapInfo);
        g_queue_pop_head(data->queue);
        gst_buffer_unref(gstBuffer);
    } else {
        GST_DEBUG("Failed to map GstBuffer.");
    }
}

sauwming:
Ah, I misunderstood the C++ question.
Yes, for Java, you need to make sure you pass it correctly. Passing values between languages can be tricky indeed (for example, in Python, the endian-ness may be different, check the sample app), so make sure you test it with some sample values in the Java layer, and check if the values are the same in C++.

semwal0808:
@sauwming  there is another problem too when ever we connect the call between to android device ............ it only takes input from mic audio.... due some reason it i cant listen custom audio in other end

Issue #3575, opened 2023-05-22
Crash in ssl_write with threading and TLS connections


creslin2877:
### Describe the bug

It would appear that there is an issue with a lack of synchronization around openssl functions.  While using pjproject in a multithreaded environment (Asterisk 18.15.1) with TLS connected endpoints, a crash was triggered when two threads attempted to utilize same SSL object in ssl_sock_ossl.c on the same socket at the same time (specifically when triggering a write to the write BIOs, I think in this crash's case).  The thread that crashed was responding to an inbound subscribe request:
```
Thread 1 (Thread 0x7f090c696700 (LWP 6617)):
#0  0x00007f09b97ff73d in __memcpy_ssse3_back () at ../sysdeps/x86_64/multiarch/memcpy-ssse3-back.S:1631
#1  0x00007f09bb2829e4 in mem_write () at /lib64/libcrypto.so.10
#2  0x00007f09bb2819ab in BIO_write () at /lib64/libcrypto.so.10
#3  0x00007f09bb5f3722 in ssl3_write_pending (s=s@entry=0x7f09050d5100, type=type@entry=23, buf=<optimized out>, len=<optimized out>) at s3_pkt.c:1111
#4  0x00007f09bb5f3ac0 in do_ssl3_write (s=s@entry=0x7f09050d5100, type=type@entry=23, buf=0x7f09863e2c60 "", 
    buf@entry=0x7f098829cdd8 "SIP/2.0 481 Call/Transaction Does Not Exist\r\nVia: SIP/2.0/TLS 10.10.10.101:12581;rport=7738;received=72.69.152.133;branch=z9hG4bK483307485\r\nCall-ID: 0_2375866139@10.10.10.101\r\nFrom: \"110\" <sip:110-Cor"..., len=<optimized out>, create_empty_fragment=create_empty_fragment@entry=0) at s3_pkt.c:1083
#5  0x00007f09bb5f3e16 in ssl3_write_bytes (s=0x7f09050d5100, type=<optimized out>, buf_=0x7f098829cdd8, len=<optimized out>) at s3_pkt.c:833
#6  0x00007f09bc1c47a7 in ssl_write (ssock=0x7f0986123158, data=0x7f098829cdd8, size=424, nwritten=0x7f090c695848) at ../src/pj/ssl_sock_ossl.c:2383
[...]

PR #3577, opened 2023-05-24
Don't call SSL_shutdown() when receiving SSL_ERROR_SYSCALL or SSL_ERROR_SSL

### LINKED ISSUE #3576: Potentially improper use of SSL_shutdown() when receiving SSL_ERROR_SYSCALL or SSL_ERROR_SSL
### Describe the bug

In the man page for SSL_get_error, specifically with regards to SSL_ERROR_SYSCALL and SSL_ERROR_SSL, it states:

https://www.openssl.org/docs/man1.0.2/man3/SSL_get_error.html

>SSL_ERROR_SYSCALL
>Some non-recoverable, fatal I/O error occurred. The OpenSSL error queue may contain more information on the error. For >socket I/O on Unix systems, consult errno for details. If this error occurs then no further I/O operations should be performed >on the connection and SSL_shutdown() must not be called.
>
>SSL_ERROR_SSL
>A non-recoverable, fatal error in the SSL library occurred, usually a protocol error. The OpenSSL error queue contains more >information on the error. If this error occurs then no further I/O operations should be performed on the connection and >SSL_shutdown() must not be called.

It would appear that there's a CVE against openssl that caused this change in behavior:

https://forge.etsi.org/rep/cyber/103523_MSP/tlmsp/tlmsp-openssl/-/commit/ab874dfd3e22a7c6ea3d45bc352294546af5afff

In ssl_sock_ossl.c, if we receive either of these errors when performing an SSL_read(), it would appear that we need to ensure that ssl_reset_sock_state() does not call SSL_shutdown() when those errors are received to comply with this documented behavior.

### Steps to reproduce

Receive SSL_ERROR_SSL on TLS connection.

### PJSIP version

2.12.1+

### Context

n/a

### Log, call stack, etc

```shell
n/a
```

trengginas:
This is to fix #3576.
Before calling `SSL_shutdown()`, a check to SSL_ERROR_SYSCALL and SSL_ERROR_SSL is required.

### DIFF
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -1695,7 +1695,12 @@ static void ssl_destroy(pj_ssl_sock_t *ssock)
 /* Reset SSL socket state */
 static void ssl_reset_sock_state(pj_ssl_sock_t *ssock)
 {
-    ossl_sock_t *ossock = (ossl_sock_t *)ssock;
+    ssl_reset_sock_state_with_error(ssock, PJ_TRUE);
+}
+
+static void ssl_reset_sock_state_with_error(pj_ssl_sock_t* ssock, pj_bool_t check_error)
+{
+    ossl_sock_t* ossock = (ossl_sock_t*)ssock;
 
     /* Detach from SSL instance */
     if (ossock->ossl_ssl) {
@@ -1706,12 +1711,21 @@ static void ssl_reset_sock_state(pj_ssl_sock_t *ssock)
      * Avoid calling SSL_shutdown() if handshake wasn't completed.
      * OpenSSL 1.0.2f complains if SSL_shutdown() is called during an
      * SSL handshake, while previous versions always return 0.
+     * Don't send notify when the last error is SSL_ERROR_SYSCALL or SSL_ERROR_SSL.
      */
     if (ossock->ossl_ssl && SSL_in_init(ossock->ossl_ssl) == 0) {
-        int ret = SSL_shutdown(ossock->ossl_ssl);
-        if (ret == 0) {
-            /* Flush data to send close notify. */
-            flush_circ_buf_output(ssock, &ssock->shutdown_op_key, 0, 0);
+        pj_bool_t send_notify = PJ_TRUE;
+        if (check_error && (ssock->last_err == SSL_ERROR_SYSCALL ||
+            ssock->last_err == SSL_ERROR_SSL))
+        {
+            send_notify = PJ_FALSE;
+        }
+        if (send_notify) {
+            int ret = SSL_shutdown(ossock->ossl_ssl);
+            if (ret == 0) {
+                /* Flush data to send close notify. */
+                flush_circ_buf_output(ssock, &ssock->shutdown_op_key, 0, 0);
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1695,7 +1695,12 @@ static void ssl_destroy(pj_ssl_sock_t *ssock)
 /* Reset SSL socket state */
 static void ssl_reset_sock_state(pj_ssl_sock_t *ssock)
 {
-    ossl_sock_t *ossock = (ossl_sock_t *)ssock;
+    ssl_reset_sock_state_with_error(ssock, PJ_TRUE);
+}
+
+static void ssl_reset_sock_state_with_error(pj_ssl_sock_t* ssock, pj_bool_t check_error)
+{
+    ossl_sock_t* ossock = (ossl_sock_t*)ssock;
```
sauwming, line 1703: Minor: our PointerAlignment is actually `Right`, i.e. see the initial code:
`ossl_sock_t *ossock = (ossl_sock_t *)ssock`

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1706,12 +1711,21 @@ static void ssl_reset_sock_state(pj_ssl_sock_t *ssock)
      * Avoid calling SSL_shutdown() if handshake wasn't completed.
      * OpenSSL 1.0.2f complains if SSL_shutdown() is called during an
      * SSL handshake, while previous versions always return 0.
+     * Don't send notify when the last error is SSL_ERROR_SYSCALL or SSL_ERROR_SSL.
```
sauwming, line 1714: Replace `send notify` with `call SSL_shutdown()`.

Yes, SSL_shutdown() will send notify, but it's only one of the steps.
`Note that SSL_shutdown() must not be called if a previous fatal error has occurred on a connection i.e. if SSL_get_error() has returned SSL_ERROR_SYSCALL or SSL_ERROR_SSL.

The shutdown procedure consists of two steps: sending of the close_notify shutdown alert, and reception of the peer's close_notify shutdown alert. The order of those two steps depends on the application.`
https://www.openssl.org/docs/manmaster/man3/SSL_shutdown.html

Also change the boolean below.

creslin2877:
The only concern I have is that last_err might be set from an ssl_write context while ssl_read is being executed (due to concerns brought up in #3575).  I just pushed my patch up for it, so we'll probably have a merge conflict to deal with whenever your branch or mine goes in, but in any case, I think the locking in the fix for #3575 will make things a bit better.

sauwming:
Since #3583 has been integrated, you can merge with the latest master to resolve the conflict.

creslin2877 (changes requested):
As mentioned in my previous comment, we need to cover the case where we get an SSL_ERROR_SSL from the SSL_read() function too.

sauwming:
> As mentioned in my previous comment, we need to cover the case where we get an SSL_ERROR_SSL from the SSL_read() function too.

Since, after merging, it now already uses the locking in `ssl_read()`, let us know if there are further modifications required.

sauwming:
@creslin2877, let us know if there's any issue with the latest patch, so we can merge it for 2.13.1.

sauwming:
I just realised that the addition of `ssl_reset_sock_state_with_error()` will break other SSL backends. Can we avoid modifying the internal API?

creslin2877:
I think you've covered the case that triggered my concern.  Thanks!

sauwming (approved):

nanangizz (approved):

Issue #2143, opened 2018-08-30
Investigate AEC info to be added into call info & statistics dump


pjsipbot:
_2018-08-30 05:59:22: @nanangizz_ _created the issue on trac ticket [2143](https://trac.pjsip.org/repos/ticket/2143)_

Thank you Jure Erznožnik for the suggestion.

AEC info will be added in pjsua_call_dump(). There's also a new API pjsua_get_ec_stat() to query the AEC statistics.

___
_2020-01-28 06:58:46: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2020-01-28 06:58:46: @mingteluu_ _set **resolution** to **fixed**_

___
_2020-01-28 06:58:46: @mingteluu_ _commented_

In [r6140](../commit/bde1167cc37d23e9720fadfd44d918c2bb6e9372):
Closed #2143: Add AEC info to call info & statistics dump

___
_2020-01-28 07:00:57: @mingteluu_ _edited the issue description_

PR #3582, opened 2023-05-24
Opus codec stat

vsedyshev-intermedia:
- OPUS codec stat summary
 - FreeSwitch code on FEC stat calculation 
 - WebRTC code on FEC stat calculation 
 - PJMEDIA stream provide codec API


sauwming:
Thank you for the contribution. We appreciate the submission of the patch.

Regrettably, I don't believe it is prepared for integration into the main codebase at this time. Here are some points to consider:

- The modifications made to Opus appear reasonable.
- However, the proposed new API in the stream (as well as the pjsua/pjsua2 API) require further refinement. For instance, the function `pjmedia_stream_get_stat_codec()`, as its name suggests, should encompass all codecs, not only Opus. If you'd like to check a reference example, you can refer to the Echo Cancellation stat implementation at https://github.com/pjsip/pjproject/issues/2143.
- If third-party code from FreeSwitch and WebRtc is to be included, appropriate acknowledgement of the source and inclusion of the original licensing terms are necessary.
- Minor changes unrelated to the pull request, such as config modifications, version.mak, and GitHub workflows, need not be included.

In summary, there is still significant work remaining before the patch can be incorporated into the library. If you wish to continue working on it, please let us know. However, if you prefer to keep it as it is and utilize it solely for your own application, that is also acceptable.

sauwming:
Will close this due to the above reasons. Please feel free to reopen if needed.

PR #3583, opened 2023-05-24
Locking fix so that SSL_shutdown and SSL_write are not called at same time

creslin2877:
Adds locking to prevent simultaneous execution of SSL_write and SSL_shutdown.  OpenSSL documentation says that multiple threads simultaneously calling SSL_* related functions on the same SSL object is not supported.

It'd be nice if this bug fix got into 2.13.x as well.

Fix for #3575

### DIFF
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -2452,9 +2452,6 @@ static pj_status_t ssl_read(pj_ssl_sock_t *ssock, void *data, int *size)
                 pj_lock_release(ssock->write_mutex);
                 /* Unfortunately we can't hold the lock here to reset all the state.
                   * We probably should though.
-                  *
-                  * TODO: if we get an SSL_ERROR_SSL or SSL_ERROR_SYSCALL, we need to not perform
-                  *       an SSL_shutdown in ssl_reset_sock_state according to openssl docs (due to some CVE or other)
                   */
                 ssl_reset_sock_state(ssock);
                 return status;
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -1695,7 +1695,15 @@ static void ssl_destroy(pj_ssl_sock_t *ssock)
 /* Reset SSL socket state */
 static void ssl_reset_sock_state(pj_ssl_sock_t *ssock)
 {
+    int post_unlock_flush_circ_buf = 0;
+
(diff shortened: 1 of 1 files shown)


pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -2436,15 +2449,25 @@ static pj_status_t ssl_read(pj_ssl_sock_t *ssock, void *data, int *size)
                 /* Reset SSL socket state, then return PJ_FALSE */
                 status = STATUS_FROM_SSL_ERR2("Read", ssock, size_,
                                               err, len);
+                pj_lock_release(ssock->write_mutex);
+                /* Unfortunately we can't hold the lock here to reset all the state.
+                  * We probably should though.
+                  *
+                  * TODO: if we get an SSL_ERROR_SSL or SSL_ERROR_SYSCALL, we need to not perform
```
sauwming, line 2456: Since we're going to immediately address this in the other PR, I suppose this comment can be removed?

sauwming (approved):

pjlib/src/pj/ssl_sock_ossl.c
creslin2877: Ok, just pushed a fix up that removes the TODO portion of the comment.

trengginas (approved):

creslin2877:
Any chance of this making it into a 2.13.x release?

sauwming:
Coincidentally, we have been planning to release 2.13.1 to include the recent security fixes: https://github.com/pjsip/pjproject/security/advisories

So yes, we can include this as well (and probably #3577 too?)

creslin2877:
That sounds good.  Just to update, so far I haven't seen any crashes in this code on our 30'ish Asterisk instances it's deployed to so far (and we'd have seen a few by now), so I think it's pretty safe.  This is also good, as it strongly looks like this bug was probably causing general heap/memory corruption as well, which can manifest itself as lots of weird bugs.

I didn't see any response in #3577 about not calling SSL_shutdown when reading an SSL_ERROR_SSL from an SSL_read() call (in the openssl adapter code for TLS connected SIP).  Should I make a separate code review to fix that issue or is @trengginas going to take care of that?

sauwming:
Sorry for the delay, I have merged your fix with #3577.

One more thing, I usually inform George Joseph or Joshua Colp prior to any release, so I take it you will be the one informing them?

creslin2877:
Probably yes - deepest apologies - I actually don't work at Sangoma/Digium anymore, so it might be best to do any pre-release discussions you typically have with them without me.

PR #3584, opened 2023-05-30
Xamarin samples for Android and iOS

trengginas:
This PR will add samples for Xamarin app on Android and iOS.

The Xamarin app will use PJSUA2 API with SWIG binding.

### Requirements
- Build requirements for Android and iOS.
  - Android : https://docs.pjsip.org/en/latest/get-started/android/build_instructions.html
  - iOS : https://docs.pjsip.org/en/latest/get-started/ios/build_instructions.html
- Microsoft Visual Studio ([ref](https://learn.microsoft.com/en-us/xamarin/cross-platform/get-started/requirements))

### Building PJSIP
- Android : https://docs.pjsip.org/en/latest/get-started/android/build_instructions.html#building-pjsip
- iOS : https://docs.pjsip.org/en/latest/get-started/ios/build_instructions.html#building-pjsip

### Trying our sample application and creating your own

**Building our Xamarin Android sample app**
- Make sure SWIG is in the build environment PATH.

- Run make from directory [pjsip-apps/src/swig](https://github.com/pjsip/pjproject/blob/master/pjsip-apps/src/swig) (note that the Android NDK root should be in the PATH), e.g:
  ```
  $ cd /path/to/your/pjsip/dir
  $ cd pjsip-apps/src/swig
  $ make
  ```
  This step should produce:
  - native library libpjsua2.so in pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin.Android/lib/arm64-v8a
  - pjsua2 C# interface (a lot of .cs files) in pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin/pjsua2

- Make sure any library dependencies are copied to pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin.Android/lib/arm64-v8a (or the appropriate target ABI directory), e.g: libopenh264.so for video support.

- Open pjsua2xamarin app solution in Visual Studio, it is located in pjsip-apps/src/swig/csharp/pjsua2xamarin.

- For video support, the make command will copy additional supporting .java files (PJCamera*.java). From the Visual Studio, right click the files and set the Build Action as Java Native Source. 

- Build the pjsua2xamarin.Android project

- Run it.

**Building our Xamarin iOS sample app**
- Make sure SWIG is in the build environment PATH.

- Run make from directory [pjsip-apps/src/swig](https://github.com/pjsip/pjproject/blob/master/pjsip-apps/src/swig) e.g:
  ```
  $ cd /path/to/your/pjsip/dir
  $ cd pjsip-apps/src/swig
  $ make
  ```
  This step should produce:
  - native library libpjsua2.so in pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin.iOS/lib/arm64-v8a
  - pjsua2 Java interface (a lot of .cs files) in pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin/pjsua2

- Make sure any library dependencies are copied to pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin.iOS/lib/arm64 (or the appropriate target ABI directory), e.g: libopenh264.so for video support.

- Open pjsua2xamarin app solution in Visual Studio, it is located in pjsip-apps/src/swig/csharp/pjsua2xamarin.

- Build the pjsua2xamarin.iOS project

- Run it.

**Creating your own application**

For developing Xamarin application, you should use [pjsua2 API](https://docs.pjsip.org/en/latest/api/pjsua2/ref.html) whose C# interface available via SWIG C# binding.

1. First, build pjproject libraries as described above.

1. Also build pjsua2xamarin SWIG binding, this step is required to generate the pjsua2 c# interface and the native library.

1. Create Xamarin application outside the PJSIP sources for your project.

1. Get pjsua2 C# interface and native library from pjsua2 sample application:

1. Copy pjsua2 C# interface files from pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin/pjsua2 to your project’s folder, e.g:
   ```
   $ cd $YOUR_PROJECT_DIR/
   $ cp -r $PJSIP_DIR/pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin/pjsua2 .
   ```

1. Copy native library libpjsua2.so and libc++_shared.so from pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin.[platform]/lib/ to your project’s libs folder:
   ```
   $ cd $YOUR_PROJECT_DIR/lib
   $ cp -r $PJSIP_DIR/{pjsip-apps/src/swig/csharp/pjsua2xamarin/pjsua2xamarin.[platform]/lib .
   ```

Notes:
- Android App
  - When adding a static library to the project, set the library as Java native libray as the build action. This includes the pjsua2 library and additional supporting library.
  - When adding supporting java files (e.g: PJCamera*.java), set the build action to Java native source.
  - Load the static library on your code using JavaSystem.LoadLibrary().
    ```
    JavaSystem.LoadLibrary("c++_shared");
    JavaSystem.LoadLibrary("pjsua2");    
    ```
- iOS App
  - From https://learn.microsoft.com/en-us/xamarin/ios/platform/native-interop
    ```
    Since you can only use static libraries on iOS, there is no external shared library to link with, so the path parameter in the 
    DllImport attribute needs to use the special name __Internal (note the double underscore characters at the start of the 
    name) as opposed to the path name.

    This forces DllImport to look up the symbol of the method that you are referencing in the current program, instead of 
    trying to load it from a shared library.
    ```
    Implications:
    - The C# interface files will be different from the one generated for Android. Make sure that you are using the correct C# interface files generated specific for the platform.
  - For iOS app, the framework used needed to be specified in the additional build arguments.
    ```
    -cxx -gcc_flags "-L${ProjectDir}/lib/arm64/ -lpjsua2 -force_load ${ProjectDir}/lib/arm64/libpjsua2.a -framework 
    CoreMedia -framework CoreAudio -framework CoreVideo -framework AVFoundation -framework AudioToolbox - 
    framework VideoToolbox -framework SystemConfiguration -framework Security"
    ```
- Since the app is build independently, it is recommended to disable/unload the other projects that is not intended to be built. This is to avoid any errors unrelated to build target. e.g: when building for Android, unload the pjsua2xamarin.iOS project.

Start writing your application, please check [pjsua2 docs](http://www.pjsip.org/docs/book-latest/html/index.html) for reference.


sauwming:
- Some of the indentation seems to be incorrect.
- Bitrise fails
- How big is the size of this sample project?

How about the video? Is it smooth or is there any noticeable lag?

For the audio, since it happens on both iOS and Android, it's unlikely to be caused by a particular audio backend. So perhaps the audio troubleshooting can help? Such as:
- Is the audio loopback okay?
- How about the CPU usage?
- Does the audio issue worsen if there's video?

trengginas:
> * Some of the indentation seems to be incorrect.
- Just fixed the indentation.
> * Bitrise fails
- I checked here locally and it builds (Android, iOS, mac) without issue. The report suggested SWIG build failed, but I cannot check the detail from bitrise.
> * How big is the size of this sample project?
- The added source is ~3Mb
> 
> How about the video? Is it smooth or is there any noticeable lag?
- Didn't notice any issue
> 
> For the audio, since it happens on both iOS and Android, it's unlikely to be caused by a particular audio backend. So perhaps the audio troubleshooting can help? Such as:
> 
> * Is the audio loopback okay?
- I can hear stutters on audio loopback
> * How about the CPU usage?
> * Does the audio issue worsen if there's video?

- The audio issue is noticeable without video. With video, it doesn't get any worse. Trying to use the ExtraAudioDevice to fix the timing issue and it seems much better.

sauwming:
> > * Bitrise fails
> * I checked here locally and it builds (Android, iOS, mac) without issue. The report suggested SWIG build failed, but I cannot check the detail from bitrise.

I tested locally and the SWIG build fails if executed from the default directory (i.e. `pjsip-apps/src/swig`).

sauwming (approved):

sauwming:
So what happens with extra audio device setup? Does it work fine now without it? What was the issue then?

trengginas:
> So what happens with extra audio device setup? Does it work fine now without it? What was the issue then?

I retested and use the normal sound device, and the audio works fine (no stutters). My guess is there's a network issue on the first test. There are few access points (one with weak signal) here on the environment. If one of the devices connected the weak access point, media issue is observable. On the latest test, I make sure both endpoints are using the normal/good access point.

PR #3590, opened 2023-06-05
Add option for dual stack IPv4&IPv6 account config

### LINKED ISSUE #2146: Revisit IPv4/IPv6 settings and behavior in pjsua
_2018-09-05 03:51:22: @trengginas_ _created the issue on trac ticket [2146](https://trac.pjsip.org/repos/ticket/2146)_

This a continuation from #2078

Currently, in the account settings, SIP and media has to be predetermined whether it wants to use a particular IP version, IPv4 or IPv6. As a result, typically two accounts are required for each IP version, one for IPv4, and one for IPv6.

It is desirable that a more flexible/smart detection mechanism is employed, such as:

- Ability to handle both IP versions in SIP and media, especially in the case of incoming offer where we can check what the remote wants.
- Specify a preference of one IP version over the other, but still can handle both.
- Match the IP version for SIP and media, i.e. if SIP signalling is done via IPv6, then media will also use IPv6.

Thanks to Alexander Traud for the suggestion.

sauwming:
To close #2146 .

Add the following options in pjsua for SIP/media account config:
```
/**
 * Specify how IPv6 transport should be used in account config.
 * IP version preference only applies for outgoing direction, for incoming
 * direction, we will check the corresponding message/offer and match it.
 */
typedef enum pjsua_ipv6_use
{
    /* IPv6 is not used. */
    PJSUA_IPV6_DISABLED,
    /* IPv6 is enabled. */
    PJSUA_IPV6_ENABLED = 1,
    PJSUA_IPV6_ENABLED_NO_PREFERENCE = 1,
    /* IPv6 is enabled, but IPv4 is preferable.*/
    PJSUA_IPV6_ENABLED_PREFER_IPV4,
    /*IPv6 is enabled and preferable. */
    PJSUA_IPV6_ENABLED_PREFER_IPV6,
    /* Only IPv6 is enabled, IPv4 will not be used. */
    PJSUA_IPV6_ENABLED_USE_IPV6_ONLY
} pjsua_ipv6_use;
```

And add field `pjsua_acc_config.ipv6_sip_use` in addition to the already existing `ipv6_media_use`.

For backward compatibility and interoperability reasons, the default settings of the fields are:
- `ipv6_sip_use`: `PJSUA_IPV6_ENABLED_NO_PREFERENCE`, which means that for SIP transport, it will depend on the address resolution result returned by the OS (`pj_getaddrinfo()`) or by the resolver. Note that this is the current behavior.
- `ipv6_media_use`: `PJSUA_IPV6_ENABLED_PREFER_IPV4`, which means that outgoing offer will still use IPv4, but it's capable of receiving IPv6 media offer without any additional modification. For interoperability reason, IPv4 preference is recommended since not all endpoints support IPv6.

Note: IPv6 usage setting may not apply for ICE media transport, depending on STUN configuration. This behavior is observable before this PR. We may need to modify ICE media transport IPv6 usage spec in the future, which is beyond the scope of this PR.

### DIFF
--- a/pjsip/include/pjsip/sip_transport.h
+++ b/pjsip/include/pjsip/sip_transport.h
@@ -213,14 +213,22 @@ typedef enum pjsip_tpselector_type
  */
 typedef enum pjsip_tpselector_ip_ver
 {
+    /** IPv4 only. */
     PJSIP_TPSELECTOR_USE_IPV4_ONLY,
 
+    /**
+     * No preference. IP version used will depend on the order of addresses
+     * returned by pjsip_resolver.
+     */
     PJSIP_TPSELECTOR_NO_PREFERENCE,
 
+    /** IPv4 is preferred. */
     PJSIP_TPSELECTOR_PREFER_IPV4,
 
+    /** IPv6 is preferred. */
     PJSIP_TPSELECTOR_PREFER_IPV6,
 
+    /** IPv6 only. */
     PJSIP_TPSELECTOR_USE_IPV6_ONLY
 
 } pjsip_tpselector_ip_ver;
@@ -260,7 +268,10 @@ typedef struct pjsip_tpselector
      */
     pj_bool_t disable_connection_reuse;
 
-    /** Union representing the transport/listener criteria to be used. */
+    /**
+     * Union representing the transport/listener/IP version criteria
+     * to be used.
+     */
     union {
         pjsip_transport         *transport;
         pjsip_tpfactory         *listener;
--- a/pjsip/include/pjsip/sip_transport.h
+++ b/pjsip/include/pjsip/sip_transport.h
@@ -262,10 +262,10 @@ typedef struct pjsip_tpselector
 
     /** Union representing the transport/listener criteria to be used. */
     union {
-        pjsip_transport *transport;
-        pjsip_tpfactory *listener;
-        void            *ptr;
-        pjsip_tpselector_ip_ver ip_ver;
+        pjsip_transport         *transport;
+        pjsip_tpfactory         *listener;
+        pjsip_tpselector_ip_ver  ip_ver;
+        void                    *ptr;
     } u;
 
 } pjsip_tpselector;
--- a/pjsip/include/pjsip/sip_transport.h
+++ b/pjsip/include/pjsip/sip_transport.h
@@ -203,7 +203,24 @@ typedef enum pjsip_tpselector_type
     /** Use the specific listener to send request. */
     PJSIP_TPSELECTOR_LISTENER,
 
+    /** Use the specific listener to send request. */
+    PJSIP_TPSELECTOR_CONFIG,
+
 } pjsip_tpselector_type;
+
+typedef enum pjsip_tpselector_config
+{
+    PJSIP_TPSELECTOR_USE_IPV4_ONLY,
+
+    PJSIP_TPSELECTOR_NO_PREFERENCE,
+
+    PJSIP_TPSELECTOR_PREFER_IPV4,
+
+    PJSIP_TPSELECTOR_PREFER_IPV6,
+
+    PJSIP_TPSELECTOR_USE_IPV6_ONLY
+
+} pjsip_tpselector_config;
 
 
 /**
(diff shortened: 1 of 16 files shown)


pjsip/include/pjsip/sip_transport.h
```diff
@@ -233,6 +250,8 @@ typedef struct pjsip_tpselector
     /** The type of data in the union */
     pjsip_tpselector_type   type;
 
+    pjsip_tpselector_config config;
+
```
nanangizz, line 254: - perhaps this can be categorized as criteria, so it can be moved to be inside `union u`?
- also, term `config` is perhaps too broad and currently the criteria is an enum (while `config` is more like a struct that can be expanded in the future), so what if using term like `ip_ver` instead?

pjsip/include/pjsip/sip_transport.h
```diff
@@ -245,6 +265,7 @@ typedef struct pjsip_tpselector
         pjsip_transport *transport;
         pjsip_tpfactory *listener;
         void            *ptr;
+        pjsip_tpselector_ip_ver ip_ver;
```
nanangizz, line 268: minor: `ptr` seems to be the least important field in the union (for generic/unclear purpose :D ), so perhaps move it to the last?

nanangizz (approved):
just a very minor proposal below.

sauwming:
Final testing before merging.

Test environment: dual stack Mac OS

SIP registration tests:
1. Scenario 1: normal registration with default setting, no IPv6 SIP transports created, getaddrinfo() returns IPv6 first
Pre patch: Initial reg will fail to get contact source port (due to no IPv6 transport)
[normal_pre_patch.txt](https://github.com/pjsip/pjproject/files/11701879/normal_pre_patch.txt)
Post patch: No issue (since there is fallback mechanism, if IPv6 fails use IPv4, and vice versa)
[normal_after_patch.txt](https://github.com/pjsip/pjproject/files/11701874/normal_after_patch.txt)

2. Scenario 2: normal registration + IPv6 SIP transports
NO_PREFERENCE: getaddrinfo() returns IPv6 first, IPv6 will be used
[no_pref_ipv6.txt](https://github.com/pjsip/pjproject/files/11701886/no_pref_ipv6.txt)
NO_PREFERENCE: getaddrinfo() returns IPv4 first, IPv4 will be used
[no_pref_ipv4.txt](https://github.com/pjsip/pjproject/files/11701888/no_pref_ipv4.txt)
PREF_IPV4: getaddrinfo() returns IPv4 first, IPv4 will be used
[prefer_ipv4.txt](https://github.com/pjsip/pjproject/files/11701894/prefer_ipv4.txt)
PREF_IPV6: getaddrinfo() returns IPv4 first, IPv6 will be used
[prefer_ipv6.txt](https://github.com/pjsip/pjproject/files/11701895/prefer_ipv6.txt)

2. Scenario 2: normal registration + IPv6 SIP transports, use nameserver
NO_PREF, PREF_IPV6: initial reg use IPv4 Contact and Via (note that contact_use_src_port doesn't work if nameserver is configured), nameserver returns IPv6 as priority, so reg sent using IPv6 transport, then contact update will correct the addresses to IPv6
[pref_ipv6_ns.txt](https://github.com/pjsip/pjproject/files/11701897/pref_ipv6_ns.txt)
PREF_IPV4: nameserver returns IPv6 as priority, reg sent using IPv4 transport
[pref_ipv4_ns.txt](https://github.com/pjsip/pjproject/files/11701899/pref_ipv4_ns.txt)

Media tests:
Caller pref IPv6 vs callee pref IPv4
[caller_pref_ipv6.txt](https://github.com/pjsip/pjproject/files/11701918/caller_pref_ipv6.txt)
[callee_pref_ipv4.txt](https://github.com/pjsip/pjproject/files/11701919/callee_pref_ipv4.txt)
Caller pref IPv6 vs callee pref IPv4, using TLS and SRTP
[caller_pref_ipv6_tls_srtp.txt](https://github.com/pjsip/pjproject/files/11701921/caller_pref_ipv6_tls_srtp.txt)

PR #3594, opened 2023-06-12
Check STUN tdata before sending the message

trengginas:
It is reported that a crash might happen on `stun_tsx_on_send_msg()`
```
    0x100fb2118 <+28>:  bl     0x100fb5794               ; pj_stun_client_tsx_get_data
->  0x100fb2124 <+40>:  ldr    x8, [x8, #0x18]
    0x100fb2134 <+56>:  bl     0x100ebfac4               ; pj_grp_lock_acquire
    0x100fb2150 <+84>:  bl     0x100ebfb0c               ; pj_grp_lock_release
```

```
pjsua_0 (14)#0	0x0000000100fb2124 in stun_tsx_on_send_msg ()
#1	0x0000000100fb5d64 in tsx_transmit_msg ()
#2	0x0000000100fb527c in retransmit_timer_callback ()
#3	0x0000000100ed48d0 in pj_timer_heap_poll ()
#4	0x00000001011d73e8 in pjsip_endpt_handle_events2 ()
#5	0x000000010118cfac in pjsua_handle_events ()
#6	0x000000010118c3c0 in worker_thread ()
#7	0x0000000100eb2f28 in thread_main ()
#8	0x00000001da496060 in _pthread_start ()
```

The log:
```
00:16:55.995        utsx0x1068f5628  .STUN client transaction created
00:16:55.995        utsx0x1068f5628  .STUN sending message (transmit count=1)
00:16:56.656         stun_session.c  .tdata 0x1068f54a8 destroy request, force=0, tsx=0x1068f5628, destroying=0
00:16:56.657        utsx0x1068f5628  .STUN transaction 0x1068f5628 schedule destroy
00:16:56.958        utsx0x1068f5628 !STUN client transaction 0x1068f5628 stopped, ref_cnt=15
00:16:57.321         stun_session.c !tdata 0x1068f54a8 destroy request, force=1, tsx=0x1068f5628, destroying=0
00:16:57.321        utsx0x1068f5628  STUN client transaction 0x1068f5628 stopped, ref_cnt=12
00:16:57.321         stun_session.c  STUN transaction 0x1068f5628 destroyed
00:16:57.321        utsx0x1068f5628 !STUN sending message (transmit count=2)
```

This indicates that while STUN transaction is destroyed.

https://github.com/pjsip/pjproject/blob/9a7d4d17202fb7821f3fbf1736c4f1468bdacb36/pjnath/src/pjnath/stun_session.c#L145-L163

The retransmit callback is called but blocked waiting for the group lock:
https://github.com/pjsip/pjproject/blob/9a7d4d17202fb7821f3fbf1736c4f1468bdacb36/pjnath/src/pjnath/stun_transaction.c#L347-L354

Afterwards, the tdata was reset to NULL and after the retransmit will resume and crash caused by invalid tdata.

### DIFF
--- a/pjnath/src/pjnath/stun_session.c
+++ b/pjnath/src/pjnath/stun_session.c
@@ -507,6 +507,9 @@ static pj_status_t stun_tsx_on_send_msg(pj_stun_client_tsx *tsx,
     pj_status_t status;
 
     tdata = (pj_stun_tx_data*) pj_stun_client_tsx_get_data(tsx);
+    if (!tdata)
+        return PJ_EGONE;
+     
     sess = tdata->sess;
 
     /* Lock the session and prevent user from destroying us in the callback */
--- a/pjnath/src/pjnath/stun_transaction.c
+++ b/pjnath/src/pjnath/stun_transaction.c
@@ -47,6 +47,7 @@ struct pj_stun_client_tsx
     pj_timer_heap_t     *timer_heap;
 
     pj_timer_entry       destroy_timer;
+    pj_bool_t            is_destroying;
 
     void                *last_pkt;
     unsigned             last_pkt_size;
@@ -109,7 +110,12 @@ PJ_DEF(pj_status_t) pj_stun_client_tsx_schedule_destroy(
     PJ_ASSERT_RETURN(tsx && delay, PJ_EINVAL);
     PJ_ASSERT_RETURN(tsx->cb.on_destroy, PJ_EINVAL);
 
+    if (tsx->is_destroying)
+        return PJ_SUCCESS;
+
     pj_grp_lock_acquire(tsx->grp_lock);
+
+    tsx->is_destroying = PJ_TRUE;
 
     /* Cancel previously registered timer */
     pj_timer_heap_cancel_if_active(tsx->timer_heap, &tsx->destroy_timer,
@@ -291,6 +297,11 @@ PJ_DEF(pj_status_t) pj_stun_client_tsx_send_msg(pj_stun_client_tsx *tsx,
 
     pj_grp_lock_acquire(tsx->grp_lock);
 
+    if (tsx->is_destroying) {
+        pj_grp_lock_release(tsx->grp_lock);
+        return PJ_SUCCESS;
+    }
+
     /* Encode message */
     tsx->last_pkt = pkt;
     tsx->last_pkt_size = pkt_len;
@@ -353,6 +364,11 @@ static void retransmit_timer_callback(pj_timer_heap_t *timer_heap,
     PJ_UNUSED_ARG(timer_heap);
     pj_grp_lock_acquire(tsx->grp_lock);
 
+    if (tsx->is_destroying) {
+        pj_grp_lock_release(tsx->grp_lock);
+        return;
+    }
+
     if (tsx->transmit_count >= PJ_STUN_MAX_TRANSMIT_COUNT) {
         /* tsx may be destroyed when calling the callback below */
         pj_grp_lock_t *grp_lock = tsx->grp_lock;
@@ -398,6 +414,9 @@ PJ_DEF(pj_status_t) pj_stun_client_tsx_retransmit(pj_stun_client_tsx *tsx,
     if (tsx->destroy_timer.id != 0) {
         return PJ_SUCCESS;
     }
+
+    if (tsx->is_destroying)
+        return PJ_SUCCESS;
 
     if (mod_count) {
         pj_timer_heap_cancel_if_active(tsx->timer_heap, &tsx->retransmit_timer,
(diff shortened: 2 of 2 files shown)


sauwming:
I don't think checking `tdata` will be helpful if the STUN tsx itself has been destroyed, since `retransmit_timer_callback()` will still access tsx.

I believe we need to make sure that the timer is really cancelled, since:
`pj_stun_client_tsx_schedule_destroy()` and `pj_stun_client_tsx_stop()` will both call:
`    pj_timer_heap_cancel_if_active(tsx->timer_heap, &tsx->retransmit_timer,
                                   TIMER_INACTIVE);`

pjnath/src/pjnath/stun_session.c
```diff
@@ -507,6 +507,9 @@ static pj_status_t stun_tsx_on_send_msg(pj_stun_client_tsx *tsx,
     pj_status_t status;
 
     tdata = (pj_stun_tx_data*) pj_stun_client_tsx_get_data(tsx);
+    if (!tdata)
+        return PJ_EGONE;
+
```
nanangizz, line 512: Looks like a race condition happened. What if the context switching happens after the `tdata` validation (just added by this patch) instead of before?

trengginas: The method is called within a grup lock
https://github.com/pjsip/pjproject/blob/9a7d4d17202fb7821f3fbf1736c4f1468bdacb36/pjnath/src/pjnath/stun_transaction.c#L347-L354

and 

https://github.com/pjsip/pjproject/blob/9a7d4d17202fb7821f3fbf1736c4f1468bdacb36/pjnath/src/pjnath/stun_session.c#L145-L163

It shouldn't context switch within the group lock.

trengginas:
> I don't think checking `tdata` will be helpful if the STUN tsx itself has been destroyed, since `retransmit_timer_callback()` will still access tsx.
> 
> I believe we need to make sure that the timer is really cancelled, since: `pj_stun_client_tsx_schedule_destroy()` and `pj_stun_client_tsx_stop()` will both call: ` pj_timer_heap_cancel_if_active(tsx->timer_heap, &tsx->retransmit_timer, TIMER_INACTIVE);`

I think the callback is already called but blocked in  
https://github.com/pjsip/pjproject/blob/9a7d4d17202fb7821f3fbf1736c4f1468bdacb36/pjnath/src/pjnath/stun_transaction.c#L347-L354

Once the callback acquire the group lock, it gets the tdata and crash if it's already set to NULL.

pjnath/src/pjnath/stun_session.c
nanangizz: Ah okay.

Just curious, it seems that theoritically they can be using different locks, one is tsx's and the other is sess's. So, what if the tdata query & check is moved down to inside the sess's lock, will it improve safety or even introduce some risk?

nanangizz:
> I think the callback is already called but blocked in
> Once the callback acquire the group lock, it gets the tdata and crash if it's already set to NULL.

IMO, perhaps better to add this scenario into the PR desc too?

pjnath/src/pjnath/stun_session.c
trengginas: We can't get the sess's lock without getting the tdata reference first. In this case we can't move the query/check

nanangizz (approved):

sauwming:
> > I think the callback is already called but blocked in

This is a recurrent problem with our timer implementation :(
Cancelling can race with the callback being called.
We probably need to improve our timer, but this will be a much more difficult task.

The current proposal still has a few problems:
1. The race can still happen
True that `retransmit_timer_callback()` is holding the tsx group lock, but as pointed out by @nanangizz, the other thread destroying the tsx can be holding a different lock, so they both can run together at the same time (as shown in the log where both have the exact timestamp `00:16:57.321`).
2. Another retransmit timer can get scheduled even though tsx has been scheduled to be destroyed.

A simpler fix is perhaps to just use `is_destroying` flag?
i.e. in `pj_stun_client_tsx_schedule_destroy()`, set `tsx->is_destroying = PJ_TRUE`

in `retransmit_timer_callback()`, `pj_stun_client_tsx_send_msg()`, and `pj_stun_client_tsx_retransmit()`: `if (tsx->is_destroying) return`

pjnath/src/pjnath/stun_transaction.c
```diff
@@ -399,6 +415,9 @@ PJ_DEF(pj_status_t) pj_stun_client_tsx_retransmit(pj_stun_client_tsx *tsx,
         return PJ_SUCCESS;
```
sauwming, line 415: minor: just remove this (or combine it with the below check) since it's more or less the same.

sauwming (approved):

Issue #3514, opened 2023-04-19
Documentation of TransportConfig::publicAddress unclear


wosrediinanatour:
### Describe the bug

The documentation says: "Optional address to advertise as the address of this transport".

It is not clear from the documentation what advertise actually mean, e.g. whether this will be used in the contact header of SIP messages.

### Steps to reproduce

(documentation only)

### PJSIP version

2.13

### Context

Does this setting influence the contact header of SIP messages?

### Log, call stack, etc

```shell
(documentation only)
```

sauwming:
In the context of network communication, "advertising" refers to making a network address or endpoint known to other entities on the network to allow them to reach/communicate with you using that address.

The field `publicAddress` you mention is a generic transport config, it can be used for SIP or media transport. So If you configure it for SIP transport, then yes, it means that the Contact address will be using that address. If for media transport, then it will be advertised in the SDP to allow remote endpoint to send media to you using that address.

wosrediinanatour:
@sauwming: Hence, if I set the `TransportConfig`'s `publicAddress` to an IP and apply it via `Endpoint::transportCreate(PJSIP_TRANSPORT_UDP, tcfg);`, it should be used in the contacts header of outgoing SIP REGISTERs and SIP INVITEs? If yes, I have to make a bug report...

sauwming:
It looks like you're right. #2205 only checks and uses the bound address if configured, but doesn't check for public address.

sauwming:
@wosrediinanatour , please check #3595 and let us know if it resolves the issue.

wosrediinanatour:
> @wosrediinanatour , please check #3595 and let us know if it resolves the issue.

I will! (continuing in https://github.com/pjsip/pjproject/pull/3595)

wosrediinanatour:
Follow-up question: @sauwming, is there a reason the contact header is of a SIP response is not set to the `publicAddress`? Or is it only  missing?

Issue #2205, opened 2019-05-29
Use bound address for outgoing UDP SIP messages


pjsipbot:
_2019-05-29 09:01:42: @mingteluu_ _created the issue on trac ticket [2205](https://trac.pjsip.org/repos/ticket/2205)_

For TCP&TLS, we already have ticket #1575, which will use the configured bind address for outgoing messages. However, for UDP, this is not the case. So even though the user has specified the bound_addr when calling pjsua_transport_create(UDP,...), the outgoing messages may still use different Via and Contact addresses.

___
_2019-05-29 09:02:05: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2019-05-29 09:02:05: @mingteluu_ _set **owner** to **@mingteluu**_

___
_2019-05-29 09:02:05: @mingteluu_ _set **resolution** to **fixed**_

___
_2019-05-29 09:02:05: @mingteluu_ _commented_

In [r6015](../commit/58593ba22cee4241d9aaf224af9b93d167057341):
Fixed #2205: Use bound address for outgoing UDP SIP messages

PR #3595, opened 2023-06-14
Use configured bound or public address for outgoing UDP SIP messages

sauwming:
As reported in #3514, the public address specified in the config to create UDP transport may not be used in the outgoing messages.

Tracing back, we already have #2205 which will use the configured bound address, but we haven't done so for public address.


trengginas (approved):

wosrediinanatour:
First test using the sample app looks promising. I will use it in my app within the next days.

PR #3599, opened 2023-06-15
Only validate public address when IP address is supplied on TCP/TLS

trengginas:
When public address is supplied when starting TCP/TLS transport(e.g: `pjsip_tcp_transport_start2()`), the address will be validated. 
The validation is not applied to UDP transport.
If the address supplied is unresolvable FQDN address, the transport will fail to start.
This PR will only validate the publish address if it is an IP address.


sauwming:
Also update the doc of `cfg.addr_name` and `a_name` to mention that it can be hostname.

Make sure to test that it works well. Also, I'm curious, how does Contact and Via rewrite work with FQDN published address?

sauwming (approved):

nanangizz (approved):

Issue #1943, opened 2016-07-11
AES-GCM crypto support for SRTP


pjsipbot:
_2016-07-11 08:44:06: @mingteluu_ _created the issue on trac ticket [1943](https://trac.pjsip.org/repos/ticket/1943)_

For GNU targets:

1. AES GCM crypto requires OpenSSL, so make sure OpenSSL is enabled. Also this requires PJSIP 2.5, as this version has newer libsrtp version. Then make sure that AES GCM is supported by `libcrypto` by checking the `configure` output:
```
checking for EVP_aes_128_gcm in -lssl... yes
OpenSSL has AES GCM support, SRTP will use OpenSSL
```

2. Enable the desired AES GCM cryptos in your config_site.h (by default, it's not enabled), for example:
```
    #define PJMEDIA_SRTP_HAS_AES_GCM_256      1
    #define PJMEDIA_SRTP_HAS_AES_GCM_128      1
```

3. Rebuild PJSIP.

For Windows:

1. [Install OpenSSL](https://trac.pjsip.org/repos/wiki/Getting-Started/Installing-OpenSSL-Windows).

2. In Visual Studio, go to `libsrtp` project, into folder `crypto`.

3. Multiple-select `cipher/aes_gcm_ossl.c, cipher/aes_icm_ossl.c, hash/hmac_ossl.c, rng/rand_source_ossl.c` <BR />
 Right-click, then click Properties. Change Configuration to `"All Configurations"` and Platform to `"All Platforms"`. Go to `Configuration Properties - General`, then change `Excluded From Build` from "Yes" to "No".

4. Multiple-select `cipher/aes.c cipher/aes_icm.c cipher/aes_cbc.c hash/sha1.c hash/hmac.c rng/rand_source.c rng/prng.c rng/ctr_prng.c` <BR />
 Right-click, then click Properties. Change Configuration to `"All Configurations"` and Platform to `"All Platforms"`. Go to `Configuration Properties - General`, then change `Excluded From Build` from "No" to "Yes".

5. Click on libsrtp project. <BR />
 Right-click, then click Properties. Change Configuration to `"All Configurations"` and Platform to `"All Platforms"`. Go to `Configuration Properties - C/C++ - Preprocessor`, then choose Preprocessor Definitions, and add `";OPENSSL"`.

6. Enable the desired AES GCM cryptos in your config_site.h, for example:
```
    #define PJMEDIA_SRTP_HAS_AES_GCM_256      1
[...]

PR #3603, opened 2023-06-19
Support OpenSSL3 with "no-deprecated" build option

trengginas:
When building OpenSSL with "no-deprecated" option, the library will fail due to use of deprecated method.
- `SSL_get_peer_certificate()`
- `X509_get_notBefore()`
- `X509_get_notAfter()`

This PR will fix the error by using the recommended/updated method.

This PR will also update the bundled `libsrtp` since it might also use OpenSSL.

Due to the changes, some updates also needed to support AES-GCM in SRTP (#1943).
The updates are required in Visual Studio (no changes on GNU targets).

1. [Install OpenSSL](https://docs.pjsip.org/en/latest/get-started/windows/build_instructions.html#installing-openssl).

1. In Visual Studio, go to `libsrtp` project, into folder crypto.

1. Multiple select `cipher/aes_gcm_ossl.c, cipher/aes_icm_ossl.c` 

   Right-click, then click Properties. Change Configuration to "All Configurations" and Platform to "All Platforms". Go to 
   Configuration Properties - General, then change Excluded From Build from "Yes" to "No".

1. Select `cipher/aes_icm.c`

   Right-click, then click Properties. Change Configuration to "All Configurations" and Platform to "All Platforms". Go to 
   Configuration Properties - General, then change Excluded From Build from "No" to "Yes".

1. Click on `libsrtp` project.

   Right-click, then click Properties. Change Configuration to "All Configurations" and Platform to "All Platforms". Go to 
   Configuration Properties - C/C++ - Preprocessor, then choose Preprocessor Definitions, and add `";OPENSSL;GCM"`.

1. Enable the desired AES GCM cryptos in your config_site.h, for example:
    ```
    #define PJMEDIA_SRTP_HAS_AES_GCM_256      1
    #define PJMEDIA_SRTP_HAS_AES_GCM_128      1
    ```

1. Rebuild `libsrtp`.


third_party/srtp/PJSIP_NOTES
```diff
@@ -0,0 +1,28 @@
+Local changes:
+1. Error build "Platform not recognized"
```
sauwming, line 2: need to be more specific here. On which platform(s) did it fail?

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -100,6 +100,17 @@
 
 #   include <openssl/obj_mac.h>
 
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L || \
+    (defined(OPENSSL_API_COMPAT) && OPENSSL_API_COMPAT >= 0x10100000L)
+#  define X509_get_notBefore(x)     X509_getm_notBefore(x)
```
sauwming, line 105: There are similar definitions below, so perhaps combine them, to make sure there's no redefinition warning/error?

sauwming:
> Due to the changes, some updates also needed to support AES-GCM in SRTP (#1943). The updates are required in Visual Studio (no changes on GNU targets).
> 
> 1. [Install OpenSSL](https://docs.pjsip.org/en/latest/get-started/windows/build_instructions.html#installing-openssl).
> 2. In Visual Studio, go to `libsrtp` project, into folder crypto.
> 3. Multiple select `cipher/aes_gcm_ossl.c, cipher/aes_icm_ossl.c`
>    Right-click, then click Properties. Change Configuration to "All Configurations" and Platform to "All Platforms". Go to
>    Configuration Properties - General, then change Excluded From Build from "Yes" to "No".
> 4 ....

Are all these steps supposed to be done by user??

third_party/srtp/PJSIP_NOTES
trengginas: The reason is build error on every platform.
On `datatypes.h` there's a check for certain config. 
(https://github.com/pjsip/pjproject/blob/affc21f07cf610a861b3da82b96e41542b9bb4b4/third_party/srtp/crypto/include/datatypes.h#L57-L63).

The current code doesn't check if those configs are not defined.
https://github.com/pjsip/pjproject/blob/a5023f7fdd4506808e3cc50da2fdbf7b47bdda0b/third_party/srtp/crypto/include/datatypes.h#L57-L63

These configs are defined in `srtp_config.h`. 
Alternatively we can add `#include <srtp_config.h>` to `transport_srtp.c`
https://github.com/pjsip/pjproject/blob/affc21f07cf610a861b3da82b96e41542b9bb4b4/pjmedia/src/pjmedia/transport_srtp.c#L101

This way no changes to datatypes.h is needed

sauwming (approved):

nanangizz (approved):

PR #3609, opened 2023-06-26
Add custom headers to invite/resonse in pjsua

joeygo:
Add custom invite/response headers to pjsua app
  --add-invite-header header:value            Add custom invite header
  --add-response-header header:value     Add custom response header


sauwming:
Thank you for the patch.

A couple of notes:
- We need the CLA to be signed
- Unnecessary whitespace modification is better avoided, to make review more compact and simpler.

PR #3612, opened 2023-06-27
Support Python3 for github CI

trengginas:
Github CI no longer support Python2.7 ([ref](https://github.com/actions/setup-python/issues/672)), making some tests failing.
This PR will update the unit tests using Python3, it will no longer work on Python2.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -32,10 +32,10 @@ jobs:
       run: make
     - name: swig bindings
       run: cd pjsip-apps/src/swig && make
-    - name: set up Python 2.7 for pjsua test
+    - name: set up Python 3.10 for pjsua test
       uses: actions/setup-python@v2
       with:
-        python-version: 2.7
+        python-version: '3.10'
     - name: unit tests
       run: make pjlib-test-ci pjlib-util-test pjmedia-test pjsua-test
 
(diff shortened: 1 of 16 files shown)


sauwming:
The CI tests still failed.

nanangizz:
There are a lot of warnings:
`/home/runner/work/pjproject/pjproject/tests/pjsua/run.py:2: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
  import imp`

This may be useful: https://stackoverflow.com/questions/69884878/replacing-imp-with-importlib-maintaining-old-behavior

nanangizz:
In CI Windows:
- `UnicodeEncodeError: 'charmap' codec can't encode character '\x80' in position 482: character maps to <undefined>`
- Some failures in initiating/running pjsua.

Saw some encoding errors in CI Linux too.

sauwming (approved):

nanangizz (approved):
pjsua tests in CI Windows produce about 6x errors (out of 2xx tests), mostly about failure in starting pjsua and `FileNotFoundError: [WinError 2] The system cannot find the file specified`. But as it seems to also happens before this Python version upgrade, I think it is not within the scope of this PR.

trengginas:
> pjsua tests in CI Windows produce about 6x errors (out of 2xx tests), mostly about failure in starting pjsua and `FileNotFoundError: [WinError 2] The system cannot find the file specified`. But as it seems to also happens before this Python version upgrade, I think it is not within the scope of this PR.

The failure related to sipp tests and AMRNB codec use. sipp tests will fail because it needs cygwin to run on windows. AMRNB build fails in windows.

PR #3619, opened 2023-07-02
Add custom headers to invite/response in pjsua

joeygo:
Add custom headers to invite/200 OK response in pjsua.
Those parameters can be called more than once.
To add a custom invite header append the following to pjsua --add-invite-header "header:value"
To add a custom response header, append the following to pjsua --add-response-header "header:value"

### DIFF
--- a/pjsip-apps/src/pjsua/pjsua_app_common.h
+++ b/pjsip-apps/src/pjsua/pjsua_app_common.h
@@ -92,12 +92,12 @@ typedef struct pjsua_app_config
     pjsua_buddy_config      buddy_cfg[PJSUA_MAX_BUDDIES];
 
     unsigned                inv_hdr_cnt;
-    pj_str_t                inv_hname_cfg[16];
-    pj_str_t                inv_hvalue_cfg[16];
+    pj_str_t                inv_hname_cfg[PJSUA_MAX_CUSTOM_HEADERS];
+    pj_str_t                inv_hvalue_cfg[PJSUA_MAX_CUSTOM_HEADERS];
 
     unsigned                res_hdr_cnt;
-    pj_str_t                res_hname_cfg[16];
-    pj_str_t                res_hvalue_cfg[16];
+    pj_str_t                res_hname_cfg[PJSUA_MAX_CUSTOM_HEADERS];
+    pj_str_t                res_hvalue_cfg[PJSUA_MAX_CUSTOM_HEADERS];
 
     app_call_data           call_data[PJSUA_MAX_CALLS];
 
--- a/pjsip-apps/src/pjsua/pjsua_app_common.h
+++ b/pjsip-apps/src/pjsua/pjsua_app_common.h
@@ -91,6 +91,14 @@ typedef struct pjsua_app_config
     unsigned                buddy_cnt;
     pjsua_buddy_config      buddy_cfg[PJSUA_MAX_BUDDIES];
 
+    unsigned                inv_hdr_cnt;
+    pj_str_t                inv_hname_cfg[16];
+    pj_str_t                inv_hvalue_cfg[16];
+
+    unsigned                res_hdr_cnt;
+    pj_str_t                res_hname_cfg[16];
+    pj_str_t                res_hvalue_cfg[16];
+
     app_call_data           call_data[PJSUA_MAX_CALLS];
 
     pj_pool_t              *pool;
--- a/pjsip-apps/src/pjsua/pjsua_app.c
+++ b/pjsip-apps/src/pjsua/pjsua_app.c
@@ -297,9 +297,12 @@ static void on_incoming_call(pjsua_acc_id acc_id, pjsua_call_id call_id,
                              pjsip_rx_data *rdata)
 {
     pjsua_call_info call_info;
+    pjsua_msg_data msg_data;
 
     PJ_UNUSED_ARG(acc_id);
     PJ_UNUSED_ARG(rdata);
+
+    pjsua_msg_data_init(&msg_data);
 
     pjsua_call_get_info(call_id, &call_info);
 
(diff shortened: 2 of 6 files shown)


nanangizz:
Could you please also describe some sample scenario where this new options/features may be useful for? Also why the custom headers are only added for INVITE requests and responses (for example, for completeness, perhaps there should be options for other methods and/or any method too)?

joeygo:
Hi nanangizz,

Sure! Here are some examples of scenarios where custom SIP headers may be useful:

1. Call Priority and QoS (Quality of Service) Management:
   Custom headers can be used to indicate the priority of a call or specify specific QoS requirements. For example, a custom header like "X-Call-Priority" can be included in the SIP INVITE request to prioritize certain types of calls, such as emergency or high-priority business calls, over regular calls.

2. Service Differentiation:
   Custom headers can be used to differentiate between different types of services or service providers. For instance, a custom header like "X-Service-Type" can indicate the requested service type, such as voice, video, messaging, or data, allowing the SIP server to handle the request accordingly.

3. Authentication and Authorization:
   Custom headers can be utilized for authentication and authorization purposes. For example, a custom header like "X-Auth-Token" can contain a unique token or key associated with the user, which the SIP server can validate to ensure the user's authenticity and grant appropriate access permissions.

4. Billing and Accounting:
   Custom headers can be used to track and manage billing and accounting information related to SIP calls. A custom header like "X-Call-Cost" can include the cost of the call, while another header like "X-Account-ID" can specify the user's account ID or billing reference for accurate tracking and invoicing.

5. Call Recording and Monitoring:
   Custom headers can be employed to control call recording and monitoring functionalities. For instance, a custom header like "X-Record-Call" can be used to indicate whether the call should be recorded. In contrast, another header like "X-Monitor-Call" can enable call monitoring for quality assurance or compliance purposes.

These are just a few examples of how custom SIP headers can be utilized to enhance the functionality and customization of pjsua toward SIP-based communication systems. The specific custom headers and their usage can be tailored to meet the requirements and objectives of any application or service.

Regarding your second comment, I am not familiar with other clients providing the ability to extend other sip methods with custom headers, and honestly, I don't think it is needed. :-)

trengginas:
Could you please modify the code to limit the code width to 80 characters?

pjsip-apps/src/pjsua/pjsua_app_common.h
```diff
@@ -91,6 +91,14 @@ typedef struct pjsua_app_config
     unsigned                buddy_cnt;
     pjsua_buddy_config      buddy_cfg[PJSUA_MAX_BUDDIES];
 
+    unsigned                inv_hdr_cnt;
+    pj_str_t                inv_hname_cfg[16];
+    pj_str_t                inv_hvalue_cfg[16];
```
wosrediinanatour, line 96: Here an above: please don't use magic numbers. Where does 16 come from?

joeygo: I've added a define PJSUA_MAX_CUSTOM_HEADERS to pjsua.h

joeygo:
> Could you please modify the code to limit the code width to 80 characters?

Done.

nanangizz:
> Sure! Here are some examples of scenarios where custom SIP headers may be useful:
>
> ...
>
> These are just a few examples of how custom SIP headers can be utilized to enhance the functionality and customization of pjsua toward SIP-based communication systems. The specific custom headers and their usage can be tailored to meet the requirements and objectives of any application or service.

Right, I am aware of custom SIP headers as some PJSIP users use private SIP headers in their app (developed on PJSIP lib, not on pjsua app). But I should have been more specific, have you seen the need of adding custom/private headers specifically for **pjsua app** which is a command-line/telnet app, because IIRC I haven't seen any query/request of that for pjsua app, even for testing purposes.

> Regarding your second comment, I am not familiar with other clients providing the ability to extend other sip methods with custom headers, and honestly, I don't think it is needed. :-)

How about messaging or registration that may use authentication or may need to manage QoS? For example, some private headers added to registration [here](https://datatracker.ietf.org/doc/html/rfc7315#section-4.3.2.3).

Also note that the value/token set to the header may be dynamic (changing between requests).

joeygo:
> > Regarding your second comment, I am not familiar with other clients providing the ability to extend other sip methods with custom headers, and honestly, I don't think it is needed. :-)
> 
> How about messaging or registration that may use authentication or may need to manage QoS? For example, some private headers added to registration [here](https://datatracker.ietf.org/doc/html/rfc7315#section-4.3.2.3).
> 
> Also note that the value/token set to the header may be dynamic (changing between requests).

Maybe, they question now, is that blocking the review/merge, or the PR can be accepted and merged and if needed in the future we'll implement it?

nanangizz:
> Maybe, they question now, is that blocking the review/merge, or the PR can be accepted and merged and if needed in the future we'll implement it?

Here I have been trying to identify the pro (e.g: the usefulness) and the cons (e.g: increasing code size/complexity, future maintenance (possible bug/enhancement)) of this feature, before start reviewing the code/patch :-) Honestly initially adding this feature sounds great, but then considering the low demand and not really convinced the real world usage of custom INVITE headers in **pjsua app** (e.g: something like *to enable XYZ functionalities in ABC SIP provider/server*), perhaps this feature is not that urgent for now, but other reviewers may say otherwise :-)

sauwming:
I tend to agree with @nanangizz . The feature appears to have a specific focus, and it might be more suitable to include it in the user app rather than integrating it into the sample app provided by the library.

We appreciate the effort put into the patch, and I hope it can serve as a valuable reference for other users seeking to implement similar features. Thank you.

For now, I will mark this as closed, but we're open to reconsidering and reopening it in the future if there is an increased demand or requests for such feature.

PR #3625, opened 2023-07-10
Fix compile errors on old compilers

nanangizz:
This PR fixes:
- C90 compatibility issue: variable declaration in the middle of block (unfortunately in third-party libsrtp code).
- C++98 compatibility issue: Vector::data() is not available.
- MSVC2005 project misses some new libsrtp files.

### DIFF
--- a/pjsip/src/pjsua2/media.cpp
+++ b/pjsip/src/pjsua2/media.cpp
@@ -288,7 +288,16 @@ static pj_status_t get_frame(pjmedia_port *port, pjmedia_frame *frame)
     mport->onFrameRequested(frame_);
     frame->type = frame_.type;
     frame->size = PJ_MIN(frame_.buf.size(), frame_.size);
+
+#if ((defined(_MSVC_LANG) && _MSVC_LANG <= 199711L) || __cplusplus <= 199711L)
+    /* C++98 does not have Vector::data() */
+    if (frame->size > 0)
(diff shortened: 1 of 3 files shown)


pjsip/src/pjsua2/media.cpp
```diff
@@ -288,7 +288,16 @@ static pj_status_t get_frame(pjmedia_port *port, pjmedia_frame *frame)
     mport->onFrameRequested(frame_);
     frame->type = frame_.type;
     frame->size = PJ_MIN(frame_.buf.size(), frame_.size);
+
+#if ((defined(_MSVC_LANG) && _MSVC_LANG <= 199711L) || __cplusplus <= 199711L)
```
sauwming, line 292: Do we really need to support a compiler that's already a quarter century old?

Besides, C++98 doesn't guarantee vectors to be contagious (it was only added in C++03), so even though it can compile, theoretically it can crash?

nanangizz: Well, officially we still support MSVC2005, don't we :)

They say the C++98 standard has been updated (C++98-TC?), also implementations very likely implement it using contiguous buffer, except for `bool`, from [here](https://stackoverflow.com/questions/849168/are-stdvector-elements-guaranteed-to-be-contiguous).

sauwming (approved):

trengginas (approved):

PR #3626, opened 2023-07-10
Add SSL socket info to TLS verification callback

nanangizz:
The TLS verification callback, `pjsip_tls_setting.on_verify_cb`, can be used by app to implement its own TLS verification. Currently the callback pass some TLS info via callback param `pjsip_tls_on_verify_param` and there is a request for an additional info: pre-verification status of peer certificate. This PR add SSL socket info (which also contains pre-verification status) to the callback param, so application can make use all available info for its verification logic.

This PR also modify the behavior of verification callback in SSL socket and TLS transport, i.e: the callback will not be called if verification is disabled (i.e: `pj_ssl_sock_param.verify_peer==PJ_FALSE`, `pjsip_tls_setting.verify_client/server==PJ_FALSE`).

Thanks Peter Koletzki for the request/feedback.

### DIFF
--- a/pjlib/src/pj/ssl_sock_ossl.c
+++ b/pjlib/src/pj/ssl_sock_ossl.c
@@ -946,15 +946,6 @@ static int verify_cb(int preverify_ok, X509_STORE_CTX *x509_ctx)
         goto on_return;
     }
 
-    if (ssock->param.cb.on_verify_cb) {
-        update_certs_info(ssock, x509_ctx, &ssock->local_cert_info, 
-                          &ssock->remote_cert_info, PJ_TRUE);
-        preverify_ok = (*ssock->param.cb.on_verify_cb)(ssock, 
-                                                       ssock->is_server);
-
-        goto on_return;
-    }
-
     /* Store verification status */
     err = X509_STORE_CTX_get_error(x509_ctx);
     switch (err) {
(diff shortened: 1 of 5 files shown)


pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1025,11 +1016,18 @@ static int verify_cb(int preverify_ok, X509_STORE_CTX *x509_ctx)
...
-    /* When verification is not requested just return ok here, however
-     * application can still get the verification status.
+    /* When verification is not requested just return ok here.
+     * Otherwise, invoke app's verification callback, if configured.
+     * Anyway, application can still get the verification status.
      */
-    if (PJ_FALSE == ssock->param.verify_peer)
+    if (PJ_FALSE == ssock->param.verify_peer) {
         preverify_ok = 1;
+    } else if (ssock->param.cb.on_verify_cb) {
```
sauwming, line 1025: Not sure if this is intentional, but there seems to be a behavior change with the patch.
Previously, the callback will be called regardless of the value of `verify_peer`, but now only if `verify_peer` is `PJ_TRUE`.

nanangizz: Forgot to mention this in the desc, yes it is intentional. The callback invocation needs to be moved down to get preverification status. Then I thought the callback should also regard the `verify_peer` setting as otherwise there may be contradiction, i.e: TLS connection may be dropped due to verification while `verify_peer` is disabled (when callback implementation in app does not check if verification is disabled).

Alternatively, we can add some docs in `pj_ssl_sock_param.verify_peer` and `pjsip_tls_setting.verify_client/server` to explain the behavior, i.e: if verification callback is set/implemented, those flags will be ignored.

sauwming (approved):
Okay, as long as clarified in the doc.

trengginas (approved):

nanangizz:
Replaced by #3630.

PR #3636, opened 2023-07-17
Split dropped packet transport logging

### LINKED ISSUE #3476: Invalid inbound binary data should not be logged with %s, and probably not as ERROR
### Describe the bug

It appears my Asterisk is being probed for security vulnerabilities, and I am getting messages in the Asterisk log file like this one:

image

(1) The binary data dumped into the log file "as is" with %s is a bad practice for obvious reasons, It should be formatted as hex with printable ASCII and limited in length (e.g., to 64 or 128 bytes). The [current code](https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_transport.c#L2093) is below:

`            if (tmp.slen) {
                PJ_LOG(1, (THIS_FILE, 
                      "Error processing %ld bytes packet from %s %s:%d %.*s:\n"
                      "%.*s\n"
                      "-- end of packet.",
                      msg_fragment_size,
                      rdata->tp_info.transport->type_name,
                      rdata->pkt_info.src_name, 
                      rdata->pkt_info.src_port,
                      (int)tmp.slen, tmp.ptr,
                      (int)msg_fragment_size,
                      rdata->msg_info.msg_buf));
            }` 

(2) It also shouldn't be logged as ERROR because at this level it's hard to get rid of these messages which in my case don't present an error at all but a nuisance. It could be logged as ERROR if it happened in a context of a valid connection. If it's just a random packet, it should be logged as a NOTICE or, better, DEBUG.

Thanks!

### Steps to reproduce

N/A

### PJSIP version

2.12.1, but the code is still present in master.

### Context

N/A

### Log, call stack, etc

```shell
N/A
```

mbradeen:
To fix #3476.

Split logging to keep Warning regarding dropped packet at level 1 while moving the packet contents to 4.

The existing fix moves the entire log entry to 4. The fact that there was a dropped packet should still be logged at level 1.

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -2103,15 +2103,17 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
              * which were sent to keep NAT bindings.
              */
             if (tmp.slen) {
-                PJ_LOG(4, (THIS_FILE, 
-                      "Dropping %ld bytes packet from %s %s:%d %.*s:\n"
-                      "%.*s\n"
-                      "-- end of packet.",
+                PJ_LOG(2, (THIS_FILE,
+                      "Dropping %d bytes packet from %s %s:%d %.*s\n",
                       msg_fragment_size,
                       rdata->tp_info.transport->type_name,
-                      rdata->pkt_info.src_name, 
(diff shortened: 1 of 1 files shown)


sauwming (approved):

nanangizz (approved):

sauwming:
For future reference, the log level is increased because the packet source information is deemed important to identify any rogue senders.

PR #3641, opened 2023-07-20
Fixed pjmedia wsola test failure

### LINKED ISSUE #3638: The pjmedia-test item is abnormal.
### Describe the bug

The pjmedia-test item is abnormal in versions later than 2.13.
Through comparison and locating, it is found that the problem is caused by the different parameters of the pjmedia_wsola_create(pool, clock_rate, samples_per_frame, channel_count, opt, &wp->wsola); opt function in the pjproject/pjmedia/src/test/mips_test.c file.

### Steps to reproduce

After the compilation is complete, run the make pjmedia-test command.

### PJSIP version

2.13

### Context

There is no log for the infinite loop.

### Log, call stack, etc

```shell
There is no log for the infinite loop.
```

sauwming:
To fix #3638 .

There are a couple of issues identified here:
- Bug in the unit test itself, which will still call `pjmedia_wsola_generate()` even though `loss_pct` is 0.
- Infinite loop in `wsola` if `pjmedia_wsola_generate()` is called when `wsola` is created with option `PJMEDIA_WSOLA_NO_PLC`. This is because `expand_sr_min_dist` and `expand_sr_max_dist` are only initialised if the option is not specified:
```
    if ((options & PJMEDIA_WSOLA_NO_PLC) == 0) {
        wsola->min_extra = wsola->hanning_size;
        wsola->expand_sr_min_dist = (pj_uint16_t)
                                    (EXP_MIN_DIST * wsola->samples_per_frame);
        wsola->expand_sr_max_dist = (pj_uint16_t)
                                    (EXP_MAX_DIST * wsola->samples_per_frame);
    }
```
so `find_pitch(pj_int16_t *frm, pj_int16_t *beg, pj_int16_t *end, ...)` will be called with `beginning` equalling `end`, and get stuck in infinite loop: `find_pitch(templ, templ - wsola->expand_sr_max_dist, templ - wsola->expand_sr_min_dist, ...)`. 

This PR will fix both.

### DIFF
--- a/pjmedia/src/pjmedia/wsola.c
+++ b/pjmedia/src/pjmedia/wsola.c
@@ -632,11 +632,15 @@ static void expand(pjmedia_wsola *wsola, unsigned needed)
         templ = reg1 + reg1_len - wsola->hanning_size;
         CHECK_(templ - reg1 >= wsola->hist_size);
 
-        start = find_pitch(templ, 
-                           templ - wsola->expand_sr_max_dist, 
-                           templ - wsola->expand_sr_min_dist,
-                           wsola->templ_size, 
-                           1);
+        if ((wsola->options & PJMEDIA_WSOLA_NO_PLC) == 0) {
+            start = find_pitch(templ,
+                               templ - wsola->expand_sr_max_dist,
+                               templ - wsola->expand_sr_min_dist,
+                               wsola->templ_size,
+                               1);
+        } else {
+            start = PJ_MAX(templ - needed + generated, reg1);
+        }
 
         /* Should we make sure that "start" is really aligned to
          * channel #0, in case of stereo? Probably not necessary, as
--- a/pjmedia/src/test/mips_test.c
+++ b/pjmedia/src/test/mips_test.c
@@ -1022,7 +1022,7 @@ static pj_status_t wsola_plc_get_frame(struct pjmedia_port *this_port,
     pj_status_t status;
 
 
-    if ((pj_rand() % 100) > wp->loss_pct) {
+    if ((pj_rand() % 100) >= wp->loss_pct) {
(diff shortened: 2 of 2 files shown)


nanangizz:
Re: second issue. First, the docs of `PJMEDIA_WSOLA_NO_PLC` says
> WSOLA will not be used for PLC

IMHO it is rather ambiguous as the generate/PLC functionality seems to still be provided by WSOLA. After checking the code, perhaps the PLC functionality is supposed to omit the 'waveform similarity search' part, but still use the 'overlap-and-add'. If this assumption is correct, the fix should be: in `expand()`, skip `find_pitch()` (the similarity search algo) altogether, just assign the `start` based on the `needed` (i.e: `start = templ - needed`). But it seems that `start` cannot be earlier than history region (please check), if so, perhaps just readjust the `start` (the frame generation will take more than one iterations).

nanangizz (approved):

sauwming:
Changing it to:
```
        if ((wsola->options & PJMEDIA_WSOLA_NO_PLC) == 0) {
            start = find_pitch(templ, 
                               templ - wsola->expand_sr_max_dist, 
                               templ - wsola->expand_sr_min_dist,
                               wsola->templ_size, 
                               1);
        } else {
            start = PJ_MAX(templ - needed, reg1);
        }
```
seems to work (no infinite loop).
Is this okay and should I include it in this PR?

nanangizz:
Looks okay to me, an update in case there are two or more iterations:
```start = PJ_MAX(templ - needed + generated, reg1);```

Yes, I guess it is okay to also include it here.

nanangizz (approved):

trengginas (approved):

PR #3643, opened 2023-07-21
Fixed potentially duplicate RTCP FB attribute on response

### LINKED ISSUE #3642: Duplicate NACK attribute on answer
### Describe the bug

Since I setup `acc->cfg.rtcp_fb_cfg.cap_count` with 1 caps NACK and receive incoming INVITE with NACK see an two attributes.

### Steps to reproduce

1. Make PJSUA register and ready to answer
2. Make call
3. See an SDP on 200 OK response

### PJSIP version

2.13.1

### Context

 - Simple mock SIP server

### Log, call stack, etc

```shell
- Think that issue raise up since pjsua_media.c:2939 and pjsua_media.c:2954 where we invoke `pjmedia_rtcp_fb_encode_sdp` and each one invoke `add_sdp_attr_rtcp_fb` and setup attribute.
```

sauwming:
To fix #3642.

For answer, we will list all the capabilities in RTCP FB settings twice, which result in duplication.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -2961,8 +2961,10 @@ pj_status_t pjsua_media_channel_create_sdp(pjsua_call_id call_id,
                              call_id, mi));
             }
 
-            /* Add any other RTCP-FB setting configured in account setting */
-            if (acc->cfg.rtcp_fb_cfg.cap_count) {
+            /* For offer, add any other RTCP-FB setting configured in account
+             * setting.
+             */
+            if (!rem_sdp && acc->cfg.rtcp_fb_cfg.cap_count) {
                 pj_bool_t tmp = rtcp_cfg.dont_use_avpf;
                 rtcp_cfg = acc->cfg.rtcp_fb_cfg;
                 rtcp_cfg.dont_use_avpf = tmp;


nanangizz:
It looks like we haven't implemented SDP nego for RTCP-FB attributes (negotiation only implemented for the transport, `RTP/AVP` vs `RTP/AVPF`), so we don't have match-the-offer mechanism yet?

https://www.rfc-editor.org/rfc/rfc4585.html#section-4.2:
```
   When used in conjunction with the offer/answer model [8], the offerer
   MAY present a set of these AVPF attributes to its peer.  The answerer
   MUST remove all attributes it does not understand as well as those it
   does not support in general or does not wish to use in this
   particular media session.  The answerer MUST NOT add feedback
   parameters to the media description and MUST NOT alter values of such
   parameters.
```

sauwming:
Ah you are right, I thought we already have the offer matching, and the duplication is because of offer match + list capabilities.
It turns out it's because for answer, we will list the capabilities twice, once in `pjsua_media_channel_create_sdp()` and then again in `pjsua_media_channel_update()`.

So for the duplication issue, I believe we should remove either one of them.
As for the offer matching, I suppose it will be a TODO.

nanangizz (approved):

trengginas (approved):

PR #3644, opened 2023-07-25
Fix bug that causes typeError when encoding sipURI in python

zooda56:
Instead of adding port as a string this if block tried to call the port as a function.


nanangizz:
Thanks for the PR. But I am afraid that this Python module has been deprecated and no longer maintained, please check [here](https://trac.pjsip.org/repos/wiki/Python_SIP_Tutorial) for more info.

PR #3653, opened 2023-08-09
Fixed RTCP packet lost calculation

sauwming:
Suppose RTP packets arrive in the following order: seq 5, 10, 6, 7, 8, 9

Currently, upon receiving seq 10, we will calculate it as 4 lost packets (6 until 9) in `pjmedia_rtcp_rx_rtp2()`: `sess->stat.rx.loss += (seq_st.diff - 1)`.

According to the RFC https://datatracker.ietf.org/doc/html/rfc3550#appendix-A.3, we should calculate it as `lost = expected - s->received` instead. But we commented this code in https://github.com/pjsip/pjproject/commit/84f45890c4afb189bc06f835902ed41cce6e6ab5 because we have done the calculation earlier in `pjmedia_rtcp_rx_rtp2()`. So we probably should use the later calculation instead.

### DIFF
--- a/pjmedia/src/pjmedia/rtcp.c
+++ b/pjmedia/src/pjmedia/rtcp.c
@@ -372,7 +372,11 @@ PJ_DEF(void) pjmedia_rtcp_rx_rtp2(pjmedia_rtcp_session *sess,
     /* Only mark "good" packets */
     ++sess->received;
 
-    /* Calculate loss periods. */
+    /* Calculate packet lost and loss periods.
+     * We should not calculate packet lost here and do it when we're sending
+     * outbound RTCP RR.
+     */
+    /*
     if (seq_st.diff > 1) {
         unsigned count = seq_st.diff - 1;
         unsigned period;
@@ -380,16 +384,16 @@ PJ_DEF(void) pjmedia_rtcp_rx_rtp2(pjmedia_rtcp_session *sess,
         period = count * sess->dec_pkt_size * 1000 / sess->clock_rate;
         period *= 1000;
 
-        /* Update packet lost. 
-         * The packet lost number will also be updated when we're sending
-         * outbound RTCP RR.
-         */
+        // Update packet lost. 
+        // The packet lost number will also be updated when we're sending
+        // outbound RTCP RR.
         sess->stat.rx.loss += (seq_st.diff - 1);
         TRACE_((sess->name, "%d packet(s) lost", seq_st.diff - 1));
 
-        /* Update loss period stat */
+        // Update loss period stat
         pj_math_stat_update(&sess->stat.rx.loss_period, period);
     }
+    */
 
 
     /*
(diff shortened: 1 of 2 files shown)


sauwming:
The first commits introduce a behavioural change, i.e. in that the packet lost will be calculated even if the probation is not over. For example, RTP packets received in order: seq 1, 12, 13, 14, ... will previously be considered 0 loss, but now will be 10 losses (2 until 11).
This is because previously `pjmedia_rtp_seq_update()` will not consider the seq diff when we're still in probation.

According to the RFC https://datatracker.ietf.org/doc/html/rfc3550#appendix-A.1:
```
               if (s->probation == 0) {
                   init_seq(s, seq);
                   s->received++;
...
```
we only init the sequence and start counting the packets received once the probation has ended, and the seq is inited to the last seq received during probation, not the first seq.

sauwming:
The third commit is to fix #3655, to handle seq rollover during probation.

pjmedia/src/pjmedia/rtcp.c
```diff
@@ -991,12 +996,22 @@ PJ_DEF(void) pjmedia_rtcp_build_rtcp(pjmedia_rtcp_session *sess,
...
+
+        sess->stat.rx.loss = count;
+        TRACE_((sess->name, "%d packet(s) lost", count));
+
+        /* Update loss period stat */
+        pj_math_stat_update(&sess->stat.rx.loss_period, period);
+    } else {
         sess->stat.rx.loss = 0;
-    */
+    }
```
nanangizz, line 1014: What if this calculation is done on each incoming RTP instead of in building RTCP for faster stats update (by moving this block there perhaps), because the RTCP interval is configurable so it can be long, default is not that long though, 5 seconds.

Also, perhaps only update/calculate it when out-of-order flag is not set.

sauwming: Okay, done in the latest commit.

There are a couple of drawbacks though:
- Calculating only when not out of order means that the packet loss stat may not always be accurate.
- There is also complication with updating the loss period. The stats such as avg, max, std, etc will be inaccurate.

nanangizz: Re: first drawback should be acceptable, as I think most RTP packets are ordered (e.g: every second there should be at least two properly ordered packets).

Re: second drawback. Currently the `count` is the total loss count since the beginning of the session? If it is, the period should not be counted using `count`. IMO the loss period represents the consecutive loss packets (in time unit) of every packet loss occurance. So the calculation is perhaps: when total loss count is increased, compare the current loss  count vs previous loss count, the difference should be the period?

pjmedia/src/pjmedia/rtcp.c
```diff
@@ -384,17 +384,20 @@ PJ_DEF(void) pjmedia_rtcp_rx_rtp2(pjmedia_rtcp_session *sess,
         expected = last_seq - sess->seq_ctrl.base_seq;
 
         if (expected > sess->received)
-            count = expected - sess->received;
+            loss = expected - sess->received;
 
+        if (loss > sess->stat.rx.loss)
+            count = loss - sess->stat.rx.loss;
         period = count * sess->dec_pkt_size * 1000 / sess->clock_rate;
         period *= 1000;
```
nanangizz, line 392: minor: move this period calculation to inside the `loss_period` update block below?

nanangizz (approved):

Issue #2386, opened 2020-04-13
Meson build system


vit1251:
Could you please start distribute PJSIP modules as meson subproject?

sauwming:
From https://mesonbuild.com:
```
Meson is an open source build system meant to be both extremely fast, and, even more importantly, as user friendly as possible.

Features
- multiplatform support for Linux, macOS, Windows, GCC, Clang, Visual Studio and others 
```

Sounds interesting, unfortunately we have no plan to support this currently. If there's more interest, then perhaps we'll reconsider in the future.

vit1251:
Meson is most important replace autotools and provide reuse via subprojects. I think that it way to simplyfy distribution channel. I already use meson with PJ and may provide basic example if usage in fork. But I use only on Linux and make meson on top of make output.

vit1251:
Let's I make fork for example current release version 2.10 and we return back to that conversation about using meson?

sauwming:
Yes, please share with us your Linux meson build. It should serve as a good starting point for us or others to expand it to other platforms later in the future.

vit1251:
Greetings. I made fork and start branch ```2.10+meson``` on https://github.com/vit1251/pjproject/commit/2119deaf7007ddebf010f36b3cf26b444826114e 

I tried to build with classic ```./configure```, ```make dep```, ```make```. As a result of the work of these scenarios, I received script output with ```gcc``` calls with parameters. I began to build my meson assembly over that script output. I understand that this is a simplified solution, but I don’t have many resources right now to deeply explore the original build system. If you like this direction, it will make sense to continue improve the assembly on the meson.

The assembly began with the link of the static library ```pjlib```. In the beginning I replace macro definition PJ_AUTOCONF on PJ_LINUX. In the future, we can create configuring meson code with environment research and create the necessary files and add variables for compilation. See more on https://mesonbuild.com/Configuration.html

Taking the first steps I immediately got a compilation error:

```
[14/47] Compiling C object 'pjlib@sta/pjlib_src_pj_sock_bsd.c.o'.
FAILED: pjlib@sta/pjlib_src_pj_sock_bsd.c.o 
[...]

PR #3659, opened 2023-08-10
Add Meson build system welcomes

vit1251:
Let's try second attempt with meson build integration started at discussion https://github.com/pjsip/pjproject/issues/2386 a few years ago.

I simple copy output autotools and invite people to start contribute and add more architectures and compile options.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -52,6 +52,10 @@ add_global_arguments('-Werror=format-security', language : 'c')
 add_global_arguments('-ggdb3', language : 'c')
 
 add_global_arguments('-DPJ_DEBUG=1', language : ['c', 'cpp'])
+
+add_global_arguments('-DPJ_IS_BIG_ENDIAN=0', language : ['c', 'cpp', 'objc'])
+add_global_arguments('-DPJ_IS_LITTLE_ENDIAN=1', language : ['c', 'cpp', 'objc'])
+
 add_global_arguments('-DPJ_SOCK_HAS_GETADDRINFO=1', language : ['c', 'cpp'])
 add_global_arguments('-DPJ_HAS_IPV6=0', language : ['c', 'cpp'])
 add_global_arguments('-DPJ_SOCK_HAS_INET_PTON=0', language : ['c', 'cpp'])
@@ -73,6 +77,8 @@ add_global_arguments('-DPJMEDIA_HAS_ILBC_CODEC=0', language : ['c', 'cpp'])
 add_global_arguments('-DPJMEDIA_HAS_GSM_CODEC=0', language : ['c', 'cpp'])
 add_global_arguments('-DPJMEDIA_HAS_L16_CODEC=0', language : ['c', 'cpp'])
 add_global_arguments('-DPJMEDIA_HAS_G722_CODEC=0', language : ['c', 'cpp'])
+
+system_dep = []
 
 openssl_dep = dependency('openssl')
 
@@ -132,7 +138,8 @@ resample_dep = declare_dependency(link_with : [
 # PJLIB
 
 pjlib_inc = include_directories('pjlib/include')
-pjlib_sources = [
+if os == 'linux'
+    pjlib_sources = [
 	'pjlib/src/pj/ioqueue_select.c',
 	'pjlib/src/pj/file_access_unistd.c',
 	'pjlib/src/pj/file_io_ansi.c',
@@ -178,20 +185,68 @@ pjlib_sources = [
 	'pjlib/src/pj/string.c',
 	'pjlib/src/pj/timer.c',
 	'pjlib/src/pj/types.c',
-]
-if os == 'darwin'
-    pjlib_sources += [
-        'pjlib/src/pj/sock_qos_darwin.c',
-        'pjlib/src/pj/os_core_darwin.m',
+	'pjlib/src/pj/os_core_unix.c',
     ]
-else
-    pjlib_sources += [
-        'pjlib/src/pj/os_core_unix.c',
+elif os == 'darwin'
+    pjlib_sources = [
+	'pjlib/src/pj/ioqueue_select.c',
+	'pjlib/src/pj/file_access_unistd.c',
+	'pjlib/src/pj/file_io_ansi.c',
+	'pjlib/src/pj/os_core_unix.c',
+	'pjlib/src/pj/os_error_unix.c',
+	'pjlib/src/pj/os_time_unix.c',
+	'pjlib/src/pj/os_timestamp_posix.c',
+	'pjlib/src/pj/os_core_darwin.m',
+	'pjlib/src/pj/sock_qos_darwin.c',
+	'pjlib/src/pj/sock_qos_bsd.c',
+	'pjlib/src/pj/ssl_sock_apple.m',
+	'pjlib/src/pj/guid_darwin.c',
+	'pjlib/src/pj/addr_resolv_sock.c',
+	'pjlib/src/pj/log_writer_stdout.c',
+	'pjlib/src/pj/os_timestamp_common.c',
+	'pjlib/src/pj/pool_policy_malloc.c',
+	'pjlib/src/pj/sock_bsd.c',
+	'pjlib/src/pj/sock_select.c',
+	'pjlib/src/pj/activesock.c',
+	'pjlib/src/pj/array.c',
+	'pjlib/src/pj/config.c',
+	'pjlib/src/pj/ctype.c',
+	'pjlib/src/pj/errno.c',
+	'pjlib/src/pj/except.c',
+	'pjlib/src/pj/fifobuf.c',
+	'pjlib/src/pj/guid.c',
+	'pjlib/src/pj/hash.c',
+	'pjlib/src/pj/ip_helper_generic.c',
+	'pjlib/src/pj/list.c',
+	'pjlib/src/pj/lock.c',
+	'pjlib/src/pj/log.c',
+	'pjlib/src/pj/os_time_common.c',
+	'pjlib/src/pj/os_info.c',
+	'pjlib/src/pj/pool.c',
+	'pjlib/src/pj/pool_buf.c',
+	'pjlib/src/pj/pool_caching.c',
+	'pjlib/src/pj/pool_dbg.c',
+	'pjlib/src/pj/rand.c',
+	'pjlib/src/pj/rbtree.c',
+	'pjlib/src/pj/sock_common.c',
+	'pjlib/src/pj/sock_qos_common.c',
+	'pjlib/src/pj/ssl_sock_common.c',
+	'pjlib/src/pj/ssl_sock_ossl.c',
+	'pjlib/src/pj/ssl_sock_gtls.c',
+	'pjlib/src/pj/ssl_sock_dump.c',
+	'pjlib/src/pj/ssl_sock_darwin.c',
+	'pjlib/src/pj/string.c',
+	'pjlib/src/pj/timer.c',
+	'pjlib/src/pj/types.c',
     ]
+    system_dep += dependency('CoreFoundation')
+    system_dep += dependency('AppKit')
+    
 endif
 pjlib = library('pjlib', pjlib_sources, include_directories: [
 	pjlib_inc,
 ], dependencies: [
+	system_dep,
 	openssl_dep,
 	thread_dep,
 	uuid_dep,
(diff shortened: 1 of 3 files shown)


sauwming:
Thank you for your contribution. Perhaps you can update the PR to include what platforms (and architectures) this has been tested to work on, and what platforms and archs are still Todo.

I tried this on Mac and got the following results:
- srtp2 should not be required, by default we should use the bundled SRTP
- Error:
```
meson.build:17:0: ERROR: Expecting rparen got id.
conf_data = configuration_data({
^
For a block that started at 1,7
project('pjproject', ['c', 'cpp'], version: '2.13.1'
```

vit1251:
Let's try to build PJSIP for a macOS computer. It's a pretty wild platform for me but ok.

Meson has the ability to use external dependencies so we can use srtp from the library
developer at https://github.com/cisco/libsrtp/blob/main/meson.build

Of course, it is more convenient to use an external dependency, since we can save
resources on instrumenting the build.

Regarding your error, I don’t really understand the reason for its occurrence,
but tell me which version of the meson (meson -v) do you use? I am using
version 1.0.1 under linux platform.

sauwming:
We already have libsrtp included in `third_party`, but sure, we can download and use the latest external SRTP if it's convenient.

I already use `1.2.1`:
```
The Meson build system
Version: 1.2.1
Source dir: /Users/pjsip/pjproject-mac
Build dir: /Users/pjsip/pjproject-mac/builddir
Build type: native build

meson.build:17:0: ERROR: Expecting rparen got id.
conf_data = configuration_data({
^
For a block that started at 1,7
project('pjproject', ['c', 'cpp'], version: '2.13.1'
       ^
```

vit1251:
Hmmm... I write `meson-sh` shell script. Try to check on your MacOS system. My output is

```
The Meson build system
Version: 1.2.1
Source dir: /Users/.../pjproject
Build dir: /Users/.../pjproject/builddir
Build type: native build
Project name: pjproject
Project version: 2.13.1
C compiler for the host machine: cc (clang 14.0.0 "Apple clang version 14.0.0 (clang-1400.0.29.202)")
C linker for the host machine: cc ld64 820.1
C++ compiler for the host machine: c++ (clang 14.0.0 "Apple clang version 14.0.0 (clang-1400.0.29.202)")
C++ linker for the host machine: c++ ld64 820.1
Host machine cpu family: x86_64
Host machine cpu: x86_64
Configuring config_site.h using configuration
Found pkg-config: /usr/local/bin/pkg-config (0.29.2)
Run-time dependency openssl found: YES 3.1.2
Run-time dependency threads found: YES
Run-time dependency uuid found: YES 1.0
Library m found: YES
Found CMake: /usr/local/bin/cmake (3.27.2)
...
```

sauwming:
I tested it here and it seems to work reasonably well now.

The only missing parts now seem to be the bundled support for srtp and Speex, which should be the default options, since this is also the default if we use the configure script.

meson.build
```diff
@@ -0,0 +1,1207 @@
...
+    os_auto.set('PJ_HAS_ERRNO_H', 1)
+endif
+os_auto.set('PJ_HAS_FCNTL_H', 1)
+if compiler.has_header('limits.h')
+  os_auto.set('PJ_HAS_LIMITS_H', 1)
+endif
+#undef PJ_HAS_LINUX_SOCKET_H */
+if compiler.has_header('malloc.h')
+  os_auto.set('PJ_HAS_MALLOC_H', 1)
+endif
```
eli-schwartz, line 40: meson's own tools/ac_converter.py will do this:
```meson
check_headers = [
    'arpa/inet.h',
    'assert.h',
    'errno.h',
    'limits.h',
    'malloc.h',
]

foreach h : check_headers
    if cc.has_header(h)
        cdata.set('HAVE_' + h.underscorify().to_upper(), 1)
    endif
endforeach
```

Minor style tweaks aside, this can be very helpful in the general case by reducing the complexity of adding an additional header check.

meson.build
```diff
@@ -0,0 +1,1207 @@
...
+os_auto.set('PJ_THREAD_SET_STACK_SIZE', 0)
+os_auto.set('PJ_THREAD_ALLOCATE_STACK', 0)
+os_auto.set('PJ_HAS_SSL_SOCK', 1)
+os_auto.set('PJ_SSL_SOCK_IMP', 'PJ_SSL_SOCK_IMP_OPENSSL')
+#undef PJ_HAS_PTHREAD_NP_H */
+#define PJ_HAS_PTHREAD_SETNAME_NP 1
+#undef PJ_HAS_PTHREAD_SET_NAME_NP */
+#
+configure_file(output : 'os_auto.h', configuration : os_auto)
+run_command('/bin/cp', 'builddir/os_auto.h', 'pjlib/include/pj/compat/os_auto.h', check : true)
```
eli-schwartz, line 148: Why?

meson.build
```diff
@@ -0,0 +1,1207 @@
...
+speex = get_option('speex')
+if speex == 'system'
+  speex_dep = dependency('speex', required : true)
+  add_global_arguments('-DPJMEDIA_HAS_SPEEX_AEC=0', language : ['c', 'cpp'])
+  add_global_arguments('-DPJMEDIA_HAS_SPEEX_CODEC=0', language : ['c', 'cpp'])
+elif speex == 'external'
+  # TODO - use https://gitlab.xiph.org/xiph/speex/-/blob/master/meson.build
+elif speex == 'bundled'
+  # TODO - last chance use bundled version...
+endif
```
eli-schwartz, line 240: These options seem entirely too complex, and meson can handle this automatically via `-Dwrap_mode=forcefallback` etc to force using the bundled copy instead of the system one. I don't see the need to reinvent the significant effort that the meson developers have put into making this polished.

Other build systems don't have an integrated way of handling this well, so third_party directories with hardcoded vendor copies are used as a hack.

meson.build
```diff
@@ -0,0 +1,1207 @@
...
+], include_directories : [
+	pjlib_inc,
+	pjlib_util_inc,
+	pjmedia_inc,
+	pjnath_inc,
+	pjsip_inc,
+])
+
+# PJSUA-APPS
+
```
eli-schwartz, line 1066: Might be worth making these into a for loop?

```meson
apps = [
    'auddemo', 'aviplay', 'aectest', 'clidemo',
    'confsample', # .....
]
foreach app: apps
    executable(app, 'pjsip-apps/src/samples/@0@.c'.format(app), dependencies: pjproject_dep)
endforeach

File: meson.build (hunk 1 above)
Comment (vit1251) [reply meson.build:40]: I completely agree. Let's use this approach. The original PR was only in the dumb output to meson transaltion. Of course you offer great improvements.

File: meson.build (hunk 2 above)
Comment (vit1251) [reply meson.build:148]: I did not understand how meson can publish the file in the source code. Recreating the directory tree in the `builddir` seemed like a very time consuming task for me. So I chose a simple copy here. What do you suggest?

File: meson.build (hunk 3 above)
Comment (vit1251) [reply meson.build:240]: I don't quite understand what exactly you are suggesting. Now there is some directory structure and source code in PJPROEJCT. Which other developers use since 2005, so I still don't understand how it is possible to support `third_party`, external source and using the system library without significant tree changes. Please tell me your idea in more detail.

File: meson.build (hunk 4 above)
Comment (vit1251) [reply meson.build:1066]: Yep, it may. All of these suggested ideas look like optimizations and improvements.

I generally welcome advanced use of meson tools, but first I would like to understand how generally able idea of using `meson` in `PJPROJECT` as a alternative build system.

If the welcome `meson` integration step is successful pass, we can continue to improve and extend `meson.build` in accordance with best practices and provie more architecture and platofrms.

At this stage, this is the first step (actually second) I would like to show support Linux and macOS (actually by direct PJPROJECT request and does not have any plan about support macOS platform) clarify for myself in general the possibility of using the `pjproject` as an external project in my source code.

Since I does not have feedback about this step I also does not want to spend more steps. Let I remind that the issue of integration with meson was raised about two years ago and while we are still on the beginning of road.

File: meson.build (hunk 2 above)
Comment (eli-schwartz) [reply meson.build:148]: I'm not certain I understand what you mean by "publish", as it will always be recreated from meson when you setup a build directory.

- Why does this file need to be in the source directory? You can include it just fine from the build directory.
- How do you know the build directory is called "builddir"? What if people use `meson setup builddir-debug/` and `meson setup builddir-release/` and possibly also use `builddir-clang/` or `builddir-gcc/`? This is why meson has `meson.current_build_dir()`.

I would probably use `configure_file()` from inside `pjlib/include/pj/compat/meson.build`. A C/C++ include will find that file from anywhere on the compiler search path, it doesn't have to be in the same directory tree as other files in `#include <pj/compat/*.h>`.

File: meson.build (hunk 3 above)
Comment (eli-schwartz) [reply meson.build:240]: I guess the issue is that it's not entirely clear what the goal of each option is.

- The system option has an obvious purpose -- many people like building against system shared libraries for consistency and applying security fixes once globally, e.g. buildroot would want to do this
- The external option would use a meson wrap, I guess, which is useful if you don't have a system copy available or you want to build statically linked executables for deployment or just want to test against whichever versions were tested as wraps
- The bundled option would *not* use a meson wrap, but it would... use the third_party/ directory and you'd write your own meson.build files for each one?

What is the difference between options 2 and 3? When would one prefer 2 over 3, and when would one prefer 3 over 2?

File: meson.build (hunk 2 above)
Comment (vit1251) [reply meson.build:148]: > Why does this file need to be in the source directory?

See article https://trac.pjsip.org/repos/wiki/Getting-Started/Build-Preparation

> How do you know the build directory is called "builddir"? What if people use ...

My version `meson-sh` use that directory. Quick answer is people make refer changes ...

File: meson.build (hunk 3 above)
Comment (vit1251) [reply meson.build:240]: > What is the difference between options 2 and 3? When would one prefer 2 over 3, and when would one prefer 3 over 2?

Now the architecture of the project is exactly like this and I think that there are some good reasons for this. In particular, I think that the goal is to ensure the autonomy of the assembly. In any case, I don't think it's in my power to make any structure changes.

I think you may wait for the first integration of meson.build and open discussions and make suggestions. Now I make reference version, with demonstration of options for solving the problem with dependencies.

**Comment by sauwming** [comment]
Having never used Meson myself, personally I don't have much issue with the build script itself, as long as it can successfully build.

But as mentioned above, the bundled third party libraries need to be supported before the PR can be considered for review.

**Comment by vit1251** [comment]
> Having never used Meson myself, personally I don't have much issue with the build script itself, as long as it can successfully build.
> 
> But as mentioned above, the bundled third party libraries need to be supported before the PR can be considered for review.

I see some library already use meson:

- https://gitlab.xiph.org/xiph/speex/-/blob/master/meson.build
- https://github.com/cisco/libsrtp/blob/main/meson.build

In general, I also noticed that some `third-party` code may contain proprietary source code additions, so this makes it a little difficult to reuse ship code from external sources.

I assumes in `meson.build` several approaches `system`, `external` and `boundled`. Two approaches `system` and `external` is easy to implement but `boundled` require create `meson.build` in `third-party` directory manually.

If you get green, I can try to write these `meson.build`, but it is look like reinvent already exists `meson.build` provided by that module owners.

I could probably try to reuse ready-made modules from the manufacturer with minor changes.

**Comment by sauwming** [comment]
Okay, I understand your point about the third party build.

I just retested using the Meson build and IMO, it's still far from usable.
For example, I can't seem to be able to build with video enabled, and when I tried to run pjsua with audio-only build, it triggered an assertion:
```
 builddir % ./pjsua 
Assertion failed: (a->addr.sa_family == PJ_AF_INET || a->addr.sa_family == PJ_AF_INET6), function pj_sockaddr_get_len, file sock_common.c, line 393.
```

So the current build script does provide a good starting point, and is good for experimental purpose, but if it's not a complete and ready build system, it will only frustrate users and create more problems than help.

In conclusion, unless the Meson build system can have more or less the same capability as the current build system that uses configure script, we will have to put off its integration to the codebase.

I understand that you were not planning to complete this yourself and invite people to contribute, so hopefully someone can continue building on this until it's ready.

vit1251:
> Assertion failed: (a->addr.sa_family == PJ_AF_INET || a->addr.sa_family == PJ_AF_INET6), function pj_sockaddr_get_len, file sock_common.c, line 393.

Hmmm... Require additional source code research and investigate root cause assesion on common C const.
I think that there is wrong or unset common constants but I does not found any reference in original product about it.

PR #3661, opened 2023-08-15
Add configuration for default incoming payload type number for telephone-event

trengginas:
RFC [2833](https://datatracker.ietf.org/doc/html/rfc2833#section-3.3) specify that payload type number for telephone-event/DTMF is set dynamically and out-of-band.
```
  The RTP payload format is designated as "telephone-event", the MIME
   type as "audio/telephone-event". The default timestamp rate is 8000
   Hz, but other rates may be defined. In accordance with current
   practice, this payload format does not have a static payload type
   number, but uses a RTP payload type number established dynamically
   and out-of-band.
```

And further, RFC [4733](https://datatracker.ietf.org/doc/html/rfc4733#section-2.5.1.1) stated this:
```
   For example, if the payload format uses the payload type number 100,
   and the implementation can handle the DTMF tones (events 0 through
   15) and the dial and ringing tones (assuming as an example that these
   were defined as events with codes 66 and 70, respectively), it would
   include the following description in its SDP message:

      m=audio 12346 RTP/AVP 100
      a=rtpmap:100 telephone-event/8000
      a=fmtp:100 0-15,66,70
```

Unfortunately, some endpoint doesn't specify the payload type number of the telephone-event in its SDP.
Hence, the incoming DTMF might be ignored.
This patch will introduce a new configuration to set the default incoming telephone-event/DTMF payload type number.

### DIFF
--- a/pjmedia/src/pjmedia/stream_info.c
+++ b/pjmedia/src/pjmedia/stream_info.c
@@ -304,7 +304,7 @@ static pj_status_t get_audio_codec_info_param(pjmedia_stream_info *si,
 
 
     /* Get incomming payload type for telephone-events */
-    si->rx_event_pt = -1;
+    si->rx_event_pt = PJMEDIA_DTMF_TEL_EVENT_RX_PT;
     for (i=0; i<local_m->attr_count; ++i) {
         pjmedia_sdp_rtpmap r;
 
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -459,6 +459,16 @@
 #  define PJMEDIA_DTMF_DURATION_MSEC            200
 #endif
 
+/**
+ * The payload type number of the DTMF/telephone-event is set dynamically out-of-band. 
+ * In case that remote endpoint doesn't specify the the payload type number 
+ * in the SDP, the incoming DTMF might be ignored (bad RTP PT number). Set this as the 
+ * default payload type number for incoming DTMF/telephone-event if no payload type number
+ * is specified on the SDP.
+ */
+#ifndef PJMEDIA_DTMF_TEL_EVENT_RX_PT
+#  define PJMEDIA_DTMF_TEL_EVENT_RX_PT         -1
+#endif
 
 /**
  * Number of RTP packets received from different source IP address from the


pjmedia/src/pjmedia/stream_info.c
```diff
@@ -304,7 +304,7 @@ static pj_status_t get_audio_codec_info_param(pjmedia_stream_info *si,
 
 
     /* Get incomming payload type for telephone-events */
-    si->rx_event_pt = -1;
+    si->rx_event_pt = PJMEDIA_DTMF_TEL_EVENT_RX_PT;
```
nanangizz, line 307: IMHO this approach sounds hackish because tel-event PT is basically a dynamic PT, which may change between calls or even between SDP offer/answer in the same call (e.g: when codecs are enabled/disabled).

If we want to accept DTMF when remote SDP answer does not specify it in its SDP, perhaps a better approach is to handle it in the SDP negotiation (i.e: do not remove it from the local offer). Then perhaps introduce a config setting for enabling/disabling this behaviour.

Another idea is to allow app overriding `si->rx_event_pt` via PJSUA callback `on_stream_precreate` (and PJSUA2). A disadvantage compared to prev approach is: it may not be straightforward as app may need to check the initial local offer to know the offered PT (which is later removed by the SDP nego as remote does not have tel-event).

sauwming: > If we want to accept DTMF when remote SDP answer does not specify it in its SDP, perhaps a better approach is to handle it in the SDP negotiation (i.e: do not remove it from the local offer). Then perhaps introduce a config setting for enabling/disabling this behaviour.

This sounds okay, but since this is remote's issue, perhaps handling it in the app level would be better? App can modify the SDP answer and add the tel-event manually.

trengginas: This is the SDP nego scenario:
Offer
```
v=0
o=- 3899294638 3899294638 IN IP4 192.168.254.200
s=pjmedia
b=AS:84
t=0 0
a=X-nat:0
m=audio 23002 RTP/AVP 96 97 98 99 3 0 8 9 120 121 122
c=IN IP4 192.168.254.200
b=TIAS:64000
a=rtcp:23003 IN IP4 192.168.254.200
a=sendrecv
a=rtpmap:96 speex/16000
a=rtpmap:97 speex/8000
a=rtpmap:98 speex/32000
a=rtpmap:99 iLBC/8000
a=fmtp:99 mode=30
a=rtpmap:3 GSM/8000
a=rtpmap:0 PCMU/8000
a=rtpmap:8 PCMA/8000
a=rtpmap:9 G722/8000
a=rtpmap:120 telephone-event/16000
a=fmtp:120 0-16
a=rtpmap:121 telephone-event/8000
a=fmtp:121 0-16
a=rtpmap:122 telephone-event/32000
a=fmtp:122 0-16
a=ssrc:415131524 cname:074d4dc8644366bb
```

Answer:
```
v=0
o=- 2616112480 1 IN IP4 192.168.254.6
c=IN IP4 192.168.254.6
t=0 0
m=audio 50080 RTP/AVP 0
a=rtpmap:0 PCMU/8000
a=ptime:20
```

Later, remote/callee sends DTMF with PT 122. Not clear how they choose the PT based on the offer, I assume it is the latest tel-event.

sauwming: The latest commit has even more limited usage, since it only allows the last telephone-event PT.

I would first suggest the app to add the answers themselves, This should be done considering their understanding of the problematic remote. As the issue lies with the remote itself, there is no need for a library fix at the moment.

ilogixxDE: If it is possible for the app to override the telefone-event PT, than this would be the best approach, at least for us. So we can handle this differently for different PBXes.

trengginas: @sauwming, I leaning to agree with you. It's not really clear how the remote choose the PT used. @ilogixxDE , app can modify the SDP answer to include the telephone-event used.

ilogixxDE: It would be great, if we can handle the SDP in PJSUA or at least have an option at runtime to set the expected PT Type in case we are facing a system that behaves like the one we have described. We expect to run into similar issues with other PBXes too, so it would be great if we have an easy way to make DTMF work, similar to this PR. Can this be added somehow as an option in PJSUA, please?

nanangizz:
In [RFC3264 section 6.1](https://www.rfc-editor.org/rfc/rfc3264.html#section-6.1):
```
   The answerer MUST send using a media format in the offer
   that is also listed in the answer, and SHOULD send using the most
   preferred media format in the offer that is also listed in the answer.
```
So I agree with @sauwming, it is a remote/answerer's issue. If we want to accomodate this, the modification/patch has to be a simple one, e.g.: few liners in one location, so it is relatively easy to read & maintain (note that if we decide to merge it, we will need to maintain it for bugs, compatibility, etc). And unfortunately it looks like the current patch is a bit more complex than I expected.

So IMHO, the alternatives are now perhaps:
 - Allow app overriding `si->rx_event_pt` via PJSUA callback `on_stream_precreate` as mentioned before, the patch in the lib should be simple and low risk (one liner in code plus updated docs perhaps). To avoid multiple PTs offered, `PJMEDIA_TELEPHONE_EVENT_ALL_CLOCKRATES` can be disabled/set-to-zero. Then application can simply override  `si->rx_event_pt` to `PJMEDIA_RTP_PT_TELEPHONE_EVENTS`.
 - Handle this in the app as suggested by @sauwming by modifying the SDP answer.

sauwming:
> * Allow app overriding `si->rx_event_pt` via PJSUA callback `on_stream_precreate` as mentioned before, the patch in the lib should be simple and low risk (one liner in code plus updated docs perhaps). To avoid multiple PTs offered, `PJMEDIA_TELEPHONE_EVENT_ALL_CLOCKRATES` can be disabled/set-to-zero. Then application can simply override  `si->rx_event_pt` to `PJMEDIA_RTP_PT_TELEPHONE_EVENTS`.

For @nanangizz 's proposal above, you can try this patch below that I created a few weeks ago:
[aud_rx_event_pt.txt](https://github.com/pjsip/pjproject/files/12594039/aud_rx_event_pt.txt)
Let us know if it works for you.

sauwming:
We used the approach with `on_stream_precreate()` callback in #3705

PR #3673, opened 2023-08-29
Update video handling in Android PJSUA2/SWIG sample apps

nanangizz:
The update includes:
- simplify the code for readibility (combine surface callback handler for incoming video & local preview, remove show/hide preview button, etc)
- remote video surface is shown after format change event (usually indication of first successful decoding) to avoid black/green screen
- maintain video aspect ratio for both incoming video and local preview
- add video capability in Kotlin sample app
- add run-time Android permission check
- update the Android build config version (gradle version, etc)

### DIFF
--- a/pjsip-apps/src/swig/java/android/app/src/main/java/org/pjsip/pjsua2/app/MyApp.java
+++ b/pjsip-apps/src/swig/java/android/app/src/main/java/org/pjsip/pjsua2/app/MyApp.java
@@ -33,6 +33,7 @@ interface MyAppObserver
     abstract void notifyCallMediaState(MyCall call);
     abstract void notifyBuddyState(MyBuddy buddy);
     abstract void notifyChangeNetwork();
+    abstract void notifyCallMediaEvent(MyCall call, OnCallMediaEventParam prm);
 }
 
 
@@ -51,10 +52,14 @@ class MyCall extends Call
     public VideoWindow vidWin;
     public VideoPreview vidPrev;
 
+    public boolean vidPrevStarted;
+
     MyCall(MyAccount acc, int call_id)
     {
         super(acc, call_id);
         vidWin = null;
+        vidPrev = null;
+        vidPrevStarted = false;
     }
 
     @Override
@@ -65,6 +70,9 @@ class MyCall extends Call
             if (ci.getState() == 
                 pjsip_inv_state.PJSIP_INV_STATE_DISCONNECTED)
             {
+                if (vidPrevStarted)
+                    vidPrev.stop();
+
                 MyApp.ep.utilLogWrite(3, "MyCall", this.dump(true, ""));
             }
         } catch (Exception e) {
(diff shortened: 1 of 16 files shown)


sauwming:
Bitrise fails to build, although it shows 404 everytime I click on it.

I tried to build the sample app locally and got a few build errors, such as:
```
pjsip-apps/src/swig/java/android/app/src/main/java/org/pjsip/pjsua2/app/MyApp.java:135: error: bad operand types for binary operator '&'
                if ((cmi.getDir() & pjmedia_dir.PJMEDIA_DIR_ENCODING) != 0) {
                                  ^
```

nanangizz:
Strange, it did not happen here, just retried on both Android Studio & SWIG `make java`.

Just in case, `cmi.getDir()` returns `int` here:
```
  public int getDir() {
    return pjsua2JNI.CallMediaInfo_dir_get(swigCPtr, this);
  }
```
And `pjmedia_dir.PJMEDIA_DIR_ENCODING` also `int`:
```
public final static int PJMEDIA_DIR_ENCODING = 1;
```
Could you check them on yours?

nanangizz:
Also the desktop CI Actions should also build the SWIG bindings including Java?

sauwming:
Ah, thanks for the pointer, there are conflicting definition on my workdir here, which doesn't get cleaned by make clean or realclean, so I remove it manually:
```
./pjsip-apps/src/swig/java/android/app/src/main/java/org/pjsip/pjsua2/CallMediaInfo.java:  public pjmedia_dir getDir() {
./pjsip-apps/src/swig/java/android/pjsua2/src/main/java/org/pjsip/pjsua2/CallMediaInfo.java:  public int getDir() {
```

The next problem I encountered is:
`The project is using an incompatible version (AGP 7.4.2) of the Android Gradle plugin. Latest supported version is AGP 7.3.1`
Which seems to require me to install a later version of Android Studio.
But changing it to 7.3.1 seems to work okay, so I suppose we should reduce the dependency requirement as minimal as possible.

Everything works fine here now.

pjsip-apps/src/swig/java/android/app/src/main/java/org/pjsip/pjsua2/app/MyApp.java
```diff
@@ -373,6 +408,9 @@ public void init(MyAppObserver obs, String app_dir,
             System.out.println(e);
         }
 
+        // TCP & TLS transports with default SIP ports
+        // sometimes not working fine on Android somehow.
```
sauwming, line 412: It works fine on my device, so I have to uncomment it here.
Perhaps you can try using non-default ports rather than disabling this entirely?

nanangizz:
Re: gradle version. Not sure which gradle version should be used here, there are too many, checked the web it looks like the latest is 8.3, so 7.x actually sounds rather old already. Considering Android Studio is a bit aggressive in suggesting updates, I guess updating IDE/SDK/tools versions is like a daily routine for proper Android developers (definitely not me :D)

sauwming:
Everything looks okay to me, but the bitrise failure will need to be fixed.

sauwming (approved):

trengginas (approved):

PR #3679, opened 2023-09-01
update rust bindgen 0.60.1 -> 0.66.1

JRazek:
When generating Rust bindings with llvm-16, one can observe an error with invalid Ident:

```
  "pj_timestamp_struct_(unnamed_at_/usr/include/pj/types_h_135_5)" is not a valid Ident
```
This is a bug, that was fixed in the 0.66.1 version of Rust bindgen.
This PR updates the dependency.

Original upstream issue: 
https://github.com/rust-lang/rust-bindgen/issues/2312

### DIFF
--- a/pjsip-apps/src/rust/demo/Cargo.lock
+++ b/pjsip-apps/src/rust/demo/Cargo.lock
@@ -3,59 +3,33 @@
 version = 3
 
 [[package]]
-name = "aho-corasick"
-version = "0.7.20"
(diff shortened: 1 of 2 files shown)


sauwming (approved):

nanangizz (approved):

sauwming:
I believe this should also fix https://github.com/pjsip/pjproject/security/dependabot/1

PR #3683, opened 2023-09-04
Add on_rejected_incoming_call() callback

trengginas:
Consider the scenario: (on iOS)
- User agent received push notification of a new call
- the call is registered to iOS CallKit, and show to user notification about incoming call. wait for SIP INVITE
- SIP INVITE is received, however is rejected by the library due to invalid message
- Notification to the incoming call still shown since the application is not aware of the call rejection

This PR will add a new callback `on_rejected_incoming_call()` to notifiy the application in case incoming call is rejected.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -1106,7 +1106,8 @@ typedef struct pjsua_on_rejected_incoming_call_param {
     pj_str_t        st_text;
 
     /** 
-     * The original INVITE message.
+     * The original INVITE message, if it's not available this will be set
+     * to NULL.
      */
     pjsip_rx_data  *rdata;
     
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -1085,6 +1085,12 @@ typedef struct pjsua_call_setting
  * \a pjsua_on_rejected_incoming_call_cb.
  */
 typedef struct pjsua_on_rejected_incoming_call_param {
+    /**
+     * The incoming call id. This will be set to PJSUA_INVALID_ID when there is
+     * no available call slot at the time.
+     */
+    pjsua_call_id   call_id;
+
     /** 
      * Local URI.
      */
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -2019,7 +2019,9 @@ typedef struct pjsua_callback
     void (*on_media_event)(pjmedia_event *event);
 
     /**
-     * This callback is called when an incoming call is rejected.
+     * This callback will be invoked when the library implicitly rejects
+     * an incoming call.
+     * 
      * In addition to being declined explicitly using the 
      * #pjsua_call_answer()/#pjsua_call_answer2() method,
      * the library may also automatically reject the incoming call due 
@@ -2028,7 +2030,7 @@ typedef struct pjsua_callback
      * - no available account to handle the call.
      * - when an incoming INVITE is received with, for instance, a message 
      *   containing invalid SDP.
-     * 
+     *
      * See also #pjsua_on_rejected_incoming_call_cb.
      */
     pjsua_on_rejected_incoming_call_cb on_rejected_incoming_call;
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -1097,17 +1097,17 @@ typedef struct pjsua_on_rejected_incoming_call_param {
     /** 
      * Rejection code.
      */
-    int             code;
+    int             st_code;
 
     /**
-     * Original INVITE message.
+     * Rejection text.
      */
-    pjsip_rx_data *rdata;
+    pj_str_t        st_text;
 
     /** 
-     * The rejection message.
+     * The original INVITE message.
      */
-    pjsip_tx_data *tdata;
+    pjsip_rx_data  *rdata;
     
     /** 
      * Internal.
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -1081,6 +1081,50 @@ typedef struct pjsua_call_setting
 
 
 /**
+ * This will contain the information of the callback \a on_rejected_incoming_call.
+ */
+typedef struct pjsua_on_rejected_incoming_call_param {
+    /** 
+     * Local URI 
+     */
+    pj_str_t        local_info;
+
+    /** 
+     * Remote URI 
+     */
+    pj_str_t        remote_info;
+
+    /** 
+     * Rejection code 
+     */
+    int             code;
+
+    /** 
+     * Original rejection message 
+     */
+    pjsip_tx_data  *tdata;
+    
+    /** 
+     * Internal.
+     */
+    struct {
+        char    local_info[PJSIP_MAX_URL_SIZE];
+        char    remote_info[PJSIP_MAX_URL_SIZE];
+    } buf_;
+
+} pjsua_on_rejected_incoming_call_param;
+
+/**
+  * Type of callback to be called when incoming call is rejected.
+  *
+  * @param param      The rejected call information.
+  *
+  */
+typedef void (*pjsua_on_rejected_incoming_call_cb)(
+                                      const pjsua_on_rejected_incoming_call_param *param);
+
+
+/**
  * This structure describes application callback to receive various event
  * notification from PJSUA-API. All of these callbacks are OPTIONAL,
  * although definitely application would want to implement some of
@@ -1966,6 +2010,20 @@ typedef struct pjsua_callback
      * @param event     The media event.
      */
     void (*on_media_event)(pjmedia_event *event);
+
+    /**
+     * This callback is called when an incoming call is rejected.
+     * In addition to being declined explicitly using the #pjsua_call_answer() method,
+     * the library may also automatically reject the incoming call due 
+     * to different scenarios:
+     * - when an incoming INVITE is received with, for instance, a message 
+     *   containing invalid SDP.
+     * - when issues occur during the #pjsua_call_answer() process, 
+     *   such as a failure to initialize call media.
+     * 
+     * See also #pjsua_on_rejected_incoming_call_cb.
+     */
+    pjsua_on_rejected_incoming_call_cb on_rejected_incoming_call;
 
 } pjsua_callback;
 
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -2020,11 +2020,11 @@ typedef struct pjsua_callback
      * This callback is called when an incoming call is rejected.
      * In addition to being declined explicitly using the #pjsua_call_answer() method,
      * the library may also automatically reject the incoming call due 
-     * to different scenarios:
+     * to different scenarios, e.g:
+     * - no available call slot.
+     * - no available account to handle the call.
      * - when an incoming INVITE is received with, for instance, a message 
      *   containing invalid SDP.
-     * - when issues occur during the #pjsua_call_answer() process, 
-     *   such as a failure to initialize call media.
      * 
      * See also #pjsua_on_rejected_incoming_call_cb.
      */
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -1085,24 +1085,29 @@ typedef struct pjsua_call_setting
  */
 typedef struct pjsua_on_rejected_incoming_call_param {
     /** 
-     * Local URI 
+     * Local URI.
      */
     pj_str_t        local_info;
 
     /** 
-     * Remote URI 
+     * Remote URI.
      */
     pj_str_t        remote_info;
 
     /** 
-     * Rejection code 
+     * Rejection code.
      */
     int             code;
 
-    /** 
-     * Original rejection message 
+    /**
+     * Original INVITE message.
      */
-    pjsip_tx_data  *tdata;
+    pjsip_rx_data *rdata;
+
+    /** 
+     * The rejection message.
+     */
+    pjsip_tx_data *tdata;
     
     /** 
      * Internal.
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -1081,7 +1081,8 @@ typedef struct pjsua_call_setting
 
 
 /**
- * This will contain the information of the callback \a on_rejected_incoming_call.
+ * This will contain the information passed from the callback 
+ * \a pjsua_on_rejected_incoming_call_cb.
  */
 typedef struct pjsua_on_rejected_incoming_call_param {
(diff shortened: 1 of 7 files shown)


wosrediinanatour:
Does this handle INVITE crossover, i.e. SIP 491?

Having the SIP message, that has been rejected, would/could be useful. Maybe by using sip_event.

Isn't it implemented intentionally in PJSUA2?

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -1080,6 +1080,50 @@ typedef struct pjsua_call_setting
...
+
+    /** 
+     * Rejection code 
+     */
+    int             code;
+
+    /** 
+     * Original rejection message 
+     */
+    pjsip_tx_data  *tdata;
```
nanangizz, line 1105: Perhaps it'd be better to provide `rdata` too if possible, in case app wants to examine the original request.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -2129,6 +2129,60 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
...
+            param.local_info.ptr = param.buf_.local_info;
+            uri = (pjsip_sip_uri*)pjsip_uri_get_uri(to_hdr->uri);
+            param.local_info.slen = pjsip_uri_print(PJSIP_URI_IN_FROMTO_HDR,
+                                                     uri,
+                                                     param.buf_.local_info,
+                                                     sizeof(param.buf_.local_info));
+        }
+
+        pjsua_var.ua_cfg.cb.on_rejected_incoming_call(&param);
+    }
```
nanangizz, line 2183: This should be called on initial INVITE only, usually no To-tag.

sauwming:
This approach seems to be more complicated than expected.

Some alternative ideas:
a. Add `on_incoming_invite()`. We simply call this new callback in the beginning of `pjsua_call_on_incoming()`.
Pro: very simple. Con: Even though app can know if an incoming INVITE is eventually rejected by the library, it will not know the exact reason why.
b. Invoke `on_rejected_incoming_call()` from `pjsua_call_on_incoming()` instead.
Pro: allow more customization (app can change the status code/reason, perhaps even override the rejection, etc).

Personally, I like option b.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -1453,6 +1453,155 @@ on_incoming_call_med_tp_complete(pjsua_call_id call_id,
...
+            uri = (pjsip_sip_uri*)pjsip_uri_get_uri(from_hdr->uri);
+            param.remote_info.slen = pjsip_uri_print(PJSIP_URI_IN_FROMTO_HDR,
+                uri,
+                param.buf_.remote_info,
+                sizeof(param.buf_.remote_info));
+        }
+
+        to_hdr = rdata->msg_info.to;
+        if (to_hdr) {
+            param.local_info.ptr = param.buf_.local_info;
```
wosrediinanatour, line 1488: As far as I understand, `buf_` is only used locally, and has to live for the lifetime of the callback below.
So why not having it locally only (nicer `pjsua_on_rejected_incoming_call_param`)? Why not using a memory pool?

sauwming:
A couple of spec questions to consider:
- Is there any need for the user to get the actual rejection message in the form of `tdata`? So far, pjsua callback usually only provides the status code (and st_text) of the answer. If we don't need the tdata, it will be much simpler, you don't need to copy the implementation of `pjsip_endpt_respond/respond_stateless()` into `pjsua_call`.
- Do we want to call the callback before, or after, rejecting the call?
  If before, the callback will allow room for customisation, such as permitting user to modify the status code and text. But we may need to change the callback name since the call has not actually been rejected yet.
  If after, the callback will serve as informational purpose only.

trengginas:
> A couple of spec questions to consider:
> 
> * Is there any need for the user to get the actual rejection message in the form of `tdata`? So far, pjsua callback usually only provides the status code (and st_text) of the answer. If we don't need the tdata, it will be much simpler, you don't need to copy the implementation of `pjsip_endpt_respond/respond_stateless()` into `pjsua_call`.
> * Do we want to call the callback before, or after, rejecting the call?
>   If before, the callback will allow room for customisation, such as permitting user to modify the status code and text. But we may need to change the callback name since the call has not actually been rejected yet.
>   If after, the callback will serve as informational purpose only.



> A couple of spec questions to consider:
> 
> * Is there any need for the user to get the actual rejection message in the form of `tdata`? So far, pjsua callback usually only provides the status code (and st_text) of the answer. If we don't need the tdata, it will be much simpler, you don't need to copy the implementation of `pjsip_endpt_respond/respond_stateless()` into `pjsua_call`.
> * Do we want to call the callback before, or after, rejecting the call?
>   If before, the callback will allow room for customisation, such as permitting user to modify the status code and text. But we may need to change the callback name since the call has not actually been rejected yet.
>   If after, the callback will serve as informational purpose only.

- I agree to not include `tdata` on the callback param. It will simplify the solution. Adding `st_text` as parameter should also be useful.
- I think we should not allow application to override the/change the status code and it should be informational only.

pjsip/src/pjsua-lib/pjsua_call.c
trengginas: It is using the same pattern as in `pjsua_call_info` (using `buf_` as internal buffer).

wosrediinanatour:
> > A couple of spec questions to consider:
> > 
> > * Is there any need for the user to get the actual rejection message in the form of `tdata`? So far, pjsua callback usually only provides the status code (and st_text) of the answer. If we don't need the tdata, it will be much simpler, you don't need to copy the implementation of `pjsip_endpt_respond/respond_stateless()` into `pjsua_call`.
> > * Do we want to call the callback before, or after, rejecting the call?
> >   If before, the callback will allow room for customisation, such as permitting user to modify the status code and text. But we may need to change the callback name since the call has not actually been rejected yet.
> >   If after, the callback will serve as informational purpose only.
> 
>     * I agree to not include `tdata` on the callback param. It will simplify the solution. Adding `st_text` as parameter should also be useful.
> 
>     * I think we should not allow application to override the/change the status code and it should be informational only.

The transmitted message is very useful for tracing, when onCallTsxState(...) is not called.

pjsip/src/pjsua-lib/pjsua_call.c
wosrediinanatour: I see. I am curious why. :-)

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -1967,6 +2016,20 @@ typedef struct pjsua_callback
      */
     void (*on_media_event)(pjmedia_event *event);
 
+    /**
+     * This callback is called when an incoming call is rejected.
```
sauwming, line 2020: Fix the doc.

pjsip/include/pjsua2/endpoint.hpp
```diff
@@ -436,6 +436,37 @@ struct OnMediaEventParam
...
+     * Local URI.
+     */
+    std::string     localInfo;
+
+    /**
+     * Remote URI.
+     */
+    std::string     remoteInfo;
+
+    /**
```
sauwming, line 454: Use statusCode and reason, to be consistent with other pjsua2 callbacks

pjsip/include/pjsua2/endpoint.hpp
```diff
@@ -1898,6 +1929,14 @@ class Endpoint
      */
     virtual pj_status_t onCredAuth(OnCredAuthParam &prm);
 
+    /**
+     * Callback when an incoming call is rejected.
```
sauwming, line 1933: Complete the doc.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -1475,6 +1519,9 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
     pjmedia_sdp_session *offer=NULL;
     pj_bool_t should_dec_dlg = PJ_FALSE;
     pjsip_tpselector tp_sel;
+    char reason_buf[32] = {0};
```
sauwming, line 1522: what's the buffer for? doesn't seem to be used

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -1453,6 +1453,50 @@ on_incoming_call_med_tp_complete(pjsua_call_id call_id,
     return on_incoming_call_med_tp_complete2(call_id, info, NULL, NULL, NULL);
 }
 
+static void rejected_incoming_call_cb(pjsip_rx_data *rdata, int st_code, 
+                                      pj_str_t *st_text)
+{
+    /* Call on_rejected_incoming_call() only if to-tag is not set. */
```
sauwming, line 1459: No need to check To tag since it's not called from `on_tx_response()`.

sauwming:
> The transmitted message is very useful for tracing, when onCallTsxState(...) is not called.

By "tracing", you mean for info only, correct?
Since it doesn't have any practical purpose, it's not strong enough reason to include `tdata` within the callback, considering the complexity to do so.

wosrediinanatour:
> > The transmitted message is very useful for tracing, when onCallTsxState(...) is not called.
> 
> By "tracing", you mean for info only, correct? Since it doesn't have any practical purpose, it's not strong enough reason to include `tdata` within the callback, considering the complexity to do so.

Yes.

pjsip/include/pjsua-lib/pjsua.h
sauwming: The callback is now called only when it's internally rejected by the library, right? Will using `pjsua_call_answer()` also trigger the callback?

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -1475,6 +1516,8 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
     pjmedia_sdp_session *offer=NULL;
     pj_bool_t should_dec_dlg = PJ_FALSE;
     pjsip_tpselector tp_sel;
+    pj_str_t st_reason = pj_str("");
+    int st_code = 200;
```
nanangizz, line 1520: Why not initialize `st_code` with zero?

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -1965,6 +2023,7 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
             pjsua_perror(THIS_FILE, "Error initializing media channel", status);
             
             pjsip_dlg_inc_lock(dlg);
+            st_code = sip_err_code;
```
nanangizz, line 2026: When `pjsua_media_channel_init()` is done async-ly or returning `EPENDING`, e.g: due to waiting for TURN allocations, the library may reject the call if the media init fails, should we invoke the callback too?

sauwming: This is an interesting scenario indeed. And note that `on_incoming_call()` has been called for such case, whereas for other cases, it hasn't.

pjsip/src/pjsua-lib/pjsua_call.c
trengginas: It's following the original code on https://github.com/pjsip/pjproject/blob/05d03ad9ee4eb4d93f63d2eb7b4345e1ad49e615/pjsip/src/pjsua-lib/pjsua_call.c#L1561

pjsip/src/pjsua-lib/pjsua_call.c
trengginas: Since the `on_incoming_call()` has been called, the new callback (`on_rejected_incoming_call()`) should not be called again. Another case would be if the called is answered with 200OK, however there's an issue (e.g: media init) and the call is rejected. The callback would not be called, however app still be informed from `on_call_state()`.

sauwming: Make sure the behavior is consistent on pjsua2 as well, since on pjsua2 onIncomingCall() is called earlier.

pjsip/src/pjsua-lib/pjsua_call.c
nanangizz: > It's following the original code on
> 
> https://github.com/pjsip/pjproject/blob/05d03ad9ee4eb4d93f63d2eb7b4345e1ad49e615/pjsip/src/pjsua-lib/pjsua_call.c#L1561

Oh, "the original code" is actually the default answer code for replaced dialog (may be overriden by app from the replaced dlg callback). IMO better not share the same var with the replaced dialog, it serves different purpose, it may cause confusion in the future.

pjsip/src/pjsua-lib/pjsua_call.c
nanangizz: As it is async, I guess there can be race between `on_incoming_call()` and `on_incoming_call_med_tp_complete()`, in other word, we cannot be sure whether `on_incoming_call()` has been called?

nanangizz (approved):

pjsip/include/pjsua-lib/pjsua.h
sauwming: Minor:
make it clear in the doc that the callback is only called for calls that are internally rejected by the library. And the callback won't be called if user rejects the call using `pjsua_call_answer/answer2()`.

trengginas: No, the callback will not trigger when `pjsua_call_answer()` is used to reject the call

sauwming (approved):

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -1625,8 +1684,8 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
     } else {
         acc_id = call->acc_id = pjsua_acc_find_for_incoming(rdata);
         if (acc_id == PJSUA_INVALID_ID) {
-            pjsip_endpt_respond_stateless(pjsua_var.endpt, rdata,
-                                          PJSIP_SC_TEMPORARILY_UNAVAILABLE,
+            ret_st_code = PJSIP_SC_TEMPORARILY_UNAVAILABLE;
```
wosrediinanatour, line 1687: Here you are using enums, in other files integers. Is there a reason for different approaches?

trengginas: Usually we avoid unnecessary changes, however with the use of the new var we can also change the status code to using the enum.

sauwming:
Sorry for rather late comment. But maybe we should consider adding `call_id` to the callback? So app can easily tell: - which call is getting internally rejected by the library, and - whether `on_incoming_call()` has ever been called (I suppose if `call_id == PJSUA_INVALID_ID` it means that app has not been notified of such call before).

sauwming:
looks okay.

sauwming:
Just a few additional notes:
The last commit won't be able to help app know if the callback `on_incoming_call()`  has been called.
And I still have some reservations about whether there can be any issue caused by the overlap between `on_call_state(DISCONNECTED)` and `on_rejected_incoming_call()`.

But I think the PR can be merged first for now.

If there's any feedback @anikitin-intermedia, please feel free to inform us.

PR #3690, opened 2023-09-08
Enable returning IPv4 mapped IPv6 address on getaddrinfo()

### LINKED ISSUE #3681: iOS and IPv6 problems (again)
### Describe the bug

If we use IPv6 on iOS (15 and 16) we get several 'bugs'

1. sip_reg.c Invalid Contact: "<sip:xxxxx@2a05:123:4556::1:52081;ob>"
We don't set the contact manually. The [] are missing (my opinion)

2. If we use a name as proxy it will not resolved in IPv6
We can work arround this by getting the IPV6 address 'by hand' (getaddrinfo() with hint AF_INET6)
and set it by acc_cfg.proxy.0 = pj_str("sip:[xxxxx]:5061")
This works. Without [] it does not work. Against the documentation where mentioned that no [] should be used.

But if we have a network change we can not detect which transport is now used to set the new proxy.
(we need to set the new proxy since we wrote the IP address as proxy and not the name)
How can we detect the current used transport for an account?
We tried different things like pjsua_transport_get_info() and looked at usage_count,
the on_ip_change_progress_cb and looked at transport_id but all transports are affected and not only the current used one.

### Steps to reproduce

Set a proxy by name and use IPv6 in an iOS app.

### PJSIP version

2.13.1

### Context

several iPhones with iOS 15 and iOS 16

### Log, call stack, etc

```shell
sip_reg.c Invalid Contact: "<sip:xxxxx@2a05:123:4556::1:52081;ob>"


sip_resolve.c DNS resolver not available, target 'xxx.ednt.eu:5061' type=UDP will be resolved with getaddrinfo()
sip_resolve.c Failed to resolve 'xxx.ednt.eu': gethostbyname() has returned error (PJ_ERESOLVE)
```

trengginas:
This is to fix #3681 

This ticket will add `AI_V4MAPPED` and `AI_ALL` to `ai_flags` when calling `getaddrinfo()` for IPv6.
It will enable `getaddrinfo()` to return IPv4 mapped IPv6 addresses, by default it will return the AAAA IPv6 only addresses.

```
If the AI_ALL flag is used in conjunction with the AI_V4MAPPED flag,
         and only used with the IPv6 address family.  When AI_ALL is logically
         or'd with AI_V4MAPPED flag then the caller wants all addresses: IPv6
         and IPv4-mapped IPv6.  A query is first made for AAAA records and if
         successful, the IPv6 addresses are returned.  Another query is then
         made for A records and any found are returned as IPv4-mapped IPv6
         addresses.  h_length will be 16.  Only if both queries fail does the
         function return a NULL pointer.  This flag is ignored unless af
         equals AF_INET6.  If both AI_ALL and AI_V4MAPPED are specified,
         AI_ALL takes precedence.
```
[ref](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getipnodebyname.3.html)

### DIFF
--- a/pjlib/src/pj/addr_resolv_sock.c
+++ b/pjlib/src/pj/addr_resolv_sock.c
@@ -178,6 +178,9 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
     /* Call getaddrinfo() */
     pj_bzero(&hint, sizeof(hint));
     hint.ai_family = af;
+    if (af == PJ_AF_INET6) {
+        hint.ai_flags = AI_V4MAPPED | AI_ALL;
+    }
     /* Zero value of ai_socktype means the implementation shall attempt
      * to resolve the service name for all supported socket types */
     hint.ai_socktype = 0;


sauwming:
Closing this as this patch is not verified to be able to fix the problem.

PR #3691, opened 2023-09-11
build: resolve aconfigure location

gjasny:
Come package managers like Conan prefer to use `configure` over `aconfigure`. Doing so right now fails if the working directory is not in the root of the source tree.

### DIFF
--- a/configure
+++ b/configure
@@ -1,5 +1,7 @@
 #!/bin/sh
-./aconfigure "$@"
+
+script_dir=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd -P)
+"${script_dir}/aconfigure" "$@"
 
 # Note:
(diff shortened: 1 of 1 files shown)


nanangizz:
I am not familiar with Conan, just curious, why not change the working directory first before invoking `configure` as instructed by [the docs](https://docs.pjsip.org/en/latest/get-started/posix/build_instructions.html#configure): `cd pjproject` then `./configure`? The `aconfigure` script itself also seems to require to be executed from the root of the source tree (e.g: `ac_pjdir` is set using `pwd` [here](https://github.com/pjsip/pjproject/blob/ad152c2f42c3e067b8b280f18572a009a13d32d4/aconfigure.ac#L74-L86)).

sauwming:
Closing this due to lack of follow-up/update. Please reopen if there's more information.

Issue #1773, opened 2014-06-19
Crash caused by race condition between transport destroy and callback


pjsipbot:
_2014-06-19 10:39:18: @nanangizz_ _created the issue on trac ticket [1773](https://trac.pjsip.org/repos/ticket/1773)_

Call stack:
```
Thread 14 Crashed:
pj_list_insert_nodes_before (list_i.h:25)
int_parse_msg (sip_parser.c:1033)
pjsip_parse_rdata (sip_parser.c:761)
pjsip_tpmgr_receive_packet (sip_transport.c:1651)
on_data_read (sip_transport_tcp.c:1292)
ioqueue_on_read_complete (activesock.c:495)

Thread 15:
pj_mutex_lock (os_core_unix.c:1243)
pj_ioqueue_unregister (ioqueue_common_abs.c:1326)
pj_activesock_close (activesock.c:305)
tcp_destroy (sip_transport_tcp.c:775)
tcp_destroy_transport (sip_transport_tcp.c:724)
destroy_transport (sip_transport.c:1092)
pj_timer_heap_poll (timer.c:748)
```

The call stack shows that the transport callback is being invoked when transport destroy is on progress. As reported, a possible scenario is that the `tp_info` of the `rdata` is accessed after the memory pool has been released.

Thanks Itay Bianco for the report.
___
_2014-06-19 10:42:04: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2014-06-19 10:42:04: @nanangizz_ _set **resolution** to **fixed**_

___
_2014-06-19 10:42:04: @nanangizz_ _commented_

In [r4862](../commit/6cbe3651b59c4747ebba409d65c9e00cf5b7df94):
Fix #1773: Added group lock to SIP transport to avoid race condition between transport callback and destroy.

PR #3694, opened 2023-09-12
rx_data_clone-check need check transport's state,  to avoid crashes in subsequent use.

Tinet-mucw:
In a multi-threaded environment, when thread1 executes pjsip_transport_shutdown and thread2 executes pjsip_tpmgr_receive_packet: 1)The entire process in pjsip_tpmgr_receive_packet does not validate the state of the transport. 2) In pjsip_tpmgr_receive_packet, subsequent to (*mod->on_rx_request)(rdata), when executing pjsip_rx_data_clone, tp->ref_cnt happens to be 0, and is_shutdown is TRUE. 3) Later, when generating a response from the received request (rx) and executing pjsip_endpt_send_response2, it leads to a crash. I believe that when executing pjsip_rx_data_clone, it is necessary to check the transport state to avoid crashes in subsequent use.

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -773,16 +773,28 @@ PJ_DEF(pj_status_t) pjsip_rx_data_clone( const pjsip_rx_data *src,
     pj_pool_t *pool;
     pjsip_rx_data *dst;
     pjsip_hdr *hdr;
+    pj_status_t clone_status;
+    pjsip_transport *src_tp = (pjsip_transport*)src->tp_info.transport;
 
     PJ_ASSERT_RETURN(src && flags==0 && p_rdata, PJ_EINVAL);
+
+    pj_lock_acquire(src_tp->lock);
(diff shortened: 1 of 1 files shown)


nanangizz:
The `pjsip_tpmgr_receive_packet()` is normally invoked from ioqueue read callback which should be proceeded by add-ref to the transport group lock (shared group lock with ioqueue key), code is [here](https://github.com/pjsip/pjproject/blob/ad152c2f42c3e067b8b280f18572a009a13d32d4/pjlib/src/pj/ioqueue_select.c#L1088C18-L1088C18), so theoretically the group lock ref-count should never reach 0 there. The `tp->ref_cnt` may reach 0 but the transport pool/atomic/etc will not be released until group lock ref-count reaches 0, so it should not cause crash.

Currently all built-in PJSIP SIP transports are equipped with group lock, shared with its lower level socket/ioqueue-key. Any custom SIP transports are strongly adopt the approach as otherwise there may be other synchronization issues (this case seems to be one of them). See also #1773.

Tinet-mucw:
Thank you for your response.

PR #3707, opened 2023-09-19
Fixed Coverity and build warnings

sauwming:
Fixed various Coverity and build warnings:
- Unused variable values
- Uninitialised AudioMediaPort class members
- printf warnings

### DIFF
--- a/pjlib/src/pj/ssl_sock_imp_common.c
+++ b/pjlib/src/pj/ssl_sock_imp_common.c
@@ -1134,7 +1134,6 @@ static pj_bool_t ssock_on_accept_complete (pj_ssl_sock_t *ssock_parent,
                                                    ssock->param.grp_lock);
         if (status != PJ_SUCCESS) {
             ssock->timer.id = TIMER_NONE;
-            status = PJ_SUCCESS;
         }
     }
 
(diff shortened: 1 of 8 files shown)


pjlib/src/pj/ssl_sock_imp_common.c
```diff
@@ -1134,7 +1134,6 @@ static pj_bool_t ssock_on_accept_complete (pj_ssl_sock_t *ssock_parent,
                                                    ssock->param.grp_lock);
         if (status != PJ_SUCCESS) {
             ssock->timer.id = TIMER_NONE;
-            status = PJ_SUCCESS;
```
nanangizz, line 1137: Won't this change the behavior (more than fixing a warning)?
The original code seems to start the handshake without a timeout when the timer scheduling fails?
Now, it will trigger handshake failure?

pjlib/src/pjlib-test/ioq_tcp.c
```diff
@@ -402,10 +402,6 @@ static int compliance_test_0(const pj_ioqueue_cfg *cfg)
                 return -52;
             }
             pending_op -= status;
-
-            if (pending_op == 0) {
-                status = 0;
-            }
```
nanangizz, line 408: Also, this does not look like a fixing-warning?
`pending_op == 0` seems to mean that all operations have been completed successfully.

sauwming: The intention of the original code is to reset the status to zero when the loop `while (pending_loop)` completes.

But status will always be overwritten by `status = pj_ioqueue_poll()` below.

nanangizz (approved):

PR #3722, opened 2023-09-29
Fix build errors & warnings on MSVC2005

nanangizz:


pjlib/src/pj/string.c
```diff
@@ -310,7 +310,7 @@ PJ_DEF(pj_status_t) pj_strtol2(const pj_str_t *str, long *value)
         return PJ_ETOOSMALL;
     }
 
-    *value = is_negative ? -retval : retval;
+    *value = is_negative ? -(long)retval : retval;
```
sauwming, line 313: This will cause regression of #3315 and #3426.

pjlib/src/pjlib-test/sleep.c
```diff
@@ -103,7 +103,7 @@ static int sleep_duration_test(void)
     /* Test pj_thread_sleep() and pj_gettimeofday() */
     for (i=0, avg_diff=0, max_diff=0; i<PJ_ARRAY_SIZE(duration); ++i) {
         pj_time_val start, stop;
-        pj_uint32_t msec;
+        long msec;
```
sauwming, line 106: This will cause a different warning:
```
warning: format specifies type 'int' but the argument has type 'long' [-Wformat]
                      msec, duration[i], MAX_SLIP));
```

pjlib/src/pj/string.c
nanangizz: Ah right, tried a different approach in the new commit.

sauwming: Perhaps just changing it to `retval > PJ_MAXLONG` is enough, as described in #3315?
This way, no additional if block is needed.

nanangizz: Right, thought about that, but it will change behavior (i.e: `PJ_MINLONG` will no longer accepted) which generally should be avoid in this kind of patch, so is changing behavior acceptable here?

sauwming: Ah I see.
How about the original patch, i.e. (long)-retval? Will it still trigger warning?

nanangizz: Same warning as before this patch: `warning C4146: unary minus operator applied to unsigned type, result still unsigned`

sauwming (approved):

trengginas (approved):

PR #3729, opened 2023-10-09
Fixed issue when using setting to use AVPF and receive AVP

sauwming:
It is reported that when `pjmedia_rtcp_fb_setting.dont_use_avpf` is set to `PJ_FALSE` (note that the default is PJ_TRUE), we will reject the offer that uses `RTP/AVP` only.

### DIFF
--- a/pjlib/src/pj/ssl_sock_imp_common.h
+++ b/pjlib/src/pj/ssl_sock_imp_common.h
@@ -230,7 +230,7 @@ inline static void io_read(pj_ssl_sock_t *ssock, circ_buf_t *cb,
                            pj_uint8_t *dst, pj_size_t len)
 {
     PJ_UNUSED_ARG(ssock);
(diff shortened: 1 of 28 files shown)


nanangizz:
Just a note (for future ref).

The behavior is different to SRTP. In SRTP we have the _optional_ mode which is lenient (always generate offer using `AVP` and generate answer flexibly, e.g: if offer is using `SAVP` we generates answer using `SAVP` too), the _mandatory_ mode which enforces `SAVP` and the _disabled_ mode which enforces `AVP`.

Here in RTCP-FB, we seems to always generate offer & answer using `AVP` (when `dont_use_avpf==true`) or `AVPF` (when `dont_use_avpf=false`).

nanangizz (approved):

trengginas (approved):

PR #3732, opened 2023-10-09
Fixed DTLS handshake issue when RTCP address changes

### LINKED ISSUE #3724: transport_srtp_dtls RTP/RTCP handshaking
### Describe the bug

Hello, I'm working with PJSIP using SRTP/DTLS.

Commit a0ac247 introduces separate RTP and RTCP channels for DTLS handshaking.

However, I noticed that when the remote address changes during the handshake, the `dtls_on_recv` function reattaches the member transport with potentially incorrect RTCP information (when multiplexing is not enabled).

Since there are now two channels, `dtls_on_recv` gets called separately for RTP and RTCP in no particular order. The logic that handles the transport reattachment only applies to the RTP channel. And when it detects an address change, it attempts to fill in the RTCP address in the attach parameters, but that address is not yet (ever?) valid.

This is the code I'm referring to inside `dtls_on_recv`:
```
if (idx == RTP_CHANNEL && !ds->use_ice && !ds->nego_completed[idx]) {
    pjmedia_transport_info info;
    pjmedia_transport_get_info(ds->srtp->member_tp, &info);
    if (pj_sockaddr_cmp(&ds->rem_addr, &info.src_rtp_name)) {
        pjmedia_transport_attach_param ap;
        pj_status_t status;

        pj_bzero(&ap, sizeof(ap));
        ap.user_data = ds->srtp;
        pj_sockaddr_cp(&ds->rem_addr, &info.src_rtp_name);
        pj_sockaddr_cp(&ap.rem_addr, &ds->rem_addr);
        ap.addr_len = pj_sockaddr_get_len(&ap.rem_addr);
        if (pj_sockaddr_cmp(&info.sock_info.rtp_addr_name,
                            &info.sock_info.rtcp_addr_name) == 0)
        {
            /* Using RTP & RTCP multiplexing */
            pj_sockaddr_cp(&ds->rem_rtcp, &ds->rem_addr);
            pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_rtcp);
        } else if (pj_sockaddr_has_addr(&ds->rem_rtcp)) {
            pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_rtcp);
        } else {
            pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_addr);
            pj_sockaddr_set_port(&ap.rem_rtcp,
                                    pj_sockaddr_get_port(&ds->rem_addr)+1);
        }

        status = pjmedia_transport_attach2(&ds->srtp->base, &ap);
```

From the caller's point of view, the `else if (pj_sockaddr_has_addr(&ds->rem_rtcp))` line is never true, and falls through to adding +1 to the port, which may not be the actual originating RTCP port coming from the callee.

This seems to cause a lot of a delay in the handshake as both sides attempt to renegotiate.

Perhaps `ds->rem_rtcp` needs to be set when `idx == RTCP_CHANNEL` and only when `rem_rtcp` is fully valid, should the main `RTP_CHANNEL` reattach the member transport. Again, only when multiplexing is disabled.

Maybe I'm misinterpreting this mechanism, but in attempt to fix it, I modified this function to set `rem_rtcp` during RTCP recv and for the RTP channel to reattach the transport only when both remote addresses are valid. The handshke now completes a lot more quickly.

Perhaps this is not a critical bug, but more so a discussion on this negotiation mechanism. Please let me know if there is a better place to discuss this.

### Steps to reproduce

1. Connect two endpoints using SRTP/DTLS.
2. Notice a longer delay in SSL handshaking after a0ac247.

### PJSIP version

a0ac247

### Context

- Custom embedded device.

### Log, call stack, etc

```shell
N/A
```

sauwming:
To fix #3724.

When multiplexing is not enabled, it is reported that DTLS reattaches the member transport with potentially incorrect RTCP information during handshake.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -1267,8 +1267,17 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
     /* Check remote address info, reattach member tp if changed */
     if (idx == RTP_CHANNEL && !ds->use_ice && !ds->nego_completed[idx]) {
         pjmedia_transport_info info;
+        pj_bool_t use_rtcp_mux;
+
         pjmedia_transport_get_info(ds->srtp->member_tp, &info);
-        if (pj_sockaddr_cmp(&ds->rem_addr, &info.src_rtp_name)) {
+        use_rtcp_mux = (pj_sockaddr_cmp(&info.sock_info.rtp_addr_name,
+                        &info.sock_info.rtcp_addr_name))? PJ_FALSE: PJ_TRUE;
+        if (pj_sockaddr_cmp(&ds->rem_addr, &info.src_rtp_name) ||
+            (!use_rtcp_mux &&
+             pj_sockaddr_has_addr(&info.src_rtcp_name) &&
+             (!pj_sockaddr_has_addr(&ds->rem_rtcp) ||
+              pj_sockaddr_cmp(&ds->rem_rtcp, &info.src_rtcp_name))))
+        {
             pjmedia_transport_attach_param ap;
             pj_status_t status;
 
@@ -1277,13 +1286,12 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
             pj_sockaddr_cp(&ds->rem_addr, &info.src_rtp_name);
             pj_sockaddr_cp(&ap.rem_addr, &ds->rem_addr);
             ap.addr_len = pj_sockaddr_get_len(&ap.rem_addr);
-            if (pj_sockaddr_cmp(&info.sock_info.rtp_addr_name,
-                                &info.sock_info.rtcp_addr_name) == 0)
-            {
+            if (use_rtcp_mux) {
                 /* Using RTP & RTCP multiplexing */
                 pj_sockaddr_cp(&ds->rem_rtcp, &ds->rem_addr);
                 pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_rtcp);
-            } else if (pj_sockaddr_has_addr(&ds->rem_rtcp)) {
+            } else if (pj_sockaddr_has_addr(&info.src_rtcp_name)) {
+                pj_sockaddr_cp(&ds->rem_rtcp, &info.src_rtcp_name);
                 pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_rtcp);
             } else {
                 pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_addr);
@@ -1300,11 +1308,15 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
 #if DTLS_DEBUG
             {
                 char addr[PJ_INET6_ADDRSTRLEN];
+                char addr2[PJ_INET6_ADDRSTRLEN];
                 PJ_LOG(2,(ds->base.name, "Re-attached transport to update "
-                          "remote addr=%s:%d",
+                          "remote addr=%s:%d remote rtcp=%s:%d",
                           pj_sockaddr_print(&ap.rem_addr, addr,
                                             sizeof(addr), 2),
-                          pj_sockaddr_get_port(&ap.rem_addr)));
+                          pj_sockaddr_get_port(&ap.rem_addr),
+                          pj_sockaddr_print(&ap.rem_rtcp, addr2,
+                                            sizeof(addr2), 2),
+                          pj_sockaddr_get_port(&ap.rem_rtcp)));
             }
 #endif
         }


amubiera (commented):
Hi, I was the one that originally reported the issue on #3724. I was just about to create a pull request, but I see that your change is a bit more thorough than mine.

I believe the `ds->rem_rtcp` address was not available in the original version, but now I see that you're setting it on line 1294.

One question though, since `dtls_on_recv` gets called for each channel in no particular order, the handshake for RTCP may begin with an empty `rem_rtcp` address. But maybe that's ok. One of the endpoints doesn't end up calling `get_rem_addrs` so `rem_addr` and `rem_rtcp` are empty and only set via this recv function.

What I had working was for both channels to update their own address and then check if the everything is valid for a tp change. Whoever is last would submit the reattach. Which seems safe especially now that there is a `DTLS_LOCK` around the entire function.

(The separate `use_rtcp_mux` bool makes it more clear as you have it).

I will try your version soon.

( edit: Now I see that you're using `src_rtcp_name` as the address rather than `ds->rem_rtcp`. That seems more consistent )
( edit: Actually, I don't believe `src_rtcp_name` is valid for the case that RTP comes in first )

```
/* Check remote address info, reattach member tp if changed */
if (!ds->use_ice && !ds->nego_completed[idx]) {
    pjmedia_transport_info info;
    pjmedia_transport_get_info(ds->srtp->member_tp, &info);

    /* Store remote RTP/RTCP addresses as they are received */
    if (idx == RTP_CHANNEL && !pj_sockaddr_has_addr(&ds->rem_addr))
        pj_sockaddr_cp(&ds->rem_addr, &info.src_rtp_name);
    else if (idx == RTCP_CHANNEL && !pj_sockaddr_has_addr(&ds->rem_rtcp))
        pj_sockaddr_cp(&ds->rem_rtcp, &info.src_rtcp_name);

    /* Reattach member tp when:
        *     Have valid RTP address
        * and Have valid src RTP address
        * and RTP address is different than src RTP address
        * and Using multiplexing OR have valid RTCP address
        */
    if (pj_sockaddr_has_addr(&ds->rem_addr) &&
        pj_sockaddr_has_addr(&info.src_rtp_name) &&
        pj_sockaddr_cmp(&ds->rem_addr, &info.src_rtp_name) != 0 &&
        (pj_sockaddr_cmp(&info.sock_info.rtp_addr_name, &info.sock_info.rtcp_addr_name) == 0 ||
         pj_sockaddr_has_addr(&ds->rem_rtcp))) {
        pjmedia_transport_attach_param ap;
        pj_status_t status;

        pj_bzero(&ap, sizeof(ap));
        ap.user_data = ds->srtp;
        pj_sockaddr_cp(&ap.rem_addr, &ds->rem_addr);
        ap.addr_len = pj_sockaddr_get_len(&ap.rem_addr);
        if (pj_sockaddr_cmp(&info.sock_info.rtp_addr_name,
                            &info.sock_info.rtcp_addr_name) == 0)
        {
            /* Using RTP & RTCP multiplexing */
            pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_addr);
        } else if (pj_sockaddr_has_addr(&ds->rem_rtcp)) {
            pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_rtcp);
        } else {
            pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_addr);
            pj_sockaddr_set_port(&ap.rem_rtcp,
                                    pj_sockaddr_get_port(&ds->rem_addr)+1);
        }
```

amubiera:
Hello. Made some modifications and ended up with this version. The test comes through for both channels. Note the missing `idx == RTP_CHANNEL` on the outer `if`. The re-attach will happen once the last channel detects that both addresses are valid and one or both are different.

I believe the PR version has a potential for calling attach twice if I'm not mistaken. As well as attaching when the `info.src_rtcp_name` is not yet valid, which would fall through to the port+1 case.

```
/* Check remote address info, reattach member tp if changed */
if (!ds->use_ice && !ds->nego_completed[idx]) {
    pjmedia_transport_info info;
    pj_bool_t use_rtcp_mux;

    pjmedia_transport_get_info(ds->srtp->member_tp, &info);
    use_rtcp_mux = (pj_sockaddr_cmp(&info.sock_info.rtp_addr_name,
                    &info.sock_info.rtcp_addr_name))? PJ_FALSE: PJ_TRUE;

    /*     Have RTP address
     * AND Have RTCP address (when not using mux)
     * AND RTP address differs
     *  OR RTCP address differs (when not using mux)
     */
    if (pj_sockaddr_has_addr(&info.src_rtp_name) &&
        (use_rtcp_mux || pj_sockaddr_has_addr(&info.src_rtcp_name)) &&
        (pj_sockaddr_cmp(&ds->rem_addr, &info.src_rtp_name) ||
            (!use_rtcp_mux && pj_sockaddr_cmp(&ds->rem_rtcp, &info.src_rtcp_name)))) {
        pjmedia_transport_attach_param ap;
        pj_status_t status;

        /* Set RTP and RTCP address */
        pj_sockaddr_cp(&ds->rem_addr, &info.src_rtp_name);
        if (use_rtcp_mux) {
            /* Using RTP & RTCP multiplexing */
            pj_sockaddr_cp(&ds->rem_rtcp, &ds->rem_addr);
        } else if (pj_sockaddr_has_addr(&info.src_rtcp_name)) {
            pj_sockaddr_cp(&ds->rem_rtcp, &info.src_rtcp_name);
        } else {
            pj_sockaddr_cp(&ds->rem_rtcp, &ds->rem_addr);
            pj_sockaddr_set_port(&ds->rem_rtcp, pj_sockaddr_get_port(&ds->rem_addr)+1);
        }

        /* Attach member transport */
        pj_bzero(&ap, sizeof(ap));
        ap.user_data = ds->srtp;
        pj_sockaddr_cp(&ap.rem_addr, &ds->rem_addr);
        pj_sockaddr_cp(&ap.rem_rtcp, &ds->rem_rtcp);
        ap.addr_len = pj_sockaddr_get_len(&ap.rem_addr);
        status = pjmedia_transport_attach2(&ds->srtp->base, &ap);
        if (status != PJ_SUCCESS) {
            DTLS_UNLOCK(ds);
            return status;
        }
```

sauwming:
Thanks for the proposed suggested fix.
Unfortunately, it doesn't work on my local machine. I will conduct further test tomorrow to figure out why.

sauwming:
I end up using a combination of your proposed first and second patch.

The second approach when we wait until both RTP and RTCP addresses are obtained may not always work, for example against an older version of PJSIP that doesn't even initiate RTCP negotiation.

Also fixed a deadlock problem I encounter as well during testing.

Please test it on your side and let us know if there's any issue.

amubiera:
Thanks for the update. I'll try the new version and let you know.

> The second approach when we wait until both RTP and RTCP addresses are obtained may not always work, for example against an older version of PJSIP that doesn't even initiate RTCP negotiation.

I wondered about that. Wouldn't an older PJSIP always be in multiplex mode sending all RTP to the same port, and cause `info.sock_info.rtp_addr_name` and `info.sock_info.rtcp_addr_name` to be identical when connected to a newer PJSIP endpoint?

> Calling pjmedia_clock_stop() can be done while ioqueue is holding the group lock.

I saw your update to the description. I ran into that clock deadlock as well but it was hard to reproduce. That's a good catch.

One minor observation. Can the calls to `pj_sockaddr_print` in various places use the 3 flag instead of 2 to avoid the extra calls to `pj_sockaddr_get_port`?

sauwming:
> I wondered about that. Wouldn't an older PJSIP always be in multiplex mode sending all RTP to the same port, and cause `info.sock_info.rtp_addr_name` and `info.sock_info.rtcp_addr_name` to be identical when connected to a newer PJSIP endpoint?

PJSIP won't automatically use rtcp mux unless explicitly configured to. For older PJSIP without multiplexing, DTLS still works but without RTCP negotiation.

> I saw your update to the description. I ran into that clock deadlock as well but it was hard to reproduce. That's a good catch. Is the while-sleep loop needed since the clock will continue retrying every 4Hz ? Perhaps try lock + return is enough?

Without the loop, if the try lock always coincides with another thread holding DTLS_LOCK, then it's possible the clock callback is actually never executed at all.

> One minor observation. Can the calls to `pj_sockaddr_print` in various places use the 3 flag instead of 2 to avoid the extra calls to `pj_sockaddr_get_port`?

Ok, changed.

amubiera:
Thanks for the explanation.

So far so good with this version. Doing a bit more testing here.

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1676,12 +1674,10 @@ static pj_status_t dtls_encode_sdp( pjmedia_transport *tp,
...
             char addr2[PJ_INET6_ADDRSTRLEN];
-            PJ_LOG(2,(ds->base.name, "Attached transport, remote addr=%s:%d "
-                                     "remote rtcp=%s:%d",
-                      pj_sockaddr_print(&ap.rem_addr, addr, sizeof(addr), 2),
-                      pj_sockaddr_get_port(&ap.rem_addr),
-                      pj_sockaddr_print(&ap.rem_rtcp, addr2, sizeof(addr2), 2),
-                      pj_sockaddr_get_port(&ap.rem_rtcp)));
+            PJ_LOG(2,(ds->base.name, "Attached transport, remote addr=%s "
+                                     "remote rtcp=%s",
+                      pj_sockaddr_print(&ap.rem_addr, addr2, sizeof(addr), 3),
```
amubiera, line 1679: Should be:
```
pj_sockaddr_print(&ap.rem_addr, addr, sizeof(addr), 3),
pj_sockaddr_print(&ap.rem_rtcp, addr2, sizeof(addr2), 3)));
```

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1863,13 +1859,11 @@ static pj_status_t dtls_media_start( pjmedia_transport *tp,
...
-                          "remote addr=%s:%d remote rtcp=%s:%d",
+                          "remote addr=%s remote rtcp=%s",
                           pj_sockaddr_print(&ap.rem_addr, addr,
-                          sizeof(addr), 2),
-                          pj_sockaddr_get_port(&ap.rem_addr),
+                          sizeof(addr), 3),
                           pj_sockaddr_print(&ap.rem_rtcp, addr2,
-                          sizeof(addr), 2),
-                          pj_sockaddr_get_port(&ap.rem_rtcp)));
+                          sizeof(addr), 3)));
```
amubiera, line 1866: Should be: `sizeof(addr2)`

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -2028,9 +2022,8 @@ PJ_DEF(pj_status_t) pjmedia_transport_srtp_dtls_start_nego(
 #if DTLS_DEBUG
     {
         char addr[PJ_INET6_ADDRSTRLEN];
-        PJ_LOG(2,(ds->base.name, "Attached transport, remote addr=%s:%d",
-                  pj_sockaddr_print(&ap.rem_addr, addr, sizeof(addr), 2),
-                  pj_sockaddr_get_port(&ap.rem_addr)));
+        PJ_LOG(2,(ds->base.name, "Attached transport, remote addr=%s",
```
amubiera, line 2025: Perhaps also add rtcp address like the others.

pjmedia/src/pjmedia/transport_srtp_dtls.c
sauwming: the usage of addr and addr2 is intentionally reversed here to avoid exceeding 80 columns.

nanangizz (approved):

PR #3733, opened 2023-10-09
Fixed bug in SRTP ROC checking when using ICE

sauwming:
Fixed bug in SRTP ROC checking when remote specifies ICE ufrag in the session descriptor, instead of within the media.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -3400,8 +3400,9 @@ static pj_bool_t is_ice_running(pjmedia_transport *tp)
 static void check_srtp_roc(pjsua_call *call,
                            unsigned med_idx,
                            const pjsua_stream_info *new_si_,
-                           const pjmedia_sdp_media *local_sdp,
-                           const pjmedia_sdp_media *remote_sdp)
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3400,8 +3400,9 @@ static pj_bool_t is_ice_running(pjmedia_transport *tp)
 static void check_srtp_roc(pjsua_call *call,
                            unsigned med_idx,
                            const pjsua_stream_info *new_si_,
-                           const pjmedia_sdp_media *local_sdp,
-                           const pjmedia_sdp_media *remote_sdp)
+                           const pjmedia_sdp_media *local_m,
+                           const pjmedia_sdp_session *rem_sdp,
+                           const pjmedia_sdp_media *rem_m)
```
nanangizz, line 3405: minor:
- as there are `rem_sdp` & `med_idx` params already, perhaps `rem_m` is not necessary?
- also, what if the same mechanism (check media level first, then check session level) is also applied for the local SDP, for example to handle the case when the SDP is generated/modified by the app?

nanangizz (approved):

trengginas (approved):

Issue #2141, opened 2018-08-28
Add TCP initial receive timeout for server connection


pjsipbot:
_2018-08-28 06:36:34: @trengginas_ _created the issue on trac ticket [2141](https://trac.pjsip.org/repos/ticket/2141)_

This ticket will add a timeout (compile time setting `PJSIP_TCP_INITIAL_TIMEOUT`) to disconnect a TCP server connection if it doesn't receive any data following a successful connect.

Note that when a TCP server connection is idle or not referred anymore, idle timeout will disconnect the connection, refer to `PJSIP_TRANSPORT_SERVER_IDLE_TIME`

Thanks to Peter Koletzki for the suggestion.
___
_2018-08-28 06:42:30: @trengginas_ _changed **status** from **new** to **closed**_

___
_2018-08-28 06:42:30: @trengginas_ _set **resolution** to **fixed**_

___
_2018-08-28 06:42:30: @trengginas_ _commented_

In [r5869](../commit/2c3e39b2f519b38d422e060ee882bd4f3830308c):
Close #2141: Add TCP initial receive timeout for server connection.

___
_2018-08-28 08:05:46: @trengginas_ _commented_

In [r5870](../commit/0d6e7d6037897ce9969ff3d02b343ec3af2ec6a9):
Re #2141: Fix variable declaration misplacement.

PR #3744, opened 2023-10-20
Add TLS initial receive timeout for server connection

trengginas:
For TCP there is a setting to close the incoming connection if there's no data received after a successful accept (`PJSIP_TCP_INITIAL_TIMEOUT`).

There is `PJSIP_TRANSPORT_SERVER_IDLE_TIME` which will close the connection when it's idle or not referred anymore.
However a different/shorter timeout (after connection is established) is useful to protect against any DDOS attempt trying to exhaust resources.

### DIFF
--- a/pjsip/include/pjsip/sip_config.h
+++ b/pjsip/include/pjsip/sip_config.h
@@ -677,16 +677,17 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 
 
 /**
- * Initial timeout interval to be applied to incoming transports (i.e. server
- * side) when no data received after a successful connection for TLS. 
- * Value is in seconds. Disable the timeout by setting it to 0.
+ * Initial timeout interval to be applied to incoming TCP/TLS transports
+ * (i.e. server side) when no valid data received after a successful
+ * connection. Value is in seconds. Disable the timeout by setting it to 0.
  *
- * For TCP, refer to a\ PJSIP_TCP_INITIAL_TIMEOUT
- * 
  * Note that even when this is disable, the connection might still get closed
  * when it is idle or not referred anymore. Have a look at \a
- * PJSIP_TRANSPORT_SERVER_IDLE_TIME
+ * PJSIP_TRANSPORT_SERVER_IDLE_TIME.
  * 
+ * Notes:
+ * - keep-alive packet is not considered as a valid message.
+ *
  * Default: 0
 */
 #ifndef PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST
@@ -803,14 +804,19 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 
 
 /**
- * Initial timeout interval to be applied to incoming transports (i.e. server
- * side) when no data received after a successful connection. Value is in
- * seconds. Disable the timeout by setting it to 0.
+ * Initial timeout interval to be applied to incoming TCP transports
+ * (i.e. server side) when no valid data received after a successful
+ * connection. Value is in seconds. Disable the timeout by setting it to 0.
  *
  * Note that even when this is disable, the connection might still get closed
  * when it is idle or not referred anymore. Have a look at \a
- * PJSIP_TRANSPORT_SERVER_IDLE_TIME
+ * PJSIP_TRANSPORT_SERVER_IDLE_TIME.
  *
+ * Notes:
+ * - keep-alive packet is not considered as a valid message.
+ * - This macro is exclusive to TCP usage. It will be prioritize over
+ *   a\ PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST if both are set.
+ * 
  * Default: 0 (disabled)
  */
 #ifndef PJSIP_TCP_INITIAL_TIMEOUT
--- a/pjsip/include/pjsip/sip_config.h
+++ b/pjsip/include/pjsip/sip_config.h
@@ -677,6 +677,23 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 
 
 /**
+ * Initial timeout interval to be applied to incoming transports (i.e. server
+ * side) when no data received after a successful connection for TLS. 
+ * Value is in seconds. Disable the timeout by setting it to 0.
+ *
+ * For TCP, refer to a\ PJSIP_TCP_INITIAL_TIMEOUT
+ * 
+ * Note that even when this is disable, the connection might still get closed
+ * when it is idle or not referred anymore. Have a look at \a
+ * PJSIP_TRANSPORT_SERVER_IDLE_TIME
+ * 
+ * Default: 0
+*/
+#ifndef PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST
+#   define PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST     0
+#endif
+
+/**
  * Maximum number of usages for a transport before a new transport is
  * created. This only applies for ephemeral transports such as TCP.
  *
--- a/pjsip/include/pjsip/sip_config.h
+++ b/pjsip/include/pjsip/sip_config.h
@@ -677,11 +677,12 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 
 
 /**
- * Initial timeout interval to be applied to incoming TCP/TLS transports
- * (i.e. server side) when no valid data received after a successful
- * connection. Value is in seconds. Disable the timeout by setting it to 0.
+ * The initial timeout interval for incoming TCP/TLS transports
+ * (i.e. server side) in the event that no valid SIP message is received
+ * following a successful connection. The value is in seconds.
+ * Disable the timeout by setting it to 0.
  *
- * Note that even when this is disable, the connection might still get closed
+ * Note that even if this is disabled, the connection might still get closed
  * when it is idle or not referred anymore. Have a look at \a
  * PJSIP_TRANSPORT_SERVER_IDLE_TIME.
  * 
@@ -804,18 +805,19 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 
 
 /**
- * Initial timeout interval to be applied to incoming TCP transports
- * (i.e. server side) when no valid data received after a successful
- * connection. Value is in seconds. Disable the timeout by setting it to 0.
+ * The initial timeout interval for incoming TCP transports
+ * (i.e. server side) in the event that no valid SIP message is received
+ * following a successful connection. The value is in seconds.
+ * Disable the timeout by setting it to 0.
  *
- * Note that even when this is disable, the connection might still get closed
+ * Note that even if this is disabled, the connection might still get closed
  * when it is idle or not referred anymore. Have a look at \a
  * PJSIP_TRANSPORT_SERVER_IDLE_TIME.
  *
  * Notes:
  * - keep-alive packet is not considered as a valid message.
- * - This macro is exclusive to TCP usage. It will be prioritize over
- *   a\ PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST if both are set.
+ * - This macro is specific to TCP usage and takes precedence over
+ *   a\ PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST when both are set.
  * 
  * Default: 0 (disabled)
  */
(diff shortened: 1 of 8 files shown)


nanangizz:
The specs need to be clarified, e.g:
- should it close the connection when no incoming message or no **valid** incoming SIP message?
- if a transport is to be destroyed due to first-timeout and there is some message in the buffer (incomplete/invalid SIP message), should we invoke `tp_drop_data_cb()` first before destroying the transport?
- as there is already `PJSIP_TCP_INITIAL_TIMEOUT`, instead of using `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST` for TLS only, here are IMO a couple alternatives:
  - for TLS, consider to name it `PJSIP_TLS_INITIAL_TIMEOUT`
  - for both TCP & TLS, use `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST` (and then perhaps deprecate `PJSIP_TCP_INITIAL_TIMEOUT` and/or override `PJSIP_TCP_INITIAL_TIMEOUT` when `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST` is set).

trengginas:
* For TCP initial timer, the connection will be closed when no incoming message. For TLS the connection is closed when no
  valid message is received.
* agreed that `tp_drop_data_cb()` should be called
* agree to use `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST` for both TCP and TLS. With some implications:
  - Change the spec for `PJSIP_TCP_INITIAL_TIMEOUT`, to close the connection when there's no **valid** message is received.
  - Remove the use of `initial_timer` in TCP and instead use `idle_timer`. Both TCP and TLS will schedule `idle_timer` in `on_accept_complete`. It will simplify the timer usage on the transport.
  - Maintain the `PJSIP_TLS_INITIAL_TIMEOUT` by setting it as a default value for `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST` for TCP.

nanangizz:
> * For TCP initial timer, the connection will be closed when no incoming message. For TLS the connection is closed when no
>       valid message is received.

I assume this is the specs (to be implemented), why are TCP & TLS have different behavior here?

>       * Change the spec for `PJSIP_TCP_INITIAL_TIMEOUT`, to close the connection when there's no **valid** message is received.

Right, behavioral change, I think it is okay/acceptible (still need to be mentioned somewhere).
Note that there is keep alive message (e.g: newline & configurable), perhaps we need to also specify whether keep-alive is acceptable as valid (initial) message.

>       * Remove the use of `initial_timer` in TCP and instead use `idle_timer`. Both TCP and TLS will schedule `idle_timer` in `on_accept_complete`. It will simplify the timer usage on the transport.

Or perhaps the timer is started right after TCP/TLS client transport is registered (in `pjsip_transport_register()`).

PeterKoletzki:
let me try to explain my point of view:

1. TLS entered handshake but there are no more data to perform the handshake -> timeout error
2. no data (TLS after handshake success,  TCP) -> timeout error 
3. TLS/TCP receiving some garbage data e.g. "asdfghjk" -> timeout error, malformed packet error
4. TLS/TCP received a lots of garbage until the buffer limit is reached -> timeout error, malformed packet error 
2+3+4 shall only checked until a valid SIP message has been received. After that the normal server idle timer is taken as usual.

trengginas:
The implementation to support `PJSIP_TCP_INITIAL_TIMEOUT` is part of #2141.

It uses `PJSIP_TCP_INITIAL_TIMEOUT` as a default value for runtime TCP configuration `pjsip_tcp_transport_cfg::initial_timeout`
Later it will be use as a delay value for `initial_timer`.

The proposed changes:
1. `pjsip_tcp_transport_cfg::initial_timeout` is a public API, we need to maintain it.
   - Add new configuration `initial_timeout` for TLS.
2. Both configuration will use `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST` as a default `initial_timeout` value.
   For TCP, use `PJSIP_TCP_INITIAL_TIMEOUT` whenever it is set.
3. Remove `initial_timer` in TCP and use `idle_timer` for both TCP/TLS.
4. Store the runtime configuration for the `initial_timeout` in a new field in `pjsip_transport`.
5. Change `idle_timer` timeout value:
   - When no data is received, use the `initial_timeout` value stored. 
   - After data is received, use `PJSIP_TRANSPORT_SERVER_IDLE_TIME` compile time settings
6. Scheduling `idle_timer`:
   * TLS: 
     * the `idle_timer` will start `on_accept_complete()` (already implemented). 
   * TCP: 
     * schedule the `idle_timer` by calling `pjsip_transport_add_ref()/pjsip_transport_dec_ref()` after the transport is registered.
7. Call `tp_drop_data_cb()` on the timer callback if there's some invalid data received after the timer started.
8. Add docs related to behavioral change on `PJSIP_TCP_INITIAL_TIMEOUT` and keep-alive packet is not considered as valid/initial message.

nanangizz:
> let me try to explain my point of view:
> 
>     1. TLS entered handshake but there are no more data to perform the handshake -> timeout error

So for TLS, the timer is started right after the TCP accept and before the TLS handshake, makes sense.

> 
>     2. no data (TLS after handshake success,  TCP) -> timeout error
> 
>     3. TLS/TCP receiving some garbage data e.g. "asdfghjk" -> timeout error, malformed packet error

Right, I believe both are the initial goals of this PR.

> 
>     4. TLS/TCP received a lots of garbage until the buffer limit is reached -> timeout error, malformed packet error

For the scenario of buffer limit reached, perhaps we can immediately stop the transport as malformed packet error without waiting for timeout?

>        2+3+4 shall only checked until a valid SIP message has been received. After that the normal server idle timer is taken as usual.

Yes.

nanangizz:
> The implementation to support `PJSIP_TCP_INITIAL_TIMEOUT` is part of #2141.
> 
> It uses `PJSIP_TCP_INITIAL_TIMEOUT` as a default value for runtime TCP configuration `pjsip_tcp_transport_cfg::initial_timeout` Later it will be use as a delay value for `initial_timer`.
> 
> The proposed changes:
> 
>     1. `pjsip_tcp_transport_cfg::initial_timeout` is a public API, we need to maintain it.
>        
>        * Add new configuration `initial_timeout` for TLS.
> 
>     2. Both configuration will use `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST` as a default `initial_timeout` value.
>        For TCP, use `PJSIP_TCP_INITIAL_TIMEOUT` whenever it is set.
> 
>     3. Remove `initial_timer` in TCP and use `idle_timer` for both TCP/TLS.
> 
>     4. Store the runtime configuration for the `initial_timeout` in a new field in `pjsip_transport`.
> 
>     5. Change `idle_timer` timeout value:
>        
>        * When no data is received, use the `initial_timeout` value stored.
>        * After data is received, use `PJSIP_TRANSPORT_SERVER_IDLE_TIME` compile time settings
> 
>     6. Scheduling `idle_timer`:
>        
>        * TLS:
>          
>          * the `idle_timer` will start `on_accept_complete()` (already implemented).
>        * TCP:
>          
>          * schedule the `idle_timer` by calling `pjsip_transport_add_ref()/pjsip_transport_dec_ref()` after the transport is registered.
> 
>     7. Call `tp_drop_data_cb()` on the timer callback if there's some invalid data received after the timer started.
> 
>     8. Add docs related to behavioral change on `PJSIP_TCP_INITIAL_TIMEOUT` and keep-alive packet is not considered as valid/initial message.

Looks okay to me.

Please also incorporate a couple of new points/scenarios from @PeterKoletzki (start timer right after TCP-accept before TLS-handshake, and buffer limit reached).

PeterKoletzki:
Regarding scenario 4
right, we should immediately stop the transport.

trengginas:
> > let me try to explain my point of view:
> > ```
> > 1. TLS entered handshake but there are no more data to perform the handshake -> timeout error
> > ```
> 
> So for TLS, the timer is started right after the TCP accept and before the TLS handshake, makes sense.
> 

There is already a handshake timer which handles this. The timeout setting is from `pj_ssl_sock_param::timeout`. 
The timer will notify app by calling `pjsip_tls_setting::on_accept_fail_cb` if such scenario occured.

PeterKoletzki:
> > > let me try to explain my point of view:
> > > ```
> > > 1. TLS entered handshake but there are no more data to perform the handshake -> timeout error
> > > ```
> > 
> > 
> > So for TLS, the timer is started right after the TCP accept and before the TLS handshake, makes sense.
> 
> There is already a handshake timer which handles this. The timeout setting is from `pj_ssl_sock_param::timeout`. The timer will notify app by calling `pjsip_tls_setting::on_accept_fail_cb` if such scenario occured.

Right, worked as expected. 
I put that in the scenarios cases to have all cases which can happen in one view. 
Regardless if that case is handled or not.

pjsip/src/pjsip/sip_transport.c
```diff
@@ -2048,6 +2077,15 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
...
                         (*mgr->tp_drop_data_cb)(&dd);
                     }
+
+                    if (rdata->tp_info.transport->idle_timer.id == 
+                                                         INITIAL_IDLE_TIMER_ID)
+                    {
+                        /* We are not getting the first valid data as expected,
+                         * close the transport.
+                         */
+                        pjsip_transport_shutdown(rdata->tp_info.transport);
```
nanangizz, line 2087: better print log?

pjsip/src/pjsip/sip_transport.c
```diff
@@ -1179,16 +1200,24 @@ PJ_DEF(pj_status_t) pjsip_transport_dec_ref( pjsip_transport *tp )
...
-                delay.sec = (tp->dir==PJSIP_TP_DIR_OUTGOING) ?
-                                PJSIP_TRANSPORT_IDLE_TIME :
-                                PJSIP_TRANSPORT_SERVER_IDLE_TIME;
+                if (tp->dir == PJSIP_TP_DIR_OUTGOING) {
+                    delay.sec = PJSIP_TRANSPORT_IDLE_TIME;
+                } else {
+                    delay.sec = PJSIP_TRANSPORT_SERVER_IDLE_TIME;
+                    if (tp->last_recv_ts.u64 == 0 && tp->initial_timeout) {
+                        timer_id = INITIAL_IDLE_TIMER_ID;
+                        delay.sec = tp->initial_timeout;
```
nanangizz, line 1218: print log that initial timer has started so we know it is enabled?

pjsip/src/pjsip/sip_transport.c
```diff
@@ -1079,8 +1087,21 @@ static void transport_idle_callback(pj_timer_heap_t *timer_heap,
...
+            if (tp->last_recv_len > 0 && tp->tpmgr->tp_drop_data_cb) {
+                pjsip_tp_dropped_data dd;
+                pj_bzero(&dd, sizeof(dd));
+                dd.tp = tp;
+                dd.data = NULL;
+                dd.len = tp->last_recv_len;
+                dd.status = PJ_ESOCKETSTOP;
+                (*tp->tpmgr->tp_drop_data_cb)(&dd);
+            }
+        }
```
nanangizz, line 1104: add log that transport is about to be destroyed due to idle, for both idle timers: initial and consecutive?

pjsip/src/pjsip/sip_transport.c
```diff
@@ -2211,6 +2249,13 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
         }
         */
 
+        /* We have a valid message, cancel the initial timer. */
+        if (rdata->tp_info.transport->idle_timer.id == INITIAL_IDLE_TIMER_ID) {
+            rdata->tp_info.transport->idle_timer.id = PJ_FALSE;
+            pjsip_endpt_cancel_timer(mgr->endpt,
+                                     &rdata->tp_info.transport->idle_timer);
```
nanangizz, line 2256: print log?

nanangizz (approved):
Just minors about printing logs for debugging. For transport shutdown/destroy scenarios, better use lower log verbosity level, e.g: 2?

pjsip/include/pjsip/sip_config.h
```diff
@@ -676,6 +676,24 @@ PJ_INLINE(pjsip_cfg_t*) pjsip_cfg(void)
 #endif
 
 
+/**
+ * Initial timeout interval to be applied to incoming TCP/TLS transports
+ * (i.e. server side) when no valid data received after a successful
```
sauwming, line 681: Minor: doc clarification
The word "data" and "message" seem to be used interchangeable throughout the document, which may cause confusion.

According to the RFC 3261, "SIP messages are either requests or responses", while data may not always refer to an entire message. So better clarify and be precise here.

I would also suggest to use word processor or AI to catch grammatical or spelling errors.

sauwming (approved):

PR #3748, opened 2023-10-25
base64: support encode/decode with URL and Filename Safe Alphabet

jimying:
as title

[rfc4648 section-5, Base 64 Encoding with URL and Filename Safe Alphabet](https://datatracker.ietf.org/doc/html/rfc4648#section-5)

### DIFF
--- a/pjlib-util/include/pjlib-util/base64.h
+++ b/pjlib-util/include/pjlib-util/base64.h
@@ -93,6 +93,18 @@ PJ_DECL(pj_status_t) pj_base64url_encode(const pj_uint8_t *input, int in_len,
 PJ_DECL(pj_status_t) pj_base64_decode(const pj_str_t *input, 
                                       pj_uint8_t *out, int *out_len);
 
+/**
+ * Decode base64 (URL and Filename Safe) string.
+ *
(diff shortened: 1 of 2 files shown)


sauwming:
Need to postpone this, currently in the process of releasing 2.14.

nanangizz:
I'll vote for the second solution because [the RFC](https://datatracker.ietf.org/doc/html/rfc4648#section-5) explicitly mentions that they should be treated as different encodings:
```
   This encoding may be referred to as "base64url".  This encoding
   should not be regarded as the same as the "base64" encoding and
   should not be referred to as only "base64".
```

sauwming:
For the second approach, do we need to add a helper function such as `pj_base64_isbase64url_encoding()` or something like that?

nanangizz:
CMIIW I don't see how they can be ambiguous/mixed-use as the use case are different (if the encoded string is to be used-as or is fetched-from URL, use `base64url`, otherwise use `base64`). So I don't think such helper function is needed.

sauwming (approved):

trengginas (approved):

nanangizz (approved):

PR #3751, opened 2023-10-31
Add SIP transport data received callback

sauwming:
Initial patch provided by Peter Koletzki in the first commit:
Add callback `pjsip_tp_on_custom_parser_cb()` and API `pjsip_tpmgr_set_custom_parser_cb()` for custom parsing.

Proposed modification in the subsequent commits:
* Change the cb and API names to `pjsip_tp_on_rx_data_cb()` and `pjsip_tpmgr_set_recv_data_cb()`, in order to:
  - make it more general purpose. The callback is not specifically for parsing, but can also be used for other purposes such as filtering, logging of potential malicious activities, or other processing
  - make the name more similar to the existing `pjsip_tp_on_rx_dropped_cb()` and `pjsip_tpmgr_set_drop_data_cb()`.
* Also put the callback parameters into a struct `pjsip_tp_rx_data`, for extensibility purpose.

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -142,9 +142,9 @@ struct pjsip_tpmgr
 #endif
     void           (*on_rx_msg)(pjsip_endpoint*, pj_status_t, pjsip_rx_data*);
     pj_status_t    (*on_tx_msg)(pjsip_endpoint*, pjsip_tx_data*);
-    pjsip_tp_state_callback tp_state_cb;
+    pjsip_tp_state_callback   tp_state_cb;
     pjsip_tp_on_rx_dropped_cb tp_drop_data_cb;
-    pjsip_tp_on_custom_parser_cb tp_custom_parser_cb;
+    pjsip_tp_on_rx_data_cb    tp_rx_data_cb;
 
     /* Transmit data list, for transmit data cleanup when transport manager
      * is destroyed.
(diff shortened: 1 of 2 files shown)


trengginas (approved):

pjsip/src/pjsip/sip_transport.c
```diff
@@ -2066,6 +2067,26 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
...
+                rd.status = PJ_EIGNORED;
+
+                if ((*mgr->tp_rx_data_cb)(&rd) == PJ_SUCCESS) {
+                    /* Sanity check */
+                    if (rd.len > remaining_len)
+                        rd.len = remaining_len;
+                    total_processed += rd.len;
+                    current_pkt += rd.len;
+                    remaining_len -= rd.len;
+                    continue;
```
nanangizz, line 2086: IMHO, perhaps custom processings here is a bit too much, some default nice features such as initial idle timer, parsing error logging, drop data callback, will be gone (need to be re-implemented by app if wanted). Moreover, if app is allowed to do custom rdata processings by itself, I think we need to describe more things in the docs, e.g:
- app may need to invoke `pjsip_endpt_process_rx_data()` since `mgr->on_rx_msg/endpt_on_rx_msg()` is no longer called.
- describe all processings done by the lib, so the developer will be aware, what they do or are for, which processings are mandatory/optional/modifiable.

Perhaps allowing app to verify the data or shift the start of the data (e.g: for skipping garbage) is sufficient for now?

nanangizz (approved):

nanangizz:
One more thing, should we add to the docs about how app can shutdown the transport (for example after if find out that the data is suspicious/garbage), e.g: should app set `rd.len` to zero and shutdown the transport from the callback?

pjsip/src/pjsip/sip_transport.c
trengginas: Could you clarify:
```
      remaining_len = rd.len;

      msg_fragment_size = remaining_len - rd.len;
      total_processed += msg_fragment_size;
      current_pkt += msg_fragment_size;
```
this means `msg_fragment_size` is 0, and `current_pkt` will not be increased.

sauwming: > this means `msg_fragment_size` is 0, and `current_pkt` will not be increased.

Good catch. I incorrectly moved it to the top of the block.

PR #3752, opened 2023-10-31
Fixed race condition in ACK handling of INVITE message

sauwming:
Currently we handle the ACK of INVITE in two places:
- `sip_inv.c`, with the flow:
```
pjsip_endpt_process_rx_data()
mod_ua_on_rx_request()
pjsip_dlg_on_rx_request()->acquire dialog lock
mod_inv_on_rx_request()
pjsip_tsx_terminate()->acquire tsx lock
```
- `sip_tsx.c`, with the flow:
```
pjsip_endpt_process_rx_data()
mod_tsx_layer_on_rx_request()
pjsip_tsx_recv_msg()->acquire tsx lock
tsx_on_state_completed_uas()
tsx_set_state()-mod_ua_on_tsx_state()-pjsip_dlg_on_tsx_state()-acquire dlg lock
```

According to RFC 3261 section 17 (https://datatracker.ietf.org/doc/html/rfc3261#section-17):
"In the case of a transaction where the
   request was an INVITE (known as an INVITE transaction), the
   transaction also includes the ACK only if the final response was not
   a 2xx response.  If the response was a 2xx, the ACK is not considered
   part of the transaction."

But currently we do not enforce this (I assume the reason is for leniency purpose) and allow either component (sip inv or tsx) to handle the ACK, i.e. if the ACK can be matched to INVITE tsx, tsx will handle it, otherwise the dialog will handle it. Unfortunately, if we receive two ACKs, one belonging to INVITE tsx, and one not, this can potentially introduce deadlock as shown above.

The issue is reproducible in a multithreaded environment with the following SIPP scenario:
[uac-double_ack.txt](https://github.com/pjsip/pjproject/files/13213632/uac-double_ack.txt)

Note that the problem is complicated since there doesn't seem any way to ensure the lock ordering between the dialog and tsx. In the first flow, dialog can't acquire tsx lock because the received ACK is not associated with any tsx.

So, the proposed fix is to sacrifice the leniency, i.e. by only allowing one component to handle it as per the RFC. If the INVITE status is 2xx, sip tsx must not handle it and pass it to dialog, while if status is not 2xx, sip dialog must not handle it.

Because of this, there may be potential backward incompatibility consequences.

### DIFF
--- a/pjsip/include/pjsip/sip_transaction.h
+++ b/pjsip/include/pjsip/sip_transaction.h
@@ -530,6 +530,18 @@ PJ_DECL(void) pjsip_tsx_layer_dump(pj_bool_t detail);
 PJ_DECL(const char *) pjsip_tsx_state_str(pjsip_tsx_state_e state);
 
 /**
+ * Schedule tsx timer.
+ * @param tsx       The transaction
+ * @param entry     The timer entry
+ * @param delay     The timer delay
+ * @param active_id The timer id
+ */
+PJ_DECL(pj_status_t) pjsip_tsx_schedule_timer(pjsip_transaction *tsx,
+                                              pj_timer_entry *entry,
+                                              const pj_time_val *delay,
+                                              int active_id);
+
+/**
  * Get the role name.
  * @param role  Role.
  */
--- a/pjsip/include/pjsip/sip_transaction.h
+++ b/pjsip/include/pjsip/sip_transaction.h
@@ -410,9 +410,24 @@ PJ_DECL(pj_status_t) pjsip_tsx_create_key( pj_pool_t *pool,
  *
  * @param tsx       The transaction.
  * @param code      The status code to report.
+ *
+ * @return          PJ_SUCCESS or the appropriate error code.
  */
 PJ_DECL(pj_status_t) pjsip_tsx_terminate( pjsip_transaction *tsx,
                                           int code );
+
+
+/**
+ * Force terminate transaction asynchronously, using the transaction
+ * internal timer.
+ *
+ * @param tsx       The transaction.
+ * @param code      The status code to report.
+ *
+ * @return          PJ_SUCCESS or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsip_tsx_terminate_async(pjsip_transaction *tsx,
+                                               int code );
 
 
 /**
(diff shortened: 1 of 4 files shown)


pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -698,13 +698,20 @@ static pj_bool_t mod_inv_on_rx_request(pjsip_rx_data *rdata)
...
-                                inv->invite_tsx->status_code);
+
+            /* In the case of an INVITE transaction, if the response was not
+             * 2xx, the ACK is considered part of the transaction, so
+             * should be handled by the transaction.
+             */
+            if (inv->invite_tsx->status_code/100 == 2) {
+                pjsip_tsx_terminate(inv->invite_tsx,
+                                    inv->invite_tsx->status_code);
+            }
```
nanangizz, line 709: If the tsx status code is 3xx and the remote wrongly put a bad Via branch in its ACK (i.e: not equal to the branch in its INVITE) so the ACK has to be handled here (by the invite session), but by this patch, the INVITE tsx will be no longer terminated immediately (will be terminated later by tsx timeout timer?).

Not sure if it is a problem (note that there should be only one INVITE tsx at any time), but perhaps it is better if it can be avoided (maintaining the current behavior). So an idea, what if the `pjsip_tsx_terminate()` is moved to a timer instead (and use the tsx group lock for the timer scheduling to avoid tsx premature destroy)?

pjsip/include/pjsip/sip_transaction.h
```diff
@@ -529,6 +529,18 @@ PJ_DECL(void) pjsip_tsx_layer_dump(pj_bool_t detail);
...
+ * Schedule tsx timer.
+ * @param tsx       The transaction
+ * @param entry     The timer entry
+ * @param delay     The timer delay
+ * @param active_id The timer id
+ */
+PJ_DECL(pj_status_t) pjsip_tsx_schedule_timer(pjsip_transaction *tsx,
+                                              pj_timer_entry *entry,
+                                              const pj_time_val *delay,
+                                              int active_id);
```
nanangizz, line 542: "Valid" values for `active_id` are actually defined internally, so perhaps a specific function is cleaner than a generic one (like this one), e.g: `pjsip_tsx_terminate2()/pjsip_tsx_terminate_async()`?

pjsip/src/pjsip/sip_transaction.c
```diff
@@ -237,6 +244,19 @@ PJ_DEF(const char *) pjsip_tsx_state_str(pjsip_tsx_state_e state)
...
+ */
+PJ_DEF(pj_status_t) pjsip_tsx_schedule_timer( pjsip_transaction *tsx,
+                                              pj_timer_entry *entry,
+                                              const pj_time_val *delay,
+                                              int active_id)
+{
+    tsx_cancel_timer(tsx, entry);
+
+    return tsx_schedule_timer(tsx, entry, delay, active_id);
+}
```
nanangizz, line 258: Protect timer cancel/schedule with `lock/unlock_timer(tsx)`?

pjsip/src/pjsip/sip_transaction.c
```diff
@@ -139,6 +139,7 @@ static int max_retrans_count = -1;
 #define TIMEOUT_TIMER           2
 #define TRANSPORT_ERR_TIMER     3
 #define TRANSPORT_DISC_TIMER    4
+#define TERMINATE_TIMER         PJSIP_TSX_STATE_TERMINATED
```
nanangizz, line 142: Not sure about this, `PJSIP_TSX_STATE_TERMINATED` has implicitly defined value, IMO better use `5` here?

nanangizz (approved):

trengginas (approved):

PR #3753, opened 2023-11-01
Check for buffer EOF in scanner

sauwming:
In continuation of https://github.com/pjsip/pjproject/security/advisories/GHSA-fq45-m3f7-3mhj, we are striving to incorporate an EOF check within our scanner. This initiative is aimed at achieving the following objectives:
- Improving the robustness of the scanner (and reducing dependence on NULL buffer termination)
- Addressing our internal issues resulting from our own failures in conforming to the specification that mandates the use of a NULL-terminated buffer while invoking the API.

The EOF check has already been employed in similar functions such as `pj_scan_get_until()`, `pj_scan_get_until_ch()`, `pj_scan_get_until_chr()`.

Thanks to Peter Koletzki for the report and Rene Heuven for the initial patch.

### DIFF
--- a/pjlib-util/src/pjlib-util/scanner.c
+++ b/pjlib-util/src/pjlib-util/scanner.c
@@ -277,11 +277,7 @@ PJ_DEF(void) pj_scan_get( pj_scanner *scanner,
 
     do {
         ++s;
-    } while (pj_cis_match(spec, *s));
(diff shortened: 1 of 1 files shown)


nanangizz:
Should we also update `pj_scan_peek()` & `pj_scan_skip_whitespace()`?

nanangizz (approved):

trengginas (approved):

PR #3756, opened 2023-11-01
Hide private symbols

gjasny:
Those symbols are needlessly exposed and can be easily hidden.


sauwming (approved):

nanangizz (approved):
`third_party/srtp/srtp/srtp.c` seems to be part of libsrtp, so if someday we upgrade the libsrtp version, those symbols may get exposed again (because we may forget to reapply these modifications :)

PR #3757, opened 2023-11-02
sdp.c: don't pass NULL pointer to log string format

brad0:
Passing a NULL pointer to log string format will result in a crash on OpenBSD.

https://flak.tedunangst.com/post/its-hard-work-printing-nothing

### DIFF
--- a/pjmedia/src/pjmedia/sdp.c
+++ b/pjmedia/src/pjmedia/sdp.c
@@ -611,13 +611,13 @@ PJ_DEF(pj_status_t) pjmedia_sdp_rtpmap_to_attr(pj_pool_t *pool,
     len = pj_ansi_snprintf(tempbuf, sizeof(tempbuf), 
                            "%.*s %.*s/%u%s%.*s",
                            (int)rtpmap->pt.slen,
-                           rtpmap->pt.ptr,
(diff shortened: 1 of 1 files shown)


sauwming:
First of all, we're actually using `"%.*s"` here, so I'm surprised that it will cause crash if the string length is zero and the string pointer is NULL. Since if the length is zero, any reasonable implementation shouldn't even try to access the pointer, and should print nothing at all.

Now, let's say that the NULL string will indeed cause issue, the PR should be modified as follows:
-  The encoding name can't be NULL, so the check is unnecessary.
```
PJ_ASSERT_RETURN(rtpmap->enc_name.slen && rtpmap->clock_rate, PJMEDIA_SDP_EINRTPMAP);
```
- I believe this implies that pt shouldn't be NULL too, so perhaps better add it to the above assertion instead?
- For the rtp param, it's better to check the slen instead, i.e. `rtpmap->param.slen ? rtpmap->param.ptr : "(null)")` . Since it may be possible that `pj_str_t.slen` is 0, but `pj_str_t.ptr` just points to some random location instead of NULL.

nanangizz (approved):

trengginas (approved):

PR #3758, opened 2023-11-02
Switch LibreSSL code path to utilizing OpenSSL 1.1 API

brad0:
Not worrying about version checks as these API additions date back to
LibreSSL 2.7 from almost 6 years ago.


sauwming:
Perhaps some description about why this is necessary?

nanangizz:
Perhaps LibreSSL version check is also needed now?

sauwming (approved):

brad0:
I updated the commit message and description to provide more information.

nanangizz (approved):
Agree, 6 year old for a TLS/security lib should be quite dated.

Issue #3769, opened 2023-11-07
Answering a re-INVITE failed for UAC


wosrediinanatour:
### Describe the bug

Issue #3537 causes if-branch https://github.com/pjsip/pjproject/blame/ca5255795a330f88d72a0504d902de8bbddcdfeb/pjsip/src/pjsua-lib/pjsua_call.c#L2747 to be taken, for UAC, which wants to answer a re-INVITE.

### Steps to reproduce

1. Make a call (app shall handle answering of SIP re-INVITEs)
2. Transmit a re-INVITE to the UAC of the call
3. Try to answer the re-INVITE

### PJSIP version

2.14

### Context

The issue happens on Linux.

### Log, call stack, etc

```shell
-
```

PR #3770, opened 2023-11-07
Fix answering a re-INVITE

wosrediinanatour:
Commit f3a261345f448dd63b98d66c06dd5f283840e17b introduces a problem with answering incoming re-INVITE manually by the app in cases, where the UAC receives the re-INVITE.

Issue #3769


sauwming (approved):

nanangizz (commented):
ups, just realized that review I wrote long time ago is actually not submitted (still in pending state :)

sauwming (approved):

nanangizz (approved):

Issue #3771, opened 2023-11-07
Loop transport does not work with PJSUA(2)


wosrediinanatour:
### Describe the bug

Commit [6dc9b8c181aff39845f02b4626e0812820d4ef0d ](https://github.com/pjsip/pjproject/commit/6dc9b8c181aff39845f02b4626e0812820d4ef0d) of https://github.com/pjsip/pjproject/security/advisories/GHSA-f76w-fh7c-pc66 cause loop transport to fail.

### Steps to reproduce

Make a call (cf. Context). 

### PJSIP version

2.14

### Context

- Linux
- Mocked pjsua_aud ("3rd party media stack")
- Using transport loop

### Log, call stack, etc

```shell
SDP negotiation done: Success 
...
Call 0: updating media.. 
...
Audio updated, stream #0: PCMU (sendrecv)
Media stream call00:1 is destroyed 
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2193131==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000030 (pc 0x000001083a14 bp 0x000000000000 sp 0x7fae5b9e2290 T3)
==2193131==The signal is caused by a READ memory access.
==2193131==Hint: address points to the zero page.
    #0 0x1083a14 in grp_lock_dec_ref src/pj/lock.c:562:14
    #1 0x1083a14 in pj_grp_lock_dec_ref src/pj/lock.c:654:12
    #2 0x1046611 in transport_destroy src/pjmedia/transport_loop.c:254:5
    #3 0xfd7c0f in pjmedia_transport_close pjmedia/include/pjmedia/transport.h:1036:17
...
```

sauwming:
Fixed by #3773

PR #3773, opened 2023-11-07
Fix transport loop

wosrediinanatour:
Transport "loop" need set the `base.grp_lock`, as also shown in `transport_adapter_sample.c` to have destroying of the loop transport working.

Issue #3771

### DIFF
--- a/pjmedia/src/pjmedia/transport_loop.c
+++ b/pjmedia/src/pjmedia/transport_loop.c
@@ -187,6 +187,7 @@ pjmedia_transport_loop_create2(pjmedia_endpt *endpt,
     if (status != PJ_SUCCESS)
         return status;
 
+    tp->base.grp_lock = grp_lock;
     pj_grp_lock_add_ref(grp_lock);
     pj_grp_lock_add_handler(grp_lock, pool, tp, &tp_loop_on_destroy);
 


pjmedia/src/pjmedia/transport_loop.c
```diff
@@ -239,7 +240,6 @@ static void tp_loop_on_destroy(void *arg)
     struct transport_loop *loop = (struct transport_loop*) arg;
 
     PJ_LOG(4, (loop->base.name, "Loop transport destroyed"));
-    pj_pool_release(loop->pool);
```
trengginas, line 242: No need to erase this, it will cause leak.

wosrediinanatour: But then it is released twice: the second time in https://github.com/pjsip/pjproject/blob/ca5255795a330f88d72a0504d902de8bbddcdfeb/pjlib/src/pj/lock.c#L404.

trengginas: the issue is that, the pool is released when the tp is still in use. If the tp is not used any longer, then it will be safe to release the pool. Tested here without removing the pool release, and there's no crash

wosrediinanatour: But then we have two releases of pool - or do I miss something?
`pj_grp_lock_dec_ref()` -> `grp_lock_dec_ref()`-> `grp_lock_destroy()`:
*  `cb->handler()`-> `tp_loop_on_destroy()` -> `pj_pool_release()`.
*  `pj_pool_release()`.


I have continue debugging...

Anyway... due to https://github.com/pjsip/pjproject/commit/6dc9b8c181aff39845f02b4626e0812820d4ef0d I get a crash for a code that worked at least since 2.11.

wosrediinanatour: You are right  - it is working for me too.

sauwming (approved):

trengginas (approved):

nanangizz:
IIRC the group lock of transport loop is not exported to `base.grp_lock` on purpose, i.e: it does not really have socket/ioqueue key and it destroys synchronously (not like ICE which may destroy async-ly for TURN deallocation). After the stream/transport is destroyed, I assume the app does not send any packets (and as it is a loop tp, it also won't receive any packet, so there should be no race condition between receive & destroy). So if you see a crash, perhaps the problem is somewhere else. Unfortunately the issue and this PR do not seem to describe the problem detail, e.g: race condition causing use-after-free.

wosrediinanatour:
I will try to provide more information and look deeper into the problem.

wosrediinanatour:
@nanangizz: In https://github.com/pjsip/pjproject/blob/9287ac21e6dda91f68a0e30dea88e846c960a549/pjmedia/src/pjmedia/transport_loop.c#L186 the grp_lock is created, but the pointer `grp_lock` is nowhere stored in this function.

https://github.com/pjsip/pjproject/blob/9287ac21e6dda91f68a0e30dea88e846c960a549/pjmedia/src/pjmedia/transport_loop.c#L254 tries to access`tp->grp_lock` - which was never set.
E.g. the pool address is stored by `tp->pool = pool;` at https://github.com/pjsip/pjproject/blob/9287ac21e6dda91f68a0e30dea88e846c960a549/pjmedia/src/pjmedia/transport_loop.c#L180.

nanangizz (approved):
@wosrediinanatour You're absolutely right! Thanks for pointing it out.

PR #3784, opened 2023-11-21
add PJSUA_DETECT_MERGED_REQUEST parameter

igorolhovskiy:
In my case I'm using pjsua as a test tool, and I need to accept all type of calls including same transaction calls, but different branches. As Merged Requests mechanism is relies on From tag,   Call-ID, and CSeq, different branches are detected as Merged request.
This option adds a possibility to control it in compile-time.


sauwming (approved):

sauwming:
Looks okay to me. I just made a minor doc update.

nanangizz (approved):

PR #3788, opened 2023-11-24
Revert "Fix warnings"

### LINKED ISSUE #3785: Cross compile doesn't work
### Describe the bug

On Ubuntu 22.04 cross compilation for arm has many errors of long int vs int.

### Steps to reproduce

$ ./configure --host=arm-linux-gnueabi --disable-sdl --disable-ffmpeg --disable-v4l2 --disable-openh264 --disable-libwebrtc --disable-libyuv --disable-floating-point --disable-pjsua2
$ make dep
$ make lib`

### PJSIP version

2.14

### Context

$ uname -a
Linux prchalpc 6.2.0-36-generic #37\~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Oct  9 15:34:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy
$ arm-linux-gnueabi-gcc --version
arm-linux-gnueabi-gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

### Log, call stack, etc

```shell
In file included from ../src/pj/pool_caching.c:21:
../src/pj/pool_caching.c: In function ‘cpool_dump_status’:
../src/pj/pool_caching.c:282:27: error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 3 has type ‘pj_size_t’ {aka ‘unsigned int’} [-Werror=format=]
  282 |     PJ_LOG(3,("cachpool", "   Capacity=%lu, max_capacity=%lu, used_cnt=%lu", \
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  283 |                              cp->capacity, cp->max_capacity, cp->used_count));
      |                              ~~~~~~~~~~~~
      |                                |
      |                                pj_size_t {aka unsigned int}
../include/pj/log.h:430:50: note: in definition of macro ‘pj_log_wrapper_3’
  430 |     #define pj_log_wrapper_3(arg)       pj_log_3 arg
      |                                                  ^~~
../src/pj/pool_caching.c:282:5: note: in expansion of macro ‘PJ_LOG’
  282 |     PJ_LOG(3,("cachpool", "   Capacity=%lu, max_capacity=%lu, used_cnt=%lu", \
      |     ^~~~~~
../src/pj/pool_caching.c:282:42: note: format string is defined here
  282 |     PJ_LOG(3,("cachpool", "   Capacity=%lu, max_capacity=%lu, used_cnt=%lu", \
      |                                        ~~^
      |                                          |
      |                                          long unsigned int
      |                                        %u
In file included from ../src/pj/pool_caching.c:21:
../src/pj/pool_caching.c:282:27: error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 4 has type ‘pj_size_t’ {aka ‘unsigned int’} [-Werror=format=]
  282 |     PJ_LOG(3,("cachpool", "   Capacity=%lu, max_capacity=%lu, used_cnt=%lu", \
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  283 |                              cp->capacity, cp->max_capacity, cp->used_count));
      |                                            ~~~~~~~~~~~~~~~~
      |                                              |
      |                                              pj_size_t {aka unsigned int}

and many times repeated...
```

sauwming:
Reverts pjsip/pjproject#3778

To fix #3785

### DIFF
--- a/pjlib-util/src/pjlib-util/stun_simple_client.c
+++ b/pjlib-util/src/pjlib-util/stun_simple_client.c
@@ -345,7 +345,7 @@ PJ_DEF(pj_status_t) pjstun_get_mapped_addr2(pj_pool_factory *pf,
         }
     }
 
-    TRACE_((THIS_FILE, "  Pool usage=%ld of %ld", pj_pool_get_used_size(pool),
+    TRACE_((THIS_FILE, "  Pool usage=%d of %d", pj_pool_get_used_size(pool),
             pj_pool_get_capacity(pool)));
 
     pj_pool_release(pool);
--- a/pjlib/src/pj/hash.c
+++ b/pjlib/src/pj/hash.c
@@ -187,7 +187,7 @@ static pj_hash_entry **find_entry( pj_pool_t *pool, pj_hash_table_t *ht,
 
         entry = PJ_POOL_ALLOC_T(pool, pj_hash_entry);
         PJ_LOG(6, ("hashtbl", 
-                   "%p: New p_entry %p created, pool used=%lu, cap=%lu", 
+                   "%p: New p_entry %p created, pool used=%u, cap=%u", 
                    ht, entry,  pj_pool_get_used_size(pool), 
                    pj_pool_get_capacity(pool)));
     }
--- a/pjlib/src/pj/pool.c
+++ b/pjlib/src/pj/pool.c
@@ -52,7 +52,7 @@ static pj_pool_block *pj_pool_create_block( pj_pool_t *pool, pj_size_t size)
     PJ_CHECK_STACK();
     pj_assert(size >= sizeof(pj_pool_block));
 
-    LOG((pool->obj_name, "create_block(sz=%lu), cur.cap=%lu, cur.used=%lu", 
+    LOG((pool->obj_name, "create_block(sz=%u), cur.cap=%u, cur.used=%u", 
          size, pool->capacity, pj_pool_get_used_size(pool)));
 
     /* Request memory from allocator. */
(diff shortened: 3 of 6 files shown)


nanangizz:
Right, I think those originally are compile warnings treated as errors (e.g: due to `-Werror=format`). `pj_size_t/pj_ssize_t` can be `long` or not, so print format can be tricky. So I guess if we decide to use `%ld/%lu`, we need to ignore the format warnings when `pj_size_t/pj_ssize_t` is not using `long`?

sauwming:
Yes, it is tricky.
Initially I want to revert this commit and resort to using cast instead, i.e. `"%u", (unsigned) ....`, but then I realise that we have been using `%lu` everywhere else, so the problem is not with this commit.

It turns out that we need to use `%zu` instead for `size_t`:
https://cplusplus.com/reference/cstdio/printf/
https://www.tutorialspoint.com/what-is-the-correct-way-to-use-printf-to-print-a-size-t-in-c-cplusplus

nanangizz:
The `%zu` may have a drawback as it seems to be introduced in C99, while AFAIK the library has been trying to conform to C89.

See https://copyprogramming.com/howto/is-the-zu-specifier-required-for-printf

sauwming:
Ah okay, I guess we shouldn't use `zu`, it seems that it may cause issues with VS and mingw as well.

So I suppose the safest approach for us is to cast it to unsigned long, i.e. `printf("%lu\n", (unsigned long)n)`

PR #3789, opened 2023-11-24
pjsua2: add destroyPlayer to AudioMediaPlayer

### LINKED ISSUE #3787: pjsua2: player cannot be destroyed when Normal call clearing
### Describe the bug

When user closes the call while wav is playing, I received the `PJSIP_INV_STATE_DISCONNECTED` event with reason `Normal call clearing`, but when I need to make another call I can't because I received `PJ_ETOOMANY`

When I receive the event, when I tried to stop the transmission an error invalid sink happens in pjsua_conf_disconnect.

Debugging I found that player is not destroyed in this case in function `processStateChange` https://github.com/pjsip/pjproject/blob/9287ac21e6dda91f68a0e30dea88e846c960a549/pjsip/src/pjsua2/call.cpp#L983 when call is disconnected.

Adding this code at the end of the if condition helps because I'm using only one max call and one player in config_site:

```c
/* Destroy file players */
        for (i=0; i<PJ_ARRAY_SIZE(pjsua_var.player); ++i) {
            if (pjsua_var.player[i].port) {
                PJ_LOG(6,(THIS_FILE, "!!!!!!!!!!!!!! Destroying player id %d", i));
                pjsua_player_destroy(i);
            }
        }
```

But not safe with multiple calls.

### Steps to reproduce

1. Compile pj with config_site.h with:
```
#define PJSUA_MAX_CALLS 1
#define PJSUA_MAX_ACC 1
#define PJSUA_MAX_PLAYERS 1
#define PJSUA_MAX_RECORDERS 1
```
2. make call with pjsua2
3. Close the call when audio is playing
4. Try to stop transmission, destroy player and AudioMedia (this it's not possible)
5. Try to make another call

In 5 I have the problem: PJ_ETOOMANY is received because previous player was not destroyed, also I can't destroy it when Normal call clearing happens.

### PJSIP version

2.14 commit 6157218806094deeab818cee721484906b3219fe

### Context

Using Ubuntu 22.04

I compile pjsua2 wrapper for c#

config_site.h as above, to allow one call and one player.

### Log, call stack, etc

```shell
Pending...
```

xhit:
With this, the dev can destroy a specific player when needed without wait GC in some languages with not Dispose

### DIFF
--- a/pjsip/src/pjsua2/media.cpp
+++ b/pjsip/src/pjsua2/media.cpp
@@ -362,10 +362,12 @@ AudioMediaPlayer::~AudioMediaPlayer()
     }
 }
 
-void AudioMediaPlayer::destroyPlayer(const int &playerID)
+void AudioMediaPlayer::destroyPlayer()
 {
-    PJ_LOG(4,(THIS_FILE, "Destroying specific player %d", playerID));
-    pjsua_player_destroy(playerId);
+    if (playerId != PJSUA_INVALID_ID) {
+        PJSUA2_CATCH_IGNORE( unregisterMediaPort() );
+        pjsua_player_destroy(playerId);
+    }
 }
 
 void AudioMediaPlayer::createPlayer(const string &file_name,
@@ -463,7 +465,6 @@ AudioMediaPlayerInfo AudioMediaPlayer::getInfo() const PJSUA2_THROW(Error)
     info.payloadBitsPerSample   = pj_info.payload_bits_per_sample;
     info.sizeBytes              = pj_info.size_bytes;
     info.sizeSamples            = pj_info.size_samples;
-    info.playerId               = playerId;
 
     return info;
 }
--- a/pjsip/src/pjsua2/media.cpp
+++ b/pjsip/src/pjsua2/media.cpp
@@ -362,6 +362,12 @@ AudioMediaPlayer::~AudioMediaPlayer()
     }
 }
 
+void AudioMediaPlayer::destroyPlayer(const int &playerID)
+{
+    PJ_LOG(4,(THIS_FILE, "Destroying specific player %d", playerID));
+    pjsua_player_destroy(playerId);
+}
+
 void AudioMediaPlayer::createPlayer(const string &file_name,
                                     unsigned options)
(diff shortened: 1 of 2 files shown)


sauwming:
This looks like a clear violation of Object Oriented Programming.

You're not supposed to destroy something that belongs to another object.

pjsip/src/pjsua2/media.cpp
```diff
@@ -289,15 +289,15 @@ static pj_status_t get_frame(pjmedia_port *port, pjmedia_frame *frame)
...
+#if ((defined(_MSVC_LANG) && _MSVC_LANG <= 199711L) || __cplusplus <= 199711L)
+    /* C++98 does not have Vector::data() */
+    if (frame->size > 0)
         pj_memcpy(frame->buf, &frame_.buf[0], frame->size);
 #else
-    /* Newer than C++98 */
+    /* Newer than C++98 */
     pj_memcpy(frame->buf, frame_.buf.data(), frame->size);
-#endif
+#endif
```
xhit, line 300: Note: these lines only changes CRLF to LF

xhit:
Fixed.

Also updated comment to specify use with other languages, like Golang, to not wait for GC to acts.

sauwming:
It's essentially the same as what the class destructor does.

I can understand though why this is needed, especially for GC languages such as Java, Python, C#; similar to implementing `IDisposable` in C# like what you described (or `AutoClosable` in Java).

Unfortunately, in my opinion, it's unlikely to be merged to the master branch since the implementation is only limited to AudioMediaPlayer, while if we really do need this feature, it probably needs to be more comprehensive than this, i.e.:
- encompass all PJSUA2 classes (or at least, all the classes that require this feature)
- need safeguard against double destruction, and against race condition (i.e. manual disposal and GC-initiated destruction).

xhit:
In C# or Java not problem because the Dispose, but in Go I need to wait the GC.

You are correct about expand more this, I only covers my use case.

And about race conditions, GC will be triggered only if not referenced, so, calling a virtual destructor like this cannot interfere with GC, because to be called the object need to be referenced. I think this should be in dev side.

PR #3805, opened 2023-12-15
Prevent immediate tsx termination upon transport error

sauwming:
It is reported that we will immediately terminate the transaction and the INVITE session upon transport error:
```
11:09:45.912         inv0x10953d5a8  ..Sending Response msg 200/INVITE/cseq=24793 (tdta0x109535da8)
11:09:45.912         dlg0x10953d5a8  ...Sending Response msg 200/INVITE/cseq=24793 (tdta0x109535da8)
11:09:45.912         tsx0x1099a0b28  ...Sending Response msg 200/INVITE/cseq=24793 (tdta0x109535da8) in state Proceeding
11:09:45.913         tsx0x1099a0b28  ....Failed to send Response msg 200/INVITE/cseq=24793 (tdta0x109535da8)! err=120111 (Connection refused)
11:09:45.913         tsx0x1099a0b28  ....State changed from Proceeding to Terminated, event=TRANSPORT_ERROR
11:09:45.913         dlg0x10953d5a8  .....Transaction tsx0x1099a0b28 state changed to Terminated
11:09:45.913         inv0x10953d5a8  ......State changed from INCOMING to DISCONNECTED, event=TSX_STATE
```

However, according to RFC 3261 (https://www.rfc-editor.org/rfc/rfc3261), section 13.3.1.4, we are supposed to periodically retry the 2xx:
```
The UAS core generates a 2xx response.
...
   Once the response has been constructed, it is passed to the INVITE
   server transaction.  Note, however, that the INVITE server
   transaction will be destroyed as soon as it receives this final
   response and passes it to the transport.  Therefore, it is necessary
   to periodically pass the response directly to the transport until the
   ACK arrives.  The 2xx response is passed to the transport with an
   interval that starts at T1 seconds and doubles for each
   retransmission until it reaches T2 seconds (T1 and T2 are defined in
   [Section 17](https://www.rfc-editor.org/rfc/rfc3261.html#section-17)).  Response retransmissions cease when an ACK request for
   the response is received.  This is independent of whatever transport
   protocols are used to send the response.
```

Which is in line with section 8.1.3.1:
```
8.1.3.1 Transaction Layer Errors
If a fatal transport error is reported by the transport layer
(generally, due to fatal ICMP errors in UDP or connection failures in
TCP), the condition MUST be treated as a 503 (Service Unavailable)
status code.
```

And our own comment in https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_transaction.c#L2181
```
[...]

PR #3806, opened 2023-12-19
Terminate client transaction upon transport error

trengginas:
#3805 will prevent immediate transaction termination upon transport error.
This will introduce new behavioral change which is an issue.

**Provided an outgoing INVITE  without patch**
```
22:27:26.194    inv05253D64  .UAC invite session created for dialog dlg05253D64
22:27:26.194     sip_util.c  .Request msg INVITE/cseq=8155 (tdta05262D64) created.
22:27:26.194    inv05253D64  ..Sending Request msg INVITE/cseq=8155 (tdta05262D64)
22:27:26.194    dlg05253D64  ...Sending Request msg INVITE/cseq=8155 (tdta05262D64)
22:27:26.194    tsx0530BEE4  ....Transaction created for Request msg INVITE/cseq=8155 (tdta05262D64)
22:27:26.194    tsx0530BEE4  ...Sending Request msg INVITE/cseq=8155 (tdta05262D64) in state Null
22:27:26.194  sip_resolve.c  ....Target '192.168.1.114:0' type=TCP resolved to '192.168.1.114:5060' type=TCP (TCP transport)
22:27:26.194   tcpc06C06114  ....TCP client transport created
22:27:26.194   tcpc06C06114  ....TCP transport 192.168.100.51:63251 is connecting to 192.168.1.114:5060...
22:27:26.194   pjsua_core.c  ....TX 955 bytes Request msg INVITE/cseq=8155 (tdta05262D64) to TCP 192.168.1.114:5060:
22:27:26.198    tsx0530BEE4  ....State changed from Null to Calling, event=TX_MSG
22:27:26.198    dlg05253D64  .....Transaction tsx0530BEE4 state changed to Calling
22:27:26.198    inv05253D64  ......State changed from NULL to CALLING, event=TSX_STATE
22:27:26.198    pjsua_app.c  .......Call 0 state changed to CALLING
22:27:26.198    pjsua_app.c  .......Tsx tsx0530BEE4 state changed: Calling code: 0
22:27:31.204   sound_port.c !EC suspended because of inactivity
22:27:47.244   tcpc06C06114  TCP connect() error: [code=130060]: Connection timed out (WSAETIMEDOUT)
22:27:47.244    tsx0530BEE4  Failed to send Request msg INVITE/cseq=8155 (tdta05262D64)! err=130060 (Connection timed out (WSAETIMEDOUT))
22:27:47.244    tsx0530BEE4  State changed from Calling to Terminated, event=TRANSPORT_ERROR
22:27:47.244    dlg05253D64  .Transaction tsx0530BEE4 state changed to Terminated
22:27:47.244    inv05253D64  ..State changed from CALLING to DISCONNECTED, event=TSX_STATE
22:27:47.244  pjsua_media.c  ...Call 0: deinitializing media..
```

**With patch**
```
22:33:35.824    inv05451564  .UAC invite session created for dialog dlg05451564
22:33:35.824     sip_util.c  .Request msg INVITE/cseq=7217 (tdta05460564) created.
22:33:35.824    inv05451564  ..Sending Request msg INVITE/cseq=7217 (tdta05460564)
22:33:35.824    dlg05451564  ...Sending Request msg INVITE/cseq=7217 (tdta05460564)
22:33:35.824    tsx0550BEE4  ....Transaction created for Request msg INVITE/cseq=7217 (tdta05460564)
22:33:35.824    tsx0550BEE4  ...Sending Request msg INVITE/cseq=7217 (tdta05460564) in state Null
22:33:35.824  sip_resolve.c  ....Target '192.168.1.114:0' type=TCP resolved to '192.168.1.114:5060' type=TCP (TCP transport)
22:33:35.824   tcpc06E06114  ....TCP client transport created
22:33:35.824   tcpc06E06114  ....TCP transport 192.168.100.51:63302 is connecting to 192.168.1.114:5060...
22:33:35.824   pjsua_core.c  ....TX 955 bytes Request msg INVITE/cseq=7217 (tdta05460564) to TCP 192.168.1.114:5060:
22:33:35.837    tsx0550BEE4  ....State changed from Null to Calling, event=TX_MSG
22:33:35.837    dlg05451564  .....Transaction tsx0550BEE4 state changed to Calling
22:33:35.837    inv05451564  ......State changed from NULL to CALLING, event=TSX_STATE
22:33:35.837    pjsua_app.c  .......Call 0 state changed to CALLING
22:33:35.837    pjsua_app.c  .......Tsx tsx0550BEE4 state changed: Calling code: 0
22:33:56.871   tcpc06E06114  TCP connect() error: [code=130060]: Connection timed out (WSAETIMEDOUT)
22:33:56.871    tsx0550BEE4  Failed to send Request msg INVITE/cseq=7217 (tdta05460564)! err=130060 (Connection timed out (WSAETIMEDOUT))
22:33:56.871   tcpc06E06114  TCP send() error, sent=-130060
22:33:56.872    pjsua_app.c  SIP TCP transport is disconnected from 192.168.1.114:5060: Connection timed out (WSAETIMEDOUT) [status=130060]
22:33:56.872    pjsua_acc.c  Disconnected notification for transport tcpc06E06114
22:33:56.872 sip_transport.  .Transport tcpc06E06114 shutting down, force=0
22:33:56.873 sip_transport.  Transport tcpc06E06114 is being destroyed due to timeout in idle timer
22:33:56.873   tcpc06E06114  TCP transport destroyed with reason 130060: Connection timed out (WSAETIMEDOUT)
22:34:07.848    tsx0550BEE4  Timeout timer event
22:34:07.848    tsx0550BEE4  .State changed from Calling to Terminated, event=TIMER
22:34:07.848    dlg05451564  ..Transaction tsx0550BEE4 state changed to Terminated
22:34:07.848    inv05451564  ...State changed from CALLING to DISCONNECTED, event=TSX_STATE
22:34:07.848  pjsua_media.c  ....Call 0: deinitializing media..
```

The outgoing INVITE transaction will be terminated by timeout timer and the call disconnected notification will require additional time.

As stated in RFC 6026 (https://www.rfc-editor.org/rfc/rfc6026.html), section 7.1:
```
A server transaction MUST NOT discard transaction state based only on
   encountering a non-recoverable transport error when sending a
   response.  Instead, the associated INVITE server transaction state
   machine MUST remain in its current state.  (Timers will eventually
   cause it to transition to the "Terminated" state).  This allows
   retransmissions of the INVITE to be absorbed instead of being
   processed as a new request.
```

And section 7.2:
```
Note that it is not necessary to preserve client transaction state
   upon the detection of unrecoverable transport errors.  Existing
   requirements ensure the TU has been notified, and the new
   requirements in this document ensure that any received retransmitted
   response will be dropped since there will no longer be any matching
   transaction state.
```

Upon transport error, only server transaction that must not be discarded/terminated.

This PR will terminate client transaction upon transport error.

### DIFF
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -2162,6 +2162,7 @@ static void send_msg_callback( pjsip_send_state *send_state,
             char errmsg[PJ_ERR_MSG_SIZE];
             pjsip_status_code sc;
             pj_str_t err;
+            pj_bool_t terminate_tsx = PJ_FALSE;
 
             tsx->transport_err = (pj_status_t)-sent;
 
@@ -2187,12 +2188,21 @@ static void send_msg_callback( pjsip_send_state *send_state,
             else
                 sc = PJSIP_SC_TSX_TRANSPORT_ERROR;
 
-            /* For client transaction, we terminate the transaction.
-             * Otherwise, we terminate the transaction for 502 error.
-             * For 503, we will retry it.
+            /* Terminate transaction on these conditions:
+             * - client transaction
+             * - server transaction with 502 error. We will retry for 503.
+             * - server transaction with status code 100 on
+             *   PJSIP_TSX_STATE_TRYING state
              * See https://github.com/pjsip/pjproject/pull/3805
              */
-            if ((tsx->role == PJSIP_ROLE_UAC || sc == PJSIP_SC_BAD_GATEWAY) &&
+            if ((tsx->role == PJSIP_ROLE_UAC) ||
+                (sc == PJSIP_SC_BAD_GATEWAY) ||
+                (tsx->state == PJSIP_TSX_STATE_TRYING && 
+                 tsx->status_code == 100))
+            {
+                terminate_tsx = PJ_TRUE;
+            }
+            if (terminate_tsx && 
                 tsx->state != PJSIP_TSX_STATE_TERMINATED &&
                 tsx->state != PJSIP_TSX_STATE_DESTROYED)
             {
@@ -2440,6 +2450,9 @@ static pj_status_t tsx_send_msg( pjsip_transaction *tsx,
         
         char errmsg[PJ_ERR_MSG_SIZE];
         pj_str_t err;
+
+        if (tsx->transport == NULL)
+            status = PJ_ENOTFOUND;
 
         if (status == PJ_SUCCESS) {
             pj_assert(!"Unexpected status!");
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -2170,7 +2170,6 @@ static void send_msg_callback( pjsip_send_state *send_state,
             char errmsg[PJ_ERR_MSG_SIZE];
             pjsip_status_code sc;
             pj_str_t err;
-            pj_bool_t terminate_tsx = PJ_FALSE;
 
             tsx->transport_err = (pj_status_t)-sent;
 
@@ -2196,21 +2195,12 @@ static void send_msg_callback( pjsip_send_state *send_state,
             else
                 sc = PJSIP_SC_TSX_TRANSPORT_ERROR;
 
-            /* Terminate transaction on these conditions:
-             * - client transaction
-             * - server transaction with 502 error. We will retry for 503.
-             * - server transaction with status code 100 on
-             *   PJSIP_TSX_STATE_TRYING state
-             * See https://github.com/pjsip/pjproject/pull/3805
+            /* For UAC tsx, we directly terminate the transaction.
+             * For UAS tsx, we terminate the transaction for 502 error,
+             * and will retry for 503.
+             * See #3805 and #3806.
              */
-            if ((tsx->role == PJSIP_ROLE_UAC) ||
-                (sc == PJSIP_SC_BAD_GATEWAY) ||
-                (tsx->state == PJSIP_TSX_STATE_TRYING && 
-                 tsx->status_code == 100))
-            {
-                terminate_tsx = PJ_TRUE;
-            }
-            if (terminate_tsx && 
+            if ((tsx->role == PJSIP_ROLE_UAC || sc == PJSIP_SC_BAD_GATEWAY) &&
                 tsx->state != PJSIP_TSX_STATE_TERMINATED &&
                 tsx->state != PJSIP_TSX_STATE_DESTROYED)
             {
@@ -2285,7 +2275,16 @@ static void transport_callback(void *token, pjsip_tx_data *tdata,
     pj_grp_lock_acquire(tsx->grp_lock);
     tsx->transport_flag &= ~(TSX_HAS_PENDING_TRANSPORT);
 
-    if (sent > 0) {
+    if (sent > 0 || tsx->role == PJSIP_ROLE_UAS) {
+        if (sent < 0) {
+            /* For UAS transactions, we just print error log
+             * and continue as per normal.
+             */
+            PJ_PERROR(2,(tsx->obj_name, (pj_status_t)-sent,
+                          "Transport failed to send %s!",
+                          pjsip_tx_data_get_info(tdata)));
+        }
+
         /* Pending destroy? */
         if (tsx->transport_flag & TSX_HAS_PENDING_DESTROY) {
             tsx_set_state( tsx, PJSIP_TSX_STATE_DESTROYED,
@@ -2318,7 +2317,7 @@ static void transport_callback(void *token, pjsip_tx_data *tdata,
     }
     pj_grp_lock_release(tsx->grp_lock);
 
-    if (sent < 0) {
+    if (sent < 0 && tsx->role == PJSIP_ROLE_UAC) {
         pj_time_val delay = {0, 0};
 
         PJ_PERROR(2,(tsx->obj_name, (pj_status_t)-sent,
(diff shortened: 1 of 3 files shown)


sauwming:
Sorry, it looked like the previous PR introduced backward incompatibility and failed tests. Could you please fix it as well so that it passed the CI tests?

trengginas:
Failed pjsip-test:
* now some transactions will not get immediately terminated, so we need to increase the end test time
* assertion in https://github.com/pjsip/pjproject/blob/bc5e8267ca28a2a4633d70fdf16b741363468bcb/pjsip/src/pjsip/sip_transaction.c#L2458
  due to no transport being set on `send_msg_callback()`
* In case there's a transport error when sending provisional responses, retransmit will only be done to responses greater to 100.
https://github.com/pjsip/pjproject/blob/bc5e8267ca28a2a4633d70fdf16b741363468bcb/pjsip/src/pjsip/sip_transaction.c#L3029
For 100 response, we need to immediately terminate it.
https://datatracker.ietf.org/doc/html/rfc3261#section-17.2.1

sauwming:
>   For 100 response, we need to immediately terminate it.
>   https://datatracker.ietf.org/doc/html/rfc3261#section-17.2.1

Why does 100 response need to be treated differently?

trengginas:
Currently, retransmission will be done only for > 100 responses.
https://github.com/pjsip/pjproject/blob/bc5e8267ca28a2a4633d70fdf16b741363468bcb/pjsip/src/pjsip/sip_transaction.c#L3029
This is based on https://datatracker.ietf.org/doc/html/rfc3261#section-17.2.1 (see Figure 7: INVITE server transaction)

This will make 100 response to linger (no timeout/retransmission) if it fails to send.
Based on the above figure, then 100 responses should be terminated immediately. 
The other alternative would be to retransmit it.

sauwming:
> This is based on https://datatracker.ietf.org/doc/html/rfc3261#section-17.2.1 (see Figure 7: INVITE server transaction)
> This will make 100 response to linger (no timeout/retransmission) if it fails to send. Based on the above figure, then 100 responses should be terminated immediately. The other alternative would be to retransmit it.

Ah okay, I understand now that currently we don't retransmit 100, but why doesn't the timeout work?

After re-reading both RFCs, I'm still not entirely sure as to what we should do upon 100 failure. Based on my understanding: - the server tsx shouldn't be terminated, - timeout must still run. Not sure about retransmission, I believe we need to do this also, but maybe there's a reason right now we intentionally don't do this?

trengginas:
Because timer timeout isn't scheduled for provisional responses.
On tests, response code > 100 will be terminated when sending retransmission failed:
https://github.com/pjsip/pjproject/blob/bc5e8267ca28a2a4633d70fdf16b741363468bcb/pjsip/src/pjsip/sip_transaction.c#L2467
Regarding rfc6062, the title stated "Correct Transaction Handling for 2xx Responses". so, it should only apply to 2xx responses?

sauwming:
I believe the title of RFC 6026 states its primary or main content, which doesn't necessarily mean that it will only affect 2xx responses. Section 8 of the RFC elaborates all the revisions, and Figure 7 shows the overall updated INVITE server transaction flow.

So, I believe the sip_tsx modification in the previous commit https://github.com/pjsip/pjproject/pull/3806/commits/035bad7f2a73871888942b16726e459652559d91 was already correct. We don't need to terminate upon 1xx failure.

As for the 100 retransmission, I suggest to comment the 100 condition to activate retransmission for 100, and test if it introduces compatibility issue.
https://github.com/pjsip/pjproject/blob/bc5e8267ca28a2a4633d70fdf16b741363468bcb/pjsip/src/pjsip/sip_transaction.c#L3029
If it doesn't cause incompatibility, I think it should be okay.

nanangizz:
> I believe the title of RFC 6026 states its primary or main content, which doesn't necessarily mean that it will only affect 2xx responses. Section 8 of the RFC elaborates all the revisions, and Figure 7 shows the overall updated INVITE server transaction flow.

Yes, the updates on the diagrams seem to affect other responses/tsx-states too.


> So, I believe the sip_tsx modification in the previous commit [035bad7](https://github.com/pjsip/pjproject/commit/035bad7f2a73871888942b16726e459652559d91) was already correct. We don't need to terminate upon 1xx failure.
> 

Agree.


> As for the 100 retransmission, I suggest to comment the 100 condition to activate retransmission for 100, and test if it introduces compatibility issue.
> 
> https://github.com/pjsip/pjproject/blob/bc5e8267ca28a2a4633d70fdf16b741363468bcb/pjsip/src/pjsip/sip_transaction.c#L3029
> 
> If it doesn't cause incompatibility, I think it should be okay.

Not sure about this, but I won't vote for 100 retransmission. It is supposed to be unreliable, immediate (to cease INVITE retransmission) and hop-by-hop (101-199 will be forwarded by proxy, but 100 will not), even PRACK's RFC prohibits 100 reliability ([ref](https://www.rfc-editor.org/rfc/rfc3262.html#section-3)), so I guess it is better to leave it as is (i.e: no retransmission).

---

Btw, it seems we now support rfc6026 partially in the recent changes, haven't really checked the details myself, but would we consider to support it fully (e.g: add new tsx state "Accepted", etc)?

sauwming:
> Not sure about this, but I won't vote for 100 retransmission. It is supposed to be unreliable, immediate (to cease INVITE retransmission) and hop-by-hop (101-199 will be forwarded by proxy, but 100 will not), even PRACK's RFC prohibits 100 reliability ([ref](https://www.rfc-editor.org/rfc/rfc3262.html#section-3)), so I guess it is better to leave it as is (i.e: no retransmission).

Okay, let's go with this.

> Btw, it seems we now support rfc6026 partially in the recent changes, haven't really checked the details myself, but would we consider to support it fully (e.g: add new tsx state "Accepted", etc)?

For now, I plan to put in our roadmap for further discussion, especially since it seems quite major.

sauwming:
List of changes in the latest commit:
* sip transaction
  - `transport_callback()` need to handle the case to ignore sending failure for UAS tsx.
  -` tsx_send_msg()` should not treat the error as permanent for UAS tsx.

* pjsip test
  Now the expected test result will be timeout, instead of transport failure 
  - For test 10 (failed transport in `TRYING` state), the expected result is timeout by app (since tsx will not timeout before it sends a final response).
  - For test 11 and 12 (failed transport in `PROCEEDING/COMPLETED` state), the timeout is by the tsx (`PJSIP_SC_REQUEST_TIMEOUT`).

nanangizz (approved):

sauwming (approved):

PR #3809, opened 2023-12-22
Fix return value for pjsua_set_snd_dev2() with no sound device option

trengginas:
The `pjsua_set_snd_dev2()` method will always return -1 when `snd_param->capture_dev` and `snd_param->playback_dev` is set to `PJSUA_SND_NO_DEV`.
This will fix the return code of such case.


pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -2265,9 +2265,12 @@ PJ_DEF(pj_status_t) pjsua_set_snd_dev2(const pjsua_snd_dev_param *snd_param)
         snd_param->playback_dev == PJSUA_SND_NO_DEV)
     {
         PJSUA_UNLOCK();
+        status = PJ_SUCCESS;
```
sauwming, line 2268: Just put the initialisation above, to replace
`pj_status_t status = -1`

sauwming (approved):

nanangizz (approved):

PR #3817, opened 2024-01-02
Enable PendingJob use on Python

sauwming:
Continuing #3774.

First of all, using `PendingJob` is only useful on Python.

`Endpoint::utilAddPendingJob()` will *always* execute the pending job directly unless `mainThreadOnly` is set to `true` (the default is false) (and thread count 0).
```
     * Utility to register a pending job to be executed by main thread.
     * If EpConfig::UaConfig::mainThreadOnly is false, the job will be
     * executed immediately.
    void utilAddPendingJob(PendingJob *job);
```
```
    /**
     * When this flag is non-zero, all callbacks that come from thread
     * other than main thread will be posted to the main thread and
     * to be executed by Endpoint::libHandleEvents() function. This
     * includes the logging callback. Note that this will only work if
     * threadCnt is set to zero and Endpoint::libHandleEvents() is
     * performed by main thread. By default, the main thread is set
     * from the thread that invoke Endpoint::libCreate()
     *
     * Default: false
     */
    bool                mainThreadOnly;
```
```
    /* See if we can execute immediately */
    if (!mainThreadOnly || pj_thread_this()==mainThread) {
        job->execute(false);
        delete job;
        return;
    }
```

This makes the problem much less complicated since `mainThreadOnly` typically won't be enabled on Java nor C# (and enabling this on Android is definitely a no no due to the main thread being the GUI thread).
Therefore, the simplest solution has been provided by @nanangizz in (https://github.com/pjsip/pjproject/pull/3774#issuecomment-1803227688) using `__disown__()`.

This turns out to be a classical "elements within a container" problem, where we want the container to own the elements. And just for future reference, in case that we also encounter similar issues for Java and C#, the solutions are readily available in the docs:
- https://www.swig.org/Doc4.1/Java.html#Java_memory_management_objects (Memory management for objects passed to the C++ layer)
- https://www.swig.org/Doc4.1/CSharp.html#CSharp_memory_management_objects (Memory management for objects passed to the C++ layer)

For Python, the solution can be found here:
- https://www.swig.org/Doc4.1/Python.html#Python_nn35 (Ownership and object destruction)
- https://stackoverflow.com/questions/50259152/adding-swig-pythoncode-to-set-thisown-flag-on-python-object

### DIFF
--- a/pjsip-apps/src/swig/pjsua2.i
+++ b/pjsip-apps/src/swig/pjsua2.i
@@ -121,7 +121,13 @@ using namespace pj;
 %feature("director") FindBuddyMatch;
 %feature("director") AudioMediaPlayer;
 %feature("director") AudioMediaPort;
-%feature("director") PendingJob;
+// PendingJob is only used on Python
+#ifdef SWIGPYTHON
+    %feature("director") PendingJob;
+#else
+    %ignore pj::PendingJob;
+    %ignore pj::Endpoint::utilAddPendingJob;
+#endif
 
 //
 // STL stuff.
--- a/pjsip-apps/src/swig/pjsua2.i
+++ b/pjsip-apps/src/swig/pjsua2.i
@@ -121,6 +121,7 @@ using namespace pj;
 %feature("director") FindBuddyMatch;
 %feature("director") AudioMediaPlayer;
 %feature("director") AudioMediaPort;
+%feature("director") PendingJob;
 
 //
 // STL stuff.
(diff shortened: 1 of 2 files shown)


sauwming:
Test output:
```
Creating job 1
Job created id: 4312656208 value: 17136
Adding job 1
Executing job value: 17136 False
Job deleted id: 4312656208 value: 17136
Creating job 2
Job created id: 4311885776 value: 84049
Adding job 2
Collecting gc 1
Handling events
Executing job value: 84049 True
Collecting gc 2
Job deleted id: 4311885776 value: 84049
```

pjsip-apps/src/swig/pjsua2.i
```diff
@@ -121,6 +121,7 @@ using namespace pj;
 %feature("director") FindBuddyMatch;
 %feature("director") AudioMediaPlayer;
 %feature("director") AudioMediaPort;
+%feature("director") PendingJob;
```
nanangizz, line 124: If this is only for Python, shouldn't we enable the SWIG director for Python only?
Even perhaps disable/remove the `PendingJob` class & `utilAddPendingJob()` method on other languages?

trengginas (approved):

PR #3823, opened 2024-01-08
Fixed Service-Route update upon registration

### LINKED ISSUE #3790: [PJSUA][ACC] Service-Route header not cleared
### Describe the bug

I recently encountered a call routing problem, which prevented outgoing calls to be made.
The scenario is this:

- UA registers to registrar via SIP proxy (actually a bunch of proxies (A, B, C), one of which is selected via DNS SRV lookup).
For this, there is an outbound proxy configured for the account.
Please note, the proxies have different priorities: A (high), B (middle), C (low).

- UA is registered to Proxy A, no Service-Route headers received in the response
- Everything working fine
- For some reason, registration fails and UA registers via Proxy B
- Registration is ok, but this time the response contains a "Service-Route" header which contains the IP address of proxy B
- The Service-Route header is then added to the "Route" header list of the account
The account now has two route header: The preconfigured outbound proxy and the Route learned with Service-Route.

All outgoing calls via this account will carry these two Route headers.

**The problem I encountered:**

- After a while the DNS-SRV lookup returns a new set of proxies in which A is different (A-new, B, C) (though it doesn't matter that A is different now, I think)
- PJSUA send register to A-new
- Registration is ok, but this time the registrar does NOT send any Service-Route headers

- Outgoing call attempts fail from this point on because the account still injects the Route header learned from the other proxy

**The general problem seems to be:**

The route learned via Service-Route in REGISTER is NOT updated \
in subsequent REGISTER responses/results (be they success or failure).

A failure or UNREGISTER should clear the learned Route headers and a successful response should ALWAYS update the header list,
this includes the case above, where a route has been learned through Service-Route but subsequent responses didn't contain any Service-Route headers.

I'm sorry I don't have a PR handy; I've just analyzed this issue and it's not so easy for me to reproduce this at the moment;
I'm putting this here so this issue gets documented.

### Steps to reproduce

1. Register to a registrar where the DNS SRV lookup returns two entries (registrar A (high prio), registrar B (low prio))
2. Let UA register to A (A must not send Service-Route headers in its response)
3. Kill Registrar A and wait for UA to refresh the registration
4. UA should register to B now; B sends Service-Route header in its response
5. Resurrect A and wait for UA to refresh its registration
6. UA should register to A now (A must not send Service-Route headers in its response)
7. Outgoing calls still carry the Route header entry learned via Service-Route in 4

### PJSIP version

2.14

### Context

Any platform

### Log, call stack, etc

```shell
Sorry, I cannot share the log I've got at hand because it contains private data.
```

sauwming:
To fix #3790.

According to the RFC (https://datatracker.ietf.org/doc/html/rfc3608.html#section-6.1):
```
If
   the UA refreshes the registration, the stored value of the Service-
   Route is updated according to the Service-Route header field of the
   latest 200 class response.  If there is no Service-Route header field
   in the response, the UA clears any service route for that address-
   of-record previously stored by the UA.  If the re-registration
   request is refused or if an existing registration expires and the UA
   chooses not to re-register, the UA SHOULD discard any stored service
   route for that address-of-record.
```

There are a couple issues with the current implementation:
- We do not clear the service route if there is no Service-Route header found in the response.
- We do not clear the service route upon registration failure.

Thanks to @andreas-wehrmann for the report and investigation.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -2002,7 +2002,7 @@ static void update_service_route(pjsua_acc *acc, pjsip_rx_data *rdata)
     pj_size_t rcnt;
 
     /* Find and parse Service-Route headers */
-    for (;;) {
+    for (;(rdata);) {
         char saved;
         int parsed_len;
 
@@ -2053,9 +2053,6 @@ static void update_service_route(pjsua_acc *acc, pjsip_rx_data *rdata)
         if ((void*)hsr == (void*)&rdata->msg_info.msg->hdr)
             break;
     }
-
-    if (uri_cnt == 0)
-        return;
 
     /* 
      * Update account's route set 
@@ -2416,6 +2413,9 @@ static void regc_cb(struct pjsip_regc_cbparam *param)
                    (int)param->reason.slen, param->reason.ptr));
         destroy_regc(acc, PJ_TRUE);
 
+        /* Clear Service-Route header */
+        update_service_route(acc, NULL);
+
         /* Stop keep-alive timer if any. */
         update_keep_alive(acc, PJ_FALSE, NULL);
     } else if (PJSIP_IS_STATUS_IN_CLASS(param->code, 200)) {
@@ -2446,9 +2446,6 @@ static void regc_cb(struct pjsip_regc_cbparam *param)
                 acc_check_nat_addr(acc, (acc->cfg.contact_rewrite_method & 3),
                                    param))
             {
-                /* Check and update Service-Route header */
-                update_service_route(acc, param->rdata);
-
                 PJSUA_UNLOCK();
                 pj_log_pop_indent();
 
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -2446,6 +2446,9 @@ static void regc_cb(struct pjsip_regc_cbparam *param)
                 acc_check_nat_addr(acc, (acc->cfg.contact_rewrite_method & 3),
                                    param))
             {
+                /* Check and update Service-Route header */
+                update_service_route(acc, param->rdata);
+
                 PJSUA_UNLOCK();
                 pj_log_pop_indent();
 


sauwming:
Attached is the SIPP xml for testing the issue:
[uas-register.xml.txt](https://github.com/pjsip/pjproject/files/13868567/uas-register.xml.txt)

Test flow:
1. REGISTER-200 OK with Service Route header
2. Confirm that outgoing call contains that header
3. Re-REGISTER-200 OK without any Service Route header
4. Confirm that outgoing call now contains no Route header

1. REGISTER-200 OK with Service Route header
2. Confirm that outgoing call contains that header
3. Re-REGISTER-failed
4. Confirm that outgoing call now contains no Route header

nanangizz (approved):

trengginas (approved):

trengginas (approved):

andreas-wehrmann:
Though I haven't tested it, I believe there may be another issue with pjsua_acc_modify().
From the looks of it, it seems to assume that the route set consists always of the global outbound proxy list followed by the account specific proxy list (if any).

See here (global outbound proxies first):
https://github.com/pjsip/pjproject/blob/4169142bd6b5a3d5215bbfdce04b6bda411d5634/pjsip/src/pjsua-lib/pjsua_acc.c#L1171
and here (account proxies always last):
https://github.com/pjsip/pjproject/blob/4169142bd6b5a3d5215bbfdce04b6bda411d5634/pjsip/src/pjsua-lib/pjsua_acc.c#L1191

It doesn't seem to account for any additional routes injected via Service-Route
in which case it may be best to simply discard any headers learned via Service-Route?

andreas-wehrmann:
update_service_route() appends the headers to the route set of an account.
So the route set is made up like this: global proxies + account proxies + service route.

My worry is, that the code here:
https://github.com/pjsip/pjproject/blob/4169142bd6b5a3d5215bbfdce04b6bda411d5634/pjsip/src/pjsua-lib/pjsua_acc.c#L1191

removes the service route headers (if there are any) when it actually wants to remove the proxies of the account.
Or am I not understanding this code correctly?

sauwming:
You're right. I understand the problem now.

andreas-wehrmann:
Ok, so the questions remains what to do when?

- Forget all Service-Routes on account modify unconditionally and let it relearn?
- Forget Service-Routes only when there is a change to the global/account proxy list?
- Forget Service-Routes only when there is change to either proxy list and/or the registrar (what about the Contact header and possibly other fields?)

In my naiveté I tend towards the first since it seems to be the option which is least error prone.

sauwming:
I'm currently leaning towards the third option (only change in either proxy list or the registrar).
And I will probably open a new PR since it's a different issue (albeit related to this one).

PR #3825, opened 2024-01-10
Fixed account's route set update when modifying account

sauwming:
There's an issue with account modification via `pjsua_acc_modify()` when updating account's route_set.

Account's route set is a list that consists of:  global proxies + account proxies + service route learnt from `update_service_route()`.
But when modifying account, the learnt service route is never taken into account, causing incorrect list size calculation, and hence incorrect element removal.

The fixes in this PR:
- Learnt route headers is only removed upon registrar change. Account proxy is only modified upon change.
- Remove global outbound proxy modification.
 The current implementation seems to be buggy and not working as intended for the following reasons:
  - there's no API to modify the global proxy,
  - even if app modifies pjsua_var.ua_cfg.outbound_proxy internally (such as by including pjsua_internal.h), the route set is built from the list pjsua_var.outbound_proxy, which actually has different content and is never updated.
- Add new list APIs pj_list_insert_list_before/after().

Thank you to @andreas-wehrmann for the report and fix recommendation.

### DIFF
--- a/pjlib/include/pj/list.h
+++ b/pjlib/include/pj/list.h
@@ -171,6 +171,27 @@ PJ_IDECL(void) pj_list_insert_nodes_after(pj_list_type *lst,
 
 
 /**
+ * Insert the node to the list before the specified element position.
+ *
+ * @param pos   The element to which the node will be inserted before.
+ * @param lst   The list to be inserted.
+ */
+PJ_IDECL(void) pj_list_insert_list_before(pj_list_type *pos,
+                                          pj_list_type *lst);
+
+
+/**
+ * Insert the node to the list after the specified element position.
+ *
+ * @param pos   The element in the list which will precede the inserted
+ *              list.
+ * @param lst   The list to be inserted.
+ */
+PJ_IDECL(void) pj_list_insert_list_after(pj_list_type *pos,
+                                         pj_list_type *lst);
+
+
+/**
  * Remove elements from the source list, and insert them to the destination
  * list. The elements of the source list will occupy the
  * front elements of the target list. Note that the node pointed by \a list2
--- a/pjlib/include/pj/list.h
+++ b/pjlib/include/pj/list.h
@@ -171,7 +171,7 @@ PJ_IDECL(void) pj_list_insert_nodes_after(pj_list_type *lst,
 
 
 /**
- * Insert the node to the list before the specified element position.
+ * Insert a list to another list before the specified element position.
  *
  * @param pos   The element to which the node will be inserted before.
  * @param lst   The list to be inserted.
@@ -181,7 +181,7 @@ PJ_IDECL(void) pj_list_insert_list_before(pj_list_type *pos,
 
 
 /**
- * Insert the node to the list after the specified element position.
+ * Insert a list to another list after the specified element position.
  *
  * @param pos   The element in the list which will precede the inserted
  *              list.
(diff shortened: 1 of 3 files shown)


nanangizz (commented):
This change seems to drop the existing memory optimization, e.g: currently (shallow) cloning global proxy will only occur if global proxy is changed & (deep) cloning account proxy only occur if account proxy is changed, now both clonings will always occur when whichever proxy setting is changed or reg URI is changed. If some app for some reason modifies registrar URI regularly, this may cause unnecessary/avoidable "memory leak" (as account pool is never reset?).

If the memory optimization can be maintained, it will be ideal, but the new behavior may also be acceptable considering the memory leak scenario seems to be a rare scenario.

sauwming:
Changes in the latest commit:
- Learnt route headers is only removed upon registrar change. Account proxy is only modified upon change.
- Remove global outbound proxy modification.
The current implementation seems to be buggy and not working as intended for the following reasons:
  - there's no API to modify the global proxy,
  - even if app modifies `pjsua_var.ua_cfg.outbound_proxy` internally (such as by including `pjsua_internal.h`), the route set is built from the list `pjsua_var.outbound_proxy`, which actually has different content and is never updated.
- Add new list APIs `pj_list_insert_list_before/after()`. The current APIs `pj_list_insert_nodes_before/after()` seem to be useless since our list typically has an empty element as its head.

pjlib/include/pj/list_i.h
```diff
@@ -54,6 +54,22 @@ PJ_IDEF(void) pj_list_insert_nodes_before(pj_list_type *pos, pj_list_type *lst)
...
 }
 
+PJ_IDEF(void) pj_list_insert_list_after(pj_list_type *pos, pj_list_type *lst)
+{
+    if (!pj_list_empty(lst)) {
+        pj_list *lst_last = (pj_list *) ((pj_list*)lst)->prev;
+        pj_list *pos_next = (pj_list *) ((pj_list*)pos)->next;
+
+        pj_link_node(pos, (pj_list *) ((pj_list*)lst)->next);
+        pj_link_node(lst_last, pos_next);
```
nanangizz, line 64: reset/init the list after inserted, to avoid nodes/elements pointed by two different lists?

pjlib/include/pj/list.h
```diff
@@ -170,6 +170,27 @@ PJ_IDECL(void) pj_list_insert_nodes_after(pj_list_type *lst,
                                           pj_list_type *nodes);
 
 
+/**
+ * Insert the node to the list before the specified element position.
```
nanangizz, line 174: docs: insert `list` or `node`?

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -370,7 +370,9 @@ static pj_status_t initialize_acc(unsigned acc_id)
         }
         pj_list_push_back(&acc->route_set, r);
     }
-
+       pj_list_push_back(&acc->route_set, pjsip_hdr_shallow_clone(acc->pool, pjsua_var.outbound_proxy.next));
+       pj_list_push_back(&acc->route_set, pjsip_hdr_shallow_clone(acc->pool, pjsua_var.outbound_proxy.prev));
+
```
nanangizz, line 375: Adding first & last global route to acc `route_set` list, what for? Perhaps also add some comments?

pjlib/include/pj/list.h
```diff
@@ -170,6 +170,27 @@ PJ_IDECL(void) pj_list_insert_nodes_after(pj_list_type *lst,
...
+ *
+ * @param pos   The element to which the node will be inserted before.
+ * @param lst   The list to be inserted.
+ */
+PJ_IDECL(void) pj_list_insert_list_before(pj_list_type *pos,
+                                          pj_list_type *lst);
+
+
+/**
+ * Insert the node to the list after the specified element position.
```
nanangizz, line 184: docs: similar to above.

nanangizz (approved):

PR #3841, opened 2024-01-25
Add Metal renderer for Mac and iOS

sauwming:
The current OpenGL ES renderer has been deprecated for iOS since iOS 12.0
(https://developer.apple.com/documentation/opengles).

Thus we need to migrate to the new Metal API. The advantage of the new API is that it's also available for Mac OS (we use APIs that's supported for 10.14+ (https://developer.apple.com/documentation/metal/using_metal_to_draw_a_view_s_contents?language=objc)). This makes SDL no longer required for Mac and we can use the native Metal renderer instead.

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -712,6 +712,7 @@ ac_pjmedia_video_has_qt
 ac_darwin_cflags
 ac_pjmedia_video_has_ios_opengl
 ac_pjmedia_video_has_vtoolbox
+ac_pjmedia_video_has_metal
 ac_pjmedia_video_has_darwin
 ac_dshow_ldflags
 ac_dshow_cflags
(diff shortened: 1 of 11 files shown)


sauwming:
The CI is currently failing with "Illegal instruction: 4". It looks like the CI machine doesn't support the Metal API. Let me double check this.

sauwming:
Fixed it in the latest commit:
- NSWindow is automatically released when closed (https://developer.apple.com/documentation/appkit/nswindow/1419062-isreleasedwhenclosed)
- Fixed video device test incorrect loop (the renderer search loop only starts from `i+1`, which will fail if renderers are listed first).

nanangizz (approved):

trengginas (approved):

sbarcelona11:
@sauwming I can't see these changes in the latest library version. Why? 
The branch is removed. Can you upload it again?

PR #3845, opened 2024-01-29
Activate Bitrise iOS build test

sauwming:


sauwming:
Done.
https://github.com/pjsip/pjproject/pull/3845/checks?check_run_id=20967335964

Currently, (I think) our Bitrise plan only allows one build summary, so we can only display either iOS or Android. Both are still being test-built though, it's just in Github, we will see only one.

nanangizz:
Android build perhaps can be done in Github itself, haven't checked much but the runner seems to have Android NDK, see [here](https://github.com/actions/runner-images?tab=readme-ov-file#support-policy).

sauwming:
True, I believe both Android and iOS builds can be done on Github, but since we're going to try until device testing stage, we need to use Bitrise.

nanangizz (approved):

PR #3754, opened 2023-11-01
Implement Dialog Event subscription (client only)

goodicus:
Implement Dialog Event subscription (client only) for dialog monitoring purpose, which can be useful for features such as BLF (Busy Lamp Field) that displays the state of the calls associated with the Sip extension.

The most important RFCs related to this are:
- 3856 - A Presence Event Package for Session Initiation Protocol (SIP)
- 4235 - An INVITE-Initiated Dialog Event Package for the Session Initialisation Protocol (SIP)

PR #3846, opened 2024-01-30
Fixed dialog event Coverity warnings

sauwming:
Re #3754 , Coverity yields a couple of warnings:
```
>>>     CID 1533986:  Concurrent data access violations  (MISSING_LOCK)
>>>     Accessing "dlgev->status" without holding lock "pj_mutex_t.mutex". Elsewhere, "pjsip_dlg_event.status" is written to with "pj_mutex_t.mutex" held 1 out of 1 times.
567             for (i=0; i<dlgev->status.info_cnt; ++i) {
568                 dlgev->[status.info](http://status.info/)[i].dialog_node = NULL;
```

```
353             info->dialog_info_state.ptr = info->buf_ + total;
>>>     CID 1533985:    (OVERRUN)
>>>     Overrunning array of 1024 bytes at byte offset 1024 by dereferencing pointer "info->dialog_info_state.ptr".
```

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_pres.c
+++ b/pjsip/src/pjsua-lib/pjsua_pres.c
@@ -349,7 +349,9 @@ pjsua_buddy_get_dlg_event_info( pjsua_buddy_id buddy_id,
     pj_strncpy(&info->uri, &buddy->uri, sizeof(info->buf_)-total);
     total += info->uri.slen;
 
-    if (buddy->dlg_ev_status.info[0].dialog_info_state.slen > 0) {
+    if (buddy->dlg_ev_status.info[0].dialog_info_state.slen > 0 &&
+        total < sizeof(info->buf_))
+    {
         info->dialog_info_state.ptr = info->buf_ + total;
         pj_strncpy(&info->dialog_info_state,
             &buddy->dlg_ev_status.info[0].dialog_info_state,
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsua-lib/pjsua_pres.c
```diff
@@ -349,128 +349,142 @@ pjsua_buddy_get_dlg_event_info( pjsua_buddy_id buddy_id,
...
+            info->dialog_duration.ptr = info->buf_ + total;
+            pj_strncpy(&info->dialog_duration,
+                       &buddy->dlg_ev_status.info[0].dialog_duration,
+                       buddy->dlg_ev_status.info[0].dialog_duration.slen);
+            total += info->dialog_duration.slen;
+        }
+
+        if (buddy->dlg_ev_status.info[0].local_identity.slen > 0 &&
+            total < sizeof(info->buf_))
+        {
```
nanangizz, line 431: - Raising assertion and/or printing log when buffer is insufficient may be useful for developer?
- The checks & string copy ops seem similar, perhaps a helper function or a macro can improve readibility & correctness, something like:
```
#define COPY_BUF(field)
  if (status->field.slen > 0) {
    if (total + status->field.slen < size(info->buf_)) {
      info->field.ptr = info->buf_ + total;
      strncpy(..);
      total += field.slen;
    } else {
      PJ_LOG(..);
      pj_assert(..);
    }
  }
```

sauwming: Great idea!
Now it becomes much more compact.

I don't think there should be an assertion there though, since if remote sends something that's larger than the buffer size, we don't want to assert.

nanangizz (approved):

trengginas (approved):

PR #3847, opened 2024-01-30
Prevent data race in PJSIP presence

sauwming:
Add mutex to prevent data race in PJSIP presence, as per the discussion in #3754.

Also in this PR (https://github.com/pjsip/pjproject/pull/3754#discussion_r1467315047):
Moved `_is_valid` flag to presence struct, and move `_is_valid` checking in `pres_process_rx_notify()`.

### DIFF
--- a/pjsip/src/pjsip-simple/presence.c
+++ b/pjsip/src/pjsip-simple/presence.c
@@ -83,7 +83,9 @@ struct pjsip_pres
     pjsip_pres_status    status;        /**< Presence status.               */
     pj_pool_t           *tmp_pool;      /**< Pool for tmp_status            */
     pjsip_pres_status    tmp_status;    /**< Temp, before NOTIFY is answred.*/
+    pj_bool_t            is_ts_valid;   /**< Is tmp_status valid?           */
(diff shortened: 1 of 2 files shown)


nanangizz (approved):

pjsip/src/pjsip-simple/presence.c
```diff
@@ -389,14 +396,20 @@ PJ_DEF(pj_status_t) pjsip_pres_get_status( pjsip_evsub *sub,
     pres = (pjsip_pres*) pjsip_evsub_get_mod_data(sub, mod_presence.id);
     PJ_ASSERT_RETURN(pres!=NULL, PJSIP_SIMPLE_ENOPRESENCE);
 
-    if (pres->tmp_status._is_valid) {
+    if (pres->mutex)
+        pj_mutex_lock(pres->mutex);
+
+    if (pres->is_ts_valid) {
         PJ_ASSERT_RETURN(pres->tmp_pool!=NULL, PJSIP_SIMPLE_ENOPRESENCE);
```
trengginas, line 403: we should unlock when assert is disabled?

trengginas (approved):

PR #3853, opened 2024-02-06
Miscellaneous fixes

nanangizz:
Place for various minor fixes/enhancements, e.g: compile warning fixes, typos.


sauwming:
For a placeholder, perhaps we can create an issue instead.
So the issue will remain open, and separate PRs which link to the issue can be freely merged and closed, instead of having to wait until pre-release.

nanangizz:
> For a placeholder, perhaps we can create an issue instead. So the issue will remain open, and separate PRs which link to the issue can be freely merged and closed, instead of having to wait until pre-release.

Good idea! Also perhaps we can use a label like "misc/minor-fix" instead of creating an misc-issue for every milestone?

sauwming:
Ok, yes that's better so there's no open outstanding issue.

sauwming (approved):

PR #3854, opened 2024-02-07
Fix bad access in cancelling timer

nanangizz:
Reported that in a valgrind test when `PJ_POOL_DEBUG` is set, PJSIP transaction accesses to bad memory in cancelling a timer entry:
```
Invalid read of size 8
    at 0x49BA530: cancel_timer (in /usr/local/lib64/libpj.so.2)
    by 0x48E49B5: tsx_set_state (in /usr/local/lib64/libpjsip.so.2)
```
After investigation, it looks like that when cancelling a timer entry that is never scheduled (i.e: the entry has `_timer_id == -1`), it may cause the timer to access bad location. This does not seem to occur when `PJ_POOL_DEBUG` is not set.

Thanks Taisto Qvist for the report.

### DIFF
--- a/pjlib/src/pj/timer.c
+++ b/pjlib/src/pj/timer.c
@@ -534,11 +534,14 @@ static int cancel( pj_timer_heap_t *ht,
 
     PJ_CHECK_STACK();
 
-    // Check to see if the timer_id is out of range
+    // Check to see if the timer_id is out of range.
+    // Moved to cancel_timer() as it needs to validate _timer_id earlier
+    /*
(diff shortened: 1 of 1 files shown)


sauwming (approved):
I suppose this also only happens if PJ_TIMER_DEBUG is enabled (which is the default)?

nanangizz:
> I suppose this also only happens if PJ_TIMER_DEBUG is enabled (which is the default)?

Perhaps you mean `PJ_TIMER_USE_COPY`, which also by default is enabled. `GET_TIMER()` macro involves access to `_timer_id` when `PJ_TIMER_USE_COPY` is set, so yes it only happens in this condition.

trengginas (approved):

PR #3857, opened 2024-02-08
Header editing feature

goodicus:
Allow headers edition

The purpose of this PR is to edit headers in outgoing messages.

### DIFF
--- a/pjlib/src/pj/timer.c
+++ b/pjlib/src/pj/timer.c
@@ -534,11 +534,14 @@ static int cancel( pj_timer_heap_t *ht,
 
     PJ_CHECK_STACK();
 
-    // Check to see if the timer_id is out of range
+    // Check to see if the timer_id is out of range.
+    // Moved to cancel_timer() as it needs to validate _timer_id earlier
+    /*
     if (entry->_timer_id < 1 || (pj_size_t)entry->_timer_id >= ht->max_size) {
         entry->_timer_id = -1;
         return 0;
     }
+    */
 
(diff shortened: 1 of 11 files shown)


andreas-wehrmann:
I'm not entirely sure what this PR is supposed to do.
I mean, if you want the ability to manipulate incoming/outgoing messages, this can be implemented via modules.
Wouldn't this be the cleaner solution? To have a separate module for header manipulation only?
That way you'd also keep responsibilities nicely separated and can avoid changes to PJSUA itself.

But maybe I'm just not understanding something.
This PR seems to introduce a significant number of changes;
Can you elaborate on what problem you're trying to solve?

goodicus:
Hi, @andreas-wehrmann @sauwming!
 I would like to introduce one more useful feature to PjSUA tool. It's a "header modification feature".
It allows to extend an original and to emulate non-standard PJSUA behaviour, like:

- add an optional header to outgoing messages or an optional tag in a header field (many cases during registration);
- remove necessary and optional headers in outgoing messages;
- substitute headers by another value ( "User-Agent: PJSUA v2.14" -> "User-Agent: Yealink SIP-T42S").

sauwming:
We already have such features via configs such as `pjsua_acc_config.reg_hdr_list`, `pjsua_msg_data.hdr_list`, `pjsua_config.user_agent`. While it's true that not all headers can be modified using those configs, but it has proven to be sufficient for most use cases of our users. And if custom modification is required, as @andreas-wehrmann  stated above, user can implement their own pjsip module.

PR #3863, opened 2024-02-17
Add extra logging for dummy resample imp

deeagle001:
When I first started to use pjproject, I disabled everything I assumed I did not need. This mistakenly included `--disable-resample`. I thought it would disable the resampling functionality as is, but I got a mysterious error about the channel can not be created and raised `PJ_EINVALIDOP`. After a day of setting up the debug environment and debugging the issue, it turned out that the resampling functionality is mandatory for SIP audio calls, but applying `--disable-resample` will switch to a dummy resample implementation, which just returns the error code mentioned above. The first channel creation was just a victim of this fact, and the real problem (What operation was invalid?) was already lost at this high level.

This PR tries to help future users not to fall into the same trap I did, by making an error log in the exact place with the proper message.

The PR also contains some minor, non-functional fixes like a typo fix and comment alignment fixes.

### DIFF
--- a/pjmedia/src/pjmedia/resample_resample.c
+++ b/pjmedia/src/pjmedia/resample_resample.c
@@ -1,4 +1,4 @@
-/*
+/* 
  * Copyright (C) 2008-2011 Teluu Inc. (http://www.teluu.com)
  * Copyright (C) 2003-2008 Benny Prijono <benny@prijono.org>
  *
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia/resample_resample.c
```diff
@@ -320,25 +320,27 @@ PJ_DEF(pj_status_t) pjmedia_resample_create( pj_pool_t *pool,
     PJ_UNUSED_ARG(samples_per_frame);
     PJ_UNUSED_ARG(p_resample);
 
+    PJ_LOG(3, (THIS_FILE, "Resampler creation cancelled due to missing"
+                          "resample implementation");
```
sauwming, line 324: missing closing bracket

sauwming (approved):

nanangizz (approved):
Honestly trailing whitespace removal is quite annoying for code reviewing (I know some IDEs have this feature enabled by default). Hopefully the commit can be separated from the bug fix/enhancement patch.

PR #3866, opened 2024-02-20
Add an option to disable Message Composition Indication feature (MCI)

goodicus:
The Message Composition Indication feature (RFC-3994) is always enabled currently. And it leads to sending massive messages while typing.
The PR switches off the feature by default and adds an option to a configuration file to enable the feature
```
--set-mci
```


nanangizz:
Usually we don't change default behaviors except for a strong reason, e.g: security issues. So what if the new option is to disable the MCI instead?

goodicus:
@nanangizz I got your point and will rework the implementation.

goodicus:
Add an option to disable the default pjsua behaviour to not send Message Composition Information (RFC-3994).
The option name is "--no-mci".

nanangizz (approved):

sauwming (approved):

PR #3867, opened 2024-02-21
Add TLS/SSL backend: Windows Schannel

nanangizz:
### Introduction

Currently the library has implemented TLS/SSL using various backends such as OpenSSL, BoringSSL, GnuTLS, Apple Network framework. This PR will add another backend: [Windows SSP Schannel](https://learn.microsoft.com/en-us/windows-server/security/tls/tls-ssl-schannel-ssp-overview), which is a native TLS/SSL security provider for Windows. As usual, this backend is implemented for the PJLIB [SSL socket API](https://docs.pjsip.org/en/latest/api/generated/pjlib/group/group__PJ__SSL__SOCK.html). 

### How to enable

Enable TLS/SSL and use Schannel backend implementation. In [`config_site.h`](https://docs.pjsip.org/en/latest/get-started/guidelines-development.html#config-site-h), specify:
```
#define PJ_HAS_SSL_SOCK 1
#define PJ_SSL_SOCK_IMP PJ_SSL_SOCK_IMP_SCHANNEL
```
For app, generally this should be similar to using the other backends (see [here](https://docs.pjsip.org/en/latest/specific-guides/security/ssl.html#using-sip-tls-transport)), except for few things:
- TLS certificates are looked up from the [Certificate Store](https://learn.microsoft.com/en-us/dotnet/framework/wcf/feature-details/working-with-certificates#certificate-stores), specifically the **current user** store and the **local machine** store. Application can specify the certificate lookup criteria via `pj_ssl_cert_lookup_criteria`, which currently supports lookup by subject, by fingerprint, or by friendly name. Currently certificates specified via file, path, or buffer (i.e: `pj_ssl_cert_load_from_files/files2()` and `pj_ssl_cert_load_from_buffer()`) will be ignored.
- For manual verification, it will lookup the certificate chain in the **current user** store only in folder **Trusted Root Certification Authorities**.
- Enabling/disabling cipher suite and curves should be configured via OS settings (global scope), see [here](https://learn.microsoft.com/en-us/windows-server/security/tls/manage-tls).

### Limitations

- Currently this does not cover DTLS-SRTP key negotiation for SRTP media transport.
- Specifying enabled cipher suites and curves cannot be done via PJLIB `pj_ssl_sock_param.ciphers/curves` or PJSIP/PJSUA `pjsip_tls_setting.ciphers/curves`. This is because the Schannel API for enable/disable/prioritize cipher suites & curves seems to be for global scope instead of per TLS session (e.g: see [here](https://learn.microsoft.com/en-us/windows/win32/secauthn/prioritizing-schannel-cipher-suites)).

### Notes
- TLS 1.0 and TLS 1.1 will be deprecated soon, see [here](https://learn.microsoft.com/en-us/windows/win32/secauthn/tls-10-11-deprecation-in-windows).
- TLS 1.3 is supported since Windows 11 or Windows Server 2022, more info [here](https://learn.microsoft.com/en-us/windows/win32/secauthn/protocols-in-tls-ssl--schannel-ssp-).

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -182,21 +182,6 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@master
-    - name: get openssl
-      run: Invoke-WebRequest -Uri "https://github.com/pjsip/third_party_libs/raw/main/openssl-1.1.1s-win.zip" -OutFile ".\openssl.zip"
-      shell: powershell
-    - name: expand openssl
-      run: |
-        Expand-Archive -LiteralPath .\openssl.zip -DestinationPath .; pwd
-        cd openssl_build
-        Add-Content ..\openssl_dir.txt $pwd.Path
-      shell: powershell
-    - name: check openssl folder
-      run: |
-        set /P OPENSSL_DIR=<openssl_dir.txt
-        dir "%OPENSSL_DIR%\include"
-        dir "%OPENSSL_DIR%\lib"
-      shell: cmd
     - name: get vpx
       run: Invoke-WebRequest -Uri "https://github.com/pjsip/third_party_libs/raw/main/vpx-1.12-win.zip" -Outfile "vpx.zip"
       shell: powershell
@@ -231,6 +216,8 @@ jobs:
       run: |
         cd pjlib/include/pj; cp config_site_test.h config_site.h
         Add-Content config_site.h "#define PJ_HAS_SSL_SOCK 1"
+        Add-Content config_site.h "#define PJ_SSL_SOCK_IMP PJ_SSL_SOCK_IMP_SCHANNEL"
+        Add-Content config_site.h "#undef PJMEDIA_SRTP_HAS_DTLS"
         Add-Content config_site.h "#define PJMEDIA_HAS_VIDEO 1"
         Add-Content config_site.h "#define PJMEDIA_VIDEO_DEV_HAS_DSHOW 1"
         Add-Content config_site.h "#define PJMEDIA_HAS_LIBYUV 1"
@@ -258,7 +245,7 @@ jobs:
         set /P SDL_DIR=<sdl_dir.txt
         cd tests/pjsua/tools
         set INCLUDE=%INCLUDE%;%OPENSSL_DIR%\include;%VPX_DIR%\include;%SDL_DIR%\include
-        set LIB=%LIB%;%OPENSSL_DIR%\lib;%VPX_DIR%\lib;%SDL_DIR%\lib\x86
+        set LIB=%LIB%;%VPX_DIR%\lib;%SDL_DIR%\lib\x86
         call "%PROGRAMFILES%\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat"
         msbuild cmp_wav.vcxproj /p:PlatformToolset=v143 /p:Configuration=Release /p:Platform=win32 /p:UseEnv=true
       shell: cmd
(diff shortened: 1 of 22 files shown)


pjsip/include/pjsua2/siptypes.hpp
```diff
@@ -174,6 +174,26 @@ struct TlsConfig : public PersistentObject
      */
     string              privKeyBuf;
 
+    /**
+     * Lookup certificate from OS certificate store, this setting will
+     * specify the field type to lookup.
+     *
+     * Currently only TLS backend Windows Schannel support this and this
```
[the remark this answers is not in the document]
nanangizz: Right, moved most of the text to `pj_ssl_cert_load_from_store()` and add link to `pj_ssl_cert_load_from_store()` in those places (PJSUA2, TURN & TLS transport).

pjlib/include/pj/ssl_sock.h
```diff
@@ -273,6 +331,25 @@ PJ_DECL(pj_status_t) pj_ssl_cert_load_from_buffer(pj_pool_t *pool,
                                         const pj_str_t *privkey_pass,
                                         pj_ssl_cert_t **p_cert);
 
+/**
+ * Create credential from OS certificate store, this function will lookup
+ * certificate using the specified criterias.
+ *
+ * Currently this is used by Windows Schannel backend only, it will lookup
+ * in the Current User store first, if not found it will lookup in the
+ * Local Machine store.
```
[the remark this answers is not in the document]
nanangizz: Right, had a thought about this a bit, that's why `pj_ssl_cert_load_from_store()` uses an 'expandable' param `pj_ssl_cert_lookup_criteria` for future use.

Nevertheless, tried to do a bit research for Android. From [here](https://developer.android.com/privacy-and-security/keystore#ExtractionPrevention), it seems that a cert (& its private key) cannot be extracted out to be used by third-party security lib (such as OpenSSL), so the keystore seems to be for native SSL backend only. And from [here](https://android-developers.googleblog.com/2012/03/unifying-key-store-access-in-ics.html), it looks like accessing cert in the keystore system is done via UI prompt dialog for the user selecting the cert to use, then application can assign an "alias" (perhaps some kind of "friendly-name" on Windows) for the cert for future use (so app don't need to show the UI prompt dialog again). So if in the future we have a native Android SSL backend, perhaps `pj_ssl_cert_load_from_store()` can be used to lookup TLS cert using the "alias". The `pj_ssl_cert_lookup_criteria` & `pj_ssl_cert_load_from_store()` may not be able to handle any access auth though, which can be password, PIN, biometrics, etc.

sauwming (approved):
Include it in CI test:
Perhaps can replace the GnuTLS, or combine it with `windows-with-video-libvpx-unit-test-1`
Running the pjlib-test only should suffice.

.github/workflows/ci-win.yml
```diff
@@ -182,21 +182,6 @@ jobs:
     runs-on: windows-latest
```
sauwming, line 182: Nice, it looks like the test passed.

Add comment and/or change the workflow name as well so we remember that this unit test includes Schannel SSL backend.

pjlib/src/pj/ssl_sock_schannel.c
```diff
@@ -0,0 +1,1631 @@
...
+        goto on_return;
+
+    /* Initialize output circular buffer */
+    status = circ_init(pf, &ssock->circ_buf_output, write_cap);
+    if (status != PJ_SUCCESS)
+        goto on_return;
+
+    /* Initialize decrypted circular buffer */
+    status = circ_init(pf, &sch_ssock->decrypted_buf, read_cap);
+    if (status != PJ_SUCCESS)
```
trengginas, line 346: Can be deleted here?

nanangizz: Describe a bit more please, e.g: what can be deleted, what is the potential problem.

trengginas: There's no need to check `status` since the next line is the `on_return` label.

nanangizz: Oh, I didn't see the label. IIRC it is on purpose, it relatively gives better readibility & avoid bug when new code is added below that. Hopefully compiler optimization can "delete" it (haven't checked further, but something like [this](https://stackoverflow.com/questions/15301711/stop-goto-optimization-g))?

PR #3868, opened 2024-02-22
Video CI test on iOS device

sauwming:
Video test on iOS device using Bitrise and Firebase.
`Device testing with Firebase` https://devcenter.bitrise.io/en/testing/device-testing-with-firebase.html

Test setup:
- Using `ipjsua-swift` sample app.
- Video capture: colorbar. Video renderer: Metal

Test scenario:
- Local video call utilising a single iOS device only (video call initiated to local account `sip:localhost:5080`).
   **UI test (Automated)**: Simulate triple tap on `Destination:` text field, type in the text `sip:localhost:5080`. Verify using `XCTAssert` that the text is correct. If yes, click the button `Make call`.
- The device will only display one video, so only one party will transmit the video, while the other one will receive and show the video.
   **Verification**: Make sure that the incoming video view is not `nil` (i.e. check if `pjsua_vid_win_info.hwnd.info.ios.window != NULL`).


nanangizz:
Add some description about the test scenarios & scope? E.g: on video call, check if video/colorbar appears/disappears at both sides.

nanangizz (approved):

PR #3875, opened 2024-03-02
Allow to record only a part of a call

goodicus:


sauwming:
Not sure what the intention is here, but I'm afraid the current implementation will confuse, rather than help, users.
A few notes:
- We already have `--rec-file=file` option, which will do the same thing.
- We can't specify the filename here.
- Start/stop is not quite correct, since here we only create/destroy.

goodicus:
@sauwming This feature allows you to record only a part of a call. It is used with "-rec-file=file" option.

PR #3889, opened 2024-03-18
PJ_LOG_HAS_FILENAME

oliverepper:
When using the filename inside the project as the sender for the pj_log function it get's truncated from the left adding '...' at the beginning to indicate.

### DIFF
--- a/pjlib/include/pj/log.h
+++ b/pjlib/include/pj/log.h
@@ -82,7 +82,8 @@ enum pj_log_decoration
     PJ_LOG_HAS_LEVEL_TEXT = 2048, /**< Include level text string [no]         */
     PJ_LOG_HAS_THREAD_ID  = 4096, /**< Include thread identification [no]     */
     PJ_LOG_HAS_THREAD_SWC = 8192, /**< Add mark when thread has switched [yes]*/
-    PJ_LOG_HAS_INDENT     =16384  /**< Indentation. Say yes! [yes]            */
+    PJ_LOG_HAS_INDENT     =16384, /**< Indentation. Say yes! [yes]            */
+    PJ_LOG_HAS_FILENAME   =32768, /**< Like HAS_SENDER but truncated left     */
 };
 
 /**
--- a/pjlib/src/pj/log.c
+++ b/pjlib/src/pj/log.c
@@ -416,6 +416,24 @@ PJ_DEF(void) pj_log( const char *sender, int level,
                 *pre++ = *sender++;
         }
     }
+    if (log_decor & PJ_LOG_HAS_FILENAME) {
+        enum { SENDER_WIDTH = PJ_LOG_SENDER_WIDTH };
+        pj_size_t sender_len = strlen(sender);
+        if (pre!=log_buffer) *pre++ = ' ';
+        if (sender_len <= SENDER_WIDTH) {
+            while (sender_len < SENDER_WIDTH)
+                *pre++ = ' ', ++sender_len;
+            while (*sender)
+                *pre++ = *sender++;
+        } else {
+            int i;
+            for (i=0; i<3; ++i)
+                *pre++ = '.';
+            char *truncated_begin = sender + sender_len - SENDER_WIDTH + 3;
+            for (i=0; i<SENDER_WIDTH -3; ++i)
+                *pre++ = *truncated_begin++;
+        }
+    }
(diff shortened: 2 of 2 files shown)


oliverepper:
I like to use a project relative filename as sender. But since they are often times too long and the important information is at the end I added this.

Bildschirmfoto 2024-03-18 um 16 56 36

nanangizz:
I think the proper approach should be lengthening the macro setting [`PJ_LOG_SENDER_WIDTH`](https://docs.pjsip.org/en/latest/api/generated/pjlib/group/group__pj__config.html#c.PJ_LOG_SENDER_WIDTH), this way any type of sender won't be (too much) truncated. As you are aware already, sender can also be objects such as transaction & mutex which use prefix 'tsx' or 'mtx' and generally missing the prefix info (due to truncation) may significantly increase debugging difficulties.

oliverepper:
You're probably right. I wasn't aware of the prefixes and that this would kill them. I could check for them, though? The problem will go away once Apple updates their clang. The problem I had is that `__FILE__` gives the full path to a file with Apple clang (homebrew is more recent and gives only the name). I find it annoying to `#define` a variable for the current filename just to get proper logging. These things are 100% forgotten to update, once a file gets refactored/renamed and then they make hard to debug situations.

I could update the logic to not do any processing when the sender does not contain path separators?

sauwming:
> I could update the logic to not do any processing when the sender does not contain path separators?

Sounds like a decent idea.
I tested it here and I can understand that it's annoying for the log to be printed something like this:
`15:56:49.784 /home/user/pjsip/pjsua-lib/p  test log`
Would be even worse if the pathnames are longer.

Perhaps a flag like `PJ_LOG_HAS_SENDER_FILECHECK` may help? So if it's a file (contains file separator either '/' or '\'), SENDER will be truncated left, otherwise it's the current behavior.

What do you think, @nanangizz ?

oliverepper:
Keep in mind, that the problem goes away when Apple updates their clang. In upstream clang the `__FILE__` macro gives only the filename, not the complete path. The flag could be useful anyways. It could still be useful to use a left-truncated project relative path as the sender.

nanangizz:
Isn't it a bit too much for a logger (i.e: scanning sender string for '/' on each log)? If app wants to "normalize" sender string, we'll better leave it to app (e.g: app creates a function to do that which may be called it each time printting a log or once in some initialization, or via a compile-time macro if possible). So basically it calls the logger with the sender already preformated/normalized.

sauwming:
Makes sense. I guess we should not change anything since the problem will eventually disappear in the future, and doing additional processing will be highly inefficient for frequent operations such as logging.

I believe we can close this then? Let me know if you think otherwise.

oliverepper:
I think it should be closed, too. @nanangizz is right. Logging should be kept clean. Might be alright to provide the option of right, or left truncation but every more complex operation should be kept out of this path, I guess. I still don't like the `#define` of the current filename, though. Maybe autoconf can be used to fill in the filenames?

oliverepper:
On second thought: Apple should update their clang and everything is fine. Compiler should provide a macro for the current filename, as it does upstream.

nanangizz:
Just checked [this](https://stackoverflow.com/questions/8487986/file-macro-shows-full-path), looks like there is no simple & portable compile-time macro for getting the file name only.

As you may be aware, we decided to use simple plain `THIS_FILE` in many files for logging.

Here is an idea that may be portable and relatively not so much overhead. In a C header file, define a macro named `THIS_FILE2` for example, then every C source file can include the header and use `THIS_FILE2` macro as log sender, have not tested myself though.
```
static const char *fname;

#define THIS_FILE2 get_file_name()

inline const char *get_file_name()
{
  if (!fname)
    fname = strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__;

  return fname;
}
```

oliverepper:
```C++
#if defined(_WIN32) || defined(WIN32)
constexpr char pathSeparator = '\\';
#else
constexpr char pathSeparator = '/';
#endif

constexpr std::string_view filename(std::string_view maybe_full_path = __FILE__) {
    size_t last_slash = maybe_full_path.find_last_of(pathSeparator);
    if (last_slash != std::string_view::npos)
        return maybe_full_path.substr(last_slash + 1);
    return maybe_full_path;
}
```

PR #3893, opened 2024-03-21
Add some missing unlocks

xhit:
Add some missing unlocks in the code.

Untested in local, I found some playing with the library, If some are unnecessary we can to edit.

### DIFF
--- a/pjlib-util/src/pjlib-util/resolver.c
+++ b/pjlib-util/src/pjlib-util/resolver.c
@@ -1840,16 +1840,20 @@ PJ_DEF(pj_status_t) pj_dns_resolver_add_entry( pj_dns_resolver *resolver,
     pj_bzero(&key, sizeof(struct res_key));
     if (pkt->hdr.anscount) {
         /* Make sure name is not too long. */
-        PJ_ASSERT_RETURN(pkt->ans[0].name.slen < PJ_MAX_HOSTNAME, 
-                         PJ_ENAMETOOLONG);
+        if (pkt->ans[0].name.slen >= PJ_MAX_HOSTNAME){
+            pj_grp_lock_release(resolver->grp_lock);
+            return PJ_ENAMETOOLONG;
+        }
 
         init_res_key(&key, pkt->ans[0].type, &pkt->ans[0].name);
 
     } else {
         /* Make sure name is not too long. */
-        PJ_ASSERT_RETURN(pkt->q[0].name.slen < PJ_MAX_HOSTNAME, 
-                         PJ_ENAMETOOLONG);
-
+        if (pkt->q[0].name.slen >= PJ_MAX_HOSTNAME){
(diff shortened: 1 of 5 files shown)


pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1840,16 +1840,20 @@ PJ_DEF(pj_status_t) pj_dns_resolver_add_entry( pj_dns_resolver *resolver,
     pj_bzero(&key, sizeof(struct res_key));
     if (pkt->hdr.anscount) {
         /* Make sure name is not too long. */
-        PJ_ASSERT_RETURN(pkt->ans[0].name.slen < PJ_MAX_HOSTNAME, 
-                         PJ_ENAMETOOLONG);
+        if (pkt->ans[0].name.slen >= PJ_MAX_HOSTNAME){
+            pj_grp_lock_release(resolver->grp_lock);
+            return PJ_ENAMETOOLONG;
+        }
```
nanangizz, line 1846: I think we should maintain it as an assertion check, for example: 
```
PJ_ASSERT_ON_FAIL(pkt->ans[0].name.slen < PJ_MAX_HOSTNAME, 
                  { pj_grp_lock_release(resolver->grp_lock); return PJ_ENAMETOOLONG; });
```

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1840,16 +1840,20 @@ PJ_DEF(pj_status_t) pj_dns_resolver_add_entry( pj_dns_resolver *resolver,
...
     } else {
         /* Make sure name is not too long. */
-        PJ_ASSERT_RETURN(pkt->q[0].name.slen < PJ_MAX_HOSTNAME, 
-                         PJ_ENAMETOOLONG);
-
+        if (pkt->q[0].name.slen >= PJ_MAX_HOSTNAME){
+            pj_grp_lock_release(resolver->grp_lock);
+            return PJ_ENAMETOOLONG;
+        }
+
```
nanangizz, line 1856: same as above

nanangizz (commented):
Good catch, thanks.
Just a couple notes.

pjlib-util/src/pjlib-util/resolver.c
wosrediinanatour: Just an idea I head when seeing this PR: In other places a return is "deferred" in case of error:

```cpp
//...

if (error) 
   goto on_error;

//...
return;

on_error:
   //... some cleanup ...

return;
```

This is usually quite useful with locks in C since there is nothing like RAII.

xhit: done ee6beea6f4e6bd440d36b09155f8c3ab968d54ca

nanangizz (approved):

sauwming (approved):

PR #3896, opened 2024-03-22
Fix warnings for 32-bit compiler and misc fixes.

amubiera:
- speex_codec: Include speex/speex.h when codec is enabled.
- clock_thread: Fix for ‘ts.u64’ may be used uninitialized in this function.
- ice_session: Fix for format '%ld' expects argument of type 'long int', but argument 3 has type 'int'.
- pjsua_core: Fix for format '%ld' expects argument of type 'long int', but argument 3 has type 'int'.

### DIFF
--- a/pjmedia/src/pjmedia-codec/speex_codec.c
+++ b/pjmedia/src/pjmedia-codec/speex_codec.c
@@ -22,7 +22,6 @@
 #include <pjmedia/errno.h>
 #include <pjmedia/endpoint.h>
 #include <pjmedia/port.h>
-#include <speex/speex.h>
 #include <pj/assert.h>
 #include <pj/log.h>
 #include <pj/pool.h>
@@ -34,6 +33,7 @@
  */
 #if defined(PJMEDIA_HAS_SPEEX_CODEC) && PJMEDIA_HAS_SPEEX_CODEC!=0
 
+#include <speex/speex.h>
 
 #define THIS_FILE   "speex_codec.c"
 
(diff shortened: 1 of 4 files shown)


third_party/srtp/srtp/srtp.c
```diff
@@ -506,7 +506,7 @@ static srtp_err_status_t srtp_stream_clone(
     srtp_session_keys_t *session_keys = NULL;
     const srtp_session_keys_t *template_session_keys = NULL;
 
-    debug_print(mod_srtp, "cloning stream (SSRC: 0x%08x)", ntohl(ssrc));
+    debug_print(mod_srtp, "cloning stream (SSRC: 0x%08x)", (unsigned int)ntohl(ssrc));
```
nanangizz, line 509: We usually don't change third-party code. If it is relatively important (e.g: compile failure or major bug), we apply the change & add some notes in a separate text file, e.g: see [here](https://github.com/pjsip/pjproject/blob/master/third_party/build/yuv/Notes.txt). For compile-warnings, we usually add ignore warning settings, e.g: see [here](https://github.com/pjsip/pjproject/blob/a7c4d83807ceb25c5067c1dcb5d863becc7986c3/third_party/build/srtp/srtp_config.h#L144-L154).

amubiera: Ok. Removed the change for srtp.c and repushed.

nanangizz (approved):

sauwming (approved):

PR #3899, opened 2024-03-24
removed a bashism that prevented configuration

oliverepper:
To my understanding `+=` is a bashism that a real `sh` does not understand. In order to successfully configure with `--enable-libwebrtc-aec3` on FreeBSD I needed this.

P.S.: There are more bashisms in the aconfigure script, but this made the script fail, because the variable was empty.

P.P.S.: In order to build with `--enable-libwebrtc-aec3` on FreeBSD there are more things to do. Some types need to be corrected. `pid_t` is 4 bytes on FreeBSD, for example.

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -10362,7 +10362,11 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 		     *)
 			;;
 		  esac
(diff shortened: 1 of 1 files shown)


sauwming:
You need to modify `aconfigure.ac` instead, then generate `aconfigure` via `autoconf aconfigure.ac > aconfigure`.

oliverepper:
Of course, you're right. I've just copied this from my FreeBSD ports patch that I just applied to the resulting script. I will close this and make a new PR. Thanks.

PR #3913, opened 2024-04-04
Support Push Notification in iOS sample app

sauwming:
Provide an example on how to implement Push Notification using PushKit and CallKit frameworks in ipjsua sample app.

Also in this PR:
Add sample on how to call PJSIP functions from a PJLIB registered thread via timer.


sauwming:
> Is two/more incoming calls scenario handled, e.g:
> 
> * concurrently, user may want to choose which call to answer, or
> * another incoming call in the middle of an active call.

Done.

nanangizz (approved):
Looks okay to me.

Btw, the patch seems to also add sample code of how to call PJSIP functions from a PJLIB registered thread via timer, perhaps add desc about this too?

PR #3925, opened 2024-04-15
Add function to initialize MediaFormat audio & video

sauwming:
Currently, the object `MediaFormatAudio/Video` will be uninitalised upon creation (i.e. `type == PJMEDIA_TYPE_NONE`).

To initialise it, user needs to explicitly set its parent's `type` to `PJMEDIA_TYPE_AUDIO/VIDEO`, otherwise the object will still considered to be uninitialised.

For example:
```
VideoPreview prev(videoDevIdx);
VideoPreviewOpParam param = VideoPreviewOpParam();
param.format.id = PJMEDIA_FORMAT_I420;
param.format.width = 1080;
param.format.height = 1920;
param.format.fpsNum = 15;
param.format.fpsDenum = 1;
/* Despite all the above initialisations, param.format.type is still PJMEDIA_TYPE_NONE and hence, will just be ignored. */
prev.start(param);
```

### DIFF
--- a/pjsip-apps/src/samples/pjsua2_demo.cpp
+++ b/pjsip-apps/src/samples/pjsua2_demo.cpp
@@ -205,11 +205,7 @@ void MyCall::onCallMediaState(OnCallMediaStateParam &prm)
         med_port = new MyAudioMediaPort();
 
         MediaFormatAudio fmt;
-        fmt.type = PJMEDIA_TYPE_AUDIO;
-        fmt.clockRate = 16000;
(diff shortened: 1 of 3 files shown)


nanangizz (approved):
Notes:
- The default media type in the PR description is `PJMEDIA_TYPE_UNKNOWN` while it should be `PJMEDIA_TYPE_NONE` instead (they have different values)?
- Better also quickly scan our sample apps for possible bug caused by this behavior.

PR #3928, opened 2024-04-16
Asynchronous conference bridge operation

nanangizz:
In #3183, video conference bridge adopts asynchronous ports operations to handle non-uniform lock ordering issues. This PR integrates the same mechanism for audio conference bridge.

Beside avoiding deadlock, this change may improve performance a little bit as audio data processing in the conference bridge will be performed without holding a mutex.

There are some pending tasks (e.g: integrate this into audio switchboard).

### Potential backward compatibility issues (major):

1. Pool release immediately after port removal, consider this code:
    ```
    pool = pj_pool_create(...);
    pjmedia_tonegen_create(pool, ..., &tonegen_port);
    pjmedia_conf_add_port(conf, ..., tonegen_port, ..., &tonegen_port_id);
    ...
    pjmedia_conf_remove_port(conf, tonegen_port_id);
    pjmedia_port_destroy(tonegen_port);
    pj_pool_release(pool); /* <-- This is now a problem!
                                  It was fine with a synchronous pjmedia_conf_remove_port() */
    ```
    Note that the PJMEDIA tonegen port **does not have its own pool** (it allocates its own instances and all internal states using the supplied memory pool), so releasing the pool immediately after removing the port from conference bridge may cause crash as the port removal is now asynchronous and it may be still being accessed by the conference bridge.

    Unfortunately, the real removal operation will not invoke any callback, so application cannot be so sure the safe time for releasing the pool. FYI the actual removal is done by the conference bridge clock which usually is a sound device, so if the clock is running, adding `pj_thread_sleep(100)` before relasing pool may avoid the crash (note that clock interval is usually 20ms, considering possible audio device burst, 100ms delay is generally sufficient).

    Now all PJMEDIA ports have been updated to have their own pool, except bidirectional & stereo ports (currently not used). Any third-party PJMEDIA ports that are registered to the conference bridge will need to be made sure that they have their own pools.

2. PJMEDIA Stream pattern.
    PJMEDIA Stream exports a PJMEDIA port interface which can be queried using `pjmedia_stream_get_port()`. Unfortunately the port does not implements `pjmedia_port_destroy()`, it is destroyed using `pjmedia_stream_destroy()` instead, which will release the memory pool. Similar to above issue, if pjmedia_stream_destroy() is invoked immediately after `pjmedia_conf_remove_port()`, it may cause crash.

    Now PJMEDIA Stream has a group lock, so `pjmedia_stream_destroy()` will just decrement the reference counter, the real destroy will be done when the reference counter reaches zero.

3. Premature caching pool destroy.
    When all PJMEDIA ports have their own pools, there is still possibility that application uses its own pool factory for creating the ports and destroy the pool factory (e.g: invoking `pj_caching_pool_destroy()`) prematurely before the actual port removal is done. For example, in application shutdown, the conference bridge clock has been stopped so actual port removal can only be done later by `pjmedia_conf_destroy()/pjsua_destroy()`, but in between, application may destroy its pool factory which may cause crash.

    A possible solution is to delay the pool factory destroy to PJLIB deinitialization via `pj_atexit()`.

4. Premature pool release when pool created using `pjsua_pool_create()`.
    Similar to premature caching pool destroy above, any application pool that is used for creating PJMEDIA ports needs to be released after the conference bridge (PJSUA internal object) is destroyed, this can be accomplished by scheduling the pool release using `pjsip_endpt_atexit()` (don't use `pj_atexit()` as it will be too late, PJSUA destroys its caching pool factory before calling `pj_shutdown()`), e.g:
    ```
    static pj_pool_t *app_pool;

    static void app_cleanup(pjsip_endpoint *endpt) {
      PJ_UNUSED_ARG(endpt);
      pj_pool_safe_release(&app_pool);
    }

    int main() {
      app_pool = pjsua_pool_create(...);
      pjsip_endpt_atexit(pjsua_get_pjsip_endpt(), &app_cleanup);
      ...
    }
    ```


sauwming:
Since issue 2 is already handled, I assume it's no longer a problem (i.e. no longer causes backward compatibility)?

Which only leaves issue 1. I understand it will be quite a lot to change all existing pjmedia_port(s) to have its own pool, but since it's necessary for this feature, I vote to continue.

And if 1&2 are already resolved, the change in pjsua_app.c is no longer necessary, correct? (i.e. app will not need to change anything).

nanangizz:
> Since issue 2 is already handled, I assume it's no longer a problem (i.e. no longer causes backward compatibility)?
> 
> Which only leaves issue 1. I understand it will be quite a lot to change all existing pjmedia_port(s) to have its own pool, but since it's necessary for this feature, I vote to continue.

The backward compatibility issues cannot be completely avoided, e.g:
- if any custom/proprietary pjmedia ports are used, need to check/update them,
- applications may need to be adjusted to avoid issue 3 (pool factory premature destroy).

> 
> And if 1&2 are already resolved, the change in pjsua_app.c is no longer necessary, correct? (i.e. app will not need to change anything).

Yes for `pjsua_app.c`, other apps may need adjustment (as mentioned above).

sauwming:
Regarding issue 3, I don't think app is allowed to call `pj_caching_pool_destroy()` before `pjmedia_conf_destroy()` since the conference creation uses the pool, i.e. `pjmedia_conf_create(pool, ...)`?
Also, using `pj_atexit()` doesn't seem possible since pjsip can be restarted (instead of shutdown).

The important thing here seems to be that `pjmedia_conf_destroy()` must guarantee the completion of all pending async operations, if any.

nanangizz:
> Regarding issue 3, I don't think app is allowed to call `pj_caching_pool_destroy()` before `pjmedia_conf_destroy()` since the conference creation uses the pool, i.e. `pjmedia_conf_create(pool, ...)`? Also, using `pj_atexit()` doesn't seem possible since pjsip can be restarted (instead of shutdown).

Here is a sample scenario. App using PJSUA creates its own pool factory (instead of using PJSUA's) for some reason, and it uses the pool/factory to create a PJMEDIA port. It has to destroy the pool factory before shutting down PJSUA.

Re: lib (PJSUA?) restart, `pj_atexit()` will still be called right?

> 
> The important thing here seems to be that `pjmedia_conf_destroy()` must guarantee the completion of all pending async operations, if any.

Yes, `pjmedia_conf_destroy()` flushes all pending async ops.

sauwming:
> Here is a sample scenario. App using PJSUA creates its own pool factory (instead of using PJSUA's) for some reason, and it uses the pool/factory to create a PJMEDIA port. It has to destroy the pool factory before shutting down PJSUA.

I see.

> Re: lib (PJSUA?) restart, `pj_atexit()` will still be called right?

Ah, right. It's called upon `pj_shutdown()`.

pjmedia/src/pjmedia/vid_conf.c
```diff
@@ -93,6 +93,7 @@ typedef struct vconf_port
     unsigned             idx;           /**< Port index.                    */
     pj_str_t             name;          /**< Port name.                     */
     pjmedia_port        *port;          /**< Video port.                    */
+    pj_bool_t            is_new;        /**< Port newly added?              */
```
sauwming, line 96: I'm still not quite clear with the "new port" check. Since it also applies to video, is it considered a bug fix for the current code?

nanangizz: Yes, it is kind of bug fix (for some reason, for video it causes less/no harm).

Currently adding a port to conf needs to be done synchronously to avoid premature destroy (if async, the conf may end-up adding/accessing a destroyed port). However, we cannot increment/change `conf->port_cnt` there as it is accessed by conf `get_frame()` without mutex protection, e.g: https://github.com/pjsip/pjproject/blob/ed32a91a4a27fe4ebf3e53478f40a2a18396544d/pjmedia/src/pjmedia/conference.c#L2387
So, the "new port" check is introduced to be able to increment the `conf->port_cnt` safely.

sauwming (approved):

trengginas (approved):

trengginas:
There's an issue when testing this.
Scenario using pjsua sample app:
- set `auto-answer=200`
- on incoming call, the ring port is connected to the sound device port. (the connect port is queued)
- the call is answered, and the ring port is disconnected from the sound dev port (port connection does not exist)
- the connect port action in the queue is processed

```
08:00:36.107    inv04AA1314  .......State changed from NULL to INCOMING, event=TSX_STATE
08:00:36.107    pjsua_aud.c  ..Conf connect: 4 --> 0
...
08:00:36.170   conference.c  ....Connect ports 4->0 requested
08:00:36.170   conference.c  ....Connect ports 4->0 queued
08:00:36.170   pjsua_call.c  ..Answering call 0: code=200
...
08:00:36.176    pjsua_app.c  .....Call 0 media 0 [type=audio], status is Active
08:00:36.177    pjsua_aud.c  .....Conf disconnect: 4 -x- 0
08:00:36.177   conference.c  .......Disconnect ports 4->0 requested
08:00:36.177   conference.c  .......Ports connection 4->0 does not exist
...
08:00:36.208   conference.c !.New port 3 (ringback) is added
08:00:36.209   conference.c  .New port 4 (ring) is added
08:00:36.209   conference.c  .New port 5 (sip:192.168.100.51) is added
08:00:36.210   conference.c  .Port 4 (ring) transmitting to port 0 (Wave mapper)
08:00:36.210   conference.c  .Port 5 (sip:192.168.100.51) transmitting to port 0 (Wave mapper)
08:00:36.210   conference.c  .Port 0 (Wave mapper) transmitting to port 5 (sip:192.168.100.51)
```

sauwming:
I did the pre-release test and this seems to fail the double-hold scenario. Only video seems to be affected (video will not fully recover).

nanangizz:
> I did the pre-release test and this seems to fail the double-hold scenario. Only video seems to be affected (video will not fully recover).

See PR #4164

sauwming:
Thanks for fixing it!

Also, I noticed these additional two lines in pjsua app during shutdown:
```
13:13:33.343         sip_endpoint.c  .Endpoint 0x10ae0a2a8 destroyed
13:13:33.343               ringback  .Pool is not released by application, releasing now
13:13:33.343                   ring  .Pool is not released by application, releasing now
```
I suppose this is unavoidable due to issue 1 in the PR description above, correct?

nanangizz:
> Also, I noticed these additional two lines in pjsua app during shutdown:
> 
> ```
> 13:13:33.343         sip_endpoint.c  .Endpoint 0x10ae0a2a8 destroyed
> 13:13:33.343               ringback  .Pool is not released by application, releasing now
> 13:13:33.343                   ring  .Pool is not released by application, releasing now
> ```
> 
> I suppose this is unavoidable due to issue 1 in the PR description above, correct?

Unfortunately no. It is a new bug, see #4166.

LeonidGoltsblat:
Hi, everybody! I would like to leave feedback here.
Asynchronous commutation is a great idea!
And contrary to what the author wrote "this change may improve performance a little", it gives a serious performance improvement on a heavily loaded system where the high-priority conference bridge thread does not leave time for another thread to hold the conference mutex.

However, we are encountering several unexpected behavior changes and issues. Some of them are outlined below:

- Memory Leak Issue: A new cause of memory leak has been identified due to implicit pool creation with each media and conference port creation. An example of a buggy application is pjmedia_test. Further details and fixes can be found in #4235.

- Deadlock Risk: The solution is still not fully free from deadlocks. The OP_REMOVE_PORT function calls the group lock handler while holding the conf->mutex. As with any callback invoked under a lock, there is a potential risk of deadlock. We cannot predict what actions the user’s application may take in the callback, for instance, it might acquire the SIP dialog mutex, which could lead to a lock ordering issue similar to the one described in bug report #3183.

- Loss of control over data lifetime: The application loses control over the data lifetime. The recommendation to add a 100ms delay before destroying data works fine for applications handling a few simultaneous calls. However, this approach limits performance to only 10 operations per second, which is insufficient for high-load systems. I think we need an API to add/remove the grp_lock port conf handle - it's simple and gives the application the ability to safely close all the resources it needs.

nanangizz:
> Hi, everybody! I would like to leave feedback here. Asynchronous commutation is a great idea! And contrary to what the author wrote "this change may improve performance a little", it gives a serious performance improvement on a heavily loaded system where the high-priority conference bridge thread does not leave time for another thread to hold the conference mutex.
> 

First, thanks for the feedbacks!

Re: peformance improvement note, I believe so especially for such condition. Will update the desc.

> However, we are encountering several unexpected behavior changes and issues. Some of them are outlined below:
> 
>     * Memory Leak Issue: A new cause of memory leak has been identified due to implicit pool creation with each media and conference port creation. An example of a buggy application is pjmedia_test. Further details and fixes can be found in [Memory leak caused by change in media port behavior due to implementation of asynchronous conference bridge #4235](https://github.com/pjsip/pjproject/pull/4235).

Thanks for the bug report & the fix.

> 
>     * Deadlock Risk: The solution is still not fully free from deadlocks. The OP_REMOVE_PORT function calls the group lock handler while holding the conf->mutex. As with any callback invoked under a lock, there is a potential risk of deadlock. We cannot predict what actions the user’s application may take in the callback, for instance, it might acquire the SIP dialog mutex, which could lead to a lock ordering issue similar to the one described in bug report [Lock ordering issue in video conference bridge #3183](https://github.com/pjsip/pjproject/pull/3183).
>       I don't think it's necessary to call handle_op_queue under the mutex lock. Mutex protection is only required to protect conf->op_queue, not the handle_op_queue() function itself. This way, we can invoke the callback outside the conference mutex lock.

You're right, will try to fix for both video & audio conf.

> 
>     * Loss of control over data lifetime: The application loses control over the data lifetime. The recommendation to add a 100ms delay before destroying data works fine for applications handling a few simultaneous calls. However, this approach limits performance to only 10 operations per second, which is insufficient for high-load systems. I think we need an API to add/remove the grp_lock port conf handle - it's simple and gives the application the ability to safely close all the resources it needs.

Ideally app must check & update its ports to have its own pool (btw, the PJMEDIA tonegen and almost all PJMEDIA ports have been updated). It is not supposed to be a recommendation, just a sample of a possible workaround.

nanangizz:
Re: deadlock risk, please check #4243.

LeonidGoltsblat:
@nanangizz 
> Ideally app must check & update its ports to have its own pool (btw, the PJMEDIA tonegen and almost all PJMEDIA ports have been updated). It is not supposed to be a recommendation, just a sample of a possible workaround.

Real world applications may use pjsip as one of many other libraries and may not be written only in C/C++. The application manages resources outside the pjsip logic and may not be able to use the pool API to allocate memory, but in some cases the closing of application resources must be synchronized with the destruction of conference bridge ports.

For example, an application may implement a special media port proxy to play audio data stored in a BLOB field of a database. With an asynchronous switch, the application cannot close the database connection not only when performing a disconnect, but also when calling remove port, because the conference bridge may try to receive the next frame of data after that. The application must know the exact moment when to release resources, close the database connection, etc. This moment is the "grp_lock handle call".

nanangizz:
@LeonidGoltsblat

You're right, so far I missed the idea and your suggestion that group lock destroy handler can be used as a proper solution to avoid premature pool destroy! Updated the PR desc.

Btw, app can access the port's group lock directly and use the group lock APIs to add/remove handler. Anyway, added helper APIs in #4244. I cannot add you as a reviewer somehow, but feel free to review. Thanks again!

PR #3930, opened 2024-04-17
Added ability in stream to modify video codec framerate

shubhamfuloria:
Openh264 codec supports dynamic change of `bitrate` and `framerate`. This pull request contains functionality with which we can modify framerate of encoding stream after the codec is open with `SetOption` API of openh264.


sauwming:
Unfortunately modifying framerate is not that simple, since it not only involves the codec, but also other components such as video stream, otherwise the video stream may not know that the framerate has changed. There's also the issue with SDP negotiation which may limit the framerate that can be sent, so bypassing this may not be a good idea.

So, for major modifications such as resolution and/or fps, it may be better to send re-INVITE with REINIT_MEDIA instead.

PR #3935, opened 2024-04-23
Fixed PJSUA2 API to get/set Opus config

### LINKED ISSUE #3934: The makefile for SWIG does not pass parameters for conditional compilation
### Describe the bug

Endpoint.hpp is used to compile pjsua2_wrap.cpp, but there is a condition there  
`#if defined(PJMEDIA_HAS_OPUS_CODEC) && (PJMEDIA_HAS_OPUS_CODEC!=0)`
And to build pjsua2_wrap.cpp which would contain methods `getCodecOpusConfig()` and `setCodecOpusConfig()` I had to edit the `pjproject-2.14.1/pjsip-apps/src/swig/python/Makefile`. I have shown it in green on the screenshot. Overall, it was too non-obvious. Could you please improve this process or mention somewhere in the instructions about this. Thanks!

image

### Steps to reproduce

Use these instructions: https://docs.pjsip.org/en/latest/pjsua2/building.html#building-python-swig-module

### PJSIP version

2.14

### Context

This affects the SWIG building for all high level languages (Java Python C#)

### Log, call stack, etc

```shell
(.pjproject-venv) romk@RG:~/pjproject-2.14.1/pjsip-apps/src/swig/python$ make
swig -I../../../../pjlib/include -I../../../../pjlib-util/include -I../../../../pjmedia/include -I../../../../pjsip/include -I../../../../pjnath/include -c++  -w312 -threads -DSWIG_NO_EXPORT_ITERATOR_METHODS -python -o pjsua2_wrap.cpp ../pjsua2.i
python3 setup.py build
```

sauwming:
To fix #3934.

Checking the macro `PJMEDIA_HAS_OPUS_CODEC` within the PJSUA2 API declaration will complicate the build process, such as for SWIG binding.
We should check the macro in the implementation instead.

### DIFF
--- a/pjsip/src/pjsua2/endpoint.cpp
+++ b/pjsip/src/pjsua2/endpoint.cpp
@@ -2490,15 +2490,16 @@ void Endpoint::codecSetParam(const string &codec_id,
     PJSUA2_CHECK_EXPR( pjsua_codec_set_param(&codec_str, &pj_param) );
 }
 
-#if defined(PJMEDIA_HAS_OPUS_CODEC) && (PJMEDIA_HAS_OPUS_CODEC!=0)
 
 CodecOpusConfig Endpoint::getCodecOpusConfig() const PJSUA2_THROW(Error)
 {
-   pjmedia_codec_opus_config opus_cfg;
    CodecOpusConfig config;
+#if defined(PJMEDIA_HAS_OPUS_CODEC) && (PJMEDIA_HAS_OPUS_CODEC!=0)
+   pjmedia_codec_opus_config opus_cfg;
 
    PJSUA2_CHECK_EXPR(pjmedia_codec_opus_get_config(&opus_cfg));
    config.fromPj(opus_cfg);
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsua2/endpoint.cpp
```diff
@@ -2515,9 +2517,10 @@ void Endpoint::setCodecOpusConfig(const CodecOpusConfig &opus_cfg)
 
    PJSUA2_CHECK_EXPR(pjmedia_codec_opus_set_default_param(&new_opus_cfg,
                                                           &param));
-}
-
+#else
+    PJ_UNUSED_ARG(opus_cfg);
 #endif
```
nanangizz, line 2522: What if it throws some exception (e.g: `PJ_ENOTSUP`) when Opus is not enabled? (also in `getCodecOpusConfig()`)

nanangizz (approved):

Issue #3938, opened 2024-04-26
Crash in ioqueue_common_abs.c pj_ioqueue_sendto when checking IPv6 address length


rickdialpad:
### Describe the bug

Bug:
Assertion when checking address length in pj_ioqueue_sendto in ioqueue_common_abs.c of an IPv6 address. Looks like the `PJ_ASSERT_RETURN(addrlen <= (int)sizeof(pj_sockaddr_in), PJ_EBUG);` is old code that has not been updated to account for IPv6 addresses.

My Fix:
Had to change line ioqueue_common_abs.c:1059.
```
/*
 * Check that address storage can hold the address parameter.
 */
PJ_ASSERT_RETURN(addrlen <= (int)sizeof(pj_sockaddr_in), PJ_EBUG);
```

to 

```
/*
 * Check that address storage can hold the address parameter.
 */
const pj_sockaddr *a = (const pj_sockaddr*) addr;
PJ_ASSERT_RETURN(a->addr.sa_family == PJ_AF_INET ||
                 a->addr.sa_family == PJ_AF_INET6, 0);
PJ_ASSERT_RETURN(
  (a->addr.sa_family == PJ_AF_INET) ?
    (addrlen <= (int)sizeof(pj_sockaddr_in)) :
    (addrlen <= (int)sizeof(pj_sockaddr_in6)),
  PJ_EBUG
);
```

To make sure we do not crash if we ever get into pj_ioqueue_sendto with an IPv6 address. Previously it only checks IPv4 address lengths. If an IPv6 address was compared and had an `addrlen` of 28, it would crash.

There may be a better fix than my small patch that I do not know.

### Steps to reproduce

Happens extremely rarely. I was unable to reproduce the crash without changing the code to skip the `if (pj_list_empty(&key->write_list) {` conditional statement. But, we have about 200 crashes for users over the course of the last few months.

### PJSIP version

2.14

### Context

Not important because I found the problem and provided a fix. There may be a better way to fix this that I do not know.

### Log, call stack, etc

```shell
Not important because I found the problem and provided a fix. There may be a better way to fix this that I do not know.
```

PR #3941, opened 2024-04-29
Fix bad address length check in pj_ioqueue_sendto()

nanangizz:
Fix issue #3938:

> Assertion when checking address length in pj_ioqueue_sendto in ioqueue_common_abs.c of an IPv6 address. Looks like the `PJ_ASSERT_RETURN(addrlen <= (int)sizeof(pj_sockaddr_in), PJ_EBUG)` is old code that has not been updated to account for IPv6 addresses.

Thanks to @rickdialpad for the detailed report and the proposed patch.

PS: I don't think verifying the address family is necessary here as the check seems to concern about the storage capacity only. Btw, also updated the storage type (`write_operation.rmt_addr`) to `pj_sockaddr`.

### DIFF
--- a/pjlib/src/pj/ioqueue_common_abs.c
+++ b/pjlib/src/pj/ioqueue_common_abs.c
@@ -1056,7 +1056,7 @@ retry_on_restart:
     /*
      * Check that address storage can hold the address parameter.
      */
-    PJ_ASSERT_RETURN(addrlen <= (int)sizeof(pj_sockaddr_in), PJ_EBUG);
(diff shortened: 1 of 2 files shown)


sauwming (approved):
CI failed, but doesn't seem related.

trengginas (approved):

nanangizz:
> CI failed, but doesn't seem related.

`PJMEDIA_SRTP_HAS_DTLS` set to 1 in `config_site_test.h`, while configure cannot detect OpenSSL, so the error is quite as expected.

Not sure why other previous CI sessions didn't fail here, perhaps OpenSSL were available most of the time (not always)?

PR #3942, opened 2024-05-02
Print log after sending SIP message successfully

nanangizz:
Sending a SIP message is done asynchronously, so when the log says "TX n bytes Request msg.." actually the message may not be on the wire yet. Even for TCP/TLS the transport, the connection may not be established yet. Having a log line printed for a successful send can be useful for debugging purpose.

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -877,6 +877,12 @@ static void transport_send_callback(pjsip_transport *transport,
 
     PJ_UNUSED_ARG(transport);
 
+    /* Print log on successful sending */
+    if (size > 0) {
+        PJ_LOG(5,(transport->obj_name,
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsip/sip_transport.c
```diff
@@ -877,6 +877,12 @@ static void transport_send_callback(pjsip_transport *transport,
 
     PJ_UNUSED_ARG(transport);
```
sauwming, line 878: So this can be removed now?

sauwming (approved):

pjsip/src/pjsip/sip_transport.c
nanangizz: Yes, but as `transport` is only used for logging with level 5, if `PJ_LOG_MAX_LEVEL` is set to <5, the unused var warning will reappear. So I guess better leave it as is for now?

PR #3960, opened 2024-05-14
pjsip_find_msg: Log warning if Content-Length field not found

pespin:
The pjsip_find_msg() is used when in TCP to find the message boundaries in the stream. For that, it uses the Content-Length field, which is only mandatory in SIP TCP according to spec. Since this function is used only in TCP transport, that's totally fine.
However, a SIP test written in TTCN-3 was recently sending a message without Content-Length field (because there was no body at all in the message). That made Asterisk discard the message without any kind of feedback, as if it was a black hole.
Write down some log line explaining the user that a Content-Length value is expected.

### DIFF
--- a/pjsip/src/pjsip/sip_parser.c
+++ b/pjsip/src/pjsip/sip_parser.c
@@ -937,6 +937,8 @@ PJ_DEF(pj_status_t) pjsip_find_msg( const char *buf, pj_size_t size,
 
     /* Found Content-Length? */
     if (content_length == -1) {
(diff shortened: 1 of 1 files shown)


trengginas:
Right, incoming message without `Content-Length:` header will be dropped without any information, and the `tp_drop_data_cb()` is not called either. Users should be notified either from the log or via the callback.

pespin:
@trengginas what you suggest is done in my other PR (https://github.com/pjsip/pjproject/pull/3961). I still think it's fine to merge this PR as it is now, so that logging provides the exact reason of the failure.

nanangizz (approved):

sauwming (approved):

PR #3961, opened 2024-05-14
sip_transport: tcp: Notify app if unable to keep decoding stream

pespin:
If pjsip_find_msg() starts failing because there's some missing header like Content-Length, log the issue and shutdown the transport to avoid staying in a broken state.

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -2156,40 +2156,67 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
 
             msg_status = pjsip_find_msg(current_pkt, remaining_len, PJ_FALSE, 
                                         &msg_fragment_size);
-            if (msg_status != PJ_SUCCESS) {
-                if (remaining_len == PJSIP_MAX_PKT_LEN) {
-                    mgr->on_rx_msg(mgr->endpt, PJSIP_ERXOVERFLOW, rdata);
-                    
-                    /* Notify application about the message overflow */
-                    if (mgr->tp_drop_data_cb) {
-                        pjsip_tp_dropped_data dd;
-                        pj_bzero(&dd, sizeof(dd));
-                        dd.tp = tr;
-                        dd.data = current_pkt;
-                        dd.len = msg_fragment_size;
-                        dd.status = PJSIP_ERXOVERFLOW;
-                        (*mgr->tp_drop_data_cb)(&dd);
-                    }
-
-                    if (rdata->tp_info.transport->idle_timer.id == 
-                                                         INITIAL_IDLE_TIMER_ID)
-                    {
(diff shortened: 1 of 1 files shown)


trengginas:
Shouldn't calling the callback be enough? The user will be notified and shutdown the transport if needed.

pespin:
@trengginas I'm fine with your proposal. TBH I care more about the "logging" side of things to start with, to at least report that the stream fails to parse and becomes broken. Before that, there was no message at all, everything happened silently.

I'll submit a  new version dropping the transport shutdown.

pjsip/src/pjsip/sip_transport.c
```diff
@@ -2156,41 +2156,70 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
...
-                    {
-                        /* We are not getting the first valid SIP message 
-                         * as expected, close the transport.
-                         */
-                        PJ_LOG(4, (THIS_FILE, "Unexpected data was received "\
-                            "while waiting for a valid initial SIP messages. "\
-                            "Shutting down transport %s",
-                            rdata->tp_info.transport->obj_name));
-
-                        pjsip_transport_shutdown(rdata->tp_info.transport);
```
pespin, line 2185: @trengginas I added the transport_shutdown in the new extended code path because I saw in the other code path it was already being done. So not sure if then this path was wrong then?
Let me know whether you want to merge as it, drop the pjsip_transport_shutdown() from the new path or from both paths.

nanangizz:
I also vote for callback approach, shutting down transport may not be preferred in some cases so IMO better leave it for app to decide.

nanangizz:
Hmm.. honestly the flow resturcturing make me nervous about possibility of bug or major behavior change.

pespin:
@nanangizz it's just a matter of looking at pjsip_find_msg() and seeing the error codes returned in each scenario, and act on them.

The whole "if (msg_status != PJ_SUCCESS) {" path is transformed in the first case so we get rid of a whole indentation level.

Then, the other msg_status are further discriminated. I'm keeping the old code path (the one returning PJSIP_EPARTIALMSG) the same. I'm simply extending the other error codes to handle them, since they were ignored in the past.

nanangizz:
> @nanangizz it's just a matter of looking at pjsip_find_msg() and seeing the error codes returned in each scenario, and act on them.

Originally non-success cases are evaluated based on `remaining_len` instead of error codes. To me this is not "just a matter of ...", it definitely needs a very careful check to make sure no behavior change is introduced. Note that this code is part of the core of the lib and have been used and tested for a long time. This is what makes me nervous.

> 
> The whole "if (msg_status != PJ_SUCCESS) {" path is transformed in the first case so we get rid of a whole indentation level.

Indentation is not a very strong reason to modify the flow, the code is only few lines and not that hard to read. CMIIW the new proposed feature in this PR seems to be able to be inserted with much less risk and easier to review :)

Issue #3986, opened 2024-06-13
pjsua_core.c Set default cfg ka_data to double CRLF


space88man:
### Describe the feature

RFE: Set the default `cfg.ka_data` to double-CRLF so that programming `pjsua_acc_config` will not require overriding the default.

Observation from the `pjsua` command line app
```
# REGISTER reponse has Flow-Timer; transport is TLS
# This is sending a bogus SIP message

17:40:42.318            pjsua_acc.c  Sending 2 bytes keep-alive packet for acc 3 to 10.0.0.1:5061
```

The command line app bootstraps `pjsua_acc_config` from the defaults but `ka_data` is single-CRLF.
Without overriding this the UAS will reject this 2-byte SIP message. This invalid message is sent by
`pjsua_acc.c:keep_alive_timer_cb()` due to `Flow-Timer` in the response header.

### Describe alternatives you've considered

Whenever creating `pjsua_acc_config` from defaults always remember to set ka_data to double-CRLF

### Additional context

This results in a bug with pjsua app as it doesn't explicity set `pjsua_acc_config:ka_data` to double-CRLF (it just uses the default value).

```
pjsua_core.c:
    338     cfg->lock_codec = 1;
    339     cfg->ka_interval = 15;
    340     cfg->ka_data = pj_str("\r\n");  // <---- bogus SIP message here
    341     cfg->vid_cap_dev = PJMEDIA_VID_DEFAULT_CAPTURE_DEV;
    342     cfg->vid_rend_dev = PJMEDIA_VID_DEFAULT_RENDER_DEV;
```

PR #3987, opened 2024-06-13
Set default ka_data to double-CRLF - so it is a valid SIP message

space88man:
Set the default `.ka_data` field to double-CRLF, i.,e., a valid SIP message.

Users of the default `pjsua_acc_config` have a bogus `.ka_data` field — affects `pjsua` CLI application when the RESPONSE has `Flow-Timer`.

Addresses #3986

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_core.c
+++ b/pjsip/src/pjsua-lib/pjsua_core.c
@@ -337,7 +337,7 @@ PJ_DEF(void) pjsua_acc_config_default(pjsua_acc_config *cfg)
     cfg->timer_setting = pjsua_var.ua_cfg.timer_setting;
     cfg->lock_codec = 1;
     cfg->ka_interval = 15;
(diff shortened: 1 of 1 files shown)


sauwming:
Need to modify the doc in `pjsua.h` as well.

According to https://www.rfc-editor.org/rfc/rfc5626#section-4.4.1:
```
4.4.1 Keep-Alive with CRLF
...
   For this mechanism, the client "ping" is a double-CRLF sequence, and
   the server "pong" is a single CRLF, ...
```

So although there's nothing wrong with the current default of a single CRLF, using double CRLF seems to be better.

sauwming:
Sorry, I skipped the part that said:

```
4.4.1 Keep-Alive with CRLF

This approach MUST only be used with connection oriented transports
   such as TCP or SCTP; it MUST NOT be used with connection-less
   transports such as UDP.
```
So for UDP, we're actually recommended not to use this.

PR #3990, opened 2024-06-17
pj_dns_srv_resolve() returns query object when it shouldn't

andreas-wehrmann:
According to the documentation of [pj_dns_srv_resolve()](https://github.com/pjsip/pjproject/blob/28c54a42f6978e6c5b8d051bb9ae984a3a6130d7/pjlib-util/include/pjlib-util/srv_resolver.h#L183) the optional parameter _p_query_ \
can be used to retrieve the query object _if one was started_.

It's supposed to return NULL when the response is available from the cache. \
I noticed that the function **always** returns the internal query object, no matter if the response is in the cache or not
which makes it weird to deal with the case where the response is available immediately \
(because the callback will be called before the function returns). \
This PR adds an additional check to see whether a query was actually started and returns the object only in that case (if any).

### DIFF
--- a/pjlib-util/include/pjlib-util/resolver.h
+++ b/pjlib-util/include/pjlib-util/resolver.h
@@ -406,7 +406,7 @@ PJ_DECL(pj_status_t) pj_dns_resolver_destroy(pj_dns_resolver *resolver,
  * @param user_data Arbitrary user data to be associated with the query,
  *                  and which will be given back in the callback.
  * @param p_query   Optional pointer to receive the query object, if one
- *                  was started. If this pointer is specified, a NULL may
+ *                  was started. If this pointer is specified, a NULL will
  *                  be returned if response cache is available immediately.
  *
  * @return          PJ_SUCCESS if either an asynchronous query has been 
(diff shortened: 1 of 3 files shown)


sauwming:
The doc uses the word `may`, so perhaps it's better to change it to a more definitive term, e.g.:
If this pointer is specified, a NULL `will` be returned if response cache is available immediately.

andreas-wehrmann:
Ok, I updated the doc, also for [pj_dns_resolver_start_query()](https://github.com/pjsip/pjproject/blob/28c54a42f6978e6c5b8d051bb9ae984a3a6130d7/pjlib-util/include/pjlib-util/resolver.h#L416) which is used internally and follows the same semantics.

I didn't understand `may` as in _may or may not be NULL if a response is available_ but rather: _If_ you specify a pointer then under certain circumstances NULL may be returned; The circumstance being the response is available immediately.

nanangizz:
Btw, the function implementation never really set it to `NULL`? In other words, it seems to rely on app to initialize it with `NULL`, if so, perhaps we need to fix this too?

andreas-wehrmann:
Good catch! \
I saw that [pj_dns_resolver_start_query()](https://github.com/pjsip/pjproject/blob/28c54a42f6978e6c5b8d051bb9ae984a3a6130d7/pjlib-util/src/pjlib-util/resolver.c#L867) does this as well, so I updated the function accordingly.

nanangizz (approved):

sauwming (approved):

PR #4004, opened 2024-06-27
Remove OpenSSL<1.1.0 ssl_sock_ossl.c

r-barnes:
I can't imagine this version of SSL is secure any more or that this code is getting well-tested. Let's remove it and focus on the modern forms of the library.


sauwming:
Unfortunately, some of our customers still report using 1.0.2

PR #4013, opened 2024-07-09
Simple command line argument parsing utility in PJLIB

bennylp:
The PJSIP suite contains copy of `getopt` implementation, but it is in PJLIB-UTIL and also has different licensing than PJSIP.

This PR implements a simple, header only argument parsing utilities that can be (and will be) used by `pjlib-test` and all unit testing apps to replace current manual parsing. 

## Usage

Argparse provides these few utilities:

- `pj_argparse_exists(..)`: check if argument exists
- `pj_argparse_get_bool(..)`: get an argument and remove it from argc/argv if it exists
- `pj_argparse_get_str(..)`: get an argument and its value and remove the argument-value pair from argc/argv if it exists
- `pj_argparse_get_int(..)`: get an argument and its int value and remove the argument-value pair from argc/argv if it exists
- `pj_argparse_peek_next_option(..)`: get first option (argument with preceeding `-`, or `--`) in argv, to check if there is remaining unrecognized option

### Note

The difference between `pj_argparse_exists(..)` and `pj_argparse_get_bool()` is the latter removes the argument from argc/argv if the argument exists.

There is a different semantic between the return value of `pj_argparse_get_bool()` and `pj_argparse_get_str()` or `pj_argparse_get_int()`. 

`pj_argparse_get_bool()` returns "the value", i.e. TRUE if the the argument exists.

`pj_argparse_get_bool()` and `pj_argparse_get_str()` returns `pj_status_t`, i.e. `PJ_SUCCESS` if everything is okay (an argument has been successfully parsed, or the argument does not exist), or error otherwise. In case of error, the error string would have been printed to the screen.

This design decision is chosen for convenience.

### Sample

For example, to parse these options:

```
Usage:
  pjlib-test [OPTION] [test_to_run] [test_to_run]...

where OPTIONS:
  -h, --help             Show this help screen
  -p, --port PORT        Use port PORT for echo port
  -s, --server SERVER    Use SERVER as echo server

test_to_run is the name of the test, e.g. rand_test, os_test, etc.
```
the following code can be used:

```
int main(int argc, char *argv[])
{
    int i, port = 0;
    char *server = NULL;

    if (pj_argparse_get_bool(&argc, argv, "-h") ||
        pj_argparse_get_bool(&argc, argv, "--help"))
    {
        usage();
        return 0;
    }

    if (pj_argparse_get_int(&argc, argv, "-p", &port) ||
        pj_argparse_get_int(&argc, argv, "--port", &port))
    {
        usage();
        return 1;
    }

    if (pj_argparse_get_str(&argc, argv, "-s", &server) ||
        pj_argparse_get_str(&argc, argv, "--server", &server))
    {
        usage();
        return 1;
    }

    printf("Request to run %d test(s)\n", argc-1);
    for (i=1; i<argc; ++i)
        printf("  - %s", argv[i]);
    puts("");

    ...
}
```

## Limitations

First, the utility only supports white space(s) as separator between an option and its value. Equal sign is not supported:

```
# this will not work
$ ./app --server=127.0.0.1

# this works
$ ./app --server 127.0.0.1
```
Second, the utility does not support double dash (`--`) as separator between options and operands (see [this discussion in stackexchange](https://unix.stackexchange.com/questions/11376/what-does-double-dash-double-hyphen-mean) on the meaning of `--`)

### DIFF
--- a/pjlib/include/pj/argparse.h
+++ b/pjlib/include/pj/argparse.h
@@ -1,0 +1,198 @@
+/*
+ * Copyright (C) 2008-2024 Teluu Inc. (http://www.teluu.com)
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+#ifndef __PJ_ARGPARSE_H__
+#define __PJ_ARGPARSE_H__
+
+/**
+ * @file argparse.h
+ * @brief Command line argument parser
+ */
+#include <pj/ctype.h>
+#include <pj/errno.h>
+#include <pj/string.h>
+
+PJ_BEGIN_DECL
+
+/**
+ * Define function to display parsing error.
+ */
+#ifndef PJ_ARGPARSE_ERROR
+#  include <stdio.h>
+#  define PJ_ARGPARSE_ERROR(fmt, arg) printf(fmt "\n", arg)
+#endif
+
+
+/**
+ * @defgroup PJ_ARGPARSE Command line argument parser
+ * @ingroup PJ_MISC
+ * @{
+ *
+ * This module provides header only utilities to parse command line arguments.
+ * This is mostly used by PJSIP test and sample apps. Note that there is
+ * getopt() implementation in PJLIB-UTIL (but it's in PJLIB-UTIL, so it can't
+ * be used by PJLIB)
+ */
+
+/**
+ * Peek the next possible option from argv. An argument is considered an
+ * option if it starts with "-" and followed by at least another letter that
+ * is not digit or starts with "--" and followed by a letter.
+ *
+ * @param argv      The argv, which must be null terminated.
+ *
+ * @return next option or NULL.
+ */
+static char* pj_argparse_peek_next_option(char *const argv[])
+{
+    while (*argv) {
+        const char *arg = *argv;
+        if ((*arg=='-' && *(arg+1) && !pj_isdigit(*(arg+1))) ||
+            (*arg=='-' && *(arg+1)=='-' && *(arg+2)))
+        {
+            return *argv;
+        }
+        ++argv;
+    }
+    return NULL;
+}
+
+/**
+ * Check that an option exists, without modifying argv.
+ *
+ * @param opt       The option to find, e.g. "-h", "--help"
+ * @param argv      The argv, which must be null terminated.
+ *
+ * @return PJ_TRUE if the option exists, else PJ_FALSE.
+ */
+static pj_bool_t pj_argparse_exists(const char *opt, char *const argv[])
+{
+    int i;
+    for (i=1; argv[i]; ++i) {
+        if (pj_ansi_strcmp(argv[i], opt)==0)
+            return PJ_TRUE;
+    }
+    return PJ_FALSE;
+}
+
+/**
+ * Check for an option and if it exists, returns PJ_TRUE remove that option
+ * from argc/argv.
+ *
+ * @param opt       The option to find, e.g. "-h", "--help"
+ * @param argc      Pointer to argc.
+ * @param argv      Null terminated argv.
+ *
+ * @return PJ_TRUE if the option exists, else PJ_FALSE.
+ */
+static pj_bool_t pj_argparse_get_bool(const char *opt, int *argc, char *argv[])
+{
+    int i;
+    for (i=1; argv[i]; ++i) {
+        if (pj_ansi_strcmp(argv[i], opt)==0) {
+            pj_memmove(&argv[i], &argv[i+1], ((*argc)-i)*sizeof(char*));
+            (*argc)--;
+            return PJ_TRUE;
+        }
+    }
+    return PJ_FALSE;
+}
+
+/**
+ * Check for an option and if it exists, get the value and remove both
+ * the option the the value from argc/argv. Note that the function only
+ * supports whitespace as separator between option and value (i.e. equal
+ * sign is not supported).
+ *
+ * @param opt           The option to find, e.g. "-t", "--type"
+ * @param argc          Pointer to argc.
+ * @param argv          Null terminated argv.
+ * @param ptr_value     Pointer to receive the value.
+ *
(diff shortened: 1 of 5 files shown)


pjlib/include/pj/argparse.h
```diff
@@ -0,0 +1,198 @@
...
+ * @param argv      The argv, which must be null terminated.
+ *
+ * @return next option or NULL.
+ */
+PJ_INLINE(char*) pj_argparse_peek_next_option(char *const argv[])
+{
+    while (*argv) {
+        const char *arg = *argv;
+        if ((*arg=='-' && *(arg+1) && !pj_isdigit(*(arg+1))) ||
+            (*arg=='-' && *(arg+1)=='-' && *(arg+2)))
```
sauwming, line 65: according to the spec, *(arg+2) also needs to be a letter, so need to check `!pj_isdigit()` as well?

pjlib/include/pj/argparse.h
```diff
@@ -0,0 +1,198 @@
...
+    return NULL;
+}
+
+/**
+ * Check that an option exists, without modifying argv.
+ *
+ * @param opt       The option to find, e.g. "-h", "--help"
+ * @param argv      The argv, which must be null terminated.
+ *
+ * @return PJ_TRUE if the option exists, else PJ_FALSE.
```
sauwming, line 80: minor: indentation
typically we also align the `@return` content with the `@param` content, i.e.
```
@param argv      The argv, which must be null terminated.

@return          PJ_TRUE if the option exists, else PJ_FALSE.
```

pjlib/include/pj/argparse.h
```diff
@@ -0,0 +1,198 @@
...
+ */
+PJ_INLINE(pj_bool_t) pj_argparse_exists(const char *opt, char *const argv[])
```
sauwming, line 82: I could be mistaken here, but usually it's the other way around, `pj_argparse_exists(argv, opt)`, with the object being searched (i.e. the main object) being the first parameter.

pjlib/include/pj/argparse.h
```diff
@@ -0,0 +1,198 @@
...
+    int i;
+    for (i=1; argv[i]; ++i) {
+        if (pj_ansi_strcmp(argv[i], opt)==0)
+            return PJ_TRUE;
+    }
+    return PJ_FALSE;
+}
+
+/**
+ * Check for an option and if it exists, returns PJ_TRUE remove that option
```
sauwming, line 93: `return` PJ_TRUE `and` remove

pjlib/include/pj/argparse.h
```diff
@@ -0,0 +1,198 @@
...
+            (*argc)--;
+            return PJ_TRUE;
+        }
+    }
+    return PJ_FALSE;
+}
+
+/**
+ * Check for an option and if it exists, get the value and remove both
+ * the option the the value from argc/argv. Note that the function only
```
sauwming, line 117: `and` the value

Also below.

pjlib/include/pj/argparse.h
```diff
@@ -0,0 +1,198 @@
...
+ * supports whitespace as separator between option and value (i.e. equal
```
sauwming, line 118: a single whitespace or whitespaces?

nanangizz:
> `pj_argparse_get_bool()` and `pj_argparse_get_str()` returns `pj_status_t`, i.e. `PJ_SUCCESS` if everything is okay (an argument has been successfully parsed, or the argument does not exist), or error otherwise. In case of error, the error string would have been printed to the screen.
> 
> This design decision is chosen for convenience.

IMHO it is easier to have error returned when argument does not exist because:
- More intuitive (perhaps subjective) as expected-but-not-exist is usually not a 'success'.
- App does not need a separate check (using `pj_argparse_exists()`), especially for `int` option. Btw, currently the `str` version also does not seem to return `NULL` when option does not exist, is this intentional?
- A bit more efficient because a separate existance check performs an additional loop.

pjlib/include/pj/argparse.h
bennylp: The comment meant to say that a negative number (e.g. `-1`) will not be considered as an option. But thinking about it again, and checking the behavior of some programs e.g. `gcc` or `clang`, they both consider `-1` as an option, e.g.:

```
$ gcc -1
gcc: error: unrecognized command-line option ‘-1’
```

So in the latest commit, any arguments that start with dash is considered an option (and comment updated accordingly)

pjlib/include/pj/argparse.h
bennylp: Done

pjlib/include/pj/argparse.h
bennylp: I think that's a good point. Rearranged the argument (also in `pj_argparse_get_bool()/_str()/_int()` to the same convention) in latest commit

pjlib/include/pj/argparse.h
bennylp: Done

pjlib/include/pj/argparse.h
bennylp: Done

pjlib/include/pj/argparse.h
bennylp: Added "limitations" section in this PR and in argparse doxygen to clarify the problem. Also doxygen comments updated to use "whitespace(s)".

bennylp:
> > (an argument has been successfully parsed, or the argument does not exist), or error otherwise. In case of error, the error string would have been printed to the screen.
> > This design decision is chosen for convenience.
> 
> IMHO it is easier to have error returned when argument does not exist because:
> 
>     * More intuitive (perhaps subjective) as expected-but-not-exist is usually not a 'success'.
> 

Agreed that this is rather subjective, and it is in my view that since options are by definition optional, it is not wrong to consider lack of it as "success" condition.

But I also understand your point, i.e. if we issue `"_get_something()"`, then one should expect that (s)he gets something or otherwise error should be reported.

In the end, I chose current approach because it is easier to use (we'll discuss more below) and while taking this approach probably is not idealistic, it's not wrong.

Btw initially the API uses that approach i.e. to return `PJ_ENOTFOUND` when the option does not exist (see sample [here](https://github.com/pjsip/pjproject/blob/13db0c6fe8ce43833562a8a37118569f190c0116/pjlib/src/pjlib-test/main.c#L124)). But this makes using it more verbose, especially when both short and long options are used.


>     * App does not need a separate check (using `pj_argparse_exists()`), especially for `int` option. Btw, currently the `str` version also does not seem to return `NULL` when option does not exist, is this intentional?
> 

It is indeed intentional for `get_str()` and `_get_int()`, i.e. if the option does not exist then the return value will not be touched. By convention, app should initialize the option value with a default value, which can be used to check if the value is changed by user via cmd line.

Also this allows handling short and long options like this:

```
    int port = -1;
    
    if (pj_argparse_get_int("-p", &argc, argv, &port) ||
        pj_argparse_get_int("--port", &argc, argv, &port))
    {
        usage();
        return 1;
    }
```

So in most cases, checking if the default value has changed (e.g. not `-1` anymore in above sample) can be used instead of `pj_argparse_exists()`.

>     * A bit more efficient because a separate existance check performs an additional loop.

See above.

sauwming (approved):

nanangizz (approved):

PR #4026, opened 2024-07-31
SIP transport: call callback when data is dropped due to invalid headers

sauwming:
To continue #3961.

Currently, the app is not notified if incoming data is dropped due to errors of `PJSIP_EMISSINGHDR` or `PJSIP_EINVALIDHDR`. So the patch will now call the callbacks in such scenario.

The original patch is created by @pespin.
This PR is to minimise the flow/code changes.

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -2157,20 +2157,43 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
             msg_status = pjsip_find_msg(current_pkt, remaining_len, PJ_FALSE, 
                                         &msg_fragment_size);
             if (msg_status != PJ_SUCCESS) {
-                if (remaining_len == PJSIP_MAX_PKT_LEN) {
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsip/sip_transport.c
```diff
@@ -2157,20 +2157,43 @@ PJ_DEF(pj_ssize_t) pjsip_tpmgr_receive_packet( pjsip_tpmgr *mgr,
...
-                    
-                    /* Notify application about the message overflow */
+                pj_status_t dd_status = msg_status;
+
+                if (msg_status == PJSIP_EPARTIALMSG) {
+                    if (remaining_len != PJSIP_MAX_PKT_LEN) {
+                        /* Not enough data in packet. */
+                        return total_processed;
+                    }
+                    dd_status = PJSIP_ERXOVERFLOW;
```
nanangizz, line 2167: - Better put comment for future ref about the two cases of `EPARTIALMSG`, i.e: really partial message or overflow (buffer too small or packet too large)?
- And/or print log (as in the other errors)?

nanangizz (approved):

PR #4042, opened 2024-08-19
Add option to unescape or not during SIP URI parsing

sauwming:


pjsip/src/pjsip/sip_uri.c
```diff
@@ -404,7 +408,13 @@ static pj_status_t pjsip_url_compare( pjsip_uri_context_e context,
      * This includes userinfo containing passwords or formatted as 
      * telephone-subscribers.
      */
-    if (pj_strcmp(&url1->user, &url2->user) != 0)
+    str[0].ptr = buf[0];
```
trengginas, line 411: Should it protect this with PJSIP_UNESCAPE_WHEN_PARSING_URI when comparing unescape user part?

sauwming:
A different approach was adopted in #4043.

PR #4043, opened 2024-08-19
Save the original user and password part when parsing URI

sauwming:
We have received the following report:
- Remote sends a URI that escapes an unreserved character, which is actually allowed by [RFC 2396: Uniform Resource Identifiers (URI)](https://datatracker.ietf.org/doc/html/rfc2396#section-2.4.2) (as well as the newer [RFC 3896](https://datatracker.ietf.org/doc/html/rfc3986#section-2.4)):
```
In some cases, data that could be represented by an unreserved
   character may appear escaped; for example, some of the unreserved
   "mark" characters are automatically escaped by some systems.
```
For example: `sip:%2auser@host.com` (which is decoded as `sip:*user@host.com`).

- PJSIP will parse the URI, unescape it, and then store and use it in its unescaped form.
- But the remote will reject this URI when unescaped.

Although both URIs should be identical according to the RFC, it may be preferable if we use the URI in its original encoding.

### DIFF
--- a/pjsip/include/pjsip/sip_uri.h
+++ b/pjsip/include/pjsip/sip_uri.h
@@ -337,6 +337,7 @@ PJ_INLINE(void*) pjsip_uri_clone( pj_pool_t *pool, const void *uri )
 typedef struct pjsip_sip_uri
 {
     pjsip_uri_vptr *vptr;               /**< Pointer to virtual function table.*/
+    pj_str_t        original;           /**< Optional pointer to original user and password. */
     pj_str_t        user;               /**< Optional user part. */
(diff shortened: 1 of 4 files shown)


trengginas (approved):

nanangizz (approved):

sauwming:
Can you elaborate? The patch will only make the number appear encoded, *if* it's originally encoded by the remote, it shouldn't make it disappear.

umerov1999:
if (url->orig_userpass.slen)
     copy_advance_check(buf, url->orig_userpass);

I think the problem is in this piece of code. My usernames are set in Cyrillic

PR #4067, opened 2024-09-06
Update IP version choosing logic in media transport creation for generating SDP offer.

nanangizz:
This will update the behavior of account config's [`ipv6_media_use`](https://docs.pjsip.org/en/latest/api/generated/pjsip/group/group__PJSUA__LIB__ACC.html#_CPPv4N16pjsua_acc_config14ipv6_media_useE):
- `PJSUA_IPV6_DISABLED`, media will always use IPv4 regardless (may use loopback/local-link IP address).
- `PJSUA_IPV6_ENABLED_USE_IPV6_ONLY`, media will always use IPv6 regardless (may use loopback/local-link IP address).
- `PJSUA_IPV6_ENABLED_PREFER_IPV4`, media will use IPv4 if usable, otherwise it may fallback to IPv6 (if it is usable). If none is usable, it will prioritize IPv4 (if available).
- `PJSUA_IPV6_ENABLED_PREFER_IPV6,` media will use IPv6 if usable, otherwise it may fallback to IPv4 (if it is usable). If none is usable, it will prioritize IPv6 (if available).
- `PJSUA_IPV6_ENABLED` or `PJSUA_IPV6_ENABLED_NO_PREFERENCE`, media will find any usable IP version, and prioritizing IPv4 (e.g: if both are usable/available).

By useable, it means that the host has an IP address that is not loopback, local-link, nor disabled/invalid.
By available, it means that the host has an IP address for the specified IP version, including loopback or local-link address.

### DIFF
--- a/pjlib/include/pj/addr_resolv.h
+++ b/pjlib/include/pj/addr_resolv.h
@@ -180,6 +180,51 @@ PJ_DECL(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *name,
                                     unsigned *count, pj_addrinfo ai[]);
 
 
+/**
+ * Enumeration of IP address type.
+ */
+typedef enum pj_addr_type
+{
+    /**
+     * Disabled: 0.0.0.0/8 or ::/128
+     */
+    PJ_ADDR_TYPE_DISABLED       = 1,
+
+    /**
+     * Loopback: 127.0.0.0/8 or ::1/128
+     */
+    PJ_ADDR_TYPE_LOOPBACK       = 2,
+
+    /**
+     * Link-local: 169.254.0.0/16 or fe80::/64
+     */
+    PJ_ADDR_TYPE_LINK_LOCAL     = 4,
+
+    /**
+     * Private: 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16 or fc00::/7
+     */
+    PJ_ADDR_TYPE_PRIVATE        = 8,
+
+    /**
+     * Multicast: 224.0.0.0/3 or ff00::/8
+     */
+    PJ_ADDR_TYPE_MULTICAST      = 16,
+
+} pj_addr_type;
+
+
+
+/**
+ * Check IP address type.
+ *
+ * @param addr      The IP address to check, must be IPv4 or IPv6 address.
+ * @param type      Bit mask of address type pj_addr_type.
+ *
+ * @return          PJ_TRUE if the type of specified address \addr match to
+ *                  the specified types \a type.
+ */
+PJ_DECL(pj_bool_t) pj_check_addr_type(const pj_sockaddr *addr, unsigned type);
+
 
 /** @} */
 
--- a/pjlib/src/pj/sock_common.c
+++ b/pjlib/src/pj/sock_common.c
@@ -730,6 +730,7 @@ PJ_DEF(pj_status_t) pj_sockaddr_parse( int af, unsigned options,
     return status;
 }
 
+
 /* Resolve the IP address of local machine */
 PJ_DEF(pj_status_t) pj_gethostip(int af, pj_sockaddr *addr)
 {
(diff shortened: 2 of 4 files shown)


sauwming (approved):
This is great improvement!

trengginas (approved):

andreas-wehrmann:
> * `PJSUA_IPV6_ENABLED_PREFER_IPV4`, media will use IPv4 if usable, otherwise it may fallback to IPv6 (if it is usable). If none is usable, it will prioritize IPv4 (if available).
> 
> * `PJSUA_IPV6_ENABLED_PREFER_IPV6,` media will use IPv6 if usable, otherwise it may fallback to IPv4 (if it is usable). If none is usable, it will **prioritize IPv4** (if available)

Is this true?
My expectation would be that for `PJSUA_IPV6_ENABLED_PREFER_IPV6` it will prioritize IPv6 if none is available.

nanangizz:
> > * `PJSUA_IPV6_ENABLED_PREFER_IPV4`, media will use IPv4 if usable, otherwise it may fallback to IPv6 (if it is usable). If none is usable, it will prioritize IPv4 (if available).
> > * `PJSUA_IPV6_ENABLED_PREFER_IPV6,` media will use IPv6 if usable, otherwise it may fallback to IPv4 (if it is usable). If none is usable, it will **prioritize IPv4** (if available)
> 
> Is this true? My expectation would be that for `PJSUA_IPV6_ENABLED_PREFER_IPV6` it will prioritize IPv6 if none is available.

IMO the "preference" is originally for good scenario (i.e: both usable). But when there is a problem such as "both are not usable" (we expect media won't work), I assume that IPv4 may have a little bit better chance to work than IPv6 (e.g: because we used to deal with IPv4 more than IPv6)? I am open to this, let's hear what others may say.

Btw, as described in the definitions, "none is available" is different to "none is usable". But yes, even when both is not available, it will prioritize IPv4 for the same reason above.

sauwming:
Since the user has explicitly specified its IPv6 preference, IMO following the user would be more reasonable.

andreas-wehrmann:
> Since the user has explicitly specified its IPv6 preference, IMO following the user would be more reasonable.

Right, that was my expectation when I read `PREFER_IPV6`.
But a fallback to IPv4 might still make sense, when an IPv4 is _usable_ but IPv6 is only _available_.
The reverse is true for IPv4 to IPv6 fallback.

I'd like to hear what others think about this.

nanangizz:
> > Since the user has explicitly specified its IPv6 preference, IMO following the user would be more reasonable.
> 
> Right, that was my expectation when I read `PREFER_IPV6`. But a fallback to IPv4 might still make sense, when an IPv4 is _usable_ but IPv6 is only _available_. The reverse is true for IPv4 to IPv6 fallback.
> 
> I'd like to hear what others think about this.

I've updated the behavior based on your comments in https://github.com/pjsip/pjproject/pull/4067/commits/76496f0f28e7b073828da7ccd5fe1f714714d4fc. Thanks for the feedback :)

sauwming (approved):

wosrediinanatour:
Maybe it may sense to split the enum/flags up in two or three. Could be more expressive.

nanangizz:
> Maybe it may sense to split the enum/flags up in two or three. Could be more expressive.

For example?

wosrediinanatour:
> For example?

```c
enum {
PJSUA_IPV4 = 1,
PJSUA_IPV6 = 1 << 1
};
enum {
PJSUA_NO_PREFERENCE = 0,
PJSUA_PREFER_IPV4 = 1,
PJSUA_PREFER_ IPV6 = 2
};
```

For me, having 2 enums make it more expressive and it easier to reason about the cases.

andreas-wehrmann:
Having two enums adds a further conditional and in fact makes it somewhat unclear (at least to me); I prefer just one enum and am fine with the suggestion made by @nanangizz [above](https://github.com/pjsip/pjproject/pull/4067#issue-2510158032).

What if neither PJSUA_IPV4 or PJSUA_IPV6 are set?
Having IPV4 to be the default seems to make sense because it has been the last couple of decades.

The suggested enum also reminded me of what you can set as socket option on dual-use sockets (IPV6_V6ONLY);
which made it look and feel familiar (at least to me).

Furthermore, how would you make it configurable in the app? Would you have two separate parameters? This seems unergonomic and not very user friendly (also remember, we're talking about the PJSUA layer here, which is pretty high level); and if you were to merge them together so that there's only one parameter? You'd probably end up with the suggestion above anyway.

nanangizz:
Looks like what you want is to separate settings of IP versions & fallback methods. Actually that is out of this PR scope.

Note that changing the enum names (and as mentioned by @andreas-wehrmann, perhaps also changing the setting into two separate params) may cause a backward compatibility issue, thing we try to avoid. Also note that the enum is shared by media & SIP signaling.

wosrediinanatour:
I am convinced :-)

PR #4080, opened 2024-09-17
Add swig option to include doc for java and python wrapper

trengginas:
This PR will add the option to include the doc from the `C++` code to the `java` and `python` wrapper.
It is nice to have since on some IDE, the doc will be shown as tooltip information.

Notes:
- Requires `SWIG 4` or later
- Only supports `Java` and `Python`
- On `Python`, the fields doc might get exported as property doc and not shown properly on the tooltip (e.g.: on VS Code). You can use the script in [here](https://github.com/user-attachments/files/16749800/transform_docstrings.txt) to transform the doc further.


nanangizz (approved):

sauwming (approved):

Axiinos:
Glad to see i was able to help with supplying the transformation script. 

It helps on both VSCode and Pycharm.

PR #4086, opened 2024-09-23
ifs log enhancements

krishanifs:
- In pj_assert put logs in non-debug modes
- Added PJ_ASSERT_LOG 
- In pjsua_reconfigure_logging gets log file name from config
- Adding more logs

### DIFF
--- a/pjlib/include/pj/assert.h
+++ b/pjlib/include/pj/assert.h
@@ -44,11 +44,44 @@
  *
  * @param expr      The expression to be evaluated.
  */
+#ifndef _DEBUG  
+#ifndef pj_assert
+#include "pj/log.h"
+#include <string.h>
+#ifdef _WIN32
+#define __FILENAME__ (strrchr(__FILE__, '\\') ? strrchr(__FILE__, '\\') + 1 : __FILE__)
+#else
+#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
+#endif
+#define pj_assert(expr) \
+        do { \
+                if (!(expr)) { PJ_LOG(1,(__FILENAME__, "Assert failed: " #expr)); } \
+        } while (0)
+#endif
+#else
 #ifndef pj_assert
 #   define pj_assert(expr)   assert(expr)
 #endif
+#endif
 
+/**
+ * @hideinitializer
+ * for all buils log the message
+ * Check during debug build that an expression is true. If the expression
+ * computes to false during run-time, then the program will stop at the
+ * offending statements.
+ * For release build, this macro only print message on the log.
+ * @param expr	    The expression to be evaluated.
+ * @param ...	    file name,The format string for the log message ("config.c", " PJ_VERSION: %s", PJ_VERSION)
+ */
+#ifndef PJ_ASSERT_LOG
+#include "pj/log.h"
+#define PJ_ASSERT_LOG(expr,...)    \
+            do { \
+                if (!(expr)) { PJ_LOG(1,(__VA_ARGS__)); assert(expr); } \
+            } while (0)
 
+#endif
 /**
  * @hideinitializer
  * If the expression yields false, assertion will be triggered
--- a/pjlib/src/pj/os_core_win32.c
+++ b/pjlib/src/pj/os_core_win32.c
@@ -54,7 +54,7 @@
 #else
 #   define LOG_MUTEX(expr)  PJ_LOG(6,expr)
 #endif
-
+#   define LOG_MUTEX_WARN(expr)  PJ_LOG(1,expr)
 #define THIS_FILE       "os_core_win32.c"
 
 /*
@@ -1093,10 +1093,18 @@ PJ_DEF(pj_status_t) pj_mutex_lock(pj_mutex_t *mutex)
         status = PJ_STATUS_FROM_OS(GetLastError());
 
 #endif
+    if (status == PJ_SUCCESS)
+    {
     LOG_MUTEX((mutex->obj_name, 
-              (status==PJ_SUCCESS ? "Mutex acquired by thread %s" : "FAILED by %s"),
+            "Mutex acquired by thread %s",
               pj_thread_this()->obj_name));
-
+    }
+    else
+    {
+        LOG_MUTEX_WARN((mutex->obj_name,
+           "FAILED by %s",
+            pj_thread_this()->obj_name));
+    }
 #if PJ_DEBUG
     if (status == PJ_SUCCESS) {
         mutex->owner = pj_thread_this();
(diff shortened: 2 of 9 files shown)


pjlib/include/pj/assert.h
```diff
@@ -44,11 +44,44 @@
  *
  * @param expr      The expression to be evaluated.
  */
+#ifndef _DEBUG  
+#ifndef pj_assert
+#include "pj/log.h"
+#include <string.h>
+#ifdef _WIN32
```
nanangizz, line 51: For Windows, it is possible to use MinGW, so need to confirm whether MinGW compiler has `_WIN32` and uses `\\` instead of `/` for path separator.

pjlib/src/pj/os_core_win32.c
```diff
@@ -1093,10 +1093,18 @@ PJ_DEF(pj_status_t) pj_mutex_lock(pj_mutex_t *mutex)
...
+            "Mutex acquired by thread %s",
               pj_thread_this()->obj_name));
-
+    }
+    else
+    {
+        LOG_MUTEX_WARN((mutex->obj_name,
+           "FAILED by %s",
+            pj_thread_this()->obj_name));
+    }
```
nanangizz, line 1107: As now it is separated, perhaps print the status code as well (e.g: using `PJ_PERROR` instead of `PJ_LOG`)?

pjmedia/src/pjmedia/stream.c
```diff
@@ -1546,7 +1546,7 @@ static pj_status_t put_frame_imp( pjmedia_port *port,
         pjmedia_rtp_hdr *rtp = (pjmedia_rtp_hdr*) channel->out_pkt;
 
         rtp->m = 1;
-        PJ_LOG(5,(stream->port.info.name.ptr,"Start talksprut.."));
+	PJ_LOG(5,(stream->port.info.name.ptr,"Starting talksprut.."));
```
nanangizz, line 1549: Minor: we no longer use tab for indentation (see #3210), please avoid using tab for indent (looks like some places in this PR are still using tab :)

pjsip/src/pjsip/sip_transaction.c
```diff
@@ -1952,6 +1952,7 @@ static void send_msg_callback( pjsip_send_state *send_state,
             tsx_update_transport(tsx, send_state->cur_transport);
 
             /* Update remote address. */
+		pj_assert(tdata->dest_info.cur_addr < (sizeof(tdata->dest_info.addr.entry)/sizeof(tdata->dest_info.addr.entry[0])));
```
nanangizz, line 1955: Minor: please limit column width to max 80 (old convention, just for consistency).

pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -834,6 +835,8 @@ pj_status_t pjsua_aud_channel_update(pjsua_call_media *call_med,
 
 on_return:
     pj_log_pop_indent();
+    if (status != PJ_SUCCESS)
+        PJ_LOG(2, (THIS_FILE, "pjsua_aud_channel_update failed"));
```
nanangizz, line 839: While adding this, use `PJ_PERROR()` to also print the status code?

pjsip/src/pjsua-lib/pjsua_core.c
```diff
@@ -728,7 +728,7 @@ PJ_DEF(pj_status_t) pjsua_reconfigure_logging(const pjsua_logging_config *cfg)
     pjsua_logging_config_dup(pjsua_var.pool, &pjsua_var.log_cfg, cfg);
 
     /* Redirect log function to ours */
-    pj_log_set_log_func( &log_writer );
+    pj_log_set_log_func( (cfg && cfg->cb && !cfg->log_filename.slen) ? cfg->cb : &log_writer );
```
nanangizz, line 731: - We missed `cfg != NULL` verification, better add that, e.g: `PJ_ASSERT_RETURN(cfg, PJ_EINVAL);`
- Re: skipping the `log_writer`, it may change the behavior of the lib (potential backward compatibility issue), could you specify the reason?

nanangizz (commented):
Thank you for the contributions :)

pjsip/src/pjsua-lib/pjsua_core.c
sauwming: Done for the first one, the second one reverted to current code.

pjsip/src/pjsip/sip_transaction.c
sauwming: Done. Modified using PJ_ARRAY_SIZE.

pjlib/include/pj/assert.h
```diff
@@ -44,11 +44,44 @@
...
+#define __FILENAME__ (strrchr(__FILE__, '\\') ? strrchr(__FILE__, '\\') + 1 : __FILE__)
+#else
+#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
```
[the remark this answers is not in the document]
sauwming: Removed this in #4098 due to a couple of reasons:
- platform difficulty (as commented above).
- if we want to extract the filename only and get rid of its path, perhaps we need to do it in the log instead. It has been previously discussed in #3889.

sauwming (commented):
Will continue this in #4098.

pjsip/src/pjsua-lib/pjsua_core.c
krishanifs: We need to pass a different log_writer call back function that write pj logs into our own application logs. how are we going to achieve that without this implementation.

sauwming: The function log_writer() will call you callback.

sauwming:
Since we already create a separate PR to continue this, you need to create a new PR if you want to push a new commit, otherwise it will potentially cause conflicts when merged.

krishanifs:
sorry about the late commit but (https://github.com/pjsip/pjproject/pull/4086/commits/41c00823080ccdbc2161e002a13f84bc0372bc3d) is also part of the log enhancement.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -2468,7 +2468,15 @@ PJ_DECL(pjsua_msg_data*) pjsua_msg_data_clone(pj_pool_t *pool,
...
-
+/*
+ * Instantiate pjsua application. Application must call this function before
+ * calling any other functions, to make sure that the underlying libraries
+ * are properly initialized. Once this function has returned success,
+ * application must call pjsua_destroy() before quitting.
+ *
+ * @return		PJ_SUCCESS on success, or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsua_create2(const pjsua_logging_config *log_cfg);
```
sauwming, line 2479: Since we can pass the log config to pjsua_init(), why is this new API needed?

pjsip/src/pjsua-lib/pjsua_core.c
krishanifs: this is an optimization we added. it directly call 'cfg->cb' if user has defined that; without calling it through log_writer. But it is OK to stick with current implementation.

Issue #4092, opened 2024-09-28
SWIG error when using make


mohannadjamal:
### Describe the bug

I get the following error on Windows 10 when running make in python SWIG
`C:\Users\MOHANN~1\AppData\Local\Temp\cciE8lgl.o:cc_mingw.c:(.text+0x5f): undefined reference to `pj_ansi_strxcpy'
collect2.exe: error: ld returned 1 exit status
make: *** [Makefile:43: gcc.exe] Error 1`

### Steps to reproduce

1. Add config_site.h to pjlib/include/pj/ including the config_site_sample.h
2. Build project using Visual Studio 2022
3. Go to pjsip-apps/src/swig/python/ 
4. Run make
5. Get the error

### PJSIP version

2.14.1

### Context

This issue happens on Windows 10, with SWIG version 4.0.0 or SWIG version 4.2.1, GCC version (MinGW.org GCC-6.3.0-1) 6.3.0 ,Python version 3.12.5. 

### Log, call stack, etc

```shell
'sed' is not recognized as an internal or external command,
operable program or batch file.
gcc -o gcc.exe cc_mingw.c
cc_mingw.c: In function 'find_gcc':
cc_mingw.c:44:5: warning: implicit declaration of function 'pj_ansi_strxcpy' [-Wimplicit-function-declaration]
     pj_ansi_strxcpy(spath, p, sizeof(spath));
     ^~~~~~~~~~~~~~~
C:\Users\MOHANN~1\AppData\Local\Temp\cciE8lgl.o:cc_mingw.c:(.text+0x5f): undefined reference to `pj_ansi_strxcpy'
collect2.exe: error: ld returned 1 exit status
make: *** [Makefile:43: gcc.exe] Error 1
```

Issue #4054, opened 2024-08-26
Build error for SWIG 4.2.1 Python 3.12.5, potential syntax error in importsym.py


Axiinos:
### Describe the bug

The build error is due to the original SWIG build, and it needs to be refreshed by rerunning the input/output formatter manually to build properly. running make in either swig folder or swig/python will produce errors.

### Steps to reproduce

The steps are more to fix the bug that is still in the source code.

1. install latest SWIG version by building it form their own source.
"Required libraries"
sudo apt install build-essential libpcre2-dev libpcre3-dev autoconf automake libtool bison git libboost-dev golang-go guile-2.2-dev nodejs node-gyp libwebkit2gtk-4.0-dev lua5.3 liblua5.3-dev mono-devel octave liboctave-dev default-jdk-headless php-cli php-dev python3-dev r-base ruby ruby-dev tcl-dev scilab libxml2-dev
"this can also be done in a env instead of system python."

2. install python 3.12.5

3. install setuptools with pip, it does not come with normal python install and is required.

4. install all the required build libraries that SWIG needs to run and rebuild the output file for pjsua2.i and its links that turn into pjsua2_wrap.ccp and pjsua2_wrap.h. 
"gcc is needed"

5.run ./configure CFLAGS="-fPIC" in the root project folder
"if you are running a virtual environment like me, run with the following arg : --build:x86_64-ubuntu-linux"

6. after it is done, run make dep && make to build the libraries.

7. return to the swig folder, and run swig to rebuild the generated swig install file. "swig -python -c++ -I../../../pjsip/include -I../../../pjlib/include -I../../../pjlib-util/include -I../../../pjmedia/include -I../../../pjnath/include -o python/pjsua2_wrap.cpp pjsua2.i"
this then rebuilds the wrapper properly, its needed due to the way that it needs to work between latest SWIG version and python 3.12* versions.

8. go to the python folder and do a "make" inside of it.
"this should now build properly, if you try to run make in swig, you will get an error about linking issues. the errors have been attached"

9. install the package with "pip install ." this will build it and install it as a wheel to either system or venv, depending on if you have activated a venv ofcourse.

### PJSIP version

2.14.1

### Context

System: Ubuntu 24.04 LTS in a HyperV environment, Windows 10 latest build.
"i specified in the ./configure what host to use, which was :  "./configure CFLAGS="-fPIC" --build=x86_64-ubuntu-linux" and it build succesfully."

[...]

PR #4093, opened 2024-10-01
Update for swig/python

Axiinos:
i made changes to make importsym more adaptive, along with checking if the system is win32, if it is, it will ask the user the registered path is correct, if not change it to the correct one, then edit and comment #include<pj/compat/socket.h> for them.

at the same time made changes to setup.py, including automatic execution of transform_docs.py, adding code comments for the user without any hassle.

I have tested the changes against linux and windows, and they seem to function as expected.

however i do still happen to hit a roadblock of the actual process function having issues with _fake_typedefs.h "see screenshot for reference.", which is likely just a setup issue on my end/versioning issues....

Linux version: 
image

Windows version:
image

there is also probably some methods i could have done better, tho i would like to hear feedback on the solution!

### DIFF
--- a/pjsip-apps/src/swig/importsym.py
+++ b/pjsip-apps/src/swig/importsym.py
@@ -22,6 +22,7 @@ import pycparser
 from pycparser import c_generator
 import sys
 import os
+from modules.win32fix import *
 
 def which(program):
 	import os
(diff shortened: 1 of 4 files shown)


sauwming:
I tested it here on Mac and it didn't work either. Unfortunately, I also don't know how to fix it.

Axiinos:
> I tested it here on Mac and it didn't work either. Unfortunately, I also don't know how to fix it.

I have tried multiple things, and yes i have followed the build steps to the letter from the guides "they seem sorta lacking when it comes to pjsua2 swig building, like no mention of c++xx version for compiler, which version of pycparser or python 2.x version to use" which could be the potential issue here?

would need more info from those who work on the swig/python side of things to clear it up.

when I use make, I do also get similar result as this bug report does: https://github.com/pjsip/pjproject/issues/4092#issue-2554146276

so if I really want to build the module, I have to just trust a potentially outdated symbolic link file and just force swig to process it.

sauwming:
Thanks for the contribution. But we can't merge it at the moment since there are still some issues with the current patch.

We understand that the existing `importsym` has some problems that you described in #4054, such as the hardcoded dir and compatibility with Python 3, but `importsym.py` is not supposed to be executed every time, so it's not urgent to be fixed/changed. Thus, I will close the PR and the issue.

PR #4098, opened 2024-10-04
Added more logs

sauwming:
To continue PR #4086.

Thanks to @krishanifs for contributing the original patch.

### DIFF
--- a/pjlib/include/pj/assert.h
+++ b/pjlib/include/pj/assert.h
@@ -40,14 +40,50 @@
  * Check during debug build that an expression is true. If the expression
  * computes to false during run-time, then the program will stop at the
  * offending statements.
- * For release build, this macro will not do anything.
+ * For release build, this macro will only log the assertion, while the
+ * program continues running.
  *
  * @param expr      The expression to be evaluated.
  */
-#ifndef pj_assert
-#   define pj_assert(expr)   assert(expr)
+#if PJ_DEBUG==0
+
+#   ifndef pj_assert
+#       include "pj/log.h"
+#       define pj_assert(expr) \
+            do { \
+                if (!(expr)) { \
+                    PJ_LOG(1, (__FILE__, "Assert failed: %s", #expr)); \
+                } \
+            } while (0)
+#   endif
+
+#else /* PJ_DEBUG != 0 */
+
+#   ifndef pj_assert
+#       define pj_assert(expr)   assert(expr)
+#   endif
+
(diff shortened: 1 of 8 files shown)


pjlib/include/pj/assert.h
```diff
@@ -40,14 +40,50 @@
...
+ * For all builds, log the message.
+ * Check during debug build that an expression is true. If the expression
+ * computes to false during run-time, then the program will stop at the
+ * offending statements.
+ * For release build, this macro only print message on the log.
+ * @param expr	    The expression to be evaluated.
+ * @param ...	    File name. The format string for the log message
+ *                  ("config.c", " PJ_VERSION: %s", PJ_VERSION)
+ */
+#ifndef PJ_ASSERT_LOG
```
sauwming, line 79: This doesn't seem to be used anywhere. Should we remove it?

pjlib/include/pj/assert.h
```diff
@@ -40,14 +40,50 @@
...
-#ifndef pj_assert
-#   define pj_assert(expr)   assert(expr)
+#if PJ_DEBUG==0
+
+#   ifndef pj_assert
+#       include "pj/log.h"
+#       define pj_assert(expr) \
+            do { \
+                if (!(expr)) { \
+                    PJ_LOG(1, (__FILE__, "Assert failed: %s", #expr)); \
```
sauwming, line 55: This is actually a good idea, but I think the reason we have never done this is that our logging functionality requires thread registration. So this may unintentionally crash the app in release build?

pjlib/include/pj/assert.h
nanangizz: As it looks simple and can be useful for some apps/users, perhaps it is okay?

nanangizz (approved):

trengginas (approved):

pjlib/include/pj/assert.h
krishanifs: with __FILE__ sometimes important part of the path is not visible

sauwming: As mentioned in the comment https://github.com/pjsip/pjproject/pull/4086#discussion_r1787067009, we had a precious PR #3889.
In summary, the proper solution is to increase PJ_LOG_SENDER _WIDTH. Alternatively, we should put the filename extraction mechanism in log.c/h since this doesn't only affect assertion but all the log printing.

Issue #4091, opened 2024-09-26
PJSUA: pjsua_call_send_dtmf ignores DTMF duration paramter for RFC2833


R-Jeske:
### Describe the bug

When using `pjsua_call_send_dtmf()` it is assumable, that it is possible to configure `duration` inside the `pjsua_call_send_dtmf_param` structure. But regardless which value in this structure is set, `PJMEDIA_DTMF_DURATION` or `PJMEDIA_DTMF_DURATION_MSEC` will be used which are defined in the `config.h`. This is a surprising behavior.

### Steps to reproduce

1. create a call with call_id id
2. send a DTMF character
```
      char buf[2];
      buf[0]='2';
      buf[1]='\0';

      const pj_str_t str=pj_str(buf);
      pjsua_call_send_dtmf_param dtmf_param;
      pjsua_call_send_dtmf_param_default(&dtmf_param);
      dtmf_param.digits = str;
      dtmf_param.duration = 500;
      dtmf_param.method = PJSUA_DTMF_METHOD_RFC2833;
      rc=pjsua_call_send_dtmf(id, &dtmf_param);
```
3. listen to `RTP EVENT` packets. The duration of the `SIP Event` is not 500.

### PJSIP version

2.14.1

### Context

This happens on all platforms. The behavior is caused by the fact, that `pjsua_call_send_dtmf` calls `pjsua_call_dial_dtmf` while losing the `duration` information. This could be avoided by setting an additional target-`duration` in `pjmedia_stream_dial_dtmf`, which should be compared with `digit->duration` in `create_dtmf_payload` if the target duration is set, instead of comparing it with `stream->dtmf_duration` which is defined by `PJMEDIA_DTMF_DURATION...`

### Log, call stack, etc

```shell
pjsua_call.c Call 0 sending DTMF 2 using RFC2833 method
   pjsua_aud.c Call 0 dialing DTMF 2 Sent DTMF 2
```

sauwming:
You're right. We should be more clear in the doc that the duration field is only used for SIP INFO method.

PR #4099, opened 2024-10-04
Fixed doc for sending DTMF

sauwming:
As reported in #4091.


nanangizz (approved):

R-Jeske:
This could be reverted due to #4124

PR #4105, opened 2024-10-14
Implement async callback for Android mediacodec

trengginas:
This is the current pattern for encoding/decoding using Android media codec:
1. check the input buffer availability
2. once an input buffer is available, input the data to encode/decode
3. check the output buffer availability
4. once the output buffer is available, it is the encoded/decoded data

Issues related to this pattern/when using polling:
1. We are using polling when checking the input/output buffer availability. This is necessary because the API won't block to wait for the buffer to be available.
2. A supply to input buffer doesn't always produce from the output buffer. Sometimes it needs to be fed more than one input buffer before the output is available, so it might waste some polling time/no guarantee of an output.
3. When the poll returns error, the codec might need to be restarted ([ref](https://github.com/pjsip/pjproject/blob/34500899bfc46d8c2660198984dbb3f127262ca3/pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp#L1250)). However, there are instances where this doesn't resolve the issue, and the codec fails to recover even after being re-created. 

This patch will use an async callback to the implementation.
1. Set new_input_buffer_avail() and new_output_buffer_avail() callback.
2. The callback will notify the index of the input/output buffer available to use. The buffer information will be stored in an atomic queue.
4. The encode/decode process will only use the buffer information provided from the callback on the queue.
5. This will eliminate the polling requirements.

Notes:
- The patch will use the atomic queue from `oboe_dev.cpp`
- The async callback is supported for Android API level 28 and later.
- There's noticeable lag/delay when using the Android MediaCodec backend. Unfortunately, this patch doesn't address the issue.
- To completely avoid the lag/delay issue, you can use an alternative video codec backend, such as OpenH264. This solution is also applicable to older devices.


pjmedia/include/pjmedia/atomic_queue.hpp
```diff
@@ -0,0 +1,81 @@
...
+     */
+    bool inc_ptr_read_if_not_yet(unsigned old_ptr);
+
+    /* Increment write pointer */
+    unsigned inc_ptr_write(unsigned old_ptr);
+
+    AtomicQueue() {}
+};
+
+#endif
```
nanangizz, line 81: - Headers in `include` dir are meant to be public API, so by convention it should use a specific prefix such as `pj` or `pjmedia`.
- If this is public API, IMO, the interface better uses C (as used by the rest), e.g: `pj_atomic_queue`, and perhaps part of PJLIB instead of PJMEDIA, note that the implementation can still use C++.

pjlib/include/pj/atomic_queue.h
```diff
@@ -0,0 +1,81 @@
...
+     */
+    bool inc_ptr_read_if_not_yet(unsigned old_ptr);
+
+    /* Increment write pointer */
+    unsigned inc_ptr_write(unsigned old_ptr);
+
+    pj_atomic_queue() {}
+};
+
+#endif
```
nanangizz, line 81: This is definitely still C++ (with C naming style :)
Please check `ioqueue_symbian.cpp` which use `ioqueue.h`.

pjlib/src/pj/atomic_queue.cpp
```diff
@@ -0,0 +1,99 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+#include <pj/assert.h>
+#include <pj/atomic_queue.h>
+#include <pj/string.h>
+
+#if defined(PJ_ANDROID) && PJ_ANDROID != 0
+
+//#include <android/log.h>
```
nanangizz, line 24: For tracing purpose, perhaps it can use the usual approach, e.g: using PJLIB log (instead of Android), tracing is by default disabled.

nanangizz (approved):

pjlib/include/pj/atomic_queue.h
```diff
@@ -0,0 +1,93 @@
...
+
+PJ_BEGIN_DECL
+
+/* Atomic queue (ring buffer) for single consumer & single producer.
+ *
+ * Producer invokes 'pj_atomic_queue_put(item)' to put an item to the back of
+ * the queue.
+ * Consumer invokes 'pj_atomic_queue_get(item)' to get an item from the head of
+ * the queue.
+ */
```
nanangizz, line 36: This does not seem to be part of the docs?
Note that the "single consumer & single producer" condition needs to be explicitly specified in the docs.

pjlib/src/pj/atomic_queue.cpp
```diff
@@ -0,0 +1,181 @@
...
+ * For producer, there is write pointer 'ptrWrite' that will be incremented
+ * every time a item is queued to the back of the queue. If the queue is
+ * almost full (the write pointer is right before the read pointer) the
+ * producer will forcefully discard the oldest item in the head of the
+ * queue by incrementing read pointer.
+ */
+PJ_DEF(pj_status_t) pj_atomic_queue_put(pj_atomic_queue_t *atomic_queue,
+                                        void *item)
+{
+    PJ_ASSERT_RETURN(atomic_queue && atomic_queue->aQ, PJ_EINVAL);
```
nanangizz, line 163: Check `item` too?
Also below.

pjlib/src/pj/atomic_queue.cpp
```diff
@@ -0,0 +1,181 @@
...
+    AtomicQueue    *aQ;
+};
+
+PJ_DEF(pj_status_t) pj_atomic_queue_create(pj_pool_t *pool,
+                                           unsigned max_item_cnt,
+                                           unsigned item_size,
+                                           const char* name,
+                                           pj_atomic_queue_t **atomic_queue)
+{
+    pj_atomic_queue_t *aqueue;
```
nanangizz, line 138: Add pre-verification/`PJ_ASSERT_RETURN()`?
Also below.

nanangizz (approved):

pjlib/include/pj/atomic_queue.h
```diff
@@ -0,0 +1,95 @@
...
+ * Destroy the Atomic Queue.
+ *
+ * @param atomic_queue  The Atomic Queue to be destroyed.
+ *
+ * @return              PJ_SUCCESS if success.
+ */
+PJ_DECL(pj_status_t) pj_atomic_queue_destroy(pj_atomic_queue_t *atomic_queue);
+
+/**
+ * Put a item to the back of the queue. If the queue is almost full
```
sauwming, line 67: an item.

What does "almost" full mean?

pjlib/src/pj/atomic_queue.cpp
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * Destructor
+     */
+    ~AtomicQueue()
+    {
+        delete [] buffer;
+    }
+
+    /**
+     * Get a item from the head of the queue
```
sauwming, line 56: an item.
Also below.

pjlib/src/pj/atomic_queue.cpp
```diff
@@ -0,0 +1,183 @@
...
+                ptrRead(0),
+                buffer(NULL),
+                name_(name)
+    {
+        buffer = new char[maxItemCnt_ * itemSize_];
+
+        /* Surpress warning when debugging log is disabled */
+        PJ_UNUSED_ARG(name_);
+
+        // PJ_LOG(5, (name_, "Created AtomicQueue: maxItemCnt=%d itemSize=%d",
```
sauwming, line 43: Rather than comment it, better disable it as @nanangizz suggested, such as using macro TRACE, so we can easily enable it when logging is needed.

sauwming (commented):
Also add in the PR description as to why this PR is needed, e.g. is there any performance issue with the current polling approach?
If yes, does using atomic queue manage to improve performance/reduce video lags (especially when compared to using other codec backend such as OpenH264)?

trengginas:
> Also add in the PR description as to why this PR is needed, e.g. is there any performance issue with the current polling approach? If yes, does using atomic queue manage to improve performance/reduce video lags (especially when compared to using other codec backend such as OpenH264)?

Unfortunately, the lag/delay issue still noticeable.
Tested scenario:
- Using the configuration specified [here](https://source.android.com/docs/core/media/low-latency-media)
- Using hw codec

sauwming (approved):

PR #4107, opened 2024-10-17
Assertion in sending BYE after transaction timeout

nanangizz:
When a request within dialog gets timeout, the invite session will terminate the session, and in the attempt of sending BYE, an assertion is raised:
```
src/pjsip-ua/sip_inv.c:287: pjsip_inv_add_ref: Assertion `inv && inv->ref_cnt' failed.
```
with call stack:
```
pjsip_inv_add_ref()
pjsip_inv_send_msg()
handle_uac_tsx_response()
mod_inv_on_tsx_state()
pjsip_dlg_on_tsx_state()
tsx_set_state()
tsx_on_state_calling()
tsx_timer_callback()
pj_timer_heap_poll()
```

After investigation, the transaction timeout handler in the invite sesion shift the invite sesion state to DISCONNECTED before sending BYE. The DISCONNECTED state will decrement the reference counter and may destroy the invite session.

The transaction timeout handler code is actually quite old, it did not raise an assertion because the invite session destructor is not as complex as now (basically it was only decrementing the dialog's session counter, code [here](https://github.com/pjsip/pjproject/blob/0b4c57b8de7e782fe4c18849f3511460cb84ab34/pjsip/src/pjsip-ua/sip_inv.c#L218-L228)).

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -4673,6 +4673,10 @@ static pj_bool_t handle_uac_tsx_response(pjsip_inv_session *inv,
         /* Send BYE */
         status = pjsip_dlg_create_request(inv->dlg, pjsip_get_bye_method(), 
                                           -1, &bye);
+        if (status == PJ_SUCCESS && tsx->status_text.slen) {
+            status = add_reason_warning_hdr(bye, tsx->status_code,
+                                            &tsx->status_text);
+        }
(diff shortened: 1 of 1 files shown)


sauwming (approved):

trengginas (approved):

TheQue42:
Would you mind adding a Reason header to the BYE, so that its easier to detect why it was triggered?

TheQue42:
Thanks for a quick fix :-)

PR #4113, opened 2024-10-24
Fix crash on incoming TLS call with only unsupported cipher offered

Mokolea:
A null-pointer exception occurs if we get an incoming TLS/SRTP call with only cipher offered which we do not support in our PJSIP application.

The cipher which are offered are: AEAD_AES_256_GCM, AEAD_AES_256_GCM_8
We have not activated these ciphers in our app

We are using PJSIP v2.14.1 on Linux and Windows and developing with C/C++ (pjsua/pjsua2)

This is just a quick-fix right where the null-pointer exception happens
Maybe this situation has to be properly handled in a different software layer/module of PJSIP

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_call.c
+++ b/pjsip/src/pjsua-lib/pjsua_call.c
@@ -2045,7 +2045,7 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
                                       NULL);
                 }               
 
-                if (call->inv->dlg) {
+                if (call->inv && call->inv->dlg) {
                     pjsip_inv_terminate(call->inv, sip_err_code, PJ_FALSE);
                 }
                 pjsip_dlg_dec_lock(dlg);


sauwming:
Well, the check was actually removed due to Coverity warning here:
https://github.com/pjsip/pjproject/commit/dd15679ae4909063aa8dbf2d4833e8dd4b97612d#diff-38326e41666262ac676da6de5d00a2f2ef0741b5ed851d1f1fae46aded951bf1L1964

The check was deemed unnecessary since there is an assignment `call->inv = inv` above:
https://github.com/pjsip/pjproject/blame/master/pjsip/src/pjsua-lib/pjsua_call.c#L1961
and no flow would change the value.

sauwming:
I tested here with our sample app, and it will correctly return the error:
`10:33:22.103           pjsua_call.c  ..Error creating SDP answer: Unsupported SRTP crypto-suite (PJMEDIA_SRTP_ENOTSUPCRYPTO) [status=220222]`

So not sure why it crashed on your app. To investigate, we need more info:
- the stack trace
- the PJSIP log, which shows the problematic incoming INVITE

sauwming (approved):

nanangizz (approved):

PR #4118, opened 2024-10-28
Add full support for SHA-256 and SHA-512-256 digest algorithms

gtjoseph:
There are no breaking changes for this work however several structures
were extended with new fields.  See below.

In order to use the new algorithms, you MUST set the new
pjsip_cred_info.ext.algorithm_type field to the appropriate value
when the credential data type is PJSIP_CRED_DATA_DIGEST and when
acting as a server, you must also use pjsip_auth_srv_challenge2()
to send challenges so you can specify algorithms other than MD5.

Summary of changes:

* Added enum pjsip_auth_algorithm_type which list all digest algorithms
supported.

* Added struct pjsip_auth_algorithm which defines parameters for each
algorithm including its IANA name, OpenSSL name, digest length and
digest string representation length.

* Added pjsip_auth_algorithm_type to the pjsip_cred_info structure
so the digest algorithm can be specified when the cred data type
is PJSIP_CRED_DATA_DIGEST.

* Added pjsip_auth_algorithm_type to the pjsip_cached_auth_hdr
structure so we can match on specific algorithm.

* Added functions pjsip_auth_get_algorithm_by_type(),
pjsip_auth_get_algorithm_by_iana_name(), and
pjsip_auth_is_digest_algorithm_supported() to find and search
for supported algorithms.

* Added pjsip_authorization_hdr to the pjsip_auth_lookup_cred_param
structure so we can look up credentiials by specific algorithm.

* Added the pjsip_auth_srv_challenge2() function that takes
a pjsip_auth_algorithm_type so users can create challenges with
specific algorithms instead of defaulting to MD5.

* pjsip_auth_create_digest() was heavily refactored to use the
new algorithm_type contained in pjsip_cred_info to determine the
algorithm to use when creating the digest.  The function is now
generic and can use any supported algorithm.  If OpenSSL isn't
available, it will fall back to the internal MD5 implementation.

* pjsip_auth_create_digestSHA256() is now marked as deprecated and
simply calls the new function with PJSIP_AUTH_ALGORITHM_SHA256.

* sip_auth_client.c and sip_auth_server.c were refactored to support
multiple digest algorithms.

* sip_auth_client was updated to allow the AKEv2-MD5 algorithm
to pass through to the callback specified in pjsip_cred_info.

* A bug was fixed with the PJSIP_AUTH_ALLOW_MULTIPLE_AUTH_HEADER
option where the default setting of 0 prevented sip_auth_client
from responding to WWW/Proxy-Authenticate headers from different
realms.  The RFCs state that this behavior should be allowed.
The comment for this option in sip_config.h was also updated to
indicate that setting this option to 1 is probably not a good idea
for security reasons.

Resolves: #4119

### DIFF
--- a/pjsip/include/pjsip/sip_auth.h
+++ b/pjsip/include/pjsip/sip_auth.h
@@ -42,11 +42,65 @@ PJ_BEGIN_DECL
  * @{
  */
 
-/** Length of digest MD5 string. */
+/**
+ * Length of digest MD5 string.
+ * \deprecated Use #pjsip_auth_algorithm::digest_str_length instead.
+ */
 #define PJSIP_MD5STRLEN         32
 
-/** Length of digest SHA256 string. */
+/**
+ * Length of digest SHA256 string.
+ * \deprecated Use #pjsip_auth_algorithm::digest_str_length instead.
+ */
 #define PJSIP_SHA256STRLEN      64
+
+/**
+ * The length of the buffer needed to contain the largest
+ * supported algorithm's digest.
+ */
+#define PJSIP_AUTH_MAX_DIGEST_BUFFER_LENGTH      64
+
+/**
+ * Digest Algorithm Types.
+ * \warning These entries must remain in order with
+ * no gaps and with _NOT_SET = 0 and _COUNT as the last entry.
+ *
+ * The MD5, SHA-256, and SHA-512/256 algorithms are described
+ * in RFC 7616 and RFC 8760.
+ * The AKA algorithms are described in RFC 3310 and RFC 4169
+ * and 3GPP TS 33.203.
+ */
+typedef enum pjsip_auth_algorithm_type
+{
+    PJSIP_AUTH_ALGORITHM_NOT_SET = 0,  /**< Algorithm not set.          */
+    PJSIP_AUTH_ALGORITHM_MD5,          /**< MD5 algorithm.              */
+    PJSIP_AUTH_ALGORITHM_SHA256,       /**< SHA-256 algorithm.          */
+    PJSIP_AUTH_ALGORITHM_SHA512_256,   /**< SHA-512/256 algorithm       */
+    PJSIP_AUTH_ALGORITHM_AKAV1_MD5,    /**< AKA v1 with MD5 algorithm.  */
+    PJSIP_AUTH_ALGORITHM_AKAV2_MD5,    /**< AKA v2 with MD5 algorithm.  */
+    PJSIP_AUTH_ALGORITHM_COUNT,        /**< Number of algorithms.       */
+} pjsip_auth_algorithm_type;
+
+
+/**
+ * Authentication Digest Algorithm
+ *
+ * This structure describes a digest algorithm used in
+ * SIP authentication.
+ *
+ */
+typedef struct pjsip_auth_algorithm
+{
+    pjsip_auth_algorithm_type algorithm_type; /**< Digest algorithm type     */
+    pj_str_t iana_name;                       /**< IANA/RFC name used in
+                                                   SIP headers               */
+    const char *openssl_name;                 /**< The name used by OpenSSL's
+                                                   EVP_get_digestbyname()    */
+    unsigned digest_length;                   /**< Length of the raw digest
+                                                   in bytes                  */
+    unsigned digest_str_length;               /**< Length of the digest HEX
+                                                   representation            */
+} pjsip_auth_algorithm;
 
 
 /** Type of data in the credential information in #pjsip_cred_info. */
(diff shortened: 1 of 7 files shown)


gtjoseph:
This PR is in draft mode because we're still testing with Asterisk and older versions of OpenSSL.  I wanted to get it up to receive comments sooner rather than later though.

gtjoseph:
Oh you're kidding.  We still need to support building without OpenSSL?

gtjoseph:
I still need to clean up the OpenSSL/nonm-OpenSSL things and remove some testing things I did to get a successful Windows build but it's looking much better.

gtjoseph:
Rebased on current master.  All asterisk tests pass.

gtjoseph:
The last force push contained just documentation updates in sip_auth.h and sip_auth_aka.h

sauwming (approved):

pjsip/src/pjsip/sip_auth_client.c
```diff
@@ -162,158 +196,107 @@ static void digestNtoStr(const unsigned char digest[], int n, char *output)
...
-    if ((cred_info->data_type & PASSWD_MASK) == PJSIP_CRED_DATA_PLAIN_PASSWD)
+    if (PJSIP_CRED_DATA_IS_PASSWD(cred_info))
     {
+        AUTH_TRACE_((THIS_FILE, " Using plain text password for %.*s digest",
+                (int)algorithm->iana_name.slen, algorithm->iana_name.ptr));
         /***
-         *** ha1 = SHA256(username ":" realm ":" password)
+         *** ha1 = (digest)(username ":" realm ":" password)
          ***/
-#if OPENSSL_VERSION_NUMBER < 0x30000000L
```
nanangizz, line 306: Looks like all codes for OpenSSL < 3 is removed, we expect it is still compatible with OpenSSL 1.1.x, can you confirm or have you tested?

pjsip/include/pjsip/sip_auth.h
```diff
@@ -115,9 +179,15 @@ struct pjsip_cred_info
                                      challenges.                            */
     pj_str_t    scheme;         /**< Scheme (e.g. "digest").                */
     pj_str_t    username;       /**< User name.                             */
-    int         data_type;      /**< Type of data (0 for plaintext passwd). */
+    int         data_type;      /**< Type of data \ref pjsip_cred_data_type */
```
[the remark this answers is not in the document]
gtjoseph: Because that's the way it was historically and I didn't want to change the signature with this patch.

pjsip/src/pjsip/sip_auth_client.c
```diff
@@ -162,158 +196,107 @@ static void digestNtoStr(const unsigned char digest[], int n, char *output)
...
-        pj_memcpy( ha1, cred_info->data.ptr, cred_info->data.slen );
-    } else {
-        pj_assert(!"Invalid data_type");
-        pj_bzero(result->ptr, result->slen);
-        result->slen = 0;
+    if (!(result && nonce && uri && realm && cred_info && method)) {
+        PJ_LOG(4, (THIS_FILE,
+                "result, nonce, uri, realm, cred_info and method are all required"));
         return PJ_EINVAL;
     }
```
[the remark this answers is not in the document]
gtjoseph: Odd.  I remember using PJ_ASSERT_RETURN here.  I wonder why I changed it to an "if". Anyway, will be fixed shortly.

pjsip/src/pjsip/sip_auth_client.c
gtjoseph: The code was a bit weird when it came to OpenSSL versions.  The things that were only enabled with OpenSSL >= 3.0.0 (EVP_MD, EVP_Digest, etc) were all available with earlier versions back to 1.1.0 so I'm not sure why those SHA256_ alternatives were there.

I did build Asterisk with this patch on CentOS7 with OpenSSL 1.1.0 and tested both UAC and UAS authentication, including detecting that OpenSSL 1.1.0 didn't support SHA-512-256.  Everything worked fine.

trengginas (approved):

pjsip/src/pjsip/sip_auth_client.c
nanangizz: From a manual page [here](https://docs.openssl.org/1.0.2/man3/EVP_DigestInit/#history), `EVP_DigestInit_ex()` seems to be introduced in 0.9.7. So perhaps those alternatives were actually for OpenSSL older than 0.9.7, which is too old by now, especially for a security library :)

Also thanks for the confirmation that it works fine with 1.1.0.

nanangizz (approved):

PR #4121, opened 2024-10-30
Removed unnecessary prechecks in async conference bridge operations

nanangizz:
After #3928, reported that when connecting & disconnecting ports requests are issued in a very short interval (e.g: when auto answer with resp code 200 is enabled in pjsua app), the disconnect request may fail which cause the ports to remain connected (undesirable). The failure in disconnection request happens because there is precheck of connection availability. Those prechecks are still be performed in the operation execution, but not in the operation request. This PR will remove such prechecks in all async operations requests.

This PR also contain some minor updates, e.g:
- add check for possible failure in memory allocation (for operation entry)
- use ```on_return``` label to clean up error handling.
- logging update: add word "video" when referring a video port's slot without port's name.

### DIFF
--- a/pjmedia/src/pjmedia/vid_conf.c
+++ b/pjmedia/src/pjmedia/vid_conf.c
@@ -594,8 +594,11 @@ PJ_DEF(pj_status_t) pjmedia_vid_conf_remove_port( pjmedia_vid_conf *vid_conf,
 {
     vconf_port *cport;
     op_entry *ope;
+    pj_status_t status = PJ_SUCCESS;
 
-    PJ_LOG(5,(THIS_FILE, "Port %d remove requested", slot));
+    pj_log_push_indent();
+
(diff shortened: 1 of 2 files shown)


pjmedia/src/pjmedia/vid_conf.c
```diff
@@ -594,8 +594,11 @@ PJ_DEF(pj_status_t) pjmedia_vid_conf_remove_port( pjmedia_vid_conf *vid_conf,
 {
     vconf_port *cport;
     op_entry *ope;
+    pj_status_t status = PJ_SUCCESS;
+
+    pj_log_push_indent();
 
-    PJ_LOG(5,(THIS_FILE, "Port %d remove requested", slot));
+    PJ_LOG(5,(THIS_FILE, "Remove video port %d remove requested", slot));
```
sauwming, line 601: Minor: duplicate 'remove'

sauwming (approved):

trengginas (approved):

PR #4132, opened 2024-11-05
Add siprec extension: handle incoming INVITE with siprec

sorooshm78:
I had a Session Record Client (SRC) that utilized the SIPREC module from the OpenSIPS SIP server, and I intended to develop a Session Record Server (SRS) capable of supporting SIPREC. The goal was for the SRS to handle requests effectively. However, since PJSIP does not support the SIPREC extension, I added SIPREC extension support to PJSIP.

image (3)

### INVITE Request
```
Request-Line: INVITE sip:192.168.10.20 SIP/2.0

Message Headers:
    Via: SIP/2.0/UDP 192.168.10.20:5060;branch=z9hG4bK656b.7f03f537.0
    To: <sip:192.168.10.20>
    From: <sip:192.168.10.20>;tag=85c717edf7d282027df52a83293e4957-ce21
    User-Agent: OpenSIPS (3.5.0-dev (x86_64/linux))
    Require: siprec  
    Content-Type: multipart/mixed; boundary=OSS-unique-boundary-42
    Contact: <sip:192.168.21.89:5060>; +sip.src
    ...

Session Description Protocol (SDP):
    Version: 0
    Media Description (m): audio 35001 RTP/AVP 8 0 101
    Media Attributes (a): 
        sendonly
        label:0
    ...
    Media Description (m): audio 35002 RTP/AVP 8 101
    Media Attributes (a): 
        sendonly
        label:1
    ... 
```

### 420 Bad Extension Response
```
Status-Line: SIP/2.0 420 Bad Extension

Message Header
    ...
    Unsupported: siprec
    Supported: replaces, 100rel, timer, norefersub, trickle-ice
    Content-Length:  0
```

## Senario with siprec extension:
In the SIPREC module of PJSIP, when an INVITE is received with the header `Require: siprec` and two media attribute labels in the offer SDP, it can generate an appropriate `200 OK` response that includes those two media attribute labels in the SDP answer.

### INVITE Request
```
Request-Line: INVITE sip:192.168.10.20 SIP/2.0

Message Headers:
    Via: SIP/2.0/UDP 192.168.10.20:5060;branch=z9hG4bK656b.7f03f537.0
    To: <sip:192.168.10.20>
    From: <sip:192.168.10.20>;tag=85c717edf7d282027df52a83293e4957-ce21
    User-Agent: OpenSIPS (3.5.0-dev (x86_64/linux))
    Require: siprec  
    Content-Type: multipart/mixed; boundary=OSS-unique-boundary-42
    Contact: <sip:192.168.21.89:5060>; +sip.src
    ...

Session Description Protocol (SDP):
    Version: 0
    Media Description (m): audio 35001 RTP/AVP 8 0 101
    Media Attributes (a): 
        sendonly
        label:0
    ...
    Media Description (m): audio 35002 RTP/AVP 8 101
    Media Attributes (a): 
        sendonly
        label:1
    ... 
```

### 200 OK Response
```
Message Headers:
    Via: SIP/2.0/UDP 192.168.10.20:5060; received=192.168.10.20; branch=z9hG4bK656b.7f03f537.0
    From: <sip:192.168.10.20>; tag=85c717edf7d282027df52a83293e4957-ce21
    To: <sip:192.168.10.20>; tag=7de0bfd1-94b6-4462-929f-53955d386d91
    Allow: PRACK, INVITE, ACK, BYE, CANCEL, UPDATE, INFO, SUBSCRIBE, NOTIFY, REFER, MESSAGE, OPTIONS
    Supported: replaces, 100rel, timer, siprec, norefersub
    ...

Session Description Protocol (SDP):
    Version: 0
    Session Name (s): pjmedia
    Media Description (m): audio 4000 RTP/AVP 8 101
    Media Attributes (a): 
        label:0
        recvonly
    ...

    Media Description (m): audio 4002 RTP/AVP 8 101
    Media Attributes (a): 
        label:1
        recvonly
    ...

```


SmWork78:
I couldn't debug the `Bitrise` error and figure out where the issue in my code is. Please guide me to resolve the problem.

sorooshm78:
I couldn't debug the Bitrise error and figure out where the issue in my code is. Please guide me to resolve the problem.

sauwming:
Thank you for the submission.

The support seems to be pretty basic, i.e. it allows incoming request with siprec to be accepted (instead of declined). There are a few concerns though:
- The setting `enable_multimedia` seems misleading. PJSIP already supports multimedia, i.e. audio and video, even multiple audio and multiple video.
- The patch seems to assume that only audio is supported.

sorooshm78:
Thank you for your response.  
I have a few questions:  
1. What do you mean by "pretty basic," and how can I perform more extensive checks?  
2. Is your issue only with the name `enable_multimedia`? For example, if it is changed to `enable_multiple_audio`, would that resolve the issue? Or, if PJSIP itself supports multiple audio streams, where can I configure it?  
3. Yes, for now, it only works with audio, but adding video support shouldn’t be too difficult.  

In the SIPREC protocol, it is possible to send information in the body in XML format with the `Content-Type: application/rs-metadata+xml`. I am working on another commit to add this feature.

sauwming:
1. It's basic in the sense that it is more directed towards the UAS, only to accept INVITE that requires siprec. But it lacks other verification, for example, in the RFC, it is mentioned that:
`When an SRS receives a new INVITE, the SRS MUST only consider the SIP
   session as an RS when both the "+sip.src" feature tag and the
   "siprec" option tag are included in the INVITE request`.
While for the UAC, it also lacks features such as how to specify the labelling, etc.

2. PJSIP supports multiple audio media. If you specify `call->opt.aud_cnt` to 2, there will be two audio streams.

sorooshm78:
1. I will add more checks for the necessary headers in the next few days.
2. I have made some changes to the code. My assumption is that pjsip should work without any issues or changes in both SIPREC mode and normal mode. This means when an INVITE with a `SIPREC` header (I will add other checks as well) arrives, the call should be established correctly. Specifically, we will modify the value of `call->opt.aud_cnt` and add a `label` to its SDP. When a call is established in normal mode (without SIPREC), it should also work without any issues or changes, and should not cause any malfunction in pjsip.
3. I am implementing SIPREC as a UAS, and if I have time, I will also implement UAC.

sorooshm78:
I have added more checks based on [RFC 7866](https://www.ietf.org/rfc/rfc7866.txt) , but my code lacks appropriate comments, which I will add in the next commit.

sorooshm78:
I added the appropriate comment.

sauwming:
Just to give you a heads up, we are preparing for 2.15 release within a few weeks time, so we won't be able to merge this until then.
But I will give you some preliminary comments about the current patch.

build.symbian/pjsip_ua.mmp
```diff
@@ -40,6 +40,7 @@ SOURCE	sip_replaces.c
 SOURCE	sip_xfer.c
 SOURCE	sip_100rel.c
 SOURCE	sip_timer.c
+SOURCE	sip_siprec.c
```
sauwming, line 43: no need to modify Symbian build. It's deprecated and no longer used.

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,109 @@
...
+
+
+/**
+ * Create a=label attribute.
+ *
+ * @param media          The SDP media.
+ *
+ * @return               SDP label attribute.
+ */
+PJ_DEF(pjmedia_sdp_attr*) pjmedia_sdp_attr_get_label(pjmedia_sdp_media *sdp_media);
```
sauwming, line 44: PJ_DECL, also for all others below.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -1802,6 +1802,44 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
...
+        }
+
+        goto on_return;
+    }
+        
+    /* Check if request supports PJSIP_INV_REQUIRE_SIPREC. If so
+     * set the number of call active streams to the number of media in the SDP offer.
+     */
+    if(options & PJSIP_INV_REQUIRE_SIPREC){
+        call->opt.aud_cnt = offer->media_count;
```
sauwming, line 1840: This is what I mentioned about the patch assuming audio only. For example, if you check the example in the RFC 7866: https://www.rfc-editor.org/rfc/rfc7866.html#section-7.1.1, there are two videos and two audios in the incoming offer, so this code is obviously incorrect.

build.symbian/pjsip_ua.mmp
sorooshm78: fix it

pjsip/include/pjsip-ua/sip_siprec.h
sorooshm78: fix it

pjsip/src/pjsua-lib/pjsua_call.c
sorooshm78: I will fix the issue within the next few days.

sorooshm78:
Resolved the audio count issue and added video support. Successfully tested with a video call using MicroSIP.

pjsip/src/pjsua-lib/pjsua_call.c
sorooshm78: Resolved the issue

SmWork78:
In the SIPREC protocol, it is now possible to send information in the body in XML format with the Content-Type: application/rs-metadata+xml. This feature has been added.

sorooshm78:
In the SIPREC protocol, it is now possible to send information in the body in XML format with the Content-Type: application/rs-metadata+xml. This feature has been added.

andreas-wehrmann:
Just out of curiosity: Why didn't you implement this on top of PJSUA? Or do you not use PJSUA but rather pjsip?

I'm asking because I implemented an SRC on top of PJSUA and a little SRS test tool as well.
That way PJSUA handles all the "annoying low level" media stuff and I just have to inspect the media parameters (and the XML in case of an SRS) in the callbacks.
As for setting/manipulating SDP, this can be done in the existing callbacks also.

This also cleanly separates the different tasks and concerns of each module
(i.e. the SIPREC module is able to understand and handle snapshot requests).

sorooshm78:
> Just out of curiosity: Why didn't you implement this on top of PJSUA? Or do you not use PJSUA but rather pjsip?
> 
> I'm asking because I implemented an SRC on top of PJSUA and a little SRS test tool as well. That way PJSUA handles all the "annoying low level" media stuff and I just have to inspect the media parameters (and the XML in case of an SRS) in the callbacks. As for setting/manipulating SDP, this can be done in the existing callbacks also.
> 
> This also cleanly separates the different tasks and concerns of each module (i.e. the SIPREC module is able to understand and handle snapshot requests).

I initially tried to do the same thing, but it didn’t work. I was using OpenSIPS for the SRC, and in the INVITE message that OpenSIPS sent, there was a "Require: siprec" header, which caused a "bad extension" error since PJSIP didn’t support SIPREC. I didn’t want to make any changes in OpenSIPS.

andreas-wehrmann:
I see; didn't adding "siprec" with [pjsip_endpt_add_capability()](https://docs.pjsip.org/en/latest/api/generated/pjsip/group/group__PJSIP__ENDPT.html#_CPPv426pjsip_endpt_add_capabilityP14pjsip_endpointP12pjsip_moduleiPK8pj_str_tjA_K8pj_str_t) work for you?

sorooshm78:
> I see; didn't adding "siprec" with [pjsip_endpt_add_capability()](https://docs.pjsip.org/en/latest/api/generated/pjsip/group/group__PJSIP__ENDPT.html#_CPPv426pjsip_endpt_add_capabilityP14pjsip_endpointP12pjsip_moduleiPK8pj_str_tjA_K8pj_str_t) work for you?


Yes, it works, but I tried to implement SIPREC in the appropriate place as an extension, similar to other PJSIP extensions like replaces, 100rel, timer and so on.

sauwming:
@nanangizz, do you think we should integrate the patch, before we start a detailed review?

nanangizz:
> @nanangizz, do you think we should integrate the patch, before we start a detailed review?

I'd vote for yes, it appears to be a widely supported extension.

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -1546,6 +1555,7 @@ PJ_DEF(pj_status_t) pjsip_inv_verify_request3(pjsip_rx_data *rdata,
             } else if (pj_stricmp(&req_hdr->values[i], &STR_REPLACES)==0) {
                 pj_bool_t supp;
                 
+
```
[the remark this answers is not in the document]
sorooshm78: fix it

andreas-wehrmann:
Will there be a possibility to build PJ without this extension by means of a `#define` for example?

sorooshm78:
> Will there be a possibility to build PJ without this extension by means of a `#define` for example?

The current implementation does not affect non-SIPREC calls, and calls are established without any issues. The added code checks the condition without introducing significant overhead, and it can be ignored. I will incorporate your suggestion later if I have the time.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -2083,6 +2115,16 @@ pj_bool_t pjsua_call_on_incoming(pjsip_rx_data *rdata)
     }
 */
 
+    /* Check if request supports PJSIP_INV_REQUIRE_SIPREC. If so
```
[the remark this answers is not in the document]
sorooshm78: Yes, I think the same. fixed it.

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,135 @@
...
+
+
+/*
+ * Counts the number of audio and video streams in the SDP
+ * 
+ * @param sdp         The SDP media.
+ * @param maudcnt     media audio count.
+ * @param mvidcnt     media video count
+ */
+PJ_DECL(void) pjsip_siprec_count_media(pjmedia_sdp_session *sdp,
```
sauwming, line 122: Remove this since the API is not used anywhere, and also shouldn't be the responsibility of siprec.

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Find siprec metadata information from the message body
+ * with "rs-metadata+xml" Content-Type.
+ *
+ * @param pool               Pool to allocate memory.
+ * @param body               The message body.
+ * @param metadata           The siprec metadata
+ */
+PJ_DECL(void) pjsip_siprec_find_metadata(pj_pool_t *pool,
```
sauwming, line 110: A more appropriate name should be:
`get_metadata()`.

So there are two alternatives:
a. The metadata as a return value:
PJ_DECL(pj_str_t *) pjsip_siprec_get_metadata()
or
b. The function returns `pj_status_t` to indicate any error during parsing and/or retrieval, and the metadata is put as output parameter (need to specify this in the doc):
PJ_DECL(pj_status_t) pjsip_siprec_get_metadata(..., pj_str_t *metadata)

pjsip/src/pjsip-ua/sip_siprec.c
```diff
@@ -0,0 +1,276 @@
...
+    if (status != PJ_SUCCESS)
+        return status;
+
+    /* Register deinit module to be executed when PJLIB shutdown */
+    if (pjsip_endpt_atexit(endpt, &pjsip_siprec_deinit_module) != PJ_SUCCESS) {
+        /* Failure to register this function may cause this module won't 
+         * work properly when the stack is restarted (without quitting 
+         * application).
+         */
+        pj_assert(!"Failed to register Session Timer deinit.");
```
sauwming, line 51: Incorrect assertion and log message.

pjsip/src/pjsip-ua/sip_siprec.c
```diff
@@ -0,0 +1,276 @@
+#include <pjsip-ua/sip_siprec.h>
```
sauwming, line 1: Put copyright and licensing info at the top of the file, also in the header file.

You can put your name as an author (optional, if you wish so).

pjsip/src/pjsip-ua/sip_siprec.c
```diff
@@ -0,0 +1,276 @@
...
+    pj_list_init(&res_hdr_list);
+
+    /* Find Require header */
+    req_hdr = (pjsip_require_hdr*) pjsip_msg_find_hdr_by_name(rdata->msg_info.msg, &str_require, NULL);
+
+    if(!req_hdr || (pjsip_siprec_verify_require_hdr(req_hdr) == PJ_FALSE)){
+        return PJ_SUCCESS;
+    }
+    
+    /* Find Conatct header */
```
sauwming, line 133: Contact,
also the variable name

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * @param endpt         The SIP endpoint instance.
+ *
+ * @return              PJ_SUCCESS if module is successfully initialized.
+ */
+PJ_DECL(pj_status_t) pjsip_siprec_init_module(pjsip_endpoint *endpt);
+
+
+/**
+ * Create a=label attribute.
```
sauwming, line 38: Create or get?

Based on your implementation, I think you need both, but in separate APIs.

Get is to get the label, create is to add it in the response.

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * @param media          The SDP media.
+ *
+ * @return               SDP label attribute.
+ */
+PJ_DECL(pjmedia_sdp_attr*) pjmedia_sdp_attr_get_label(pjmedia_sdp_media *sdp_media);
```
nanangizz, line 44: This does not look right, a PJMEDIA SDP API in PJSIP, should remove this or move it to PJMEDIA SDP.
The implementation looks simple/one-liner, so I think you can just use the generic API `pjmedia_sdp_media_find_attr()` directly without this API.

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,135 @@
...
+ * @defgroup PJSIP_SIPREC SIP Session Recording Protocol (siprec) support (RFC 7866 - Session Recording Protocol in SIP)
+ * @brief SIP Session Recording Protocol support (RFC 7866 - Session Recording Protocol in SIP)
+ * @{
+ *
+ * \section PJSIP_SIPREC_REFERENCE References
+ *
+ * References:
+ *  - <A HREF="http://www.ietf.org/rfc/rfc7866.txt">RFC 7866: Session Recording Protocol (siprec)
+ *    in the Session Initiation Protocol (SIP)</A>
+ */
```
nanangizz, line 23: Minor: column width < 80 chars.
This also applies for all comments & codes below.

pjsip/src/pjsip-ua/sip_siprec.c
```diff
@@ -0,0 +1,276 @@
...
+PJ_DEF(pj_status_t) pjsip_siprec_verify_request(pjsip_rx_data *rdata, 
+                                                pj_str_t *metadata,
+                                                pjmedia_sdp_session *sdp_offer,                                     
+                                                unsigned *options,
+                                                pjsip_dialog *dlg,
+                                                pjsip_endpoint *endpt,
+                                                pjsip_tx_data **p_tdata)
+{
+    pjsip_require_hdr *req_hdr;
+    pjsip_contact_hdr *conatct_hdr;
```
nanangizz, line 111: Minor: should be `contact_hdr`?

nanangizz (commented):
From a quick look, the implementation is rather small parts of siprec features, i.e: verification and response SIP & SDP signaling for recording server (CMIIW), I think this should be clarified in the PR title & desc, even better if it also (briefly) describes what to do next to implement a recording server with full functionality.

pjsip/include/pjsip-ua/sip_siprec.h
sorooshm78: remove it

pjsip/src/pjsip-ua/sip_siprec.c
sorooshm78: This was a bad mistake. it fixed.

pjsip/src/pjsip-ua/sip_siprec.c
sorooshm78: This was a bad mistake. it fixed.

pjsip/include/pjsip-ua/sip_siprec.h
sorooshm78: I removed the pjmedia_sdp_attr_get_label function, but I am still using the pjmedia_sdp_media_find_attr function siprec.c

pjsip/include/pjsip-ua/sip_siprec.h
sorooshm78: fix it

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,116 @@
...
+
+
+/**
+ * Checks if there is an attribute label for each media in the SDP.
+ * 
+ * @param media          The SDP media.
+ * 
+ * @return               PJ_TRUE if a label exists in the SDP.
+ */
+PJ_DECL(pj_status_t) pjsip_siprec_verify_sdp_attr_label(pjmedia_sdp_session *sdp);
```
nanangizz, line 54: Well, this is still strange :)
Why not just put the implementation inline?

pjsip/include/pjsip-ua/sip_siprec.h
sorooshm78: fix it

pjsip/include/pjsip-ua/sip_siprec.h
sorooshm78: You're right, I fixed it.
Do you think I should define `STR_LABEL` inside the function or at the top of the header file?
One more thing, what do you think about making the `pjsip_siprec_verify_require_hdr` function inline?

sorooshm78: Did I understand your point correctly?

Or are you suggesting that instead of creating the `pjsip_siprec_verify_sdp_attr_label` function, I should place it directly at `pjsip_siprec_verify_request`?

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -1951,12 +1951,12 @@ static pj_status_t call_media_init_cb(pjsua_call_media *call_med,
         goto on_return;
     }
 
-    /* Check if media is deinitializing */
-    if (call_med->call->async_call.med_ch_deinit || !call_med->tp) {
-        status = PJ_ECANCELLED;
-        goto on_return;
-    }
-
+    /* Check if media is deinitializing */
```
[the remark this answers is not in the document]
sorooshm78: fix it

pjsip/include/pjsip-ua/sip_siprec.h
nanangizz: Re: `STR_LABEL`, it should be in the `.c` file instead of header file. If it is used only once or twice, putting it in inside the function is okay, otherwise, better in the (top of) file.

Re: `pjsip_siprec_verify_require_hdr()`, it is okay inline or not.

Re: `pjsip_siprec_verify_sdp_attr_label()`, this is more like an SDP API, so better the latter approach, remove the API & place the implementation directly inside `pjsip_siprec_verify_request()`.

sorooshm78: fix it.

sorooshm78: The review of labels is a more general topic than SIPREC. In other words, when we have multiple media, there must always be a label, and later on, I will remove this review from SIPREC and place it in a more general section.

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,129 @@
+/* 
+ * Copyright (C) 2024 Green and Silver Leaves. (https://github.com/BSVN)
```
sauwming, line 2: Please put Teluu copyright line here as well (so there will be two lines).
Example: https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia-audiodev/android_jni_dev.c

Also in siprec.c

pjsip/src/pjsip-ua/sip_siprec.c
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    /* Checks if the body exists */
+    if (!rdata->msg_info.msg->body) {
+        code = PJSIP_SC_BAD_REQUEST;
+        warn_text = "SIPREC INVITE must have a body";
+        goto on_return; 
+    }
+
+    /* Currently, SIPREC INVITE requests without SPD are not supported. */
```
sauwming, line 148: SDP

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,129 @@
...
+                                                pjsip_tx_data **p_tdata);
+
+
+/**
+ * Find siprec metadata information from the message body
+ * with "rs-metadata" Content-Type.
+ *
+ * @param pool               Pool to allocate memory.
+ * @param body               The message body.
+ * @param metadata           The siprec metadata
```
sauwming, line 111: Specify that this is output parameter.

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3003,6 +3003,15 @@ pj_status_t pjsua_media_channel_create_sdp(pjsua_call_id call_id,
             continue;
         }
 
+        /* Check if request supports PJSIP_INV_REQUIRE_SIPREC. If so
+         * Get label attribute in SDP offer and add label attribute to SDP answer
+         */
+        if (call->inv && (call->inv->options & PJSIP_INV_REQUIRE_SIPREC)) {
+            const pj_str_t STR_LABEL_ATTR = {"label", 5};
+            m->attr[m->attr_count++] = pjmedia_sdp_media_find_attr(
```
sauwming, line 3011: I think you need to duplicate/clone the attribute, instead of just taking it from the offer, since you can't rely on the offer's lifetime.

pjsip/include/pjsip-ua/sip_siprec.h
sorooshm78: fix it

pjsip/src/pjsip-ua/sip_siprec.c
sorooshm78: fix it

pjsip/include/pjsip-ua/sip_siprec.h
sorooshm78: fix it

sauwming (commented):
I think we need an option use_siprec in Pjsua to enable/disable this. Check use_timer and require_100rel for examples.
By default, it should be disabled. 

Also need the corresponding option in pjsua2.

sorooshm78:
> I think we need an option use_siprec in Pjsua to enable/disable this. Check use_timer and require_100rel for examples. By default, it should be disabled.
> 
> Also need the corresponding option in pjsua2.

Sure, I will add it in the next few days.

pjsip/src/pjsua-lib/pjsua_media.c
sorooshm78: fix it

pjmedia/include/pjmedia/sdp.h
```diff
@@ -370,6 +370,18 @@ PJ_DECL(pjmedia_sdp_attr*) pjmedia_sdp_attr_create_ssrc(pj_pool_t *pool,
...
+/**
+ * Create a=label attribute.
+ *
+ * @param pool          Pool to create the attribute.
+ * @param attr          Attribute to create.
+ *
+ * @return              SDP label attribute.
+ */
+PJ_DECL(pjmedia_sdp_attr*) pjmedia_sdp_attr_create_label( pj_pool_t *pool,
+                                                pjmedia_sdp_attr *attr);
```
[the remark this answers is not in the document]
sorooshm78: fix it

pjsip/src/pjsua2/account.cpp
```diff
@@ -634,6 +634,7 @@ void AccountConfig::toPj(pjsua_acc_config &ret) const
     ret.call_hold_type          = callConfig.holdType;
```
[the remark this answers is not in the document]
sorooshm78: fix it

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -2343,6 +2369,9 @@ typedef struct pjsua_config
      */
     pjsua_sip_timer_use use_timer;
 
+
```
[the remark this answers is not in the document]
sorooshm78: fix it

pjsip/include/pjsip-ua/sip_inv.h
```diff
@@ -388,7 +388,21 @@ enum pjsip_inv_option
...
 
+    /**
+     * Indicate support for siprec
+     */
+    PJSIP_INV_SUPPORT_SIPREC      = 2048,
+    
+    /**
+     * Indicate not support for siprec
+     */
+    PJSIP_INV_NOT_SUPPORT_SIPREC  = 4096,
```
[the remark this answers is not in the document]
sorooshm78: `PJSIP_INV_NOT_SUPPORT_SIPREC` is required because when the user sets `use_siprec` to `PJSUA_SIP_SIPREC_INACTIVE`, I reject SIPREC calls with a Bad Extension error.

pjsip/include/pjsip-ua/sip_siprec.h
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Checks if the INVITE request is SIPREC.
+ * 
+ * @param rdata         The incoming request to be verified.
+ * 
+ * @return               If the request is SIPREC, return PJ_TRUE;
+ *                       otherwise, return PJ_FALSE.
+ */
+PJ_DECL(pj_status_t) pjsip_siprec_check_request(pjsip_rx_data *rdata);
```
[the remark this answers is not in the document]
sorooshm78: I did this for code cleanliness; otherwise, I would have had to handle it inside the verify_request() method using a variable.

pjsip/include/pjsip-ua/sip_inv.h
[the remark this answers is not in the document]
sauwming: By default, no flag means unsupported. Check the other flags above.

pjsip/include/pjsip-ua/sip_siprec.h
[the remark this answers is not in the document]
sauwming: Since it's only used internally in `sip_siprec.c`, you can create the function there and it won't be necessary to create a public API.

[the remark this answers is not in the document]
sorooshm78: You're right. fix it

pjsip/src/pjsua-lib/pjsua_media.c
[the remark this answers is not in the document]
sorooshm78: I had previously fixed the issue, but it has returned and is now resolved again.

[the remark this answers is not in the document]
sorooshm78: I had previously fixed the issue, but it has returned and is now resolved again.

sorooshm78:
I don't know what the issue with `windows-with-video-libvpx-schannel-unit-test-1` is. Could you please guide me?

sorooshm78:
I have considered three modes for `use_siprec`:

1. **optional**: In this mode, calls will be established regardless of whether they are SIPREC or not.  
2. **mandatory**: In this mode, only SIPREC calls are allowed. If a regular call is attempted, it will result in a "Bad Request" error.  
3. **inactive**: In this mode, only regular calls are allowed. If a SIPREC call is attempted, it will result in a "Bad Extension" error.

I use two variables to determine the state for the three modes above.

|            | PJSIP_INV_SUPPORT_SIPREC | PJSIP_INV_REQUIRE_SIPREC |
|:----------:|:-------------------------:|:-------------------------:|
| **inactive** | no                      | no                       |
| **option**   | yes                     | no                       |
| **mandatory**| yes                     | yes                      |

When `use_siprec` is set to **inactive**, neither the `PJSIP_INV_SUPPORT_SIPREC` flag nor the `PJSIP_INV_REQUIRE_SIPREC` flag is enabled, as shown in the table above.

And in the code below, I attempt to return a "Bad Extension" error:

`pjsip/src/pjsip-ua/sip_inv.c`

``` c
else if (!(*options & PJSIP_INV_SUPPORT_SIPREC) && 
        pj_stricmp(&req_hdr->values[i], &STR_SIPREC)==0)
    {
        unsupp_tags[unsupp_cnt++] = req_hdr->values[i];
    
    }
```

In the code above, when the `PJSIP_INV_SUPPORT_SIPREC` option is not set, the execution enters the `if` block and returns a "Bad Extension" error.

The pjsip_inv_verify_request3 function is called three times for each call. The first time, it includes the desired options, but for the next two calls, the option value is zero. Therefore, I added the following code to resolve this issue:

 `verify_request` function in `pjsip/src/pjsua-lib/pjsua_call.c`
```c
if (status == PJ_SUCCESS) {
    unsigned options = 0;

    /* Add SIPREC support to prevent the "bad extension" error */
    options |= PJSIP_INV_SUPPORT_SIPREC; // add this line

    /* Verify that we can handle the request. */
    status = pjsip_inv_verify_request3(rdata,
                                        call->inv->pool_prov, &options, 
                                        offer, answer, NULL, 
                                        pjsua_var.endpt, response);
```


If there's a better solution to address the issue mentioned, please guide me. 

and I don't know what the issue with `windows-with-video-libvpx-schannel-unit-test-1` is. Could you please guide me?

---

In my previous commit, to resolve this issue, I added `PJSIP_INV_NOT_SUPPORT_SIPREC`. 

`pjsip/src/pjsip-ua/sip_inv.c`

``` c
else if (!(*options & PJSIP_INV_NOT_SUPPORT_SIPREC) && 
        pj_stricmp(&req_hdr->values[i], &STR_SIPREC)==0)
    {
        unsupp_tags[unsupp_cnt++] = req_hdr->values[i];
    
    }
```

sauwming:
For the failed CI test, we have fixed it in #4234, so don't worry about it.

pjmedia/include/pjmedia/sdp.h
```diff
@@ -664,6 +676,10 @@ PJ_DECL(pjmedia_sdp_media*) pjmedia_sdp_media_clone_deactivate(
                                                 const pjmedia_sdp_media *rhs);
 
 
+PJ_DEF(pjmedia_sdp_attr*) pjmedia_sdp_attr_get_label(
```
sauwming, line 679: PJ_DECL

The signature also needs to match other `pjmedia_sdp_attr_get_*()`
For example:
`PJ_DECL(pj_status_t) pjmedia_sdp_attr_get_rtcp(const pjmedia_sdp_attr *attr,
                                               pjmedia_sdp_rtcp_attr *rtcp);
`

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -1539,14 +1543,19 @@ PJ_DEF(pj_status_t) pjsip_inv_verify_request3(pjsip_rx_data *rdata,
             {
                 rem_option |= PJSIP_INV_REQUIRE_100REL;
 
+            } else if (!(*options & PJSIP_INV_SUPPORT_SIPREC) &&
```
sauwming, line 1546: this seems to be incorrect, or at least not uniform with other extension checking (such as 100rel, timer, etc).

pjmedia/include/pjmedia/sdp.h
sorooshm78: I removed the `pjmedia_sdp_attr_get_label` method because methods like`pjmedia_sdp_attr_get_*` are used for parsing a generic SDP attribute

pjsip/src/pjsip-ua/sip_inv.c
sorooshm78: The operation of the SIPREC mode with 100rel and the timer is different. I tried to explain it in the text above.

sorooshm78:
> I have considered three modes for `use_siprec`:
> 
> 1. **optional**: In this mode, calls will be established regardless of whether they are SIPREC or not.
> 2. **mandatory**: In this mode, only SIPREC calls are allowed. If a regular call is attempted, it will result in a "Bad Request" error.
> 3. **inactive**: In this mode, only regular calls are allowed. If a SIPREC call is attempted, it will result in a "Bad Extension" error.
> 
> I use two variables to determine the state for the three modes above.
> 
> PJSIP_INV_SUPPORT_SIPREC	PJSIP_INV_REQUIRE_SIPREC
> **inactive**	no	no
> **option**	yes	no
> **mandatory**	yes	yes
> When `use_siprec` is set to **inactive**, neither the `PJSIP_INV_SUPPORT_SIPREC` flag nor the `PJSIP_INV_REQUIRE_SIPREC` flag is enabled, as shown in the table above.
> 
> And in the code below, I attempt to return a "Bad Extension" error:
> 
> `pjsip/src/pjsip-ua/sip_inv.c`
> 
> ```c
> else if (!(*options & PJSIP_INV_SUPPORT_SIPREC) && 
>         pj_stricmp(&req_hdr->values[i], &STR_SIPREC)==0)
>     {
>         unsupp_tags[unsupp_cnt++] = req_hdr->values[i];
>     
>     }
> ```
> 
> In the code above, when the `PJSIP_INV_SUPPORT_SIPREC` option is not set, the execution enters the `if` block and returns a "Bad Extension" error.
> 
> The pjsip_inv_verify_request3 function is called three times for each call. The first time, it includes the desired options, but for the next two calls, the option value is zero. Therefore, I added the following code to resolve this issue:
> 
> `verify_request` function in `pjsip/src/pjsua-lib/pjsua_call.c`
> 
> ```c
> if (status == PJ_SUCCESS) {
>     unsigned options = 0;
> 
>     /* Add SIPREC support to prevent the "bad extension" error */
>     options |= PJSIP_INV_SUPPORT_SIPREC; // add this line
> 
>     /* Verify that we can handle the request. */
>     status = pjsip_inv_verify_request3(rdata,
>                                         call->inv->pool_prov, &options, 
>                                         offer, answer, NULL, 
>                                         pjsua_var.endpt, response);
> ```
> 
> If there's a better solution to address the issue mentioned, please guide me.
> 
> and I don't know what the issue with `windows-with-video-libvpx-schannel-unit-test-1` is. Could you please guide me?
> 
> In my previous commit, to resolve this issue, I added `PJSIP_INV_NOT_SUPPORT_SIPREC`.
> 
> `pjsip/src/pjsip-ua/sip_inv.c`
> 
> ```c
> else if ((*options & PJSIP_INV_NOT_SUPPORT_SIPREC) && 
>         pj_stricmp(&req_hdr->values[i], &STR_SIPREC)==0)
>     {
>         unsupp_tags[unsupp_cnt++] = req_hdr->values[i];
>     
>     }
> ```

@sauwming , @nanangizz If there's a better solution to address the issue mentioned, please guide me.

pjsip/src/pjsip-ua/sip_inv.c
sauwming: To me, they look the same.

So, basically in this particular block, if local doesn't support a particular extension, while remote requires it, it should go to this block:
```
            } else if (!pjsip_endpt_has_capability(endpt, PJSIP_H_SUPPORTED,
                                                   NULL, &req_hdr->values[i]))
            {
                /* Unknown/unsupported extension tag!  */
                unsupp_tags[unsupp_cnt++] = req_hdr->values[i];
            }
```
and will cause us to respond with "Bad Extension".

Therefore, I believe the siprec checking should be more or less similar with other extensions, such as 100rel or timer. I don't understand why it should be treated differently.

sorooshm78: Thank you for your response. I will try to be like the other modules. However, before implementing, I would like to check one thing with you and apply it if you agree.

When the user sets `use_siprec` to `inactive`, what action should I take? 
I think that instead of using "bad extension" I should do nothing. This means that whether the request includes `Required: siprec` or not, I take no action. In other words, I neither check for the siprec condition nor add a label.

sorooshm78: If you have any better suggestions, please let me know. Otherwise, with your approval, I will implement this approach.

sauwming: I'm not quite clear with the question.

If user sets `use_siprec` to inactive, and remote signals `Require: siprec` in the offer, then I believe the correct behaviour should be to reject with "Bad extension".

sorooshm78: When I call the `pjsip_endpt_add_capability` function within the `pjsip_siprec_init_module`, `pjsip_endpt_has_capability` always returns `true`, so it never enters the condition below:
```
            } else if (!pjsip_endpt_has_capability(endpt, PJSIP_H_SUPPORTED,
                                                   NULL, &req_hdr->values[i]))
            {
                /* Unknown/unsupported extension tag!  */
                unsupp_tags[unsupp_cnt++] = req_hdr->values[i];
            }
```

sorooshm78: The `pjsip_endpt_remove_capability` function does not exist to be called when `inactive` that the value of the `pjsip_endpt_has_capability` function becomes `false`.

sauwming: Ah okay, now I understand what you mean.
It looks like you're right, so no modification required for this and I will mark the conversation resolved.

nanangizz (commented):
As the implementation is not a full functionality, I think the PR title should be changed to represent to coverage, e.g: something like "Add siprec extension: handle incoming INVITE with siprec", I can do this if you want :)

Also better if the PR desc briefly describes what to do next to implement a recording server with full functionality, unfortunately I think this needs to be done by you as I believe you've done it in real app.

Thanks for the submission & all the works!

sorooshm78:
To implement simple Session Recording Server (SRS) using the SIPREC module, you can follow the steps below.I am using `pjsua2` to implement SRS.

## Configure
To configure the SIPREC module, you can proceed as in the code below.
```c++
    AccountConfig accountConfig;
    accountConfig.callConfig.siprecUse = PJSUA_SIP_SIPREC_OPTIONAL;
```

## Audio
Also, to save audio, you can use the following code.
```c++
class MyCall : public pj::Call
{
    void onCallMediaState(pj::OnCallMediaStateParam& params)
    {
        pj::CallInfo callInfo = getInfo();
        for (unsigned media_index = 0; media_index < callInfo.media.size(); media_index++)
        {
            if (callInfo.media[media_index].type == PJMEDIA_TYPE_AUDIO)
            {
                pj::AudioMedia audioMedia = getAudioMedia(media_index);
                recorder.createRecorder(path);
                audioMedio.startTransmit(recorder);
            }
        }
    }
};
```

## Metadata
And to access metadata, you can use the following code.
```c++
class MyCall : public pj::Call
{
    void saveMetadata()
    {
        int id = getId();
        pjsua_call *call = &pjsua_var.calls[id];
        std::string metadata = std::string(call->siprec_metadata.ptr, call->siprec_metadata.slen);
    }    
};
```

## Unsupported Features in the SIPREC Module
A list of items that are not present in the SIPREC module:

* `Update` requests are not supported.
* `Reinvite` requests are not supported.
* Metadata in `BYE` requests is not supported.
* `UAS` can request the `UAC` to send metadata, which is not supported.
* Currently, the `SIPREC` module supports only `UAC` mode.


I will try to implement some of the above items in another PR.

nanangizz (approved):

Issue #985, opened 2009-11-05
Write to freed memory area with IoCompletionPort (IOCP) ioqueue backend may lead to application crash


pjsipbot:
_2009-11-05 13:40:55: @bennylp_ _created the issue on trac ticket [985](https://trac.pjsip.org/repos/ticket/985)_

Using IoCompletionPort (IOCP) ioqueue backend, which is the **default** backend on Win32 on Release build, may lead to write access to memory area that has been freed before. This potentially could crash the application, or random memory corruption.

This symptom can be detected with Visual Studio 2005 (but not with Visual Studio 6).

Explanation about this bug is as follows.

 1. When socket is closed, Windows automatically unregisters the socket from the IOCP. Any pending operations to the socket are also cancelled.
 1. But when the next time the IOCP is polled with GetQueuedCompletionStatus(), Windows would still report an event to the pending operations that have been cancelled above.
 1. If the memory that corresponds to the `WSAOVERLAPPED` structure that was registered for the asynchronous operation has been freed (for example, the pool is released), Windows will write some values to these area that was previously occupied by the `WSAOVERLAPPED` structure.

___
_2010-02-25 11:54:47: @bennylp_ _changed **milestone** from **release-1.6** to **release-1.7**_

___
_2010-04-28 01:16:16: @bennylp_ _changed **status** from **new** to **closed**_

___
_2010-04-28 01:16:16: @bennylp_ _set **resolution** to **worksforme**_

___
_2010-04-28 01:16:16: @bennylp_ _changed **milestone** from **release-1.7** to **release-1.5.5**_

___
_2010-04-28 01:16:16: @bennylp_ _commented_

Turns out this has been "fixed" in milestone:release-1.6 by ticket #1015, by disabling IOCP backend altogether.

Changing the milestone to 1.5.5 and closing the ticket.

___
_2010-04-28 01:16:52: @bennylp_ _commented_

(The above should read milestone:release-1.5.5 instead of 1.6).

nanangizz:
I am afraid currently no, we haven't got an idea for solution to the problem described in the bug explanation above.

nanangizz:
> I have tested CancelIoEx(), not working, GetQueuedCompletionStatus() still report an event after closesocket.

Thanks for letting us know, haven't got a chance to setup test env myself.

[...]

PR #4136, opened 2024-11-07
iocp: fix crash, GetQueuedCompletionStatus() write freed WSAOVERLAPPED memory

jimying:
Try to fix issue #985. The idea is to call [CancelIoEx()](https://learn.microsoft.com/en-us/windows/win32/fileio/cancelioex-func) for the unregistering socket/key to cancel all pending operations of the key. However, as `CancelIoEx()` is basically asynchronous, this also makes the key unregistration asynchronous, so here are some consequences:
- The memory for operation key must not be freed until the operation cancellation is completed, so instead of using app supplied operation key, the IOCP will use an internal copy.
- Key must always have a group lock (can be supplied by app), which will handle key's resources release.

### DIFF
--- a/pjlib/src/pj/ioqueue_winnt.c
+++ b/pjlib/src/pj/ioqueue_winnt.c
@@ -754,7 +754,6 @@ static void cancel_all_pending_op(pj_ioqueue_key_t *key)
             if (rc)
                 PJ_PERROR(2, (THIS_FILE, PJ_RETURN_OS_ERROR(GetLastError()), "cancel io error"));
         }
-        fnCancelIoEx(key->hnd, NULL);
         pj_mutex_unlock(key->mutex);
     }
 }
--- a/pjlib/src/pj/ioqueue_winnt.c
+++ b/pjlib/src/pj/ioqueue_winnt.c
@@ -28,7 +28,7 @@
 #include <pj/errno.h>
 #include <pj/compat/socket.h>
 
-#define THIS_FILE "iocp"
+#define THIS_FILE "ioq_winnt"
 
 typedef BOOL (WINAPI *FnCancelIoEx)(HANDLE hFile, LPOVERLAPPED lpOverlapped);
 static FnCancelIoEx fnCancelIoEx = NULL;
@@ -154,7 +154,6 @@ struct pj_ioqueue_t
 
     pj_ioqueue_key_t  active_list;
     pj_ioqueue_key_t  free_list;
-    pj_ioqueue_key_t  closing_list;
 
     /* These are to keep track of connecting sockets */
 #if PJ_HAS_TCP
@@ -165,9 +164,6 @@ struct pj_ioqueue_t
     pj_ioqueue_key_t *connecting_keys[MAXIMUM_WAIT_OBJECTS+1];
 #endif
 };
-
-/* Prototype */
-static void scan_closing_keys(pj_ioqueue_t *ioqueue);
 
 #if PJ_HAS_TCP
 /*
@@ -190,6 +186,7 @@ static void ioqueue_on_accept_complete(pj_ioqueue_key_t *key,
                         SO_UPDATE_ACCEPT_CONTEXT, 
                         (char*)&key->hnd, 
                         sizeof(SOCKET));
+    (void)status;
     /* SO_UPDATE_ACCEPT_CONTEXT is for WinXP or later.
      * So ignore the error status.
      */
@@ -405,7 +402,6 @@ PJ_DEF(pj_status_t) pj_ioqueue_create2(pj_pool_t *pool,
      */
     pj_list_init(&ioqueue->active_list);
     pj_list_init(&ioqueue->free_list);
-    pj_list_init(&ioqueue->closing_list);
 
     /* Preallocate keys according to max_fd setting, and put them
      * in free_list.
@@ -445,7 +441,7 @@ PJ_DEF(pj_status_t) pj_ioqueue_create2(pj_pool_t *pool,
 
     *p_ioqueue = ioqueue;
 
-    PJ_LOG(4, ("pjlib", "WinNT IOCP I/O Queue created (%p)", ioqueue));
+    PJ_LOG(4, (THIS_FILE, "WinNT IOCP I/O Queue created (%p)", ioqueue));
     return PJ_SUCCESS;
 }
 
@@ -478,13 +474,6 @@ PJ_DEF(pj_status_t) pj_ioqueue_destroy( pj_ioqueue_t *ioqueue )
     /* Destroy reference counters */
     key = ioqueue->active_list.next;
     while (key != &ioqueue->active_list) {
-        pj_atomic_destroy(key->ref_count);
-        pj_mutex_destroy(key->mutex);
-        key = key->next;
-    }
-
-    key = ioqueue->closing_list.next;
-    while (key != &ioqueue->closing_list) {
         pj_atomic_destroy(key->ref_count);
         pj_mutex_destroy(key->mutex);
         key = key->next;
@@ -552,11 +541,6 @@ PJ_DEF(pj_status_t) pj_ioqueue_register_sock2(pj_pool_t *pool,
     PJ_ASSERT_RETURN(pool && ioqueue && cb && key, PJ_EINVAL);
 
     pj_lock_acquire(ioqueue->lock);
-
-    /* Scan closing list first to release unused keys.
-     * Must do this with lock acquired.
-     */
-    scan_closing_keys(ioqueue);
 
     /* If safe unregistration is used, then get the key record from
      * the free list.
@@ -683,7 +667,7 @@ static void decrement_counter(pj_ioqueue_key_t *key)
         pj_time_val_normalize(&key->free_time);
 
         pj_list_erase(key);
-        pj_list_push_back(&key->ioqueue->closing_list, key);
+        pj_list_push_back(&key->ioqueue->free_list, key);
 
         pj_lock_release(key->ioqueue->lock);
     }
@@ -702,6 +686,7 @@ static struct pending_op *alloc_pending_op(pj_ioqueue_key_t *key, pj_ioqueue_op_
     op->pending_key.overlapped.wsabuf.buf = (CHAR *)buf;
     op->pending_key.overlapped.wsabuf.len = (ULONG)len;
 
+    pj_atomic_inc(key->ref_count);
     pj_mutex_lock(key->mutex);
     pj_list_push_back(&key->pending_list, op);
     pj_mutex_unlock(key->mutex);
@@ -716,6 +701,7 @@ static void release_pending_op(pj_ioqueue_key_t *key, struct pending_op *op)
     pj_list_erase(op);
     pj_mutex_unlock(key->mutex);
     pj_pool_release(op->pool);
+    decrement_counter(key);
 }
 
 static void cancel_all_pending_op(pj_ioqueue_key_t *key)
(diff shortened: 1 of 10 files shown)


nanangizz:
> Note: this only fix the issue when PJ_IOQUEUE_HAS_SAFE_UNREG=1; PJ_IOQUEUE_HAS_SAFE_UNREG=0 still has memory error

When `PJ_IOQUEUE_HAS_SAFE_UNREG==1`, the key won't be reused for 500ms (configurable via `PJ_IOQUEUE_KEY_FREE_DELAY`), this is a bit risky actually, e.g: on high CPU load, the cancellation may take longer?

nanangizz:
Tried to running `pjlib-test` with this PR patch on VS2005 and pool debugging enabled (so pool uses normal `malloc/free()`, by setting `PJ_POOL_DEBUG` in `config_site.h`), I got an assertion:
```
 	_wassert(const wchar_t * expr=0x004d1e90, const wchar_t * filename=0x004d1cf0, unsigned int lineno=588) Line 212	C
 	pj_ioqueue_register_sock2(pj_pool_t * pool=0x023a29b8, pj_ioqueue_t * ioqueue=0x023a2a2c, long sock=388, pj_grp_lock_t * grp_lock=0x00000000, void * user_data=0x00000002, const pj_ioqueue_callback * cb=0x0019fb54, pj_ioqueue_key_t * * key=0x0019fbec) Line 588 + 0x2c bytes	C
 	pj_ioqueue_register_sock(pj_pool_t * pool=0x023a29b8, pj_ioqueue_t * ioqueue=0x023a2a2c, long sock=388, void * user_data=0x00000002, const pj_ioqueue_callback * cb=0x0019fb54, pj_ioqueue_key_t * * key=0x0019fbec) Line 665 + 0x1f bytes	C
 	unregister_test(const pj_ioqueue_cfg * cfg=0x0019fd34) Line 554 + 0x23 bytes	C
 	udp_ioqueue_test_imp(const pj_ioqueue_cfg * cfg=0x0019fd34) Line 1190 + 0x9 bytes	C
 	udp_ioqueue_test() Line 1255 + 0x9 bytes	C
 	test_inner() Line 171 + 0x2a bytes	C
 	test_main() Line 245 + 0x5 bytes	C
```
Not sure if this is the same issue, but this assertion does not happen when using ioqueue select.

nanangizz:
> @nanangizz no this patch, Is there this assert?

Yes, same assert without this patch.

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -698,6 +715,49 @@ static void decrement_counter(pj_ioqueue_key_t *key)
...
+{
+    if (!fnCancelIoEx)
+        fnCancelIoEx = (FnCancelIoEx)GetProcAddress(GetModuleHandle(PJ_T("Kernel32.dll")), "CancelIoEx");
+    if (fnCancelIoEx) {
+        struct pending_op *op;
+        pj_mutex_lock(key->mutex);
+        for (op = key->pending_list.next; op != &key->pending_list; op = op->next) {
+            BOOL rc = fnCancelIoEx(key->ioqueue->iocp, (LPOVERLAPPED)&op->pending_key);
+            if (rc)
+                PJ_PERROR(2, (THIS_FILE, PJ_RETURN_OS_ERROR(GetLastError()), "cancel io error"));
```
nanangizz, line 755: - actually when `rc ==1`, it succeeds
- shouldn't the handle `key->hnd` instead of `key->ioqueue->iocp`?

sauwming:
Hi @jimying, please let us know whether you plan to incorporate @nanangizz suggestions above.
In my opinion, the first two (enforcing`PJ_IOQUEUE_HAS_SAFE_UNREG` and zero pending-op)  are important and shouldn't require much modification.
While for the last suggestion (memory pool), I think it's optional if you don't have the time necessary to implement it. Just put a note in the code for future optimisation.

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -133,12 +133,9 @@ struct pj_ioqueue_key_t
     int                 connecting;
 #endif
 
-#if PJ_IOQUEUE_HAS_SAFE_UNREG
     pj_atomic_t        *ref_count;
```
sauwming, line 136: I don't think we should remove all the code related to `PJ_IOQUEUE_HAS_SAFE_UNREG`. I believe what @nanangizz meant was that we need to make sure `PJ_IOQUEUE_HAS_SAFE_UNREG` is set to 1.

Since by default PJ_IOQUEUE_HAS_SAFE_UNREG is already 1, IMO adding a check below should be sufficient, just to make sure that users don't accidentally disable it:
#if PJ_IOQUEUE_HAS_SAFE_UNREG == 0
    /* IOCP only works with PJ_IOQUEUE_HAS_SAFE_UNREG enabled, otherwise we will get memory error in ... */
    #error "PJ_IOQUEUE_HAS_SAFE_UNREG must be enabled to use ioqueue IOCP"
#endif

Another alternative is to force enable it in `pjlib/config.h`, something like:
```
#if PJ_IOQUEUE_IMP == PJ_IOQUEUE_IOCP
#   define PJ_IOQUEUE_HAS_SAFE_UNREG    1
#endif
```
But this is more complicated since we need to define `PJ_IOQUEUE_IMP` similar to `PJ_SSL_SOCK_IMP`.

sauwming: It's still there in all other ioq backends.

nanangizz: Perhaps unsafe unreg option will not be used for a very long time in IOCP (until there is a way to cancel op immediately), so perhaps it is okay to remove that.

Btw, I think I will try to join the development, may I commit to the branch directly @jimying ?

nanangizz:
I think this is ready for review @jimying , @sauwming , @trengginas.

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -48,6 +49,13 @@
 /* For GetAcceptExSockaddrs() on MSVC2005 */
 #pragma comment(lib, "mswsock.lib")
 
+#if 0
+#  define TRACE(args) PJ_LOG(1,args)
```
sauwming, line 53: Usually we use level 3 for trace.

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -117,6 +137,7 @@ struct pj_ioqueue_key_t
 {
     PJ_DECL_LIST_MEMBER(struct pj_ioqueue_key_t);
 
+    pj_pool_t*          pool;
```
sauwming, line 140: Minor: pointer (*) placement. The convention is `*pool`.

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -378,11 +412,24 @@ PJ_DEF(pj_status_t) pj_ioqueue_create2(pj_pool_t *pool,
     rc = sizeof(union operation_key);
 
     /* Check that sizeof(pj_ioqueue_op_key_t) makes sense. */
-    PJ_ASSERT_RETURN(sizeof(pj_ioqueue_op_key_t)-sizeof(void*) >= 
+    PJ_ASSERT_RETURN(sizeof(pj_ioqueue_op_key_t)-3*sizeof(void*) >=
```
sauwming, line 415: Why 3 times? Perhaps some doc would help, so we can easily recall in the future.

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -461,7 +485,11 @@ PJ_DEF(pj_status_t) pj_ioqueue_destroy( pj_ioqueue_t *ioqueue )
 #if PJ_HAS_TCP
     unsigned i;
 #endif
-    pj_ioqueue_key_t *key;
+    pj_ioqueue_key_t *key, *next;
+    pj_time_val stop;
+
+    /* Timeout for keys cancelled ops */
+    enum { WAIT_KEY_MS = 5000 };
```
sauwming, line 492: Since the value is quite large (5 sec), better define a macro in the beginning of the file, easier for user to find out and adjust (if necessary).

pjlib/src/pjlib-test/ioq_iocp_unreg_test.c
```diff
@@ -0,0 +1,188 @@
+/*
```
sauwming, line 1: Test code still requires copyright info.

pjlib/src/pjlib-test/ioq_iocp_unreg_test.c
```diff
@@ -0,0 +1,188 @@
...
+
+    /* create send socket */
+    status = pj_activesock_create_udp(pool, NULL, &cfg, test->ioq, &cb, NULL,
+                                      &test->asock_send, NULL);
+    if (status != PJ_SUCCESS) {
+        status = -902;
+        goto on_error;
+    }
+
+    /* create sockets to recevie */
```
sauwming, line 115: receive

pjlib/src/pjlib-test/ioq_stress_test.c
```diff
@@ -333,7 +333,7 @@ static void on_accept_complete(pj_ioqueue_key_t *key,
     status = pj_ioqueue_register_sock2(test->state.pool,
                                        test->state.ioq,
                                        test->state.socks[SERVER],
-                                       test->state.grp_lock,
+                                       NULL, //test->state.grp_lock,
```
sauwming, line 336: Why the change?

nanangizz:
The last commit should cover all review comments above.

@jimying, re: copyright text, feel free to change the name :)
Btw, as mentioned in the top of the file, the test can repro the issue on Win 10 & MSVC2005, but cannot on Windows 11 & MSVC 2022. Also the test will be run for all ioqueue types (originally for iocp only), just in case.

pjlib/src/pjlib-test/ioq_iocp_unreg_test.c
```diff
@@ -0,0 +1,208 @@
+/*
+ * Copyright (C) 2024 jimying at github dot com.
+ * Copyright (C) 2024 Teluu Inc. (http://www.teluu.com)
```
sauwming, line 3: If I'm not mistaken, Teluu is usually put above, for two reasons: 1. to signify that the original author has agreed to contribute it to Teluu (as per CLA), 2. to make it easier to update copyright year (i.e. only the latest/first copyright info will get updated, the rest will remain the same).

sauwming (approved):

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -48,11 +49,27 @@
...
  */
 #define ACCEPT_ADDR_LEN     (sizeof(pj_sockaddr_in)+16)
 
+
+/* Timeout for cancelling pending operations in ioqueue destroy.
+ * Upon ioqueue destroy, all keys must be unregistered and all pending
+ * operations must be cancelled. As cancelling ops is asynchronous,
+ * IOCP destroy may need to wait for the maximum time specified here.
+ */
+#define TIMEOUT_CANCEL_OP   5000
```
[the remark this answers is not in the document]
nanangizz: Actually WAIT_KEY_MS should have been replaced by TIMEOUT_CANCEL_OP, so WAIT_KEY_MS is unused (and undefined).

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -402,54 +464,31 @@ PJ_DEF(pj_status_t) pj_ioqueue_create2(pj_pool_t *pool,
...
-            return rc;
-        }
+        /* Initialize pending op lists */
+        pj_list_init(&key->pending_list);
+        pj_list_init(&key->free_pending_list);
 
         pj_list_push_back(&ioqueue->free_list, key);
     }
-#endif
+    ioqueue->max_fd = pj_list_size(&ioqueue->free_list); // max_fd;
```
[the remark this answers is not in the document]
nanangizz: oops right, forgot to revert.

pjlib/src/pj/ioqueue_winnt.c
```diff
@@ -476,34 +516,43 @@ PJ_DEF(pj_status_t) pj_ioqueue_destroy( pj_ioqueue_t *ioqueue )
...
+        pending_key_cnt = ioqueue->max_fd - pj_list_size(&ioqueue->free_list);
+        if (!pending_key_cnt)
+            break;
+
+        pj_ioqueue_poll(ioqueue, &timeout);
+
+        pj_gettickcount(&timeout);
+        if (PJ_TIME_VAL_GTE(timeout, stop)) {
+            PJ_LOG(3, (THIS_FILE, "Warning, IOCP destroy timeout in waiting "
+                       "for cancelling ops, after %dms, pending keys=%d",
```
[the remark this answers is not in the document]
nanangizz: fixed

PR #4140, opened 2024-11-07
pcaputil: Fix infinite write on skewed timestamp

### LINKED ISSUE #4139: pcaputil writes huge output when timestamp is skewed
### Describe the bug

I have a pcap file in which the timestamp in one packet is *less than*  the previous one. On this case, `ts_gap` is huge, and the util can write gigabytes of output.

image

### Steps to reproduce

With [this](https://github.com/user-attachments/files/17667960/bad.zip) pcap file, run:
```
pcaputil --codec=PCMA/8000/1 bad.pcap bad.wav
```

pcaputil now takes 100% CPU and writes an infinite output file until aborted.

### PJSIP version

master

### Context

N/A

### Log, call stack, etc

```shell
N/A
```

orgads:
If a packet timestamp is *less than* the previous packet, pcaputil considers this a huge gap, and tries to fill it by writing endless blank samples.

Add protection to avoid filling a gap if it is more than 30 sec.

Fixes #4139

### DIFF
--- a/pjsip-apps/src/samples/pcaputil.c
+++ b/pjsip-apps/src/samples/pcaputil.c
@@ -21,6 +21,9 @@
 #include <pjmedia.h>
 #include <pjmedia-codec.h>
 
+/* Ignore gap >30s */
+#define GAP_IGNORE_SECONDS 30
+
 static const char *USAGE =
 "pcaputil [options] INPUT OUTPUT\n"
 "\n"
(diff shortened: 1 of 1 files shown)


nanangizz (approved):

sauwming:
Why not specifically handle the problematic issue only, i.e. `if a packet timestamp is less than the previous packet` -> `(pj_ntohl(pkt1.rtp->ts) < pj_ntohl(pkt0.rtp->ts)`?

What if there's a genuine lack of packet for, say, 11 sec, perhaps during call hold? If we indeed want this feature (i.e. ignoring gaps > certain period), the 10s seems a bit arbitrary, so perhaps make it configurable (e.g. define macro `#define GAP_IGNORE_PERIOD 10000`).

orgads:
I don't mind changing it to 30s, and adding a define (would you like a command-line arg for that?). Checking if it is smaller is not good enough, because it could have been a "much larger" value as well. The change between these packets looks random.

sauwming:
Command line is not necessary at the moment.

sauwming (approved):

orgads:
Fixed the calculation. `ts_gap` is not measured in milliseconds. Had to divide it by clock_rate to get the value in seconds (I preferred to multiply the other side of the equation).

PR #4142, opened 2024-11-09
Fix to compile error when excluding AKA algos in sip_auth_client auth_find_cred.

amubiera:
Fixes error for: `pjsip/src/pjsip/sip_auth_client.c:739:21: error: overlapping comparisons always evaluate to true [-Werror,-Wtautological-overlap-compare]`


amubiera:
Seems already fixed by #4141

PR #4149, opened 2024-11-12
Add software clock to sound port

nanangizz:
This is a better approach to handle [sound device timing problem](https://docs.pjsip.org/en/latest/specific-guides/media/audio_flow.html#sound-device-timing-problem), as features such as auto-close and stereo will work.

Application can enable this via PJSUA settings:
- Compile time, set `PJSUA_DEFAULT_SND_USE_SW_CLOCK` to `PJ_TRUE` in `config_site.h`.
   Note: In `config_site_sample.h` it is enabled by default for Android (i.e: when `PJ_CONFIG_ANDROID == 1`).
- Run time, set `pjsua_media_config.snd_use_sw_clock`  to `PJ_TRUE` in PJSUA or `MediaConfig.sndUseSwClock` to `true` in PJSUA2.

### DIFF
--- a/pjlib/include/pj/config_site_sample.h
+++ b/pjlib/include/pj/config_site_sample.h
@@ -390,8 +390,10 @@
     /* Fine tune Speex's default settings for best performance/quality */
     #define PJMEDIA_CODEC_SPEEX_DEFAULT_QUALITY 5
 
-    /* Software clock for sound device can improve quality */
-    #define PJSUA_DEFAULT_SND_USE_SOFT_CLOCK    PJ_TRUE
+    /* Using software clock for media flow can improve quality,
+     * i.e: more consistent RTP timing and less jitter/burst.
+     */
+    #define PJSUA_DEFAULT_SND_USE_SW_CLOCK      PJ_TRUE
 
     /*
      * PJSIP settings.
--- a/pjlib/include/pj/config_site_sample.h
+++ b/pjlib/include/pj/config_site_sample.h
@@ -389,7 +389,10 @@
 
     /* Fine tune Speex's default settings for best performance/quality */
     #define PJMEDIA_CODEC_SPEEX_DEFAULT_QUALITY 5
-    
+
+    /* Software clock for sound device can improve quality */
+    #define PJSUA_DEFAULT_SND_USE_SOFT_CLOCK    PJ_TRUE
+
     /*
      * PJSIP settings.
      */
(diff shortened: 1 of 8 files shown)


pjmedia/src/pjmedia/sound_port.c
```diff
@@ -322,24 +383,79 @@ static pj_status_t start_sound_device( pj_pool_t *pool,
...
+        if (!snd_port->frame_buf) {
+            status = PJ_ENOMEM;
+            goto on_error;
         }
+
+        status = pjmedia_clock_start(snd_port->clock);
+        if (status != PJ_SUCCESS)
+            goto on_error;
+
+        PJ_LOG(4,(THIS_FILE, "Sound port does not use native clock"));
```
sauwming, line 441: Explicit would be better here, as to not leave the users confused (the users know it doesn't use native clock but then they don't know what's being used). So "Sound port uses internal (or software) clock" should be more clear.

pjlib/include/pj/config_site_sample.h
```diff
@@ -389,7 +389,10 @@
 
     /* Fine tune Speex's default settings for best performance/quality */
     #define PJMEDIA_CODEC_SPEEX_DEFAULT_QUALITY 5
-    
+
+    /* Software clock for sound device can improve quality */
```
sauwming, line 393: Add more explanation about what quality improvement that's expected, such as: Using software clock for (audio) media flow can improve quality (more consistent RTP timing and less jitter/burst).

pjmedia/include/pjmedia/sound_port.h
```diff
@@ -71,7 +71,15 @@ enum pjmedia_snd_port_option
...
-    PJMEDIA_SND_PORT_NO_AUTO_START = 1
+    PJMEDIA_SND_PORT_NO_AUTO_START = 1,
+
+    /**
+     * Sound device uses \ref PJMEDIA_CLOCK instead of native sound device
+     * clock, generally this will be able to reduce jitter and clock skew.
+     *
+     * This option is not applicable for encoded/non-PCM format.
+     */
+    PJMEDIA_SND_PORT_USE_SOFT_CLOCK = 2
```
sauwming, line 82: Perhaps using SW_CLOCK would be more intuitive? It's inline with what we've been using, such as USE_SW_ECHO, SW_VID_ENC/DEC.

sauwming (approved):

trengginas (approved):

Issue #4150, opened 2024-11-12
Correct CPU (compiler) features detection during cross-compile


aol-nnov:
### Describe the bug

While cross-compiling cpu features (like neon, sse2, etc.) are not properly detected, which leads to compilation failures.

### Steps to reproduce

Compile pjsip on linux amd64 for armhf

### PJSIP version

2.14.1

### Context

Debian 12 amd64 with cross-compiler for armhf

### Log, call stack, etc

```shell
../../webrtc/src/webrtc//common_audio/signal_processing/vector_scaling_operations.c 
arm-linux-gnueabihf-gcc -Wall -DPJ_AUTOCONF=1 -Wdate-time -D_FORTIFY_SOURCE=2 -g -O2 -ffile-prefix-map=/builds/foreign/telephony/pjsip=. -fstack-protector-strong -Wformat -Werror=format-security -DPJ_IS_BIG_ENDIAN=0 -DPJ_IS_LITTLE_ENDIAN=1 -fPIC -Iyes/include -Iyes/include -Iyes/include -Iyes/include   -Wdate-time -D_FORTIFY_SOURCE=2 -g -O2 -ffile-prefix-map=/builds/foreign/telephony/pjsip=. -fstack-protector-strong -Wformat -Werror=format-security -I. -I../../webrtc/src -I../../../pjlib/include -fexceptions -DWEBRTC_POSIX=1  \
	-c -o output/libwebrtc-arm-unknown-linux-gnueabihf/modules/audio_processing/aec/aec_core_sse2.o \
	../../webrtc/src/webrtc//modules/audio_processing/aec/aec_core_sse2.c 
arm-linux-gnueabihf-gcc -Wall -DPJ_AUTOCONF=1 -Wdate-time -D_FORTIFY_SOURCE=2 -g -O2 -ffile-prefix-map=/builds/foreign/telephony/pjsip=. -fstack-protector-strong -Wformat -Werror=format-security -DPJ_IS_BIG_ENDIAN=0 -DPJ_IS_LITTLE_ENDIAN=1 -fPIC -Iyes/include -Iyes/include -Iyes/include -Iyes/include   -Wdate-time -D_FORTIFY_SOURCE=2 -g -O2 -ffile-prefix-map=/builds/foreign/telephony/pjsip=. -fstack-protector-strong -Wformat -Werror=format-security -I. -I../../webrtc/src -I../../../pjlib/include -fexceptions -DWEBRTC_POSIX=1  \
[...]

PR #4151, opened 2024-11-12
Correct cpu features detection during cross-compiation

aol-nnov:
Closes https://github.com/pjsip/pjproject/issues/4150

### DIFF
--- a/aconfigure.ac
+++ b/aconfigure.ac
@@ -2646,6 +2646,23 @@ AC_ARG_ENABLE(libyuv,
     AC_MSG_RESULT([Checking if libyuv is disabled...no])
 )
 
+dnl proper neon detector
+SAVED_CFLAGS="$CFLAGS"
+case $target_cpu in
+	arm*)
+        CFLAGS="-mfpu=neon $CFLAGS"
+	;;
+	aarch64*)
+        CFLAGS="-march=armv8-a+simd $CFLAGS"
+	;;
+esac
+AC_COMPILE_IFELSE(
+    [AC_LANG_PROGRAM()],
+    [ax_cv_support_neon_ext=yes],
+    [ax_cv_support_neon_ext=no]
+)
+CFLAGS="$SAVED_CFLAGS"
+
 dnl # Include webrtc
 AC_SUBST(ac_no_webrtc)
 AC_SUBST(ac_webrtc_instset)
@@ -2711,11 +2728,23 @@ AC_ARG_ENABLE(libwebrtc,
             ;;
             *win32* | *w32* | *darwin* | *linux*)
                 case $target in
-                    armv7l*gnueabihf)
-                        ac_webrtc_instset=neon
-                        ac_webrtc_cflags="-DWEBRTC_ARCH_ARM_V7 -mfloat-abi=hard -mfpu=neon"
+                    arm*gnueabihf)
+                        if test "x$ax_cv_support_neon_ext" = "xyes"; then
+                            ac_webrtc_instset=neon
+                            ac_webrtc_cflags="-DWEBRTC_ARCH_ARM_V7 -mfloat-abi=hard -mfpu=neon"
+                        else
+                            ac_webrtc_instset=generic
+                        fi
                     ;;
-                    arm-apple-darwin* | aarch64*)
+                    aarch64*)
+                        if test "x$ax_cv_support_neon_ext" = "xyes"; then
+                            ac_webrtc_instset=neon
+                            ac_webrtc_cflags="-DWEBRTC_ARCH_ARM64"
+                        else
+                            ac_webrtc_instset=generic
+                        fi
+                    ;;
+                    arm-apple-darwin*)
                         ac_webrtc_instset=neon
                         ac_webrtc_cflags="-DWEBRTC_ARCH_ARM64"
                     ;;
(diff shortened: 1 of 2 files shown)


aconfigure.ac
```diff
@@ -2480,10 +2489,12 @@ AC_ARG_ENABLE(libwebrtc,
...
                          case $target in
-                             armv7l*gnueabihf)
-                                 ac_webrtc_instset=neon
-                                 ac_webrtc_cflags="-DWEBRTC_ARCH_ARM_V7 -mfloat-abi=hard -mfpu=neon"
-                                 ;;
+                             armv*gnueabihf)
+							 	if test "x$ax_cv_support_neon_ext" = "xyes"; then
+									ac_webrtc_instset=neon
+									ac_webrtc_cflags="-DWEBRTC_ARCH_ARM_V7 -mfloat-abi=hard -mfpu=neon"
+								fi
```
sauwming, line 2496: Do we need an "else" here?
Otherwise `ac_webrtc_instset` will be unset.

aconfigure.ac
```diff
@@ -2480,10 +2489,12 @@ AC_ARG_ENABLE(libwebrtc,
 			;;
 		     *win32* | *w32* | *darwin* | *linux*)
                          case $target in
-                             armv7l*gnueabihf)
-                                 ac_webrtc_instset=neon
-                                 ac_webrtc_cflags="-DWEBRTC_ARCH_ARM_V7 -mfloat-abi=hard -mfpu=neon"
-                                 ;;
+                             armv*gnueabihf)
+							 	if test "x$ax_cv_support_neon_ext" = "xyes"; then
```
sauwming, line 2493: Also need to change the detection for webrtc_aec3 below?

sauwming (commented):
You also need to commit the modified `aconfigure` as well (i.e. `autoconf aconfigure.ac > aconfigure`).

aconfigure.ac
aol-nnov: Not sure, though. With those minimal changes it compiles quite okay.

aconfigure.ac
aol-nnov: What do you propose for "else"? Empty variable?

aol-nnov:
> You also need to commit the modified aconfigure

Should I include `aclocal.m4` as well? `AX_CHECK_COMPILE_FLAG` is an external macro, provided by autoconf-archive package in Debian.

aconfigure.ac
sauwming: webrtc_aec3 is not included to be built by default.
So test it with `--enable-libwebrtc-aec3` option.

aconfigure.ac
sauwming: Either ac_webrtc_instset=sse2 or generic.
I'm not sure what it's supposed to be if neon is not supported in ARM machine.

sauwming:
> > You also need to commit the modified aconfigure
> Should I include `aclocal.m4` as well? `AX_CHECK_COMPILE_FLAG` is an external macro, provided by autoconf-archive package in Debian.

Then this could be a problem. I check my Mac machine and it doesn't have `autoconf-archive` installed by default. Try to replace it with readily available macros instead, such as `AC_COMPILE...` or something else.

aol-nnov:
> Then this could be a problem

But why? If I run aclocal, all external macro are copied over to self-sufficient aclocal.m4 and I can include it into the commit as well.

aol-nnov:
Also, autoconf-archive is available on Mac via HomeBrew: https://formulae.brew.sh/formula/autoconf-archive#default

aconfigure.ac
aol-nnov: > Either ac_webrtc_instset=sse2


it was sse2 initially (before this patch), but arm devices do not support sse2 and compilation was failing due to that.

sauwming: If this is the case, then perhaps we don't need the neon detection above? (since there is no failover if it doesn't support neon).

Just modify `armv7l*gnueabihf` to `armv*gnueabihf`, wouldn't this suffice?

aol-nnov: > Just modify armv7l*gnueabihf to armv*gnueabihf, wouldn't this suffice?

That was my initial approach, but I've talked to colleagues and they proposed checking compiler flags, too.

Also, there's a typo in `armv*gnueabihf` - `v` is also not needed, I'm on my way with updates.

aol-nnov: Also, armv6 does not have neon support, AFAIK. So, let's stick to that check.

aconfigure.ac
aol-nnov: Implemented the same way as above. The change is really needed too, as long as --enable-libwebrtc-aec3 is present.

aconfigure.ac
aol-nnov: added fallback to `generic`

aol-nnov:
@sauwming Let's do another round. I've implemented your comments.

Also, please check aclocal.m4 if it works for you that way.

sauwming:
The `aclocal.m4` seems to work okay for me, as well as on the CI machines.
Although quite frankly, it still leaves me worried.

####
aclocal.m4 itself is not inherently portable across different environments without considering some nuances. 

Portability Considerations:
- Generated by aclocal:
aclocal.m4 is typically generated on the build machine where aclocal runs. The macros it includes are influenced by the m4 macro files available on that machine.
- Environment-Specific Dependencies:
If aclocal.m4 depends on custom or third-party macros (e.g., from the Autoconf Archive), it may reference macros that are not present on another system. This can lead to portability issues when transferring the aclocal.m4 file to a different environment without the necessary macro definitions.
- Not a Replacement for Autoconf and Automake:
While you can transfer aclocal.m4 to another system, it’s not a full replacement for running aclocal, automake, or autoconf. The environment where the configure script is regenerated must have the necessary tools and macro files.

Conclusion:
aclocal.m4 is not highly portable by itself because it relies on macros defined by the environment in which it was generated. For distributing software, ensure that you distribute the configure script and any other build system files needed for an end-user to compile the code without requiring Autotools.
###

I know I can install autoconf-archive, but I don't want to do this on every machine and I don't wish to make this a requirement, especially if it's not really necessary.

So, I would prefer if we check the availability of compile flags using the standard macros instead, like everywhere else in the currently existing configure script.

aol-nnov:
So, you want me to copy and paste a local.mc contents inside the aconfigure.ac? Okay, will do that tomorrow.

I'm out for today and will be back online in next 12 hours or so.

Thank you for the cooperation!

sauwming:
I believe you can replace it with `AC_COMPILE_IFELSE`.

aol-nnov:
@sauwming I've rewrote the patch with AC_COMPILE_IFELSE as you proposed earlier. Please review.

As for factoring out `aarch64*)` into separate branch, I had a small chat with our HW team and they had some uncertanities about the fact, that all aarch64 CPUS have neon support. On the other hand, I have no access to `arm-apple-darwin*` hw to check.

So, it's the least intrusive patch for now.

sauwming (approved):

nanangizz (approved):

Issue #4152, opened 2024-11-12
Incorrect pkg-config executable search during cross-compilation


aol-nnov:
### Describe the bug

Missing pkg-config during cross-compilation as it is named `$host_cpu-$host_os-pkg-config`. Failing to find it leads to compilation failure.

### Steps to reproduce

run cross-compilation, obsserve failure

### PJSIP version

2.14.1

### Context

cross-compilation follow-up for https://github.com/pjsip/pjproject/issues/4150

### Log, call stack, etc

```shell
none
```

PR #4153, opened 2024-11-12
Fix pkg-config invocation during cross-compilation

aol-nnov:
Also, honour system locations for `--with-sdl` and `--with-ffmpeg`

Closes https://github.com/pjsip/pjproject/issues/4152

### DIFF
--- a/aconfigure.ac
+++ b/aconfigure.ac
@@ -115,6 +115,7 @@ AC_CHECK_LIB(winmm,puts)
 AC_CHECK_LIB(socket,puts)
 AC_CHECK_LIB(rt,puts)
 AC_CHECK_LIB(m,sin)
+AC_CHECK_PROGS(PKG_CONFIG,$host_cpu-$host_os-pkg-config $host-pkg-config pkg-config "python pkgconfig.py",none)
 
 dnl
 dnl libuuid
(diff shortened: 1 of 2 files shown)


aconfigure.ac
```diff
@@ -2021,7 +2023,7 @@ AC_ARG_ENABLE(ssl,
     		    fi
 
         	    AC_CHECK_PROGS(PKG_CONFIG,
-                       $host-pkg-config pkg-config "python pkgconfig.py",
+                       $host_cpu-$host_os-pkg-config pkg-config "python pkgconfig.py",
```
sauwming, line 2026: Googling around (such as:
https://autotools.info/pkgconfig/cross-compiling.html)
, `$host-pkg-config` seems to be commonly used as well, so better not remove it.

sauwming (commented):
Like the other PR, also need to include the `aconfigure`.

aconfigure.ac
aol-nnov: Good point, will keep that, too.

aol-nnov:
> Like the other PR, also need to include the aconfigure.

Agreed, let's finish that one and then I will rebase this on top and regenerate `aconfigure` as well.

aconfigure.ac
aol-nnov: done

aol-nnov:
@sauwming I've discovered two `AC_CHECK_PROGS(PKG_CONFIG,...)` calls along the `aconfigure.ac`, so I decided to combine them and move closer to the top of the file.

Also, as I mentioned earlier, system locations for `--with-sdl` and `--with-ffmpeg` are now honoured if those options are provided without arguments.

`aconfigure` is also re-generated according to the introduced changes. Please review.

sauwming (approved):

nanangizz (approved):

aol-nnov:
Thank you for the prompt responses @sauwming @nanangizz ! It's a pleasure to work with you!

PR #4155, opened 2024-11-13
Add support of out-of-dialog SIP transactions

wosrediinanatour:
This commit adds support of out-of-dialog transactions to PJSUA and PJSUA2. The requests are sent in the context of an account. The API is similar to that of an in-dialog transaction.

Example: The new interface can be used to send SIP OPTIONS.

Issue: https://github.com/pjsip/pjproject/issues/4156

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -1197,6 +1197,19 @@ typedef struct pjsua_callback
                               pjsip_event *e);
 
     /**
+     * Notify application when a transaction started by pjsua_acc_send_request()
+     * has been completed,i.e. when a response has been received.
+     *
+     * @param acc_id   Account identification.
+     * @param token    Arbitrary data owned by the application
+     *                 that was specified when sending the request.
+     * @param event    Transaction event that caused the state change.
+     */
+    void (*on_acc_send_request)(pjsua_acc_id acc_id,
+                                void *token,
+                                pjsip_event *event);
+
+    /**
      * Notify application when media state in the call has changed.
      * Normal application would need to implement this callback, e.g.
      * to connect the call's media to sound device. When ICE is used,
@@ -4921,6 +4934,30 @@ PJ_DECL(pj_status_t) pjsua_acc_get_config(pjsua_acc_id acc_id,
 PJ_DECL(pj_status_t) pjsua_acc_modify(pjsua_acc_id acc_id,
                                       const pjsua_acc_config *acc_cfg);
 
+/**
+ * Send arbitrary out-of-dialog requests from an account, e.g. OPTIONS.
+ * The application should use the call or presence API to create
+ * dialog-related requests.
+ *
+ * @param acc_id        The account ID.
+ * @param dest_uri      URI to be put into the To header (normally is the same
+ *                      as the target URI).
+ * @param method        The SIP method of the request.
+ * @param options       This is for future use (currently only NULL is supported).
+ * @param token         Arbitrary token (user data owned by the application)
+ *                      to be passed back to the application in callback
+ *                      on_acc_send_request().
+ * @param msg_data      Optional headers etc. to be added to the outgoing
+ *                      request, or NULL if no custom header is desired.
+ *
+ * @return              PJ_SUCCESS or the error code.
+ */
+PJ_DECL(pj_status_t) pjsua_acc_send_request(pjsua_acc_id acc_id,
+                                            const pj_str_t *dest_uri,
+                                            const pj_str_t *method,
+                                            void *options,
+                                            void *token,
+                                            const pjsua_msg_data *msg_data);
 
 /**
  * Modify account's presence status to be advertised to remote/presence
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -800,6 +800,37 @@ public:
 };
 
 /**
+ * This structure contains parameters for Account::sendRequest()
+ */
+struct SendRequestParam
+{
+    /**
+     * Token or arbitrary user data ownd by the application,
+     * which will be passed back in callback Account::onSendRequest().
+     */
+    Token        userData;
+
+    /**
+     * SIP method of the request.
+     */
+    string       method;
+
+    /**
+     * Message body and/or list of headers etc. to be included in
+     * the outgoing request.
+     */
+    SipTxOption  txOption;
+
+public:
+    /**
+     * Default constructor initializes with zero/empty values.
+     */
+    SendRequestParam();
+};
+
+
+
+/**
  * SRTP crypto.
  */
 struct SrtpCrypto
@@ -1726,6 +1757,24 @@ struct OnMwiInfoParam
 };
 
 /**
+ * This structure contains parameters for Account::onSendRequest() callback.
+ */
+struct OnSendRequestParam
+{
+    /**
+     * Token or arbitrary user data owned by the application,
+     * which was passed to Endpoint::sendRquest() function.
+     */
+    Token               userData;
+
+    /**
+     * Transaction event that caused the state change.
+     */
+    SipEvent    e;
+};
+
+
+/**
  * Parameters for presNotify() account method.
  */
 struct PresNotifyParam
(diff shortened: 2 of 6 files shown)


pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -1196,6 +1196,19 @@ typedef struct pjsua_callback
...
 
+    /**
+     * This is a general notification callback which is called whenever
+     * a transaction within the account has changed state. Application can
+     * implement this callback for example to monitor the state of
+     * outgoing requests, or to answer unhandled incoming requests
+     * with a final response.
+     *
+     * @param acc_id   Account identification.
+     * @param e         Transaction event that caused the state change.
```
sauwming, line 1207: event

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -1196,6 +1196,19 @@ typedef struct pjsua_callback
...
                               pjsip_event *e);
 
+    /**
+     * This is a general notification callback which is called whenever
+     * a transaction within the account has changed state. Application can
+     * implement this callback for example to monitor the state of
+     * outgoing requests, or to answer unhandled incoming requests
+     * with a final response.
+     *
+     * @param acc_id   Account identification.
```
sauwming, line 1206: Minor: indentation ('A' should align with 'T')

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -4921,6 +4934,25 @@ PJ_DECL(pj_status_t) pjsua_acc_get_config(pjsua_acc_id acc_id,
...
                                       const pjsua_acc_config *acc_cfg);
 
+/**
+ * Send arbitrary requests using the account. Application should only use
+ * this function to create auxiliary requests outside dialog, such as
+ * OPTIONS, and use the call or presence API to create dialog related
+ * requests.
+ *
+ * @param acc_id        The account ID.
+ * @param dst_uri       URI to be put in the To header (normally is the same
```
sauwming, line 4944: dest_uri, method_str

pjsip/include/pjsua2/account.hpp
```diff
@@ -2082,6 +2131,16 @@ class Account
     virtual void onInstantMessageStatus(OnInstantMessageStatusParam &prm)
     { PJ_UNUSED_ARG(prm); }
 
+    /**
+     * Notify application when call state has changed.
```
sauwming, line 2135: Incorrect documentation.

pjsip/src/pjsip/sip_transaction.c
```diff
@@ -34,7 +34,7 @@
 
 #define THIS_FILE   "sip_transaction.c"
 
-#if 0
+#if 1
```
sauwming, line 37: need to revert back to 0.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -4921,6 +4934,25 @@ PJ_DECL(pj_status_t) pjsua_acc_get_config(pjsua_acc_id acc_id,
...
+ * @param method        The SIP method of the request.
+ * @param msg_data      Optional headers etc to be added to outgoing
+ *                      request, or NULL if no custom header is desired.
+ *
+ * @return              PJ_SUCCESS or the error code.
+ */
+PJ_DECL(pj_status_t) pjsua_acc_send_request(pjsua_acc_id acc_id,
+                                            const pj_str_t *dest_uri,
+                                            const pj_str_t *method_str,
+                                            const pjsua_msg_data *msg_data);
```
nanangizz, line 4955: Add few params which perhaps can be useful in the future:
- app user-data/token (void *),
- generic options/settings (void *): for now perhaps must be NULL, for extensibility in the future, so we don't need to create `pjsua_acc_send_request2()` :)

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -1196,6 +1196,19 @@ typedef struct pjsua_callback
...
+     * implement this callback for example to monitor the state of
+     * outgoing requests, or to answer unhandled incoming requests
+     * with a final response.
+     *
+     * @param acc_id   Account identification.
+     * @param e         Transaction event that caused the state change.
+     */
+    void (*on_acc_tsx_state)(pjsua_acc_id acc_id,
+                             pjsip_event *event);
+
```
nanangizz, line 1211: Additional param:
- user data/token.

Also for consistency, perhaps the name should be something like `on_acc_send_request` instead of `on_tsx_state`, as the callback is actually called once, i.e: on transaction termination, instead of every tsx state shift.

wosrediinanatour:
Incorporated feedback in a fixup commit. PTAL.

pjsip/include/pjsua-lib/pjsua.h
nanangizz: Looks like only token/user-data was added and it is specified as "unused" while it should be usable (as opposed to must-be-NULL :)

Actually user-data/token and settings are actually different:
- user-data/token, basically it needs to be passed back in the callback, so app can correspond the tsx to something, just check `pjsip_endpt_send_request()` or `pjsua_call_make_call()` for reference.
- settings, it is for future use, must be NULL for now.

wosrediinanatour: Sorry..I misunderstood.

So you are saying that the token should already work and shouldn't be for future use cases, i e. that an object is allocated in the pool, which stores the acc_id and the token (a pointer). (In current PR only acc_id is stored).

wosrediinanatour:
PTAL (https://github.com/pjsip/pjproject/pull/4155/commits/9114869eb8bef7ce356829bc79decbed20236f5e)

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -1510,6 +1510,70 @@ PJ_DEF(pj_status_t) pjsua_acc_modify( pjsua_acc_id acc_id,
     return status;
 }
 
+typedef struct pjsua_send_request_data
+{
+    pjsua_acc_id acc_id;
+    void *token;
+} pjsua_send_request_data;
```
nanangizz, line 1517: A minor: by convention, `pj*_` prefixes should only be used for API. As this is internal type, it should not use `pjsua_` prefix (e.g: just `send_request_data`).

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -1510,6 +1510,70 @@ PJ_DEF(pj_status_t) pjsua_acc_modify( pjsua_acc_id acc_id,
...
+
+static void on_acc_tsx_state(void *request_data, pjsip_event *event)
```
nanangizz, line 1519: Minor: to avoid possible confusion in the future, perhaps don't use `tsx_state` term too here, `on_send_request` for example?

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -1510,6 +1510,70 @@ PJ_DEF(pj_status_t) pjsua_acc_modify( pjsua_acc_id acc_id,
...
+{
+    pjsua_send_request_data *request_data_ = (pjsua_send_request_data*) request_data;
+    if (request_data_)
+        (pjsua_var.ua_cfg.cb.on_acc_send_request)(request_data_->acc_id, request_data_->token, event);
```
nanangizz, line 1523: Add check if `on_acc_send_request` is set?

pjsip/src/pjsua2/account.cpp
```diff
@@ -117,6 +119,13 @@ void RtcpFbConfig::writeObject(ContainerNode &node) const PJSUA2_THROW(Error)
 
 ///////////////////////////////////////////////////////////////////////////////
 
+SendRequestParam::SendRequestParam()
+: method("")
```
nanangizz, line 123: Minor: better initialize `token` (to NULL) too, some code analyzer usually complains :)

pjsip/src/pjsua2/account.cpp
```diff
@@ -1078,7 +1078,7 @@ void Account::sendRequest(const pj::SendRequestParam& prm) PJSUA2_THROW(Error)
     pjsua_msg_data msg_data;
     prm.txOption.toPj(msg_data);
 
-    PJSUA2_CHECK_EXPR(pjsua_acc_send_request(id, &dest_uri, &method, prm.userData, NULL, &msg_data));
+    PJSUA2_CHECK_EXPR(pjsua_acc_send_request(id, &dest_uri, &method, prm.userData, prm.userData, &msg_data));
```
nanangizz, line 1081: Minor: the fourth param (i.e: `options` param) should be NULL?

nanangizz (approved):
Looks good to me, just a minor above.
Thank you @wosrediinanatour!

pjsip/src/pjsua2/account.cpp
```diff
@@ -24,6 +24,8 @@
 using namespace pj;
 using namespace std;
 
+#include <pjsua-lib/pjsua_internal.h>   /* For retrieving pjsua threads */
```
sauwming, line 27: is this still needed?

wosrediinanatour:
Incorporated feedback in one fixup commit and then autosquashed the fixups.

Just a question: Is support of C++98 still a thing? E.g. nowadays for me it feels strange to still use NULL instead of nullptr (and similar things).

sauwming (approved):

sauwming:
Quite frankly, we use NULL out of habit (since we start from C) rather than because we want to compatible with older C++.
You're right though, `nullptr` is generally preferable in C++.

Issue #4117, opened 2024-10-28
DTMF duration is wrong for G722 on receiver with RFC2833


R-Jeske:
### Describe the bug

When a DTMF Signal is transmitted with RFC2833 and DTMF events are used on the receiver, the duration from the callback `cb_on_dtmf_event` differs between G722 and PCMA codecs. When using the PCMA codec, the event is 200ms long and the duration from callback says 200ms (640 inside the packet), but when using the G722 Codec, the event is 200ms long and the duration from the callback says 100ms (640 inside the packet). I tested it with other lengths, and it is always the half in case of G722. Since the `clock_rate` of both PCMA and G722 is 8000, I would assume, that a value of 640 inside the packet is correct. Thus there might be a correction missing (like the `pjmedia_stream_create` `PJMEDIA_HANDLE_G722_MPEG_BUG`)

### Steps to reproduce

1. Call callee directly using G722 Codec
2. send DTMF with RFC2833 

```
pjsua_call_send_dtmf_param dtmf_param;
pjsua_call_send_dtmf_param_default(&dtmf_param);
dtmf_param.method = PJSUA_DTMF_METHOD_RFC2833;
char buf[2];
buf[0]='1' ;
buf[1]='\0';
const pj_str_t str=pj_str(buf);
dtmf_param.digits = str;
pjsua_call_send_dtmf(id, &dtmf_param);
```
3. use callback to read duration:
```
void cb_on_dtmf_event(pjsua_call_id call_id,
                                        const pjsua_dtmf_event *event)
{
  bool dtmfEnd = (event->flags & PJMEDIA_STREAM_DTMF_IS_END) != 0;
  if (p->pCb_!=NULL && dtmfEnd)
  {
    a = event->duration;
  }
}
```

### PJSIP version

2.14.1

### Context

Should occur on all platforms, that support G722

### Log, call stack, etc

-

trengginas:
The event duration calculation is done [here](https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream.c#L1801).
G722 is using 16000 clockrate, and PCMA is using 8000, it is expected.
ref:
- https://datatracker.ietf.org/doc/html/rfc3551#section-4.5.2

R-Jeske:
Clock Rate is 8000 
> RTP clock rate for the G722 payload format is 8,000 Hz

Sample Rate is 16000
> sampling rate for G.722 audio is 16,000 Hz

[...]

PR #4159, opened 2024-11-14
fix timestamp and duration for received dtmf events for G722 and Opus (#4117)

R-Jeske:
This fixes: #4117: DTMF duration is wrong for G722 on receiver with RFC2833

For G722 and Opus, the sample rate may differ from the clock rate. Thus, the calculation of duration and timestamp of a received RFC2833 DTMF event must be recalculated based on the real clock rate specified as specified in the standards.

### DIFF
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -1771,6 +1771,7 @@ static void handle_incoming_dtmf( pjmedia_stream *stream,
     pjmedia_stream_dtmf_event dtmf_event;
     pj_bool_t is_event_end;
     pj_bool_t emit_event;
+    float ts_modifier = 1.0;
 
     /* Check compiler packing. */
     pj_assert(sizeof(pjmedia_rtp_dtmf_event)==4);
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia/stream.c
```diff
@@ -1841,10 +1855,23 @@ static void handle_incoming_dtmf( pjmedia_stream *stream,
      * the DTMF digits in the buffer.
      */
     if (stream->dtmf_event_cb) {
+        /* Correct duration and timestamp for codecs where clock_rate
```
sauwming, line 1858: since this is essentially the same code as above, should we put it in the outer block instead, to avoid duplication?

R-Jeske: yes, that sounds useful. I added a commit to reduce redundancy: 65f91a8b

pjmedia/src/pjmedia/stream.c
```diff
@@ -1797,6 +1797,19 @@ static void handle_incoming_dtmf( pjmedia_stream *stream,
     event_duration = pj_ntohs(event->duration);
     is_event_end = (event->e_vol & PJMEDIA_RTP_DTMF_EVENT_END_MASK) != 0;
 
+    /* Correct duration and timestamp for codecs where clock_rate
+     * differs from sample rate
+     */
+    float ts_modifier = 1.0;
```
sauwming, line 1803: Thank you. Only one small modification required. The variable declaration needs to be at the beginning of the block.

sauwming (approved):

trengginas (approved):

PR #4164, opened 2024-11-20
Fix audio & video conference bridge: adding port may not update port counter

nanangizz:
This is to fix https://github.com/pjsip/pjproject/pull/3928#issuecomment-2484827371.

After investigation, it is caused by invalid port counter (e.g: zero or `-1` while there are actually some ports added). Conference bridge uses the port counter to iterates the ports, so when it is invalid, no media flow will happen. Furthermore, such invalid port counter case may happen when a port is removed before the add port operation, including increasing the port counter, is completed. The add port operation can never be completed when no async operation is queued.

This PR fix it by:
- introducing a new async operation for add port operation,
- only decrement port counter when port add is completed (i.e: internal port state `is_new` is false).

### DIFF
--- a/pjmedia/src/pjmedia/conference.c
+++ b/pjmedia/src/pjmedia/conference.c
@@ -1046,7 +1046,7 @@ static void op_add_port(pjmedia_conf *conf, const op_param *prm)
     cport->is_new = PJ_FALSE;
     ++conf->port_cnt;
 
-    PJ_LOG(5,(THIS_FILE, "Added port %d (%.*s), port count=%d",
+    PJ_LOG(4,(THIS_FILE, "Added port %d (%.*s), port count=%d",
               port, (int)cport->name.slen, cport->name.ptr, conf->port_cnt));
 }
 
--- a/pjmedia/src/pjmedia/conference.c
+++ b/pjmedia/src/pjmedia/conference.c
@@ -392,10 +392,15 @@ static pj_status_t create_conf_port( pj_pool_t *parent_pool,
 {
     struct conf_port *conf_port;
     pj_pool_t *pool = NULL;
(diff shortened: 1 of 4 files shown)


sauwming (dismissed):

sauwming:
pjsua-test failed, and it seems genuine, already rerun it twice.

nanangizz:
Just tried the failing Python test using this branch on MacOS (13.2/Intel), still not reproducible.

sauwming:
Yes, that's strange, I also tried both tests here on my Mac and they completed successfully.

But it's most certainly because of this PR since this is the only one that failed the CI tests in https://github.com/pjsip/pjproject/actions. I have also repeated the CI tests many times over the night and the failed results are consistent.

sauwming:
Your intuition is correct. The '%' causes the issue.
Passing the URI as format specifier doesn't make sense: 
`snprintf(.., ..., "sip:I%20have%20spaces@example.net", ...)`

Although not sure why only in this PR it raises the error.

Perhaps we should change it to `strstr(.., "%p)`?

Another potential issue is that `pjmedia_conf_add_port()` never specifies that `port_name` must be NULL terminated, so we probably shouldn't call `pj_pool_create(..., name->ptr)`

nanangizz:
> Perhaps we should change it to `strstr(.., "%p")`?

Agree. But it may still have a problem, when the name accidentally has `%p`, the pool name will not be as expected. Perhaps we should also escape/replace any unintended `%`, but this needs to be done in the apps.

> Another potential issue is that `pjmedia_conf_add_port()` never specifies that `port_name` must be NULL terminated, so we probably shouldn't call `pj_pool_create(..., name->ptr)`

Agree.

sauwming (approved):

trengginas (approved):

PR #4173, opened 2024-11-24
Fix build failure on RISC-V architecture

imguoguo:
The issue was due to missing architectural definitions in WebRTC’s typedefs.h. After applying the provided patch, the package successfully compiles on Fedora using the [pjproject.spec](https://src.fedoraproject.org/rpms/pjproject/blob/rawhide/f/pjproject.spec) file from the Fedora rawhide repository.

Error messages are as follows without this patch:
```
../../webrtc/src/webrtc/typedefs.h:65:2: error: #error Please add support for your architecture in typedefs.h
../../webrtc/src/webrtc/typedefs.h:69:2: error: #error Define either WEBRTC_ARCH_LITTLE_ENDIAN or WEBRTC_ARCH_BIG_ENDIAN
```

The package was compiled successfully on Fedora Rawhide for the RISC-V architecture using the updated [pjproject.spec](https://src.fedoraproject.org/rpms/pjproject/blob/rawhide/f/pjproject.spec)(with this patch).

WebRTC Patch Source: [Link](https://gitlab.freedesktop.org/pulseaudio/webrtc-audio-processing/-/merge_requests/19.patch)


sauwming:
Thanks, it seems in line with WebRTC AEC3 check here:
https://github.com/pjsip/pjproject/blob/c239fc89e97f50e9ef83021bdaf83cee95130afc/third_party/webrtc_aec3/src/rtc_base/system/arch.h#L70-L75

But just to confirm, can you please check if WebRTC AEC3 is also able to build successfully on RISC-V? (you need to enable it using `./configure --enable-libwebrtc-aec3` and rebuild PJSIP).

sauwming (approved):

trengginas:
Could you please add the changes in (arch.h) https://github.com/imguoguo/pjproject/blob/master/third_party/webrtc_aec3/PJSIP_NOTES

sauwming:
The flag `-mfma` is used when the instruction set is `sse2`:
https://github.com/pjsip/pjproject/blob/master/third_party/build/os-auto.mak.in#L145

So you need to add your target architecture here:
https://github.com/pjsip/pjproject/blob/master/aconfigure.ac#L2832
and probably use either `neon` or `generic` instset.

Also, the PJSIP_NOTES has a conflict with the current master due to our recent change.

imguoguo:
> The flag `-mfma` is used when the instruction set is `sse2`: https://github.com/pjsip/pjproject/blob/master/third_party/build/os-auto.mak.in#L145
> 
> So you need to add your target architecture here: https://github.com/pjsip/pjproject/blob/master/aconfigure.ac#L2832 and probably use either `neon` or `generic` instset.
> 
> Also, the PJSIP_NOTES has a conflict with the current master due to our recent change.

After multiple tests, it was discovered that the same modification was needed in `webrtc`. Interestingly, this issue was not identified during previous tests using spec compilation. Thank you for pointing it out and helping. The update has been committed. After the modification, no errors occurred during testing. Currently, I only have riscv64 boards, so the testing was only conducted and passed on riscv64 platform.

sauwming (approved):

trengginas (approved):

Issue #4180, opened 2024-11-26
heap-use-after-free error detected by address sanitizer on cloned sip headers


ablangy:
### Describe the bug

In pjsip/src/pjsip/sip_msg.c, `pjsip_generic_int_hdr* pjsip_generic_int_hdr_create( pj_pool_t *pool, const pj_str_t *hnames, unsigned value)` allocate space from the provided pj_pool_t to store the header name.  

In the same file `pjsip_generic_int_hdr* pjsip_generic_int_hdr_clone( pj_pool_t *pool, const pjsip_generic_int_hdr *rhs)`  which is the clone function defined in `generic_int_hdr_vptr`, copy the `rhs` header name adress instead of allocating new space in the provided pj_pool_t and duplicate `rhs->name` (same problem with `rhs->sname`).   

In pjsip/src/pjsip/sip_msg.c, `pjsip_generic_array_hdr* pjsip_generic_array_hdr_create( pj_pool_t *pool, const pj_str_t *hnames)` do the same by allocating space from the provided pj_pool_t to store the header name.  
`pjsip_generic_array_hdr* pjsip_generic_array_hdr_clone( pj_pool_t *pool, const pjsip_generic_array_hdr *rhs)` copy the header name address of `rhs` instead of allocating new space in the provided pool.  

### Steps to reproduce

Our application is compiled with asan and ubsan.  
It is composed of two thread:
- one waiting for the reception of sip messages by calling `pj_status_t pjsip_endpt_handle_events(pjsip_endpoint *endpt, const pj_time_val *max_timeout)`
- the other one is consuming those received message

The reception thread is cloning the received message to pass them to the second one, using `pj_status_t pjsip_rx_data_clone(const pjsip_rx_data *src, unsigned flags, pjsip_rx_data **p_rdata)`.  

The call stack provided below is caused by a reception of a SIP INVITE request by thread T3 and handled by thread T6 which try to send an answer.  
The SIP INVITE request has a "Recv-Info" header.  
A custom header parser function is added to pjsip parser by calling  `pj_status_t pjsip_register_hdr_parser( const char *hname, const char *hshortname, pjsip_parse_hdr_func *fptr);`.  
This custom parser create the pjsip header by calling `pjsip_generic_array_hdr* pjsip_generic_array_hdr_create(pj_pool_t *pool,const pj_str_t *hname)`.  

### PJSIP version

2.14

### Context

The issue happens on linux but should not be specific to this platform.  

[...]

PR #4182, opened 2024-11-26
Issue 4180 heap use after free cloned hdr

ablangy:
This pull request should fix the [issue 4180](https://github.com/pjsip/pjproject/issues/4180)

### DIFF
--- a/pjsip/src/pjsip/sip_msg.c
+++ b/pjsip/src/pjsip/sip_msg.c
@@ -898,8 +898,11 @@ static int pjsip_generic_int_hdr_print( pjsip_generic_int_hdr *hdr,
 static pjsip_generic_int_hdr* pjsip_generic_int_hdr_clone( pj_pool_t *pool, 
                                                    const pjsip_generic_int_hdr *rhs)
 {
-    pjsip_generic_int_hdr *hdr = PJ_POOL_ALLOC_T(pool, pjsip_generic_int_hdr);
-    pj_memcpy(hdr, rhs, sizeof(*hdr));
+    pjsip_generic_int_hdr *hdr;
+
+    hdr = pjsip_generic_int_hdr_create(pool, &rhs->name, rhs->ivalue);
+    hdr->type = rhs->type;
+
     return hdr;
 }
 
--- a/pjsip/src/pjsip/sip_msg.c
+++ b/pjsip/src/pjsip/sip_msg.c
@@ -26,6 +26,7 @@
 #include <pj/pool.h>
 #include <pj/assert.h>
 #include <pj/log.h>
+#include <pj/math.h>
 #include <pjlib-util/string.h>
 
 PJ_DEF_DATA(const pjsip_method) pjsip_invite_method =
@@ -982,10 +983,12 @@ static pjsip_generic_array_hdr* pjsip_generic_array_hdr_clone( pj_pool_t *pool,
                                                  const pjsip_generic_array_hdr *rhs)
 {
     unsigned i;
-    pjsip_generic_array_hdr *hdr = PJ_POOL_ALLOC_T(pool, pjsip_generic_array_hdr);
+    pjsip_generic_array_hdr *hdr;
 
-    pj_memcpy(hdr, rhs, sizeof(*hdr));
-    for (i=0; i<rhs->count; ++i) {
+    hdr = pjsip_generic_array_hdr_create(pool, &rhs->name);
+    hdr->type = rhs->type;
+    hdr->count = PJ_MIN(rhs->count, PJSIP_GENERIC_ARRAY_MAX_COUNT);
+    for (i=0; i<hdr->count; ++i) {
         pj_strdup(pool, &hdr->values[i], &rhs->values[i]);
     }
 


nanangizz (dismissed):

sauwming:
This failed the CI tests

pjsip/src/pjsip/sip_msg.c
```diff
@@ -898,8 +898,7 @@ static int pjsip_generic_int_hdr_print( pjsip_generic_int_hdr *hdr,
 static pjsip_generic_int_hdr* pjsip_generic_int_hdr_clone( pj_pool_t *pool, 
                                                    const pjsip_generic_int_hdr *rhs)
 {
-    pjsip_generic_int_hdr *hdr = PJ_POOL_ALLOC_T(pool, pjsip_generic_int_hdr);
-    pj_memcpy(hdr, rhs, sizeof(*hdr));
+    pjsip_generic_int_hdr *hdr = pjsip_generic_int_hdr_create(pool, &rhs->name, rhs->ivalue);
```
nanangizz, line 901: Just realized that it is a generic header, using `pjsip_generic_int_hdr_create()` will always reset the header type to `PJSIP_H_OTHER`. So from the original code, only adding header name duplication (especially if the type is `PJSIP_H_OTHER`) is sufficient?

pjsip/src/pjsip/sip_msg.c
```diff
@@ -979,10 +978,10 @@ static pjsip_generic_array_hdr* pjsip_generic_array_hdr_clone( pj_pool_t *pool,
                                                  const pjsip_generic_array_hdr *rhs)
 {
     unsigned i;
-    pjsip_generic_array_hdr *hdr = PJ_POOL_ALLOC_T(pool, pjsip_generic_array_hdr);
+    pjsip_generic_array_hdr *hdr = pjsip_generic_array_hdr_create(pool, &rhs->name);
```
nanangizz, line 981: Similar to above.

ablangy: `pjsip_generic_array_hdr_create()` does the same by calling `pjsip_generic_array_hdr_init()` which allocate new space in the provided `pj_pool_t` for the header name of the cloned header via `pj_strdup()` (pjsip/src/pjsip/sip_msg.c:939).  
The problem doesn't exist with header which are not *PJSIP_H_OTHER* because the header name points to a static memory space defined by `const pjsip_hdr_name_info_t pjsip_hdr_names[];`.

ablangy:
Alright, I didn't get your point, now it's clear, you are right.  
This correction should do the job the same way the other "clone" functions do, but I missed the header type point.  
If I keep the code as it was and just add the missing `pj_strdup()`, then the cloned header list pointers still point to the source and not to the cloned header, which can also be an issue.  
A deep clone function shall create a new header and then copy the specific informations from the source, that way we do not have to be aware of the content of the 'pjsip_hdr' structure which can evolved with less impact on the existing code. This is what is more or less done in the other "clone" functions in the sip_msg.c file.  
I propose a solution close to the one applied to `pjsip_generic_string_hdr`, if you think it is not apropriate, I will restore the initial `pj_memcpy()` and set the cloned header type in the clone functions.

nanangizz:
> If I keep the code as it was and just add the missing `pj_strdup()`, then the cloned header list pointers still point to the source and not to the cloned header, which can also be an issue.

You're right.

> A deep clone function shall create a new header and then copy the specific informations from the source, that way we do not have to be aware of the content of the 'pjsip_hdr' structure which can evolved with less impact on the existing code. This is what is more or less done in the other "clone" functions in the sip_msg.c file. I propose a solution close to the one applied to `pjsip_generic_string_hdr`, if you think it is not apropriate, I will restore the initial `pj_memcpy()` and set the cloned header type in the clone functions.

You're right again.

Just perhaps creating `pjsip_generic_int_hdr_init2()+create2()` may not sound ideal because `generic` is basically corresponding to `PJSIP_H_OTHER` (including in `pjsip_generic_string_hdr`) and now it will be variable/settable, and the naming gives an impression like _overloading_ existing APIs while they are actually internal function helpers. IMO using the existing `pjsip_generic_int_hdr_create()` and overriding the header type should be sufficient and acceptable?

Also, just a minor, it is better to maintain the column width limit <= 80.

ablangy:
Alright, I have removed `pjsip_generic_int_hdr_init2()+create2()` and override the header type directly in the clone functions.

nanangizz (approved):

pjsip/src/pjsip/sip_msg.c
```diff
@@ -979,10 +982,12 @@ static pjsip_generic_array_hdr* pjsip_generic_array_hdr_clone( pj_pool_t *pool,
...
 {
     unsigned i;
-    pjsip_generic_array_hdr *hdr = PJ_POOL_ALLOC_T(pool, pjsip_generic_array_hdr);
+    pjsip_generic_array_hdr *hdr;
 
-    pj_memcpy(hdr, rhs, sizeof(*hdr));
-    for (i=0; i<rhs->count; ++i) {
+    hdr = pjsip_generic_array_hdr_create(pool, &rhs->name);
+    hdr->type = rhs->type;
+    hdr->count = rhs->count;
```
sauwming, line 989: for added security, `hdr->count = PJ_MIN(rhs->count, PJSIP_GENERIC_ARRAY_MAX_COUNT)`

ablangy:
`hdr->count = PJ_MIN(rhs->count, PJSIP_GENERIC_ARRAY_MAX_COUNT)` added

sauwming (approved):

PR #3970, opened 2024-05-24
Assertion on call hangup from DTMF callback

nanangizz:
To fix #3956.

Stream & transport still access their internal states after being destroyed by the call hangup invoked from within transport callback.

The fix is to move pool release to the transport's group lock handler. Also for added protection, add check for UDP transport validity after invoking callback.

Thanks @R-Jeske for the detailed report.

Issue #3956, opened 2024-05-14
Hangup inside a DTMF callback causes assertion


R-Jeske:
### Describe the bug

When i try to hangup a call inside a DTMF callback, the application crashes with an assertion in debug build and Segfault in release build.

### Steps to reproduce

add hangup in DTMF callback e.g. in PJSua App

```
static void call_on_dtmf_callback2(pjsua_call_id call_id, 
                                   const pjsua_dtmf_info *info)
{    
    char duration[16];
    char method[16];

    duration[0] = '\0';

    switch (info->method) {
    case PJSUA_DTMF_METHOD_RFC2833:
        pj_ansi_snprintf(method, sizeof(method), "RFC2833");
        break;
    case PJSUA_DTMF_METHOD_SIP_INFO:
        pj_ansi_snprintf(method, sizeof(method), "SIP INFO");
        pj_ansi_snprintf(duration, sizeof(duration), ":duration(%d)", 
                         info->duration);
        break;
    };    
    PJ_LOG(3,(THIS_FILE, "Incoming DTMF on call %d: %c%s, using %s method", 
           call_id, info->digit, duration, method));
+
+   pjsua_call_hangup(call_id, PJSIP_SC_GONE, NULL, NULL);
}
```
**Scenario:**

- A calls B
- B accepts call
- A sends DTMF
- B hangs up call in DTMF callback
- B cashes after return of `call_on_dtmf_callback2`

### PJSIP version

2.14.1

### Context

Appears on all tested platforms: 
- Linux x86_64, ARM, ARM64
- Win32

seems to be caused by destruction of transaction in `pjmedia/src/transport_udp.c` caused by `pjsua_call_hangup()`:

```
static void on_rx_rtp(pj_ioqueue_key_t *key,
                      pj_ioqueue_op_key_t *op_key,
                      pj_ssize_t bytes_read)
...
        //if (!discard && udp->attached && cb)
        if (!discard && 
            (-bytes_read != PJ_STATUS_FROM_OS(PJ_BLOCKING_ERROR_VAL))) 
        {
              TRACE_(("on_rx_rtp()","on_rx_rtp() 1d"));
            call_rtp_cb(udp, bytes_read, &rem_switch);  // <<< hangup is called in this callback
                TRACE_(("on_rx_rtp()","on_rx_rtp() 1e"));
        }
```

but afterwards following code is called without check, if it even exists anymore:

```
[...]

PR #4184, opened 2024-11-28
Fix stream lingering issues due to using transport's group lock

nanangizz:
To fix https://github.com/pjsip/pjproject/pull/3970#issuecomment-2499583512 by using its own group lock instead of transport's group lock.

Also, maintain ref to transport just in case stream still tries to send something after returning from callback which has just destroyed the stream.

### DIFF
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -2357,7 +2357,6 @@ static void on_rx_rtcp( void *data,
     }
 
     pjmedia_rtcp_rx_rtcp(&stream->rtcp, pkt, bytes_read);
-    pjmedia_rtcp_rx_rtcp(&stream->rtcp, pkt, bytes_read);
 }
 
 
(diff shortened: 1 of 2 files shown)


sauwming (approved):
Pls make sure that this doesn't cause issue #3956 to reappear. Thx.

nanangizz:
Tested with both scenarios: #3956 & https://github.com/pjsip/pjproject/pull/3970#issuecomment-2499583512, looks fine.

pjmedia/src/pjmedia/stream.c
```diff
@@ -2351,6 +2357,7 @@ static void on_rx_rtcp( void *data,
     }
 
     pjmedia_rtcp_rx_rtcp(&stream->rtcp, pkt, bytes_read);
+    pjmedia_rtcp_rx_rtcp(&stream->rtcp, pkt, bytes_read);
```
sauwming, line 2360: What is this for?

trengginas (approved):

Issue #4194, opened 2024-12-04
Default auth algorithm isn't being set when pjsip_cred_info.data_type is PJSIP_CRED_DATA_DIGEST


gtjoseph:
### Describe the bug

The new algorithm_type field in the pjsip_cred_info structure isn't being defaulted to PJSIP_AUTH_ALGORITHM_MD5 as it should be.  As a result if the user specifies a data_type of PJSIP_CRED_DATA_DIGEST but doesn't explicitly set algorithm_type, it will be left at PJSIP_AUTH_ALGORITHM_NOT_SET which will cause authentication to fail.

### Steps to reproduce

N/A

### PJSIP version

2.15

### Context

The issue happens if pjsip_cred_info.data_type is set to PJSIP_CRED_DATA_DIGEST but pjsip_cred_info.algorithm_type is not set.

### Log, call stack, etc

```shell
N/A
```

PR #4195, opened 2024-12-04
Set default pjsip_cred_info.auth_algorithm to PJSIP_AUTH_ALGORITHM_MD5

gtjoseph:
The new algorithm_type field in the pjsip_cred_info structure wasn't being
defaulted to PJSIP_AUTH_ALGORITHM_MD5 as it should be. As a result if the
user specified a data_type of PJSIP_CRED_DATA_DIGEST but didn't explicitly
set algorithm_type, it was left at PJSIP_AUTH_ALGORITHM_NOT_SET which
will caused authentication to fail.

* pjsip_cred_info.auth_algorithm is now correctly defaulted to
PJSIP_AUTH_ALGORITHM_MD5 when supplied via the pjsip_auth_lookup_cred and
pjsip_auth_lookup_cred2 callbacks and via the pjsip_auth_clt_set_credentials,
and pjsip_auth_create_digest functions.

* pjsip_cred_info.auth_algorithm now defaults to PJSIP_AUTH_ALGORITHM_SHA256
for the pjsip_auth_create_digestSHA256 function.

* Documentation was updated for those functions and callbacks to indicate the
defaults.

NOTE: The pjsip_auth_create_digest and pjsip_auth_create_digestSHA256
defaults are actually set in pjsip_auth_create_digest2 because those two
functions are now deprecated wrappers that pass cred_info (which is a const) to
pjsip_auth_create_digest2.  The documentation for pjsip_auth_create_digest2
however, doesn't mention defaults on purpose because it's a generic function
that handles multiple algorithms and we want users to specifify exactly what
algorithms they want to use to avoid ambiguity.

Resolves: #4194

### DIFF
--- a/pjsip/include/pjsip/sip_auth.h
+++ b/pjsip/include/pjsip/sip_auth.h
@@ -373,6 +373,10 @@ PJ_DECL(int) pjsip_cred_info_cmp(const pjsip_cred_info *cred1,
 /**
  * Type of function to lookup credential for the specified name.
  *
+ * \note If pjsip_cred_info::data_type is set to PJSIP_CRED_DATA_DIGEST and
+ * pjsip_cred_info::algorithm_type is left unset (0), algorithm_type will
+ * default to #PJSIP_AUTH_ALGORITHM_MD5.
+ *
  * @param pool          Pool to initialize the credential info.
  * @param realm         Realm to find the account.
  * @param acc_name      Account name to look for.
@@ -405,6 +409,10 @@ typedef struct pjsip_auth_lookup_cred_param
 
 /**
  * Type of function to lookup credential for the specified name.
+ *
+ * \note If pjsip_cred_info::data_type is set to PJSIP_CRED_DATA_DIGEST and
+ * pjsip_cred_info::algorithm_type is left unset (0), algorithm_type will
+ * default to #PJSIP_AUTH_ALGORITHM_MD5.
  *
  * @param pool          Pool to initialize the credential info.
  * @param param         The input param for credential lookup.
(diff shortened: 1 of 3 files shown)


sauwming (approved):

trengginas (approved):

nanangizz:
This behavior seems to be specified in `pjsip_auth_create_digest2()` declaration:
https://github.com/pjsip/pjproject/blob/8117bc593929b50f5f55402eae68b0bd7e4350ef/pjsip/include/pjsip/sip_auth.h#L797-L798)
while the 'defaulting' action seems to be done in `pjsip_auth_clt_set_credentials()`, so I think the specification needs to be moved to `pjsip_auth_clt_set_credentials()`?
Also, perhaps the spec should mention about "when data_type is set to `PJSIP_CRED_DATA_DIGEST`".

gtjoseph:
@nanangizz I updated the documentation, PR description and commit message.  Since the pjsip_auth_create_digest* functions are public, I also updated them.  See the "NOTE" in the PR description and see if that makes more sense.

nanangizz:
> ... and we want users to specifify exactly what
> algorithms they want to use to avoid ambiguity

Agree. Honestly I was wondering why not just enforce it, e.g: `pjsip_auth_clt_set_credentials()` returns error when `pjsip_cred_info::algorithm_type` is not set (or even when it is set but not supported).

nanangizz (approved):

gtjoseph:
> > ... and we want users to specifify exactly what
> > algorithms they want to use to avoid ambiguity
> 
> Agree. Honestly I was wondering why not just enforce it, e.g: `pjsip_auth_clt_set_credentials()` returns error when `pjsip_cred_info::algorithm_type` is not set (or even when it is set but not supported).

I didn't want to break backwards compatibility (any more than I already did). :)

PR #4200, opened 2024-12-06
Fixed AudioMediaPort use after free

sauwming:
As described in #3928 point 1, due to the now async nature of the conference bridge, we need to make sure that the port is not prematurely destroyed.

```
==41294==ERROR: AddressSanitizer: heap-use-after-free on address 0x6190000f3178 at pc 0x00010296eecc bp 0x00016dec8bb0 sp 0x00016dec8ba8
READ of size 8 at 0x6190000f3178 thread T19
    #0 0x10296eec8 in op_remove_port conference.c:1728
    #1 0x10296e1b4 in handle_op_queue conference.c:355
    #2 0x102969ae0 in get_frame conference.c:2416

0x6190000f3178 is located 1016 bytes inside of 1056-byte region [0x6190000f2d80,0x6190000f31a0)
freed by thread T0 here:
    #1 0x1024b1388 in MyAudioMediaPort::~MyAudioMediaPort()

### DIFF
--- a/pjmedia/src/pjmedia/port.c
+++ b/pjmedia/src/pjmedia/port.c
@@ -163,8 +163,12 @@ PJ_DEF(pj_status_t) pjmedia_port_init_grp_lock( pjmedia_port *port,
      * If media port is using app's pool, and the app destroys the port
      * and then destroys the pool immediately, it may cause crash as the port
      * may have not really been destroyed and may still be accessed.
-     * If port uses app's pool, it needs to implement on_destroy() for releasing
-     * the pool, so here we check availability of on_destroy implementation.
+     *
+     * Thus, media port needs to create and manage its own pool. Since
+     * app MUST NOT free this pool, port typically needs to implement
+     * on_destroy() for releasing the pool. Alternatively, port can also
+     * add a group lock handler for this purpose, but since we can't check
+     * this, here we just check the availability of on_destroy implementation.
      */
     if (port->on_destroy == NULL) {
         PJ_LOG(2,(THIS_FILE, "Warning! Port %s on_destroy() not found. To "
--- a/pjmedia/src/pjmedia/port.c
+++ b/pjmedia/src/pjmedia/port.c
@@ -166,10 +166,12 @@ PJ_DEF(pj_status_t) pjmedia_port_init_grp_lock( pjmedia_port *port,
      * If port uses app's pool, it needs to implement on_destroy() for releasing
      * the pool, so here we check availability of on_destroy implementation.
      */
-    if (!grp_lock && port->on_destroy == NULL) {
-        PJ_LOG(2,(THIS_FILE, "Warning! media port %s MUST either: - implement"
-                             " on_destroy() and release the pool there, or"
-                             " - supply its own group lock.",
+    if (port->on_destroy == NULL) {
+        PJ_LOG(2,(THIS_FILE, "Warning! Port %s on_destroy() not found. To "
+                             "avoid premature destroy, media port must "
+                             "manage its own pool, which can only be "
+                             "released in on_destroy() or in its grp lock "
+                             "handler. See PR #3928 for more info.",
                              port->info.name.ptr));
     }
 
--- a/pjmedia/src/pjmedia/port.c
+++ b/pjmedia/src/pjmedia/port.c
@@ -159,16 +159,17 @@ PJ_DEF(pj_status_t) pjmedia_port_init_grp_lock( pjmedia_port *port,
     PJ_ASSERT_RETURN(port && pool, PJ_EINVAL);
     PJ_ASSERT_RETURN(port->grp_lock == NULL, PJ_EEXISTS);
 
-    /* We need to be caution on ports that do not have its own pool,
-     * such port is likely using app's pool, so if the app destroys the port
+    /* We need to be extra cautious here!
+     * If media port is using app's pool, and the app destroys the port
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by sauwming** [comment]
To be honest, the existing port warning confused me.

If I just comment `port.on_destroy` in the patch, it will give the warning: `Warning, media port %s is using group lock, but it does not seem to have a pool.`, even though I already use my own pool, and it's not obvious what users should do next to fix the situation.

I'm not sure if my proposed fix is right though, so please correct it if it's inaccurate.

Please also help check for the other media ports in pjsua2 `media.cpp`, because I see similar patterns in Tonegen and AudioMedia, but since they use group lock, they should be safe?

**Comment by nanangizz** [comment]
> To be honest, the existing port warning confused me.
> 
> If I just comment `port.on_destroy` in the patch, it will give the warning: `Warning, media port %s is using group lock, but it does not seem to have a pool.`, even though I already use my own pool, and it's not obvious what users should do next to fix the situation.

Right, it sounds a bit confusing indeed. Let me try to clarify it, and then help me clarify it for port developers :)

To avoid premature destroy, port **must** use its own pool, never use app's pool (as app may call `pjmedia_port_destroy()` & release app's pool immediately after `pjmedia_conf_remove_port()`). Unfortunately the bridge cannot verify this (i.e: whether the port has & uses its own pool). So, I was thinking to piggybacking the helper function `pjmedia_port_init_grp_lock()` to check if the port has its own pool by checking `on_destroy` implementation (normally it is the place to release the pool, isn't it?), that's it :)

As a side note, port's group lock, while being needed to properly time the port's pool release, is actually optional, because the bridge can check and install automatically if there is none.

File: pjsip/src/pjsua2/media.cpp [hunk 1]
```diff
@@ -321,7 +326,8 @@ static pj_status_t get_frame(pjmedia_port *port, pjmedia_frame *frame)
 
 static pj_status_t put_frame(pjmedia_port *port, pjmedia_frame *frame)
 {
-    AudioMediaPort *mport = (AudioMediaPort *) port->port_data.pdata;
+    struct port_data *pdata = (struct port_data *)port->port_data.pdata;
+    AudioMediaPort *mport = pdata->mport;
```
nanangizz, line 330: Note that `get/put_frame()` may be called after the `AudioMediaPort` instance is destroyed, so perhaps `mport` field need to be reset in `AudioMediaPort` destructor. Even perhaps better to protect all access to the field with (group) lock.

sauwming:
> Right, it sounds a bit confusing indeed. Let me try to clarify it, and then help me clarify it for port developers :)

I would suggest:
(a) Warning, port %s on_destroy() not found. To avoid premature destroy, media port must manage its own pool. App shall not release this pool, but let the port do so in on_destroy() or in its grp lock handler.

(b) Warning, port %s on_destroy() not found. To avoid premature destroy, media port must manage its own pool, which can only be released in on_destroy() or in its grp lock handler.

(c) Warning, port %s on_destroy() not found. To avoid premature destroy, media port must control its own lifetime. See PR #3928.

pjsip/src/pjsua2/media.cpp
sauwming: Ouch, this is more complicated than I anticipated and never seems to be mentioned anywhere.
Perhaps need to add this in the PR desc: Due to the async nature of conf port removal, media port must be prepared to receive further get/put_frame() callbacks until the removal completes.

nanangizz: Added to backward compatibility issue in PR & 2.15 release note.

nanangizz:
> > Right, it sounds a bit confusing indeed. Let me try to clarify it, and then help me clarify it for port developers :)
> 
> I would suggest: (a) Warning, port %s on_destroy() not found. To avoid premature destroy, media port must manage its own pool. App shall not release this pool, but let the port do so in on_destroy() or in its grp lock handler.
> 
> (b) Warning, port %s on_destroy() not found. To avoid premature destroy, media port must manage its own pool, which can only be released in on_destroy() or in its grp lock handler.
> 
> (c) Warning, port %s on_destroy() not found. To avoid premature destroy, media port must control its own lifetime. See PR #3928.

I'd vote for (b), with "see PR #3928 for more info" perhaps?

pjmedia/src/pjmedia/port.c
```diff
@@ -159,16 +159,19 @@ PJ_DEF(pj_status_t) pjmedia_port_init_grp_lock( pjmedia_port *port,
...
     PJ_ASSERT_RETURN(port->grp_lock == NULL, PJ_EEXISTS);
 
-    /* We need to be caution on ports that do not have its own pool,
-     * such port is likely using app's pool, so if the app destroys the port
+    /* We need to be extra cautious here!
+     * If media port is using app's pool, and the app destroys the port
      * and then destroys the pool immediately, it may cause crash as the port
      * may have not really been destroyed and may still be accessed.
-     * When port has a pool, it usually implements on_destroy() for releasing
+     * If port uses app's pool, it needs to implement on_destroy() for releasing
```
nanangizz, line 166: "If port uses app's pool" implies that port may still use app's pool, while it must not. Even perhaps we should clarify somewhere here that port must create & use its own pool?

nanangizz (approved):
Did quick check on the code, Tonegen & AudioMedia seems okay. Also done a quick test for Tonegen using pjsua2_demo.cpp, while AudioMedia should already have been tested when testing Android.

PR #4201, opened 2024-12-06
Fixed alsa dev read pcm bug

freeforf:
Reading audio data from alsa may get stuck when plugging or unplugging the sound card device. So check if there is a frame of audio data in the cache before reading it.

### DIFF
--- a/pjmedia/src/pjmedia-audiodev/alsa_dev.c
+++ b/pjmedia/src/pjmedia-audiodev/alsa_dev.c
@@ -611,6 +611,15 @@ static int ca_thread_func (void *arg)
     while (!stream->quit) {
         pjmedia_frame frame;
 
+        result = snd_pcm_avail_update(pcm);
+        if (result == 0) {
+            snd_pcm_uframes_t avail;
+            avail = snd_pcm_avail(pcm);
+            if (avail > 1 && avail <= nframes) {
+                pj_thread_sleep(10);
+                continue;
+            }
+        }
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia-audiodev/alsa_dev.c
```diff
@@ -611,6 +611,15 @@ static int ca_thread_func (void *arg)
     while (!stream->quit) {
         pjmedia_frame frame;
 
+        result = snd_pc_avail_update(pcm);
```
sauwming, line 614: typo? `pc` should be `pcm`?

sauwming (commented):
Can you provide us a reference about this issue?
I can't seem to find any info in the official doc: https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#ga4c2c7bd26cf221268d59dc3bbeb9c048

pjmedia/src/pjmedia-audiodev/alsa_dev.c
freeforf: snd_pcm_avail_update

freeforf:
see demo https://www.alsa-project.org/alsa-doc/alsa-lib/_2test_2latency_8c-example.html#a33  line 817
use snd_pcm_wait wait for next event to read latency data or use snd_pcm_avail to read avail data .
maybe can use snd_pcm_wait  instead of pj_thread_sleep, i will test tomorrow.

PR #4205, opened 2024-12-09
pcaputil: Support parsing decoder options from FMTP input

orgads:
pcaputil: Support parsing decoder options from FMTP input

Some codecs need specific flags that are given in the SDP media description. For example, AMR has octet-align=1.

Instead of adding distinct flags for each codec, allow pcaputil to accept the fmtp content and parse the attributes from there.

Example:
```
pcaputil \
  --dst-port=52422 \
  --codec=AMR/8000 \
  --codec-fmtp='mode-set=0,1,2,3,4,5,6,7;octet-align=1' \
  amr.pcap amr.wav
```

### DIFF
--- a/pjsip-apps/src/samples/pcaputil.c
+++ b/pjsip-apps/src/samples/pcaputil.c
@@ -54,6 +54,8 @@ static const char *USAGE =
 "                           AES_CM_128_HMAC_SHA1_80 \n"
 "                           AES_CM_128_HMAC_SHA1_32\n"
 "  --srtp-key=KEY, -k     Set the base64 key to decrypt SRTP packets.\n"
+"  --codec-fmtp=FMTP      Set the fmtp input for parsing codec options.\n"
+"                         For example: \"mode-set=0;octet-align=1\".\n"
 #if PJMEDIA_HAS_OPUS_CODEC
 "  --opus-ch=CH           Opus channel count                            \n"
 "  --opus-clock-rate=CR   Opus clock rate                               \n"
@@ -93,6 +95,7 @@ struct args
     pjmedia_aud_dev_index dev_id;
     pj_str_t srtp_crypto;
     pj_str_t srtp_key;
+    pj_str_t codec_fmtp;
 #if PJMEDIA_HAS_OPUS_CODEC
     int opus_clock_rate;
     int opus_ch;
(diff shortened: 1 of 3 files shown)


trengginas (approved):

sauwming:
It's rather strange to pass the entire SDP as a parameter.
Why not just pass the codec specific parameters only?
e.g. `--codec-fmtp='octet-align=1' --codec-fmtp='mode-set=0,1,2,3,4,5,6,7'`

sauwming:
https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream_common.c#L110

You can make it simpler by requiring the parameter to have no whitespace, so:
- find '='
- if '=' exists {anything before '=' is fmtp name, anything after '=' is fmtp val } else {everything is fmtp val}

orgads:
> https://github.com/pjsip/pjproject/blob/master/pjmedia/src/pjmedia/stream_common.c#L110
> 
> You can make it simpler by requiring the parameter to have no whitespace, so:
> 
> * find '='
> * if '=' exists {anything before '=' is fmtp name, anything after '=' is fmtp val } else {everything is fmtp val}

So do you want me to factor this out? What about passing this to the codec? Should I also factor out parts of `pjmedia_stream_info_parse_fmtp`?

sauwming:
What I mean is to add new parameter `--codec-fmtp` and provide very simple parsing in `pcaputil` to put it in `dec_fmtp`.

orgads:
> What I mean is to add new parameter `--codec-fmtp` and provide very simple parsing in `pcaputil` to put it in `dec_fmtp`.

Done, with a small refactoring.

orgads:
The failed test looks unrelated to my changes. @sauwming please review.

pjsip-apps/src/samples/pcaputil.c
```diff
@@ -54,6 +54,7 @@ static const char *USAGE =
 "                           AES_CM_128_HMAC_SHA1_80 \n"
 "                           AES_CM_128_HMAC_SHA1_32\n"
 "  --srtp-key=KEY, -k     Set the base64 key to decrypt SRTP packets.\n"
+"  --codec-fmtp=FMTP      Set the fmtp input for parsing codec options.\n"
```
sauwming, line 57: Just provide an example here since it may not be obvious that it can contain multiple parameters separated by semi-colon.
For example: `'mode-set=0;octet-align=1'

sauwming (approved):

pjsip-apps/src/samples/pcaputil.c
orgads: Done

nanangizz (approved):

PR #4206, opened 2024-12-09
Fixed alsa dev read pcm bug

freeforf:
remove sound card device when capture audio data, the alsa audio cache may less than nframes, snd_pcm_readi wait for audio data forever.so when close the audio device ,the ca_thread_func can`t quite.use snd_pcm_avail to make sure audio cache has enough audio data.

### DIFF
--- a/pjmedia/src/pjmedia-audiodev/alsa_dev.c
+++ b/pjmedia/src/pjmedia-audiodev/alsa_dev.c
@@ -610,7 +610,20 @@ static int ca_thread_func (void *arg)
 
     while (!stream->quit) {
         pjmedia_frame frame;
-
+        
+        snd_pcm_state_t state;
+        state = snd_pcm_state(pcm);
+        if (state == SND_PCM_STATE_PREPARED) {
+            snd_pcm_start(pcm);
+            contine;
+        } else if (state == SND_PCM_STATE_RUNNING) {
+            snd_pcm_uframes_t avail;
+            avail = snd_pcm_avail(pcm);
+            if (avail >= 0 && avail < nframes) {
+                pj_thread_sleep(5);
+                continue;
+            }
+        }
         pj_bzero (buf, size);
         result = snd_pcm_readi (pcm, buf, nframes);
         if (result == -EPIPE) {


pjmedia/src/pjmedia-audiodev/alsa_dev.c
```diff
@@ -610,7 +610,13 @@ static int ca_thread_func (void *arg)
 
     while (!stream->quit) {
         pjmedia_frame frame;
-
+        
+        snd_pcm_uframes_t avail;
+        avail = snd_pcm_avail(pcm);
+        if (avail > 0 && avail <= nframes) {
```
sauwming, line 616: shouldn't this be `>=0 and < nframes`? We shouldn't need to sleep and continue when we already have `nframes`.

freeforf: when snd_pcm_prepare called,the snd_pcm_state is SND_PCM_STATE_PREPARED,snd_pcm_avail will return 0 always,need call snd_pcm_readi or snd_pcm_stop to change state to SND_PCM_STATE_RUNNING. i will update code with snd_pcm_state check later.

trengginas:
I tried to reproduce it here using a USB/removable sound card during a call/sound card is used, however it doesn't block instead printing this on the log:
```
ca_thread_func: error reading data!
```
Any specific scenario to trigger this issue?

freeforf:
set playback  device   plughw:usb sound card,and set capture device default.

trengginas:
> set playback device plughw:usb sound card,and set capture device default.

Still couldn't reproduce it. Based on the [doc](https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#ga4c2c7bd26cf221268d59dc3bbeb9c048) the `routine waits until all requested frames are filled`. In my case, if the sound card is removed, it will return immediately and return error.

freeforf:
i insert two same usb soundcard,if it doesn't work too,maybe my soundcard bugs.

sauwming:
Perhaps you can consider submitting your issue to https://github.com/alsa-project/alsa-lib/issues (https://www.alsa-project.org/wiki/Bug_Tracking).

PR #4154, opened 2024-11-13
Expand tab to spaces and and cleanup formatting

aol-nnov:
This is a followup for https://github.com/pjsip/pjproject/pull/3292 and https://github.com/pjsip/pjproject/pull/3210 but regarding aconfigure.ac

PR #4212, opened 2024-12-11
VS build system files fixes

LeonidGoltsblat:
- added some missing files to pjlib project
- added missing configuration to pjsystest project and pjproject-vs14.sln
- build tools set changed from v140 (VS2015) to v143 (VS2022)


sauwming (changes requested):
need a cleaner patch that only contains the difference and doesn't mark the entire file as modified.

sauwming:
> cleaned!
> 
> git ls-files --eol *.vcxproj* shows that the repository has mixed line endings in VS build system files maybe it makes sense to add .gitattributes and run at least git add --renormalize *.vcxproj*

Thanks.
Yes, tidying up the line endings can be done in a separate PR (similar to https://github.com/pjsip/pjproject/pull/4154), and then put the revision in `.git-blame-ignore-revs` so it won't show up in git blame.

sauwming:
Looks okay to me, but since I rarely touch Windows, I'll leave the approval to @nanangizz and @trengginas .

nanangizz:
Tested with VS 2022 only, looks okay.
[The docs](https://docs.pjsip.org/en/latest/get-started/windows/build_instructions.html#tools-and-sdks) says we support older VS versions (2015-2019), perhaps @trengginas can check.

trengginas:
Tested on VS2015 and VS2019

nanangizz:
> Tested on VS2015 and VS2019

Both can build the lib? Any additional steps, e.g: changing the build tools set version to older/140, required? As I read somewhere that VS2015 cannot use build tool set 143 (which is set in the property sheet).

trengginas:
For VS2019 and VS2015, change the `BuildToolset` on `pjproject-vs14-common-config.props`

https://github.com/LeonidGoltsblat/pjproject/blob/31f8676c37847060da6ace988a160f2905b3f723/build/vs/pjproject-vs14-common-config.props#L26

To v140
Notes: for VS2019 it supports buildtoolset up to v142.

nanangizz:
> For VS2019 and VS2015, change the `BuildToolset` on `pjproject-vs14-common-config.props`
> 
> https://github.com/LeonidGoltsblat/pjproject/blob/31f8676c37847060da6ace988a160f2905b3f723/build/vs/pjproject-vs14-common-config.props#L26
> 
> To v140 Notes: for VS2019 it supports buildtoolset up to v142.

Okay, perhaps it is better to also add info about how to set the property sheet to docs?

nanangizz (approved):

trengginas (approved):

PR #4222, opened 2024-12-18
Limit number of Oboe auto restart

### LINKED ISSUE #4218: The ExtraAudioDevice is constantly refreshing the log  and crash
### Describe the bug

Hello, @sauwming
Based on your suggestions in #4204 and #4199, I have used the oboe method to obtain information about the audio devices on my Android device. My current device is equipped with one built-in sound card and one USB sound card. 
I hope to route two phone calls to these two sound cards separately, for instance, using the built-in sound card for call 6001 and the USB sound card for call 6002.

Following your guidance, I adopted the ExtraAudioDevice, but I discovered that after executing ExtraAudioDevice.open(), my logcat continuously refreshes logs related to AudioHardwareTiny due to "Oboe playback got an empty queue", as detailed below:
``` 
2024-12-16 15:45:36.318   256-8186  AudioHardwareTiny       pid-256                              D  Pos: 1734335136318818200 19984
2024-12-16 15:45:36.334   256-8186  AudioHardwareTiny       pid-256                              D  Pos: 1734335136334710243 20748
2024-12-16 15:45:36.337  8128-8188  oboe_dev.cpp            com.test.pjsiptest                   W  Oboe playback got an empty queue
2024-12-16 15:45:36.350   256-8186  AudioHardwareTiny       pid-256                              D  Pos: 1734335136350786912 21520
2024-12-16 15:45:36.366   256-8186  AudioHardwareTiny       pid-256                              D  Pos: 1734335136366799121 22288
2024-12-16 15:45:36.371  8128-8188  oboe_dev.cpp            com.test.pjsiptest                   W  Oboe playback got an empty queue
```

My code is as follows:
``` kotlin
val audioDevManager= endpoint.audDevManager()
audioDevManager.setNullDev()
builtInExtraAudioDevice = ExtraAudioDevice(1,2)
builtInExtraAudioDevice.open()
usbExtraAudioDevice = ExtraAudioDevice(5,6)
usbExtraAudioDevice.open()
```
In the code, the index 1, 2, 5, and 6 are obtained based on the device data, as you can see from the logs.
```
2024-12-16 15:45:35.060  8128-8128  System.out              com.test.pjsiptest                   I  15:45:35.059           oboe_dev.cpp  ..Oboe audio device initialized with 7 device(s):
2024-12-16 15:45:35.060  8128-8128  System.out              com.test.pjsiptest                   I  15:45:35.060           oboe_dev.cpp  .. 0 (native id=0): Default (in+out)
2024-12-16 15:45:35.060  8128-8128  System.out              com.test.pjsiptest                   I  15:45:35.060           oboe_dev.cpp  .. 1 (native id=2): Builtin Speaker - B10A (out)
2024-12-16 15:45:35.060  8128-8128  System.out              com.test.pjsiptest                   I  15:45:35.060           oboe_dev.cpp  .. 2 (native id=5): Builtin Mic - B10A (in)
2024-12-16 15:45:35.061  8128-8128  System.out              com.test.pjsiptest                   I  15:45:35.061           oboe_dev.cpp  .. 3 (native id=6): HDMI - B10A (in)
2024-12-16 15:45:35.061  8128-8128  System.out              com.test.pjsiptest                   I  15:45:35.061           oboe_dev.cpp  .. 4 (native id=8): Unknown (25) - B10A (in)
2024-12-16 15:45:35.061  8128-8128  System.out              com.test.pjsiptest                   I  15:45:35.061           oboe_dev.cpp  .. 5 (native id=230): USB Headset - USB-Audio - C-Media USB Headphone Set (out)
2024-12-16 15:45:35.061  8128-8128  System.out              com.test.pjsiptest                   I  15:45:35.061           oboe_dev.cpp  .. 6 (native id=234): USB Headset - USB-Audio - C-Media USB Headphone Set (in)
```
At this point, I haven't even started a call, yet these logs are constantly being refreshed. What could be the reason for this?
If I comment out the code for ExtraAudioDevice.open(), these logs disappear.
Below are two sets of logs: one where the program crashes and another where the program does not crash but keeps refreshing the logs.
[log1.txt](https://github.com/user-attachments/files/18146693/log1.txt)
[log2.txt](https://github.com/user-attachments/files/18146694/log2.txt)

### Steps to reproduce

1

### PJSIP version

2.14.1

### Context

Android

### Log, call stack, etc

```shell
1
```

sauwming:
To fix #4218.

It is reported that Oboe will keep restarting indefinitely upon error.

Also in this PR:
minor: remove unused label `on_skip_dev`.

### DIFF
--- a/pjmedia/src/pjmedia-audiodev/oboe_dev.cpp
+++ b/pjmedia/src/pjmedia-audiodev/oboe_dev.cpp
@@ -39,6 +39,9 @@
 #include <android/log.h>
 #include <oboe/Oboe.h>
 
+/* Setting of maximum number of consecutive stream restart attempts. */
+#define MAX_RESTART     3
+
 /* Device info */
 typedef struct aud_dev_info
 {
@@ -416,7 +419,6 @@ static pj_status_t oboe_refresh(pjmedia_aud_dev_factory *ff)
 
         f->dev_count++;
 
-    on_skip_dev:
         jni_env->DeleteLocalRef(jdev_info);
     }
 
@@ -530,7 +532,7 @@ class MyOboeEngine : oboe::AudioStreamDataCallback,
 public:
     MyOboeEngine(struct oboe_aud_stream *stream_, pjmedia_dir dir_)
     : stream(stream_), dir(dir_), oboe_stream(NULL), dir_st(NULL),
-      thread(NULL), thread_quit(PJ_TRUE), queue(NULL),
+      thread(NULL), thread_quit(PJ_TRUE), nrestart(0), queue(NULL),
       err_thread_registered(false), mutex(NULL)
     {
         pj_assert(dir == PJMEDIA_DIR_CAPTURE || dir == PJMEDIA_DIR_PLAYBACK);
@@ -775,11 +777,20 @@ public:
             pj_status_t status;
 
             PJ_LOG(3,(THIS_FILE,
-                      "Oboe stream %s error (%d/%s), "
-                      "trying to restart stream..",
+                      "Oboe stream %s error (%d/%s)",
                       dir_st, result, oboe::convertToText(result)));
 
             Stop();
+
+            if (nrestart >= MAX_RESTART) {
+                PJ_LOG(3, (THIS_FILE, "Oboe %s stream permanently stopped",
+                                      dir_st));
+                pj_mutex_unlock(mutex);
+                return;
+            }
+
+            nrestart++;
+            PJ_LOG(3,(THIS_FILE, "Trying to restart Oboe %s stream", dir_st));
             status = Start();
 
             if (status != PJ_SUCCESS) {
@@ -881,6 +892,7 @@ private:
                     /* Increment timestamp */
                     pj_add_timestamp32(&this_->ts, ts_inc);
                     ++cnt;
+                    this_-> nrestart = 0;
                 }
 
                 if (stop_stream)
@@ -895,6 +907,8 @@ private:
                               this_->dir_st, cnt));
                 }
             } else {
+                this_->nrestart = 0;
+
                 /* Get audio frame from app via callback play_cb() */
                 pjmedia_frame frame;
                 frame.type = PJMEDIA_FRAME_TYPE_AUDIO;
@@ -914,6 +928,7 @@ private:
 
                 /* Increment timestamp */
                 pj_add_timestamp32(&this_->ts, ts_inc);
+                this_-> nrestart = 0;
             }
         }
 
@@ -930,6 +945,7 @@ private:
     const char                  *dir_st;
     pj_thread_t                 *thread;
     volatile pj_bool_t           thread_quit;
+    unsigned                     nrestart;
     sem_t                        sem;
     pj_atomic_queue_t           *queue;
     pj_timestamp                 ts;
--- a/pjmedia/src/pjmedia-audiodev/oboe_dev.cpp
+++ b/pjmedia/src/pjmedia-audiodev/oboe_dev.cpp
@@ -924,7 +924,6 @@ private:
 
                 /* Increment timestamp */
                 pj_add_timestamp32(&this_->ts, ts_inc);
-                this_->nrestart = 0;
             }
         }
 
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia-audiodev/oboe_dev.cpp
```diff
@@ -775,11 +777,21 @@ class MyOboeEngine : oboe::AudioStreamDataCallback,
...
-                      "Oboe stream %s error (%d/%s), "
-                      "trying to restart stream..",
+                      "Oboe stream %s error (%d/%s)",
                       dir_st, result, oboe::convertToText(result)));
 
             Stop();
+
+            if (nrestart >= MAX_RESTART) {
+                PJ_LOG(3, (THIS_FILE, "Oboe %s stream permanently stopped",
+                                      dir_st));
```
nanangizz, line 787: "permanent" because the nrestart will never be reset, so the instance is no longer usable?
Perhaps it can be avoided by adding param to `Start()`, something like `inc_counter` flag, which should be true only if it is called by error case? Not sure if it is necessary though.
This may also avoid putting `nrestart = 0` in many places?

sauwming: OK, remove the log. Now it will use the already existing log message `Oboe stream restart failed`. Will also notify the callback via `PJMEDIA_EVENT_AUD_DEV_ERROR` event, so this should be better.

Should I still modify the `nrestart` counter handling?
IMO, it still needs to be reset when the stream is working (i.e. if it successfully captures or plays an audio frame).

nanangizz: > OK, remove the log. Now it will use the already existing log message `Oboe stream restart failed`. Will also notify the callback via `PJMEDIA_EVENT_AUD_DEV_ERROR` event, so this should be better.

Okay.

Just a note: putting `++nrestart` inside `PJ_LOG(3, ..)` is not ideal, as it will not be executed (even not compiled) when `PJ_MAX_LOG_LEVEL` is set to <3?

> Should I still modify the `nrestart` counter handling? IMO, it still needs to be reset when the stream is working (i.e. if it successfully captures or plays an audio frame).

Yes, it needs to be reset, just the reset places looks rather strange (multiple lines inside the callback), can it be put just once around here for example? Or there are cases that `Start()` may succeed but somehow the callback is not invoked?
https://github.com/pjsip/pjproject/blob/43bdeaa7e8d6e4165b327bfadbd0d79e69ec68dc/pjmedia/src/pjmedia-audiodev/oboe_dev.cpp#L685-L687

sauwming: It can't be put there (in ~L685) because then Start() will immediately reset the counter (after just incremented).

I move it to onAudioReady() instead.

nanangizz: Still don't understand why it can't, as the reset is supposed to be done after the start succeeds (and not reset when fails). And putting it in callback make the reset will be done every callback instead of once only, should work just fine just rather seem strange/unnecessary :)
Also there seem to still another line that reset it in the thread loop too.

Anyway it is just a minor.

nanangizz (approved):

pjmedia/src/pjmedia-audiodev/oboe_dev.cpp
sauwming: Start() usually just returns success, this is because requestStart() is asynchronous and returns immediately.
(https://google.github.io/oboe/classoboe_1_1_audio_stream.html#a3c484e314dee8dfed1d419f487b5d601)

nanangizz: Ah okay, now it makes sense.

sauwming: Alternatively, perhaps we can use blocking start() instead?
```
virtual Result oboe::AudioStream::start(int64_t timeoutNanoseconds = kDefaultTimeoutNanos)

Start the stream. This will block until the stream has been started, an error occurs or 
timeoutNanoseconds has been reached.
```

I didn't change it because I don't know the reason why the non-blocking version was selected to be used in the first place.

nanangizz: I can't remember it either, perhaps to avoid UI hiccup, e.g: make call always tries opening sound device first. I think better to just leave it for now?

PR #4223, opened 2024-12-19
Fix wrong event state callback and re-subscribe method on event subscription.

trengginas:
There are some issues related to #3754:
- Wrong callback on event: TSX_STATE when using presence subscription
The callback `on_buddy_evsub_dlg_event_state()` is called instead of `on_buddy_evsub_state()`.
- Wrong method called to re-subscribe from timer callback
When a dialog-event subscription is sent, and responded with `496/Bad Event`, the timer is setup for re-subcription and call `pjsua_buddy_update_pres()`.


nanangizz (approved):

sauwming (approved):
For the title, better put something more descriptive rather than cite the PR number. This will help understand what the issue is, as well as make it easier to search in the future.

The related PR number can be put in the PR description (which you have already done).

PR #4236, opened 2025-01-02
Fixes GnuTLS unit test failures

sauwming:
* Fixes protocol setting in GnuTLS.
Currently, by default GnuTLS will enable all TLS protocols. This will cause unit test failure such as:
`echo test w/ incompatible proto: server TLSv1 vs client SSL3`.
The patch fixes this by only enabling the protocols specified in the ssock param.
* Fixes cipher setting in GnuTLS
Currently, by default GnuTLS will enable all ciphers.
The patch fixes this by only enabling the ciphers specified in the ssock param.
* Fixes crash if handshake failed.
```
11:02:28.798 Handshake failed on 127.0.0.1:61429 in connecting to 127.0.0.1:61428: Invalid value or argument (PJ_EINVAL)
Error: signal 11:
3   libgnutls.30.dylib                  0x0000000100c830ac gnutls_alert_send + 168
4   libgnutls.30.dylib                  0x0000000100c56d94 gnutls_bye + 92
5   pjlib-test-arm-apple-darwin24.1.0   0x00000001006b6070 ssl_destroy + 188
6   pjlib-test-arm-apple-darwin24.1.0   0x00000001006aa2f8 ssl_on_destroy + 48
7   pjlib-test-arm-apple-darwin24.1.0   0x00000001006aa988 pj_ssl_sock_close + 1152
8   pjlib-test-arm-apple-darwin24.1.0   0x000000010061d248 ssl_on_connect_complete + 2248
9   pjlib-test-arm-apple-darwin24.1.0   0x00000001006b38f4 on_handshake_complete + 4964
```
According to the doc https://linux.die.net/man/3/gnutls_bye seems to be only needed if the connection has been established.

Also in this PR:
- Disable ssl sock test with SSLv23 protocol. This used to be specific to OpenSSL backend (which is then expanded to other backends as well), but now even for OpenSSL, it has been deprecated:
https://docs.openssl.org/master/man3/SSL_CTX_new/#notes
```
SSLv23_method(), SSLv23_server_method(), SSLv23_client_method()
These functions do not exist anymore, they have been renamed to TLS_method(),
```

### DIFF
--- a/pjlib/src/pj/ssl_sock_gtls.c
+++ b/pjlib/src/pj/ssl_sock_gtls.c
@@ -407,7 +407,10 @@ static pj_status_t tls_str_append_once(pj_str_t *dst, pj_str_t *src)
         if (dst->slen + src->slen + 3 > 1024)
             return PJ_ETOOMANY;
 
-        pj_strcat2(dst, ":+");
+        if (dst->slen > 0 && dst->ptr[dst->slen-1] == ':')
+            pj_strcat2(dst, "+");
+        else
+            pj_strcat2(dst, ":+");
         pj_strcat(dst, src);
     }
     return PJ_SUCCESS;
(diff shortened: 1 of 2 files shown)


nanangizz:
Re: disabling SSLv23 protocol. Rather OOT, even TLS 1.0 & 1.1 are [deprecated](https://therecord.media/ietf-officially-deprecates-tls-1-0-and-tls-1-1) already, so perhaps we should shift our focus toward TLS 1.2 & TLS 1.3 in TLS socket backends & tests.

sauwming:
> Re: disabling SSLv23 protocol. Rather OOT, even TLS 1.0 & 1.1 are [deprecated](https://therecord.media/ietf-officially-deprecates-tls-1-0-and-tls-1-1) already, so perhaps we should shift our focus toward TLS 1.2 & TLS 1.3 in TLS socket backends & tests.

Ok, created #4237 for this. Feel free to add/edit the issue page if needed.

bennylp (approved):

PR #4239, opened 2025-01-03
Only enable DTLS if SSL backend is OpenSSL

sauwming:


nanangizz:
Rather than adding checks multiple times, what if the check is inserted into `pjmedia/config.h`, e.g: somewhere after :
https://github.com/pjsip/pjproject/blob/653ff5774a627885748a37739e2503c1425303b9/pjmedia/include/pjmedia/config.h#L1089-L1091

```
/* Currently SRTP-DTLS requires OpenSSL */
#if PJMEDIA_SRTP_HAS_DTLS
#  if PJ_SSL_SOCK_IMP != PJ_SSL_SOCK_IMP_OPENSSL
#    undef PJMEDIA_SRTP_HAS_DTLS
#    define PJMEDIA_SRTP_HAS_DTLS 0
#  endif
#endif
```

sauwming:
Overwriting the setting in config.h may not be obvious for users that has enabled `PJMEDIA_SRTP_HAS_DTLS` and expects it to work.

How about adding it at the beginning of file `transport_srtp.c` instead?
We can also add `#warn` there to notify user.

nanangizz:
> Overwriting the setting in config.h may not be obvious for users that has enabled `PJMEDIA_SRTP_HAS_DTLS` and expects it to work.

Yes, it disables the DLTS silently, but effectively so does the original patch?

> How about adding it at the beginning of file `transport_srtp.c` instead?

Okay too. The point is better avoid such long check & multiple times, for example if in the future DTLS may also use another TLS backend, so the check will get longer and prone to missing update.

> We can also add `#warn` there to notify user.

Was thinking about compile warning too, but not sure if there is a standard way (or gnu-C89 way :D) for generating it.

sauwming:
> > Overwriting the setting in config.h may not be obvious for users that has enabled `PJMEDIA_SRTP_HAS_DTLS` and expects it to work.
> Yes, it disables the DLTS silently, but effectively so does the original patch?

True, but the build error will point to `transport_srtp.c` so it's easier to figure out.
(for background info, when I tested with gnutls, the SRTP failed to build and it's not immediately obvious for me why).

> > How about adding it at the beginning of file `transport_srtp.c` instead?
> Okay too. The point is better avoid such long check & multiple times, for example if in the future DTLS may also use another TLS backend, so the check will get longer and prone to missing update.

Makes sense.

> > We can also add `#warn` there to notify user.
> Was thinking about compile warning too, but not sure if there is a standard way (or gnu-C89 way :D) for generating it.

I was considering between `#warning`, `#pragma message()` and `printf()`, and it seemed that the second one would be the most acceptable. We've also been using it in various places:
```
./pjmedia/include/pjmedia/config.h:#       pragma message("Warning: PJMEDIA_HAS_SMALL_FILTER macro is deprecated"\
./pjlib/include/pj/compat/limits.h:#  pragma message("limits.h is not found or not supported. LONG_MIN and "\
./pjlib/include/pj/config.h:#       pragma message("Warning: PJ_ENABLE_EXTRA_CHECK macro is deprecated"\
```

nanangizz (approved):

Issue #2251, opened 2019-11-14
Deadlock between PJSUA LOCK and conference mutex


pjsipbot:
_2019-11-14 06:04:55: @mingteluu_ _created the issue on trac ticket [2251](https://trac.pjsip.org/repos/ticket/2251)_

The old issue of #2115 and #1464 still persists, because it can affect other callbacks as well, not just on_call_state(). If inside any pjsua callback which holds PJSUA LOCK, the app calls any pjsua conference/video conference API (which will try to acquire conference mutex), the deadlock can potentially occur with media ports' callbacks, which hold conference mutex, and if the app calls any pjsua API from there.

The best solution seems to be not holding the conference lock when calling media port's eof callbacks. In this ticket, we will try to achieve this by using media event mechanism to call those callbacks.

-*IMPORTANT**: Note that this ticket will deprecate pjmedia_*_set_eof_cb() and pjmedia_*_set_cb() and replace it with pjmedia_*_set_eof_cb2()/set_cb2(), and the callbacks will now be called asynchronously.
___
_2019-11-29 04:19:51: @mingteluu_ 

___
_2019-11-29 04:20:11: @mingteluu_ _edited the issue description_

___
_2019-11-29 04:21:19: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2019-11-29 04:21:19: @mingteluu_ _set **owner** to **@mingteluu**_

___
_2019-11-29 04:21:19: @mingteluu_ _set **resolution** to **fixed**_

___
_2019-11-29 04:21:19: @mingteluu_ _commented_

In [r6112](../commit/e4721473789ad143ad77f41f1ccfc1fa5cfa9589):
Fixed #2251: Deadlock between PJSUA LOCK and conference mutex

___
_2019-11-29 04:23:17: @mingteluu_ _edited the issue description_

___
_2019-11-29 04:25:20: @mingteluu_ _edited the issue description_

___
_2020-01-10 07:09:52: @nanangizz_ _commented_

In [r6131](../commit/74eb0a02890ee8c2bb8f689a010a35f4dd72a583):
Re #2251: Fixed bug in memory player, wrong user data passed in invoking callback. Thanks to Marcus Froeschl for the report and the fix.

___
_2020-01-10 07:26:37: @nanangizz_ _commented_

In [r6132](../commit/85887b23678bccc4029881847786cdf1499c52dc):
Re #2251: Fixed bug in memory capture, wrong user data passed in invoking callback.

PR #4241, opened 2025-01-05
Parallel conference bridge

LeonidGoltsblat:
The strictly sequential and single-threaded conference bridge of pjsip must service all connected ports within a single timer tick, inevitably leading to limitations on the number of serviced ports and high CPU performance requirements. The goal of this pull request is to implement parallel servicing of conference bridge ports while preserving the switch behavior as much as possible and minimizing changes to the original codebase.

Parallelism is implemented using the OpenMP C/C++ Application Program Interface (API) (hereinafter referred to as OpenMP), which allows declarative parallel execution of code segments that were not initially designed for parallel processing. OpenMP is cross-platform and supported by the vast majority of C language compilers (and other languages). This solution uses a set of OpenMP tools limited to version 2.0, an older standard version. This ensures that all modern compilers can compile and execute this code correctly.

OpenMP support is not enabled by default; each development environment must be explicitly configured to use OpenMP, which is disabled by default. This ensures compatibility of the proposed solution with applications that do not require parallel switching. No changes are required for such applications.

The changes to the source code primarily involve two aspects:

- Instead of using the shared buffer provided by the master port for all ports to read data, each port reserves its own buffer and reads data into it. This enables parallel reading of data by different ports. (This approach uses memory less efficiently; in practice, a buffer is needed not for each port but for each thread.)

- Since data from different input ports may need to be mixed into the buffer of the same output port to create a conference, access to this buffer must be synchronized. For this purpose, each port creates its own lock using OpenMP tools, which is then used to synchronize the mixing of data from different sources.

The entire get_frame() function (as before) is divided into three steps:

1. Initialization
2. Reading and mixing data
3. Transmitting data

Each of steps 2 and 3 represents an OpenMP parallel execution region. The steps themselves are executed sequentially, meaning the next step begins only after all tasks from the previous step have been completed by all threads.

Unlike step 2, during step 3, the ports operate completely independently; no changes to the switching scheme affect the data processing for the ports. Therefore, the tasks in the asynchronous switching queue are executed concurrently with the main tasks of step 3. To prevent premature deletion of ports with transmitters, the grplock reference count of such ports is incremented during initialization (step 1). The reference count is decremented when the port processing is completed in step 3. This ensures that even if an OP_REMOVE_PORT operation is performed concurrently with data transmission in step 3, the physical resources will only be released once they are no longer in use.

In this version, the remaining risk of deadlocks after the introduction of asynchronous switching #3928 has also been resolved. The grplock handle, called within the OP_REMOVE_PORT operation under the protection of the conference bridge mutex, could previously initiate other locks in an unpredictable order. This was a potential source of deadlocks. However, the asynchronous switching algorithm ensures that such operations are executed by only one thread at a time. Therefore, these operations do not require additional synchronization. For this reason, the execution of asynchronous operations has been moved out from under the protection of the conference bridge mutex. This mutex is required only for the asynchronous operation queue, not for the operations themselves.

Other changes in the code are minor and not critical to the overall optimization concept.

- For instance, the aforementioned step 1 does not initialize the output port buffers. Instead, ports store a "timestamp" of the last frame loaded into the buffer. If this timestamp differs from the timestamp of the frame being loaded, the first (and possibly only) frame is loaded into the buffer. This allows for straightforward copying without prior initialization and without unnecessary summing with zero. Importantly, the timestamp itself only increases and, therefore, generally does not require separate initialization.

- The OP_ADD_PORT operation has been excluded from asynchronous operations. This version of the conference bridge does not use port counters or the is_new flag. Instead, lower and upper bounds for the range of active ports are maintained. An "active" port is defined as one that is connected to something, i.e., has a non-zero number of listeners or transmitters. Only such ports participate in the sound transmission process. Immediately after being added, a port is not yet active and does not affect the operation of the conference bridge. The appearance of such a port concurrently with the execution of get_frame() does not influence the execution of steps 1-3 in any way.

- Creating new ports can also be performed "practically" in parallel. Ideally, it would be "fully parallel," but reserving a slot in the port array, though performed in O(1) time, still requires mutex locking for a short duration (by default, pj_stack is not used). Once the slot is reserved, further creation and initialization of the port within the reserved slot can be carried out concurrently with similar actions in other slots.

The only expected change in behavior is not directly related to parallelism and involves special handling of the PJ_EEOF code returned by read_port(). This code is interpreted as a signal from the port that it no longer has and will not produce any new data (e.g., a fileplayer has reached the end of the file). In this case, rx_setting = PJMEDIA_PORT_DISABLE is triggered, ensuring that no further attempts are made to retrieve data from the port. This also prevents repeated triggering of the eof_cb() and eof_cb2() callbacks, sparing the application from unnecessary calls.

Incidentally, the asynchronous switching implementation in PR #3928 resolved many deadlock-related issues. Among other improvements, it eliminated the need to prohibit very convenient synchronous callbacks (see https://github.com/pjsip/pjproject/issues/2251).


sauwming:
I like the premise of the introduction of parallelism. I'm curious whether you have any performance speedup data.

It must be noted though, that the additional compile-time switches will undoubtedly make conference much much more difficult to maintain and debug should issues arise. So I wonder whether the performance boost is worth the additional complexity.

LeonidGoltsblat:
> I'm curious whether you have any performance speedup data.

OpenMP conference bridge was inroduced 3-4 years ago.
Previously, speech intelligibility disappeared after 30-40 ports. The OpenMP conference bridge version with OMP_NUM_THREADS=8 works fine with 240 ports without any degradation in audio quality. Obviously, this is far from the limit, but there were no tests with higher density.

sauwming:
I like the idea of adding an additional layer (i.e. OpenMP stub -- perhaps this can be put in pjlib) to avoid compilation-time switches. I believe this is necessary if we want to integrate this, otherwise the conference code will become such a nightmare to read.

So for me, I vote towards adopting this (with the condition of removing the compile-time switches). But before we proceed further, let's hear first what others think about this parallel feature. 

Also, your note about #2251 is interesting. So should we undeprecate/reactivate the callbacks `eof_cb()` now that conf is async, @nanangizz?

nanangizz:
> I like the idea of adding an additional layer (i.e. OpenMP stub -- perhaps this can be put in pjlib) to avoid compilation-time switches. I believe this is necessary if we want to integrate this, otherwise the conference code will become such a nightmare to read.
> 
> So for me, I vote towards adopting this (with the condition of removing the compile-time switches). But before we proceed further, let's hear first what others think about this parallel feature.
> 

Yes, I think it is a good idea to wrap OpenMP in PJLIB (or perhaps PJLIB-UTIL?), for readibility/maintainability & platform compatibility.

Also, we use background processing in some places already (e.g: job queue, worker thread, event manager), this new framework (background & multiprocessing) may standardize them perhaps.

> Also, your note about #2251 is interesting. So should we undeprecate/reactivate the callbacks `eof_cb()` now that conf is async, @nanangizz?

Maybe :)
The deadlock should be no longer a problem. But there is also a side effect of such synchronous callback: blocking the conf clock (even with parallel conf, as at some steps it may still need sync all workers, haven't checked the details). So IMHO it is still reasonable to encourage app to use the eof_cb2() instead of eof_cb() to avoid possible performance blocker.

bennylp:
First of all, thank you for the patch submission again. This is really interesting and exciting, especially to hear that the parallel version can achieve much higher performance than the plain one.

However, unfortunately it is implemented using OpenMP... :) OpenMP is quite a "beast" to support. It's (too) high level, too implicit, requires support from many tools (compilers, debuggers), will require changes in build commands, require another skill set to master, and last but not least, not supported by iOS and Android (at least officially). I would very much prefer it to be implemented using `pj_threads`, so it's automatically as portable as the rest of the code.

If there is something like "thread pool" in pjlib (similar to [Python's process pool](https://docs.python.org/3/library/multiprocessing.html#using-a-pool-of-workers)), would it help? (and more importantly, are you willing to change it :)

Or if you want to submit this as is, then I think the best way is to "fork" `conference.c` into new implementation. e.g. `conf_openmp.c`, activated by something like `PJMEDIA_CONF_USE_OPENMP`. But then this file will be less maintained.

LeonidGoltsblat:
Short answer: Let's try. Please wait for a more detailed answer in a couple of days. I'm a bit busy right now.

LeonidGoltsblat:
More Detailed Answer

A quick web search shows that OpenMP is supported by the Android NDK starting with r11 (though this information should be verified!). However, iOS still does not officially support OpenMP.

**> something like "thread pool" in pjlib**

Yes! I’ve always wondered how pjsip works without a thread pool! ☺

As a general rule, Windows programs rarely create threads explicitly. Instead, they register callbacks of various types (e.g., IO, events, timers, etc.) with a thread pool managed by the OS kernel ([Microsoft Docs on Thread Pools](https://learn.microsoft.com/en-us/windows/win32/procthread/thread-pools)).

With more information about the workload, the OS can manage the thread pool more efficiently than an application, for example, by deciding whether to start a new thread if all pool threads are waiting. It would be highly beneficial to have platform-dependent thread pool support in pjlib!

The main challenge preventing the use of the Windows thread pool API right now is the need to register a pj_thread_t object for each thread, which is inconvenient for threads created and managed outside the application’s control. To avoid memory leaks, a "wrapper" is required to register the thread at the start of the callback and unregister it at the end. However, it's unclear how to make this wrapper completely transparent to the application.

**> to "fork" conference.c**

The current conference implementation already includes numerous enhancements unrelated to OpenMP:

- Concurrent, synchronous creation of conference ports
- A more stable and a little bit efficient looping mechanism based on lower and upper boundaries instead of using port_counter and is_new flags
- More efficient get_frame() initialization based on timestamps, which doesn’t require zeroing
- Optimizations for 1-to-1 connections
- etc

I propose integrating these enhancements first, then creating a "forked" version, and only after that proceeding with further work on parallelism

**"In the real world"**

In practice, I use a heavily optimized IOCP queue and RTP transport. I haven’t tested the standard ioqueue implementation, so I’m unsure how it behaves in multithreaded scenarios. It’s possible that both ioqueue and transport may require optimization after parallelizing the conference bridge (in my case, the bridge was optimized last).

bennylp:
Thanks for Android OpenMP info.

The initial idea for the thread pool is a high level pool of pj threads, unlike Windows thread pool which is a low level OS object, I think. Let me check if it can be abstracted using the same API. But the main objective of pj thread pool is to execute N jobs using M threads and as replacement for OpenMP.

Yes I notice there are many other changes in conf unrelated to parallelism. It would be better to submit them as separate enhancements.

Thanks for the communication. Although our high level API is geared towards client, the "core" was intended to be high performant, hence this topic is very interesting to us.

LeonidGoltsblat:
Hi!
Apologies for the delay.

Here is the updated version that avoids the use of OpenMP functions and conditional compilation where possible. Instead, only OpenMP pragmas are utilized, though some OpenMP functions are retained for debugging purposes.

**Key notes:**

- This version does **not** require any modifications to the build system. By default, it operates without OpenMP and runs as a "normal" sequential conference bridge.
- Users who wish to enable OpenMP will need to configure their build environment independently, relieving the PJSIP team of the need to support this framework.
- Hopefully, this variant is stable enough to serve as a foundation for the next step ("standard PJ threading").
- I aim to proceed with the next step in the coming days.
- It would be great if this version could be merged into the main branch of the project to serve as a starting point for future enhancements.

**Minor change:**
For compatibility with #4253, the is_new flag had to be reinstated. While adding a port still operates synchronously and doesn’t rely on this flag, it would be ideal to explicitly check for the absence of a clock. If the clock is absent, all operations (not just OP_REMOVE_PORT) should be performed synchronously.


> The initial concept for the thread pool is a high-level pool of PJ threads, in contrast to the Windows thread pool, which seems to be a low-level OS object. Let me verify whether it can be abstracted using the same API. However, the primary goal of the PJ thread pool is to execute N jobs using M threads, serving as a replacement for OpenMP.

Could you clarify: does a PJ thread pool currently exist, or is it a planned API? Such an API would significantly simplify the implementation of a multi-threaded conference bridge.

**Regarding the Windows thread pool API:**
This is not just a low-level API—it supports operations ranging from low-level (e.g., I/O) to high-level "Work." This aligns with your description: "to execute N jobs using M threads." Relevant functions include [CreateThreadpoolWork](https://learn.microsoft.com/en-us/windows/win32/api/threadpoolapiset/nf-threadpoolapiset-createthreadpoolwork) and [SubmitThreadpoolWork](https://learn.microsoft.com/en-us/windows/win32/api/threadpoolapiset/nf-threadpoolapiset-submitthreadpoolwork). These functions seem particularly suited for parallel conference bridges.

The primary obstacle to leveraging this well-designed API is the current requirement in PJSIP to register threads without the option to unregister them. A similar challenge arises in the OpenMP version and, more broadly, in scenarios where the user program doesn’t control the creation and termination of threads.

Implementing the ability to unregister threads would be highly beneficial. This would enable support for platform-dependent pools like the Windows thread pool.

LeonidGoltsblat:
Hi, everybody!
I am pleased to announce that my promise has been fulfilled! The parallel conference bridge is now implemented using the pj_thread API!

In this commit besides conference.c:
- conf_openmp.c
- some OS dependent files introduced barrier syncronization API into pjsip.

Shortly about this two punkts, then about parallel bridge:

a) To compile the OpenMP conference bridge, the user must add
```c
#define PJMEDIA_CONF_USE_OPENMP 1
```
line to config_site.h AND configure the development environment to use OpenMP. However, the main motivation for this may be testing and comparing the performance of different implementations. My tests show that the OpenMP implementation has no advantages over the "native" pj_thread threading.

b) The native multithreading used with the parallel conference bridge uses a synchronization barrier and required the implementation of this API. The API is implemented in 5 (phew!) variants:
- 3 for different versions of Windows (Windows 8 and above uses its own barrier implementation, Vista and above uses condition variable-based implementation, and older versions (default for pjsip) use semaphore-based implementation (isn't it time to define PJ_WIN32_WINNT as at least _WIN32_WINNT_VISTA (0x0600) or even as _WIN32_WINNT_WIN8 (0x6002) ?) ,
- and 2 for Unix (POSIX pthread_barrier_t, supported since POSIX.1-2001 and classic condition variable-based)
The Windows implementation is quite debugged, the Unix implementation is based on pj_event code, looks simple and clear, but is NOT tested... I don't have access to a Linux machine with developer tools right now. Any assistance would be helpful,  please review this code mindfully!
Note about enum pj_barrier_flags: These flags are only supported in the Windows 8 and higher implementation, but manipulating them has resulted in a 5-10% reduction in CPU usage (from 40-% to 30+%). It's worth the effort!

About the parallel bridge implementation

The OpenMP-based implementation uses the processor cores allocated to it at 100%. For example, if the host has 40 cores and OMP_NUM_THREADS=8, 8 cores are used at about 100%, and 32 at about 0%, the average load is 20%. If OMP_NUM_THREADS==8 and the host has 8 cores, then the load is 100% (actually 90-100%).
On the same 8-core host, the "native" implementation with 240 ports (under Windows 10) works with a load of 20-30% with peaks of up to 40%. (Do we have a 3-4 times performance reserve?...) For OpenMP, the processor load does not depend on the application load: there is no difference between 1 port and 240 ports. The "native" practically does not use resources when there is no work. This does not tell us that the algorithm is extremely efficient, but it does tell us that OpenMP probably uses a spin loop to wait for work and is unaware of the conference bridge timer and other implementation details. Perhaps OpenMP is designed for a different type of workflow. 
The current parallel bridge algorithm is optimized for 1-to-1 switching. So we immediately mix the data in the mix_buf listener and transmit the data (call write_port()) immediately after all sources have received data for the current listener. This works fine for 1-to-1 switching, but should not work so well for conference switching (m-to-n). The current algorithm acquires a lock on the listener side to protect the mix_buf listener that has more than 1 source, so threads that have source data for the same listener will block on this lock. There is an obvious optimization: have a queue on the listener side, then sources can add a reference to their receive buffers to this queue and continue processing other listeners without contention. This optimization will not give us a noticeable reduction in CPU load, but it will increase the stability of the wake time on each timer tick and reduce the duration of this time somewhat, so it may improve the audio quality. However, this optimization does require a lock-free queue... We are waiting for pj_pool_aligned_alloc().

By default, the conference bridge is serial. At compile time, the user can define the macro PJ_CONF_BRIDGE_MAX_THREADS. This is the number of threads that the conference bridge should use. This value is used to determine whether the conference bridge should be implemented as a parallel bridge or not. If this value is 1, the conference bridge will be implemented as a serial bridge, otherwise it will be implemented as a parallel bridge. The current implementation uses a static thread pool, as an optimization, it would be better to dynamically determine the required number of threads of the thread pool.

bennylp:
Thanks! Sorry for the late reply, I'm reviewing it now and will get back with more detailed replies.

bennylp (changes requested):
I think in general this is great, thanks for the hard work! I'm okay in general. I haven't reviewed `conference.c` in detailed, because I thought we should be addressing these points below first because they will modify the patch significantly.

1. I would like to inform you about our coding style (which I just wrote in more detail, previously it was too vague :), pls have a look at https://docs.pjsip.org/en/coding-style/get-started/coding-style.html

2. If you have unit test codes, it would be good to include as well (for the barrier, and conf maybe). If not, I can help with writing it.

3. Activating the parallel feature (proposal, others pls comment):

At compile time, user should have option to disable multithreading code. This is activated by (new) macro PJMEDIA_CONF_HAS_THREADS, which default value is PJ_HAS_THREADS. I don't require that when threading is disabled we must use the old code, because this could result in too many variants of code.

At run time, user should have the option to control the number of worker threads to use, from 0-N. Zero means the operations will be done only by get_frame() thread. I propose creating new API as follows:

```
typedef struct pjmedia_conf_param
{
	unsigned max_slots;
	unsigned sampling_rate;
	unsigned channel_count;
	unsigned samples_per_frame;
	unsigned bits_per_sample;
	unsigned options;
	unsigned worker_threads;
} pjmedia_conf_param;

PJ_INLINEvoid) pjmedia_conf_param_default(*param)
{
  pj_bzero(param);
}

PJ_DEF(pj_status_t) pjmedia_conf_create2(pool, pjmedia_conf_param *, *p_port);
```

LeonidGoltsblat:
> If you have unit test codes, it would be good to include as well (for the barrier, and conf maybe). If not, I can help with writing it.

@bennylp Yes, that would be helpful. I don't have tests.

LeonidGoltsblat:
implemented compile time and runtime conf bridge multithreading control.
Compile time: PJMEDIA_CONF_THREADS (prefferably for test applications which is not pjsua/pjsua2)
Runtime: MediaConfig::confThreads and pjsua_media_config::conf_threads 
In the latter cases PJMEDIA_CONF_THREADS is the default value.
Everything is done. Please review.

pjmedia/include/pjmedia/conference.h
```diff
@@ -104,6 +118,87 @@ enum pjmedia_conf_option
...
+    /**
+     * Bitmask options to be set for the bridge. The
+     * options are constructed from #pjmedia_conf_option
+     * enumeration.
+     * The default value is zero.
+     */
+    unsigned options;
+
+    /** 
+     * The number of worker threads to use.
```
sauwming, line 173: Add doc that the value will not be used for switchboard.

Also in pjsua and pjsua2.

pjmedia/src/pjmedia/conference.c
```diff
@@ -31,7 +34,13 @@
 #include <pj/pool.h>
 #include <pj/string.h>
 
-#if !defined(PJMEDIA_CONF_USE_SWITCH_BOARD) || PJMEDIA_CONF_USE_SWITCH_BOARD==0
+#if defined(PJ_STACK_IMPLEMENTATION)
```
sauwming, line 37: Since it's not integrated yet (and still not clear whether it will be), better remove this and all the stack usages first for now.

Alternatively, you can re-create a PR for the atomic list as suggested in the previous PR so it can be merged first.

pjmedia/src/pjmedia/conference.c
```diff
@@ -240,20 +315,62 @@ struct pjmedia_conf
...
+                                         *   connected port's slot          */
+
     unsigned              connect_cnt;  /**< Total number of connections    */
     pjmedia_snd_port     *snd_dev_port; /**< Sound device port.             */
     pjmedia_port         *master_port;  /**< Port zero's port.              */
     char                  master_name_buf[80]; /**< Port0 name buffer.      */
     pj_mutex_t           *mutex;        /**< Conference mutex.              */
     struct conf_port    **ports;        /**< Array of ports.                */
-    unsigned              clock_rate;   /**< Sampling rate.                 */
+    unsigned              sampling_rate;/**< Sampling rate.                 */
```
sauwming, line 332: Not sure what the purpose for this change?

pjmedia/src/pjmedia/conference.c
```diff
@@ -824,11 +1071,48 @@ PJ_DEF(pj_status_t) pjmedia_conf_create( pj_pool_t *pool_,
...
+        pjmedia_conf_destroy(conf);
+        return status;
+    }
+#else
+    conf->unused_slots = PJ_POOL_ZALLOC_T(pool, unused_slots_cache);
+    pj_list_init(conf->unused_slots);
+#endif
+    conf->free_port_slots = pj_pool_calloc(pool, conf->max_ports, sizeof(port_slot));
+    PJ_ASSERT_ON_FAIL(conf->free_port_slots, {pjmedia_conf_destroy(conf); return PJ_ENOMEM;});
+    unsigned i = conf->max_ports;
```
sauwming, line 1087: put variable declaration on top of the block, otherwise it may fail to compile on older VS.

pjmedia/src/pjmedia/conference.c
```diff
@@ -2835,7 +3610,7 @@ static pj_status_t get_frame_pasv(pjmedia_port *this_port,
 /*
  * Recorder (or passive port) callback.
  */
-static pj_status_t put_frame(pjmedia_port *this_port, 
+static pj_status_t put_frame(pjmedia_port *this_port,
```
sauwming, line 3613: *cosmetic

pjmedia/src/pjmedia/conference.c
```diff
@@ -856,36 +1140,91 @@ static pj_status_t resume_sound( pjmedia_conf *conf )
 /**
  * Destroy conference bridge.
  */
-PJ_DEF(pj_status_t) pjmedia_conf_destroy( pjmedia_conf *conf )
+PJ_DEF(pj_status_t) pjmedia_conf_destroy(pjmedia_conf *conf)
```
sauwming, line 1143: if possible, refrain from modifying code white spaces. it creates an unnecessarily larger patch that's more difficult to review. I will mark several of these (but not all so it doesn't flood the review) with "*cosmetic" in the comment.

pjmedia/src/pjmedia/conference.c
```diff
@@ -2845,7 +3620,7 @@ static pj_status_t put_frame(pjmedia_port *this_port,
     /* Check for correct size. */
     PJ_ASSERT_RETURN( frame->size == conf->samples_per_frame *
                                      conf->bits_per_sample / 8,
-                      PJMEDIA_ENCSAMPLESPFRAME);
+                      PJMEDIA_ENCSAMPLESPFRAME );
```
sauwming, line 3623: *cosmetic

pjmedia/src/pjmedia/conference.c
LeonidGoltsblat: Yes, I will create a PR for the atomic list, but I would like to run these PRs in parallel rather than sequentially.
Currently PJ_STACK_IMPLEMENTATION is only defined in my "realword application" and should not be a problem for anyone else. Missing the related code would force me to maintain another special version of the conference bridge and would be a problem for me.
I plan to create another PR after the the atomic list PR to integrate it into conference bridge and implement m-to-n switching optimization.

pjmedia/src/pjmedia/conference.c
LeonidGoltsblat: "cosmetic" has been fixed, where found 
PJSIP requires a more complex .editorconfig

pjmedia/src/pjmedia/conference.c
LeonidGoltsblat: see #4330

LeonidGoltsblat:
Improvements:
- m-to-n switching optimization (threads that have RX data to mix in the same TX mix_buf do not wait for each other, instead a queue of mixing operations is implemented on the TX side. Only one thread at a time processes this queue, all others process other TX or RX ports)
- conference test (2-to-1 mixing) added to systest
- algorithm simplification (instead of 2 barriers, now only 1 is used, only the get_frame() thread must wait until all worker threads have finished their work)

Fixes:
- eliminated residual echo on sound port after disconnection
- we can't ignore the listener even if the transmitter doesn't have a frame to transmit. The heartbeat frame must be sent anyway.  No problems on the RX side should prevent the TX side from sending heartbeat frames.

Minor:
- all pj_statck references replaced with pj_atomic_list references
- while #4330 is approved but not yet merged, the generic slist implementation is included directly in conference.c and will be removed after merging #4330

It's time to review!

LeonidGoltsblat:
The silence has dragged on for a bit. Please provide feedback.

sauwming:
In general, the patch looks okay (at least to me).
But considering the massive size of the patch, and the highly critical area that it modifies (conference is obviously one of the most important component of an audio call), admittedly it's not easy to integrate this, even with the potentially impressive improvement it promises. Because the risk is high too, if it breaks then everything breaks.

So as @bennylp mentioned above (https://github.com/pjsip/pjproject/pull/4241#issuecomment-2589326944), this may need to be broken down further, to make it easier to review and less risky to integrate. So barrier may need to go to a separate PR, all non-parallel enhancements too. The smaller the PR size the better.

But there's no action needed from you for now (not just yet). The first step IMO is to review the atomic slist, so its ugly inclusion here in this patch can be removed. I have re-requested review from the rest of the team, hopefully we can hear from them.

bennylp:
Sorry for the lack of response on my side lately, I will review soon

bennylp:
@LeonidGoltsblat are you using the multithreaded `conference.c` or `conf_openmp.c` in your day to day app? If you're not using the multithreaded `conference.c`, then that is a concern since we have this huge modification without anyone testing/using it.

LeonidGoltsblat:
The idea behind Open MP is to quickly convert single-threaded code to multi-threaded. Although the conversion was not fast and easy, we now have native multi-threaded code. There is no point in maintaining or using the Open MP variant at this point. I only use "native" and I think it's time to remove conf_openmp.c from the repository. It should also be noted that the "native" algorithm is not so different from the OpenMP algorithm.

 

My “day to day app” currently exists as a test environment supporting heavy traffic generation inside the test environment, "native multi-threading" has been tested under high load. Some bugs have been found and fixed.

 

Are there any more bugs in parallel logic? Maybe yes, but I don't know them yet...

Currently I know some bugs in "asynchronous" logic, but they are not related to "parallel" logic.

LeonidGoltsblat:
@bennylp, it seems this PR didn't attract much interest. However, your question prompted me to reanalyze the code.
I found that my version significantly changes the behavior compared to the original (your) implementation: mine sends data (including heartbeat packets) only to connected ports (i.e., those with transmitters), while yours sends to all registered ports (except those that are disabled, etc.).
I do have a fix for this (and it's not too complex), but as usual, it's not ready for a push, since it's not documented well enough.
I'm not sure whether it makes sense to keep pursuing an idea that you and your team seem to have lost interest in.

heretic13:
@LeonidGoltsblat
I really want this feature. And I continue to monitor this topic.
But I use the library, I don't develop it.

sauwming:
We sincerely apologize for the delay. Please rest assured that it’s not due to a lack of interest. As previously discussed—and as you’ve also recently discovered—the risk of behavioral changes and compatibility breaks is significant. This puts us in a challenging position every time we revisit the integration. That said, I assure you that this time, we are fully committed to seeing it through to completion, starting with the notes and proposed changes outlined below.

To minimize the risk of introducing regressions, we’ve explored several options, such as integrating the changes into a separate branch or even releasing a different PJSIP version. Ultimately, we’ve decided to maintain a safety net by leaving the current conference implementation largely intact and placing the new feature in a separate file. This approach allows us to preserve the existing functionality while still enabling integration, usage, and more comprehensive testing of the new feature.

Here is a summary of the plan:
	1.	Please confirm whether you still require the atomic slist implementation (#4330). If so, we will review and merge it first.
	2.	For this PR:
a. Keep the changes to `conference.c` as minimal as possible.
b. Duplicate `conference.c` into a new file named `conf_thread.c`, and add the parallel thread-based feature there.
c. Remove `conf_openmp` as it is no longer needed.
d. Introduce a compile-time switch `PJMEDIA_CONF_USE_XXX` to allow users to choose the conference implementation.
e. For now, the public default will remain the existing (non-parallel) implementation. However, our internal PJSIP team will use your parallel version. This will allow us to test it extensively every time we use PJSIP. Once it has been proven stable and reliable over time, we will promote it as the default and begin deprecating the legacy implementation.

Thank you again for your patience and contributions.

LeonidGoltsblat:
@heretic13, thank you for support!

@sauwming
Ok, let's try again.

Yes, the atomic slist implementation (#4330) is required. I would also appreciate it if you could take a look at #4360 as well.

It would be helpful to understand how long you plan to continue internal testing. Overall, the plan seems reasonable, although it may result in the need to fix the same issues in two different versions.

Let’s start with the PRs mentioned above, and then we can discuss the details further.

sauwming:
The two PRs have been merged, so you may proceed.

As for how long the test would be, it's difficult to say at the moment. It depends on how many issues we'll be finding and what the users' feedback would be. If everything goes well, I suppose the transition could be completed within two cycles of PJSIP version releases.

LeonidGoltsblat:
I would like to suggest preparing two additional pull requests that are formally not related to the conference bridge logic. The first would extract the barrier code into a separate PR, and the second would cover the conference bridge tests. Although I used these tests to debug parallel logic, they are expected to work correctly with single-threaded execution and are therefore not tied to multithreading. This approach would allow us to better isolate the changes specific to the bridge itself and subsequently provide them to you for internal testing. 
I’d be glad to hear your thoughts on this approach.

LeonidGoltsblat:
Ok!

LeonidGoltsblat:
And now, please, pay attention to #4449.

LeonidGoltsblat:
I hope this commit is final and ready for review and internal testing.

- Replaced backend selection macros with PJMEDIA_CONF_BACKEND to avoid ambiguity with PJMEDIA_CONF_USE_SWITCH_BOARD.
- Removed conf_openmp.c.
- Added conf_thread.c.
- In conference.c, added pjmedia_conf_create2() and included comments about some detected (minor) issues—marked with //TODO but not fixed.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -7552,6 +7552,22 @@ struct pjsua_media_config
      */
     unsigned            max_media_ports;
 
+    /**
+     * Total number of threads that can be used by the conference bridge
+     * including get_frame() thread.
+     * This value is used to determine if the conference bridge should be
+     * implemented as a parallel bridge or not.
+     * If this value is set to 1, the conference bridge will be implemented as a
```
sauwming: Since it is unsigned, it's simpler to simply state:
"If the value is set to 0 or 1"
and remove `Should not be less than 1.`.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -7552,6 +7552,22 @@ struct pjsua_media_config
      */
     unsigned            max_media_ports;
 
+    /**
+     * Total number of threads that can be used by the conference bridge
+     * including get_frame() thread.
+     * This value is used to determine if the conference bridge should be
+     * implemented as a parallel bridge or not.
```
sauwming, line 7573: Put empty line (line containing only '*') in between long paragraphs for easier doc readability.
Also below (after `as a parallel bridge`).

pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -297,14 +297,26 @@ pj_status_t pjsua_aud_subsys_init()
         opt |= PJMEDIA_CONF_USE_LINEAR;
     }
 
+    pjmedia_conf_param param;
```
sauwming: var declaration needs to be at the beginning of block.

pjlib/include/pj/config.h
```diff
@@ -1244,6 +1244,47 @@
...
+/* Conference bridge sequential backend */
+#define PJMEDIA_CONF_SERIAL_BRIDGE_BACKEND 1
+/* Multithreaded conference bridge backend */
+#define PJMEDIA_CONF_PARALLEL_BRIDGE_BACKEND 2
+
+/**
+ * Choose which conference backend implementation to use.
+ * It is not enough to choose PJMEDIA_CONF_PARALLEL_BRIDGE_BACKEND for
+ * real parallelism. For this purpose, the programmer must
+ * - define PJMEDIA_CONF_THREADS with a value > 1 (and optionally
```
sauwming: Better reword it in a positive instruction to make it clear what users must do to take advantage of the parallelism feature (so instead of telling users what's not enough to do, tell them what to do), i.e.:

```
In order to use parallel conference bridge, users need to:
1. define PJMEDIA_CONF_BACKEND to PJMEDIA_CONF_PARALLEL_BRIDGE_BACKEND
2. define PJMEDIA_CONF_THREADS with a value greater than 1
```

Even though 1 is optional, I think it's good practice to explicitly specify it, so users have a clear sense of which backend they're currently using.

pjlib/include/pj/config.h
```diff
@@ -1244,6 +1244,47 @@
 #   endif // PJ_WIN32
 #endif  //PJ_ATOMIC_SLIST_IMPLEMENTATION
 
+/**
```
sauwming: Shouldn't this be in pjmedia config?

pjmedia/include/pjmedia/conference.h
```diff
@@ -208,6 +222,94 @@ enum pjmedia_conf_option
...
+
+
+/**
+ * Initialize conference bridge creation parameters.
+ */
+PJ_INLINE(void) pjmedia_conf_param_default(pjmedia_conf_param *param)
+{
+    pj_bzero(param, sizeof(pjmedia_conf_param));
+    /* Set the default values */
+    param->worker_threads = PJMEDIA_CONF_THREADS-1;
```
sauwming, line 298: check if PJMEDIA_CONF_THREADS < 1.

pjmedia/src/pjmedia/conference.c
```diff
@@ -617,6 +617,9 @@ static pj_status_t create_pasv_port( pjmedia_conf *conf,
     if (status != PJ_SUCCESS)
         return status;
 
+    //TODO: potential bug! Is this intended? delay_buf is created in
```
sauwming, line 643: Need to fix this, in a separate PR perhaps?

LeonidGoltsblat: The easiest way to fix this is to rename conf_thread.c to conference.c. :smile:
Let's do that soon!

On a more serious note...

The asynchronous implementation of the conference bridge has much more fundamental issues. Almost all of them stem from the coexistence of synchronous and asynchronous operations: some flags can be set but not read; some are set synchronously but used asynchronously without any protection; others are set under a mutex but accessed later without any synchronization. In some cases, mutex protection exists for synchronous reads, but not for modifications or asynchronous access.

All of this probably needs to be addressed at some point. However, we now have two backends (and as of yesterday, even three—with the addition of the SIMD backend). So I think we should wrap up this PR first, and only then decide whether it's worth continuing further development in this area.

pjlib/include/pj/config.h
LeonidGoltsblat: done

pjlib/include/pj/config.h
LeonidGoltsblat: done

pjmedia/include/pjmedia/conference.h
LeonidGoltsblat: done

pjmedia/src/pjmedia/conf_thread.c
```diff
@@ -0,0 +1,3858 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA 
+ */
+
+/* MP related modification by Leonid Goltsblat <lgoltsblat@gmail.com> 2021-2025 */
```
sauwming: Since it no longer uses OpenMP, it's not obvious what MP here stands for.

pjmedia/src/pjmedia/conf_thread.c
```diff
@@ -0,0 +1,3858 @@
...
+    * 
+    * Enforce "this frame is from the first transmitter" condition:
+    * initially we set last_timestamp to -1, so the first frame will be copied.
+    */
+    conf_port->last_timestamp.u64 = (pj_uint64_t)-1;
+
+    if (conf->is_parallel) {
+        PJ_TEST_NOT_NULL(conf_port->rx_frame_buf = (pj_int16_t*)pj_pool_zalloc(pool, conf->rx_frame_buf_cap), NULL, 
+                         {status = PJ_ENOMEM;goto on_return;});
+        PJ_TEST_SUCCESS(status=pj_lock_create_simple_mutex(pool, "tx_lock", &conf_port->tx_lock), NULL, goto on_return);
```
sauwming: Although it offers the desired functionality here, I believe unittest.h, and PJ_TEST* is not supposed to be used other than for testing purposes.

If you need it, I would suggest to define your own macro in this file.

sauwming (approved):
That's all from me. Note that I will be away the entire next week, so we will wait for others to review this and hopefully can merge it the week after.

LeonidGoltsblat:
**Inspired by PR #4468 — particularly the impressive “2400 calls” benchmark, which demonstrates performance competitive with Dialogic HMP operating entirely at the OS kernel level (with a declared capacity of up to 8000 G.711 RTP streams) — I introduced only minimal code changes to enable auto-vectorization by the Visual Studio compiler. As a result, all data transformation loops in the parallel backend (mixing, level adjustment, copying) are now automatically vectorized by MSVC.**

The outcome is remarkable: my standard 240-stream test, which previously required **8 threads**, now runs on a **single thread** (just the clock thread, no worker threads) with **0% CPU load**!

The circle is complete — **vectorization has proven to be significantly more effective than parallelization** for the conference bridge workload. In fact, the code and algorithmic optimizations implemented in this PR — which enabled easy vectorization — turned out to be more impactful than multithreading support itself.

I believe it would be beneficial to improve the precision and structure of the pjmedia and RTP-related code to support auto-vectorization — **which would require only minimal changes to the source code**. (Visual Studio handles this automatically, while GCC requires enabling specific compiler flags — these are project-level adjustments, not source-level changes.)

sauwming:
That sounds great! Though it does mean one fewer platform (i.e., Windows) for testing the multithreaded conference.

sauwming:
Note that there's a recent PR affecting conference in #4482 and #4489.

LeonidGoltsblat:
The current master branch is merged with this branch.
However, I am alone, I can't win the race with the whole community. Is it time for a merge?

and as for the Windows platform - it is the only platform where multi-threaded conferencing has been subjected to serious testing!

sauwming:
Yes, I’ll go ahead and merge this PR so we can begin testing. Since the legacy implementation remains available and largely unchanged, this should be a safe path forward for us and all PJSIP users.

LeonidGoltsblat:
More about auto-vectorization.
The initial results were overstated due to a faulty test environment. After fixing it, I confirmed that vectorization alone cannot handle 240 ports on a single thread without unacceptable delays. However, when combined with parallelism (8 threads), vectorization brings a major quality improvement: maximum tick delay drops from 0.3–0.5 s to below 0.01 s.
In summary, vectorization is very effective, but only together with parallelization, and it would be worth reviewing the media/RTP code to better support compiler auto-vectorization.

P.S. I don't see any pull requests with fixes to conf_thread.c. This could be due to one of two reasons: no bugs, or no usage.

heretic13:
> 
> P.S. I don't see any pull requests with fixes to conf_thread.c. This could be due to one of two reasons: no bugs, or no usage.

.... or lack of time. Parallel conference is a good development for the future of this library.
But I don't have time to test it at the moment (Win + Linux Arm).

sauwming:
I've been using it for regular use and internal testing, and so far it works as intended.
We've also informed our users, especially the ones that create multiple conference bridges, to try this one, but of course  since it will require them to modify their app code, it may take some time for them to adopt this.

LeonidGoltsblat:
Hi, Yaroslav (@heretic13)! I see you are the most active supporter of the parallel conf bridge! Thanks for your interest! I would be glad to get to know you better! If it is not secret: where are you from? What developments are you working on?
Best regrads, 
Leonid

saghul:
👋  Is there some documentation on where the parallel bridge is an improvement over the non-parallel one? Is that always the case? What if there are just a couple of connected ports?

heretic13:
@LeonidGoltsblat.
Hi, Leonid.
I live in the CIS. I work in the field of system programming (Windows and Linux) in C++. Several of my projects use SIP telephony, including embedded solutions. Therefore, the ability to parallelize the load between several cores of a processor of average computing power is important for me.

LeonidGoltsblat:
¡Hola, Saúl! In low density systems such as "a couple of connected ports", parallel bridging will neither improve nor reduce performance. 
Keep in mind that by default, a parallel bridge behaves like a serial bridge and should not affect the behavior of your system. You don't need parallelism unless you have performance issues.

PR #4248, opened 2025-01-13
Fixed pjsua2 unnecessary object copy

sauwming:
Fix coverity warning in pjsua2:
Variable copied when it could be moved
Unnecessary object copies can affect performance.
(https://scan6.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_COPY_INSTEAD_OF_MOVE)

Basically, when we insert (push_back) an object into a vector, we can use `std::move` (instead of relying of copy ctor), if the object will no longer be used for any other purpose.

### DIFF
--- a/pjsip/src/pjsua2/account.cpp
+++ b/pjsip/src/pjsua2/account.cpp
@@ -65,7 +65,7 @@ void RtcpFbConfig::fromPj(const pjmedia_rtcp_fb_setting &prm)
     for (unsigned i = 0; i < prm.cap_count; ++i) {
         RtcpFbCap cap;
         cap.fromPj(prm.caps[i]);
-        this->caps.push_back(std::move(cap));
+        this->caps.push_back(PJSUA2_MOVE(cap));
     }
 }
 
@@ -96,7 +96,7 @@ void RtcpFbConfig::readObject(const ContainerNode &node) PJSUA2_THROW(Error)
         NODE_READ_NUM_T         (cap_node, pjmedia_rtcp_fb_type, cap.type);
         NODE_READ_STRING        (cap_node, cap.typeName);
         NODE_READ_STRING        (cap_node, cap.param);
-        this->caps.push_back(std::move(cap));
+        this->caps.push_back(PJSUA2_MOVE(cap));
     }
 }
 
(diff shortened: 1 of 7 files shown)


pjsip/src/pjsua2/account.cpp
```diff
@@ -96,7 +96,7 @@ void RtcpFbConfig::readObject(const ContainerNode &node) PJSUA2_THROW(Error)
         NODE_READ_NUM_T         (cap_node, pjmedia_rtcp_fb_type, cap.type);
         NODE_READ_STRING        (cap_node, cap.typeName);
         NODE_READ_STRING        (cap_node, cap.param);
-        this->caps.push_back(cap);
+        this->caps.push_back(std::move(cap));
```
wosrediinanatour, line 99: General for all occurency:

std::move is a code smell. You still have two objects.

With C++11 you can use:
```
....emplace_back([&](){
   RtcpFbCap cap;
   // initialize
   return cap;
});
```
Then it the object directly, thanks to RVO.

Even better would be deligated identifier. But most probably not for the minimum wanted c++ version.

nanangizz:
`std::move()` seems to be part of C++11 (released 2011)? As we still need to support older C++ version (vs2005 & vs2008), perhaps wrap it in a macro?

sauwming:
Didn't expect such an objection against `std::move` :)
But yeah, using `std::move` doesn't seem like a good idea, although the alternatives don't seem appealing as well (either unnecessarily complex or requires newer compiler).

Not sure if using a macro can effectively suppress the warning?
Alternatively, we can simply ignore the warning since the potential optimisation is negligible. In fact, I read that some modern compilers will already perform the optimisation anyway even without the `std::move`.

wosrediinanatour:
:-) Better std::move than nothing ;-) Making a PJ_MOVE macro for instance is better than nothing.

LeonidGoltsblat:
Please note: to support the correct __cplusplus value in Visual Studio, you need to set the /Zc:__cplusplus compiler option.
The /Zc:__cplusplus option is available since Visual Studio 2017 15.7 and is disabled by default.
 (by default __cplusplus ==199711L)

as a workarround:
#if (defined(_MSVC_LANG) && _MSVC_LANG >= 201103L) || __cplusplus >= 201103L
etc

nanangizz (approved):

PR #4250, opened 2025-01-15
Fixed Coverity warnings

sauwming:
sip_parser.c: hdr->expires cannot be greater than `PJSIP_MAX_EXPIRES`
pjsua_media.c: rem_sdp can be NULL

### DIFF
--- a/pjsip/src/pjsip/sip_parser.c
+++ b/pjsip/src/pjsip/sip_parser.c
@@ -1916,8 +1916,8 @@ static void int_parse_contact_param( pjsip_contact_hdr *hdr,
             hdr->expires = pj_strtoul(&pvalue);
             if (hdr->expires == PJSIP_EXPIRES_NOT_SPECIFIED)
                 hdr->expires--;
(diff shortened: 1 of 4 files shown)


pjsip/src/pjsip/sip_parser.c
```diff
@@ -1916,8 +1916,8 @@ static void int_parse_contact_param( pjsip_contact_hdr *hdr,
             hdr->expires = pj_strtoul(&pvalue);
             if (hdr->expires == PJSIP_EXPIRES_NOT_SPECIFIED)
                 hdr->expires--;
-            if (hdr->expires > PJSIP_MAX_EXPIRES)
-                hdr->expires = PJSIP_MAX_EXPIRES;
+            //if (hdr->expires > PJSIP_MAX_EXPIRES)
+            //    hdr->expires = PJSIP_MAX_EXPIRES;
```
nanangizz, line 1920: Perhaps also remove `MAX_EXPIRES` constant as it seems to be only used here? The RFC does not seem to specify the maximum expire time either.

nanangizz (approved):

PR #4252, opened 2025-01-17
Add option whether to add port in on_stream_created2() callback

sauwming:
Currently, pjsua will always add call media port to its conference bridge. But there may be situations when users may not want to do this, for example if the conference bridge is currently not connected to any clock source.

If needed, app can store the call media port specified in `on_stream_created2()` callback and add it to the conference at a later timing.


sauwming:
Issue solved with a different approach in #4253.

PR #4254, opened 2025-01-17
Update SSL backends and unit test

sauwming:
Update SSL backends to use TLS 1.2 and 1.3 by default as older TLS versions have been deprecated. Unit tests have been updated with newer versions of TLS and ciphers.

Tested and pass with OpenSSL and GnuTLS backend.

Also in this PR:
- Fixed bug in `https_client_test()` that causes `PJ_EINVAL` error during connection. Also, activate the test since CI machine should have Internet access to install the required dependencies.
- Also fixed OpenSSL backend failure to set cipher for TLS 1.3, which uses a different API to set the ciphers (https://docs.openssl.org/1.1.1/man3/SSL_CTX_set_cipher_list/#synopsis).

Apple SSL will be updated in a separate PR since it needs more comprehensive changes.


sauwming:
@nanangizz , please help check the Windows SChannel backend.

nanangizz:
From [here](https://learn.microsoft.com/en-us/windows/win32/secauthn/protocols-in-tls-ssl--schannel-ssp-), Windows Schannel seems to already disable TLS 1.1 or older, so when `PJ_SSL_SOCK_PROTO_DEFAULT` is set, it should use TLS 1.2 & 1.3.

Btw, the [`PJ_SSL_SOCK_PROTO_DEFAULT`](https://docs.pjsip.org/en/latest/api/generated/pjlib/group/group__PJ__SSL__SOCK.html#_CPPv4N17pj_ssl_sock_proto25PJ_SSL_SOCK_PROTO_DEFAULTE) docs only says:

> Default protocol of backend. 

Perhaps also add something about TLS 1.1/older deprecation in the docs?

nanangizz (approved):

PR #4257, opened 2025-01-21
Various fixes for Apple SSL backend

sauwming:
There are several problem found in Apple SSL backend:
- Occasionally, the connection can fail to be closed, as indicated by the log `Warning: Failed to cancel SSL connection`. This may cause access to its already destroyed parent SSL socket. Switching from serial to concurrent queue seems to fix the issue and the connection cancellation seems to never fail now.
- SSL socket may not report disconnection immediately. The patch now adds EOF check in `nw_connection_receive()`.
- Fix issue that cause on_handshake_complete() to be called even if we are not in a state of handshaking.

Server socket:
- Fix issue in server's socket to incorrectly discard events.
- Fix issue that causes client cert to always be required (instead of based on `ssock->param.require_client_cert`).

SSL test:
- Fix memory leak for sockets that's scheduled to be closed after `PJ_SSL_SOCK_DELAYED_CLOSE_TIMEOUT`.

### DIFF
--- a/.github/workflows/ci-mac.yml
+++ b/.github/workflows/ci-mac.yml
@@ -144,15 +144,17 @@ jobs:
     - name: pjsip-test
       run: make pjsip-test
 
-  gnu-tls-1:
+  apple-ssl-1:
     runs-on: macos-latest
-    name: GnuTLS / pjlib,util,pjmedia,pjnath
+    name: AppleSSL / pjlib,util,pjmedia,pjnath
     steps:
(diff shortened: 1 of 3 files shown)


nanangizz (approved):

sauwming:
This has passed the CI tests. The problem is that so far Apple SSL has not been integrated into CI test because it requires user prompt to access the Keychain.
The previous commit is to test whether we can workaround the prompt. Should it fail, it will be reverted.

PR #4261, opened 2025-01-22
Add the option to check the source address of ClientHello message on DTLS-SRTP

trengginas:
There is a possible issue with `DTLS` “ClientHello” Race Conditions leading to DDoS attack.
This is done by sending a malicious DTLS ClientHello message from any
IP address to the expected port, potentially causing a “network race condition” if the malicious
messages is processed before the legitimate one.
For ICE use, this patch will add `pj_ice_sess_options::check_src_addr` and `PJ_ICE_SESS_CHECK_SRC_ADDR` to allow ICE session
to check the source against the 'verified' remote candidate. i.e.: remote candidates which has 
a completed connectivity check or received a connectivity check.
For non-ICE use, this patch will add the option (`PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR`) to enable checking the source address of the "ClientHello" message is coming from the source specified in the SDP.

### DIFF
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -1100,6 +1100,16 @@
 #   define PJMEDIA_SRTP_DTLS_OSSL_CIPHERS           "DEFAULT"
 #endif
 
+ /**
+  * Enabled this to check the source address of ClientHello message coming
+  * from a valid address.
+  *
+  * Default value: 0
+  */
+#ifndef PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR
+#   define PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR       0
+#endif
+
 
 /**
  * Maximum number of SRTP cryptos.
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -1100,13 +1100,12 @@
 #   define PJMEDIA_SRTP_DTLS_OSSL_CIPHERS           "DEFAULT"
 #endif
 
- /**
-  * Enabled this to check the source address of ClientHello message coming
-  * from a valid address. When ICE is enabled, the check will always be
-  * performed.
-  *
-  * Default value: 0
-  */
+/**
+ * Enabled this to check the source address of ClientHello message coming
+ * from a valid address. See PJ_ICE_SESS_CHECK_SRC_ADDR when ICE is used.
+ *
+ * Default value: 0
+ */
 #ifndef PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR
 #   define PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR       0
 #endif
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -1102,7 +1102,8 @@
 
  /**
   * Enabled this to check the source address of ClientHello message coming
-  * from a valid address.
+  * from a valid address. When ICE is enabled, the check will always be
+  * performed.
   *
   * Default value: 0
   */
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -1365,43 +1365,51 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
 #if defined(PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR) && PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR!=0
         pjmedia_transport_info info;
         pj_sockaddr src_addr, rem_addr;
+        pj_bool_t src_addr_avail = PJ_TRUE;
         pj_bool_t rem_addr_avail = PJ_TRUE;
+
         pjmedia_transport_get_info(ds->srtp->member_tp, &info);
         if (idx == RTP_CHANNEL) {
-            pj_sockaddr_cp(&src_addr, &info.src_rtp_name);
             if (!pj_sockaddr_has_addr(&ds->rem_addr)) {
                 rem_addr_avail = PJ_FALSE;
             } else {
                 pj_sockaddr_cp(&rem_addr, &ds->rem_addr);
             }
+            if (!pj_sockaddr_has_addr(&info.src_rtp_name)) {
+                src_addr_avail = PJ_FALSE;
+            } else {
+                pj_sockaddr_cp(&src_addr, &info.src_rtp_name);
+            }
         } else {
-            pj_sockaddr_cp(&src_addr, &info.src_rtcp_name);
             if (!pj_sockaddr_has_addr(&ds->rem_rtcp)) {
                 rem_addr_avail = PJ_FALSE;
             } else {
                 pj_sockaddr_cp(&rem_addr, &ds->rem_rtcp);
             }
-        }
-        if (!rem_addr_avail) {
-            char addr[PJ_INET6_ADDRSTRLEN];
-
-            PJ_LOG(2, (ds->base.name, "DTLS-SRTP %s ignoring %lu bytes "
-                "from [%s], remote address not available",
-                CHANNEL_TO_STRING(idx), (unsigned long)size,
-                pj_sockaddr_print(&src_addr, addr, sizeof(addr), 3)));
-
-            DTLS_UNLOCK(ds);
-            return PJ_SUCCESS;
+            if (!pj_sockaddr_has_addr(&info.src_rtcp_name)) {
+                src_addr_avail = PJ_FALSE;
+            } else {
+                pj_sockaddr_cp(&src_addr, &info.src_rtcp_name);
+            }
         }
 
-        if (pj_sockaddr_cmp(&src_addr, &rem_addr) != 0) {
-            char addr[PJ_INET6_ADDRSTRLEN], raddr[PJ_INET6_ADDRSTRLEN];
-            
-            PJ_LOG(2, (ds->base.name, "DTLS-SRTP %s ignoring message from [%s],"
-                "expecting from [%s]", 
-                CHANNEL_TO_STRING(idx), 
-                pj_sockaddr_print(&src_addr, addr, sizeof(addr), 3),
-                pj_sockaddr_print(&rem_addr, raddr, sizeof(raddr), 3)));
+        if (!src_addr_avail || !rem_addr_avail || 
+            pj_sockaddr_cmp(&src_addr, &rem_addr) != 0) 
+        {
+            char psrc_addr[PJ_INET6_ADDRSTRLEN] = "Unknown";
+            char prem_addr[PJ_INET6_ADDRSTRLEN] = "Unknown";
+
+            if (src_addr_avail) {
+                pj_sockaddr_print(&src_addr, psrc_addr, sizeof(psrc_addr), 3);
+            }
+            if (rem_addr_avail) {
+                pj_sockaddr_print(&rem_addr, prem_addr, sizeof(prem_addr), 3);
+            }
+
+            PJ_LOG(2, (ds->base.name, "DTLS-SRTP %s ignoring %lu bytes, "
+                "src addr [%s], expecting from [%s]",
+                CHANNEL_TO_STRING(idx), (unsigned long)size, 
+                psrc_addr, prem_addr));
 
             DTLS_UNLOCK(ds);
             return PJ_SUCCESS;
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -1361,6 +1361,52 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
         (ds->setup == DTLS_SETUP_ACTPASS || ds->setup == DTLS_SETUP_PASSIVE))
     {
         pj_status_t status;
+
+#if defined(PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR) && PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR!=0
+        pjmedia_transport_info info;
+        pj_sockaddr src_addr, rem_addr;
+        pj_bool_t rem_addr_avail = PJ_TRUE;
+        pjmedia_transport_get_info(ds->srtp->member_tp, &info);
+        if (idx == RTP_CHANNEL) {
+            pj_sockaddr_cp(&src_addr, &info.src_rtp_name);
+            if (!pj_sockaddr_has_addr(&ds->rem_addr)) {
+                rem_addr_avail = PJ_FALSE;
+            } else {
+                pj_sockaddr_cp(&rem_addr, &ds->rem_addr);
+            }
+        } else {
+            pj_sockaddr_cp(&src_addr, &info.src_rtcp_name);
+            if (!pj_sockaddr_has_addr(&ds->rem_rtcp)) {
+                rem_addr_avail = PJ_FALSE;
+            } else {
+                pj_sockaddr_cp(&rem_addr, &ds->rem_rtcp);
+            }
+        }
+        if (!rem_addr_avail) {
+            char addr[PJ_INET6_ADDRSTRLEN];
+
+            PJ_LOG(2, (ds->base.name, "DTLS-SRTP %s ignoring %lu bytes "
+                "from [%s], remote address not available",
+                CHANNEL_TO_STRING(idx), (unsigned long)size,
+                pj_sockaddr_print(&src_addr, addr, sizeof(addr), 3)));
+
+            DTLS_UNLOCK(ds);
+            return PJ_EIGNORED;
+        }
+
+        if (pj_sockaddr_cmp(&src_addr, &rem_addr) != 0) {
+            char addr[PJ_INET6_ADDRSTRLEN], raddr[PJ_INET6_ADDRSTRLEN];
+            
+            PJ_LOG(2, (ds->base.name, "DTLS-SRTP %s ignoring message from [%s],"
+                "expecting from [%s]", 
+                CHANNEL_TO_STRING(idx), 
+                pj_sockaddr_print(&src_addr, addr, sizeof(addr), 3),
+                pj_sockaddr_print(&rem_addr, raddr, sizeof(raddr), 3)));
+
+            DTLS_UNLOCK(ds);
+            return PJ_EIGNORED;
+        }
+#endif
         ds->setup = DTLS_SETUP_PASSIVE;
         status = ssl_handshake_channel(ds, idx);
         if (status != PJ_SUCCESS) {
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -1391,7 +1391,7 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
                 pj_sockaddr_print(&src_addr, addr, sizeof(addr), 3)));
 
             DTLS_UNLOCK(ds);
-            return PJ_EIGNORED;
+            return PJ_SUCCESS;
         }
 
         if (pj_sockaddr_cmp(&src_addr, &rem_addr) != 0) {
@@ -1404,7 +1404,7 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
                 pj_sockaddr_print(&rem_addr, raddr, sizeof(raddr), 3)));
 
             DTLS_UNLOCK(ds);
-            return PJ_EIGNORED;
+            return PJ_SUCCESS;
         }
 #endif
         ds->setup = DTLS_SETUP_PASSIVE;
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -1374,16 +1374,15 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
         (ds->setup == DTLS_SETUP_ACTPASS || ds->setup == DTLS_SETUP_PASSIVE))
     {
         pj_status_t status;
-        pj_bool_t check_hello_addr = PJ_FALSE;
 
 #if defined(PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR) && \
             PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR==1
 
-        if (!ds->use_ice)
-            check_hello_addr = PJ_TRUE;
-
-#endif
-        if (check_hello_addr) {
+        /* Check the souce address with the specified remote address from 
+         * the SDP. When ICE is used, the source address checking will be 
+         * done in ICE session.
+         */
+        if (!ds->use_ice) {
             pjmedia_transport_info info;
             pj_sockaddr src_addr;
             pj_bool_t src_addr_avail = PJ_TRUE;
@@ -1407,17 +1406,18 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
                 char psrc_addr[PJ_INET6_ADDRSTRLEN] = "Unknown";
 
                 if (src_addr_avail) {
-                    pj_sockaddr_print(&src_addr, psrc_addr, 
+                    pj_sockaddr_print(&src_addr, psrc_addr,
                                       sizeof(psrc_addr), 3);
                 }
                 PJ_LOG(2, (ds->base.name, "DTLS-SRTP %s ignoring %lu bytes, "
-                    "from src addr [%s]", CHANNEL_TO_STRING(idx), 
-                    (unsigned long)size,   psrc_addr));
+                    "from unrecognized src addr [%s]", CHANNEL_TO_STRING(idx),
+                    (unsigned long)size, psrc_addr));
 
                 DTLS_UNLOCK(ds);
                 return PJ_SUCCESS;
             }
         }
+#endif
         ds->setup = DTLS_SETUP_PASSIVE;
         status = ssl_handshake_channel(ds, idx);
         if (status != PJ_SUCCESS) {
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -117,12 +117,6 @@ typedef struct dtls_srtp_channel
     unsigned             channel;
 } dtls_srtp_channel;
 
-typedef struct dtls_srtp_ice_cand
-{
-    unsigned            cand_cnt;
-    pj_sockaddr         cand_addr[PJ_ICE_MAX_CAND];
-} dtls_srtp_ice_cand;
-
 typedef struct dtls_srtp
 {
     pjmedia_transport    base;
@@ -149,8 +143,6 @@ typedef struct dtls_srtp
 
     char                 buf[NUM_CHANNEL][PJMEDIA_MAX_MTU];
     pjmedia_clock       *clock[NUM_CHANNEL];/* Timer workaround for retrans */
-
-    dtls_srtp_ice_cand   ice_rem_cand[NUM_CHANNEL];
 
     SSL_CTX             *ossl_ctx[NUM_CHANNEL];
     SSL                 *ossl_ssl[NUM_CHANNEL];
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -1090,152 +1090,6 @@ static pj_status_t parse_setup_finger_attr(dtls_srtp *ds,
     return PJ_SUCCESS;
 }
 
-/* Parse a=candidate line */
-static pj_status_t parse_cand(dtls_srtp *ds, 
-                              const pj_str_t *orig_input,
-                              dtls_srtp_ice_cand cand[])
-{
-    pj_str_t token, delim, host;
-    int af;
-    unsigned comp_id;
-    unsigned cand_cnt;
-    pj_ssize_t found_idx;
-    pj_status_t status = PJNATH_EICEINCANDSDP;
-
-    /* Foundation */
-    delim = pj_str(" ");
-    found_idx = pj_strtok(orig_input, &delim, &token, 0);
-    if (found_idx == orig_input->slen) {
-#if DTLS_DEBUG
-        PJ_LOG(2, (ds->base.name, "Parse ICE cand: "
-            "Expecting ICE foundation in candidate"));
-#endif
-        goto on_return;
-    }
-
-    /* Component ID */
-    found_idx = pj_strtok(orig_input, &delim, &token, found_idx + token.slen);
-    if (found_idx == orig_input->slen) {
-#if DTLS_DEBUG
-        PJ_LOG(2, (ds->base.name, "Parse ICE cand: "
-            "Expecting ICE component ID in candidate"));
-#endif
-        goto on_return;
-    }
-    comp_id = (pj_uint8_t)pj_strtoul(&token) - 1;
-
-    if (comp_id > NUM_CHANNEL) {
-#if DTLS_DEBUG
-        PJ_LOG(2, (ds->base.name, "Parse ICE cand: "
-            "Invalid ICE component ID in candidate"));
-#endif
-        goto on_return;
-    }
-    cand_cnt = cand[comp_id].cand_cnt;
-
-    /* Transport */
-    found_idx = pj_strtok(orig_input, &delim, &token, found_idx + token.slen);
-    if (found_idx == orig_input->slen) {
-#if DTLS_DEBUG
-        PJ_LOG(2, (ds->base.name, "Parse ICE cand: "
-            "Expecting ICE transport in candidate"));
-#endif
-        goto on_return;
-    }
-    if (pj_stricmp2(&token, "UDP") != 0) {
-#if DTLS_DEBUG
-        PJ_LOG(2, (ds->base.name, "Parse ICE cand: "
-            "Expecting ICE UDP transport only in candidate"));
-#endif
-        goto on_return;
-    }
-
-    /* Priority */
-    found_idx = pj_strtok(orig_input, &delim, &token, found_idx + token.slen);
-    if (found_idx == orig_input->slen) {
-#if DTLS_DEBUG
-        PJ_LOG(2, (ds->base.name, "Parse ICE cand: "
-            "Expecting ICE priority in candidate"));
-#endif
-        goto on_return;
-    }
-
-    /* Host */
-    found_idx = pj_strtok(orig_input, &delim, &host, found_idx + token.slen);
-    if (found_idx == orig_input->slen) {
-#if DTLS_DEBUG
-        PJ_LOG(2, (ds->base.name, "Parse ICE cand: "
-            "Expecting ICE priority in candidate"));
-#endif
-        goto on_return;
-    }
-    /* Detect address family */
-    if (pj_strchr(&host, ':'))
-        af = pj_AF_INET6();
-    else
-        af = pj_AF_INET();
-    /* Assign address */
-    if (pj_sockaddr_init(af, &cand[comp_id].cand_addr[cand_cnt], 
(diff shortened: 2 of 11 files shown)


pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1361,6 +1361,52 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
...
+        if (!rem_addr_avail) {
+            char addr[PJ_INET6_ADDRSTRLEN];
+
+            PJ_LOG(2, (ds->base.name, "DTLS-SRTP %s ignoring %lu bytes "
+                "from [%s], remote address not available",
+                CHANNEL_TO_STRING(idx), (unsigned long)size,
+                pj_sockaddr_print(&src_addr, addr, sizeof(addr), 3)));
+
+            DTLS_UNLOCK(ds);
+            return PJ_EIGNORED;
```
nanangizz, line 1394: Perhaps this should be PJ_SUCCESS instead? As PJ_EIGNORED means leaving the packet to be processed by the parent transport (SRTP).

nanangizz (approved):
Please make sure this is tested with various scenarios, e.g:
- with & without ICE,
- ICE chooses TURN & STUN/host as valid candidate.

pjmedia/include/pjmedia/config.h
```diff
@@ -1100,6 +1100,16 @@
 #   define PJMEDIA_SRTP_DTLS_OSSL_CIPHERS           "DEFAULT"
 #endif
 
+ /**
+  * Enabled this to check the source address of ClientHello message coming
+  * from a valid address.
+  *
+  * Default value: 0
+  */
+#ifndef PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR
```
sauwming, line 1109: Why do we need a compile-time setting? and why is the default 0 if it may cause a security issue?

trengginas:
> Please make sure this is tested with various scenarios, e.g:
> 
> * with & without ICE,
> * ICE chooses TURN & STUN/host as valid candidate.

When using this option, it is recommended to use regular nomination for ICE. When enabling STUN with aggressive nomination might result in a different candidate pair type being used between endpoints and preventing the handshake.

pjmedia/include/pjmedia/config.h
trengginas: It might add a slight delay in starting the handshake (wait for the transport address and RTP source address to be available) and possible issues when using aggressive nomination.

sauwming: After reading the pdf, I believe the address check should be made mandatory for ICE, and the address should match one of the verified (valid?) candidates obtained during ICE check, so there should be no delay because ICE check has been done.

For no-ICE, then you can use the compile-time setting, because the best we can do is compare it with the SDP address, which may not be accurate.

trengginas:
This is the case when using ICE with STUN (aggressive nomination):
caller side:
===
```
14:13:01.139        icetp00  ICE negotiation success after 0s:055
14:13:01.139        icetp00   Comp 1: sending from srflx candidate 117.90.81.147:12322 to srflx candidate 117.90.81.147:12325
14:13:01.141        icetp00   Comp 2: sending from host candidate 192.168.100.51:9938 to host candidate 192.168.100.51:4003
...
14:13:02.177   dtls0326C288 !DTLS-SRTP RTCP channel ignoring 222 bytes, src addr [117.90.81.147:12327], expecting from [192.168.100.51:4003]
14:13:02.201   dtls0326C288  DTLS-SRTP negotiation for RTP channel completed!
```
===

callee side:
===
```
14:13:01.142        icetp00  ICE negotiation success after 0s:069
14:13:01.143        icetp00   Comp 1: sending from srflx candidate 117.90.81.147:12325 to srflx candidate 117.90.81.147:12322
14:13:01.143        icetp00   Comp 2: sending from srflx candidate 117.90.81.147:12327 to srflx candidate 117.90.81.147:12324
```
===

nanangizz:
For trickle ICE, the candidates may be added rapidly while app is allowed to send/recv data before the ICE nego is completed. So if the proposed feature is on, perhaps it should indeed check the source address against all remote candidates. This may be applicable for aggressive mode too.

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1082,6 +1090,152 @@ static pj_status_t parse_setup_finger_attr(dtls_srtp *ds,
     return PJ_SUCCESS;
 }
 
+/* Parse a=candidate line */
+static pj_status_t parse_cand(dtls_srtp *ds,
```
nanangizz, line 1094: I don't think it is a good idea to parse ICE SDP attr by DTLS-SRTP, also it may introduce risk of maintaining the code.
IMO it is cleaner to query the valid candidates from ICE transport, it looks like ICE transport info already provides it? If not, adding API to ICE tp (or field in ICE tp info) is perhaps the proper approach.

nanangizz:
So now, does it work for aggressive & trickle modes?

trengginas:
> So now, does it work for aggressive & trickle modes?

Yes, now working with aggressive and trickle

sauwming:
For ICE:
It seems that `enum_cands()` will enumerate all candidates regardless of the candidate status (even the failed ones?). The recommendation is to compare it with **verified** candidates (i.e. the candidates that have passed the media consent verification phase).
So there are two options:
a. modify the API to return only verified candidates: `enum_verified_cands()`, or
b. let ICE handle the verification: `pjmedia_ice_verify_remote_cand()`

If no ICE is used:
The recommendation is to compare it with the SDP address.

nanangizz:
> For ICE: It seems that `enum_cands()` will enumerate all candidates regardless of the candidate status (even the failed ones?). The recommendation is to compare it with **verified** candidates (i.e. the candidates that have passed the media consent verification phase). So there are two options: a. modify the API to return only verified candidates: `enum_verified_cands()`, or b. let ICE handle the verification: `pjmedia_ice_verify_remote_cand()`

It looks like "verified" is specific to the recommendation, not really ICE things?
If it does, adding pair-check-status-info for each candidate should be more natural than adding "verified" APIs to ICE.
Then it is app/DTLS responsibility to "translate" pair-check-status to "verified".

sauwming:
> It looks like "verified" is specific to the recommendation, not really ICE things?

Correct. ICE itself doesn't use the term "verified", and the document explains it in the media consent verification phase, which means that that the other party is the peer they claim to be. So this way, security can be achieved.

> If it does, adding pair-check-status-info for each candidate should be more natural than adding "verified" APIs to ICE. Then it is app/DTLS responsibility to "translate" pair-check-status to "verified".

Yes, this is fine too. So, it means that `enum_cands()` will also return the status of the candidates, right?

nanangizz:
> Yes, this is fine too. So, it means that `enum_cands()` will also return the status of the candidates, right?

Yes. Note that in ICE, the status belongs to pair-check instead of candidate, a remote candidate may be paired (& checked) against multiple local candidates. If DTLS just needs to list remote candidates that has been checked, it can be done but needs to be clarified in the docs.

trengginas:
There's an issue when using aggressive nomination/trickle:
The controlled and controlling side might use different candidates.
Controlling side:
```
15:15:55.315  stuse03EDF9E8  .RX 120 bytes STUN message from 192.168.100.51:4008:
--- begin STUN message ---
STUN Binding request
 Hdr: length=100, magic=2112a442, tsx_id=00002c605772139d26ca369a
 Attributes:
  ICE-CONTROLLED: length=8, data=2e4013661cd0366b
--- end of STUN message ---
15:15:55.315  stuse03EDF9E8  ..TX 88 bytes STUN message to 192.168.100.51:4008:
--- begin STUN message ---
STUN Binding success response
 Hdr: length=68, magic=2112a442, tsx_id=00002c605772139d26ca369a
--- end of STUN message ---

15:15:55.323        icetp00  Starting checklist periodic check
15:15:55.323        icetp00  .Sending connectivity check for check 1: [2] 192.168.100.51:9933-->100.100.100.100:21760
15:15:55.323  stuse03EE0F50  ...TX 124 bytes STUN message to 100.100.100.100:21760:
--- begin STUN message ---
STUN Binding request
 Hdr: length=104, magic=2112a442, tsx_id=00002d906e5d1ad445091239
 Attributes:
  USE-CANDIDATE: length=0
  ICE-CONTROLLING: length=8, data=0124305e440d491c
--- end of STUN message ---
15:15:55.323   utsx03F09E00  ...STUN client transaction created
15:15:55.323   utsx03F09E00  ...STUN sending message (transmit count=1)
15:15:55.324        icetp00  ..Check 1: [2] 192.168.100.51:9933-->100.100.100.100:21760: state changed from Waiting to In Progress
15:15:55.340  stuse03EE0F50 !.RX 120 bytes STUN message from 100.100.100.100:21760:
--- begin STUN message ---
STUN Binding request
 Hdr: length=100, magic=2112a442, tsx_id=00002c607049692c26ca369b
 Attributes:
  ICE-CONTROLLED: length=8, data=2e4013661cd0366b
--- end of STUN message ---
15:15:55.340  stuse03EE0F50  ..TX 88 bytes STUN message to 100.100.100.100:21760:
--- begin STUN message ---
STUN Binding success response
 Hdr: length=68, magic=2112a442, tsx_id=00002c607049692c26ca369b
 Attributes:
--- end of STUN message ---
15:15:55.341        icetp00  .Triggered check for check 1 not performed because it's in progress. Retransmitting
15:15:55.341   utsx03F09E00  ..STUN sending message (transmit count=1)
15:15:55.345        icetp00  Starting checklist periodic check
15:15:55.345        icetp00  .Sending connectivity check for check 3: [2] 192.168.100.51:9933-->192.168.100.51:4003
15:15:55.345  stuse03EE0F50  ...TX 124 bytes STUN message to 192.168.100.51:4003:
--- begin STUN message ---
STUN Binding request
 Hdr: length=104, magic=2112a442, tsx_id=00002d9063cb6bfc4509123a
 Attributes:
  USE-CANDIDATE: length=0
  ICE-CONTROLLING: length=8, data=0124305e440d491c
--- end of STUN message ---
15:15:55.345   utsx03F08CE0  ...STUN client transaction created
15:15:55.345   utsx03F08CE0  ...STUN sending message (transmit count=1)
15:15:55.346        icetp00  ..Check 3: [2] 192.168.100.51:9933-->192.168.100.51:4003: state changed from Frozen to In Progress
15:15:55.346  stuse03EE0F50  .RX 88 bytes STUN message from 192.168.100.51:4003:
--- begin STUN message ---
STUN Binding success response
 Hdr: length=68, magic=2112a442, tsx_id=00002d9063cb6bfc4509123a
 Attributes:
--- end of STUN message ---
15:15:55.346        icetp00  .Check 3: [2] 192.168.100.51:9933-->192.168.100.51:4003 (nominated): connectivity check SUCCESS
15:15:55.346        icetp00  .Check 3: [2] 192.168.100.51:9933-->192.168.100.51:4003: state changed from In Progress to Succeeded
15:15:55.347        icetp00  .Check 3 is successful and nominated
15:15:55.347        icetp00  .Cancelling check 1: [2] 192.168.100.51:9933-->100.100.100.100:21760 (In Progress)
15:15:55.347 stun_session.c  .tdata 03F09CDC destroy request, force=0, tsx=03F09E00, destroying=0
15:15:55.347   utsx03F09E00  .STUN transaction 03F09E00 schedule destroy
15:15:55.347        icetp00  .Check 1: [2] 192.168.100.51:9933-->100.100.100.100:21760: state changed from In Progress to Failed
15:15:55.348        icetp00  .ICE process complete, status=Success
15:15:55.348        icetp00  .Valid list
15:15:55.348        icetp00  . 0: [1] 100.100.100.100:21818-->100.100.100.100:21823 (nominated, state=Succeeded)
15:15:55.348        icetp00  . 1: [2] 192.168.100.51:9933-->192.168.100.51:4003 (nominated, state=Succeeded)
```

Controlled side:
```
15:15:55.313        icetp00  .Sending connectivity check for check 2: [1] 192.168.100.51:4008-->192.168.100.51:9937
15:15:55.313  stuse035B3770  ...TX 120 bytes STUN message to 192.168.100.51:9937:
--- begin STUN message ---
STUN Binding request
 Hdr: length=100, magic=2112a442, tsx_id=00002c605772139d26ca369a
 Attributes:
  ICE-CONTROLLED: length=8, data=2e4013661cd0366b
--- end of STUN message ---

15:15:55.313   utsx035A0A4C  ...STUN client transaction created
15:15:55.313   utsx035A0A4C  ...STUN sending message (transmit count=1)
15:15:55.313        icetp00  ..Check 2: [1] 192.168.100.51:4008-->192.168.100.51:9937: state changed from Waiting to In Progress
15:15:55.315  stuse035B3770  .RX 88 bytes STUN message from 192.168.100.51:9937:
--- begin STUN message ---
STUN Binding success response
 Hdr: length=68, magic=2112a442, tsx_id=00002c605772139d26ca369a
--- end of STUN message ---

15:15:55.315        icetp00  .Check 2: [1] 192.168.100.51:4008-->192.168.100.51:9937 (not nominated): connectivity check SUCCESS
15:15:55.315        icetp00  .Check 2: [1] 192.168.100.51:4008-->192.168.100.51:9937: state changed from In Progress to Succeeded
15:15:55.315        icetp00  .Check 3: [2] 192.168.100.51:4003-->192.168.100.51:9933: state changed from Frozen to Waiting
15:15:55.316        icetp00  .Check 2 is successful
15:15:55.316 stun_session.c  .tdata 035A092C destroy request, force=0, tsx=035A0A4C, destroying=0
15:15:55.316   utsx035A0A4C  .STUN transaction 035A0A4C schedule destroy
15:15:55.317  stuse035B3770  .RX 88 bytes STUN message from 100.100.100.100:21818:
--- begin STUN message ---
STUN Binding success response
 Hdr: length=68, magic=2112a442, tsx_id=00002c6009027bb926ca3699
--- end of STUN message ---

15:15:55.317        icetp00  .Check 0: [1] 192.168.100.51:4008-->100.100.100.100:21818 (nominated): connectivity check SUCCESS
15:15:55.317        icetp00  .Check 0: [1] 192.168.100.51:4008-->100.100.100.100:21818: state changed from In Progress to Succeeded
15:15:55.318        icetp00  .Check 1: [2] 192.168.100.51:4003-->100.100.100.100:21821: state changed from Frozen to Waiting
15:15:55.318        icetp00  .Check 0 is successful and nominated
15:15:55.318 stun_session.c  .tdata 035A3844 destroy request, force=0, tsx=035A3964, destroying=0
15:15:55.318   utsx035A3964  .STUN transaction 035A3964 schedule destroy
15:15:55.335        icetp00 !Starting checklist periodic check
15:15:55.335        icetp00  .Sending connectivity check for check 1: [2] 192.168.100.51:4003-->100.100.100.100:21821
15:15:55.336  stuse035B5568  ...TX 120 bytes STUN message to 100.100.100.100:21821:
--- begin STUN message ---
STUN Binding request
 Hdr: length=100, magic=2112a442, tsx_id=00002c607049692c26ca369b
 Attributes:
  ICE-CONTROLLED: length=8, data=2e4013661cd0366b
--- end of STUN message ---

15:15:55.336   utsx035A4ECC  ...STUN client transaction created
15:15:55.336   utsx035A4ECC  ...STUN sending message (transmit count=1)
15:15:55.336        icetp00  ..Check 1: [2] 192.168.100.51:4003-->100.100.100.100:21821: state changed from Waiting to In Progress
15:15:55.346  stuse035B5568  .RX 124 bytes STUN message from 192.168.100.51:9933:
--- begin STUN message ---
STUN Binding request
 Hdr: length=104, magic=2112a442, tsx_id=00002d9063cb6bfc4509123a
 Attributes:
  USE-CANDIDATE: length=0
  ICE-CONTROLLING: length=8, data=0124305e440d491c
--- end of STUN message ---

15:15:55.346  stuse035B5568  ..TX 88 bytes STUN message to 192.168.100.51:9933:
--- begin STUN message ---
STUN Binding success response
 Hdr: length=68, magic=2112a442, tsx_id=00002d9063cb6bfc4509123a
--- end of STUN message ---

15:15:55.346        icetp00  .Performing triggered check for check 3
15:15:55.346        icetp00  ..Sending connectivity check for check 3: [2] 192.168.100.51:4003-->192.168.100.51:9933
15:15:55.347  stuse035B5568  ....TX 120 bytes STUN message to 192.168.100.51:9933:
--- begin STUN message ---
STUN Binding request
 Hdr: length=100, magic=2112a442, tsx_id=00002c604a80187e26ca369c
 Attributes:
  ICE-CONTROLLED: length=8, data=2e4013661cd0366b
--- end of STUN message ---

15:15:55.347   utsx035A1724  ....STUN client transaction created
15:15:55.347   utsx035A1724  ....STUN sending message (transmit count=1)
15:15:55.347        icetp00  ...Check 3: [2] 192.168.100.51:4003-->192.168.100.51:9933: state changed from Waiting to In Progress
15:15:55.347  stuse035B5568  .RX 88 bytes STUN message from 100.100.100.100:21821:
--- begin STUN message ---
STUN Binding success response
 Hdr: length=68, magic=2112a442, tsx_id=00002c607049692c26ca369b
--- end of STUN message ---

15:15:55.347        icetp00  .Check 1: [2] 192.168.100.51:4003-->100.100.100.100:21821 (not nominated): connectivity check SUCCESS
15:15:55.347        icetp00  .Check 1: [2] 192.168.100.51:4003-->100.100.100.100:21821: state changed from In Progress to Succeeded
15:15:55.348        icetp00  .Check 1 is successful
15:15:55.348 stun_session.c  .tdata 035A4DAC destroy request, force=0, tsx=035A4ECC, destroying=0
15:15:55.348   utsx035A4ECC  .STUN transaction 035A4ECC schedule destroy
15:15:55.348  stuse035B5568  .RX 124 bytes STUN message from 100.100.100.100:21821:
--- begin STUN message ---
STUN Binding request
 Hdr: length=104, magic=2112a442, tsx_id=00002d906e5d1ad445091239
 Attributes:
  USE-CANDIDATE: length=0
  ICE-CONTROLLING: length=8, data=0124305e440d491c
--- end of STUN message ---

15:15:55.348  stuse035B5568  ..TX 88 bytes STUN message to 100.100.100.100:21821:
--- begin STUN message ---
STUN Binding success response
 Hdr: length=68, magic=2112a442, tsx_id=00002d906e5d1ad445091239
--- end of STUN message ---

15:15:55.348        icetp00  .Valid check 1: [2] 100.100.100.100:21760-->100.100.100.100:21821 is nominated
15:15:55.348        icetp00  .Triggered check for check 1 not performed because it's completed
15:15:55.349        icetp00  ..Check 1 is successful and nominated
15:15:55.349        icetp00  ..Cancelling check 3: [2] 192.168.100.51:4003-->192.168.100.51:9933 (In Progress)
15:15:55.349 stun_session.c  ..tdata 035A1604 destroy request, force=0, tsx=035A1724, destroying=0
15:15:55.349   utsx035A1724  ..STUN transaction 035A1724 schedule destroy
15:15:55.349        icetp00  ..Check 3: [2] 192.168.100.51:4003-->192.168.100.51:9933: state changed from In Progress to Failed
15:15:55.349        icetp00  ..ICE process complete, status=Success
15:15:55.349        icetp00  ..Valid list
15:15:55.350        icetp00  .. 0: [1] 100.100.100.100:21823-->100.100.100.100:21818 (nominated, state=Succeeded)
15:15:55.350        icetp00  .. 1: [2] 100.100.100.100:21760-->100.100.100.100:21821 (nominated, state=Succeeded)
15:15:55.350        icetp00  .. 2: [1] 192.168.100.51:4008-->192.168.100.51:9937 (not nominated, state=Succeeded)
```

Check 1 connectivity check was received late, so the controlling side will use Check 3 instead and fail the Check 1.
However on the controlled side, it uses Check 1.

sauwming:
It looks like it's caused by `PJ_ICE_CANCEL_ALL` (by default 1):
https://github.com/pjsip/pjproject/blob/f986ad87ae1f7aaecce8005802f2f0d436285c8f/pjnath/src/pjnath/ice_session.c#L1734-L1741

https://github.com/pjsip/pjproject/blob/f986ad87ae1f7aaecce8005802f2f0d436285c8f/pjnath/include/pjnath/config.h#L352-L360

I don't quite understand the reasoning why we enable this setting by default though, especially about the statement `even when it most likely will fail`.
So perhaps try disabling this setting?

trengginas:
After setting `PJ_ICE_CANCEL_ALL` 0, the check still got cancelled because it has less priority.
On the controlled side, the ICE check is already completed (received a connectivity check response) and will ignore any 
other connectivity check response. (e.g.: on the controlled side Check 3 response was received first and ICE is completed, after Check 1 response will be ignored).
If we want to mandate this feature for ICE, perhaps we can add and option to check the source address against all the candidates instead of valid candidates, to accommodate the aggressive nomination.

sauwming:
Yes, we need to check against all *verified* remote candidates. According to my understanding (let me know if this is not correct), a verified remote candidate is a remote candidate that has successfully sent an ICE check to us.

trengginas:
I assume `verified` means it has the connectivity check done successfully. If using an aggressive nomination (as above), the controlling and controlled agent might use different candidates.
The check is done on the controlling side, and it will fail the connectivity check for the candidate used by the controlled side.
The source address check will fail if using `verified` candidates.
Note: 
The document bases the ICE operation on rfc8445, which already deprecates aggressive nomination.
```
Nomination:  The process of the controlling agent indicating to the
      controlled agent which candidate pair the ICE agents will use for
      sending and receiving data.  The nomination process defined in
      this specification was referred to as "regular nomination" in [RFC](https://datatracker.ietf.org/doc/html/rfc5245)
      [5245](https://datatracker.ietf.org/doc/html/rfc5245).  The nomination process that was referred to as "aggressive
      nomination" in [RFC 5245](https://datatracker.ietf.org/doc/html/rfc5245) has been deprecated in this specification.
```

nanangizz:
From the docs:
> Media consent verification uses the ICE candidate information to verify that both WebRTC peers can
> (a) communicate with each other via a reliable path and
> (b) that the other party is the peer they claim to be. 

I think we can interpret it as:
(a) remote STUN check request has reached us (connectivity)
(b) remote uses correct user/pass in its STUN check request (authentication)

So I agree with @sauwming that _a verified remote candidate is a remote candidate that has successfully sent an ICE check to us_, with additional note, it uses the correct user & pass (I think our ICE/STUN implementation has checked the user/pass before responding, need to verify though). IIRC ICE doesn't maintain a list of them.

Furthermore, if ICE is active, what if we put the source check in ICE itself, e.g: in pj_ice_sess_on_rx_pkt(), and as usual, it is better configurable.

pjnath/src/pjnath/ice_session.c
```diff
@@ -3734,6 +3746,29 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
...
+            if (i == ice->rcand_cnt) {
+                char psrc_addr[PJ_INET6_ADDRSTRLEN] = {0};
+
+                if (pj_sockaddr_has_addr(src_addr)) {
+                    pj_sockaddr_print(src_addr, psrc_addr,
+                                      sizeof(psrc_addr), 3);
+                }
+                PJ_LOG(2, (ice->obj_name, "Ignoring incoming message for "
+                                          "component [%d], from src addr [%s]",
+                                          comp_id, psrc_addr));
```
nanangizz, line 3767: - Specify the reason of ignoring, e.g: s/"component [%d], from src addr [%s]"/"component %d because source address %s is unrecognized"? Note that `pj_sockaddr_print(..,3)` will also print bracket for IPv6.
- Won't log level 2 be too verbose for low level API, perhaps 4?

pjnath/src/pjnath/ice_session.c
```diff
@@ -3141,6 +3144,13 @@ static pj_status_t on_stun_rx_request(pj_stun_session *sess,
     uc_attr = (pj_stun_use_candidate_attr*)
               pj_stun_msg_find_attr(msg, PJ_STUN_ATTR_USE_CANDIDATE, 0);
 
+    if (uc_attr != NULL && ice->is_complete) {
+        LOG4((ice->obj_name,
+            "Ignored incoming check after ICE nego has been completed!, "
+            "no need to send response"));
+        pj_grp_lock_release(ice->grp_lock);
+        return PJ_SUCCESS;
+    }
```
nanangizz, line 3153: This part seems rather OOT and change the ICE flow/algo? Usually this kind of changes is specified (including the reason) in the PR desc.

pjnath/include/pjnath/ice_session.h
```diff
@@ -672,6 +678,15 @@ typedef struct pj_ice_sess_options
...
     pj_ice_sess_trickle trickle;
 
+    /**
+     * Specify whether to check the source address of the incoming messages.
+     * The source address will be compared to the remote candidate which has 
+     * a completed connectivity check or received a connectivity check.
+     *
+     * Default value is PJ_ICE_SESS_CHECK_SRC_ADDR.
+     */
+    unsigned check_src_addr;
```
nanangizz, line 688: why not boolean?

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -144,6 +150,8 @@ typedef struct dtls_srtp
     char                 buf[NUM_CHANNEL][PJMEDIA_MAX_MTU];
     pjmedia_clock       *clock[NUM_CHANNEL];/* Timer workaround for retrans */
 
+    dtls_srtp_ice_cand   ice_rem_cand[NUM_CHANNEL];
+
```
nanangizz, line 154: still used?

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -117,6 +117,12 @@ typedef struct dtls_srtp_channel
     unsigned             channel;
 } dtls_srtp_channel;
 
+typedef struct dtls_srtp_ice_cand
+{
+    unsigned            cand_cnt;
+    pj_sockaddr         cand_addr[PJ_ICE_MAX_CAND];
+} dtls_srtp_ice_cand;
+
```
nanangizz, line 125: still used?

pjnath/include/pjnath/ice_session.h
```diff
@@ -317,6 +317,12 @@ struct pj_ice_sess_cand
      */
     pj_sockaddr          rel_addr;
 
+    /**
+     * Indicate that remote connectivity check has been received or the check
+     * has been successful for this candidate.
+     *
+     */
+    pj_bool_t            checked;
```
nanangizz, line 325: Need to specify that this flag is only used by remote candidate?

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1361,6 +1374,50 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
         (ds->setup == DTLS_SETUP_ACTPASS || ds->setup == DTLS_SETUP_PASSIVE))
     {
         pj_status_t status;
+        pj_bool_t check_hello_addr = PJ_FALSE;
```
sauwming, line 1377: Is this still needed if the check is already done in ICE?

And if we don't use ICE, should we compare it against the SDP?
I don't mind if we don't, but we need to document it somewhere

pjnath/src/pjnath/ice_session.c
```diff
@@ -3734,6 +3746,29 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
 
         PJ_RACE_ME(5);
 
+        if (ice->opt.check_src_addr) {
+            for (i = 0; i < ice->rcand_cnt; ++i) {
```
sauwming, line 3750: Won't this slow down ICE tp considerably, if we have to iterate through all remote candidates for every packet?

nanangizz:
There were ICE errors before the last commit (perhaps due to algo change?) and it seems to still happen.

pjnath/src/pjnath/ice_session.c
```diff
@@ -3734,6 +3746,42 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
...
+            }
+            if (check_addr && 
+                (!pj_sockaddr_has_addr(raddr) || 
+                 pj_sockaddr_cmp(src_addr, raddr) != 0))
+            {
+                PJ_LOG(4, (ice->obj_name, "Ignoring incoming message for "
+                         "component [%d] because source addr [%s] unrecognized",
+                         comp_id, psrc_addr));
+                return PJ_SUCCESS;
+            }
```
nanangizz, line 3782: Once `rcand_check_addr` is set, it cannot be changed, right? Is there possibility that it may settle with a wrong source address? Note that from the code `pj_ice_sess_send_data()`, it seems to allow sending data while ICE nego is on progress (using valid check with highest prio), and the valid check may change while ICE nego is not completed.

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1361,6 +1374,50 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
...
+            pjmedia_transport_info info;
+            pj_sockaddr src_addr;
+            pj_bool_t src_addr_avail = PJ_TRUE;
+
+            pjmedia_transport_get_info(ds->srtp->member_tp, &info);
+            if (idx == RTP_CHANNEL) {
+                if (!pj_sockaddr_has_addr(&info.src_rtp_name)) {
+                    src_addr_avail = PJ_FALSE;
+                } else {
+                    pj_sockaddr_cp(&src_addr, &info.src_rtp_name);
```
sauwming, line 1395: This is RTP address, not SDP address.

trengginas: > This is RTP address, not SDP address.

Yes, it is to get the message's source address. 
The SDP address/check address is here:
https://github.com/pjsip/pjproject/blob/dcf9c6505738023ff821f38d6eb4e3695fc49d24/pjmedia/src/pjmedia/transport_srtp_dtls.c#L1181-L1191

pjmedia/src/pjmedia/transport_srtp_dtls.c
```diff
@@ -1361,6 +1374,50 @@ static pj_status_t dtls_on_recv(pjmedia_transport *tp, unsigned idx,
...
+            PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR==1
+
+        /* Check the souce address with the specified remote address from 
+         * the SDP. When ICE is used, the source address checking will be 
+         * done in ICE session.
+         */
+        if (!ds->use_ice) {
+            pjmedia_transport_info info;
+            pj_sockaddr src_addr;
+            pj_bool_t src_addr_avail = PJ_TRUE;
```
sauwming, line 1388: source address must be available, because it's the packet source.

pjnath/src/pjnath/ice_session.c
```diff
@@ -3734,6 +3738,46 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
...
+            if (!pj_sockaddr_has_addr(raddr)) {
+                for (i = 0; i < ice->rcand_cnt; ++i) {
+                    if (ice->rcand[i].comp_id == comp_id &&
+                        ice->rcand[i].checked &&
+                        pj_sockaddr_cmp(src_addr, &ice->rcand[i].addr) == 0)
+                    {
+                        /* Before ICE completed, it is still allowed to receive 
+                         * data from other candidates.
+                         */
+                        if (ice->is_complete) {
```
nanangizz, line 3759: Looks better, but adding some delay (e.g: ~1s, compile-time configurable) before fixing it to an address perhaps even better as there can be race conditions when ICE has just been completed. Note that before it is fixed, the comparison still needs to be done against all "verified" remote candidate, as already done now.

pjnath/src/pjnath/ice_session.c
```diff
@@ -3734,6 +3738,46 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
...
         PJ_RACE_ME(5);
 
+        if (ice->opt.check_src_addr) {
+            pj_bool_t check_addr = PJ_TRUE;
+            pj_sockaddr *raddr = &comp->rcand_check_addr;
+            char psrc_addr[PJ_INET6_ADDRSTRLEN] = {0};
+
+            if (pj_sockaddr_has_addr(src_addr)) {
+                pj_sockaddr_print(src_addr, psrc_addr, sizeof(psrc_addr), 3);
+            }
```
nanangizz, line 3748: Always printing address does not look very optimized, print only when needed, e.g: inside the most inner log printing block?

pjnath/src/pjnath/ice_session.c
```diff
@@ -3734,6 +3738,46 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
...
+                                       psrc_addr, comp_id));
+                        }
+                        check_addr = PJ_FALSE;
+                        break;
+                    }
+                }
+            }
+            if (check_addr && 
+                (!pj_sockaddr_has_addr(raddr) || 
+                 pj_sockaddr_cmp(src_addr, raddr) != 0))
```
nanangizz, line 3772: The check can be optimized to use only `pj_sockaddr_cmp()` I guess?

pjnath/src/pjnath/ice_session.c
```diff
@@ -3734,6 +3739,56 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
...
+
+                            if (duration > PJ_ICE_SESS_SET_RADDR_DELAY) {
+                                char psrc_addr[PJ_INET6_ADDRSTRLEN] = {0};
+
+                                if (pj_sockaddr_has_addr(src_addr)) {
+                                    pj_sockaddr_print(src_addr, psrc_addr, 
+                                                      sizeof(psrc_addr), 3);
+                                }
+                                pj_sockaddr_cp(raddr, src_addr);
+                                PJ_LOG(4, (ice->obj_name, "Using [%s] as valid"
```
nanangizz, line 3771: As discussed before, I don't think brackets are good idea for printing IP address because IPv6 may also add brackets. Also below.

nanangizz (approved):
just a minor below

pjnath/include/pjnath/config.h
```diff
@@ -394,6 +394,28 @@
 #   define PJ_ICE_NOMINATED_CHECK_DELAY             (4*PJ_STUN_RTO_VALUE)
 #endif
 
+/**
+ * Specify whether to check the source address of the incoming messages.
+ * The source address will be compared to the remote candidate which has
+ * a completed connectivity check or received a connectivity check.
+ * 
+ * Defalut: 1 (yes)
```
sauwming, line 402: Minor: default

Also below.

sauwming (approved):
minor typos

sauwming:
Also merge with master and make sure it passes the CI tests before merging.

pjmedia/include/pjmedia/config.h
nanangizz: > It might add a slight delay in starting the handshake (wait for the transport address and RTP source address to be available) and possible issues when using aggressive nomination.

Btw, is it safe now to make it enabled by default?

nanangizz:
> It might add a slight delay in starting the handshake (wait for the transport address and RTP source address to be available) and possible issues when using aggressive nomination.

Looks like the above issues are solved already (right?), and as @sauwming suggested for security reason, perhaps making it enabled by default for ICE is safe now (not sure for other transports as the source address may not match the SDP)?

trengginas:
> > It might add a slight delay in starting the handshake (wait for the transport address and RTP source address to be available) and possible issues when using aggressive nomination.
> 
> Looks like the above issues are solved already (right?), and as @sauwming suggested for security reason, perhaps making it enabled by default for ICE is safe now (not sure for other transports as the source address may not match the SDP)?

for ICE it is enabled by default `PJ_ICE_SESS_CHECK_SRC_ADDR 1`
for non ICE it is disabled, `PJMEDIA_SRTP_DTLS_CHECK_HELLO_ADDR 0`

nanangizz:
Are you trying to merge from master? Looks like something was wrong because the PR patch becomes very dirty, it is very difficult to check the changes specific to this PR.

trengginas:
> Are you trying to merge from master? Looks like something was wrong because the PR patch becomes very dirty, it is very difficult to check the changes specific to this PR.

Yes, @sauwming asked to merge with master and make sure it passes the CI test.

nanangizz:
Merging from master should not cause the patch dirty, so something was wrong. Better revert the bad merging and redo the merging.

PR #4262, opened 2025-01-22
Share an auth session between multiple dialogs/regc

jwes:
For Accounts that establish a lot of dialogs it may be beneficial to reuse a already established Authorization.
To achieve this, the account was extended to store a shared pjsip_auth_clt_sess and this is set on the relevant dialogs.

This patch adds PJSIP_SHARED_AUTH_SESSION to enable this behavior in the account and to extend the relevant api.

### DIFF
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -266,7 +266,7 @@ struct AccountSipConfig : public PersistentObject
     /**
      * If this flag is set, the authentication client framework will
      * send an empty Authorization header in each initial request.
-     * Default is no.
+     * Default is false.
      */
     bool                authInitialEmpty;
 
@@ -307,9 +307,9 @@ struct AccountSipConfig : public PersistentObject
      * enabled to work properly, and also will grow usage of the used pool for
      * the cached headers.
      *
-     * Default is no
+     * Default is false
      */
-    bool        sharedAuth;
+    bool        useSharedAuth;
 
 public:
     /**
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -303,7 +303,7 @@ struct AccountSipConfig : public PersistentObject
      * This will use the accounts credentials on outgoing requests,
      * so that less 401/407 Responses will be returned.
      *
-     * May need to have PJSIP_AUTH_AUTO_SEND_NEXT and PJSIP_AUTH_HEADER_CACHING
+     * Needs PJSIP_AUTH_AUTO_SEND_NEXT and PJSIP_AUTH_HEADER_CACHING
      * enabled to work properly, and also will grow usage of the used pool for
      * the cached headers.
      *
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -298,6 +298,19 @@ struct AccountSipConfig : public PersistentObject
      */
     pjsua_ipv6_use      ipv6Use;
 
+    /**
+     * Use a shared authorization session within this account.
+     * This will use the accounts credentials on outgoing requests,
+     * so that less 401/407 Responses will be returned.
+     *
+     * May need to have PJSIP_AUTH_AUTO_SEND_NEXT and PJSIP_AUTH_HEADER_CACHING
+     * enabled to work properly, and also will grow usage of the used pool for
+     * the cached headers.
+     *
+     * Default is no
+     */
+    bool        sharedAuth;
+
 public:
     /**
      * Read this object from a container node.
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -266,7 +266,7 @@ struct AccountSipConfig : public PersistentObject
     /**
      * If this flag is set, the authentication client framework will
      * send an empty Authorization header in each initial request.
-     * Default is false.
+     * Default is no.
      */
     bool                authInitialEmpty;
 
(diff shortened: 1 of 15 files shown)


pjsip/src/pjsip/sip_auth_client.c
```diff
@@ -712,6 +737,26 @@ static pjsip_cached_auth *find_cached_auth( pjsip_auth_clt_sess *sess,
...
+#if defined(PJSIP_SHARED_AUTH_SESSION) && PJSIP_SHARED_AUTH_SESSION
+    pj_status_t lock_status;
+    pj_bool_t with_parent = PJ_FALSE;
+    lock_status = pj_lock_acquire(sess->lock);
+    pjsip_cached_auth * pauth = NULL;
+    if (lock_status != PJ_SUCCESS) {
+        return NULL;
+    }
+    if (sess->parent) {
+         pauth = sess->parent->cached_auth.next;
```
nanangizz, line 749: No iteration to lookup the matched `realm` & `algorithm_type` on parent?
(Or perhaps recursive call to self like in `auth_find_cred()`)

sauwming:
I like the feature and I prefer a runtime option in `pjsua_acc_config` rather than a compile-time option (so you can get rid of all those annoying `#ifdef`).

But don't do any modifications yet, let's hear first what others think about this.

nanangizz:
> I like the feature and I prefer a runtime option in `pjsua_acc_config` rather than a compile-time option (so you can get rid of all those annoying `#ifdef`).
> 
> But don't do any modifications yet, let's hear first what others think about this.

Sounds like a good idea!

pjsip/src/pjsip/sip_auth_client.c
```diff
@@ -105,6 +108,28 @@ const pjsip_auth_algorithm pjsip_auth_algorithms[] = {
 #  define AUTH_TRACE_(expr)
 #endif
 
+#if defined(PJSIP_SHARED_AUTH_SESSION) && PJSIP_SHARED_AUTH_SESSION
+#define DO_ON_PARENT_LOCKED(sess, call) \
+    do { \
+        pj_status_t on_parent, lock_status; \
+        pj_bool_t with_parent = PJ_FALSE; \
+        lock_status = pj_lock_acquire(sess->lock); \
```
nanangizz, line 116: The acquired lock belongs to the child, right? While IMO it should be the parent that needs synchronization.

jwes: oh, your completely right. the lock has to be shared between all sessions.
will move the lock call behind the null check and also call sess->parent->lock

jwes:
> I like the feature and I prefer a runtime option in `pjsua_acc_config` rather than a compile-time option (so you can get rid of all those annoying `#ifdef`).
> 
> But don't do any modifications yet, let's hear first what others think about this.

yeah, that sounds good. I wasn't sure if that feature should be available without compile option, so i used a conservative approach

jwes:
is there something i need to do regarding the failed checks?

sauwming:
I'm currently working on the fix on #4263.
Just waiting for the CI test there to complete and then I'll merge to master.

sauwming:
You can merge the master branch to fix the failed tests.

jwes:
i just wanted to switch to an config parameter.
If i move it to the account config, it seems, that i need to set it up on acc_modify.

Is it safe to call pjsip_auth_ctl_deinit in acc_modify, since it destroys a pool, that it did not create itself?

sauwming:
I don't understand the purpose of doing so, but I don't think you should call `auth_clt_deinit()` in `acc_modify()`.
But anyway, you can proceed with what you think is best and we'll review it later.

nanangizz:
Generally looks okay to me, perhaps a few touch-ups may improve it a bit, for example:
- for efficiency, only parent auth sess needs lock, alternative approach from top of my head: 
  - configurable via option param of `auth_sess_init()`
  - make a new API such as `auth_sess_set_parent()`, this way the lock can be created for parent only & app does not need to do the locking.
- improve docs, e.g: `May need to have PJSIP_AUTH_AUTO_SEND_NEXT`, shouldn't it be required as it is the main purpose of this feature? As otherwise it will always wait for 401/407.
- somehow lock/unlock return code check seems to reduce readibility a bit, unless expecting issues (e.g: race between destroy & lock/unlock), we don't usually check it :)
- column width limit (80 chars).

If you want, we can do the above directly in the PR branch.

pjsip/include/pjsua2/account.hpp
```diff
@@ -298,6 +298,19 @@ struct AccountSipConfig : public PersistentObject
...
+     * Use a shared authorization session within this account.
+     * This will use the accounts credentials on outgoing requests,
+     * so that less 401/407 Responses will be returned.
+     *
+     * Needs PJSIP_AUTH_AUTO_SEND_NEXT and PJSIP_AUTH_HEADER_CACHING
+     * enabled to work properly, and also will grow usage of the used pool for
+     * the cached headers.
+     *
+     * Default is no
+     */
```
sauwming, line 311: Default is false.

jwes: ah, is searched for inspiration and found this, therefore i took no.
Should i change that too?
   ```
    /**
     * If this flag is set, the authentication client framework will
     * send an empty Authorization header in each initial request.
     * Default is no.
     */
    bool                authInitialEmpty;
    ```

sauwming: If you want to use `no`, it's also okay, but usually you need to include both the C/C++ value and the meaning.
For example:
Default is no (false).
Default is disabled/false.

For `authInitialEmpty`, if you want to change it, better do it in a separate PR, so this PR doesn't get mixed up. But leaving it unchanged is also fine.

jwes:
merged the master again, maybe that fixes something?

sauwming (approved):

nanangizz (approved):

PR #4263, opened 2025-01-23
Set CI vars in GH workflow file

sauwming:
Setting the variables in https://github.com/pjsip/pjproject/settings/variables/actions doesn't work for forks.
"`They are not passed to workflows that are triggered by a pull request from a fork.`"


bennylp:
Just to add that this changes #4014 in that we no longer use repository variables.

bennylp (approved):

sauwming:
Finally, it works

nanangizz (approved):

PR #4264, opened 2025-01-23
Handle TLS transport socket bind error

Mokolea:
When creating an endpoint TLS transport and we have a socket bind error, this is not handled immediately properly

Using PJSIP v2.14.1

### DIFF
--- a/pjsip/src/pjsip/sip_transport_tls.c
+++ b/pjsip/src/pjsip/sip_transport_tls.c
@@ -545,9 +545,11 @@ PJ_DEF(pj_status_t) pjsip_tls_transport_lis_start(pjsip_tpfactory *factory,
         if (status == PJ_SUCCESS)
             pj_sockaddr_cp(listener_addr, (pj_sockaddr_t*)&info.local_addr);
 
-    }
-    else if (status != PJ_SUCCESS)
+    } else {
+        pj_ssl_sock_close(listener->ssock);
+        listener->ssock = NULL;
         return status;
+    }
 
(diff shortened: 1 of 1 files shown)


sauwming:
I believe it's intentional not to return error.
```
    /* Start accepting incoming connections. Note that some TLS/SSL
     * backends may not support for SSL socket server.
     */    
```
We want the factory to still run (as client-only). But if this doesn't work as intended, it means the issue is perhaps somewhere else.

sauwming:
As previously said, we shouldn't return error when creating listener.
I tested it here and it seems to work fine, the only problem is the misleading log:
`SIP TLS listener is ready for incoming connections at ...`

So I believe the fix here should be to close the listener socket instead:
```
status = pj_ssl_sock_start_accept2(...)
if (status == PJ_SUCCESS || status == PJ_EPENDING) {
...
} else {
        pj_ssl_sock_close(listener->ssock);
        listener->ssock = NULL;
}
```

This way, the log will correctly indicate: `SIP TLS is ready (client only)`

The patch:
[sip_tp_tls.txt](https://github.com/user-attachments/files/18529172/sip_tp_tls.txt)

Mokolea:
We still need to return the socket bind error (WSAEADDRINUSE), if there is one, if port is already in use

Note: We configured:
```
#define PJSIP_TLS_TRANSPORT_REUSEADDR   0
```

sauwming:
Unfortunately, that's not the behaviour we're looking for. Therefore, I will close this PR.

Issue #4242, opened 2025-01-06
SendInstantMessageParam.setUserData() requires SWIGTYPE_p_void, which has protected access


YogatBear:
### Describe the feature

In my app, I want to be able to track if a message has been received by the other client. I am using `Account.onInstantMessageStatus()` to receive the status code from the server, but I also needed a way to know for which message the status belongs to. I then came across `SendInstantMessageParam.setUserData()`, which seems to be exactly what I need. However, I see that I cannot create the SWIGTYPE_p_void object without altering the source code.

I am not too familiar with SWIG and the reasoning for not being able to use `SendInstantMessageParam.setUserData()`out the box, but I noticed that changing the constructors to public works, and `OnInstantMessageStatusParam.getUserData()` returns the data correctly.

### Describe alternatives you've considered

If there is an alternative solution for tracking which instant message the status callback belongs to, I would appreciate that as well! 

### Additional context

PJSUA2 2.14 on Java Android

nanangizz:
Hi @YogatBear, could you confirm whether the solution offered by PR#4273 (i.e: the type of user data is mapped to `long`) works for you? Or if you have any feedback?

YogatBear:
Hey @nanangizz. Yes absolutely. As you mentioned, the intention is to pass and retrieve call/message ID.

PR #4273, opened 2025-01-28
Various works on SWIG Java

nanangizz:
1. Fix type mapping (`SWIGTYPE_*`):
    1. Map C `void*` & `void**` to Java `long` (was `SWIGTYPE_p_void` & `SWIGTYPE_p_p_void`, which was not really usable), this should fix issues like #4242.
    2. Map `pjmedia_aud_dev_index` to `int`.
    3. Map `unsigned char[20]` for `SslCertInfo.serialNo` to Java `short array`
    This also updates pjsua.i, e.g: tab->space, reorder things.

2. Update `swig_java_pjsua2.vcxproj`:
    1. Rename config "Debug" & "Release" to "Debug-Dynamic" & "Release-Dynamic" as the project actually builds dynamic libs.  Also fix the property sheet dependencies from `*-static` to `*-dynamic`.
    2. Update other settings, e.g: built tool version from 140 to 143.

3. Update `symbols.lst`: added missing new types, tab->space, reorder alphabetically.


sauwming:
So by mapping to long, will you be able to pass Java object as user data? (e.g. SendInstantMessageParam.setUserData(Object) )

Also check if utilTimerSchedule() and utilTimerCancel() works with this patch.

nanangizz:
> So by mapping to long, will you be able to pass Java object as user data? (e.g. SendInstantMessageParam.setUserData(Object) )

No, it is a `long` not `Object`.
I decided to map it to `long` as perhaps it is safer (for us now :), as if it is an `Object`, I guess we need to maintain the lifetime of the object, e.g: add ref, etc.
The issue reporter seems to use `SWIGTYPE_p_void` which basically wraps C pointer as `long`, so perhaps it is account ID or call ID that is used as user data instead of the account/call object.

sauwming (approved):

trengginas (approved):

pjsip-apps/src/swig/java/sample.java
```diff
@@ -124,6 +134,9 @@ private void setOutputVidWin() {}
         private static void runWorker() {
                 try {                                   
                         app.init(observer, ".", true);
+                        
+                        app.ep.utilLogWrite(3, "-TIMER-", "Scheduling timer: timeout=1000ms, user data=-99");
+                        app.ep.utilTimerSchedule(1000, -99);
```
sauwming, line 139: We can cancel this, right?

nanangizz: Yes, just tested. Added sample code.

sauwming: Great!

PR #4279, opened 2025-01-30
Fix build error when PJ_LOG_MAX_LEVEL is zero

nanangizz:
The `pj_log_get_log_func()` is not defined when `PJ_LOG_MAX_LEVEL` is set to zero.

Thanks to Giorgio Alfarano for the report.


sauwming (approved):

bennylp (approved):

bennylp:
As a (self) reminder, we need CI job to test various compile time switches like this, epoll, pool alt api, conf switch, etc.

PR #4286, opened 2025-02-04
Add API to register custom SDP comparison callback

sauwming:
Since SDP can be customized (for example, it can contain custom attributes), app may wish to add its own SDP comparison logic on top of the implementation provided by `pjmedia_sdp_session_cmp()`.

The PR will provide API which allows app to register a custom SDP comparison callback.

### DIFF
--- a/pjmedia/include/pjmedia/sdp.h
+++ b/pjmedia/include/pjmedia/sdp.h
@@ -805,7 +805,13 @@ pjmedia_sdp_session_clone( pj_pool_t *pool,
 
 
 /**
- * Compare two SDP session for equality.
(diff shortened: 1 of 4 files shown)


nanangizz:
While here, `pjmedia_sdp_session_cmp()` docs may be better to be updated too on which SDP fields it compares.

Btw, perhaps it is good idea to let `pjmedia_sdp_session_cmp()` invokes the callback too (also, make the callback to be part of `sdp_cmp` instead of `sdp_neg`)?

nanangizz (approved):

trengginas (approved):

PR #4289, opened 2025-02-04
Fixed issue with SDP version when reoffer is rejected

sauwming:
History:
- We #3322
- Then in #3393, the origin version is no longer incremented when the reoffer is rejected, but we need to revert the `last_sent` as well, otherwise the `last_sent` will contain the rejected reoffer.

Sipp script to test the scenario is available in `scripts-sipp/uas-reinv-after-failed-nego.xml`.

### DIFF
--- a/pjmedia/src/pjmedia/sdp_neg.c
+++ b/pjmedia/src/pjmedia/sdp_neg.c
@@ -1940,15 +1940,18 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_cancel_offer(pjmedia_sdp_neg *neg)
                      neg->state == PJMEDIA_SDP_NEG_STATE_REMOTE_OFFER,
                      PJMEDIA_SDPNEG_EINSTATE);
 
-    // No longer needed after #3322.
-    if (0 && neg->state == PJMEDIA_SDP_NEG_STATE_LOCAL_OFFER &&
+    if (neg->state == PJMEDIA_SDP_NEG_STATE_LOCAL_OFFER &&
         neg->active_local_sdp) 
     {
         /* Increment next version number. This happens if for example
          * the reinvite offer is rejected by 488. If we don't increment
          * the version here, the next offer will have the same version.
          */
-        neg->active_local_sdp->origin.version++;
+        // No longer needed after #3322.
+        // neg->active_local_sdp->origin.version++;
+
+        /* Revert last sent SDP. */
+        neg->last_sent = neg->active_local_sdp;
     }
 
     /* Revert back initial SDP */


pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1940,15 +1940,18 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_cancel_offer(pjmedia_sdp_neg *neg)
...
+    if (neg->state == PJMEDIA_SDP_NEG_STATE_LOCAL_OFFER &&
         neg->active_local_sdp) 
     {
         /* Increment next version number. This happens if for example
          * the reinvite offer is rejected by 488. If we don't increment
          * the version here, the next offer will have the same version.
          */
-        neg->active_local_sdp->origin.version++;
+        // No longer needed after #3322.
+        // neg->active_local_sdp->origin.version++;
```
nanangizz, line 1951: With the code is commented, the next local SDP will still have version incremented, right?

sauwming: Only if it's different

nanangizz: Different to `active_local_sdp` or the last rejected?
(I guess the comparison has to be done against any last SDP, in this case it is the rejected one?)

sauwming: With the change in #3393 that no longer increments the version, the current behavior is as follows:
SDP version 1, reoffer SDP v 2, rejected, compare with `neg->last_sent`, if same send SDP version 1 else version 2.

Without the patch, the comparison will be done to SDP in the rejected reoffer.
With the patch, the comparison will be done to `active_local_sdp`.

Refer to the PR desc of #3393.

Note: we have a Sipp script to test the scenario in `uas-reinv-after-failed-nego.xml`.

nanangizz: Okay, so the comparison for SDP version increment is always done against `active_local_sdp` (any rejected offer is ignored).

Re: Sipp script, it does not seem to verify the SDP version, just reproducing the scenario, right?

nanangizz (approved):

pjmedia/src/pjmedia/sdp_neg.c
sauwming: Correct, any rejected offer should be ignored (as the RFC says: `as if no re-INVITE/UPDATE had been issued`).
It may cause the SDP version to decrease in the scenario: `SDP version 1, reoffer SDP v 2, rejected, if we send reoffer with the same SDP, the version will be back to 1`, which may seem strange, but I suppose that's what the RFC recommends.

Yes, for reproducing only.

trengginas (approved):

PR #4295, opened 2025-02-05
Add TLS/SSL backend: Mbed TLS

kolbikds-arlo:
### Introduction

Currently the library has implemented TLS/SSL using various backends such as OpenSSL, BoringSSL, GnuTLS, Apple Network framework. This PR will add another backend: Mbed TLS, which is designed for embedded applications/uses cases. As usual, this backend is implemented for the PJLIB SSL socket API.

### Advantages of Mbed TLS over Crypto libraries (OpenSSL)
From an Embedded Development perspective, Mbed TLS is often the better choice that offers several advantages. It is ligth weight, very highly configurable, and focused on resource-constrained environmnets. At the moment in PJSIP, the only OpenSSL can be considered for embedded systems. However, OpenSSL is designed for general-purpose use, so it includes a wide range of features that can be overkill for embedded applications, especially for RTOS-based platforms.

### Platform availability
MbedTLS is highly portable and works on a wide range of platforms, what makes it a versatile choice for integrating with PJSIP across diverse environments.
A very minimum list of vendors that utilizes Mbed TLS in their SDKs:

Espressif IoT Development Framework
https://github.com/espressif/esp-idf/tree/c5865270b50529cd32353f588d8a917d89f3dba4/components/mbedtls

STMicroelectronics, Mbed TLS is part of each STM32Cube MCU and MPU Package
e.g. https://github.com/STMicroelectronics/STM32CubeN6/tree/main/Middlewares/Third_Party/mbedtls

SiliconLabs, Gecko SDK
https://github.com/SiliconLabs/gecko_sdk/tree/gsdk_4.4/util/third_party/mbedtls

etc,

### Versions tested
Mbed TLS 3.6.0 & 3.6.2

### Build and install MbedTLS (Ubuntu 22.04-1)
```
git clone https://github.com/Mbed-TLS/mbedtls.git
cd mbedtls
git checkout v3.6.2
mkdir build && cd build
cmake -DENABLE_TESTING=Off ..
make
sudo make install
```

### Activate MbedTLS support in PJSIP
`./configure --with-mbedtls=/usr/local && make dep && make`

### Limitations

- The server does not handle multiple clients. So the perf_test is disabled.
- This does not cover DTLS-SRTP key negotiation for SRTP media transport.

### Notes:

- It supports Mbed TLS 3.6.x version since it has a long-term support (LTS) branch until at least March 2027.
- TLS 1.2 and TLS 1.3 are supported.

### DIFF
--- a/pjlib/src/pj/ssl_sock_mbedtls.c
+++ b/pjlib/src/pj/ssl_sock_mbedtls.c
@@ -1,0 +1,895 @@
+/*
+ * Copyright (C) 2025 Teluu Inc. (http://www.teluu.com)
+ * Copyright (c) 2025 Arlo Technologies, Inc. (https://www.arlo.com)
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include <pj/activesock.h>
+#include <pj/assert.h>
+#include <pj/compat/socket.h>
+#include <pj/errno.h>
+#include <pj/file_io.h>
+#include <pj/list.h>
+#include <pj/lock.h>
+#include <pj/log.h>
+#include <pj/math.h>
+#include <pj/os.h>
+#include <pj/pool.h>
+#include <pj/ssl_sock.h>
+#include <pj/string.h>
+#include <pj/timer.h>
+
+/* Only build when PJ_HAS_SSL_SOCK is enabled and when the backend is
+ * MbedTLS.
+ */
+#if defined(PJ_HAS_SSL_SOCK) && PJ_HAS_SSL_SOCK != 0 && \
+    (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_MBEDTLS)
+
+#include "mbedtls/ctr_drbg.h"
+#include "mbedtls/debug.h"
+#include "mbedtls/entropy.h"
+#include "mbedtls/error.h"
+#include "mbedtls/net_sockets.h"
+#include "mbedtls/oid.h"
+#include "mbedtls/platform.h"
+#include "mbedtls/pk.h"
+#include "mbedtls/ssl.h"
+#include "mbedtls/version.h"
+
+#ifdef MBEDTLS_DEBUG_C
+#define MBEDTLS_DEBUG_VERBOSE 1
+#endif
+
+#define SSL_SOCK_IMP_USE_CIRC_BUF
+
+#include "ssl_sock_imp_common.h"
+
+#define THIS_FILE "ssl_sock_mbedtls.c"
+
+/*
+ * Secure socket structure definition.
+ */
+typedef struct mbedtls_sock_t {
+    pj_ssl_sock_t base;
+
+    mbedtls_ssl_context ssl_ctx;
+    mbedtls_ssl_config ssl_config;
+    mbedtls_entropy_context entropy;
+    mbedtls_ctr_drbg_context ctr_drbg;
+    mbedtls_x509_crt cacert;
+    mbedtls_x509_crt cert;
+    mbedtls_pk_context pk_ctx;
+} mbedtls_sock_t;
+
+#include "ssl_sock_imp_common.c"
+
+/*
+ *******************************************************************
+ * Static/internal functions.
+ *******************************************************************
+ */
+
+/* MbedTLS way of reporting internal operations. */
+static void mbedtls_print_logs(void *ctx, int level, const char *file,
+                               int line, const char *str)
+{
+    PJ_UNUSED_ARG(ctx);
+    PJ_UNUSED_ARG(level);
+
+    const char* last_slash = strrchr(file, '/');
+    const char* file_name = last_slash ? last_slash + 1 : file;
+
+    PJ_LOG(3, (THIS_FILE, "%s:%d: %s", file_name, line, str));
+}
+
+/* Convert from MbedTLS error to pj_status_t. */
+static pj_status_t ssl_status_from_err(pj_ssl_sock_t *ssock, int err)
+{
+    PJ_UNUSED_ARG(ssock);
+
+    pj_status_t status;
+
+    switch (err) {
+    case 0:
+        status = PJ_SUCCESS;
+        break;
+    case MBEDTLS_ERR_SSL_BUFFER_TOO_SMALL:
+    case MBEDTLS_ERR_SSL_ALLOC_FAILED:
+        status = PJ_ENOMEM;
+        break;
+    case MBEDTLS_ERR_SSL_UNEXPECTED_MESSAGE:
+        status = PJ_ETOOBIG;
+        break;
+    case MBEDTLS_ERR_SSL_NO_CLIENT_CERTIFICATE:
+        status = PJ_ENOTFOUND;
+        break;
+    case MBEDTLS_ERR_SSL_TIMEOUT:
+        status = PJ_ETIMEDOUT;
+        break;
+    case MBEDTLS_ERR_SSL_INTERNAL_ERROR:
+    case MBEDTLS_ERR_SSL_UNKNOWN_IDENTITY:
+        status = PJ_EBUG;
+        break;
+    case MBEDTLS_ERR_SSL_WANT_READ:
+    case MBEDTLS_ERR_SSL_WANT_WRITE:
+    case MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS:
+    case MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS:
+    case MBEDTLS_ERR_SSL_RECEIVED_EARLY_DATA:
+    case MBEDTLS_ERR_SSL_RECEIVED_NEW_SESSION_TICKET:
+        status = PJ_EPENDING;
+        break;
+
+    case MBEDTLS_ERR_SSL_UNEXPECTED_RECORD:
+    case MBEDTLS_ERR_SSL_ILLEGAL_PARAMETER:
+    case MBEDTLS_ERR_SSL_BAD_PROTOCOL_VERSION:
+    case MBEDTLS_ERR_SSL_HANDSHAKE_FAILURE:
+        status = PJ_EINVAL;
+        break;
+    default:
+        status = PJ_EUNKNOWN;
+        break;
+    }
+
+    return status;
+}
+
+static int ssl_data_push(void *ctx, const unsigned char *buf, size_t len)
+{
+    pj_ssl_sock_t *ssock = (pj_ssl_sock_t *)ctx;
+
+    pj_lock_acquire(ssock->circ_buf_output_mutex);
+    if (circ_write(&ssock->circ_buf_output, buf, len) != PJ_SUCCESS) {
+        pj_lock_release(ssock->circ_buf_output_mutex);
+
+        return MBEDTLS_ERR_SSL_WANT_WRITE;
+    }
+
+    pj_lock_release(ssock->circ_buf_output_mutex);
+
+    return len;
+}
+
+static int ssl_data_pull(void *ctx, unsigned char *buf, size_t len)
+{
+    pj_ssl_sock_t *ssock = (pj_ssl_sock_t *)ctx;
+
+    pj_lock_acquire(ssock->circ_buf_input_mutex);
+
+    if (circ_empty(&ssock->circ_buf_input)) {
+        pj_lock_release(ssock->circ_buf_input_mutex);
+
+        return MBEDTLS_ERR_SSL_WANT_READ;
+    }
+
+    pj_size_t circ_buf_size = circ_size(&ssock->circ_buf_input);
+    pj_size_t read_size = PJ_MIN(circ_buf_size, len);
+
+    circ_read(&ssock->circ_buf_input, buf, read_size);
+    pj_lock_release(ssock->circ_buf_input_mutex);
+
+    return read_size;
+}
+
+/* Get Common Name field string from a general name string */
+static void cert_get_cn(const pj_str_t *gen_name, pj_str_t *cn)
+{
+    pj_str_t CN_sign = {"CN=", 3};
+    char *p, *q;
+
+    pj_bzero(cn, sizeof(cn));
+
+    p = pj_strstr(gen_name, &CN_sign);
+    if (!p)
+        return;
+
+    p += 3; /* shift pointer to value part */
+    pj_strset(cn, p, gen_name->slen - (p - gen_name->ptr));
+    q = pj_strchr(cn, ',');
+    if (q)
+        cn->slen = q - p;
+}
+
+static void cert_get_time(pj_time_val *tv,
+                          const mbedtls_x509_time *time,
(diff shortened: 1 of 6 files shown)


sauwming:
A few (mostly minor) build warnings:
```
../src/pj/ssl_sock_mbedtls.c:379:16: warning: variable 'min_proto' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
  379 |     } else if (ssock->param.proto & PJ_SSL_SOCK_PROTO_TLS1_3) {
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/pj/ssl_sock_mbedtls.c:383:59: note: uninitialized use occurs here
  383 |     mbedtls_ssl_conf_min_tls_version(&mssock->ssl_config, min_proto);
      |                                                           ^~~~~~~~~
../src/pj/ssl_sock_mbedtls.c:379:12: note: remove the 'if' if its condition is always true
  379 |     } else if (ssock->param.proto & PJ_SSL_SOCK_PROTO_TLS1_3) {
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     ^
In file included from ../src/pj/ssl_sock_mbedtls.c:77:
../src/pj/ssl_sock_imp_common.c:91:13: warning: unused function 'circ_reset' [-Wunused-function]
   91 | static void circ_reset(circ_buf_t* cb)
      |             ^~~~~~~~~~
../src/pj/ssl_sock_imp_common.c:141:13: warning: unused function 'circ_read_cancel' [-Wunused-function]
  141 | static void circ_read_cancel(circ_buf_t* cb, pj_size_t len)
      |             ^~~~~~~~~~~~~~~~
../src/pj/ssl_sock_imp_common.c:222:12: warning: unused function 'get_ip_addr_ver' [-Wunused-function]
  222 | static int get_ip_addr_ver(const pj_str_t *host)
      |    
      ../src/pjlib-test/ssl_sock.c:1323:12: warning: unused function 'perf_test' [-Wunused-function]
 1323 | static int perf_test(unsigned clients, unsigned ms_handshake_timeout)
```

sauwming:
Thank you for the contribution. I tested it here using 3.6.2, running PJLIB-test and pjsua app (registration and call), it seems to work fine. Just a few compilation warnings above.

kolbikds-arlo:
@sauwming Thanks for the feedback. I did suppress compilation warnings for `ssl_sock_mbedtls.c`. Regarding `warning: unused function 'perf_test'` any suggestion to nicely resolve this warning or we can leave it as is?

sauwming:
You can do the same thing as with the other functions, i.e.:
#else
    PJ_UNUSED_ARG(...)

bennylp:
It would be great if you could add the following information in this PR description:

- the benefit of Mbed TLS compared to using others
- platform availability
- versions tested
- (briefly) how to install it for using it with pjsip
- how to activate this feature with pjsip (`configure` options, etc)

Thanks!

nanangizz:
> The server does not handle multiple clients.

Could you also describe a bit more about this?

kolbikds-arlo:
> > The server does not handle multiple clients.
> 
> Could you also describe a bit more about this?

As a server the current implementation supports only a single client connection, but it does not mean that implementing a server option to support multiple clients is impossible.

nanangizz:
> > > The server does not handle multiple clients.
> > 
> > 
> > Could you also describe a bit more about this?
> 
> As a server the current implementation supports only a single client connection, but it does not mean that implementing a server option to support multiple clients is impossible.

I mean I can't find a docs mentioning this limitation, do you know one? And from a quick search, it seems that it can handle multiple clients normally (an SSL context, bio, etc for each client as in other backends), for example [here](https://forums.mbed.com/t/mbedtls-ssl-read-write-concurrently-from-multiple-ssl-connections/3772/4).

kolbikds-arlo:
> > > > The server does not handle multiple clients.
> > > 
> > > 
> > > Could you also describe a bit more about this?
> > 
> > 
> > As a server the current implementation supports only a single client connection, but it does not mean that implementing a server option to support multiple clients is impossible.
> 
> I mean I can't find a docs mentioning this limitation, do you know one? And from a quick search, it seems that it can handle multiple clients normally (an SSL context, bio, etc for each client as in other backends), for example [here](https://forums.mbed.com/t/mbedtls-ssl-read-write-concurrently-from-multiple-ssl-connections/3772/4).

I'm not sure we are on the same page. To clarify, the limitations in description are specific to the `ssl_sock_mbedtls.c` implementation and are not related to the Mbed TLS library itself. Based on your message, it seems your concern is about the capabilities of the Mbed TLS library — is that correct?

nanangizz:
> > > > > The server does not handle multiple clients.
> > > > 
> > > > 
> > > > Could you also describe a bit more about this?
> > > 
> > > 
> > > As a server the current implementation supports only a single client connection, but it does not mean that implementing a server option to support multiple clients is impossible.
> > 
> > 
> > I mean I can't find a docs mentioning this limitation, do you know one? And from a quick search, it seems that it can handle multiple clients normally (an SSL context, bio, etc for each client as in other backends), for example [here](https://forums.mbed.com/t/mbedtls-ssl-read-write-concurrently-from-multiple-ssl-connections/3772/4).
> 
> I'm not sure we are on the same page. To clarify, the limitations in description are specific to the `ssl_sock_mbedtls.c` implementation and are not related to the Mbed TLS library itself. Based on your message, it seems your concern is about the capabilities of the Mbed TLS library — is that correct?

Ah okay. Correct, I thought the limitation is caused by the Mbed TLS. Thanks for the clarification.

pjlib/src/pj/ssl_sock_mbedtls.c
```diff
@@ -0,0 +1,900 @@
...
+
+    pj_lock_acquire(ssock->circ_buf_input_mutex);
+
+    if (circ_empty(&ssock->circ_buf_input)) {
+        pj_lock_release(ssock->circ_buf_input_mutex);
+
+        return MBEDTLS_ERR_SSL_WANT_READ;
+    }
+
+    pj_size_t circ_buf_size = circ_size(&ssock->circ_buf_input);
```
sauwming, line 177: put var declarations on top.
refer to:
https://docs.pjsip.org/en/latest/get-started/coding-style.html#no-declaration-after-statement

Also more below.

pjlib/src/pj/ssl_sock_mbedtls.c
```diff
@@ -0,0 +1,900 @@
...
+    if (ret != 0) {
+        PJ_LOG(1, (THIS_FILE, "Failed to mbedtls_ssl_config_defaults, "
+                              "ret = -0x%04X", -ret));
+        goto out;
+    }
+
+    ret = mbedtls_ctr_drbg_seed(&mssock->ctr_drbg, mbedtls_entropy_func,
+                                &mssock->entropy,
+                                (const unsigned char *)pers, strlen(pers));
+    if (ret != 0) {
```
sauwming, line 664: Is this also fatal?

pjlib/src/pj/ssl_sock_mbedtls.c
```diff
@@ -0,0 +1,900 @@
...
+    mbedtls_x509_crt_init(&mssock->cacert);
+    mbedtls_x509_crt_init(&mssock->cert);
+    mbedtls_pk_init(&mssock->pk_ctx);
+
+    const int endpoint =
+            ssock->is_server ? MBEDTLS_SSL_IS_SERVER : MBEDTLS_SSL_IS_CLIENT;
+    ret = mbedtls_ssl_config_defaults(&mssock->ssl_config, endpoint,
+                                      MBEDTLS_SSL_TRANSPORT_STREAM,
+                                      MBEDTLS_SSL_PRESET_DEFAULT);
+    if (ret != 0) {
```
sauwming, line 655: Is this really a fatal error?
If not, then just continue and log some warning (level 3 or 4).

sauwming (approved):

pjlib/src/pj/ssl_sock_mbedtls.c
sauwming: ok, perhaps I am using too strong term.
What I meant is that is it a non-recoverable error? Because it looks like it's just trying to get a default config. Should it fail, I believe we can still proceed, instead of failing the entire SSL creation process.

kolbikds-arlo: It will be undefined behaviour for given mbedtls config if this is ignored. The endpoint type and transport type are configured only by this API in entire ssl_sock_mbedtls.c. If this fails, these options are undefined for mbedtls config. I would treat such a case as an error.

sauwming: Ok

pjlib/src/pj/ssl_sock_mbedtls.c
kolbikds-arlo: I resolved all related warnings with -Wdeclaration-after-statement enabled.

trengginas (approved):

nanangizz (approved):

amubiera:
Awesome contribution, @kolbikds-arlo! 

I actually have my own fork with a nearly identical implementation.

I have some comments / questions:
1. `transport_srtp_sdes.c` : `generate_crypto_attr_value()` should probably have a `PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_MBEDTLS` section to use a more cryptographically secure RNG via `mbedtls_ctr_drbg_random` instead of using `pj_rand`.
2. Interesting that `mbedtls_ssl_set_hostname` needed to be called in `ssl_create` instead of `ssl_set_peer_name`. I didn't catch that.
3. I have a WiP implementation of `transport_srtp_dtls.c` using mbedTLS. It is rather large, as is the OpenSSL version. Do you also have an implementation of DTLS for mbedTLS? It can get a bit tricky.

kolbikds-arlo:
Thanks for your feedback, @amubiera!

Here are my responses to your comments/questions.

1. "mbedtls_ctr_drbg_random instead of using pj_rand"
Yes, there is such opportunity with Mbed TLS in place. I'm still torn between classic cumbersome Mbed TLS RNG boilerplate (`mbedtls_entropy_init`, `mbedtls_ctr_drbg_init`, `mbedtls_ctr_drbg_seed`, `mbedtls_ctr_drbg_random`, `mbedtls_ctr_drbg_free`, `mbedtls_entropy_free`) or standalone `mbedtls_psa_get_random` which requiers PSA Crypto.

2. "ssl_set_peer_name"
I left the comment in related API why. `ssl_set_peer_name` is returning `void`. This is the reason. `ssl_create` can handle errors returned by Mbed TLS APIs properly and pass it to top-level function.

3. DTLS for Mbed TLS
No, I don't. Unfortunately.

4. "mbedtls_psa_crypto_free concern"
That is really good catch. `psa_crypto_free()` shouldn't be called here if there are other components that utilize MbedTLS APIs. Thanks! `psa_crypto_init()` is fine, it may be called more than once. I'll fix it.

5. And perhaps the #ifdef MBEDTLS_PSA_CRYPTO_C should be instead #ifdef MBEDTLS_USE_PSA_CRYPTO.
It is so. Please kindly revisit `ssl_sock_mbedtls.c` one more time.

amubiera:
Thanks for your responses @kolbikds-arlo!

1. Yes, the classic API is rather cumbersome. It's just not a good feeling to see the `Warning: simple random generator is used for generating SRTP key` on every voice call :-)
2. Good call.
3. I may clean up my version of mbedTLS DTLS and contribute it.
4. Sounds good. The version of `psa_crypto_init` in mbedTLS v3.6.x is at least capable of being thread-safe, unliked mbedTLS v2.x.

Re: PSA:
* I suppose one question is why even initialize PSA since you're not (yet) calling `mbedtls_psa_xxx` functions from `ssl_sock_mbedtls.c`?
* `MBEDTLS_USE_PSA_CRYPTO` is not defined by default in `mbedtls_config.h` so the PSA functions are not called by mbedTLS core X.509 APIs.
* If the host application has enabled `MBEDTLS_USE_PSA_CRYPTO` via their config file, then it may be best to leave PSA initialization and free up to the app.

kolbikds-arlo:
> Thanks for your responses @kolbikds-arlo!
> 
>     1. Yes, the classic API is rather cumbersome. It's just not a good feeling to see the `Warning: simple random generator is used for generating SRTP key` on every voice call :-)
> 
>     2. Good call.
> 
>     3. I may clean up my version of mbedTLS DTLS and contribute it.
> 
>     4. Sounds good. The version of `psa_crypto_init` in mbedTLS v3.6.x is at least capable of being thread-safe, unliked mbedTLS v2.x.
> 
> 
> Re: PSA:
> 
>     * I suppose one question is why even initialize PSA since you're not (yet) calling `mbedtls_psa_xxx` functions from `ssl_sock_mbedtls.c`?
> 
>     * `MBEDTLS_USE_PSA_CRYPTO` is not defined by default in `mbedtls_config.h` so the PSA functions are not called by mbedTLS core X.509 APIs.
> 
>     * If the host application has enabled `MBEDTLS_USE_PSA_CRYPTO` via their config file, then it may be best to leave PSA initialization and free up to the app.
> 
> 
> p.s. I'm now using your version of `ssl_sock_mbedtls.c` and it's working great.

The fact that mbedtls_psa_xxx functions aren’t called directly from `ssl_sock_mbedtls.c` doesn’t mean that PSA functions aren’t invoked internally by Mbed TLS. For example, you can easily verify this with TLS 1.3. If Mbed TLS is built with `PSA_CRYPTO_C` and you attempt to establish a connection without calling `psa_crypto_init()`, you will encounter an internal error. This demonstrates that` psa_crypto_init()` is required, even if mbedtls_psa_xxx functions aren’t explicitly called.

amubiera:
> If Mbed TLS is built with PSA_CRYPTO_C and you attempt to establish a connection without calling psa_crypto_init(), you will encounter an internal error. This demonstrates that psa_crypto_init() is required, even if mbedtls_psa_xxx functions aren’t explicitly called.

I see. I thought that was the purpose of `MBEDTLS_USE_PSA_CRYPTO` and `MBEDTLS_PSA_CRYPTO_C` was merely to enable the ability to compile in `mbedtls_psa_xxx` functions.

This:
```
/**
 * \def MBEDTLS_PSA_CRYPTO_C
 *
 * Enable the Platform Security Architecture cryptography API.
 *
 * Module:  library/psa_crypto.c
 *
 * Requires: either MBEDTLS_CTR_DRBG_C and MBEDTLS_ENTROPY_C,
 *           or MBEDTLS_HMAC_DRBG_C and MBEDTLS_ENTROPY_C,
 *           or MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG.
 * Auto-enables: MBEDTLS_CIPHER_C if any unauthenticated (ie, non-AEAD) cipher
 *               is enabled in PSA (unless it's fully accelerated, see
 *               docs/driver-only-builds.md about that).
 */
#define MBEDTLS_PSA_CRYPTO_C
```

Versus:
```
/**
 * \def MBEDTLS_USE_PSA_CRYPTO
 *
 * Make the X.509 and TLS libraries use PSA for cryptographic operations as
 * much as possible, and enable new APIs for using keys handled by PSA Crypto.
 *
 * \note Development of this option is currently in progress, and parts of Mbed
 * TLS's X.509 and TLS modules are not ported to PSA yet. However, these parts
 * will still continue to work as usual, so enabling this option should not
 * break backwards compatibility.
 *
 * \warning If you enable this option, you need to call `psa_crypto_init()`
 * before calling any function from the SSL/TLS, X.509 or PK modules, except
 * for the various mbedtls_xxx_init() functions which can be called at any time.
 *
 * \note An important and desirable effect of this option is that it allows
 * PK, X.509 and TLS to take advantage of PSA drivers. For example, enabling
 * this option is what allows use of drivers for ECDSA, ECDH and EC J-PAKE in
 * those modules. However, note that even with this option disabled, some code
 * in PK, X.509, TLS or the crypto library might still use PSA drivers, if it
 * can determine it's safe to do so; currently that's the case for hashes.
 *
 * \note See docs/use-psa-crypto.md for a complete description this option.
 *
 * Requires: MBEDTLS_PSA_CRYPTO_C.
 *
 * Uncomment this to enable internal use of PSA Crypto and new associated APIs.
 */
//#define MBEDTLS_USE_PSA_CRYPTO
```

PR #4298, opened 2025-02-06
Fixed msg_data assertion in pjsua_acc_send_request() API

sauwming:
According to the doc of `pjsua_acc_send_request()`:
```
 * @param msg_data      Optional headers etc. to be added to the outgoing
 *                      request, or NULL if no custom header is desired.
```

But we raise assertion if no msg_data is provided. Furthermore, we use `msg_data->target_uri` as destination while `dest_uri` parameter itself is unused.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -1560,14 +1560,13 @@ PJ_DEF(pj_status_t) pjsua_acc_send_request(pjsua_acc_id acc_id,
     PJ_ASSERT_RETURN(dest_uri, PJ_EINVAL);
     PJ_ASSERT_RETURN(method, PJ_EINVAL);
     PJ_UNUSED_ARG(options);
-    PJ_ASSERT_RETURN(msg_data, PJ_EINVAL);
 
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -1560,14 +1560,13 @@ PJ_DEF(pj_status_t) pjsua_acc_send_request(pjsua_acc_id acc_id,
     PJ_ASSERT_RETURN(dest_uri, PJ_EINVAL);
     PJ_ASSERT_RETURN(method, PJ_EINVAL);
     PJ_UNUSED_ARG(options);
-    PJ_ASSERT_RETURN(msg_data, PJ_EINVAL);
```
nanangizz, line 1563: While here, perhaps also update acc id verification, e.g:
```
PJ_ASSERT_RETURN(pjsua_acc_is_valid(acc_id), PJ_EINVAL);
```
Or simply remove the incomplete acc id check above & add info that the acc verification will be done by `pjsua_acc_create_request()`.

Also, perhaps better to protect the acc verification & usage using PJSUA lock.

nanangizz (approved):

PR #4300, opened 2025-02-11
Audio and video stream refactoring

sauwming:
In order to support additional media, it would be better to refactor the current implementation of audio and video stream since they contains a lot of duplications, which make code maintenance tedious and more prone to bug (such as if you only fix one stream and do not apply to others).

Scope:
- a new structure `pjmedia_stream_common` is created as a parent of `pjmedia_stream` and `pjmedia_vid_stream`
- a new structure `pjmedia_stream_info_common`  which contains members that exist in both`pjmedia_stream_info` and `pjmedia_vid_stream_info`. Note that unlike the streams (`pjmedia_stream` and `pjmedia_vid_stream`) which are opaque, the  stream info definition is public, so we cannot modify its internal structure.
- common callbacks for RTP and RTCP handling, `on_rx_rtp()` and `on_rx_rtcp()`, which will call its specific stream implementation: `on_stream_rx_rtp()` (currently no specific stream's RTCP handling).
- common group lock handler `on_destroy()` which will call its specific stream's handler `on_stream_destroy()`.

Notes:
- https://github.com/pjsip/pjproject/pull/2713. The PR was only applied to audio stream, but not video stream. Now both will have the same behavior.
- With this refactoring, the setting `PJMEDIA_VID_STREAM_CHECK_RTP_PT` now MUST have the same value as `PJMEDIA_STREAM_CHECK_RTP_PT`. If not, we will print a warning message during compilation and automatically set it to be equal.

### DIFF
--- a/pjmedia/include/pjmedia/stream.h
+++ b/pjmedia/include/pjmedia/stream.h
@@ -78,79 +78,20 @@ PJ_BEGIN_DECL
  */
 
 /**
- * Opaque declaration for media channel.
- * Media channel is unidirectional flow of media from sender to
- * receiver.
- */
-typedef struct pjmedia_channel pjmedia_channel;
-
-/**
- * This structure describes media stream information. Each media stream
+ * This structure describes audio stream information. Each audio stream
  * corresponds to one "m=" line in SDP session descriptor, and it has
  * its own RTP/RTCP socket pair.
  */
 typedef struct pjmedia_stream_info
 {
-    pjmedia_type        type;       /**< Media type (audio, video)          */
-    pjmedia_tp_proto    proto;      /**< Transport protocol (RTP/AVP, etc.) */
-    pjmedia_dir         dir;        /**< Media direction.                   */
-    pj_sockaddr         local_addr; /**< Local RTP address                  */
-    pj_sockaddr         rem_addr;   /**< Remote RTP address                 */
-    pj_sockaddr         rem_rtcp;   /**< Optional remote RTCP address. If
-                                         sin_family is zero, the RTP address
-                                         will be calculated from RTP.       */
-    pj_bool_t           rtcp_mux;   /**< Use RTP and RTCP multiplexing.     */
-#if defined(PJMEDIA_HAS_RTCP_XR) && (PJMEDIA_HAS_RTCP_XR != 0)
-    pj_bool_t           rtcp_xr_enabled;
-                                    /**< Specify whether RTCP XR is enabled.*/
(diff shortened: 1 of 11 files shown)


wosrediinanatour:
Cool thing! Has https://github.com/pjsip/pjproject/tree/master/pjsip-apps/src/3rdparty_media_sample been tested?

pjmedia/include/pjmedia/stream.h
wosrediinanatour, line 1: Since there is already a vid_stream.h, I would rename this file to aud_stream.h.
Same for C files: audio_stream.c, audio_stream_info.c
 
Or even better: audio_stream.h and video_stream.h. Or are there issues with file-name length on any platform?

sauwming: Unfortunately this will break app that uses pjmedia directly. App that includes <pjmedia/stream.h> will fail to build. I can make stream.h to include aud_stream.h to avoid the build failure, but then the name of the structs and APIs is still `pjmedia_stream` instead of `pjmedia_aud_stream`, so I guess we're stuck with the old naming.

sauwming:
> Cool thing! Has https://github.com/pjsip/pjproject/tree/master/pjsip-apps/src/3rdparty_media_sample been tested?

This PR only modifies pjmedia, and it's done without changing any APIs nor how to use the existing streams (audio and video). So pjsua and application won't be affected.

pjmedia/include/pjmedia/stream_common.h
```diff
@@ -54,6 +194,89 @@ typedef struct pjmedia_stream_rtp_sess_info
...
+ *
+ * @param stream        The media stream.
+ *
+ * @param session_info  The stream session info.
+ *
+ * @return              PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t)
+pjmedia_stream_common_get_rtp_session_info(pjmedia_stream_common *stream,
+                                   pjmedia_stream_rtp_sess_info *session_info);
```
nanangizz, line 259: Are these new APIs being used somewhere or perhaps will be in the future?

sauwming: Yes, most of the new APIs will be used by the new text stream.

nanangizz (approved):

sauwming:
The latest commit is to fix failed CI tests.

This is caused by the new API `pjmedia_stream_info_common_from_sdp()` that can return PJ_SUCCESS, indicating that the SDP is actually valid and as per the specification, that the stream info is initialized, but there's no further processing required. An example of such case is that when the media port is zero.

So, I have to modify the new API to add one more parameter as an output param.

PR #4305, opened 2025-02-14
Fixed Coverity warnings

sauwming:


pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3957,6 +3957,8 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
         stream_info.info.vid = vsi;
         enc_name = &vsi.codec_info.encoding_name;
 #endif
+    } else {
+        PJ_ASSERT_RETURN(!"Invalid media!", PJ_EINVAL);
```
nanangizz, line 3961: If this function won't be called for non-audio/video, better put comment here to avoid confusion in the future?

pjmedia/src/pjmedia/vid_stream.c
```diff
@@ -1612,8 +1612,8 @@ PJ_DEF(pj_status_t) pjmedia_vid_stream_get_port(pjmedia_vid_stream *stream,
 {
     pjmedia_stream_common *c_strm = (stream? &stream->base: NULL);
```
nanangizz, line 1613: This init line looks a bit odd now, perhaps just init after the assertion check?

nanangizz (approved):

PR #4308, opened 2025-02-17
pjsua_media: Fix to apply_med_update when PJMEDIA_HAS_VIDEO is disabled.

amubiera:
When video is disabled, the code will fall through to attempt to use uninitialized `si`:
```
pjsip/pjproject/pjsip/src/pjsua-lib/pjsua_media.c:3983:11: warning: 'si' may be used uninitialized [-Wmaybe-uninitialized]
 3983 |     if (si->rtcp_mux && !call_med->enable_rtcp_mux) {
      |         ~~^~~~~~~~~~
```

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -3940,11 +3940,6 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
     pjsua_stream_info stream_info;
     pj_str_t *enc_name = NULL;
 
-    /* Sanity check. */
-    PJ_ASSERT_RETURN(call_med->type == PJMEDIA_TYPE_AUDIO ||
-                     call_med->type == PJMEDIA_TYPE_VIDEO,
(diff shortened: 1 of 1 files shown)


sauwming:
The warning itself (the uninitialized `si`) should have been fixed in #4305.

But your patch to modify the default status seems to have nothing to do with the warning?

sauwming:
> A more explicit fix could be to remove the `PJ_ASSERT_RETURN` and add an `else` after the video check and leave `pj_status_t status` uninitialized.

I prefer this approach (note that `pj_status_t status` will be left initialized to PJ_SUCCESS).

amubiera:
> > A more explicit fix could be to remove the `PJ_ASSERT_RETURN` and add an `else` after the video check and leave `pj_status_t status` uninitialized.
> 
> I prefer this approach (note that `pj_status_t status` will be left initialized to PJ_SUCCESS).

Updated PR.

sauwming (approved):

nanangizz (approved):

PR #4321, opened 2025-02-24
Add logging information when incoming message is ignored caused by unchecked source address

trengginas:
PR #4261 will add the option to check the valid source address.
On ICE, the valid source address is valid for the remote address that has received the connectivity check or has
the connectivity check successful.
Currently, any incoming check will be ignored once ICE is completed. In this case the remote address will not 
set as valid although receiving the connectivity check.
```
2025-02-21T02:14:00.1964594Z caller: 02:13:39.694                icetp00  ICE negotiation failed after 0s:000: All ICE checklists failed (PJNATH_EICEFAILED)
2025-02-21T02:14:00.1965169Z caller: 02:13:39.694          pjsua_media.c  .Call 0: ICE trickle stopped trickling as ICE nego completed
2025-02-21T02:14:00.1965673Z caller: 02:13:39.694                icetp00  .Ignored incoming check after ICE nego has been completed!
2025-02-21T02:14:00.1969105Z caller: 02:13:39.700                icetp00  Ignoring incoming message for component [1] because source addr 10.10.10.155:4008 unrecognized
2025-02-21T02:14:00.1974879Z caller: 02:13:39.720                icetp00  Ignoring incoming message for component [1] because source addr 10.10.10.155:4008 unrecognized
```
This PR will add logging information that incoming message might be ignored when source address is not checked.

### DIFF
--- a/pjnath/src/pjnath/ice_session.c
+++ b/pjnath/src/pjnath/ice_session.c
@@ -3320,9 +3320,7 @@ static void handle_incoming_check(pj_ice_sess *ice,
     pj_ice_sess_cand *rcand;
     unsigned i;
 
-    /* Check if ICE has been completed. We still need to check if remote 
-     * candidate is valid when check_src_addr is set.
-     */
+    /* Check if ICE has been completed */
     if (ice->is_complete) {
         LOG4((ice->obj_name,
               "Ignored incoming check after ICE nego has been completed!"));
--- a/pjnath/src/pjnath/ice_session.c
+++ b/pjnath/src/pjnath/ice_session.c
@@ -3323,7 +3323,7 @@ static void handle_incoming_check(pj_ice_sess *ice,
     /* Check if ICE has been completed. We still need to check if remote 
      * candidate is valid when check_src_addr is set.
      */
-    if (ice->is_complete && !ice->opt.check_src_addr) {
+    if (ice->is_complete) {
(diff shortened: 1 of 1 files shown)


pjnath/src/pjnath/ice_session.c
```diff
@@ -3450,8 +3459,6 @@ static void handle_incoming_check(pj_ice_sess *ice,
          */
         c->nominated = ((rcheck->use_candidate) || c->nominated);
 
-        rcand->checked = PJ_TRUE;
-
```
nanangizz, line 3454: What is this for?

nanangizz:
So currently when the remote check comes after ICE nego is completed the `rcand->checked` will not be set (yet the remote check will still be responded). I think simply adding this info to docs/comment & log are sufficient to avoid confusion in the future. IMHO modifying the ICE flow is generally riskier. For example, log:
```
icetp00 !Ignoring incoming message for component [1] because source addr 10.1.0.155:4008 is unrecognized or unchecked.
```

nanangizz:
For future info, better also add comment about `checked` status in this code:
https://github.com/pjsip/pjproject/blob/948fd1d2fcaf93fadba5ce2329804b768486490a/pjnath/src/pjnath/ice_session.c#L3748-L3750

nanangizz (approved):

sauwming (approved):

Issue #957, opened 2009-08-27
SSL/secure socket abstraction and SIP TLS transport rewrite


pjsipbot:
_2009-08-27 20:43:40: @nanangizz_ _created the issue on trac ticket [957](https://trac.pjsip.org/repos/ticket/957)_

**Objective:**

  To create a portable SSL/secure socket abstraction in PJLIB and to rewrite the SIP TLS transport in PJSIP.

-*Specifications:**

   1. General:
     - supports client and server socket operations (note: Symbian's CSecureSocket only supports client sockets)
   1. I/O:
     - asynchronous I/O API similar to PJLIB's [Active Socket](http://www.pjsip.org/pjlib/docs/html/group__PJ__ACTIVESOCK.htm) API
     - timeout setting for SSL negotiation
     - buffer size control
   1. Options:
     - select SSL version to use (SSLv23, TLSv1, etc)
     - select ciphersuite to use
     - verify server's certificate (note: not available on Symbian)
     - server require client certificate (note: not available on Symbian)
     - verify client's certificate (note: not available on Symbian)
     - TLS server name extension to support connecting to multihosted TLS server (note: only available on recent OpenSSL versoins)
   1. Certificate management (note: not available on Symbian):
     - specify which certificate to use
     - load the certificate, either from file(s) or OS's certificate store.
     - retrieve certificate infos.

-*Backends:**

 1. OpenSSL
 1. CSecureSocket on Symbian

___
_2009-08-27 20:56:13: @nanangizz_ _commented_

In [r2913](../commit/6c62bf450d6aed6a6dcbbe38eec6db3819e0a2ef), initial version of TLS transport for Symbian, includes:
 - Secure socket, generic abstraction and Symbian implementation (using CSecureSocket).
 - Initial rewrite of SIP TLS transport.
 - Updated symbian_ua.mmp to support SIP transport TLS (experimental).

___
_2009-10-16 03:56:57: @nanangizz_ _changed **component** from **pjsip** to **common**_

___
_2009-10-16 03:56:57: @nanangizz_ _changed **title** from **TLS transport on Symbian** to **Implement SSL socket abstraction and rewrite SIP TLS transport.**_

___
_2009-10-16 04:07:41: @nanangizz_ _commented_

In [r2950](../commit/006cc01eeb8f4e9ad3f857d68313a718b9b93bc7):
 - Added SSL socket abstraction with OpenSSL backend (source only, i.e: hasn't been integrated to build settings).
[...]

PR #4325, opened 2025-02-26
Implement audio video synchronization

nanangizz:
#### For app using PJSUA-LIB
By default it is enabled for audio & video calls. It can be disabled per call basis by setting `PJSUA_CALL_NO_MEDIA_SYNC` in `pjsua_call_setting.flag`.

#### For app using PJMEDIA
1. App creates AV sync using `pjmedia_av_sync_create()`.
2. App adds all media to be synchronized using `pjmedia_av_sync_add_media()`. If the media is a audio/video stream, use `pjmedia_stream_common_set_avsync()` instead.
3. Each time a media receives an RTCP-SR, update AV sync using`pjmedia_av_sync_update_ref()`.
4. Each time a media returns a frame to be rendered, e.g: via `port.get_frame()`, update AV sync using `pjmedia_av_sync_update_pts()`, the function may request delay adjustment: increase or decrease delay.
5. Remove media from AV sync using `pjmedia_av_sync_del_media()`.
6. Destroy the AV sync using `pjmedia_av_sync_destroy()`.

#### For app using PJMEDIA AVI File Player
By default it is enabled. It can be disabled by setting `PJMEDIA_AVI_FILE_NO_SYNC` flag in creating AVI file player.

#### Global macro settings
Configurable via [`config_site.h`](https://docs.pjsip.org/en/latest/get-started/guidelines-development.html#config-site-h).
```
/**
 * Maximum tolerable presentation lag from the earliest to the latest media,
 * in milliseconds, in inter-media synchronization. When the delay is
 * higher than this setting, the media synchronizer will request the slower
 * media to speed up. And if after a number of speed up requests the delay
 * is still beyond this setting, the fastest media will be requested to
 * slow down.
 *
 * Default: 45 ms
 */
#ifndef PJMEDIA_AVSYNC_MAX_TOLERABLE_LAG_MSEC
#   define PJMEDIA_AVSYNC_MAX_TOLERABLE_LAG_MSEC    45
#endif


/**
  * Maximum number of speed up request to synchronize presentation time,
  * before a slow down request to the fastest media is issued.
  *
  * Default: 10
  */
#ifndef PJMEDIA_AVSYNC_MAX_SPEEDUP_REQ_CNT
#   define PJMEDIA_AVSYNC_MAX_SPEEDUP_REQ_CNT       10
#endif
```

## Logic in `pjmedia_av_sync_update_pts()`
- Calculate the absolute timestamp (or NTP timestamp) of the frame playback based on the reference NTP+RTP timestamp from RTCP-SR packet and the frame RTP timestamp from RTP packet. This timestamp is usually called presentation time (PTS).
- If the PTS is the largest, store it as AV sync's maximum PTS. Otherwise, calculate the difference or the delay of this PTS to the maximum PTS, then request the media to speed up as much as the delay.
- If after a specific number of speed up requests (i.e: configurable via `PJMEDIA_AVSYNC_MAX_SPEEDUP_REQ_CNT`) and the lag is still beyond a tolerable value (i.e: configurable via `PJMEDIA_AVSYNC_MAX_TOLERABLE_LAG_MSEC`), the function will issue slow down request to the fastest media (which set the AV sync's maximum PTS).
- To prevent unoptimized delay applied to all media for the synchronization, a mechanism is implemented where slow-down requests are marked down and speed-up requests are marked up.

### DIFF
--- a/pjmedia/include/pjmedia/av_sync.h
+++ b/pjmedia/include/pjmedia/av_sync.h
@@ -155,10 +155,10 @@ PJ_DECL(pj_status_t) pjmedia_av_sync_del_media(
 
 /**
  * Update synchronizer about the last presentation timestamp of the specified
- * media. Normally this function is called when media is producing a frame
- * to be rendered (e.g: in port's get_frame() method). Upon returning, the
- * media may be requested to adjust its delay so it matches to the earliest
- * or the latest media, i.e: by speeding up or slowing down.
+ * media. Normally this function is called each time the media produces
+ * a frame to be rendered (e.g: in port's get_frame() method). Upon returning,
+ * the media may be requested to adjust its delay so it matches to the
+ * earliest or the latest media, i.e: by speeding up or slowing down.
  *
  * Initially this function will try to request slower media to speed up.
  * If after a specific number of requests (i.e: configurable via
@@ -190,7 +190,8 @@ PJ_DECL(pj_status_t) pjmedia_av_sync_update_pts(
 
 /**
  * Update synchronizer about reference timestamps of the specified media.
- * Normally this function is called after a media receives RTCP SR packet.
+ * Normally this function is called each time the media receives RTCP SR
+ * packet.
  *
  * @param av_sync_media     The media synchronization handle.
  * @param ntp               The NTP timestamp info from RTCP SR.
--- a/pjmedia/include/pjmedia/av_sync.h
+++ b/pjmedia/include/pjmedia/av_sync.h
@@ -78,6 +78,25 @@ typedef struct pjmedia_av_sync_media pjmedia_av_sync_media;
 
 
 /**
+ * Synchronizer settings.
+ */
+typedef struct {
+    /**
+     * Name of the syncrhonizer
+     */
+    char                       *name;
+
+    /**
+     * Streaming mode. If set to PJ_TRUE, the delay adjustment values will
+     * be smoothened and marked up to prevent possible delay increase on
+     * all media.
+     */
+    pj_bool_t                   is_streaming;
+
+} pjmedia_av_sync_setting;
+
+
+/**
  * Media settings.
  */
 typedef struct {
@@ -87,16 +106,30 @@ typedef struct {
     char                       *name;
 
     /**
+     * Media type.
+     */
+    pjmedia_type                type;
+
+    /**
      * Media clock rate or sampling rate.
      */
     unsigned                    clock_rate;
+
 } pjmedia_av_sync_media_setting;
 
 
 /**
+ * Get default settings for synchronizer.
+ *
+ * @param setting           The synchronizer settings.
+ */
+PJ_DECL(void) pjmedia_av_sync_setting_default(
+                                pjmedia_av_sync_setting *setting);
+
+/**
  * Get default settings for media.
  *
- * @param setting           The media setting.
+ * @param setting           The media settings.
  */
 PJ_DECL(void) pjmedia_av_sync_media_setting_default(
                                 pjmedia_av_sync_media_setting *setting);
@@ -104,15 +137,15 @@ PJ_DECL(void) pjmedia_av_sync_media_setting_default(
 /**
  * Create media synchronizer.
  *
- * @param endpt             The media endpoint.
- * @param option            Synchronization option, must be NULL for now.
+ * @param pool              The memory pool.
+ * @param option            The synchronizer settings.
  * @param av_sync           The pointer to receive the media synchronizer.
  *
  * @return                  PJ_SUCCESS on success.
  */
 PJ_DECL(pj_status_t) pjmedia_av_sync_create(
-                                pjmedia_endpt *endpt,
-                                const void *option,
+                                pj_pool_t *pool,
+                                const pjmedia_av_sync_setting *setting,
                                 pjmedia_av_sync **av_sync);
 
 
@@ -125,10 +158,21 @@ PJ_DECL(void) pjmedia_av_sync_destroy(pjmedia_av_sync *av_sync);
 
 
 /**
+ * Reset synchronization states. Any existing media will NOT be removed,
+ * but their states will be reset.
+ *
+ * @param av_sync           The media synchronizer.
+ *
+ * @return                  PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t) pjmedia_av_sync_reset(pjmedia_av_sync *av_sync);
+
+
+/**
  * Add a media to synchronizer.
  *
  * @param av_sync           The media synchronizer.
- * @param setting           The media setting.
+ * @param setting           The media settings.
  * @param av_sync_media     The pointer to receive the media synchronization
  *                          handle.
  *
--- a/pjmedia/include/pjmedia/av_sync.h
+++ b/pjmedia/include/pjmedia/av_sync.h
@@ -20,7 +20,7 @@
 
  /**
   * @file av_sync.h
-  * @brief Media Presentation Synchronization.
+  * @brief Inter-media Synchronization.
   */
 #include <pjmedia/types.h>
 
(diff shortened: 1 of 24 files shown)


nanangizz:
> An early comment since it's still a draft.

Sure, actually early feedbacks is one of the goal of creating this draft.

> The term 'av' seems to imply the sync will only apply to audio and video only, even though it looks like the description can cover general cases. Especially with the upcoming text media, which may also need to be sync-ed (although the sync may not be available in the early version, but perhaps in the future).
> 
> Perhaps removing the `av` and call it `pjmedia_sync` should be sufficient?

Yes, the sync module is generic for any media, especially delivered via RTP/RTCP and using a shared/synchronized NTP source.

IMO the 'av_sync' prefix will enhance intuitiveness, i.e: it sounds related to presentation time synchronization. While 'sync' prefix may be somewhat ambiguous, e.g: what to sync?. I recall how 'ssl_sock' term was chosen over `tls_sock` for catchy/intuitiveness reason, when it was being developed, TLS was already around replacing SSL (even PJSIP transport was already using 'TLS', see also #957). What do you think?

sauwming:
The term should be inter-media synchronization, but yes, it's not quite as catchy, so I suppose it's okay if the 'av' stays.

pjmedia/src/pjmedia/av_sync.c
```diff
@@ -330,11 +326,15 @@ PJ_DEF(pj_int32_t) pjmedia_av_sync_update_pts(
         pj_sub_timestamp(&ntp_diff, &media->last_ntp);
         ms_diff = ntp_to_ms(&ntp_diff);
 
+        /* Smoothen and round down the delay */
+        ms_diff = ((ms_diff + 19 * media->smooth_diff) / 200) * 10;
```
wosrediinanatour, line 330: Should an upper bound considered? Where does the magic numbers come from?

nanangizz: Just added the upper bound 60s in the last commit, actually not really sure if it is needed.

The magic numbers are weighting for the smoothing, i.e: last delay uses weight of 19, newly received delay uses weight of 1, then divide the sum by 20 for the new delay. Instead of by 20, it is by 200 then multiplied to 10 for rounding down to nearest 10.

Thanks for the feedback.

nanangizz:
Ready to review perhaps :)

Currently the delay adjustment is implemented in audio stream only, because:
- video playback is kind of play as soon as possible (with built-in minimum delay for stability), the delay/burst/optimal-delay calculations in jitter buffer do not really apply (due to multiple RTP packets per video frame)
- it is easier to manage delay in audio as the jitter buffer already have info about delay & optimal delay
- should be sufficient for synchronization.

andreas-wehrmann:
> Especially with the upcoming text media, which may also need to be sync-ed (although the sync may not be available in the early version, but perhaps in the future).
> 

Sorry, quick question about that: Are there any specific plans to implement this (more specifically: RTT with T.140)?
I'm asking because I will need this later (probably this year) and was thinking about implementing it myself and maybe bring this upstream, unless there are any efforts underway already which I could support?

sauwming:
Yes, I'm currently implementing it. The basic functionality is already working (can send and receive text), but there's still a lot more to do, such as redundancy, docs, etc.

sauwming:
Not yet public. Will create a branch early next week.

pjmedia/include/pjmedia/av_sync.h
```diff
@@ -0,0 +1,206 @@
...
+ * More about #pjmedia_av_sync_update_pts():
+ * - The function will return non-zero when delay adjustment in the media is
+ *   needed for synchronization. When it returns a positive number, the media
+ *   should increase the delay as many as the returned number, in millisecond.
+ *   When it returns a negative number, the media should try to decrease
+ *   the delay as many as the returned number, in millisecond.
+ *   For example in audio stream, the delay adjustment can managed using
+ *   #pjmedia_jbuf_set_min_delay().
+ * - The function will keep track on the delay adjustment progress, for
+ *   example, after some delay adjustment are requested to a media and
```
sauwming, line 64: a little bit too vague.
how many are "some" delay adjustment? how much Is relatively high?
maybe provide the constants/settings that will decide this.

pjmedia/include/pjmedia/av_sync.h
```diff
@@ -0,0 +1,206 @@
...
+ * media. Normally this function is called when media is producing a frame
+ * to be rendered (e.g: to a video renderer or audio device speaker).
+ * The media may be requested to adjust its playback delay by the number
+ * returned by the function, for example audio stream can adjust delay
+ * using #pjmedia_jbuf_set_min_delay().
+ *
+ * @param av_sync_media     The media synchronization handle.
+ * @param pts               The presentation timestamp.
+ *
+ * @return                  0  : no action is needed, because the media is
```
sauwming, line 169: consider returning pj_status_t instead (and put the `adjustment` as `pj_int32_t *` output parameter).
Because if the return value is zero, it may not be obvious whether something is wrong or truly no adjustment is needed.

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -4384,6 +4390,16 @@ pj_status_t pjsua_media_channel_update(pjsua_call_id call_id,
     pj_memcpy(call->media, call->media_prov,
               sizeof(call->media_prov[0]) * call->med_prov_cnt);
 
+    /* Create synchronizer */
+    if ((maudcnt+mvidcnt) > 1 && !call->av_sync) {
```
sauwming, line 4394: Do we need pjsua config to enable/disable the sync, as well as to choose which media will be sync?

sauwming (commented):
CR first stage, I haven't checked av_sync.c

pjsip/src/pjsua-lib/pjsua_media.c
nanangizz: Yes, enable/disable should be needed. As this is a new feature, I guess we can postpone fine-tuning settings such as selecting which media to sync, allowing two or more synchronizers in a call, based on user feedback?

pjmedia/src/pjmedia/av_sync.c
```diff
@@ -0,0 +1,410 @@
...
+    } else {
+        /* Not the fastest. */
+        pj_timestamp ntp_diff = max_ntp;
+        unsigned ms_diff;
+
+        /* First, check the delay from the fastest. */
+        pj_sub_timestamp(&ntp_diff, &media->last_ntp);
+        ms_diff = ntp_to_ms(&ntp_diff);
+
+        /* Make sure the delay is sensible, e.g: not exceeding 60s */
```
sauwming, line 329: 1 minute seems super long, even 3-5 sec is already considered long in today's standard IMO.

And shouldn't the return status be ETOOBIG?

pjsip/src/pjsua-lib/pjsua_media.c
sauwming: The thing is, I'm not sure if text stream should be synchronized by default, so that's why I proposed the setting to let the user decide.

Having said that, I'm not sure whether we should also implement text sync, so I suppose enable/disable is sufficient for now, with doc noting that it currently only applies for audio and video.

pjmedia/src/pjmedia/av_sync.c
nanangizz: The delay/`ms_diff` is the lateness from the fastest, which eventually will trigger a speed-up request. If say the check is set to 5 seconds, then when a video is lagging 6 seconds behind the audio, speed-up request won't be triggered. Originally I doubt whether such boundary check is necessary.

When the media is lagging behind, it means that the supplied `pts` is too small (from the function caller PoV)?

sauwming:
> Currently the delay adjustment is implemented in audio stream only, because:
> * video playback is kind of play as soon as possible (with built-in minimum delay for stability), the delay/burst/optimal-delay calculations in jitter buffer do not really apply (due to multiple RTP packets per video frame)
> * it is easier to manage delay in audio as the jitter buffer already have info about delay & optimal delay
> * should be sufficient for synchronization.

Actually I was thinking the other way around. Because while it's true that video is initially played asap without buffering/prefetching, eventually it will be video that will be lagging due to its high bandwidth, and overall processing power required.
And when it happens (the video lags behind), then I imagine the mechanism would be to start dropping frames.

So perhaps it should be tested in such scenario, for example using HD video and/or minimal connection speed such as mobile data.

pjsip/src/pjsua-lib/pjsua_media.c
nanangizz: In the future, text sync may be useful for streaming a video with subtitle (may require something like text-player, similar to wav/video-player). However, for now I think having text sync is not so urgent.

Just to add a sample scenario of two or more synchronizers in a call, in a conference call, there can be two sets of audio+video stream (so totally there will be 4 streams), rather than synchronize all 4 streams using one synchronizer, perhaps it is better to employ a separate synchronizer for each set (due to different original sources, they may have different codec delay, network delay, NTP source, etc). So in the future, when designing fine tune settings per media, perhaps we need to also consider this scenario.

nanangizz:
> > Currently the delay adjustment is implemented in audio stream only, because:
> > 
> > * video playback is kind of play as soon as possible (with built-in minimum delay for stability), the delay/burst/optimal-delay calculations in jitter buffer do not really apply (due to multiple RTP packets per video frame)
> > * it is easier to manage delay in audio as the jitter buffer already have info about delay & optimal delay
> > * should be sufficient for synchronization.
> 
> Actually I was thinking the other way around. Because while it's true that video is initially played asap without buffering/prefetching, eventually it will be video that will be lagging due to its high bandwidth, and overall processing power required. And when it happens (the video lags behind), then I imagine the mechanism would be to start dropping frames.
> 
> So perhaps it should be tested in such scenario, for example using HD video and/or minimal connection speed such as mobile data.

Actually the description above is a bit misleading, it was supposed to explain that video stream does not have delay management and implementing it is not so simple for now because the jitter buffer does not really help. It needs to be implemented in the video stream itself from scratch. In audio, the optimal delay calculation is already done by jbuf and increasing delay can be easily inserted into jbuf.

Yes, in real world the video tends to lag behind the audio (from codecs & delivery aspects). That's why it is sufficient to have delay adjustment implemented only in audio, as mostly we need to add delay to audio (from its optimal delay).

Btw, adding delay to video is perhaps not so complex or risky, will try to implement it.

sauwming:
I'm not sure whether the video "delay" adjustment is completed, but it seems that the current patch can only add/decrease delay to the video? I believe a more realistic usage would be to speed up the video, i.e. by dropping or skipping frames.

In other words, instead of adding delay to audio, which will increase the latency/lag of the entire av streams, there can be another option for the user, which is to speed up the video, so the streams still feel like real time.

nanangizz:
> I'm not sure whether the video "delay" adjustment is completed, but it seems that the current patch can only add/decrease delay to the video? I believe a more realistic usage would be to speed up the video, i.e. by dropping or skipping frames.
> 
> In other words, instead of adding delay to audio, which will increase the latency/lag of the entire av streams, there can be another option for the user, which is to speed up the video, so the streams still feel like real time.

Previously I assume that video decoding is always done as fast as possible, no room for decreasing the delay. However, after implementing your idea here, the log showed some frames being skipped.

nanangizz:
Next, will try to integrate the AV sync to AVI player in the source side (for streaming & local playback). Currently the exising AVI player synchronization seems to be done in the rendering side (for local playback only?).

sauwming:
Right, currently sync is only done in aviplay.

sauwming (approved):

PR #4331, opened 2025-02-28
Fix crash due to accessing deleted authentication session parent

jwes:
that occurs when the app comes to the foreground and the IP connections were closed before, for example because the power saving mode was activated on the iPhone. The reason was that the client authentication sessions of the regc object still pointed to a parent client authentication session that had already been deleted from the account.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_reg.c
+++ b/pjsip/src/pjsip-ua/sip_reg.c
@@ -1222,6 +1222,9 @@ static void regc_tsx_callback(void *token, pjsip_event *event)
             }
         }
 
+        if (regc->_delete_flag != 0) {
+            regc->auth_sess.parent = NULL;
+        }
         status = pjsip_auth_clt_reinit_req( &regc->auth_sess,
                                             rdata, 
                                             tsx->last_tx,  
--- a/pjsip/src/pjsip-ua/sip_reg.c
+++ b/pjsip/src/pjsip-ua/sip_reg.c
@@ -96,7 +96,6 @@ struct pjsip_regc
 
     /* Authorization sessions. */
     pjsip_auth_clt_sess          auth_sess;
-    pjsip_auth_clt_sess         *ext_auth_sess; /**< User defined auth session.     */
 
     /* Auto refresh registration. */
     pj_bool_t                    auto_reg;
@@ -1223,7 +1222,7 @@ static void regc_tsx_callback(void *token, pjsip_event *event)
         }
 
         if (regc->_delete_flag != 0) {
-            regc->auth_sess.parent = NULL;
+            pjsip_auth_clt_set_parent(&regc->auth_sess, NULL);
         }
         status = pjsip_auth_clt_reinit_req( &regc->auth_sess,
                                             rdata, 
--- a/pjsip/src/pjsip/sip_auth_client.c
+++ b/pjsip/src/pjsip/sip_auth_client.c
@@ -907,17 +907,17 @@ PJ_DEF(pj_status_t) pjsip_auth_clt_clone( pj_pool_t *pool,
         }
     }
 
-    if (sess->parent) {
+    if (rhs->parent) {
         pj_status_t status;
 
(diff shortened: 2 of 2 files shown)


pjsip/src/pjsip-ua/sip_reg.c
```diff
@@ -1222,6 +1222,9 @@ static void regc_tsx_callback(void *token, pjsip_event *event)
             }
         }
 
+        if (regc->_delete_flag != 0) {
+            regc->auth_sess.parent = NULL;
```
nanangizz, line 1226: - What if resetting the parent is done somewhere inside `pjsip_regc_destroy2()`?
- How about dialogs (e.g: call & presence), should their auth session parents be reset too (when acc is deleted)?
- For consistency, as setting parent is done via `pjsip_auth_clt_set_parent()`, perhaps there should also be an API to reset parent, for example modify the `pjsip_auth_clt_set_parent()` to accept param `parent==NULL` so it can also be used to reset parent?

Btw, minor & rather OOT, field `pjsip_regc.ext_auth_sess` does not seem to be used anywhere, is it okay to remove it?

jwes: >     What if resetting the parent is done somewhere inside pjsip_regc_destroy2()?

pjsip_regc_destroy2() does reset the parent, because it calls         
pjsip_auth_clt_deinit(&regc->auth_sess); which resets the pointer to NULL.

>    How about dialogs (e.g: call & presence), should their auth session parents be reset too (when acc is deleted)?

If there are dialogs running while the account is deleted, it sure is necessary to reset the parent too.
But is that possible at all? The regc is called and held active in pjsua_acc_del, so this was (though hard to find) reasonable. But are there presence subscriptions or calls that remain active while the account is deleted?

>    For consistency, as setting parent is done via pjsip_auth_clt_set_parent(), perhaps there should also be an API to reset parent, for example modify the pjsip_auth_clt_set_parent() to accept param parent==NULL so it can also be used to reset parent?

that makes a lot of sense

> Btw, minor & rather OOT, field pjsip_regc.ext_auth_sess does not seem to be used anywhere, is it okay to remove it?

wasn't sure if this was the right spot, but i think you are right. If you don't mind i will address this in a follow-up commit in this PR?

nanangizz: > > ```
> > What if resetting the parent is done somewhere inside pjsip_regc_destroy2()?
> > ```
> 
> pjsip_regc_destroy2() does reset the parent, because it calls pjsip_auth_clt_deinit(&regc->auth_sess); which resets the pointer to NULL.

Ah okay, `pjsip_auth_clt_deinit()` is indeed called for immediate destroy case (no pending tsx), but not when there is pending tsx. Is resetting the parent when there is pending tsx considered unsafe (haven't checked myself)?

> 
> > How about dialogs (e.g: call & presence), should their auth session parents be reset too (when acc is deleted)?
> 
> If there are dialogs running while the account is deleted, it sure is necessary to reset the parent too. But is that possible at all? The regc is called and held active in pjsua_acc_del, so this was (though hard to find) reasonable. But are there presence subscriptions or calls that remain active while the account is deleted?

App must not delete an account being used by an active call/presence, but IIRC it is actually possible (and perhaps can cause a crash) as you can see the `pjsua_acc_del()` does not hang up call/presence. Anyway, I guess we can leave it as is for now.

> 
> > For consistency, as setting parent is done via pjsip_auth_clt_set_parent(), perhaps there should also be an API to reset parent, for example modify the pjsip_auth_clt_set_parent() to accept param parent==NULL so it can also be used to reset parent?
> 
> that makes a lot of sense
> 
> > Btw, minor & rather OOT, field pjsip_regc.ext_auth_sess does not seem to be used anywhere, is it okay to remove it?
> 
> wasn't sure if this was the right spot, but i think you are right. If you don't mind i will address this in a follow-up commit in this PR?

As it is a minor fix, it is okay.

nanangizz (approved):

PR #4333, opened 2025-03-03
Add support to new Android DNS functions android_res_nquery android_res_nresult (API 29+)

micha102:
Add support to new Android DNS functions android_res_nquery android_res_nresult (API 29+) in pjlib/src/pj/addr_resolv_sock.c
New macro
`#define PJ_GETADDRINFO_USE_ANDROID 1`

I looked at the [Android documentation](https://developer.android.com/ndk/reference/group/networking) to enhance default getaddrinfo (very slow in recent Android APIs), and found that a native method for DNS resolution is available starting Android API 29, based on android_res_nquery and android_res_nresult functions.

I implemented it and it's definitely faster than default getaddrinfo, the delay between clicking on "Start Library" and actual TLS socket establishment is next to nothing (110 ms)

### DIFF
--- a/pjlib/src/pj/addr_resolv_sock.c
+++ b/pjlib/src/pj/addr_resolv_sock.c
@@ -187,49 +187,40 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
     return status;
 
 #elif defined(PJ_GETADDRINFO_USE_ANDROID) && PJ_GETADDRINFO_USE_ANDROID!=0
-    /* Android-specific functions are available in API >= 29
-     * android_res_nquery, android_res_nresult and android_res_cancel
-     * If we directly call the functions, even surrounded by a statement
-     * checking Android API, the older Android devices will throw the error
-     * "dlopen failed: cannot locate symbol "android_res_nquery"
-     * Hence, we prepare pointers and load functions only if API check passes
+    /* Unlike Android API functions which can be safeguarded with
+     * __builtin_available, functions from libraries like libc.so
+     * cannot be loaded directly, even surrounded by a statement
+     * checking their availability. ns_initparse and ns_parserr are only
+     * available since Android API 23. This leads to failure of compiling
+     * for Android API 21. Hence, the two functions are dynamically loaded
+     * so that the library can be compiled with the minimal API 21.
+     * Otherwise, PJLIB would only compile with APP_PLATFORM=23 at minimum.
      */
 	 
-    /* Define function pointers type for Android DNS functions */
-    typedef int (*android_res_nquery_fn)(
-        unsigned int, const char*, int, int, int);
-    typedef int (*android_res_nresult_fn)(int, int*, unsigned char*, int);
-    typedef int (*android_res_cancel_fn)(int);
 
-    /* Initialize function pointers to NULL */
-    android_res_nquery_fn android_res_nquery_ptr = NULL;
-    android_res_nresult_fn android_res_nresult_ptr = NULL;
-    android_res_cancel_fn android_res_cancel_ptr = NULL;
-
-    /* Buffer to store the SDK version as a string */
-    char sdk_value[8] = "";
-    /* Get the Android SDK version */
-    __system_property_get("ro.build.version.sdk", sdk_value);
-
-    if (atoi(sdk_value) >= 29) {
-        /* Load the libandroid.so library containing network functions */
-        void *libandroid_handle = dlopen("libandroid.so", RTLD_NOW);
-        if (libandroid_handle) {
-            /* Get the nquery, nresult and cancel functions from library */
-            android_res_nquery_ptr = (android_res_nquery_fn)
-                                     dlsym(libandroid_handle,
-                                     "android_res_nquery");
-            android_res_nresult_ptr = (android_res_nresult_fn)
-                                     dlsym(libandroid_handle,
-                                     "android_res_nresult");
-            android_res_cancel_ptr = (android_res_cancel_fn)
-                                     dlsym(libandroid_handle,
-                                     "android_res_cancel");
+    if (__builtin_available(android 29, *)) {
+        /* Define function pointers type for ns_initparse and ns_parserr.
+         * These functions are in libc.so in Android (Bionic) and not
+         * in libresolv.so like usually in Linux */
+        typedef int (*ns_initparse_fn)(const unsigned char *, int, void *);
+        typedef int (*ns_parserr_fn)(const void *, int, int, void *);
+        /* Initialize function pointers to NULL */
+        ns_initparse_fn ns_initparse_ptr = NULL;
+        ns_parserr_fn ns_parserr_ptr = NULL;
+        /* Load the libc.so library containing network functions */
+        void *libc_handle = dlopen("libc.so", RTLD_NOW);
+        if (libc_handle) {
+            /* Get the ns_initparse, ns_initparse functions from library */
+            ns_initparse_ptr = (ns_initparse_fn)
+                               dlsym(libc_handle,
+                               "ns_initparse");
+            ns_parserr_ptr = (ns_parserr_fn)
+                             dlsym(libc_handle,
+                             "ns_parserr");
 
             /* Non-null pointers mean DNS functions are properly loaded */
-            if (android_res_nquery_ptr &&
-                android_res_nresult_ptr &&
-                android_res_cancel_ptr) {
+            if (ns_initparse_ptr &&
+                ns_parserr_ptr) {
                 pj_bzero(&hint, sizeof(hint));
                 hint.ai_family = af;
                 /* Zero value of ai_socktype means the implementation shall
@@ -246,14 +237,14 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
                 struct addrinfo *result = NULL, *last = NULL;
 
                 /* Step 1: Send DNS query */
-                int resp_handle = android_res_nquery_ptr(
+                int resp_handle = android_res_nquery(
                                 netid, nodecopy, ns_c_in,
                                 af == PJ_AF_INET ? ns_t_a :
                                 af == PJ_AF_INET6 ? ns_t_aaaa :
                                 ns_t_a,
                                 0);
                 if (resp_handle < 0) {
-                    printf("android_res_nquery_ptr() failed\n");
+                    printf("android_res_nquery() failed\n");
                     return PJ_ERESOLVE;
                 }
 
@@ -265,15 +256,15 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
 
                 if (ret <= 0) {
                     PJ_LOG(4,(THIS_FILE,"Poll timeout or error"));
-                    android_res_cancel_ptr(resp_handle);
+                    android_res_cancel(resp_handle);
                     return PJ_ERESOLVE;
                 }
                 /* Step 3: Get DNS response */
-                int response_size = android_res_nresult_ptr(
+                int response_size = android_res_nresult(
                                   resp_handle, &rcode, answer, sizeof(answer));
                 if (response_size < 0) {
                     PJ_LOG(4,(THIS_FILE,
-                              "android_res_nresult_ptr() failed"));
+                              "android_res_nresult() failed"));
                     return PJ_ERESOLVE;
                 }
 
@@ -282,7 +273,7 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
                 ns_rr rr;
                 int num_records, resolved_count = 0;
 
-                if (ns_initparse(answer, response_size, &msg) < 0) {
+                if (ns_initparse_ptr(answer, response_size, &msg) < 0) {
                     PJ_LOG(4,(THIS_FILE,
                               "Failed to parse DNS response"));
                     return PJ_ERESOLVE;
@@ -301,7 +292,7 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
 
                 /* Process each answer record */
                 for (i = 0; i < num_records && resolved_count < *count; i++) {
-                    if (ns_parserr(&msg, ns_s_an, i, &rr) < 0) {
+                    if (ns_parserr_ptr(&msg, ns_s_an, i, &rr) < 0) {
                         continue;
                     }
 
@@ -319,7 +310,7 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
                             struct sockaddr_in addr;
                             pj_bzero(&addr, sizeof(addr));
                             addr.sin_family = PJ_AF_INET;
-                            memcpy(&addr.sin_addr, rdata, 4);
+                            pj_memcpy(&addr.sin_addr, rdata, 4);
                             /* Copy the sockaddr into pj_addrinfo.ai_addr */
                             pj_memcpy(&ai[resolved_count].ai_addr,
                                       &addr, sizeof(addr));
@@ -328,7 +319,7 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
                             struct sockaddr_in6 addr6;
                             pj_bzero(&addr6, sizeof(addr6));
                             addr6.sin6_family = PJ_AF_INET6;
-                            memcpy(&addr6.sin6_addr, rdata, 16);
+                            pj_memcpy(&addr6.sin6_addr, rdata, 16);
                             pj_memcpy(&ai[resolved_count].ai_addr,
                                       &addr6, sizeof(addr6));
                         }
--- a/pjlib/src/pj/addr_resolv_sock.c
+++ b/pjlib/src/pj/addr_resolv_sock.c
@@ -23,19 +23,22 @@
 #include <pj/ip_helper.h>
 #include <pj/compat/socket.h>
 
-#if defined(PJ_GETADDRINFO_USE_ANDROID) && PJ_GETADDRINFO_USE_ANDROID!=0
-#include <sys/poll.h>
-#include <resolv.h>
-#include <netdb.h>
-#include <android/multinetwork.h>
-#include <sys/system_properties.h>  // For getting Android Device API Version
-#include <dlfcn.h>
-#endif
-
 #if defined(PJ_GETADDRINFO_USE_CFHOST) && PJ_GETADDRINFO_USE_CFHOST!=0
 #   include <CoreFoundation/CFString.h>
 #   include <CFNetwork/CFHost.h>
 #endif
+
+#elif defined(PJ_GETADDRINFO_USE_ANDROID) && PJ_GETADDRINFO_USE_ANDROID!=0
+#   include <sys/poll.h>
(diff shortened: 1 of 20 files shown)


micha102:
Need to implement the check on Android API Level too
If >= 29 then run the DNS lookup with android_res_nquery/android_res_nresult
if < 29 then run the DNS lookup with getaddrinfo

micha102:
I found it's trickier than I thought to "hide" the API 29+ functions from devices with API < 29.
If I just do a check on API Level, then older devices will throw "dlopen failed: cannot locate symbol "android_res_nquery" referenced by libpjsua2" even though it shouldn't see it (because of the if statement).

So, I dynamically loaded the functions android_res_nquery, android_res_nresult and android_res_cancel and did this only in case of API >= 29
Otherwise for older APIs, I cloned the same default getaddrinfo() to be used.

I tested it on :
Xiaomi Mi 9T with LineageOS API 35 arm64-v8a : works good with API 29+ functions
Emulated Pixel 6 Pro with API 35 x86_64 : works good with API 29+ functions
Emulated Pixel 6 Pro with API 32 x86_64 : works good with API 29+ functions
Emulated Pixel 6 Pro with API 28 x86 : works good with fallback getaddrinfo()

pjlib/src/pj/addr_resolv_sock.c
```diff
@@ -23,6 +23,15 @@
...
 
+#if defined(PJ_GETADDRINFO_USE_ANDROID) && PJ_GETADDRINFO_USE_ANDROID!=0
+#include <sys/poll.h>
+#include <resolv.h>
+#include <netdb.h>
+#include <android/multinetwork.h>
+#include <sys/system_properties.h>  // For getting Android Device API Version
+#include <dlfcn.h>
+#endif
+
```
sauwming, line 34: use `#elif` here, to make it obvious that only one condition can be satisfied

also indent the `#include`.

pjlib/src/pj/addr_resolv_sock.c
```diff
@@ -174,6 +183,200 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
...
+    android_res_nquery_fn android_res_nquery_ptr = NULL;
+    android_res_nresult_fn android_res_nresult_ptr = NULL;
+    android_res_cancel_fn android_res_cancel_ptr = NULL;
+    char sdk_value[8] = "";
+    __system_property_get("ro.build.version.sdk", sdk_value);
+    if (atoi(sdk_value) >= 29) {
+        void *libandroid_handle = dlopen("libandroid.so", RTLD_NOW);
+        if (libandroid_handle) {
+            android_res_nquery_ptr = (android_res_nquery_fn) dlsym(libandroid_handle, "android_res_nquery");
```
sauwming, line 200: minor: all col needs to be <= 80

pjlib/src/pj/addr_resolv_sock.c
```diff
@@ -174,6 +183,200 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
...
+                if (resp_handle < 0) {
+                    printf("android_res_nquery_ptr() failed\n");
+                    return PJ_ERESOLVE;
+                }
+
+                // Step 2: Wait for response using poll()
+                struct pollfd fds;
+                fds.fd = resp_handle;
+                fds.events = POLLIN;
+                int ret = poll(&fds, 1, 5000); // 5-second timeout
```
sauwming, line 230: create a compile-time macro for the timeout (under all the `#include` above), so user can easily modify it should they wish to.

pjlib/src/pj/addr_resolv_sock.c
```diff
@@ -174,6 +183,200 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
...
+            continue;
+
+        if (res->ai_socktype != pj_SOCK_DGRAM() &&
+            res->ai_socktype != pj_SOCK_STREAM() &&
+            /* It is possible that the result's sock type
+             * is unspecified.
+             */
+            res->ai_socktype != 0)
+            {
+            continue;
```
sauwming, line 340: indentation

pjlib/src/pj/addr_resolv_sock.c
```diff
@@ -174,6 +183,200 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
...
+                fds.fd = resp_handle;
+                fds.events = POLLIN;
+                int ret = poll(&fds, 1, 5000); // 5-second timeout
+
+                if (ret <= 0) {
+                    PJ_LOG(4,("addr_resolv_sock.c","Poll timeout or error"));
+                    android_res_cancel_ptr(resp_handle);
+                    return PJ_ERESOLVE;
+                }
+                // Step 3: Get DNS response
```
sauwming, line 237: Use /* */ for comment.
https://docs.pjsip.org/en/latest/get-started/coding-style.html#multiline-comment-indentation

pjlib/src/pj/addr_resolv_sock.c
micha102: Thanks for your review.
I put the statement below PJ_GETADDRINFO_USE_CFHOST and changed it to `#elif`.
I also indented the `#include`

pjlib/src/pj/addr_resolv_sock.c
micha102: I reviewed all the lines where columns were over 80 and added linSorry, I've just learnt this for C. Thank you.

pjlib/src/pj/addr_resolv_sock.c
micha102: I added a macro `ANDROID_DNS_TIMEOUT_MS` for this.
Also I created the macro `THIS_FILE` as in other source files, for logging usage.

pjlib/src/pj/addr_resolv_sock.c
micha102: I corrected the comments so that they match the requirement.

pjlib/src/pj/addr_resolv_sock.c
```diff
@@ -174,6 +183,200 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
...
+    typedef int (*android_res_nresult_fn)(int, int*, unsigned char*, int);
+    typedef int (*android_res_cancel_fn)(int);
+
+    android_res_nquery_fn android_res_nquery_ptr = NULL;
+    android_res_nresult_fn android_res_nresult_ptr = NULL;
+    android_res_cancel_fn android_res_cancel_ptr = NULL;
+    char sdk_value[8] = "";
+    __system_property_get("ro.build.version.sdk", sdk_value);
+    if (atoi(sdk_value) >= 29) {
+        void *libandroid_handle = dlopen("libandroid.so", RTLD_NOW);
```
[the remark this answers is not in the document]
micha102: Thanks for your review. I corrected it in the last commit and added a comment over that block to explain the reason why I used function pointers to load API 29+ functions only when Android API Level verification is done.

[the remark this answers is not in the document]
sauwming: What will happen if you link `-landroid` from the configure script instead?
Something like this:
https://github.com/pjsip/pjproject/blob/6dabbd5768dce7b2f5f68787e33afa6f701b6634/aconfigure.ac#L1211-L1213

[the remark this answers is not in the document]
micha102: I checked the output of  `TARGET_ABI=xxxxx ./configure-android --use-ndk-cflags $OTHER_OPTIONS` for my build and I saw that :

```
checking for linux/soundcard.h... yes
checking for machine/soundcard.h... no
Checking sound device backend... Android
**Checking if OpenGL ES 2 is available... yes**
Checking if small filter is disabled... no
Checking if large filter is disabled... no
...
```
So, the statement is actually matched for my build.

Do you mean I should test the verification of the function similar to this ?
https://github.com/pjsip/pjproject/blob/6dabbd5768dce7b2f5f68787e33afa6f701b6634/aconfigure.ac#L391-L396
This would be the same issue as the NDK 27c with Target API 35 will successfully detect the functions, but the Android client.
But an older Android device will find reference to those functions and thus fails to load the pjsua2.so library.

It's a known problem for NDK as compared to iOS to ensure backward compatiblity.
https://github.com/android/ndk/issues/837

The [Android NDK documentation](https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md) mentions :
> Only symbols are affected, not libraries. The only way to conditionally depend on a library that is not available in the app's minSdkVersion is with dlopen. We do not know how to solve this in a backwards compatible manner.

So I'm not sure if we can integrate this without dynamically loading functions (dlopen and dlsym)
This is discussed in https://github.com/android/ndk/issues/837 (previously neither symbols nor libraries were possible to be "catched" with an API check.
But now symbols (headers) are possible but not library.
I've checked the header files for NDK and I can't find any headers file for the DNS functions android_res_*, it's only in the library libandroid.so itself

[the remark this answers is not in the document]
micha102: I think my PR is stuffed with standard functions and I need to revamp it (like I should use optimized PJ types instead of regular C ones or functions like poll where pj_ioqueue_poll exists)
Your help is welcome either in the form or in the substance.

pjlib/src/pj/addr_resolv_sock.c
```diff
@@ -174,6 +185,241 @@ PJ_DEF(pj_status_t) pj_getaddrinfo(int af, const pj_str_t *nodename,
...
+                    if ((type == ns_t_a && rdlen == 4) || (type == ns_t_aaaa
+						&& rdlen == 16)) {
+
+                        /* For IPv4, fill a temporary sockaddr_in */
+                        /* For IPv6 fill a sockaddr_in6. */
+                        if (type == ns_t_a) {
+                            struct sockaddr_in addr;
+                            pj_bzero(&addr, sizeof(addr));
+                            addr.sin_family = PJ_AF_INET;
+                            memcpy(&addr.sin_addr, rdata, 4);
```
sauwming, line 322: pj_memcpy

pjlib/src/pj/addr_resolv_sock.c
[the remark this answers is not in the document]
sauwming: just a couple of minors from me (pj_memcpy and indentation).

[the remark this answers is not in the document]
micha102: I'm trying to compile with weak symbols but so far I'm failing (still having the dlopen error when running on older Android).
I need to tweak configure-android script.

Weak symbols would make the code "prettier" but I don't really see it to be used elsewhere apart from system features. (Like DNS which is subject of the PR)
I would say if things can be done in Android code then it must be done inside the Android Code and not in pj library.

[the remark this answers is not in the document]
micha102: I succeeded enabling weak symbols in PJ build.
In fact, I needed to add `ENABLE_WEAK_SYMBOLS="-D__ANDROID_UNAVAILABLE_SYMBOLS_ARE_WEAK__ -Werror=unguarded-availability"`
and then 
 `export CFLAGS="${NDK_CFLAGS} ${CFLAGS} ${CPPFLAGS} ${ENABLE_WEAK_SYMBOLS}"`

However, I realized there're some issues with the code :
1. I saw in the comments of configure-android, that the APP_PLATFORM will select the latest available Android API Version.
This is a contradiction with [Android recommendation](https://developer.android.com/ndk/guides/application_mk)  which states :

> APP_PLATFORM declares the Android API level this application is built against and corresponds to the application's minSdkVersion.
> If not specified, ndk-build will target the minimum API level supported by the NDK. The minimum API level supported by the latest NDK will always be low enough to support nearly all active devices.

Yet, in PJ Android build, it's the latest API level used (not the earliest as recommended by NDK)
Using the latest API Level will, of course, contains the latest APIs (functions) provided by Android.

2. When I forced APP_PLATFORM to be the minimum (in current LTS Android NDK 27c), the minimum is Android API 21), I can see errors that android_res_nquery is only available in API Level 29 and I should surround newer functions with an
if (__builtin_available(android 29, *)) {
...
}

But I also found usage of functions from API Level 28 and above :
```
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/xxxx/pjsip-android/pjproject-2.15.1/android-ndk-r27c/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/media/NdkMediaCodec.h:499:16: note: 'AMediaCodec_setAsyncNotifyCallback' has been marked as being introduced in Android 28 here, but the deployment target is Android 21
  499 | media_status_t AMediaCodec_setAsyncNotifyCallback(
      |                ^
../src/pjmedia-codec/and_aud_mediacodec.cpp:1096:5: note: enclose 'AMediaCodec_setAsyncNotifyCallback' in a __builtin_available check to silence this warning
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/pjmedia-codec/and_aud_mediacodec.cpp:1097:5: error: 'AMediaCodec_setAsyncNotifyCallback' is only available on Android 28 or newer [-Werror,-Wunguarded-availability]
 1097 |     AMediaCodec_setAsyncNotifyCallback(codec_data->dec, async_cb, codec_data);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/xxxx/pjsip-android/pjproject-2.15.1/android-ndk-r27c/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/media/NdkMediaCodec.h:499:16: note: 'AMediaCodec_setAsyncNotifyCallback' has been marked as being introduced in Android 28 here, but the deployment target is Android 21
  499 | media_status_t AMediaCodec_setAsyncNotifyCallback(

```
This is caused by PR #4105 

So, in my opinion :
1. We can keep strong symbols, and implement functions with dlopen/dlsym (as it is now in the PR)
2. We enable weak symbols, and instead of the code

`   /* Initialize function pointers to NULL */
    android_res_nquery_fn android_res_nquery_ptr = NULL;
    android_res_nresult_fn android_res_nresult_ptr = NULL;
    android_res_cancel_fn android_res_cancel_ptr = NULL;

    /* Buffer to store the SDK version as a string */
    char sdk_value[8] = "";
    /* Get the Android SDK version */
    __system_property_get("ro.build.version.sdk", sdk_value);

    if (atoi(sdk_value) >= 29) {
        /* Load the libandroid.so library containing network functions */
        void *libandroid_handle = dlopen("libandroid.so", RTLD_NOW);
        if (libandroid_handle) {
            /* Get the nquery, nresult and cancel functions from library */
            android_res_nquery_ptr = (android_res_nquery_fn)
                                     dlsym(libandroid_handle,
                                     "android_res_nquery");
            android_res_nresult_ptr = (android_res_nresult_fn)
                                     dlsym(libandroid_handle,
                                     "android_res_nresult");
            android_res_cancel_ptr = (android_res_cancel_fn)
                                     dlsym(libandroid_handle,
                                     "android_res_cancel");

            /* Non-null pointers mean DNS functions are properly loaded */
            if (android_res_nquery_ptr &&
                android_res_nresult_ptr &&
                android_res_cancel_ptr) {
                                [...]
            } else {
                do the getaddrinfo stuff }
```

`

which are 11 lines of code, we will simply have :

```
if (__builtin_available(android 29, *)) {
...
} else {
do the getaddrinfo stuff }
```

which is cleaner and prettier.
Should I go with 1 or 2 ?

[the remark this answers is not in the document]
sauwming: I vote for 2.
That's what we usually do in other platforms (such as Mac and iOS). It will be a hassle to do approach no 1 for every new-ish API that we're going to use.

micha102:
I did commits for :

configure-android
1. Add the SDK level detection for latest NDK (starting NDK 22 the folder "platforms" [disappeared](https://github.com/android/ndk/wiki/Changelog-r21/f9de9fd32d81b9a22d1cb45e6a601657abf40472))
2. Set the platform to minimum (which is 21 in latest LTS NDK 27c)
3. Enabled weak symbols inside `configure-android`

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp and and_aud_mediacodec.cpp
1. Surround the AMediaCodec_setAsyncNotifyCallback (API 28+) with availability check

addr_pjlib/src/pj/addr_resolv_sock.c
1. Surround the new Android DNS resolution with availability check
2. I noticed that the functions ns_initparse and ns_parserr are only available [starting API 23+](https://android.googlesource.com/platform/bionic/+/2b49904/docs/status.md)

> New libc functions in M (API level 23):
> ...
> re-introduced various <resolv.h> functions: ..., ns_initparse, ..., ns_parserr

And with this, I got a problem because, unlike Android API functions which are handled by the __builtin_available, [the functions from libc cannot be handled](https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#Weak-symbols-for-API-definitions), even never called (it's already inside an if statement requiring API 29!)

> Known issues and limitations:
> APIs in bionic (libc, libm, libdl) are not currently supported. See the bug for more information. If the source compatibility issues can be resolved, that will change in a future NDK release.

So I had two choices :
- Make the code only buildable for Android API 23 and above (APP_PLATFORM=23)
- Dynamically load ns_initparse and ns_parserr which are anyway loaded and called only when API is 29+ (a constraint even harder than the API 23)


I'm ready for code review.

sauwming:
It looks okay to me, but I would prefer if you put the changes unrelated to the title (i.e. unrelated to Android DNS android_res_nquery()) in a separate PR. So, the weak symbols is okay to be put in this PR, but the min platform and Android codec should be put in another PR.

micha102:
I created two additional pull requests.
one for seeking the proper directory for newer NDK (r21+) and select minimum SDK in configure-android.
One to surround the API 28+ functions with an availability check.

sauwming (approved):

nanangizz (approved):

sauwming:
The PR has been approved, but please remove the patch unrelated to this PR so I can merge this.

sauwming:
Do you need enabling weak symbols to be included in this PR?

micha102:
Yes I need it but it's already implemented by PR #4364 that's why I removed it from this PR.
Or does my revert actually removes the changes from `configure-android` ?
In this PR, I neither need to add nor modify anything in the `configure-android` file.

trengginas (approved):

sauwming:
PR #4364 doesn't have the weak symbols

micha102:
My bad, thanks for the heads-up.
I got confused between the PRs, I added a commit to enable the weak symbols.

bennylp:
Just as additional info for people who arrive here searching for slow/blocking `getaddrinfo()` problem, PJSIP has [an asynchronous DNS resolver implementation](https://docs.pjsip.org/en/latest/api/pjlib-util/index.html#dns), similar to [c-ares](https://github.com/c-ares/c-ares). It can be activated by setting the `nameserver` fields:

- in [UaConfig.nameserver](https://docs.pjsip.org/en/latest/api/generated/pjsip/group/group__PJSUA2__UA.html#_CPPv4N2pj8UaConfig10nameserverE) field (PJSUA2)
- in [pjsua_config.nameserver](https://docs.pjsip.org/en/latest/api/generated/pjsip/group/group__PJSUA__LIB__BASE.html#_CPPv4N12pjsua_config10nameserverE) field (PJSUA LIB)

PR #4337, opened 2025-03-05
Order route set based on dialogs role

jwes:
RFC 3261 specifies, that:

>  The route set MUST be set to the list of URIs in the Record-Route
>  header field from the request, taken in order and preserving all URI
>  parameters.

-- section 12.1.1 UAS behavior

also:

>  The route set MUST be set to the list of URIs in the Record-Route
>  header field from the response, taken in reverse order and preserving
>  all URI parameters.

-- section 12.1.2 UAC behavior

Currently, just the UAC behavior was taken into account backward iterating the list.

While still continuing to iterate backwards, the route is appended in front if the role is UAS.

### DIFF
--- a/pjsip/src/pjsip/sip_dialog.c
+++ b/pjsip/src/pjsip/sip_dialog.c
@@ -1984,7 +1984,11 @@ static void dlg_update_routeset(pjsip_dialog *dlg, const pjsip_rx_data *rdata)
             pjsip_route_hdr *r;
             r = (pjsip_route_hdr*) pjsip_hdr_clone(dlg->pool, hdr);
             pjsip_routing_hdr_set_route(r);
-            pj_list_push_back(&dlg->route_set, r);
+            if (dlg->role == PJSIP_ROLE_UAC) {
+                pj_list_push_back(&dlg->route_set, r);
+            } else {
+                pj_list_push_front(&dlg->route_set, r);
+            }
         }
     }
 


sauwming:
Checking the code, `dlg_update_routeset()` is only called from `pjsip_dlg_on_rx_response()`, and only if `dlg->role == PJSIP_ROLE_UAC`. 

In fact, there's an assertion just a few lines above the patch: 
```
if (dlg->role == PJSIP_ROLE_UAC) {
...
} else {
    pj_assert(!"Should not happen");
}
```

jwes:
pjsip_method_creates_dialog does only return true on UAC's?

jwes:
so the patch would be to return early? 
my coworker modified the code to store it (like in the RFC mentioned) in the incoming order.

unfortunately he is on vacation right now, so i can't ask him for the exact scenario where that was necessary.

i can just guess, that it might have something to do with early updates in a call (or something similar).

What is the best approach here?

close the PR and get another solution?
It seems to be working for me, and it also keeps the behavior of my coworker intact.
so maybe remove the assert and expect it for UAS also?

it is also ok for my to wait and add more information when he returns in ~3 weeks.

sauwming:
Since the patch will have no effect, I suggest we close the PR for now. When you have an update about the scenario, feel free to reopen it.
In the meantime, you can continue the discussion via our support mail to figure out a proper solution if needed.

PR #4338, opened 2025-03-05
Avoid 100rel retransmit if invite transaction no longer exists

PesalaDeSilva:
Also a couple of misc fixes included in this PR:
- Add assertion in os_core_win32 `pj_thread_local_get()` if the thread index is invalid.
- Add memory allocation check in pjmedia circular buffer.

### DIFF
--- a/pjlib/src/pj/os_core_win32.c
+++ b/pjlib/src/pj/os_core_win32.c
@@ -978,6 +978,10 @@ PJ_DEF(void*) pj_thread_local_get(long index)
     //Can't check stack because this function is called
     //by PJ_CHECK_STACK() itself!!!
     //PJ_CHECK_STACK();
+    if (index == -1)
+    {
+        return 0;
+    }
 #if defined(PJ_WIN32_WINPHONE8) && PJ_WIN32_WINPHONE8
     return TlsGetValueRT(index);
 #else
--- a/pjlib/src/pj/os_core_win32.c
+++ b/pjlib/src/pj/os_core_win32.c
@@ -978,10 +978,7 @@ PJ_DEF(void*) pj_thread_local_get(long index)
     //Can't check stack because this function is called
     //by PJ_CHECK_STACK() itself!!!
     //PJ_CHECK_STACK();
-    if (index == -1)
-    {
-        return 0;
-    }
+    PJ_ASSERT_RETURN(index >= 0, NULL);
 #if defined(PJ_WIN32_WINPHONE8) && PJ_WIN32_WINPHONE8
     return TlsGetValueRT(index);
 #else
--- a/pjsip/src/pjsip-ua/sip_100rel.c
+++ b/pjsip/src/pjsip-ua/sip_100rel.c
@@ -581,7 +581,7 @@ static void on_retransmit(pj_timer_heap_t *timer_heap,
         /* Clear all pending responses */
         clear_all_responses(dd);
 
-        /* Send 500 response */
+        /* Send 504 (Server Time-out) response */
         status = pjsip_inv_end_session(dd->inv, 504, &reason, &tdata);
         if (status == PJ_SUCCESS && tdata) {
             pjsip_dlg_send_response(dd->inv->dlg, 
--- a/pjsip/src/pjsip-ua/sip_100rel.c
+++ b/pjsip/src/pjsip-ua/sip_100rel.c
@@ -582,7 +582,7 @@ static void on_retransmit(pj_timer_heap_t *timer_heap,
         clear_all_responses(dd);
 
         /* Send 500 response */
-        status = pjsip_inv_end_session(dd->inv, 500, &reason, &tdata);
+        status = pjsip_inv_end_session(dd->inv, 504, &reason, &tdata);
         if (status == PJ_SUCCESS && tdata) {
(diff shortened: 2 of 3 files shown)


pjsip/src/pjsip-ua/sip_100rel.c
```diff
@@ -582,7 +582,7 @@ static void on_retransmit(pj_timer_heap_t *timer_heap,
         clear_all_responses(dd);
 
         /* Send 500 response */
```
sauwming, line 584: Also change the comment to: `Send 504 (Server Time-out) response`

pjsip/src/pjsip-ua/sip_100rel.c
```diff
@@ -608,7 +608,10 @@ static void on_retransmit(pj_timer_heap_t *timer_heap,
             return;
         }
     } else {
-        pjsip_tsx_retransmit_no_state(dd->inv->invite_tsx, tdata);
+        if (dd->inv->invite_tsx)
```
sauwming, line 611: Perhaps you can explain the reason behind this change?
i.e. What was the original problem that was fixed (crash, assertion, etc)? if there's any supporting PJSIP log, you can also attach it.

pjlib/src/pj/os_core_win32.c
```diff
@@ -978,6 +978,10 @@ PJ_DEF(void*) pj_thread_local_get(long index)
     //Can't check stack because this function is called
     //by PJ_CHECK_STACK() itself!!!
     //PJ_CHECK_STACK();
+    if (index == -1)
```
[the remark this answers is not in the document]
PesalaDeSilva: resolved with this optimization

pjsip/src/pjsip-ua/sip_100rel.c
PesalaDeSilva: resolved with the modification

pjsip/src/pjsip-ua/sip_100rel.c
PesalaDeSilva: Originally, the issue arose because dd->inv->invite_tsx was assumed to be valid at all times, but there were cases where it could be null, causing the function call to operate on an invalid pointer. By introducing a check (if (dd->inv->invite_tsx)), we ensure that retransmission only occurs when the transaction is valid. This prevents potential crashes and ensures that tdata is not kept unnecessarily when there is no valid transaction to handle it.

nanangizz: Ideally we investigate (using PJSIP log, etc) the conditions causing `dd->inv->invite_tsx` becoming null and fix accordingly. Adding such check is generally not safe from race conditions, e.g: not-null upon checking, but becoming null immediately right after.

Attached patch is perhaps a bit improvement, safer from the above race conditions, not tested though.
[add_inv_tsx_check_in_100rel.diff.txt](https://github.com/user-attachments/files/19175652/add_inv_tsx_check_in_100rel.diff.txt)

wosrediinanatour:
Could you change the title....? It does not say anything about the topic.

pjsip/src/pjsip-ua/sip_100rel.c
PesalaDeSilva: Thanks. Resolved the fix with the suggested optimizations

pjsip/src/pjsip-ua/sip_100rel.c
```diff
@@ -608,7 +615,7 @@ static void on_retransmit(pj_timer_heap_t *timer_heap,
             return;
         }
     } else {
-        pjsip_tsx_retransmit_no_state(dd->inv->invite_tsx, tdata);
+            pjsip_tsx_retransmit_no_state(dd->inv->invite_tsx, tdata);
```
sauwming, line 618: Minor: incorrect indentation

sauwming (commented):
The patch looks okay to me, only minor indentation.

And just a couple of comments:
- For future PRs, ideally it would be better to separate the PRs for different fixes.
- Please sign the CLA. Thanks.

pjsip/src/pjsip-ua/sip_100rel.c
sauwming: Just realized that the PR is slightly different from @nanangizz's original patch.
Once you assign `invite_tsx = dd->inv->invite_tsx`, you're supposed to use the local variable `invite_tsx` instead of directly accessing `dd->inv->invite_tsx`. This way, even if `dd->inv->invite_tsx` is already NULL-ed, as long as the invite transaction is still not destroyed, the app won't crash.

PesalaDeSilva: yes. thank you for the comment I just going to change the code with these   if (dd->uas_state->retransmit_count == 1) {
      pj_status_t status;
      
      status = pjsip_tsx_send_msg(invite_tsx, tdata);
      if (status != PJ_SUCCESS) {
          PJ_PERROR(3, (THIS_FILE, status,
                       "Failed to send message"));
          return;
      }
  } else {
      pjsip_tsx_retransmit_no_state(invite_tsx, tdata);
  }

sauwming (approved):

PR #4340, opened 2025-03-06
Cancel timer of sound device idle check

nanangizz:
In pjsua app, after making a call and quit, there is this log lines:
```
   timer.c  .Dumping timer heap:
   timer.c  .  Cur size: 1 entries, max: 3070
   timer.c  .  Entries:
   timer.c  .    _id   Id      Elapsed Source
   timer.c  .    ----------------------------------
   timer.c  .    7     1       0.000   pjsip\src\pjsua-lib\pjsua_aud.c:495
```
It turns out to be sound device idle timer.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_aud.c
+++ b/pjsip/src/pjsua-lib/pjsua_aud.c
@@ -2156,7 +2156,7 @@ static void close_snd_dev(void)
     pj_log_push_indent();
 
     /* Cancel sound device idle timer. */
-    if (0 && pjsua_var.snd_idle_timer.id) {
+    if (pjsua_var.snd_idle_timer.id) {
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsua-lib/pjsua_aud.c
```diff
@@ -2155,6 +2155,12 @@ static void close_snd_dev(void)
 {
     pj_log_push_indent();
 
+    /* Cancel sound device idle timer. */
+    if (0 && pjsua_var.snd_idle_timer.id) {
```
sauwming, line 2159: if 0?

nanangizz: Was generating log to show the problem, but then forgot to revert it.

sauwming (approved):

PR #4343, opened 2025-03-10
Delay video encoding start to suppress early RTP packet lost

nanangizz:
Currently in a video call start, frequently there is early keyframe request. After investigation, it seems that some initial RTP packets are actually lost indeed. Delaying the encoding seems able to suppress it.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_vid.c
+++ b/pjsip/src/pjsua-lib/pjsua_vid.c
@@ -1219,8 +1219,18 @@ pj_status_t pjsua_vid_channel_update(pjsua_call_media *call_med,
             pjmedia_stream_common *c_strm = (pjmedia_stream_common*)
                                             call_med->strm.v.stream;
             pj_grp_lock_add_ref(c_strm->grp_lock);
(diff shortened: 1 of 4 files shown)


pjsip/src/pjsua-lib/pjsua_vid.c
```diff
@@ -1194,6 +1205,24 @@ pj_status_t pjsua_vid_channel_update(pjsua_call_media *call_med,
...
+        if (status != PJ_SUCCESS)
+            goto on_error;
+
+        /* Schedule start stream encoding */
+        if (acc->cfg.vid_out_auto_transmit) {
+            pjmedia_stream_common *c_strm = (pjmedia_stream_common*)
+                                            call_med->strm.v.stream;
+            pj_grp_lock_add_ref(c_strm->grp_lock);
+            pjsua_schedule_timer2(&timer_on_start_vid_stream_encoding,
+                                  call_med->strm.v.stream, 500);
```
sauwming, line 1223: as usual, configurable is better. at least local configurable, with 0 meaning disabled/old behavior.

sauwming (approved):

PR #4344, opened 2025-03-10
Support real-time text (RFC4103)

sauwming:
This PR encompasses the following modifications:
* `pjmedia`:
 Add a new media stream: `pjmedia_txt_stream`.

* `pjsua-lib` and `pjsua2`:
  - Add `pjsua_call_setting.txt_cnt`/`CallSetting.textCount` to create calls with text streams.
  - Add `pjsua_acc_config.txt_red_level`/`AccountTextConfig.redundancyLevel` to configure text streams' redundancy levels (default: 2 as recommended by the RFC).
  - Add API `pjsua_call_send_text()`/`Call::sendText()` to send real-time text via RTP.
  - Add callback `on_call_rx_text()`/`Call:onCallRxText()` to receive notifications of incoming text.

This PR also modifies the default value of `pjsua_call_setting.flag`/`CallSetting.flag` to 0 (previously `PJSUA_CALL_INCLUDE_DISABLED_MEDIA`). The reason is that for typical audio-only calls, operating with 3 medias with 2 being disabled seems like slightly superfluous and inefficient.


nanangizz:
Just an idea, when designing modules & APIs, better also consider adding a conferencing feature in the future, preferably using a mechanism similar to other media, e.g: connecting/disconnecting ports, has source/sink port if applicable (sink port is perhaps like a log writer callback).

sauwming:
> Just an idea, when designing modules & APIs, better also consider adding a conferencing feature in the future, preferably using a mechanism similar to other media, e.g: connecting/disconnecting ports, has source/sink port if applicable (sink port is perhaps like a log writer callback).

OK, will try to incorporate this. It will become more massive than it currently is, fortunately the stream common already has the media ports.

sauwming:
One problem is that just like the existing media stream implementation before refactoring, there's currently already a lot of overlap between audio and video conference, so adding one more duplication is going to cause quite a headache.
Will probably need to refactor it and create `pjmedia_conf_common`, but with the upcoming parralelization, this will be no easy task.

andreas-wehrmann:
> @andreas-wehrmann , as promised, here's the draft PR for RTT support. Please let us know if you have any comments or feedback, thanks.

Thanks for the heads up; I'll try to find some time in the next couple of days to take a look!

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -0,0 +1,988 @@
...
+        c_strm->cname.slen = p - c_strm->cname.ptr;
+    }
+
+    /* Create mutex to protect jitter buffer: */
+    status = pj_mutex_create_recursive(pool, NULL, &c_strm->jb_mutex);
+    if (status != PJ_SUCCESS)
+        goto err_cleanup;
+
+    /* Create jitter buffer */
+    status = pjmedia_jbuf_create(pool, &c_strm->port.info.name,
```
andreas-wehrmann, line 261: This causes an immediate floating point exception on my machine because ´stream->buf_time´ is zero which leads to a div by zero:

```
12:29:54.677          pjsua_media.c  .....Media stream call00:1 is destroyed
12:29:54.677      udp0x7fffec021da0  ....UDP media transport started
12:29:54.677            pjsua_txt.c  ....Text channel update for index 1 for call 0...

Thread 1 "pjsua-x86_64-un" received signal SIGFPE, Arithmetic exception.
0x000055555567ee95 in pjmedia_jbuf_create (pool=0x555555906a60, name=0x5555559072b0, frame_size=64, ptime=0, max_count=8, p_jb=0x555555907390) at ../src/pjmedia/jbuf.c:617
617         jb->jb_min_shrink_gap= PJMEDIA_JBUF_DISC_MIN_GAP / ptime;
(gdb) bt
#0  0x000055555567ee95 in pjmedia_jbuf_create (pool=0x555555906a60, name=0x5555559072b0, frame_size=64, ptime=0, max_count=8, p_jb=0x555555907390) at ../src/pjmedia/jbuf.c:617
#1  0x000055555566f88e in pjmedia_txt_stream_create (endpt=0x5555558c8a30, pool=0x555555906a60, info=0x7fffffffb100, tp=0x7fffec025418, user_data=0x0, p_stream=0x5555558111c8 <pjsua_var+58760>) at ../src/pjmedia/txt_stream.c:261
#2  0x00005555555c03d5 in pjsua_txt_channel_update (call_med=0x555555811160 <pjsua_var+58656>, tmp_pool=0x7fffec00ed40, si=0x7fffffffb100, local_sdp=0x5555558efa38, remote_sdp=0x7fffec011790) at ../src/pjsua-lib/pjsua_txt.c:227
#3  0x00005555555af115 in apply_med_update (call_med=0x555555811160 <pjsua_var+58656>, local_sdp=0x5555558efa38, remote_sdp=0x7fffec011790, need_renego_sdp=0x7fffffffca20) at ../src/pjsua-lib/pjsua_media.c:4266
#4  0x00005555555af9f0 in pjsua_media_channel_update (call_id=0, local_sdp=0x5555558efa38, remote_sdp=0x7fffec011790) at ../src/pjsua-lib/pjsua_media.c:4489
#5  0x000055555559778a in pjsua_call_on_media_update (inv=0x7fffec010438, status=0) at ../src/pjsua-lib/pjsua_call.c:5571
#6  0x00005555555c528d in inv_negotiate_sdp (inv=0x7fffec010438) at ../src/pjsip-ua/sip_inv.c:2275
#7  0x00005555555c5d2b in process_answer (inv=0x7fffec010438, st_code=200, tdata=0x5555558edb00, local_sdp=0x0) at ../src/pjsip-ua/sip_inv.c:2614
#8  0x00005555555c6290 in pjsip_inv_answer (inv=0x7fffec010438, st_code=200, st_text=0x0, local_sdp=0x0, p_tdata=0x7fffffffcc78) at ../src/pjsip-ua/sip_inv.c:2788
#9  0x0000555555591dda in pjsua_call_answer2 (call_id=0, opt=0x555555801660 <call_opt>, code=200, reason=0x0, msg_data=0x7fffffffcd30) at ../src/pjsua-lib/pjsua_call.c:3062
#10 0x000055555557b59d in ui_answer_call () at ../src/pjsua/pjsua_app_legacy.c:930
#11 0x000055555557df7b in legacy_main () at ../src/pjsua/pjsua_app_legacy.c:1946
#12 0x000055555556d4bd in pjsua_app_run (wait_telnet_cli=1) at ../src/pjsua/pjsua_app.c:2122
#13 0x0000555555568f78 in main_func (argc=2, argv=0x7fffffffdcc8) at ../src/pjsua/main.c:137
#14 0x000055555573626f in pj_run_app (main_func=0x555555568ef9 <main_func>, argc=2, argv=0x7fffffffdcc8, flags=0) at ../src/pj/os_core_unix.c:2165
#15 0x0000555555568ff1 in main (argc=2, argv=0x7fffffffdcc8) at ../src/pjsua/main.c:158
```

sauwming: Thanks for catching this. Should be fixed in the latest commit.

andreas-wehrmann:
@sauwming Thanks for this work! I'm still trying to get to this, but it's more like on-off because I keep getting dragged to other things.
What I noticed today was that the SDP answer (with RED enabled) _always_ indicates a RED level of 2.
- It didn't matter what RED level I set in the PJSUA test app
- Maybe add an option to make it so the own RED level is taken from the SDP offer? Maybe re-use the setting from the negotiator, which indicates whether or not to prefer the remote/local settings? See: [pjmedia_sdp_neg_set_prefer_remote_codec_order()](https://github.com/pjsip/pjproject/blob/76bd3dbf37d7d69ff47dc777f4db9a2a8329bca6/pjmedia/include/pjmedia/sdp_neg.h#L415)
Though this might cause some confusion, because this is not stricly about the "order" but rather the preferred RED level.

sauwming:
Yes, I'm currently working on it and it turns out to be more complicated than I initially expected.
The problem is that the local SDP has been generated (using call default settings) before the call is even answered, so user's red level settings won't take effect.

I have also updated the PR description with the known issues.

sauwming:
Ready for review.
Only left to do is the porting of txt account config to pjsua2, but this can follow later if everything looks okay.

pjmedia/include/pjmedia/endpoint.h
```diff
@@ -72,6 +72,16 @@ typedef struct pjmedia_endpt_create_sdp_param
...
 
+    /**
+     * Specifies whether redundancy should be offered in the SDP.
+     * Value is integer, indicating redundancy levels (0 means disabled/
+     * no redundancy).
+     * Currently it only applies to text media SDP.
+     *
+     * Default: PJMEDIA_DEFAULT_REDUNDANCY_LEVEL
+     */
+    int         red_level;
```
nanangizz, line 83: Perhaps add some hints/links about redundancy docs here or somewhere else (if not yet), e.g: what it is, what level represents & the effect/risk, valid level values (if relevant), and how it is negotiated.

pjmedia/include/pjmedia/rtp.h
```diff
@@ -105,11 +105,34 @@ struct pjmedia_rtp_hdr
 };
 #pragma pack()
 
+/**
+ * The first byte of RTP packet additional header.
```
nanangizz, line 109: As this is a kind of not ordinary header, add a brief desc about the use-case/purpose?

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -2915,6 +2952,10 @@ pj_status_t pjsua_media_channel_create_sdp(pjsua_call_id call_id,
                     m->desc.media = pj_str("video");
                     m->desc.fmt[0] = pj_str("31");
                     break;
+                case PJMEDIA_TYPE_TEXT:
+                    m->desc.media = pj_str("text");
+                    m->desc.fmt[0] = pj_str("98");
```
nanangizz, line 2957: Better not to put a dynamic-pt here to avoid pt collision?

pjmedia/include/pjmedia/txt_stream.h
```diff
@@ -0,0 +1,222 @@
...
+ *    direction),
+ *  - one instance of RTCP session (#pjmedia_rtcp_session),
+ *  - and a reference to media transport to send and receive packets
+ *    to/from the network (see @ref PJMEDIA_TRANSPORT).
+ *
+ * Text streams are created by calling #pjmedia_txt_stream_create(),
+ * specifying #pjmedia_txt_stream_info structure in the parameter. Application
+ * can construct the #pjmedia_txt_stream_info structure manually, or use 
+ * #pjmedia_txt_stream_info_from_sdp() function to construct the
+ * #pjmedia_txt_stream_info from local and remote SDP session descriptors.
```
nanangizz, line 65: Should it also mentions the RFC the implementation is based on, so users get a picture of what to expect (features, interoperability, etc)

pjmedia/include/pjmedia/txt_stream.h
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @return              PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t)
+pjmedia_txt_stream_set_rx_callback(pjmedia_txt_stream *stream,
+                                   void (*cb)(pjmedia_txt_stream*,
+                                              void *user_data,
+                                              const pjmedia_txt_stream_data *data),
+                                   void *user_data,
+                                   unsigned option);
```
nanangizz, line 199: So it is decided to use different approach for sending & receiving to audio & video streams, i.e: audio/video uses pjmedia port mechanism (`put/get_frame()`), this stream uses callback for receiving and send() function for sending.

sauwming: Text stream actually has its own multiparty/conferencing specification in RFC https://datatracker.ietf.org/doc/html/rfc9071, and after reading it, I came to the conclusion that since text stream is much more lightweight compared to audio&video, it doesn't need a complex media port and conferencing feature that we have right now with av.

So for conferencing, text stream simply needs to do this (this is possible because text stream doesn't require complex conference processing, such as format conversion, resampling, mixing, etc):
```
on_call_rx_rtp()
{
    // Broadcast the received text to the other participants
    for (I = 0; I < conf_nparty; I++)
         pjsua_call_send_text(party[I])
}
```

But I'm not quite clear with what you have in mind, so perhaps you can explain more? How do you foresee the API like? Do we need txt_port and txt_conf? How do we send the text to txt_stream, and how do we get notified for incoming text?

nanangizz: The simpler mechanism in this PR can be used to used to implement same functionalities as pjmedia port mechanism for sure (as shown above for conferencing), it is just not maintaining the same pattern as other streams. I agree using the pjmedia port mechanism is perhaps a bit overkill for most use case: simple callback & simple text send, and I think it is okay to use this simpler mechanism, we can enhance it later (to using pjmedia port mechanism) if needed.

In my mind, similar to other media, an app may use a various text sink port implementations, e.g: text-to-speech port, command parser & interpreter port. And for the source port, it may be a video subtitle port (timely prearranged text). So using the pjmedia port mechanism perhaps still makes sense.

Rather OOT, just a heads-up (to be anticipated when designing APIs), beside conferencing, this end-to-end text streaming may need to catch-up some "common" features in the future, e.g: peer-to-peer encryption (a conference must not decrypt the payload), image/video/voice sharing.

sauwming: Okay, let me try to implement the pjmedia port mechanism.
So instead of send text API, app will call `pjmedia_port_put_frame()`.
While for the incoming direction, app can have two options:
- using the callback as it is now.
- call `pjmedia_port_get_frame()`.
Let me know if there's anything else needed.

For encryption, since it's using RTP, it can use SRTP (https://www.rfc-editor.org/rfc/rfc4103.html#section-8).
What is image/video/voice sharing?

nanangizz: Actually I meant it is totally okay to use simpler mechanism for now, no urgency for pjmedia port mechanism and perhaps enhancing it later will not cause backward compatibilities issues (?). Also, the pjmedia port mechanism may require text conference bridge.

Re: encryption, a possible challenge is that our current conference bridge model, the processing is done on unencrypted data (i.e: not P2P encrypted).
Re: image/video/voice/doc sharing, it is basically sending binary data, not sure if it is part of the text streaming though, but just usually modern text chat apps also have such capabilities, so perhaps feature requests will follow in the near future :)

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -0,0 +1,1054 @@
...
+PJ_DEF(pj_status_t) pjmedia_txt_stream_destroy( pjmedia_txt_stream *stream )
+{
+    pjmedia_stream_common *c_strm = (pjmedia_stream_common *)stream;
+
+    PJ_ASSERT_RETURN(stream != NULL, PJ_EINVAL);
+
+    PJ_LOG(4,(c_strm->port.info.name.ptr, "Stream destroying"));
+
+    if (stream->clock)
+        pjmedia_clock_stop(stream->clock);
```
nanangizz, line 140: Didn't check the code. Just a heads-up, if there is possibility that the `destroy()` is called from the clock callback, the clock may not be cleaned up properly (may cause pool & handle leaks).

pjsip-apps/src/pjsua/pjsua_app_config.c
```diff
@@ -180,6 +180,11 @@ static void usage(void)
     puts  ("  --auto-play-avi     Automatically play the AVI media to call");
 #endif
 
+    puts  ("");
+    puts  ("Text Options:");
+    puts  ("  --text              Enable real-time text stream");
+    puts  ("  --text-red=N        Text stream redundancy level (default=0)");
```
nanangizz, line 186: the default seems to be 2?

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -0,0 +1,1054 @@
...
+
+    /* Init buffering time and create clock. */
+    stream->buf_time = info->buffer_time;
+    if (stream->buf_time == 0)
+        stream->buf_time = BUFFERING_TIME;
+    cparam.usec_interval = stream->buf_time * 1000;
+    cparam.clock_rate = 1000;
+    status = pjmedia_clock_create2(pool, &cparam,
+                                   PJMEDIA_CLOCK_NO_HIGHEST_PRIO,
+                                   clock_cb, stream, &stream->clock);
```
nanangizz, line 266: Looking at the clock interval & jitter buffer params (frame size & max cnt), it does not seem optimized for 'medium-sized' image/video/binary files. Anyway, I believe sending binary is not part of the current objectives.

nanangizz (commented):
Just tried the pjsua app, seems to work fine.

pjmedia/src/pjmedia/txt_stream.c
sauwming: Ah yes, I'll put a warning in the doc. As unlikely as it seems that app will want to destroy the stream upon receiving a text, I guess we'll never know what people will do.

pjsip-apps/src/pjsua/pjsua_app_config.c
sauwming: For PJSUA-LIB, the default is 2.
But for pjsua app, if not specified, the default is 0 (no redundancy will be offered).

This may be kind of similar to video and text settings itself. In PJSUA-LIB, the default of vid_cnt and txt_cnt is 1, but in pjsua app it's 0.

pjmedia/src/pjmedia/txt_stream.c
sauwming: Yes, for custom usage (such as binary), users will need to adjust some things. Because the primary usage here is of course text as per the RFC, and the RFC anticipates usage of about 30 cps only (character per second) and recommends a rather long buffering time of 300 ms.

sauwming (commented):
Modification based on comments in the last commit.
Also add pjsua2 acc config, so no more todo from my side.

pjmedia/src/pjmedia/txt_stream.c
nanangizz: Some interactive app may hangup the call when receiving a specific text/command?

pjmedia/src/pjmedia/txt_stream.c
nanangizz: Okay, so it looks like transfering binary file needs to be done via other means.

pjmedia/build/pjmedia.vcxproj
```diff
@@ -717,6 +717,7 @@
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Release-Dynamic|ARM64'">true</ExcludedFromBuild>
     </ClCompile>
     <ClCompile Include="..\src\pjmedia\transport_udp.c" />
+    <ClCompile Include="..\src\pjmedia\txt_stream.c" />
```
nanangizz, line 720: Minor: please also add the `stream_imp_common.c` to the project (with exclude from build), so it can be included in searching in files.

nanangizz (approved):
just a minor above

pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -1428,6 +1451,56 @@ static pj_status_t match_offer(pj_pool_t *pool,
...
+            a_red_level = fmtp.cnt - 1;
+            if (a_red_level > o_red_level) {
+                a = pjmedia_sdp_media_find_attr2(preanswer, "fmtp",
+                                                 &preanswer->desc.fmt[i]);
+                if (a) {
+                    int lvl = 0;
+                    for (i = 0; i < (unsigned)a->value.slen; i++) {
+                        if (*(a->value.ptr + i) == '/') {
+                            if (lvl++ >= o_red_level) {
+                                a->value.slen = i;
```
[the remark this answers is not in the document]
sauwming: Should be okay, we're going to break from the outer loop after this block.

LeonidGoltsblat:
@sauwming!  This PR is already closed, however, we currently have dozens (!) of idle clock threads. These threads are not heavily loaded — do we really need a dedicated clock thread for each stream? Is it necessary to waste resources in this way? Perhaps we could use something like pj_timer instead (i.e., share a single thread for handling all the callbacks)?

sauwming:
I did think of that and in fact, my earlier version (before I put it in the PR) of the implementation used the timer heap:
```
@@ -137,6 +129,9 @@ typedef struct pjmedia_txt_stream pjmedia_txt_stream;
+ * @param timer         Timer heap instance to be used by the resolver.
+ *                      The timer heap will function as a clock for
+ *                      the text stream.
@@ -148,6 +143,7 @@ typedef struct pjmedia_txt_stream pjmedia_txt_stream;
 PJ_DECL(pj_status_t) pjmedia_txt_stream_create(pjmedia_endpt *endpt,
                                            pj_pool_t *pool,
                                            const pjmedia_txt_stream_info *info,
+                                           pj_timer_heap_t *timer,
                                            pjmedia_transport *tp,
                                            void *user_data,
                                            pjmedia_txt_stream **p_stream);
```

At that time, because media doesn't have its own timer heap, in pjsua, I will create it using the SIP's timer heap (which of course is not the cleanest approach):
```
+        /* Create session based on session info. */
+        status = pjmedia_txt_stream_create(pjsua_var.med_endpt, NULL, si,
+                                pjsip_endpt_get_timer_heap(pjsua_var.endpt),
+                                call_med->tp, NULL,
+                                &call_med->strm.t.stream);
```

But after some consideration, I decided against it because the signature will look different with the other media streams (audio and video), and I didn't expect text stream to be used a lot.

I see your point though, so I will put this in our roadmap.

LeonidGoltsblat:
> I didn't expect text stream to be used a lot

Perhaps I’m misunderstanding the intent, but pjsua_call_setting_default() sets pjsua_call_setting.txt_cnt to 1 by default, meaning that all previously written applications will create text streams—and therefore clock threads—for every call. Is this an intended behavioral change?
If so, would it be possible to delay the creation of the clock thread until it is actually needed?

As for the second issue:
in my opinion the PJSIP clock thread implementation is designed for long-lived threads, not short-lived ones, making this usage inefficient: the clock thread is closed via a synchronous operation that typically waits for the next timer tick—on average, around 150 ms (300 ms interval / 2). This causes noticeable performance degradation. 
Would it be possible to make clock thread termination asynchronous instead?

sauwming:
Originally I wanted to put the default as zero, but again, this would cause difference with audio/video. Besides, more importantly, using zero as default requires flow change in pjsua (since in pjsua we will enable and init a media first, and disable it later when not needed).

So that's the background of why the default is one. But then, even if the default is one, I would expect users to disable it if they don't need it (yes, we probably would need to write this in the release notes during the official release).
For your case, I believe you can also simply disable it.

PR #4360, opened 2025-03-18
Aligned memory allocation fixes and enhancements

LeonidGoltsblat:
Here is minor fixes and improving to #4277 

Invalid syntax in the pool_i.h fixed: both pj_pool_aligned_alloc and pj_pool_aligned_create was defined with PJ_IDECL but should be defined with PJ_IDEF

Improving :
- there is no need to require that the current pj_pool_aligned_alloc alignment argument be >= the default pool alignment >= PJ_POOL_ALIGNMENT. (It is enough to check arg with PJ_IS_POWER_OF_TWO and nothing bad will happen if the application wants to allocate unaligned memory, i.e. with alignment == 1.)
- there is no need to round up the size of the allocated memory, this is the compiler's job, not the allocator's. (If the application requests 1 byte aligned to 256 bytes, it's enough to reserve 255 bytes before, but not after, the allocation.)


LeonidGoltsblat:
fixed! Thanks!

sauwming (approved):

LeonidGoltsblat:
In response to #4389, #4391, added a test to check PJ_POOL_ALIGN_PTR(block->cur, alignment) > block->end condition

PR #4365, opened 2025-03-19
Surround the AMediaCodec_setAsyncNotifyCallback (API 28+) with availability check

micha102:
Surround the AMediaCodec_setAsyncNotifyCallback (API 28+) with availability check
This is needed in order to allow the library to compile when targeting old Android SDK (21 for example).
Otherwise, the library can only compile with the minimum API of 28


sauwming:
Looks like the patch provided is incorrect/unclean.

micha102:
Sorry I committed and did the PR from a wrong branch.

PR #4366, opened 2025-03-20
Signed-off-by: Pesala Silva <pesala.silva@ifs.com>Media-Port-Changes-for-call-record-and-play-feature

PesalaDeSilva:
Media-Port-Changes-for-call-record-and-play-feature

### DIFF
--- a/pjmedia/include/pjmedia/conference.h
+++ b/pjmedia/include/pjmedia/conference.h
@@ -213,7 +213,19 @@ PJ_DECL(pjmedia_port*) pjmedia_conf_get_master_port(pjmedia_conf *conf);
 PJ_DECL(pj_status_t) pjmedia_conf_set_port0_name(pjmedia_conf *conf,
                                                  const pj_str_t *name);
 
+/**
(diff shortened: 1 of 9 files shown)


sauwming:
In my opinion, adding memory player&recorder to pjsua API doesn't appear very appealing. Yes, it can prove to be useful for some who require the feature, but considering the rather extensive modifications required and the very specific purpose it serves, I'm more inclined not to integrate this.

My recommendation is to move the implementation into the application level, as separate files from the PJSIP library. This way, you don't need to modify PJSIP library, and you won't be affected if there's a PJSIP version upgrade.

But first I would invite my other team to review this, perhaps they have a different opinion.

nanangizz:
IMO the pjmedia ports should not be implemented inside PJSUA-LIB.

Btw, there are already memory player & writer ports in PJMEDIA. As other pjmedia ports, application can use & manage them directly without PJSUA-LIB.

sauwming:
Thank you for the PR but our apologies that we can't accept this at the moment.

PR #4367, opened 2025-03-20
Surround the AMediaCodec_setAsyncNotifyCallback (API 28+) with availability check

micha102:
Surround the AMediaCodec_setAsyncNotifyCallback (API 28+) with availability check This is needed in order to allow the library to compile when targeting old Android SDK (21 for example). Otherwise, the library can only compile with the minimum API of 28

### DIFF
--- a/pjmedia/src/pjmedia-codec/and_aud_mediacodec.cpp
+++ b/pjmedia/src/pjmedia-codec/and_aud_mediacodec.cpp
@@ -1093,8 +1093,12 @@ static pj_status_t and_media_codec_open(pjmedia_codec *codec,
                    s->dec_setting.octet_aligned, s->dec_setting.reorder));
     }
 #endif
-    AMediaCodec_setAsyncNotifyCallback(codec_data->enc, async_cb, codec_data);
-    AMediaCodec_setAsyncNotifyCallback(codec_data->dec, async_cb, codec_data);
+    if (__builtin_available(android 28, *)) {
+        AMediaCodec_setAsyncNotifyCallback(codec_data->enc,
+                                           async_cb, codec_data);
+        AMediaCodec_setAsyncNotifyCallback(codec_data->dec,
+                                           async_cb, codec_data);
+    }
 
     status = configure_codec(codec_data, PJ_TRUE);
     if (status != PJ_SUCCESS) {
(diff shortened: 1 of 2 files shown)


sauwming (approved):

nanangizz:
It looks like the codec won't work without `AMediaCodec_setAsyncNotifyCallback()`?
Just for ref, it was added in #4105.

micha102:
Yes as it is, the codec will only work on Android fevices with 28 and above. Which is the same thing as before the PR.
The PR specifically removes the blockage of building the library with lower APP_PLATFORM values.
If the Android device is API 28+, the codec will always work, either when the library is built with APP_PLATFORM=21, or with API 28 or latest.

But I agree that fundamentally, this PJMEDIA_HAS_ANDROID_MEDIACODEC must somewhere fallback to a usable function for older devices.

I will check if there's an alternative older function that could work on Android prior to 28.
The easier way (but I'm not fan of it) is we add __ANDROID_API__ >= 28 to the whenever PJMEDIA_HAS_ANDROID_MEDIACODEC is invoked

Like in and_vid_mediacodec.h audio_codecs.c pjsua_vid.c and_aud_mediacodec.cpp and and_vid_mediacodec.cpp

```
#if defined(PJMEDIA_HAS_ANDROID_MEDIACODEC) && \
            PJMEDIA_HAS_ANDROID_MEDIACODEC != 0
```

We put 
```
#if defined(PJMEDIA_HAS_ANDROID_MEDIACODEC) && \
            PJMEDIA_HAS_ANDROID_MEDIACODEC != 0 &&  __ANDROID_API__ >= 28
```

I'm not fan because building with APP_PLATFORM >= 28 will make the library only working on Android 28+ (Older Androids will have a JNI import problem right when the app starts).
But with _builtin_available it will just works OK until the device really needs to invoke the problematic function. That's why the code must provide fallback in case API is less than 28

nanangizz:
Ah okay.

The codec should never be registered when API level is < 28 so it does not appear in the SDP to be negotiated and so on. I think we need a mechanism to check API level at run-time and add check in `pjmedia_codec_and_media_vid_init()` to avoid the codec being registered and print warning messages to the log file (indicating some codecs are disabled because device API level is < 28).

Haven't check the details, but [`pj_get_sys_info()`](https://github.com/pjsip/pjproject/blob/4d3a7f1e0f20dc8e7db82ae32b019a2ef71f16f0/pjlib/src/pj/os_info.c#L129) for Android does not seem to have the SDK info yet. And the most "portable" way to get device API level info is perhaps via `android.os.Build.VERSION.SDK_INT` using JNI (e.g: [here](https://stackoverflow.com/questions/10196361/how-to-check-the-device-running-api-level-using-c-code-via-ndk))? I can try to implement this part if you want.

PR #4369, opened 2025-03-25
Fix failed android build to to android media codec require android-28

trengginas:
Since #4105, android media codec is using async method which requires android-28.
On #4364, on NDK >= r21the APP_PLATFORM is picked by the minimum SDK version available. e.g.: on ndk r27 the minimum is android-21. This will fail the build process with this message:
```
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
This PR will change the default minimum SDK to android-28 and change how android media codec detection works in `aconfigure`.

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -9595,9 +9595,9 @@ else case e in #(
   e)
         case $target in
             *android*)
-                { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for AMediaCodec_createDecoderByType in -lmediandk" >&5
-printf %s "checking for AMediaCodec_createDecoderByType in -lmediandk... " >&6; }
-if test ${ac_cv_lib_mediandk_AMediaCodec_createDecoderByType+y}
+                { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for AMediaCodec_setAsyncNotifyCallback in -lmediandk" >&5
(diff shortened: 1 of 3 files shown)


nanangizz:
This PR seems to try to avoid build errors when native codec is enabled by enforcint the minimum SDK to android-28 or later? While normally we want the minimum SDK set to the lowest/oldest as possible for wider range of device compatibilty. So when the app is running on newer devices, it gets the Android native codecs features and when the app is running on older devices, it does not stops due to any missing library/functions. It can be achieved via [this](https://developer.android.com/ndk/guides/using-newer-apis).

nanangizz:
See also #4367.

trengginas:
The reference above also described the fallback mechanism (using old API). Ideally, we want to support wider range device compatibility, however from tests the async method provides better stability to the codec operation instead of the old polling API. 
Alternatives:
- add fallback mechanism to using older API
- revert back the minimum SDK detection. However, android media codec detection is using function checking using new API. e.g.: when the the new API is not defined, mediacodec will not be supported. User needs to check `config.log` for this.
- revert back the minimum SDK detection. However, add `__ANDROID_API__` level check runtime. If the API is below 28, the codec operation will a NO OP and log will be printed runtime.

nanangizz:
Well, the aim of #4367 seems to be the last alternative, add run-time check of device API level and use weak-reference for `AMediaCodec_setAsyncNotifyCallback()`.

PR #4372, opened 2025-03-26
Android unit test

trengginas:
This PR will add Android Unit Test using espresso for `pjsua2` app.

Changes:
- Use `androidx` instead of `android.support`. 
- Add new build flavor type (`ciTest`) specific for the instrumentation test.
- the unit test will simulate a video call to itself.

Behavior changes to pjsua2 app when build using `ciTest` flavor type:
- any call related operation is enable only from the unit test module.
- no save/load from config file when app starts/stops.

Building the unit tests from command line: (`pjsip-apps/src/swig/java/android`)
```
./gradlew assembleCiTestDebug
```

Running the unit tests from command line: (make sure you have a device/emulator running/connected)
```
./gradlew connectedCiTestDebug
```

Alternatively, you can run and build the unit test from Android Studio by choosing the correct build variant.

### DIFF
--- a/pjsip-apps/src/swig/java/android/app-kotlin/build.gradle
+++ b/pjsip-apps/src/swig/java/android/app-kotlin/build.gradle
@@ -6,6 +6,8 @@ plugins {
 android {
     compileSdk 33
 
+    buildFeatures.buildConfig true
+
     defaultConfig {
         applicationId "org.pjsip.pjsua2.app_kotlin"
         minSdkVersion 23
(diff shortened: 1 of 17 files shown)


sauwming:
Nice!
Just to be clear, this is roadmap: https://github.com/pjsip/roadmap/issues/52, right?
And is this supposed to run on Bitrise, or does it need a separate Android CI test platform?

trengginas:
Yes, it is part of https://github.com/pjsip/roadmap/issues/52.
It should work with bitrise. I only tested using bitrise CLI locally, however the workflow should not be much different.

nanangizz:
- What tests are performed by this unit test?
- Will this test report any test failures?
- Rather OOT, it looks like currently Android tests are not enabled (in Bitrise/Github CI), is enabling such tests also part of this PR?

nanangizz:
>     * Currently, android build CI is available on bitrise. 

After #4369 is merged, Android build fails on my local machine, but Bitrise/Github CI seems just fine.

nanangizz (approved):

sauwming (approved):

PR #4382, opened 2025-04-02
pool_i.h: Check available size instead of pointer wrap-around to avoid autological-compare.

amubiera:
Fixes Clang v20.1.1 generating autological warning (as error):

```
pjsip/pjproject/pjlib/include/pj/pool_i.h:61:34: error: pointer comparison always evaluates to true [-Werror,-Wtautological-compare]
   61 |         block->cur <= ptr && ptr <= ptr + size) {
      |                                  ^
1 error generated.
```

I also believe that you shouldn't need the `block->cur <= ptr` portion since `PJ_POOL_ALIGN_PTR` should always align forwards. Perhaps that should instead be an assert to validate the correctness of the macro. Or tested in a unit test.

i.e. this should be enough?
```
ptr = PJ_POOL_ALIGN_PTR(block->cur, alignment);
if (ptr + size <= block->end) {
    block->cur = ptr + size;
    return ptr;
}
```

CC: @LeonidGoltsblat

### DIFF
--- a/pjlib/include/pj/pool_i.h
+++ b/pjlib/include/pj/pool_i.h
@@ -56,9 +56,9 @@ PJ_IDEF(void*) pj_pool_alloc_from_block( pj_pool_block *block, pj_size_t alignme
     //    size = (size + alignment) & ~(alignment -1);
     //}
     ptr = PJ_POOL_ALIGN_PTR(block->cur, alignment);
-    if (ptr + size <= block->end &&
-        /* here we check pointer overflow */
-        block->cur <= ptr && ptr <= ptr + size) {
+    if (block->cur <= ptr && /* check pointer overflow */
+        block->end - ptr >= size) /* check available size */
+    {
         block->cur = ptr + size;
         return ptr;
     }


LeonidGoltsblat:
In the real world this is enough, but what about pointer overflow?
What if:
block->cur == ULONG_MAX - size/2
ptr == ULONG_MAX - size/4
block->end == ULONG_MAX - size/8
then
ptr + size == size*3/4
and even though ptr+size < block->end the memory is not in the block.
Are there any errors in this example? If not, then potentially the compiler behavior is incorrect. I think it is better to add a compiler version check here and disable "unnecessary autological comparison" for a particular compiler.

amubiera:
@LeonidGoltsblat, I see. You're trying to protect against a pointer wrap-around near the end of virtual memory. 0xFFFFFFFFFFFFFFFF ? While highly unlikely, I can see your concern.

Perhaps a better fix would be to check for available space, rather than pointer wrap-around?
e.g. `block->end - block->cur >= size`

p.s. I may try to generate a tiny example to post on the Clang forums to see why they deem this "always true" result.

LeonidGoltsblat:
Good idea! But don't forget to add the (-(pj_ssize_t)(PTR) & (ALIGNMENT-1)) term to the right side

amubiera:
> But don't forget to add the (-(pj_ssize_t)(PTR) & (ALIGNMENT-1)) term to the right side

@LeonidGoltsblat Not sure what you mean? The RHS would be the size.

So perhaps something like this:
```C
ptr = PJ_POOL_ALIGN_PTR(block->cur, alignment);
if (block->end - ptr >= size) {        
    block->cur = ptr + size;
    return ptr;
}
```

LeonidGoltsblat:
It's ok for me, but I'm not a member of the pjsip team.
Please add comments to this "if". Next time you (not me) will explain to someone why this "if" is written here. :smiley:

sauwming (approved):

nanangizz:
- agree with @LeonidGoltsblat , please add (or put back) comments about the pointer value wrap-around check for future ref.
- also while here, better move the opening brace to a new line as specified [here](https://docs.pjsip.org/en/latest/get-started/coding-style.html#use-k-r-style-brace-placement), opening braces are on the same line, except if the statement preceeding the braces is multiline.

nanangizz:
Looking at the condition again, when `ptr` is overflow, `block->end - ptr` will give even a bigger value (than when not overflow), so the overflow detection will fail here?

From [here](https://github.com/llvm/llvm-project/issues/120214), the problem root seems to be that unsigned pointer arithmetic cannot overflow. So simply cast the `size` to signed?
```
    if (ptr + size <= block->end &&
        /* here we check pointer overflow */
        block->cur <= ptr &&
        ptr <= ptr + (pj_ssize_t)size)
    {..}
```

amubiera:
@nanangizz 
I believe the overflow is protected now:
* The first checks for overflow `block->cur <= ptr`.
* The second check is then safe `block->end - ptr >= size` since `ptr` has not wrapped around.

nanangizz:
Ah okay, just realized that when wrap around happens:
- if `ptr < cur`, clear indication of wrap around
- if `ptr >= cur`, requested size must be larger than capacity (stopped by the second condition, `block->end - ptr >= size`).

nanangizz (approved):

LeonidGoltsblat:
@amubiera 
Now we have a new warning here (at least in Visual Studio):
...\pjlib\include\pj\pool_i.h(60,26): warning C4018: '>=': signed/unsigned mismatch

Please, fix!

PR #4385, opened 2025-04-08
Add AVI writer

sauwming:
For the purpose of video end-to-end quality test (as per roadmap), we need an AVI writer so we can compare the quality of the received video with the original.

Limitation:
Currently the AVI writer only supports recording uncompressed video, so the resulting file size is very large. Hence, it is more suited for experimental or testing purposes (such as to check whether video is working, or to test video quality), rather than for actual recording or archiving.

### DIFF
--- a/pjmedia/include/pjmedia/avi_stream.h
+++ b/pjmedia/include/pjmedia/avi_stream.h
@@ -23,7 +23,7 @@
  * @brief AVI file player.
  */
 #include <pjmedia/port.h>
-
+#include <pjmedia/av_sync.h>
 
 
 PJ_BEGIN_DECL
@@ -64,6 +64,13 @@ typedef pjmedia_port pjmedia_avi_stream;
  * zero or more AVI stream.
  */
 typedef struct pjmedia_avi_streams pjmedia_avi_streams;
+
+struct pjmedia_avi_streams
+{
+    pj_pool_t       *pool;
+    unsigned         num_streams;
+    pjmedia_port   **streams;
+};
 
 /**
  * Create avi streams to play an AVI file. AVI player supports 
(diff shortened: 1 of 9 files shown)


pjsip-apps/src/pjsua/pjsua_app_config.c
```diff
@@ -562,6 +567,10 @@ static pj_status_t parse_args(int argc, char *argv[],
         { "vrender-dev",  1, 0, OPT_VRENDER_DEV},
         { "play-avi",   1, 0, OPT_PLAY_AVI},
         { "auto-play-avi", 0, 0, OPT_AUTO_PLAY_AVI},
+        { "rec-avi",   1, 0, OPT_REC_AVI},
```
nanangizz, line 570: `play-avi` can be specified multiple times to play multiple AVI files, while it does not seem to be the same with `rec-avi`.
I guess it need to be described somewhere.

pjmedia/include/pjmedia/avi_stream.h
```diff
@@ -23,7 +23,7 @@
  * @brief AVI file player.
  */
 #include <pjmedia/port.h>
-
+#include <pjmedia/av_sync.h>
```
nanangizz, line 26: Is this necessary?

nanangizz (approved):
Just a couple minors.

pjsip-apps/src/pjsua/pjsua_app_config.c
sauwming: Looking at the other commands (such as for WAV --play-file and --rec-file), it seems that the description should be put in the command that can be specified multiple times, so I follow that pattern in the last commit.

PR #4397, opened 2025-04-15
Check return value when setting ALSA format

trengginas:
Currently, the library only supports 16 `bits_per_sample` when opening the sound device. 
When the sound device only supports 32 bits per sample (`S32_LE` e.g.: dsnoop), the method will crash here. https://github.com/pjsip/pjproject/blob/f80984b79b7a45f136304bba83bb7aa929efac11/pjmedia/src/pjmedia-audiodev/alsa_dev.c#L615

To avoid the crash, it is better to return an error when setting the format failed.

Note:
For devices that only uses 32 bits per sample config or multiple channel, you can use a `plug:` plugin and create an alias.
This way ALSA will provide the required conversion and allow the device to be opened with other configuration.
```
pcm.tmp_dsnoop {
    type plug
    slave {
        pcm "dsnoop:CARD=PCH,DEV=0"
    }
}
```
This way, ALSA will create a `tmp_dsnoop` alias and use it as the sound device instead.

### DIFF
--- a/pjmedia/src/pjmedia-audiodev/alsa_dev.c
+++ b/pjmedia/src/pjmedia-audiodev/alsa_dev.c
@@ -704,6 +704,7 @@ static pj_status_t open_playback (struct alsa_stream* stream,
             "playback device '%s', err: %s", format,
             stream->af->devs[param->rec_id].name, snd_strerror(result)));
 
+        snd_pcm_close (stream->pb_pcm);
         return PJMEDIA_EAUD_SYSERR;
     }    
 
@@ -849,6 +850,7 @@ static pj_status_t open_capture (struct alsa_stream* stream,
             "capture device '%s', err: %s", format,
             stream->af->devs[param->rec_id].name, snd_strerror(result)));
 
+        snd_pcm_close (stream->ca_pcm);
         return PJMEDIA_EAUD_SYSERR;
     }
 
--- a/pjmedia/src/pjmedia-audiodev/alsa_dev.c
+++ b/pjmedia/src/pjmedia-audiodev/alsa_dev.c
@@ -698,7 +698,14 @@ static pj_status_t open_playback (struct alsa_stream* stream,
         format = SND_PCM_FORMAT_S16_LE;
         break;
     }
-    snd_pcm_hw_params_set_format (stream->pb_pcm, params, format);
+    result = snd_pcm_hw_params_set_format (stream->pb_pcm, params, format);
+    if (result < 0) {
+        PJ_LOG (3,(THIS_FILE, "Unable to set format %d for "
+            "playback device '%s', err: %s", format,
+            stream->af->devs[param->rec_id].name, snd_strerror(result)));
+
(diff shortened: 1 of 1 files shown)


sauwming (approved):

nanangizz:
While here, for stability, I guess it is better to also scan all other ALSA functions in the file to make sure they have error handlings, e.g: `snd_pcm_hw_params_alloca()`, `snd_pcm_hw_params_any()`?

nanangizz (approved):
Sorry, forgot to mention that just in case, some error, if any, is supposed to be tolerable, it should just print warning instead of returning error. Anyway, make sure to test this on some ALSA devices/configs.

trengginas:
I have rechecked the sample ([ref](https://www.alsa-project.org/alsa-doc/alsa-lib/_2test_2pcm_8c-example.html#a14)) and ([ref](https://www.alsa-project.org/alsa-doc/alsa-lib/_2test_2latency_8c-example.html#a5)). And it just stopped the sample when the parameter setting fails.

nanangizz:
> I have rechecked the sample ([ref](https://www.alsa-project.org/alsa-doc/alsa-lib/_2test_2pcm_8c-example.html#a14)) and ([ref](https://www.alsa-project.org/alsa-doc/alsa-lib/_2test_2latency_8c-example.html#a5)). And it just stopped the sample when the parameter setting fails.

Well, the sample is a sample :) imho, for our use case, some parameters may be not so urgent, for example `snd_pcm_hw_params_set_buffer_size_near()`, rather than returning error & stopping the device, I guess letting the device opened & working should be preferable?

trengginas:
Then we should also ignore the error from:
- `snd_pcm_hw_params_set_rate_near()`
- `snd_pcm_hw_params_set_period_size_near()`

nanangizz:
> Then we should also ignore the error from:
> 
>     * `snd_pcm_hw_params_set_rate_near()`
> 
>     * `snd_pcm_hw_params_set_period_size_near()`

Not so sure, imo the clock rate & samples per frame (?) are urgent params, will cause problems if not set as specified?

pjmedia/src/pjmedia-audiodev/alsa_dev.c
```diff
@@ -744,6 +787,12 @@ static pj_status_t open_playback (struct alsa_stream* stream,
         tmp_buf_size = tmp_period_size * 2;
     snd_pcm_hw_params_set_buffer_size_near (stream->pb_pcm, params,
                                             &tmp_buf_size);
+    if (result < 0) {
+        PJ_LOG (3,(THIS_FILE, "Unable to set period size: %d for "
+                  "playback device '%s', err: %s", (int)tmp_buf_size,
+                  stream->af->devs[param->play_id].name, snd_strerror(result)));
```
nanangizz, line 793: Minor: perhaps better put "Warning: unable..." or "Ignored: unable...", to avoid confusion when investigating log file in the future?

nanangizz (approved):

PR #4398, opened 2025-04-15
Do not erase decoded video frame in on_rx_rtp

half-left:
Stream can try to decode another on due to ts_diff greater than `stream->dec_max_delay,` but it can be unsucsessful because of `frm_cnt < stream->dec_delay_cnt`, so it should restore already decoded frame

### DIFF
--- a/pjmedia/src/pjmedia/vid_stream.c
+++ b/pjmedia/src/pjmedia/vid_stream.c
@@ -967,9 +967,10 @@ static void on_rx_rtp( pjmedia_tp_cb_param *param)
             }
 
             if (can_decode) {
+                pj_size_t old_size = stream->dec_frame.size;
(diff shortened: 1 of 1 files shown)


half-left:
sorry, it's already solved in https://github.com/pjsip/pjproject/commit/078bfeae295464d36e9756f187e2f1b76bf2d350

aol-nnov:
Seems to be implemented in https://github.com/pjsip/pjproject/pull/4325

@nanangizz do you have any estimates in which release will it be incorporated? That PR has no milestone unfortunately..

nanangizz:
> @nanangizz do you have any estimates in which release will it be incorporated? That PR has no milestone unfortunately..

Oops, forgot to set the milestone, it should be 2.16, thanks for catching that.

Issue #4413, opened 2025-04-30
3GPP TS 33102 V16.0.0(2020-07) allows greater length of authentication parameters


wosrediinanatour:
### Describe the bug

For Digest Authentication of an PJSUA(2) account during registration process, if the password is > 16 characters, then registration will fail without any error message. 

### Steps to reproduce

Use an authentication key (K) in https://docs.pjsip.org/en/latest/api/generated/pjsip/group/group__PJSUA2__SIP__Types.html#_CPPv4N2pj12AuthCredInfo4akaKE with a length > 16 characters and try to register using Digest Authentication:
```cpp
auto auth = pj::AuthCredInfo("Digest", realm, username, PJSIP_CRED_DATA_EXT_AKA , 0);
auth.akaK = password;
config.sipConfig.authCreds.push_back(std::move(auth));  // config is of type pj::AccountConfig. config is used when creating an account.
```

### PJSIP version

2.14.1, 2.15

### Context

Latest version of the standard supports the following lengths of authentication parameters:

![Image](https://github.com/user-attachments/assets/88f12773-ae04-4787-b23b-aea01dae2513)

The length of the authentication key (K) has changed over time.

### Log, call stack, etc

```shell
Registration fails silently, when a SIP 401 is received after the initial unprotected SIP REGISTER had been sent.
```

PR #4414, opened 2025-04-30
Support latest 3GPP TS 33102

wosrediinanatour:
The length of the authentication key (K) has changed since the implementation of Digest Authentication.

Issue: #4413

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -2767,7 +2767,12 @@ static pj_status_t pjsua_regc_init(int acc_id)
     /* Set credentials
      */
     if (acc->cred_cnt) {
-        pjsip_regc_set_credentials( acc->regc, acc->cred_cnt, acc->cred);
+        status = pjsip_regc_set_credentials( acc->regc, acc->cred_cnt, acc->cred);
+        if (status != PJ_SUCCESS) {
+            pjsua_perror(THIS_FILE,
+                         "Cannot set credentials for registration",
+                         status);
+        }
     }
 
     /* Set delay before registration refresh */
--- a/pjsip/include/pjsip/sip_auth_aka.h
+++ b/pjsip/include/pjsip/sip_auth_aka.h
@@ -129,7 +129,7 @@ PJ_BEGIN_DECL
 /**
  * Length of permanent/subscriber Key (K) in bytes.
  */
-#define PJSIP_AKA_KLEN          16
+#define PJSIP_AKA_KLEN          32
 
 /**
  * Length of AKA authentication code in bytes.


sauwming (approved):

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -2767,7 +2767,12 @@ static pj_status_t pjsua_regc_init(int acc_id)
     /* Set credentials
      */
     if (acc->cred_cnt) {
-        pjsip_regc_set_credentials( acc->regc, acc->cred_cnt, acc->cred);
+        status = pjsip_regc_set_credentials( acc->regc, acc->cred_cnt, acc->cred);
+        if (status != PJ_SUCCESS) {
+            pjsua_perror(THIS_FILE,
+                         "Cannot set credentials for registration",
+                         status);
+        }
```
nanangizz, line 2775: Thanks for the report & the fix.

Rather OOT, while here, perhaps failure in setting credentials (partial/complete) should cause `pjsua_regc_init()` to fail (or returning error) as well so the app will be aware of any potential problems earlier?

Also, there are some other function calls around here without error checks, I guess better fix them as well :)
If you want, we can do it.

wosrediinanatour: I have also thought about this (.e.g. by introducing a "goto on_error_return"). 

One other topic:  `pjsip_regc_set_credentials()` fails, if one "AuthCredInfo" "out of N" fails. This is because an array instead of the memory pool is used. It would be better to just don't use the erroneous "AuthCredInfo". But I am not sure whether it is worth to do.

I you do the changes, it will be cool.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -2728,10 +2728,7 @@ static pj_status_t pjsua_regc_init(int acc_id)
             pjsua_perror(THIS_FILE, "Unable to generate suitable Contact header"
                                     " for registration", 
                          status);
-            destroy_regc(acc, PJ_TRUE);
-            pj_pool_release(pool);
-            acc->regc = NULL;
```
wosrediinanatour, line 2733: This is not moved to the `on_return`. Will it have side-effects?

pjsip/src/pjsua-lib/pjsua_acc.c
nanangizz: > I have also thought about this (.e.g. by introducing a "goto on_error_return").

Done. This introduces behavioral changes, but I guess it is necessary because skipping failures may introduce issues, e.g: unexpected behaviors, debugging difficulties.

> One other topic: `pjsip_regc_set_credentials()` fails, if one "AuthCredInfo" "out of N" fails. This is because an array instead of the memory pool is used. It would be better to just don't use the erroneous "AuthCredInfo". But I am not sure whether it is worth to do.

Ignoring erroneous credential information may lead to unexpected behavior. For example, if the app prefers to use AKA credentials over MD5, but as AKA is silently skipped due to an error, the app may not achieve its preferred security level. Therefore, I think allowing `pjsip_regc_set_credentials()` to fail seems to be the correct approach.

wosrediinanatour: Thank you! Looks good to me.

sauwming (approved):

PR #4421, opened 2025-05-07
Fix compile error with ffmpeg enable on linux

iysheng:
I try to build pjsip on fedora 38 os with ffmpeg 6.0.1 。And I do configure with command:
``./confugre CFLAGS=-I/usr/include/ffmpeg``。But when I do make command, I will get errors as below:

Screenshot from 2025-05-08 02-53-00
 
and something like ``undefined refrence to av_free_packet``。After I debug then I get these modifies to make success.

### DIFF
--- a/pjmedia/src/pjmedia-videodev/ffmpeg_dev.c
+++ b/pjmedia/src/pjmedia-videodev/ffmpeg_dev.c
@@ -55,9 +55,11 @@
 # include <libavutil/pixdesc.h>
 #endif
 
-#define MAX_DEV_CNT     8
+#define MAX_DEV_CNT      8
+#define MAX_DEV_NAME_LEN 80
 
 #ifndef PJMEDIA_USE_OLD_FFMPEG
+#  define av_free_packet av_packet_unref
 #  define av_close_input_stream(ctx) avformat_close_input(&ctx)
 #endif
 
@@ -303,8 +305,6 @@ static pj_status_t ffmpeg_factory_destroy(pjmedia_vid_dev_factory *f)
 #   pragma warning(pop)
 #endif
 
-#define MAX_DEV_NAME_LEN 80
-
 static pj_status_t dshow_enum_devices(unsigned *dev_cnt,
                                       char dev_names[][MAX_DEV_NAME_LEN])
 {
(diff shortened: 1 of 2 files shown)


sauwming (approved):

pjmedia/src/pjmedia-videodev/ffmpeg_dev.c
```diff
@@ -449,8 +449,8 @@ static pj_status_t ffmpeg_factory_refresh(pjmedia_vid_dev_factory *f)
                 continue;
 
             for(i = 0; i < ctx->nb_streams; i++) {
-                if (ctx->streams[i]->codec->codec_type == AVMEDIA_TYPE_VIDEO) {
-                    codec = ctx->streams[i]->codec;
+                if (ctx->streams[i]->codecpar->codec_type == AVMEDIA_TYPE_VIDEO) {
+                    codec = ctx->streams[i]->codecpar;
```
nanangizz, line 453: Perhaps it is better to put some version checks here? Something like 
https://github.com/pjsip/pjproject/blob/b34497b3cfcf388005bb9a203db86310508e796d/pjmedia/src/pjmedia-videodev/ffmpeg_dev.c#L180-L186

Not sure since which version `AVStream.codec` is deprecated, perhaps we can just follow [this](https://git.jerryxiao.cc/Android/scrcpy/commit/b98eb7d0fafd347704a6826d5996c59dd12ee301?files=app) (re: ffmpeg vs libav versions, the higher version should be okay as it seems to be removed much later after deprecated).

iysheng: Yeah, this is a more suite way to slove this, after I check the commit history of [https://git.ffmpeg.org/ffmpeg.git](https://git.ffmpeg.org/ffmpeg.git), I found in this commit , libavformat remove these releate code as below:
Screenshot from 2025-05-09 17-35-27
And the releate libavformat verion as below:
Screenshot from 2025-05-09 17-35-03

So I add a line code with ``LIBAVFORMAT_VER_AT_LEAST(59,0)`` here.

nanangizz: Perhaps the check version should refer to deprecation rather than removal, as some libraries can be built without the deprecated API. However, this is fine to me.

nanangizz (approved):

pjmedia/src/pjmedia-videodev/ffmpeg_dev.c
iysheng: I thought when these code is marked with deprecation, the compile process still can build. so I use the removal version to compare, just try to keep compile success.

PR #4386, opened 2025-04-08
Add support for the LoongArch64 architecture

xiaoxiaoafeifei:
1. Update config.guess and config.sub to the latest version for loongarch64
    https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD
    https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD
2. Resolve the following error for loongarch64
../../webrtc/src/webrtc/typedefs.h:65:2: error: #error Please add support for your architecture in typedefs.h
../../webrtc/src/webrtc/typedefs.h:69:2: error: #error Define either WEBRTC_ARCH_LITTLE_ENDIAN or WEBRTC_ARCH_BIG_ENDIAN

PR #4423, opened 2025-05-08
Fixed WebRTC build failure on iOS

sauwming:
Due to configure changes in #4386, WebRTC will fail to build on iOS:
`In file included from ../../webrtc/src/webrtc//modules/audio_processing/aec/aec_core_sse2.c:15:
/Applications/Xcode-14.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/14.0.3/include/emmintrin.h:14:2: error: "This header is only meant to be used on x86 and x64 architecture"`

 This is due to the target build is now named `aarch64-apple-darwin_ios` (previously `arm64`).

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -11081,7 +11081,7 @@ printf "%s\n" "Checking if libwebrtc is disabled...no" >&6; }
         case $target in
             *-apple-darwin_ios*)
                 case $target in
-                    arm64*)
+                    arm64* | aarch64*)
                         ac_webrtc_instset=neon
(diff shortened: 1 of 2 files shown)


nanangizz (approved):

sauwming:
Note that Bitrise CI test still failed due to invalid device, but the build itself is successful. This seems to be their issue since the device config is left at default setting (to avoid having to manually change device selection every time a device is outdated or iOS version is updated).

PR #4435, opened 2025-05-20
Append device id to its name in case of duplicated ProductName in PjAudioDevInfo.java

farazkhonsari:
On some devices the name of the audio devices are similar but the address is different.
Having address in the name helps device selection and debugging.


sauwming (approved):

nanangizz:
Agree, it is good for debugging, but perhaps not very ideal for UI.
What if we do comparison first, if the name is duplicated, add the device ID (or duplicate counter if not so complex) at the end? :)

nanangizz (approved):

PR #4436, opened 2025-05-20
Make the use of LowLatency as  oboe performance mode configurable

farazkhonsari:
On a car  hardware, it is not possible to have hardware echo cancellation in case PerformanceMode::LowLatency be used as performance mode, however it works fine with PerformanceMode::None.

Not sure, but I think, using LowLatency on a audio device that doesn't support MMAP will cause this issue.

PJMEDIA_OBOE_USE_LOWLATENCY has been defined for this purpose,
by default the value is 1 and PerformanceMode::LowLatency is used. By changing it to 0 , PerformanceMode::None will be used.

### DIFF
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -176,13 +176,6 @@
 #   endif
 #endif
 
-/**
- * Specify wheather LowLatency is used for Oboe as performance mode.
- * Default: 1
- */
-#ifndef PJMEDIA_OBOE_USE_LOWLATENCY
-#   define PJMEDIA_OBOE_USE_LOWLATENCY  1
-#endif
 
 /*
  * Types of WSOLA backend algorithm.
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -176,6 +176,13 @@
 #   endif
 #endif
 
+/**
+ * Specify wheather LowLatency is used for Oboe as performance mode.
+ * Default: 1
+ */
+#ifndef PJMEDIA_OBOE_USE_LOWLATENCY
+#   define PJMEDIA_OBOE_USE_LOWLATENCY  1
+#endif
 
 /*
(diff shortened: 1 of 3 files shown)


pjmedia/include/pjmedia/config.h
```diff
@@ -176,6 +176,13 @@
 #   endif
 #endif
 
+/**
```
sauwming, line 179: This should be in `pjmedia-audiodev/config.h`.

pjmedia/src/pjmedia-audiodev/oboe_dev.cpp
```diff
@@ -577,10 +578,17 @@ class MyOboeEngine : oboe::AudioStreamDataCallback,
                 dev_id = stream->f->dev_info[stream->param.play_id].id;
             }
         }
+
+        if(PJMEDIA_OBOE_USE_LOWLATENCY){
```
sauwming, line 582: minor: coding style, use space for better readability, i.e.
```
if () {
} else {
}
```

sauwming (commented):
Also update the PR description to be more specific, such as on which devices the issue was known to occur, and/or any supporting documentation references (if any) that mentions the compatibility of hardware EC with certain performance mode setting.

pjmedia/include/pjmedia/config.h
farazkhonsari: I moved the config to right config file

pjmedia/src/pjmedia-audiodev/oboe_dev.cpp
farazkhonsari: thanks for checking, done.

sauwming (approved):

nanangizz:
Could clarify whether the issue happens on recent Oboe versions, e.g: 1.9.x?

farazkhonsari:
@nanangizz I've tested with oboe 1.9.3

nanangizz (approved):

PR #4437, opened 2025-05-21
[pjmedia/pjsua] expose pjmedia_conf_configure_port() to PJSUA

### LINKED ISSUE #4431: [pjmedia/pjsua] Aren't WAV player set_pos() and get_pos() functions racy?
### Describe the bug

I was recently reviewing some code and stumbled upon the WAV player code.
The WAV player offers the methods `pjmedia_wav_player_port_set_pos()` and `pjmedia_wav_player_port_get_pos()`
to set and get the current position of the player.

This is also available when using PJSUA via `pjsua_player_set_pos()` and `pjsua_player_get_pos()`.
But reading through the code I wondered whether this is threadsafe at all.
Both functions manipulate the current position of the player,
but so does the `get_frame()` callback from the confbridge.
Because there is no locking, `file_reader` members like `fpos`, `data_left`, `eof` and others can be accessed simultaneously; possibly causing issues.

To be clear; I haven't encountered a specific problem so far, but that might just be luck.
Or is there anything I'm not seeing? I didn't find any hints in the docs, regarding this issue.

### Steps to reproduce

Read the code

### PJSIP version

2.15.1

### Context

Code review

### Log, call stack, etc

```shell
No logs
```

andreas-wehrmann:
This PR adds `pjsua_conf_configure_port()` to PJSUA as a thin wrapper for `pjmedia_conf_configure_port()`
and updates the documentation for the WAV player port functions `pjmedia_wav_player_port_set_pos()` and `pjmedia_wav_player_port_get_pos()`.

Both WAV player functions access the underlying port without synchronization which means a race condition will occur when the port is being serviced (i.e. `put_frame()` is being called by the conference bridge).

One can go around this problem by telling the confbridge to stop reading from the port via `pjmedia_conf_configure_port()` then get or set the position and then call `pjmedia_conf_configure_port()` again,
this time instructing the confbridge to continue reading from the port.
This works because port configuration happens synchronously.

Care must still be taken when an EOF callback has been specified.
My personal advise would be to try and avoid calling `get_pos()` `set_pos()` in the EOF callback.

Resolves #4431

### DIFF
--- a/pjmedia/include/pjmedia/wav_port.h
+++ b/pjmedia/include/pjmedia/wav_port.h
@@ -128,6 +128,10 @@ PJ_DECL(pj_ssize_t) pjmedia_wav_player_get_len(pjmedia_port *port);
 
 /**
  * Set the file play position of WAV player.
+ * @b Safety: This function can only be called when the player is stopped
+ * (that is: get_frame() is not being called).
+ *
+ * @sa pjmedia_conf_configure_port
  *
  * @param port          The file player port.
  * @param offset        Playback position in bytes, relative to the start of
@@ -141,6 +145,10 @@ PJ_DECL(pj_status_t) pjmedia_wav_player_port_set_pos( pjmedia_port *port,
 
 /**
  * Get the file play position of WAV player, in bytes.
+ * @b Safety: This function can only be called when the player is stopped
+ * (that is: get_frame() is not being called).
+ *
+ * @sa pjmedia_conf_configure_port
  *
  * @param port          The file player port.
  *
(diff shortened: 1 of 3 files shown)


pjmedia/include/pjmedia/wav_port.h
```diff
@@ -128,6 +128,10 @@ PJ_DECL(pj_ssize_t) pjmedia_wav_player_get_len(pjmedia_port *port);
 
 /**
  * Set the file play position of WAV player.
+ * @b Safety: This function can only be called when the player is stopped
```
sauwming, line 131: I prefer the term "not running", instead of "stopped", because the latter implies that we need to explicitly stop the player first. While actually as long as it's not running, it should be okay, such as during creation.

andreas-wehrmann: Agreed and updated

sauwming (approved):

nanangizz (approved):

PR #4446, opened 2025-05-30
Add callback to notify succesfull completion of conf/vid_conf operation

trengginas:
Since #3928, conference/video conference is done asynchronously.
The changes introduce other implications, consider the scenario:
- App connect the port1 and port2 (`pjsua_conf_connect()`)
- App immediately checks the port information after the above API returns (`pjsua_conf_get_port_info()`).

After the patch, the information from (`pjsua_conf_get_port_info()`) might not be the latest, since the operation is not 
completed.

This PR will add a callback to notify when the operation is completed.

### DIFF
--- a/pjmedia/include/pjmedia/conference.h
+++ b/pjmedia/include/pjmedia/conference.h
@@ -90,7 +90,7 @@ typedef struct pjmedia_conf_port_info
 } pjmedia_conf_port_info;
 
 /** 
- * Synchronized operation type enumeration.
+ * Conference operation type enumeration.
  */
 typedef enum pjmedia_conf_op_type
 {
@@ -122,7 +122,7 @@ typedef enum pjmedia_conf_op_type
 } pjmedia_conf_op_type;
 
 /**
- * Synchronized operation parameter.
+ * Conference operation parameter.
  */
 typedef union pjmedia_conf_op_param
 {
@@ -188,10 +188,10 @@ typedef struct pjmedia_conf_op_info
 } pjmedia_conf_op_info;
 
 /**
-  * The callback type to be called upon the successful completion 
-  * of a conference port operation.
+  * The callback type to be called upon the completion of a conference 
+  * port operation.
   *
-  * @param param      The conference op callback param.
+  * @param info      The conference op callback param.
   *
   */
 typedef void (*pjmedia_conf_op_cb)(const pjmedia_conf_op_info *info);
@@ -294,7 +294,8 @@ PJ_DECL(pj_status_t) pjmedia_conf_destroy( pjmedia_conf *conf );
  * thus application must not perform long/blocking processing in this callback.
  * 
  * @param conf          The conference bridge.
- * @param cb            Callback to be called.
+ * @param cb            Callback to be called. Set this to NULL to unregister
+ *                      the callback.
  *
  * @return              PJ_SUCCESS on success.
  */
@@ -347,8 +348,7 @@ PJ_DECL(pj_status_t) pjmedia_conf_set_port0_name(pjmedia_conf *conf,
  * port, allowing media to flow to/from the port.
  * 
  * This operation executes asynchronously, use the callback set from
- * #pjmedia_conf_set_op_cb() to receive notification upon succesfull
- * completion.
+ * #pjmedia_conf_set_op_cb() to receive notification upon completion.
  *
  * @param conf          The conference bridge.
  * @param pool          Pool to allocate buffers for this port.
@@ -451,8 +451,7 @@ PJ_DECL(pj_status_t) pjmedia_conf_configure_port( pjmedia_conf *conf,
  * pjmedia_conf_adjust_tx_level().
  * 
  * This operation executes asynchronously, use the callback set from
- * #pjmedia_conf_set_op_cb() to receive notification upon succesfull
- * completion.
+ * #pjmedia_conf_set_op_cb() to receive notification upon completion.
  *
  * @param conf          The conference bridge.
  * @param src_slot      Source slot.
@@ -479,8 +478,7 @@ PJ_DECL(pj_status_t) pjmedia_conf_connect_port( pjmedia_conf *conf,
  * Note that the operation will be done asynchronously, so application
  * should not assume that the port will no longer receive/send audio frame
  * after this function has returned. Use the callback set from
- * #pjmedia_conf_set_op_cb() to receive notification upon succesfull
- * completion.
+ * #pjmedia_conf_set_op_cb() to receive notification upon completion.
  *
  * @param conf          The conference bridge.
  * @param src_slot      Source slot.
@@ -499,8 +497,7 @@ PJ_DECL(pj_status_t) pjmedia_conf_disconnect_port( pjmedia_conf *conf,
  * Note that the operation will be done asynchronously, so application
  * should not assume that the port will no longer receive/send audio frame
  * after this function has returned. Use the callback set from
- * #pjmedia_conf_set_op_cb() to receive notification upon succesfull
- * completion.
+ * #pjmedia_conf_set_op_cb() to receive notification upon completion.
  *
  * @param conf          The conference bridge.
  * @param sink_slot     Sink slot.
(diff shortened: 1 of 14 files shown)


pjmedia/src/pjmedia/conference.c
```diff
@@ -364,23 +335,23 @@ static void handle_op_queue(pjmedia_conf *conf)
...
         /* Free op */
         pj_list_erase(op);
-        op->type = OP_UNKNOWN;
+        op->type = PJMEDIA_CONF_OP_UNKNOWN;
         pj_list_push_back(conf->op_queue_free, op);
 
         pj_mutex_unlock(conf->mutex);
 
         /* Process op */
         switch(type) {
```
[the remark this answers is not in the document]
trengginas: Some operations are called from other method. e.g.: `op_disconnect_ports()` are called from `op_remove_port()` or `op_remove_port()` from `pjmedia_conf_destroy()`.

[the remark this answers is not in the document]
trengginas: If I understand this correctly, then this callback is called for operations that are explicitly invoked by users.
Operation that is not explicitly called will not get notified by this callback.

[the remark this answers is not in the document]
nanangizz: Yes, no callback there. Or just add some callback invocations in `conf_destroy()`. Either way sounds okay to me, also better to mention this in the docs.

pjmedia/src/pjmedia/conference.c
```diff
@@ -892,9 +881,22 @@ PJ_DEF(pj_status_t) pjmedia_conf_destroy( pjmedia_conf *conf )
...
 
+    PJ_LOG(4, (THIS_FILE, "Audio conference bridge destroyed"));
+
     return PJ_SUCCESS;
 }
 
+PJ_DEF(pj_status_t) pjmedia_conf_set_op_cb(pjmedia_conf *conf,
+                                           pjmedia_conf_op_cb cb)
+{
+    PJ_ASSERT_RETURN(conf && cb, PJ_EINVAL);
```
nanangizz: What about allowing `NULL` cb to remove any existing callback?

pjmedia/include/pjmedia/conference.h
```diff
@@ -180,6 +273,20 @@ PJ_DECL(pj_status_t) pjmedia_conf_create( pj_pool_t *pool,
  */
 PJ_DECL(pj_status_t) pjmedia_conf_destroy( pjmedia_conf *conf );
 
+/**
+ * Register the callback to be called when a port operation has been successfully
+ * completed.
```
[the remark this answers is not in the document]
nanangizz: The docs need to be updated too?

pjmedia/src/pjmedia/conference.c
```diff
@@ -1470,10 +1479,12 @@ static void op_disconnect_ports(pjmedia_conf *conf,
...
             for (j=src_port->listener_cnt-1; j>=0; --j) {
                 if (src_port->listener_slots[j] == sink_slot) {
-                    op_param op_prm = {0};
+                    pjmedia_conf_op_param op_prm = {0};
                     op_prm.disconnect_ports.src = i;
                     op_prm.disconnect_ports.sink = sink_slot;
-                    op_disconnect_ports(conf, &op_prm);
+                    status = op_disconnect_ports(conf, &op_prm);
+                    if (status != PJ_SUCCESS)
+                        return status;
```
nanangizz: I think failures better be ignored here, just print logs perhaps?
Also below, and perhaps in video.

pjmedia/src/pjmedia/conference.c
```diff
@@ -1769,13 +1787,17 @@ static void op_remove_port(pjmedia_conf *conf, const op_param *prm)
     pj_bzero(&op_prm, sizeof(op_prm));
     op_prm.disconnect_ports.src = INVALID_SLOT;
     op_prm.disconnect_ports.sink = port;
-    op_disconnect_ports(conf, &op_prm);
+    status = op_disconnect_ports(conf, &op_prm);
+    if (status != PJ_SUCCESS)
+        return status;
```
nanangizz: Better ignore failure here? Also below, and in video.

pjmedia/include/pjmedia/conference.h
```diff
@@ -89,6 +89,112 @@ typedef struct pjmedia_conf_port_info
     int                 rx_adj_level;       /**< Rx level adjustment.       */
 } pjmedia_conf_port_info;
 
+/** 
+ * Synchronized operation type enumeration.
```
sauwming: I'm not sure what the word "synchronized" here is supposed to mean, since the operation is asynchronous?

Also below (parameter), and in vid conf.

pjmedia/include/pjmedia/conference.h
```diff
@@ -225,6 +345,10 @@ PJ_DECL(pj_status_t) pjmedia_conf_set_port0_name(pjmedia_conf *conf,
  * Once the media port is connected to other port(s) in the bridge,
  * the bridge will continuosly call get_frame() and put_frame() to the
  * port, allowing media to flow to/from the port.
+ * 
+ * This operation executes asynchronously, use the callback set from
+ * #pjmedia_conf_set_op_cb() to receive notification upon succesfull
```
sauwming: successful.
Also below.

pjmedia/include/pjmedia/conference.h
```diff
@@ -89,6 +89,112 @@ typedef struct pjmedia_conf_port_info
...
+     */
+    pjmedia_conf_op_param   op_param;
+
+} pjmedia_conf_op_info;
+
+/**
+  * The callback type to be called upon the successful completion 
+  * of a conference port operation.
+  *
+  * @param param      The conference op callback param.
```
sauwming: incorrect parameter.
Also in vid conf.

pjsip-apps/src/pjsua/pjsua_app.c
```diff
@@ -1537,6 +1537,7 @@ static pj_status_t app_init(void)
     app_config.cfg.cb.on_snd_dev_operation = &on_snd_dev_operation;
     app_config.cfg.cb.on_call_media_event = &on_call_media_event;
     app_config.cfg.cb.on_ip_change_progress = &on_ip_change_progress;
+
```
sauwming: unnecessary

pjmedia/include/pjmedia/conference.h
```diff
@@ -180,6 +286,20 @@ PJ_DECL(pj_status_t) pjmedia_conf_create( pj_pool_t *pool,
  */
 PJ_DECL(pj_status_t) pjmedia_conf_destroy( pjmedia_conf *conf );
 
+/**
+ * Register the callback to be called when a port operation has been
+ * completed.
+ * 
+ * The callback will most likely be called from media threads,
+ * thus application must not perform long/blocking processing in this callback.
+ *
```
sauwming, line 293: Add doc on how to remove/unset the callback.

pjsip/include/pjsua2/endpoint.hpp
```diff
@@ -482,6 +482,52 @@ struct OnRejectedIncomingCallParam
...
+     */
+    pjmedia_conf_op_type opType;
+
+    /**
+     * Represents the AudioMedia associated with the operation.
+     * For operations involving multiple AudioMedia instances
+     * (e.g., startTransmit/stopTransmit), the first AudioMedia serves
+     * as the source.
+     */
+    AudioMediaVector2    opData;
```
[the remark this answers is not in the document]
nanangizz: It is no longer AudioMediaVector but still a vector, unsigned vector :)
The param is not an array of a specific type, what if in the future there is a new field that is not an int/unsigned.
If union is a problem (in SWIG perhaps), IMO, using a flat struct/class is better than a vector, e.g:
```
struct OpParam {
  unsigned mediaId;
  unsigned targetMediaId; /* Only applicable for op: ... */
  int adjustLevel; /* Only applicable for op: ... */
};
```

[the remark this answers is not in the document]
nanangizz: When using `union` is chosen, please make sure it works with SWIG (java, python, C#).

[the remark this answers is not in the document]
trengginas: actually `MediaEventData` is using union ([ref](https://github.com/pjsip/pjproject/blob/a531322baff190348ed699dfdc94972b9c1c0060/pjsip/include/pjsua2/media.hpp#L2821)). Tested here using java, and it works

nanangizz (approved):

PR #4459, opened 2025-06-10
AVI player API for PJSUA/PJSUA2 API

trengginas:
This patch will add AVI player API for `PJSUA/PJSUA2`.
The player will read the audio/video streams in an AVI file, create a virtual video device and create/add audio/video port based 
on the streams contained in the file.

Note:
To convert a video to the supported format (video:YUY2/I420/RGB24, audio:16bit PCM), you can use ffmpeg.
```
ffmpeg -i [input.mp4] -vf scale=320:240 -c:v rawvideo -pix_fmt yuv420p -c:a pcm_alaw -ar 16000 [output.avi]
```

### DIFF
--- a/pjmedia/src/pjmedia-videodev/avi_dev.c
+++ b/pjmedia/src/pjmedia-videodev/avi_dev.c
@@ -276,7 +276,7 @@ static void reset_dev_info(struct avi_dev_info *adi)
     }
 
     if (adi->pool)
-        pj_pool_release(adi->pool);
+        pj_pool_safe_release(&adi->pool);
 
     pj_bzero(adi, sizeof(*adi));
 
@@ -671,8 +671,10 @@ static pj_status_t avi_dev_strm_destroy(pjmedia_vid_dev_stream *strm)
 
     avi_dev_strm_stop(strm);
 
-    stream->adi->strm = NULL;
-    stream->adi = NULL;
+    if (stream->adi) {
+        stream->adi->strm = NULL;
+        stream->adi = NULL;
+    }
     pj_pool_release(stream->pool);
 
     return PJ_SUCCESS;
--- a/pjmedia/src/pjmedia-videodev/avi_dev.c
+++ b/pjmedia/src/pjmedia-videodev/avi_dev.c
@@ -276,7 +276,7 @@ static void reset_dev_info(struct avi_dev_info *adi)
     }
 
     if (adi->pool)
-        pj_pool_safe_release(&adi->pool);
+        pj_pool_release(adi->pool);
 
     pj_bzero(adi, sizeof(*adi));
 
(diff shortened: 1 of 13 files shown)


pjmedia/src/pjmedia-videodev/avi_dev.c
```diff
@@ -276,7 +276,7 @@ static void reset_dev_info(struct avi_dev_info *adi)
     }
 
     if (adi->pool)
-        pj_pool_release(adi->pool);
+        pj_pool_safe_release(&adi->pool);
```
nanangizz: Minor: shouldn't the `bzero()` below also set `pool` value to `NULL` already?

pjmedia/src/pjmedia-videodev/avi_dev.c
```diff
@@ -671,8 +671,10 @@ static pj_status_t avi_dev_strm_destroy(pjmedia_vid_dev_stream *strm)
 
     avi_dev_strm_stop(strm);
 
-    stream->adi->strm = NULL;
-    stream->adi = NULL;
+    if (stream->adi) {
+        stream->adi->strm = NULL;
+        stream->adi = NULL;
+    }
```
nanangizz: Minor: the original codes seem just fine, why adding the check? for special treatments, usually comments can be helpful for future ref.

nanangizz (commented):
Just a couple minors.
Also please check the coverity scan reports (perhaps from another PR though).

pjsip/src/pjsua-lib/pjsua_vid.c
```diff
@@ -177,7 +177,12 @@ pj_status_t pjsua_vid_subsys_destroy(void)
         }
     }
 
-    /* Destroy avi file players */
```
nanangizz: For future ref, better put comment about why the vid-conf need to be destroyed before AVI player, etc?

nanangizz (approved):

pjsip/src/pjsua-lib/pjsua_core.c
```diff
@@ -73,6 +73,10 @@ static void init_data()
         pjsua_vid_win_reset(i);
     }
 #if PJSUA_HAS_VIDEO
+    for (i = 0; i < PJ_ARRAY_SIZE(pjsua_var.avi_player); ++i) {
```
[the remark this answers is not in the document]
trengginas: It's already set from `pj_bzero()` above.

sauwming:
> Comments suppressed due to low confidence (1)
> **pjsip/include/pjsua-lib/pjsua.h:8516**
> 
> * The function signature uses pjsua_player_id instead of pjsua_avi_player_id for the AVI player ID, which can cause type confusion. Update the parameter type to pjsua_avi_player_id to match the intended API.
> 
> ```
> PJ_DECL(pjsua_conf_port_id) pjsua_avi_player_get_conf_port(pjsua_player_id id,
> ```

This seems valid.

sauwming (approved):

PR #4460, opened 2025-06-11
Set default timeout for TURN TLS

sauwming:
Currently the timeout value in TURN TLS config is left unset (zero), and this may cause ICE to appear stalled for a long time when waiting for TURN TLS connection (in my local test, by default it times out only after 60 seconds).

### DIFF
--- a/pjnath/include/pjnath/config.h
+++ b/pjnath/include/pjnath/config.h
@@ -230,10 +230,10 @@
 /**
  * Default timeout value (in seconds) for TURN TLS connection.
  *
- * Default: 5 seconds
+ * Default: 10 seconds
  */
 #ifndef PJ_TURN_TLS_DEFAULT_TIMEOUT
-#   define PJ_TURN_TLS_DEFAULT_TIMEOUT              5
+#   define PJ_TURN_TLS_DEFAULT_TIMEOUT              10
 #endif
 
 /* **************************************************************************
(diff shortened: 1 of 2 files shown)


pjnath/include/pjnath/config.h
```diff
@@ -227,6 +227,15 @@
 #   define PJ_TURN_MAX_TCP_CONN_CNT                 8
 #endif
 
+/**
+ * Default timeout value (in seconds) for TURN TLS connection.
+ *
+ * Default: 5 seconds
```
nanangizz: - What about TURN allocation, will it need timeout setting too? Or is this applicable for allocation too?
- RFC6062 (TURN TCP, haven't checked TURN TLS rfc) section 5.2 & 5.3 specifies that the timeout seems to be around 30 seconds. In real world, perhaps 30 seconds is too long, but 5 seconds is perhaps too short? Note that connection need peer permisison? What about 10 seconds?

sauwming: No, this is only for connection timeout. Specifically, for `tls_cfg->ssock_param.timeout`.

OK, will change to 10 sec.

sauwming: For TURN allocation, there seems to already be a timeout mechanism in `turn_session` once the allocation succeeds.

nanangizz: Ah okay, the timeout is just for the SSL socket. So IMO "TURN TLS connection" phrase may be rather ambiguous with TURN connect (see `pj_turn_session_connect()`), to avoid ambiguity perhaps use something like `TLS socket connect timeout in contacting TURN server`, and/or change the macro name to something like `PJ_TURN_SSL_SOCK_DEFAULT_TIMEOUT`?

nanangizz (approved):

PR #4465, opened 2025-06-17
High performance conference bridge with backwards compatibility.

regularex:
Added multi-threading to the conference bridge data plane with phase sync for media proc ordering.

Main Changes

Threading

- Added a thread pool that processes ports in parallel
- Used barriers to sync the threads (they all wait for each other before moving to the next phase)
- Each worker thread gets assigned to a specific CPU core if the OS supports it
- The main thread coordinates everything and handles the final output

Work Distribution

- Started with just dividing ports equally among workers
- Added work stealing later because some workers would finish early and sit idle
- Workers can now steal work from each other's queues when they run out
- Also added periodic rebalancing based on how long each worker takes

Performance Optimizations

- Added SIMD support using AVX2 for mixing and level adjustments (way faster)
- Made sure memory is aligned properly for cache performance
- Added padding between worker data structures to avoid false sharing
- Tried to add NUMA support for big servers with multiple memory nodes

Basics

- Main thread signals all workers to start
- Each worker processes its assigned ports (mixing audio from sources)
- Workers hit a barrier and wait for everyone to finish mixing
- Then all workers write output to their ports
- Main thread combines everything for the sound device

Configuration
Added a bunch of compile flags:

- PJMEDIA_CONF_USE_MULTI_THREADING - turns the whole thing on/off
- PJMEDIA_CONF_THREAD_POOL_SIZE - how many worker threads (default 8)
- PJMEDIA_CONF_USE_SIMD - enables AVX2 optimizations
- PJMEDIA_CONF_USE_WORK_STEALING - enables the work stealing queues
- PJMEDIA_CONF_USE_NUMA - NUMA optimizations (probably overkill)

Work Stealing 

Only active when work imbalance exceeds threshold - default 25% queue capacity

- Each worker has a lock-free queue
- When a worker runs out of work, it tries to steal from others
- Used atomic operations 
- Had to add memory barriers for correctness
- Tracks statistics like steal attempts and success rate

Platform-Specific Code
Linux: Uses pthread for affinity, /sys filesystem for CPU topology
Windows: Uses Windows API for processor groups
macOS: Basic pthread support, no NUMA

Issues

- The barrier implementation was required but not all platforms have it
- SIMD code needs runtime CPU detection to avoid crashes
- Work stealing can actually make things slower if there's not enough work
- NUMA support is probably unnecessary for most use cases

Backward Compatibility
Everything is behind compile flags so if you don't enable multi-threading, it works exactly like before - no API changes needed.

What Could Be Better

- The rebalancing algorithm is pretty basic
- Could probably use better heuristics for work stealing
- The NUMA code is barely tested
- Should probably make thread count auto-detect based on CPU cores

### DIFF
--- a/pjmedia/include/pjmedia/conference.h
+++ b/pjmedia/include/pjmedia/conference.h
@@ -1,4 +1,4 @@
-/* 
+/*
  * Copyright (C) 2008-2011 Teluu Inc. (http://www.teluu.com)
  * Copyright (C) 2003-2008 Benny Prijono <benny@prijono.org>
  *
@@ -14,7 +14,7 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program; if not, write to the Free Software
- * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA 
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 #ifndef __PJMEDIA_CONF_H__
 #define __PJMEDIA_CONF_H__
@@ -90,7 +90,7 @@ typedef struct pjmedia_conf_port_info
     int                 rx_adj_level;       /**< Rx level adjustment.       */
 } pjmedia_conf_port_info;
 
-/** 
+/**
  * Conference operation type enumeration.
  */
 typedef enum pjmedia_conf_op_type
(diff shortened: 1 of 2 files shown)


pjmedia/include/pjmedia/conference.h
```diff
@@ -1,4 +1,4 @@
-/*
```
sauwming, line 1: I would suggest to avoid all these whitespace modifications to make the review much easier.

If it's not possible, you can make a separate PR instead specifically for this formatting fix, so this PR will be clean.

PR #4440, opened 2025-05-24
Synchronization barriers

LeonidGoltsblat:
This PR provides abstraction to pj_barrier_t - synchronization barrier.
It allows threads to block until all participating threads have reached the barrier, ensuring synchronization at specific points in execution. 
pj_barrier_t provides a barrier mechanism for synchronizing threads in a multithreaded application, similar to the POSIX pthread_barrier_wait or Windows EnterSynchronizationBarrier.

This functionality is extracted from #4241 without any changes.

PR #4468, opened 2025-06-17
High performance conference bridge with backwards compatibility.

regularex:
Added multi-threading to the conference bridge data plane with phase sync for media proc ordering.

Main Changes

Threading

- Added a thread pool that processes ports in parallel
- Used barriers to sync the threads (they all wait for each other before moving to the next phase)
- Each worker thread gets assigned to a specific CPU core if the OS supports it
- The main thread coordinates everything and handles the final output

Work Distribution

- Started with just dividing ports equally among workers
- Added work stealing later because some workers would finish early and sit idle
- Workers can now steal work from each other's queues when they run out
- Also added periodic rebalancing based on how long each worker takes

Performance Optimizations

- Added SIMD support using AVX2 for mixing and level adjustments (way faster)
- Made sure memory is aligned properly for cache performance
- Added padding between worker data structures to avoid false sharing
- Tried to add NUMA support for big servers with multiple memory nodes

Basics

- Main thread signals all workers to start
- Each worker processes its assigned ports (mixing audio from sources)
- Workers hit a barrier and wait for everyone to finish mixing
- Then all workers write output to their ports
- Main thread combines everything for the sound device

Configuration
Added a bunch of compile flags:

- PJMEDIA_CONF_USE_MULTI_THREADING - turns the whole thing on/off
- PJMEDIA_CONF_THREAD_POOL_SIZE - how many worker threads (default 8)
- PJMEDIA_CONF_USE_SIMD - enables AVX2 optimizations
- PJMEDIA_CONF_USE_WORK_STEALING - enables the work stealing queues
- PJMEDIA_CONF_USE_NUMA - NUMA optimizations (probably overkill)

Work Stealing 

Only active when work imbalance exceeds threshold - default 25% queue capacity

- Each worker has a lock-free queue
- When a worker runs out of work, it tries to steal from others
- Used atomic operations 
- Had to add memory barriers for correctness
- Tracks statistics like steal attempts and success rate

Platform-Specific Code
Linux: Uses pthread for affinity, /sys filesystem for CPU topology
Windows: Uses Windows API for processor groups
macOS: Basic pthread support, no NUMA

Issues

- The barrier implementation was required but not all platforms have it
- SIMD code needs runtime CPU detection to avoid crashes
- Work stealing can actually make things slower if there's not enough work
- NUMA support is probably unnecessary for most use cases

Backward Compatibility
Everything is behind compile flags so if you don't enable multi-threading, it works exactly like before - no API changes needed.

What Could Be Better

- The rebalancing algorithm is pretty basic
- Could probably use better heuristics for work stealing
- The NUMA code is barely tested
- Should probably make thread count auto-detect based on CPU cores

### DIFF
--- a/pjmedia/src/pjmedia/conference.c
+++ b/pjmedia/src/pjmedia/conference.c
@@ -1,4 +1,4 @@
-/* 
+/*
  * Copyright (C) 2008-2011 Teluu Inc. (http://www.teluu.com)
  * Copyright (C) 2003-2008 Benny Prijono <benny@prijono.org>
  *
@@ -14,7 +14,7 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with this program; if not, write to the Free Software
- * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA 
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 #include <pjmedia/conference.h>
 #include <pjmedia/alaw_ulaw.h>
@@ -25,6 +25,7 @@
 #include <pjmedia/silencedet.h>
 #include <pjmedia/sound_port.h>
 #include <pjmedia/stereo.h>
+#include <pj/os.h>
 #include <pj/array.h>
 #include <pj/assert.h>
 #include <pj/log.h>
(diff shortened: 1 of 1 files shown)


sauwming:
Thank you for submitting the PR. Improving the performance of the PJSIP conference bridge has been a topic of growing interest recently.

Since there is already an ongoing PR at #4241, could you clarify how this PR relates to that one before we proceed with the review?

regularex:
They are unrelated, although there may be some crossover in some of the obvious logic requirement. My focus was on performance with a low-to-no risk backward compatibility. 4241 has no SIMD support which would be a huge performance gain overall. My PR has more extensive dynamic rebalancing load distribution, statistical port adjustment based on load, tracks time per worker, NUMA support and graceful feature degradation when unavailable. It is modular, loosely coupled with separation of concerns, more tunable parameters with built-in fallback mechanisms. 

Frankly, we also needed this now and do not wish to maintain a separate codebase while 4241 is still in development. There are no external dependencies and has been tested with 2400 calls on an 8 core system with ~30% CPU utilization.

I also do not mind jumping in on 4241 later, but we are going to be blocked until this is dealt with and feel it is low risk.

sauwming:
There appears to be much overlap between this PR and 4241, and since 4241 has been started much earlier and is now in the final stage of review, I would recommend to wait until 4241 has been merged first. Once done, you can check the new implementation of multithreaded conference, which will now be put in a separate file `conf_thread.c`, and see if there's further improvements/optimizations required.

regularex:
Okay, agreed there is some overlap, but regardless, we can use ours in this hash for now, and if there is more to do when this gets merged, we can jump in. I will look that over and make new PRs if needed. Thanks.

LeonidGoltsblat:
You’ve done an excellent and very interesting job! The “2400 calls” result is truly impressive!

Unfortunately, I wasn’t able to compile your code (with PJMEDIA_CONF_USE_MULTI_THREADING and PJMEDIA_CONF_USE_SIMD enabled) under Visual Studio, so I couldn’t run it on my test setup. I would greatly appreciate it if you could try running my code (PR #4241) on your setup, with auto-vectorization enabled for conf_thread.c via compiler options.

You might also find helpful some of the features I introduced during work on PR #4241 — such as support for barriers (#4440) and aligned memory allocation (#4360, etc.).

sauwming:
With the merging of #4241, this will obviously needs modifications to resolve the conflicts. Please feel free to create a new PR if there are further improvements in the conference that hasn't been covered. Thanks.

PR #4470, opened 2025-06-20
chore: allow ka_cfg to be modifiable through the callback on_stream_p…

impegoraro:
Allows the keepalive configuration to be changed with the on_stream_precreate callback. Seems like it was forgotten when the support to enable/disable keepalive through the callback was added.

### DIFF
--- a/pjsip/include/pjsua2/call.hpp
+++ b/pjsip/include/pjsua2/call.hpp
@@ -698,7 +698,7 @@ struct StreamInfo
     /**
      * Keepalive interval after the stream is created.
      */
-    unsigned startIntervalKA
+    unsigned startIntervalKa;
 
     /**
      *  Keepalive sending interval.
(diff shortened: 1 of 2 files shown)


nanangizz:
I don't think it was forgotten, the callback [docs](https://docs.pjsip.org/en/latest/api/generated/pjsip/group/group__PJSUA__LIB__BASE.html#_CPPv4N14pjsua_callback19on_stream_precreateE) does not specify `ka_cfg` in the changable params.

Anyway, IMO it should be okay to add `ka_cfg`, just a few more things need to be done:
- Update the callback `on_stream_precreate` docs to include `ka_cfg`.
- Also update the code for other stream types (video & text).

nanangizz (approved):

sauwming (approved):

nanangizz:
Sorry I forgot that PJSUA2 needs to be updated too:
- add `ka_cfg` to StreamInfo (can be flattened/not-a-struct, so 3 unsigned fields), also update the `fromPj()` method.
- update `Endpoint::on_stream_precreate()`.
- update the docs of `Call::onStreamPreCreate()` to include the 3 new fields.

pjsip/include/pjsua2/call.hpp
```diff
@@ -689,6 +689,21 @@ struct StreamInfo
...
+
+    /**
+     *  Number of keepalive messages to be sent
+     */
+    unsigned            startCountKa;
+
+    /**
+     * Keepalive interval after the stream is created.
+     */
+    unsigned startIntervalKA
```
nanangizz: just a minor: shouldn't it be `startIntervalKa`?

sauwming: also missing semicolon.
Make sure you try to build it with `PJMEDIA_STREAM_ENABLE_KA` enabled.

impegoraro: oh right, thanks completely forgot about defining PJMEDIA_STREAM_ENABLE_KA.

Issue #4480, opened 2025-06-30
Assertion failure with wave file 22050Hz, ptime 10 and speex resampler.


dhoesele:
### Describe the bug

I'm encountering an assertion failure when using the Speex resampler in combination with a ptime and audio frame ptime of 10ms and a WAV file sampled at 22050 Hz and clock rate 8000. The issue arises during playback and occurs in debug builds due to the following assertion:

pj_assert(cport->rx_buf_count <= cport->rx_buf_cap);  
cport->rx_buf_count = 438, 
cport->rx_buf_cap = 418

Following wave file was used
[input.22050.zip](https://github.com/user-attachments/files/20980857/input.22050.zip)

### Steps to reproduce

1. Clone the latest main branch of pjproject.
2. Set PJSUA_DEFAULT_AUDIO_FRAME_PTIME 10
3. Build the pjsua application Debug Win32.
4. WAV file sampled at 22050 Hz (I used mono).
5. Run with following starting arguments: --ptime=10 --play-file=input.22050.wav --clock-rate=8000
6. Connect the wave port to any other port ( i used Master/sound port )
7. assertion failure occurs in the code handling resampling/buffer management.

### PJSIP version

master (commit b7cfd79b12056c5616aa6ba530f8d976ecf5b8ce)

### Context

- Windows 10
- 32 or 64 bit
- Visual Studio 2022
- Speex resampler
- Wave file with 22050 Hz sample rate
- Adjusting the ptime or using files with different sample rates (e.g., 8000 or 16000 Hz) avoids the issue.

### Log, call stack, etc

```shell
15:02:28.754 sip_endpoint.c !.Module "mod-pjsua-log" registered
15:02:28.754 sip_endpoint.c  .Module "mod-tsx-layer" registered
15:02:28.754 sip_endpoint.c  .Module "mod-stateful-util" registered
15:02:28.754 sip_endpoint.c  .Module "mod-ua" registered
15:02:28.755 sip_endpoint.c  .Module "mod-100rel" registered
15:02:28.755 sip_endpoint.c  .Module "mod-pjsua" registered
15:02:28.756 sip_endpoint.c  .Module "mod-invite" registered
15:02:28.756     null_dev.c  ..null audio initialized
15:02:28.759          pjlib  ..select() I/O Queue created (03B3406C)
15:02:28.761 sip_endpoint.c  .Module "mod-evsub" registered
15:02:28.761 sip_endpoint.c !.Module "mod-presence" registered
[...]

Issue #532, opened 2008-05-12
Avoid truncated calculation of samples per frame on some 'odd' clock rates.


pjsipbot:
_2008-05-12 18:03:51: @nanangizz_ _created the issue on trac ticket [532](https://trac.pjsip.org/repos/ticket/532)_

Sample cases:

 1. On clock rate 11025, when ptime is set to 10ms, number of samples per frame will be 110.25 which will be truncated to 110.
 2. When a port, which works on clock rate 11025Hz ptime 40ms (which is safe from truncation), is added to a conference bridge that works on clock rate 16000Hz ptime 10ms, the conference will need to resample a frame of 10ms from the port, which can cause truncation.

This may also happen in clock rate 22050Hz.

___
_2008-06-20 09:42:17: @bennylp_ _changed **milestone** from **release-0.9.0** to **Known-Issues**_

___
_2018-02-08 02:12:35: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2018-02-08 02:12:35: @nanangizz_ _set **resolution** to **wontfix**_

___
_2018-02-08 02:12:35: @nanangizz_ _commented_

TeluuCon 2018: only applicable to PCM

PR #4482, opened 2025-07-01
Check for fractional samples per frame

dhoesele:
This pull request aims to fix [issue #4480](https://github.com/pjsip/pjproject/issues/4480).
It addresses cases where certain sample rate and ptime combinations (e.g., 10ms @ 22050Hz = 220.5 samples/frame) are not supported and could cause incorrect behavior. Now, port creation will fail if the samples per frame is not an integer.

Additionally, this PR fixes a possible memory leak by ensuring any allocated resamplers are destroyed if port creation fails.


sauwming:
I think we also need to do a similar check in `pjmedia_wav_player_create()`.

nanangizz:
There was #532, but closed with "won't fix".
Adding explicit checks is definitely a good idea.

Just few minors perhaps:
- Line length <= 80 (see coding style docs [here](https://docs.pjsip.org/en/latest/get-started/coding-style.html#limit-line-length))
- From AI: better add some log & returning `PJMEDIA_ENOTCOMPATIBLE` or `PJMEDIA_ENCSAMPLESPFRAME` instead of `PJ_ENOTSUP`?

pjmedia/src/pjmedia/wav_player.c
```diff
@@ -416,6 +416,21 @@ PJ_DEF(pj_status_t) pjmedia_wav_player_port_create( pj_pool_t *pool_,
...
     /* It seems like we have a valid WAVE file. */
 
+    /* Check compatibility of sample rate and ptime.
+    * Some combinations result in a fractional number of samples per frame
+    * which we do not support.
+    * One such case would be for example 10ms @ 22050Hz which would yield
+    * 220.5 samples per frame.
+    */
+    if (0 != (ptime * wave_hdr.fmt_hdr.sample_rate *
+              wave_hdr.fmt_hdr.nchan % 1000)) {
```
nanangizz: Minor: coding style, brace has its own line for multiple lines statements/conditions. Also another one in `conference.c`.

nanangizz (approved):
Just a couple minors above.

sauwming (approved):

LeonidGoltsblat:
In my opinion, the same should be done in pjmedia_wav_playlist_create() also. Right?

PR #4489, opened 2025-07-04
Fixed Coverity warning in conference

sauwming:
Re #4482, fix warning that `conf_port` may be uninitialized.


nanangizz (approved):

LeonidGoltsblat:
Shouldn't we (with all possible caution) destroy conf_port->port->grp_lock in the same on_return block of create_conf_port()?

sauwming:
> Shouldn't we (with all possible caution) destroy conf_port->port->grp_lock in the same on_return block of create_conf_port()?

Yes, we should. Please check #4497.

PR #4490, opened 2025-07-04
Handle SSRC change after SRTP restart

sauwming:
Continuation of #4469.

In the case of IP change scenario, remote may change SSRC, so after re-INVITE and SRTP restart we may encounter error such as:
`srtp0x115f47400  Failed to unprotect SRTP, pkt size=23, err=no appropriate context found`.

This PR also fixes printing format of SSRC from signed (which can cause SSRC to be printed as negative) to unsigned integer.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp.c
+++ b/pjmedia/src/pjmedia/transport_srtp.c
@@ -986,7 +986,7 @@ static pj_status_t create_srtp_ctx(transport_srtp *srtp,
         err = srtp_set_stream_roc(ctx->srtp_tx_ctx,
                                   setting->tx_roc.ssrc,
                                   setting->tx_roc.roc);
-        PJ_LOG(4, (THIS_FILE, "Initializing SRTP TX ROC to SSRC %d with "
+        PJ_LOG(4, (THIS_FILE, "Initializing SRTP TX ROC to SSRC %u with "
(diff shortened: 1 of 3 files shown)


pjmedia/src/pjmedia/transport_srtp.c
```diff
@@ -1635,22 +1637,59 @@ static void srtp_rtp_cb(pjmedia_tp_cb_param *param)
...
+        need_restart = PJ_TRUE;
+        PJ_LOG(5, (srtp->pool->obj_name, "Restarting SRTP err=%s",
+                   get_libsrtp_errstr(err)));
+    }
+#endif
+
+#if PJMEDIA_SRTP_CHECK_SSRC_ON_RESTART
+    /* This could happen if remote changes ssrc, such as on IP address change
+     * scenario.
+     */
```
nanangizz: It is rather unclear how an IP address change may change the SSRC (AFAIK they are supposed to be independent)?

sauwming: Apparently, some implementation may change SSRC when its transport address changes (https://avt.ietf.narkive.com/l9lwWXId/ssrc-change-in-a-single-rtp-session).

But yes, the correlation may not be obvious, so I just change it to a more common reason for ssrc change.

PR #4493, opened 2025-07-07
Issue 4487: Endpoint initialization fails if one of DNS servers is BAD

alexander-m-ilyin:
Steps to reproduce:

```
...
ep_cfg.uaConfig.nameservers.push_back("some.corrupted.dns");
ep_cfg.uaConfig.nameservers.push_back("8.8.8.8"); // Google DNS
...
endpoint->libInit(ep_cfg); // Endpoint initialization fails with PJLIB_UTIL_EDNSINNSADDR error even if we still have a valid DNS server
...
```

### DIFF
--- a/pjlib-util/src/pjlib-util/resolver.c
+++ b/pjlib-util/src/pjlib-util/resolver.c
@@ -559,12 +559,8 @@ PJ_DEF(pj_status_t) pj_dns_resolver_set_ns( pj_dns_resolver *resolver,
         if (status != PJ_SUCCESS)
             status = pj_sockaddr_init(pj_AF_INET6(), &ns->addr, &servers[i], 
                                       (pj_uint16_t)(ports ? ports[i] : PORT));
-        if (status != PJ_SUCCESS) {
-            pj_grp_lock_release(resolver->grp_lock);
-            return PJLIB_UTIL_EDNSINNSADDR;
-        }
 
-        ns->state = STATE_ACTIVE;
+        ns->state = (status == PJ_SUCCESS) ? STATE_ACTIVE : STATE_BAD;
         ns->state_expiry = now;
         ns->rt_delay.sec = 10;
     }


sauwming:
Please sign the CLA, thanks.

alexander-m-ilyin:
> Please sign the CLA, thanks.

You have agreed to the CLA for pjsip/pjproject
Version: 2020-02-24

I have no idea why github still showing "not signed yet"

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -559,12 +559,8 @@ PJ_DEF(pj_status_t) pj_dns_resolver_set_ns( pj_dns_resolver *resolver,
...
-        if (status != PJ_SUCCESS) {
-            pj_grp_lock_release(resolver->grp_lock);
-            return PJLIB_UTIL_EDNSINNSADDR;
-        }
 
-        ns->state = STATE_ACTIVE;
+        ns->state = (status == PJ_SUCCESS) ? STATE_ACTIVE : STATE_BAD;
         ns->state_expiry = now;
         ns->rt_delay.sec = 10;
     }
```
nanangizz, line 566: Actually [the docs](https://docs.pjsip.org/en/latest/api/generated/pjlib_util/group/group__PJ__DNS__RESOLVER.html#_CPPv422pj_dns_resolver_set_nsP15pj_dns_resolverjA_K8pj_str_tA_K11pj_uint16_t) specifies that:
> If hostname is specified, the hostname must be resolvable with pj_gethostbyname(). 

so the behaviour of this function should already be correct.

Also note that if the failure is ignored, the specified nameserver will still be set but has no IP address (due to invalid IP address or unresolvable). Moreover, when all specified nameservers are invalid or not-resolvable, the function seems to still returning success.

So if the objective is to avoid library initialization failure (due to bad nameserver settings), I think the modification should be inside `pjsua_init()`, i.e: validate & filter the nameservers, before calling `pj_dns_resolver_set_ns()`?

alexander-m-ilyin: @nanangizz

According to the same documentation, the second nameserver (then the third and fourth) should be tried, but this will never happen if the company-specific nameserver1 is not whitelisted/accessible due to the local network firewall/router settings. As a result, we cannot fallback to the Google public DNS 8.8.8.8 in some WiFi networks. 

I agree that the pj_dns_resolver_set_ns() should return success only if at least one nameserver is marked as ACTIVE, so pjsua_init() will throw an error only if all nameservers are BAD.

nanangizz: > According to the same documentation, the second nameserver (then the third and fourth) should be tried, 

Yes, there are two independent things:
1. Initial verification: all configured nameservers must be valid, e.g: it is in a correct format if it is IP address, it is resovable if it is a hostname.
2. Nameserver selection: all configured nameservers will be tried (and marked as ACTIVE, PROBING, or BAD).

Your modification changes the first, the initial verification stage, the result is basially whether the nameservers are valid or invalid. While it looks like that what you need is the latter, which is in the operational stage, the result is nameserver state (ACTIVE, BAD, PROBING).

Note that initially, all nameservers, which have been verified, will be set as ACTIVE but expired so it will be probed immediately.

> 
> I agree that the pj_dns_resolver_set_ns() should return success only if at least one nameserver is marked as ACTIVE, so pjsua_init() will throw an error only if all nameservers are BAD.

For now I am not sure about changing the specification/behaviour of `pj_dns_resolver_set_ns()`.

IMO here are some idea for the issue:
- No lib change needed: the company specific nameserver can be supplied as an IP address rather than a hostname, it will not cause library init to fail (note that the library init only validates the address, not checking the connectivity or responsiveness).
- Lib change needed: add API in PJSUA/PJSUA2 to set the nameservers on run-time, the API will verify and filter so that at least there is a valid nameserver, then it calls `pj_dns_resolver_set_ns()`.

alexander-m-ilyin: _No lib change needed: the company-specific nameserver can be supplied as an IP address rather than a hostname_

**it is an extremely bad idea in the mobile world**, as a result, we will have a mess of mobile clients using obsolete IP addresses. 

_Initial verification: all configured nameservers must be valid, e.g: it is in a correct format if it is IP address, it is resolvable if it is a hostname._

I agree with the format, but not resolvability on the lib initialization stage: is there any reason to have 4 nameservers then? So, if the first nameserver is not resolvable in some crazy network, lib will ignore the other 3 good nameservers.

My suggestion is to use some new API in the pjlib_init(), which will do almost the same as pj_dns_resolver_set_ns() but throw the error only if all nameservers are BAD.

nanangizz: > _No lib change needed: the company-specific nameserver can be supplied as an IP address rather than a hostname_
> 
> **it is an extremely bad idea in the mobile world**, as a result, we will have a mess of mobile clients using obsolete IP addresses.

AFAIK suppying IP address for nameserver is actually quite common and sensible. Also, it does not necessarily mean that the IP address needs to be hardcoded in the app, app can do the preverifications (e.g: resolving using `gethostbyname()/getaddrinfo()`, connectivity test, etc) then pass the result to the library init.



> My suggestion is to use some new API in the pjlib_init(), which will do almost the same as pj_dns_resolver_set_ns() but throw the error only if all nameservers are BAD.

Thank you for the suggestion.

I agree, if it is to be done in the lib, the modification should be adding a new PJSUA/PJSUA2 API that specifically doing the preverification steps, which can be used for setting/updating the nameserver (during/after library is initialized). Currently we don't have a plan to implement this.

Also, the pre-verification can always be done in in the app as described above.

alexander-m-ilyin: Well, I agree with you (about IP addresses resolving on the client side) if we're talking about the IP phones and desktop clients, but not mobile apps, which can change the network every minute, so I was assuming the 4 nameservers array was introduced to have the ability to switch to the any active one at any time.

Ok, in general yes, this PR needs to be rejected.

PR #4494, opened 2025-07-07
Add initial support for build system generation with CMake

aymanalqadhi:
## Why CMake?
* Greatly improves IDE/LSP support. While almost all major IDEs have built-in support for CMake, CMake natively supports generating `compile_commands.json` files used by stand-alone tooling.
* Simplifies the build process.
* Reduces the amount of code to maintain (once this becomes mature enough), as CMake has generators for practically all widely-used build systems. The current custom build configurations for Visual Studio, and Xcode could be automatically generated using CMake.
* Enables faster compilation/recompilation cycles, and may produce potentially faster generated code when used within CMake-based projects.
* Provides better overall integration with modern tooling.

## What does this PR add?
* Near-complete support for building all modules using CMake-generated build systems (including test applications).
* Ability to use either bundled third-party dependencies or system-installed ones.
* A somewhat functional installation process, with proper (I think) handling of dyamic-linking `RPATH` values.
* Ability for building both static and shared libraries.
* Configure-time generation of configuration headers (emulating current build system).
* Cross-Platform building (needs testing). No platform-dependent tools were used, only CMake.
* Options for almost all the things that can be customized in the current build system.

## What is still needed?
* **Extended testing**: Currently only validated (best-effort) on Linux x86_64. Initial groundwork exists for other platforms, but compilation will likely require platform-specific modifications.
* **Documentation support**: Implementation of documentation builds using CMake's built-in Doxygen module (should be relatively straightforward).
* **External dependency integration**: While basic installation and application execution work, full support for using the library as an external (out-of-tree) dependency via CMake-native facilities (e.g., `find_package` in config mode) requires more work.

## Suggestions for improvements
* **Dependency packaging approach**: Why build bundled dependencies as separate libraries? Wouldn't this approach risk having conflicts with existing system-installed versions? Compiling them directly into dependents using CMake's object libraries would (in my opinion) be preferable. This would avoid the potential issues, while relatively simplifying packaging & installation.
* **Dependency include paths**: Some modules reference third-party dependencies through relative paths (e.g., #include "../../third_party/some_dep/file.h"). This prevents transparent interchangeability between system-installed and bundled dependencies, complicating their management. Standardizing include paths would resolve this.


third_party/webrtc_aec3/CMakeLists.txt
```diff
@@ -0,0 +1,205 @@
...
+  src/system_wrappers/source/metrics.cc
+  src/third_party/rnnoise/src/rnn_vad_weights.cc
+  src/third_party/pffft/src/pffft.c
+)
+
+# headers
+file(GLOB_RECURSE WEBRTC_AEC3_HEADERS
+  RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}"
+  LIST_DIRECTORIES FALSE
+  "*.h"
```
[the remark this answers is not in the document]
aymanalqadhi: I used `file(GLOB)` to collect header files because `webrtc_aec3` has 230+ header files, and the module is not maintained by pjproject anyway. The other modules, however, have their headers listed explicitly.

sauwming:
It would be great if you can list which platforms are supported (and which have been tested on), and the build steps for each platform.

aymanalqadhi:
I tested it on Linux x86_64 (as I said above).

Other platforms are not tested. However, there is code written for them, so supporting them should be simple. CMake is cross-platform, so the only missing thing (I hope) should be importing platform-specific dependencies.

### Building steps:
```sh

# create the build tree (`out` is the output directory)
cmake -B out .

# build the project :D
cmake --build out

# (Optional) install the built binaries
cmake --install out --prefix $HOME/.local

```

sauwming:
Thank you for making this.

I tested this on Mac and managed to get until 98%:
```
[ 98%] Linking CXX executable pjsua
  "_AudioComponentFindNext", referenced from:
```
which is to be expected since as you said, the OS-specific dependencies are still missing.

Unfortunately, I don't think we can merge a build system that's not yet complete. Hopefully you (or someone else) can continue this for the other platforms. I'll keep this PR open in the meantime.

a-lerion:
I have built this branch for Windows desktop x64 after some changes and have several questions.

1. Should I push the changes to https://github.com/aymanalqadhi/pjproject/tree/cmake-build for a review?
2. OpenSSL may be used in several libraries depending on options. Would it be better to move it's search to a toplevel cmake file or a new cmake include file? I'm using two find_package calls for this search, one with a forced CONFIG mode search and then the default one if that fails. In my case the default one fails and the CONFIG one succeeds when OpenSSL_DIR is set.
3. Similar question for a math library.
4. Some math library search has the following
```
find_library(MATH_LIBRARY m)
if(NOT MATH_LIBRARY)
  if(MATH_LIBRARY)
```
Line with `if(NOT MATH_LIBRARY)` should not be there, right?

aymanalqadhi:
> 1. Should I push the changes to https://github.com/aymanalqadhi/pjproject/tree/cmake-build for a review?

Sure.

> 2. OpenSSL may be used in several libraries depending on options. Would it be better to move it's search to a toplevel cmake file or a new cmake include file? I'm using two find_package calls for this search, one with a forced CONFIG mode search and then the default one if that fails. In my case the default one fails and the CONFIG one succeeds when OpenSSL_DIR is set.

I think keeping the current calls (currently just two) to `find_package(OpenSSL)` separate would be better, because each library might need a different configuration of `OpenSSL`. For example, if a library (e.g., `srtp`) needed a specific version and/or components other than the ones used by the other libraries, the other libraries using `OpenSSL` would be restricted in terms of what configuration might be used.

> 3. Similar question for a math library.

I tried to keep the libraries as loosely-coupled to each other as possible. Each library "chooses" what it links to. Some libraries require linking to `m`, but others don't.

> 4. Some math library search has the following
> 
> ```
> find_library(MATH_LIBRARY m)
> if(NOT MATH_LIBRARY)
>   if(MATH_LIBRARY)
> ```
> 
> Line with `if(NOT MATH_LIBRARY)` should not be there, right?

You're correct on this one. I just pushed the fix for it.

a-lerion:
>Sure.

@aymanalqadhi Could you add me as a collaborator then?

pjlib/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@
 
 include(Pj/Option)
 include(CMakeDependentOption)
+include(CheckIncludeFile)
```
aymanalqadhi: I think keeping this where it is being used would be clearer

pjlib/CMakeLists.txt
```diff
@@ -229,9 +230,28 @@ if (NOT (WIN32 OR CYGWIN OR MIGW))
...
+  )
+  if (PJ_HAS_WINSOCK2_H)
+    target_link_libraries(pjlib PRIVATE
+      ws2_32
+    )
+  elseif(PJ_HAS_WINSOCK_H)
+    target_link_libraries(pjlib PRIVATE
+      wsock32
+    )
+  endif()
```
aymanalqadhi: There is no need to split the call to `target_link_libraries` if the argument list is short. For this case, using a single line would be more readable

pjlib/CMakeLists.txt
```diff
@@ -347,9 +347,11 @@ block(SCOPE_FOR VARIABLES)
   elseif(WIN32 OR CYGWIN OR MINGW)
     set(PJ_WIN32 1)
     set(PJ_WIN32_WINNT 0x0501)
-    set(WIN32_LEAN_AND_MEAN)
+    set(WIN32_LEAN_AND_MEAN 1)
+    set(NOMINMAX 1)
 
-    if(CMAKE_SYSTEM_NAME MATCHES "_64-w64-mingw")
+    if(CMAKE_SYSTEM_NAME MATCHES "_64-w64-mingw"
+      OR CMAKE_HOST_SYSTEM_PROCESSOR MATCHES "AMD64")
```
aymanalqadhi, line 375: Is the value "AMD64" only used when compiling for MinGW?

pjmedia/CMakeLists.txt
```diff
@@ -732,7 +732,7 @@ add_library(pjmedia-audiodev
   src/pjmedia-audiodev/errno.c
   src/pjmedia-audiodev/null_dev.c
   src/pjmedia-audiodev/oboe_dev.cpp
-  src/pjmedia-audiodev/wasapi_dev.cpp
+  $<$<BOOL:${PJMEDIA_WITH_AUDIODEV_WASAPI}>:src/pjmedia-audiodev/wasapi_dev.cpp>
```
aymanalqadhi: This file is compiled only when `PJMEDIA_AUDIO_DEV_HAS_WASAPI` is set anyway. There is no need for this generator expression.

third_party/webrtc_aec3/CMakeLists.txt
```diff
@@ -169,6 +169,13 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
   )
 endif()
 
+if (MSVC)
+  target_compile_definitions(webrtc_aec3
+    PRIVATE
+      NOMINMAX
+  )
+endif()
+
```
aymanalqadhi: Wouldn't adding this to `cmake/DetectWebRTCParameters.cmake` be a better idea?

pjlib/CMakeLists.txt
```diff
@@ -254,7 +254,10 @@ endif()
 
 # ssl
 if(PJLIB_WITH_SSL STREQUAL "openssl")
-  find_package(OpenSSL COMPONENTS Crypto SSL)
+  find_package(OpenSSL COMPONENTS Crypto SSL CONFIG)
+  if(NOT OPENSSL_FOUND)
+    find_package(OpenSSL COMPONENTS Crypto SSL)
+  endif()
```
aymanalqadhi, line 280: CMake almost does this by default (Module mode then falls back to config mode). There is also `CMAKE_FIND_PACKAGE_PREFER_CONFIG`, which makes CMake start with config mode.

pjmedia/CMakeLists.txt
```diff
@@ -43,6 +43,13 @@ elseif(PJMEDIA_WITH_RESAMPLE STREQUAL "speex")
   endif()
 endif()
 
+if(PJLIB_WITH_SSL STREQUAL "openssl")
+  find_package(OpenSSL COMPONENTS Crypto SSL CONFIG)
+  if(NOT OPENSSL_FOUND)
+      find_package(OpenSSL COMPONENTS Crypto SSL)
+  endif()
+endif()
```
aymanalqadhi: This is redundant, since `pjlib` already finds `OpenSSL`.

pjmedia/CMakeLists.txt
```diff
@@ -546,6 +553,9 @@ target_link_libraries(pjmedia
     $<$<BOOL:${PJMEDIA_WITH_FFMPEG_AVCODEC}>:FFMPEG::avcodec>
     $<$<BOOL:${PJMEDIA_WITH_FFMPEG_AVFORMAT}>:FFMPEG::avformat>
     $<$<BOOL:${PJMEDIA_WITH_FFMPEG_AVDEVICE}>:FFMPEG::avdevice>
+
+    $<$<STREQUAL:${PJLIB_WITH_SSL},openssl>:OpenSSL::Crypto>
```
aymanalqadhi: The condition should be be combined with `PJMEDIA_WITH_SRTP`, since OpenSSL is only needed for SRTP transports.

pjsip/CMakeLists.txt
```diff
@@ -9,6 +9,13 @@ cmake_dependent_option(PJSIP_WITH_TLS
   "PJLIB_WITH_SSL" OFF
 )
 
+if(PJLIB_WITH_SSL STREQUAL "openssl")
+  find_package(OpenSSL COMPONENTS Crypto CONFIG)
+  if(NOT OPENSSL_FOUND)
+    find_package(OpenSSL COMPONENTS Crypto)
+  endif()
+endif()
+
```
aymanalqadhi: Again, no need to find `OpenSSL` if already found in `pjlib`

third_party/srtp/CMakeLists.txt
```diff
@@ -3,7 +3,10 @@
 
 option(SRTP_WITH_OPENSSL "Build SRTP with OpenSSL support" ON)
 if(SRTP_WITH_OPENSSL)
-  find_package(OpenSSL COMPONENTS Crypto SSL)
+  find_package(OpenSSL COMPONENTS Crypto SSL CONFIG)
+  if(NOT OPENSSL_FOUND)
+    find_package(OpenSSL COMPONENTS Crypto SSL)
+  endif()
```
aymanalqadhi, line 9: Relying on CMake to do this would be preferable

pjlib/CMakeLists.txt
a-lerion: It was used for linking choice which is earlier than the previous place. Do you mean move it just before the first use?

pjlib/CMakeLists.txt
a-lerion: Do you mean generator expressions with `$<IF`?

pjlib/CMakeLists.txt
a-lerion: Windows x64 usually has AMD64 for a processor name. Haven't tried it on MinGW.

pjmedia/CMakeLists.txt
a-lerion: It wasn't in my case and generated errors.

pjlib/CMakeLists.txt
a-lerion: Searching a package with OpenSSL_DIR set fails on my machine with the module only call with CMake 4.0. Does not look like a fallback.

pjmedia/CMakeLists.txt
a-lerion: > I tried to keep the libraries as loosely-coupled to each other as possible. Each library "chooses" what it links to. Some libraries require linking to m, but others don't.

That is why another search is here.

a-lerion: And this is primarily for includes for other components.

pjmedia/CMakeLists.txt
a-lerion: OK.

third_party/webrtc_aec3/CMakeLists.txt
a-lerion: OK.

pjmedia/CMakeLists.txt
a-lerion: Probalby before `CMAKE_SYSTEM_NAME MATCHES 'WindowsStore' OR CMAKE_SYSTEM_NAME MATCHES 'WindowsPhone'` condition was added.

pjmedia/CMakeLists.txt
aymanalqadhi: Since you are explicitly checking `PJLIB_WITH_SSL`, the build configuration of `pjmedia` (and `pjsip` for that matter) is already tightly-coupled to the one of `pjlib`.

One possible solution, although not ideal, is to make the `OpenSSL` dependency of `pjlib` public. This, while exposing implementation details to consumers, will remove the need to manually import `OpenSSL` in dependent libraries.

pjlib/CMakeLists.txt
aymanalqadhi: Try with `CMAKE_FIND_PACKAGE_PREFER_CONFIG` set to `ON`. This should make CMake start searching with config mode.

pjmedia/CMakeLists.txt
aymanalqadhi: So, it is not generating an error now that you have added this condition?

Also, the condition:
`PJMEDIA_WITH_AUDIODEV;WIN32 OR CYGWIN OR MINGW;CMAKE_SYSTEM_NAME MATCHES 'WindowsStore' OR CMAKE_SYSTEM_NAME MATCHES 'WindowsPhone'`
Ignores the `CYGWIN` and `MINGW` cases, since they are neither `WindowsStore` nor `WindowsPhone`.

It can be simplified to just:
`PJMEDIA_WITH_AUDIODEV;WIN32 AND CMAKE_SYSTEM_NAME MATCHES 'Windows(Store|Phone)'`

pjlib/CMakeLists.txt
aymanalqadhi: No, just instead of:
```cmake
target_link_libraries(pjlib PRIVATE
    ole32
    winmm
)
```

This is fine:
```cmake
target_link_libraries(pjlib PRIVATE ole32 winmm)
```

pjlib/CMakeLists.txt
aymanalqadhi: In the same "block" where it is used. Including it near its usage makes refactoring easier, because it can be moved with the consuming code at once.

pjmedia/CMakeLists.txt
a-lerion: It is coupled in code. Which is why this condition looks that way.

a-lerion: Yes, not ideal. As PUBLIC usually means that other library headers are exposed in public headers of the current library, not that it should be used to include the same library in other components that directly depend on it.

a-lerion: What about searching at the top file, probably via include, without specifying modules and REQUIRED?
And every component that uses it, taking into account its current options, then links to it.

pjlib/CMakeLists.txt
a-lerion: In my opinion the search should succeed by default, without special CMake options, if given enough data. And the default FindOpenSSL does not do it with OpenSSL_DIR variable set, but can succeed with paths to include directory and both built binaries.

pjmedia/CMakeLists.txt
aymanalqadhi: A more elegant solution, in my opinion, would be to add a non-exported `INTERFACE` target when configuring `pjlib` that exposes its dependencies. Any target (e.g., `pjmedia` and `pjsip`) that needs to link with `pjlib`'s dependencies, can link to it.

pjmedia/CMakeLists.txt
```diff
@@ -235,7 +235,7 @@ cmake_dependent_option(PJMEDIA_WITH_AUDIODEV_WMME
 # Windows Audio Session API (WASAPI) backend
 cmake_dependent_option(PJMEDIA_WITH_AUDIODEV_WASAPI
   "Enable WASAPI audio device backend" ON
-  "PJMEDIA_WITH_AUDIODEV;WIN32 OR CYGWIN OR MINGW;CMAKE_SYSTEM_NAME MATCHES 'WindowsStore' OR CMAKE_SYSTEM_NAME MATCHES 'WindowsPhone'" OFF
+  [[PJMEDIA_WITH_AUDIODEV AND (WIN32 OR CYGWIN OR MINGW) AND CMAKE_SYSTEM_NAME MATCHES "Windows(Store|Phone)"]] OFF
```
aymanalqadhi: Since you are requiring WindowsStore/WindowsPhone exclusively, I don't think checking for `CYGWIN` or `MINGW` is needed, is it?

pjlib/CMakeLists.txt
aymanalqadhi: It succeeds in most cases when using the default behavior. Leaving it to to CMake seems to be the more future-proof (and cleaner) approach.

sauwming:
If no objections from others, I think we can merge this, but since this is not complete yet, I propose the following:
- clearly prints a message that the build system is still experimental
- and lists the platforms it support

aymanalqadhi:
What should the message say?

Is this format appropriate?

```cmake
[=[
****************************************************
* Building `pjproject` with CMake is experimental  *
* -------------------------------------------------*
* Tested platforms:                                *
* - Linux x86_64                                   *
****************************************************
]=]
```

sauwming:
Looks good to me.

sauwming:
I can make it build on Mac with some modifications:
- add the necessary frameworks (CoreFoundation, Foundation, AudioToolbox, CoreAudio, AppKit)
- disable webrtc and aec3 (it fails to build for me)

Here's a rough patch (not sure if it's the proper way):
[mac_build.txt](https://github.com/user-attachments/files/22163022/mac_build.txt)

sauwming (approved):

aymanalqadhi:
@sauwming Check if you can build with the latest commits

sauwming:
```
CMake Error at pjlib/CMakeLists.txt:247 (target_link_libraries):
  The plain signature for target_link_libraries has already been used with
  the target "pjlib".  All uses of target_link_libraries with a target must
  be either all-keyword or all-plain.

  The uses of the plain signature are here:
   * pjlib/CMakeLists.txt:240 (target_link_libraries)

-- Configuring incomplete, errors occurred!
```

pjlib/CMakeLists.txt
```diff
@@ -0,0 +1,779 @@
...
+  elseif(PJ_HAS_WINSOCK_H)
+    target_link_libraries(pjlib PRIVATE wsock32)
+  endif()
+endif()
+
+# apple
+if(APPLE)
+  find_library(CoreFoundation_LIBRARY CoreFoundation)
+  if (CoreFoundation_LIBRARY)
+    target_link_libraries(pjlib PRIVAE "${CoreFoundation_LIBRARY}")
```
sauwming: it works on Mac.
only small typo here: PRIVATE

aymanalqadhi: It should be `PRIVATE`. This ensures that the dependency does not get leaked to any target that uses `pjlib`.

aymanalqadhi: Which target fails to build?

sauwming: Yes, there's a typo there: PRIVAE, which should be: PRIVATE.
After fixing this typo, the build is successful.

aymanalqadhi: Oh sorry 😅, I thought you meant it should've been something other than PRIVATE.

aymanalqadhi: @sauwming Check if it builds now

sauwming: Yes, it builds successfully now.

sauwming (approved):

nanangizz:
Not familiar with CMake, so I guess I did some noob mistake. A quick try on Windows for Visual Studio 17 2022 (msvc 19) using cmake-gui, configure failed:
```
CMake Error at third_party/g7221/CMakeLists.txt:34 (file):
  file failed to create symbolic link
  'D:/work/tool/cmake-4.1.1-windows-x86_64/bin/third_party/g7221/include/g7221/common':
  A required privilege is not held by the client.
```

a-lerion:
@aymanalqadhi 
Commit https://github.com/pjsip/pjproject/pull/4494/commits/87cf352d6eb53a8b5f12d661b5c1a34b4a7d0e2e removes an option to use a static C++ runtime.

By default CMake generates MSVC files with a dynamic C++ runtime usage, same as specified by the default option value.
Why remove ability to set a static one via an option?

aymanalqadhi:
> @aymanalqadhi Commit [87cf352](https://github.com/pjsip/pjproject/commit/87cf352d6eb53a8b5f12d661b5c1a34b4a7d0e2e) removes an option to use a static C++ runtime.
> 
> By default CMake generates MSVC files with a dynamic C++ runtime usage, same as specified by the default option value. Why remove ability to set a static one via an option?

Because it is not up to the library to decide which runtime to use. The code that uses the library, on the other hand, is responsible for setting such values consistently.

aymanalqadhi:
> Not familiar with CMake, so I guess I did some noob mistake. A quick try on Windows for Visual Studio 17 2022 (msvc 19) using cmake-gui, configure failed:
> 
> ```
> CMake Error at third_party/g7221/CMakeLists.txt:34 (file):
>   file failed to create symbolic link
>   'D:/work/tool/cmake-4.1.1-windows-x86_64/bin/third_party/g7221/include/g7221/common':
>   A required privilege is not held by the client.
> ```

I think this happens due to the fact that Windows symbolic links require certain privileges to be created, unlike in Unix-based operating systems.

I used symbolic links to allow including headers from `g7221` third-party library with `g7221` prefix (the library expects to find its headers under `g7221`, which would require either adding `third_party` to the include path, or using this trick).

Try compiling with the latest commit, which copies files when creating a symbolic link fails.

pjlib/CMakeLists.txt
aymanalqadhi: Should I add your OS/Architecture to the tested platforms list? If so, please reply with what should be added.

a-lerion:
> Because it is not up to the library to decide which runtime to use. The code that uses the library, on the other hand, is responsible for setting such values consistently.

It's not the library itself that decides it. The option needs to be explicitly set during project generation to have a different behaviour from the default one.

aymanalqadhi:
> > Because it is not up to the library to decide which runtime to use. The code that uses the library, on the other hand, is responsible for setting such values consistently.
> 
> It's not the library itself that decides it. The option needs to be explicitly set during project generation to have a different behaviour from the default one.

Setting `CMAKE_MSVC_RUNTIME_LIBRARY` inconsistently might have unexpected side effects. For example, if the library was used by an application that uses the dynamic runtime, but the library was linked with the static one, wouldn't there be two copies of the standard library functions that have a global state (i.e., `malloc`, `strtok`, etc.)? If so, that global state is no longer shared between the application and the library, which might cause serious bugs.

Leaving `CMAKE_MSVC_RUNTIME_LIBRARY` to the user to set is, in my opinion, the safest option.

nanangizz:
> Try compiling with the latest commit, which copies files when creating a symbolic link fails.

Thanks, the _configure_ step completed successfully now. Just FYI, there was a warning regarding the SDL2 probe (due to missing explicit settings and the absence of a CMake detection config?):
```
CMake Warning at pjmedia/CMakeLists.txt:9 (find_package):
  By not providing "FindSDL2.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "SDL2", but
  CMake did not find one.

  Could not find a package configuration file provided by "SDL2" with any of
  the following names:

    SDL2Config.cmake
    sdl2-config.cmake

  Add the installation prefix of "SDL2" to CMAKE_PREFIX_PATH or set
  "SDL2_DIR" to a directory containing one of the above files.  If "SDL2"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  pjmedia/CMakeLists.txt:269 (import_if_or_set)

[!] SDL2 was not found. setting PJMEDIA_WITH_VIDEODEV_SDL to OFF
```

Then the _generate_ step failed, looks like the `COPY_ON_ERROR` only copy the folder but not the content (tried to search around and found [this](https://gitlab.kitware.com/cmake/cmake/-/issues/25278)), error message:
```
CMake Error at third_party/g7221/CMakeLists.txt:43 (target_sources):
  Cannot find source file:

    D:/work/tool/cmake-4.1.1-windows-x86_64/bin/third_party/g7221/include/g7221/common/basic_op.h
```
Perhaps it could check whether one or more header files exist. If not, it could copy all the G7221 header files. However, I'm unsure whether this approach (especially the copying part) would behave correctly if the symbolic link is created successfully.

a-lerion:
> Leaving `CMAKE_MSVC_RUNTIME_LIBRARY` to the user to set is, in my opinion, the safest option.

That option is also for the user to set. It is simpler to set an ON/OFF option than to pass a generator expression.
Project libraries may be compiled for future use, without building an app that uses them.

> Wouldn't there be two copies of the standard library functions?

This gives errors during linking.

a-lerion:
> I used symbolic links to allow including headers from `g7221` third-party library with `g7221` prefix (the library expects to find its headers under `g7221`, which would require either adding `third_party` to the include path, or using this trick).

MSVC project files and Makefile that exist in the repo add third_party folder to include folders.

aymanalqadhi:
> > Leaving `CMAKE_MSVC_RUNTIME_LIBRARY` to the user to set is, in my opinion, the safest option.
> 
> That option is also for the user to set. It is simpler to set an ON/OFF option than to pass a generator expression. Project libraries may be compiled for future use, without building an app that uses them.

This is equivalent to adding a switch to select which compiler to use. Sure, it may work sometimes (since C has a stable ABI), but not every time. Giving users this option is like intentionally giving them a gun to shoot their feet.

> 
> > Wouldn't there be two copies of the standard library functions?
> 
> This gives errors during linking.

No, it wont. Using shared libraries does not involve compile-time linkage, but rather the libraries are dynamically-linked (hence the name) at load-time by the loader.

Feel free to correct me if I am wrong.

aymanalqadhi:
> > I used symbolic links to allow including headers from `g7221` third-party library with `g7221` prefix (the library expects to find its headers under `g7221`, which would require either adding `third_party` to the include path, or using this trick).
> 
> MSVC project files and Makefile that exist in the repo add third_party folder to include folders.

The existing build system also (mostly) includes everything to everything. Dependency isolation is one of the biggest selling points of CMake, why give it up?

a-lerion:
> No, it wont. Using shared libraries does not involve compile-time linkage, but rather the libraries are dynamically-linked (hence the name) at load-time by the loader.
> 
> Feel free to correct me if I am wrong.

E.g., linking a log4cplus library built with static runtime in a project built with dynamic runtime gives errors like the following in Visual Studio:

```
error LNK2005: "public: unsigned short const * __cdecl std::_Locinfo::_W_Getmonths(void)const " (?_W_Getmonths@_Locinfo@std@@QEBAPEBGXZ) already defined in msvcprtd.lib(MSVCP140D.dll)
error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MTd_StaticDebug' doesn't match value 'MDd_DynamicDebug' in WindowProject.obj
```

a-lerion:
> The existing build system also (mostly) includes everything to everything. Dependency isolation is one of the biggest selling points of CMake, why give it up?

Current solution looks a bit too much for this. I think changing #include directives would be better. It looks like only one file uses such includes, that start with g7221, third_party/g7221/common/defs.h, and it includes files in the same folder.

pjmedia/CMakeLists.txt
```diff
@@ -0,0 +1,903 @@
...
+    # TODO: add remaining APIs' dependencies
+
+    # transitive deps
+    $<$<BOOL:${PJMEDIA_WITH_LIBYUV}>:Pj::Dep::YUV>
+)
+
+# ##############################################################################
+#  Test application
+
+if(BIULD_TESTING)
```
nanangizz: typo? should be `BUILD_TESTING`?

pjsip/CMakeLists.txt
```diff
@@ -0,0 +1,341 @@
...
+    pjsip
+    pjsip-simple
+    pjsip-ua
+    pjsua-lib
+)
+
+# ##############################################################################
+# Test application
+
+if(BIULD_TESTING)
```
nanangizz: Similar to above, typo?

nanangizz:
Update on MSVC, it is now built successfully for the libs & and the pjsua app, however it is still failed on test apps (`pjlib-test, pjlib-util-test, pjnath-test`) with this error message:
```
1>MSVCRTD.lib(exe_main.obj) : error LNK2019: unresolved external symbol main referenced in function "int __cdecl invoke_main(void)" (?invoke_main@@YAHXZ)
```
The `CMakeFile` for test apps seem to use `main_win32.c` for Windows platforms, I think this is the cause. Replacing it with `main.c` should fix this. In fact the existing MSVC projects use `main.c`. IIRC the `main_win32.c` is for the old & deprecated Windows Mobile that does not have console UI.

Btw, `pjmedia-test` & `pjsip-test` projects are missing, I believe because of the typos (see inline code comments).

a-lerion:
> The `CMakeFile` for test apps seem to use `main_win32.c` for Windows platforms, I think this is the cause.

Yes, its use requires WIN32 option for add_executable as it's not a console app but rather a window app.

> Replacing it with main.c should fix this. In fact the existing MSVC projects use main.c.

Agree.

aymanalqadhi:
> > No, it wont. Using shared libraries does not involve compile-time linkage, but rather the libraries are dynamically-linked (hence the name) at load-time by the loader.
> > Feel free to correct me if I am wrong.
> 
> E.g., linking a log4cplus library built with static runtime in a project built with dynamic runtime gives errors like the following in Visual Studio:
> 
> ```
> error LNK2005: "public: unsigned short const * __cdecl std::_Locinfo::_W_Getmonths(void)const " (?_W_Getmonths@_Locinfo@std@@QEBAPEBGXZ) already defined in msvcprtd.lib(MSVCP140D.dll)
> error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MTd_StaticDebug' doesn't match value 'MDd_DynamicDebug' in WindowProject.obj
> ```

It seems that MSVC's linker does check for conflicting symbols (in this case, `RuntimeLibrary`, which was put specifically to prevent what I was talking about). Other compilers might handle this differently.

aymanalqadhi:
> > The existing build system also (mostly) includes everything to everything. Dependency isolation is one of the biggest selling points of CMake, why give it up?
> 
> Current solution looks a bit too much for this. I think changing #include directives would be better. It looks like only one file uses such includes, that start with g7221, third_party/g7221/common/defs.h, and it includes files in the same folder.

I had to use that method to make sure this PR is add-only, without changing the current codebase. The code inside `third_party/g7221` should not be modified if that module is to be updated sometime in the future.

aymanalqadhi:
> The `CMakeFile` for test apps seem to use `main_win32.c` for Windows platforms, I think this is the cause. Replacing it with `main.c` should fix this. In fact the existing MSVC projects use `main.c`. IIRC the `main_win32.c` is for the old & deprecated Windows Mobile that does not have console UI.

Windows builds now should use `main.c`. Should Windows Phone builds still use `main_win32.c`?

> Btw, `pjmedia-test` & `pjsip-test` projects are missing, I believe because of the typos (see inline code comments).

You're right.

aymanalqadhi:
@nanangizz While you're at it, can you try building with other toolchains like LLVM/Clang, MinGW, and/or CYGWIN?

a-lerion:
> While you're at it, can you try building with other toolchains like LLVM/Clang, MinGW, and/or CYGWIN?

Tried Visual Studio with clangcl toolset for x64. webrtc_aec3 project succeeds after AVX2 architecture is specified, `target_compile_options` with `/arch:AVX2`. Otherwise fails with errors like the following:

```
third_party\webrtc_aec3\src\common_audio\resampler\sinc_resampler_avx2.cc(25,20): error : always_inline function '_mm256_setzero_ps' requires target feature 'avx', but would be inlined into function 'Convolve_AVX2' that is compiled without support for 'avx'
third_party\webrtc_aec3\src\common_audio\resampler\sinc_resampler_avx2.cc(34,17): error : always_inline function '_mm256_fmadd_ps' requires target feature 'fma', but would be inlined into function 'Convolve_AVX2' that is compiled without support for 'fma'
third_party\webrtc_aec3\src\common_audio\resampler\sinc_resampler_avx2.cc(46,34): error : '__builtin_ia32_vextractf128_ps256' needs target feature avx
```

Other binaries are built without problems.

nanangizz:
Yes, Windows Phone may also use `main_win32.c`, but I guess we can skip that for now (the platform itself has been deprecated btw).

FYI, I experienced issues configuring for MinGW (I am a newbie, remember? :), when I use _cmake-gui_ (the same app I used for configuring for MSVC), with generator set to "MinGW Makefiles" & `CMAKE_MAKE_PROGRAM=C:/msys64/usr/bin/make.exe`, it failed:
```
The C compiler identification is GNU 14.2.0
The CXX compiler identification is GNU 14.2.0
Detecting C compiler ABI info
Detecting C compiler ABI info - failed
Check for working C compiler: C:/msys64/mingw64/bin/cc.exe
Check for working C compiler: C:/msys64/mingw64/bin/cc.exe - broken
CMake Error at D:/work/tool/cmake-4.1.1-windows-x86_64/share/cmake-4.1/Modules/CMakeTestCCompiler.cmake:67 (message):
  The C compiler

    "C:/msys64/mingw64/bin/cc.exe"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: 'D:/work/project/pjproject-tmp/cmake-mingw/CMakeFiles/CMakeScratch/TryCompile-qqqngy'
    
    Run Build Command(s): D:/work/tool/cmake-4.1.1-windows-x86_64/bin/cmake.exe -E env VERBOSE=1 C:/msys64/usr/bin/make.exe -f Makefile cmTC_76308/fast
    /usr/bin/make  -f CMakeFiles\cmTC_76308.dir\build.make CMakeFiles/cmTC_76308.dir/build
    ' m T C _ 7 6 3 0 8 . d i r '   i s   n o t   r e c o g n i z e d   a s   a n   i n t e r n a l   o r   e x t e r n a l   c o m m a n d , 
     o p e r a b l e   p r o g r a m   o r   b a t c h   f i l e . 

     make: *** [Makefile:133: cmTC_76308/fast] Error 1
    
  CMake will not be able to correctly generate this project.
Call Stack (most recent call first):
  CMakeLists.txt:3 (project)

Configuring incomplete, errors occurred!
```

I also tried to use cmake-3.30.5, installed using `pacman -S cmake` from MinGW64/Msys2 console, failed:
```
$ cmake -S D:/work/project/pjproject-tmp/ -B D:/work/project/pjproject-tmp/cmake-mingw/
-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
CMake Error at /usr/share/cmake/Modules/CMakeDetermineCompilerABI.cmake:139 (file):
  file STRINGS file
  "D:/work/project/pjproject-tmp/D:/work/project/pjproject-tmp/cmake-mingw/CMakeFiles/3.30.5/CMakeDetermineCompilerABI_C.bin"
  cannot be read.
Call Stack (most recent call first):
  /usr/share/cmake/Modules/CMakeTestCCompiler.cmake:26 (CMAKE_DETERMINE_COMPILER_ABI)
  CMakeLists.txt:3 (project)


-- Check for working C compiler: /mingw64/bin/cc.exe - skipped
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
CMake Error at /usr/share/cmake/Modules/CMakeDetermineCompilerABI.cmake:139 (file):
  file STRINGS file
  "D:/work/project/pjproject-tmp/D:/work/project/pjproject-tmp/cmake-mingw/CMakeFiles/3.30.5/CMakeDetermineCompilerABI_CXX.bin"
  cannot be read.
Call Stack (most recent call first):
  /usr/share/cmake/Modules/CMakeTestCXXCompiler.cmake:26 (CMAKE_DETERMINE_COMPILER_ABI)
  CMakeLists.txt:3 (project)


-- Check for working CXX compiler: /mingw64/bin/c++.exe - skipped
CMake Warning at CMakeLists.txt:16 (message):
   ****************************************************
   * Building `pjproject` with CMake is experimental  *
   *--------------------------------------------------*
   * Tested platforms:                                *
   * - Linux x86_64                                   *
   ****************************************************

  Set `PJ_SKIP_EXPERIMENTAL_NOTICE=ON` to disable this warning


CMake Warning (dev) at /usr/share/cmake/Modules/GNUInstallDirs.cmake:253 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  CMakeLists.txt:45 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Looking for winsock.h
-- Looking for winsock.h - found
-- Looking for winsock2.h
-- Looking for winsock2.h - found
-- Looking for mswsock.h
-- Looking for mswsock.h - found
-- Looking for ws2tcpip.h
-- Looking for ws2tcpip.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of unsigned short
CMake Error at /usr/share/cmake/Modules/CheckTypeSize.cmake:159 (file):
  file STRINGS file
  "D:/work/project/pjproject-tmp/pjlib/D:/work/project/pjproject-tmp/cmake-mingw/CMakeFiles/CheckTypeSize/CMAKE_SIZEOF_UNSIGNED_SHORT.bin"
  cannot be read.
Call Stack (most recent call first):
  /usr/share/cmake/Modules/CheckTypeSize.cmake:272 (__check_type_size_impl)
  /usr/share/cmake/Modules/TestBigEndian.cmake:55 (CHECK_TYPE_SIZE)
  /usr/share/cmake/Modules/TestBigEndian.cmake:37 (__TEST_BIG_ENDIAN_LEGACY_IMPL)
  pjlib/CMakeLists.txt:341 (test_big_endian)

...

-- Configuring incomplete, errors occurred!
```

Any idea?

a-lerion:
@nanangizz

> FYI, I experienced issues configuring for MinGW (I am a newbie, remember? :), when I use _cmake-gui_ (the same app I used for configuring for MSVC), with generator set to "MinGW Makefiles" & `CMAKE_MAKE_PROGRAM=C:/msys64/usr/bin/make.exe`, it failed

What installation package have you used? Does it correspond to your platform?
Have you installed a compiler package or was it available?
Have you used cmake from a package installer or from a host OS?

I have installed msys2-x86_64-20250830.exe, installed packages mingw-w64-x86_64-gcc git cmake ninja openssl-devel.
After cloning a repository generating project files for ninja succeeded.
Compilation failed for some files in the webrtc_aec3 target.

```
third_party/webrtc_aec3/src/rtc_base/platform_thread_types.h:47:1: error: 'PlatformThreadId' does not name a type
   47 | PlatformThreadId CurrentThreadId();
third_party/webrtc_aec3/src/rtc_base/platform_thread_types.h:52:1: error: 'PlatformThreadRef' does not name a type
   52 | PlatformThreadRef CurrentThreadRef();
```

After adding WEBRTC_WIN compiler definition for MINGW compilation failed with
```
third_party/webrtc_aec3/src/rtc_base/platform_thread_types.cc:102:3: error: '__try' was not declared in this scope
third_party/webrtc_aec3/src/rtc_base/platform_thread_types.cc:105:5: error: '__except' was not declared in this scope
```
__try/__except are Microsoft-specific.

nanangizz:
> What installation package have you used? Does it correspond to your platform? Have you installed a compiler package or was it available? Have you used cmake from a package installer or from a host OS?

I already have MinGW/msys2 installed for development, with compiler package.
Then, I install CMake, tried both without luck:
- CMake for Windows, the host OS, (version 4.1.1), there are generator options "MinGW Makefile" & "msys2 Makefile", tried both options.
- CMake for MinGW (version 3.30.5, installed from MinGW console using packet manager `pacman`).

Haven't tried to generate project files for ninja though.

a-lerion:
Tried with host OS cmake and got similar errors.
MinGW cmake worked fine. Tried relative paths with a copy of repository inside MinGW user home folder. And tried absolute paths outside of MinGW using `/` rather than `:`, e.g.
`/D/work/project/pjproject-tmp/cmake-mingw/ ` instead of `D:/work/project/pjproject-tmp/cmake-mingw/`

nanangizz:
I can generate the project files on MinGW successfully now. However, in `make`, I got the same error as you've described above (`'PlatformThreadId' does not name a type`). Also if I tried to run `make` again, it always failed:
```
pjlib/CMakeFiles/pjlib.dir/compiler_depend.make:4: *** multiple target patterns.  Stop.
make[1]: *** [CMakeFiles/Makefile2:1133: pjlib/CMakeFiles/pjlib.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

nanangizz:
Just realized that using the existing `Makefile`, WebRTC AEC3 on MinGW (e.g: `./configure --enable-libwebrtc-aec3`) fails too:
```
../../webrtc_aec3/src//rtc_base/platform_thread_types.cc:46:10: error: invalid cast from type 'pthread_t' {aka 'long long unsigned int'} to type 'rtc::PlatformThreadId' {aka 'long long int'}
   46 |   return reinterpret_cast<PlatformThreadId>(pthread_self());
```
So I guess we can just disabled the WebRTC AEC3 for MinGW for now?

PR #4497, opened 2025-07-09
Fixed conference destruction and failure handling

sauwming:


pjmedia/src/pjmedia/conference.c
```diff
@@ -403,14 +403,12 @@ static pj_status_t create_conf_port( pj_pool_t *parent_pool,
...
-        status = PJ_ENOMEM;
-        goto on_return;
-    }
+    if (!pool)
+        return PJ_ENOMEM;
 
     /* Create port. */
     conf_port = PJ_POOL_ZALLOC_T(pool, struct conf_port);
-    PJ_ASSERT_ON_FAIL(conf_port, {status = PJ_ENOMEM; goto on_return;});
+    //PJ_ASSERT_ON_FAIL(conf_port, {status = PJ_ENOMEM; goto on_return;});
```
nanangizz: Why is this commented, first allocation from the pool should always be successful?

nanangizz (approved):

pjmedia/src/pjmedia/conference.c
sauwming: Yes, if the pool initial size is greater than the first allocation size, which is the case here (`struct conf_port` is only 248 bytes I check here).

nanangizz: Okay, but note that there may still be a corner case: what if the type sizes get doubled, e.g: `unsigned/int` is 8 bytes instead of 4 bytes (e.g: in the future or [here](https://www.intel.com/content/www/us/en/docs/mpi-library/developer-guide-linux/2021-6/ilp64-support.html)), also similar for pointer, and the system almost run out of memory.

LeonidGoltsblat:
1. Now you assign conf_port->port twice in conf_thread.c (when creating grp_lock and later in the /* Save some port's infos, for convenient. */ block) and only once in conference.c. Is that intended?
2. What if media_port::grp_lock was created before create_conf_port() was called? Shouldn't we detach the new handlers and decrement the counter (rewind everything created in create_conf_port())? That's why I didn't make those changes when I first saw the problem... "with all possible caution" was about it. 
3. we get status = pj_grp_lock_add_handler() however the status value is not checked here

sauwming:
1. Yes, that's intentional.
Initially I want to remove the latter assignment `conf_port->port = port` but then, the `if else` block will be unbalanced. I also consider removing `conf_port->port = NULL` in the `else` block since it's unnecessary because conf_port is ZALLOC-ed. In the end, I decided for the most minimal modification for clarity.

2. I did decrement the ref counter, so conf_port will be deleted when the counter reaches zero. The drawback of this approach is that the destruction may be delayed, but the advantage is again the clarity, because it's the same pattern used in the library as well as other parts of the conference upon failure (`if (status != PJ_SUCCESS) if (conf_port->port && conf_port->port->grp_lock) pj_grp_lock_dec_ref() ...`  has been used in three other sections).
  The alternative `del_handler() + destroy_conf_port()` can also work, it's just not as common. But if anyone prefers this, I have no objection since this is programmatically and behaviorally correct as well.
  Additional note: I have also made sure that the first two memory allocation PJ_ENOMEM failures do not go to `on_return` to simplify things and prevent unintentional external group lock decrement.

3. I noticed that too, and checked that pj_grp_lock_add_handler() always returns success.

LeonidGoltsblat:
2. Now I think this is a good approach only in the op_remove_port(), and the other two are my mistakes

@sauwming Oops! We have one more place where we need the same code but it is not there: pjmedia_conf_add_port() when get_free_op_entry() returns NULL. (It's funny I added handling for this situation to the deprecated pjmedia_conf_add_passive_port() but not to pjmedia_conf_add_port().)

sauwming:
Okay, feel free to create a PR to rectify this.

PR #4527, opened 2025-08-05
Media port changes for pjsip 2.13

PesalaDeSilva:
Moved the media port changes to the application layer as suggested
  (Media-Port-Changes-for-call-record-and-play-feature):
https://github.com/pjsip/pjproject/pull/4366.


nanangizz:
Before reviewing the code, I think we need to discuss the changes first.
Unfortunately the PR description does not seem to clearly describe the changes.
Anyway, just skimmed the code changes and based on copilot description:

> Key changes:
>
>     * Adds `PJMEDIA_PORT_ENABLE_ALWAYS` operation mode for forcing frame processing even without listeners

Could you specify what the purpose of this feature, and can the purpose simply be achieved in the app layer?

>     * Implements new conference bridge functions for port signature comparison and info retrieval

Port signatures in pjmedia is currently mainly for internal sanity checks, so it is rarely useful for app. So similar to above, what is this feature for and can your app work without it?

>     * Updates conference logic to handle the new always-enabled port mode

Looks like similar to the first.

PesalaDeSilva:
Related to the first comment you had 
**PJMEDIA_PORT_ENABLE_ALWAYS**
Enable both TX and RX for this port, and invoke get_frame() and
put_frame() on every conference bridge tick, regardless of whether:
- there are active connections/listeners, or
- the audio is silence.
This forces continuous frame processing even without listeners
or non-silence frames.
-need continuous audio processing regardless of network or listener state.
-want recordings or playback to be uninterrupted by connection/silence logic.

trengginas:
If you want continuous frame processing, one alternative is to create a custom media port (e.g.: mem player/capture), add it to the conference bridge and connect it to the other ports.

sauwming:
Due to compatibility issue with the latest master version, we can't immediately merge this PR as is. Let me provide a summary:
1. `PJMEDIA_PORT_ENABLE_ALWAYS`
In your application, you can create a dummy port or a null port (https://docs.pjsip.org/en/latest/api/generated/pjmedia/group/group__PJMEDIA__NULL__PORT.html) and connect it to your media port. This way, you will get the continuous frame processing that you want.
2. `pjmedia_conf_get_media_port_info()`
We will accept this and combine it to the current API `pjmedia_conf_get_port_info().`
3. `pjmedia_conf_compare_port_signature()`
Since the signature will now be available in pjmedia_conf_port_info, you can then compare the signature in your application.

To follow up, there's nothing to do on your part. We will create a new PR based on the latest master that incorporates point (2) above. Please let us know if you have any questions or feedback. Thanks again for the contribution.

sauwming:
To update you, we have created issue #4555 and PR #4556 to adapt your proposed patch for point (2) to be compatible with the latest PJSIP version.

sauwming:
Completed in #4556

PR #4545, opened 2025-08-16
prevent infinite loop on reading of invalide wav/avi file

LeonidGoltsblat:
pjmedia fix

The program goes into an infinite loop when iterating over a wav or avi file, if the length field of the readed chunck header is zero.
Now this situation is tracked in the log, the loop is interrupted, and the function returns with the status code PJMEDIA_ENOTVALIDWAVE.
A similar reaction has been added for the zero chunck.id field.

### DIFF
--- a/pjmedia/src/pjmedia/avi_player.c
+++ b/pjmedia/src/pjmedia/avi_player.c
@@ -830,6 +830,32 @@ static pj_status_t skip_forward(pjmedia_port *this_port, pj_size_t frames)
         if (status != PJ_SUCCESS)
             return status;
             
+        if (!ch.len ||    /* prevent infinite loop */
+            !ch.id)       /* highly likely that the file is invalid */
+        {
+            char fourcc_name[5];
+            pj_uint32_t sig = ch.id;
+            pj_off_t pos;
+            pj_file_getpos(fport->fd, &pos);
+            if ((sig & (0xFF << 24)) && (sig & (0xFF << 16)) && (sig & (0xFF << 8)) && (sig & (0xFF << 0)))
+                PJ_LOG(2, (THIS_FILE,
+                           "%.*s. Zero length chunk of type %s found at offset=%lu (skip_forward)",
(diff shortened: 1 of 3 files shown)


pjmedia/src/pjmedia/avi_player.c
```diff
@@ -830,6 +830,32 @@ static pj_status_t skip_forward(pjmedia_port *this_port, pj_size_t frames)
...
             return status;
             
+        if (!ch.len ||    /* prevent infinite loop */
+            !ch.id)       /* highly likely that the file is invalid */
+        {
+            char fourcc_name[5];
+            pj_uint32_t sig = ch.id;
+            pj_off_t pos;
+            pj_file_getpos(fport->fd, &pos);
+            if ((sig & (0xFF << 24)) && (sig & (0xFF << 16)) && (sig & (0xFF << 8)) && (sig & (0xFF << 0)))
```
sauwming, line 840: minor: put inside braces
if ()
{
} else {
}

also for the other files

LeonidGoltsblat:
minor: code style
major: goto on_error instead of return

and AL did not understand byte by byte check correctly: pjmedia_fourcc_name requires 4_CC but not less_than_4_CC

pjmedia/src/pjmedia/avi_player.c
nanangizz: Fourcc containing a null byte is not common but still a valid fourcc, isn't it? Some people may use it for custom/proprietary/testing purpose perhaps?

LeonidGoltsblat: Even if someone uses such fourcc, pjmedia_fourcc_name() requires 4_CC, not "less_than_4_CC". pjmedia_fourcc_name() can't properly format something like "A\0B\0" (the string will be truncated to the first \0 character), so I don't use pjmedia_fourcc_name() for tracing if there are null characters in fourcc.

nanangizz (approved):

PR #4546, opened 2025-08-19
Signed-off-by: Pesala Silva <pesala.silva@ifs.com> Log_Enhacements for pjproject 2.13

PesalaDeSilva:
This PR for the Log Enhancements and modifications done for logs for better information for PJSIP 2.13 version

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_core.c
+++ b/pjsip/src/pjsua-lib/pjsua_core.c
@@ -770,7 +770,7 @@ PJ_DEF(pj_status_t) pjsua_reconfigure_logging(const pjsua_logging_config *cfg)
     pjsua_logging_config_dup(pjsua_var.pool, &pjsua_var.log_cfg, cfg);
 
     /* Redirect log function to ours */
-    pj_log_set_log_func( &log_writer );
+    pj_log_set_log_func((cfg && cfg->cb && !cfg->log_filename.slen) ? cfg->cb : &log_writer);
 
     /* Set decor */
     pj_log_set_decor(pjsua_var.log_cfg.decor);
--- a/pjlib/include/pj/assert.h
+++ b/pjlib/include/pj/assert.h
@@ -45,26 +45,24 @@
  *
  * @param expr      The expression to be evaluated.
  */
-#if PJ_DEBUG==0
-
-#   ifndef pj_assert
-#       include "pj/os.h"
-#       include "pj/log.h"
-#       define pj_assert(expr) \
-            do { \
-                if (!(expr)) { \
-                    if (pj_thread_is_registered()) \
-                        PJ_LOG(1, (__FILE__, "Assert failed: %s", #expr)); \
-                } \
-            } while (0)
-#   endif
-
-#else /* PJ_DEBUG != 0 */
-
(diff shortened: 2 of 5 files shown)


trengginas:
I think the added information is a bit excessive. The detail message content can be printed by enabling `pjsua_logging_config::msg_logging`.

pjsip/src/pjsua-lib/pjsua_core.c
```diff
@@ -770,7 +770,7 @@ PJ_DEF(pj_status_t) pjsua_reconfigure_logging(const pjsua_logging_config *cfg)
     pjsua_logging_config_dup(pjsua_var.pool, &pjsua_var.log_cfg, cfg);
 
     /* Redirect log function to ours */
-    pj_log_set_log_func( &log_writer );
+    pj_log_set_log_func((cfg && cfg->cb && !cfg->log_filename.slen) ? cfg->cb : &log_writer);
```
sauwming: As per the discussion in https://github.com/pjsip/pjproject/pull/4086#discussion_r1772695174:
it's better not to change the current behavior.

sauwming:
Thank you for the PR submission.

We’ve accepted most of the log enhancement contributions in #4086. However, regarding the additional SIP transport log in this PR, we believe it’s best to keep the current behavior. While the change does provide more detailed information, it also introduces extra overhead in terms of printing and memory buffer required.

If your application requires more detailed logging, we recommend enabling `msg_logging` as @trengginas suggested above. For further customization, you can also install your own log writer, either through the logging callback or by using the PJSIP module (this way, you don't need to modify PJSIP source base). For example, in pjsua the message logging will look like this:
```
13:36:16.747           pjsua_core.c  ....TX 1002 bytes Request msg REGISTER/cseq=41884 (tdta0x621000044db0) to TLS 139.162.62.29:5061:
REGISTER sip:sip.pjsip.org;transport=tls SIP/2.0
Via: SIP/2.0/TLS 199.77.210.25:64167;rport;branch=z9hG4bKPjC5A1C056-1E55-4F7C-B8E1-66CA46A5562F;alias
Route: <sip:sip.pjsip.org;transport=tls;lr>
Max-Forwards: 70
From: "Ming/401" <sips:401@sip.pjsip.org>;tag=736D4225-AB48-442F-8522-16B4F2FBED73
To: "Ming/401" <sips:401@sip.pjsip.org>
Call-ID: FAFB9CDE-C64F-40D0-BB71-3ED4831FD255
CSeq: 41884 REGISTER
....
```

sauwming:
Another alternative is to introduce a compile-time setting such as `PJSIP_MSG_INFO_HAS_EXTRA_DETAILS` (default is off for efficiency purposes).
So let us know whether:
a. you prefer to move the logging into your application level and leave the PJSIP code unchanged
b. modify the PJSIP code as per your proposed patch but add the compile-time setting above.

If you choose option b, there's nothing to do on your part. We will create a new PR based on your PR. If you choose a, then we will close the PR.

sauwming:
For the patch with the additional compile-time option, please check #4559 and #4560.

sauwming:
Completed with additional compile-time option in #4560

PR #4566, opened 2025-08-25
Direct OpenSSL objects for TLS credentials (key, certificate)

nanangizz:
This allow app to use OpenSSL ENGINE for loading a private key and use the `EVP_PKEY` instance as TLS credentials.

### DIFF
--- a/pjlib/include/pj/ssl_sock.h
+++ b/pjlib/include/pj/ssl_sock.h
@@ -150,6 +150,7 @@ typedef enum pj_ssl_cert_name_type
     PJ_SSL_CERT_NAME_IP
 } pj_ssl_cert_name_type;
 
+
 /**
  * Field type for looking up SSL certificate in the certificate stores.
  */
@@ -185,6 +186,7 @@ typedef enum pj_ssl_cert_lookup_type
 
 } pj_ssl_cert_lookup_type;
 
+
 /**
  * Describe structure of certificate lookup criteria.
  */
@@ -201,6 +203,61 @@ typedef struct pj_ssl_cert_lookup_criteria
     pj_str_t                keyword;
 
 } pj_ssl_cert_lookup_criteria;
+
+
+/**
+ * The SSL certificate buffer.
+ */
+typedef pj_str_t pj_ssl_cert_buffer;
+
+
+/**
+ * Data type of the direct SSL certificate.
+ */
+typedef enum pj_ssl_cert_direct_type
+{
+    /**
+     * No data.
+     */
+    PJ_SSL_CERT_DIRECT_NONE = 0,
+
+    /**
+     * Data is OpenSSL EVP_PKEY.
+     */
+    PJ_SSL_CERT_DIRECT_OPENSSL_EVP_PKEY = 1,
+
+    /**
+     * Data is OpenSSL X509 certificate.
+     */
+    PJ_SSL_CERT_DIRECT_OPENSSL_X509_CERT = 2,
+
+} pj_ssl_cert_direct_type;
+
+
+/**
+ * The direct SSL certificate. Instead of loading certificate from file,
+ * buffer, or OS certificate store, application can provide direct access
+ * to the backend specific certificate data using this structure.
+ */
+typedef struct pj_ssl_cert_direct
+{
+    /**
+     * Bit flag of the type of direct certificate data, see
+     * \ref pj_ssl_cert_direct_type.
+     */
+    unsigned type;
+    
+    /**
+     * Pointer to backend specific private key object, e.g: OpenSSL EVP_PKEY.
+     */
+    void *privkey;
+
+    /**
+     * Pointer to backend specific certificate object, e.g: OpenSSL X509.
+     */
+    void *cert;
+
+} pj_ssl_cert_direct;
 
 
 /**
@@ -254,10 +311,6 @@ typedef struct pj_ssl_cert_info {
 
 } pj_ssl_cert_info;
 
-/**
- * The SSL certificate buffer.
- */
-typedef pj_str_t pj_ssl_cert_buffer;
 
 /**
  * Create credential from files. TLS server application can provide multiple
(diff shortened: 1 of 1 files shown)


amubiera:
I’m looking forward to this feature. Currently, I have local hacks in ssl_sock_imp_common to avoid strduping the certificate buffers.

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1435,8 +1445,45 @@ static pj_status_t init_ossl_ctx(pj_ssl_sock_t *ssock)
                 BIO_free(kbio);
             }       
         }
+
+        if ((cert->direct.type && PJ_SSL_CERT_DIRECT_OPENSSL_EVP_PKEY) &&
```
amubiera: `&&` should be `==` ?

pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1435,8 +1445,45 @@ static pj_status_t init_ossl_ctx(pj_ssl_sock_t *ssock)
...
+                SSL_CTX_free(ctx);
+                ossock->ossl_ctx = NULL;
+                return status;
+            } else {
+                privkey_loaded = PJ_TRUE;
+                PJ_LOG(4,(ssock->pool->obj_name, "Direct private key set"));
+            }
+        }
+
+        if ((cert->direct.type && PJ_SSL_CERT_DIRECT_OPENSSL_X509_CERT) &&
```
amubiera: `&&` should be `==` ?

amubiera:
Instead of adding a new type of 'direct' cert structure, would it be more clear if there was a single variant type that could specify the kind of cert?

I know this would be a breaking change for backwards compatibility. But it may simplify the usage and future proof for other backends. I'm working with mbedTLS and possibly wolfSSL in the future.

For example, in `ssl_sock.h`:
```C
typedef enum pj_ssl_cert_source_type
{
    PJ_SSL_CERT_SOURCE_NONE,
    PJ_SSL_CERT_SOURCE_FILE,
    PJ_SSL_CERT_SOURCE_BUFFER,
    PJ_SSL_CERT_SOURCE_BUFFER_NO_DUP, 
    PJ_SSL_CERT_SOURCE_SCHANNEL_LOOKUP_SUBJECT,
    PJ_SSL_CERT_SOURCE_SCHANNEL_LOOKUP_FINGERPRINT,
    PJ_SSL_CERT_SOURCE_SCHANNEL_LOOKUP_FRIENDLY_NAME,
    PJ_SSL_CERT_SOURCE_OPENSSL_EVP_PKEY,
    PJ_SSL_CERT_SOURCE_OPENSSL_X509_CERT,
} pj_ssl_cert_source_type;

typedef struct pj_ssl_cert_source
{
    pj_ssl_cert_source_type type;   // type of source
    pj_str_t                str;    // filename, buffer, keyword, etc.
    void*                   data;   // backend specific data
} pj_ssl_cert_source;
```

And in `ssl_transport_tls.h`:
```C
typedef struct pjsip_tls_setting
{
    pj_str_t              ca_list_path;
    pj_ssl_cert_source    ca_list;
    pj_ssl_cert_source    cert;
    pj_ssl_cert_source    privkey;
    pj_str_t              privkey_password;
    ...
} pjsip_tls_setting;
```

Notes:
- The goal would be to have a single source for each piece except for `ca_list_path` which allows additional lookup for OpenSSL.
- Also, in my embedded system, I've been needing a mechanism so that PJSIP doesn't `strdup` large certificates that can take up a lot of memory. i.e. `PJ_SSL_CERT_SOURCE_BUFFER_NO_DUP` would assume that the `pj_str_t` remains in memory for the lifetime of the transport.
- Can also remove `pj_ssl_cert_buffer`, `pj_ssl_cert_lookup_type`, and `pj_ssl_cert_lookup_criteria`.

Just some thoughts...

pjlib/src/pj/ssl_sock_ossl.c
nanangizz: it should be `&`

pjlib/src/pj/ssl_sock_ossl.c
nanangizz: it should be `&`

nanangizz:
@amubiera Thanks for the feedback.

Actually I had similar thought before adding certificate store lookup (`pj_ssl_cert_lookup*`), we need to tidy it up a bit somehow. In the PJLIB level, i.e: `ssl_sock`, the credential types are now getting messy, also in the higher level, each `ssl_sock` app defines its own type, e.g: PJSIP `pjsip_tls_setting`, PJNATH `pj_turn_sock_tls_cfg`.

But honestly, I'm still unsure whether the urgency justifies the potential backward incompatibility issues. Anyway, if it’s going to be done, I guess we’d better do it in a separate PR.

pjlib/include/pj/ssl_sock.h
```diff
@@ -270,7 +323,10 @@ typedef pj_str_t pj_ssl_cert_buffer;
  * @param cert_file     The file of certificate.
  * @param privkey_file  The file of private key.
  * @param privkey_pass  The password of private key, if any.
- * @param p_cert        Pointer to credential instance to be created.
+ * @param p_cert        Pointer to credential instance.
+ *                      If the credential instance is NULL, a new credential
+ *                      instances will be created, otherwise the specified
```
sauwming: s/instances/instance.
Also below.

pjlib/include/pj/ssl_sock.h
```diff
@@ -344,17 +407,45 @@ PJ_DECL(pj_status_t) pj_ssl_cert_load_from_buffer(pj_pool_t *pool,
...
                                 pj_ssl_cert_t **p_cert);
 
+
+/**
+ * Create credential from the backend specific objects.
+ *
+ * For example, application can use OpenSSL ENGINE to load a private key from
+ * a hardware device, and then provide the EVP_PKEY instance to be used by
+ * OpenSSL backend via this function. Application should maintain the
+ * objects lifetime until the credential instance is no longer used
```
sauwming: Need to clarify: Since OpenSSL 3 has ref counting, is this only needed for OpenSSL < 3?

nanangizz: Actually it is a bit complicated for the docs, the library only increments the object ref counter when the certificate is assigned to an SSL socket (i.e: in `pj_ssl_sock_set_certificate()`, not really in `pj_ssl_cert_load_direct()`). Please check the updated docs.

PR #4567, opened 2025-08-25
Improve performance of pj_get_timestamp on Darwin

saghul:
Using match functions is faster than the syscall alternative. This is the implementation used on libuv (the Nodejs platform layer) amongst others.

### DIFF
--- a/pjlib/src/pj/os_timestamp_posix.c
+++ b/pjlib/src/pj/os_timestamp_posix.c
@@ -121,75 +121,18 @@ PJ_DEF(pj_status_t) pj_get_timestamp_freq(pj_timestamp *freq)
 
 #elif defined(PJ_DARWINOS) && PJ_DARWINOS != 0
 
-/* SYSTEM_CLOCK will stop when the device is in deep sleep, so we use
- * KERN_BOOTTIME instead. 
- * See ticket #2140 for more details.
- */
-#define USE_KERN_BOOTTIME 1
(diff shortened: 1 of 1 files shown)


saghul:
Not sure what the failed CIs are about, some of them are Ubuntu, which I didn't touch...

sauwming:
I have rerun the failed Ubuntu and Windows tests and now they pass, the Mac though still fails and seems relevant.

saghul:
I'll take a look, thank you for rerunning the CI!

sauwming:
It looks like it has passed all the CI tests now.
But since the function is only available at Mac 10.12 (https://developer.apple.com/documentation/kernel/1646199-mach_continuous_time) (yes, I know it's quite old and I'm not sure about iOS, I can't find the info), perhaps better add compile-time check of Mac&iOS `VERSION_MIN_REQUIRED` instead? So we don't need to completely remove the current legacy code and use it as a fallback.

saghul:
Do you still support such old macOS versions as targets? We've had this for quite a while in Nodejs / libuv without issues. May I suggest the opposite approach? We land it as is, and if people run into problems because of it, react and add the fallback. This way we can try to establish a baseline for what versions of macOS are supported. 

For reference the last macOS 10.12 version release was in 2017: https://endoflife.date/macos

sauwming (approved):

nanangizz (approved):

PR #4573, opened 2025-08-29
Fix bugs in SRTP when app manually changes crypto key

nanangizz:
Application may modify the SRTP crypto key in the SDP via the PJSUA callback `on_call_sdp_created()`. This works correctly initially. However, in subsequent SDP offer/answer cycles, the crypto key in the local SDP may revert to the original value when acting as answerer, or be replaced with kind of bad value when acting as offerer.

After investigation, I found that certain parts of the SRTP-SDES processing logic assume the remote peer will always offer crypto attributes with tags starting from 1. Additionally, SRTP-SDES may store the crypto key using a temporary or flip-flop memory pool.

### DIFF
--- a/pjmedia/src/pjmedia/transport_srtp_sdes.c
+++ b/pjmedia/src/pjmedia/transport_srtp_sdes.c
@@ -585,8 +585,15 @@ static pj_status_t sdes_encode_sdp( pjmedia_transport *tp,
 }
 
 
+/* Fill local crypto array from local SDP.
+ * If as_offerer is true, the crypto will be put in the array based on
+ * their tag number. Note that as offerer, we generate the tag number
+ * from 0..N, so the tag number is always inline with array index.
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia/transport_srtp_sdes.c
```diff
@@ -621,11 +621,14 @@ static pj_status_t fill_local_crypto(pj_pool_t *pool,
             return PJMEDIA_SRTP_ESDPINCRYPTOTAG;
 
         if (as_offerer) {
-            loc_crypto[loc_tag-1] = tmp_crypto;
+            if (loc_tag <= *count)
```
sauwming: I think Coverity may mark this as unnecessary since it's already established above that `loc_tag` cannot be > than `*count*`.

sauwming (approved):

PR #4576, opened 2025-09-01
Avoid calling `pj_log_get_level` when `level` is greater than `PJ_LOG_MAX_LEVEL`

aymanalqadhi:
`PJ_LOG` macro is never used with a runtime value as the first argument, so adding a check against `PJ_LOG_MAX_LEVEL` will ensure that the compiler always optimizes away the code generated by it.

In addition to eliminating unneeded calls to `pj_log_get_level`, this will also reduce the size of generated binaries when `level` is greater than `PJ_LOG_MAX_LEVEL`.

### DIFF
--- a/pjlib/include/pj/log.h
+++ b/pjlib/include/pj/log.h
@@ -104,7 +104,7 @@ enum pj_log_decoration
  * @hideinitializer
  */
 #define PJ_LOG(level,arg)       do { \
-                                    if (level <= pj_log_get_level()) { \
+                                    if (level <= PJ_LOG_MAX_LEVEL && level <= pj_log_get_level()) { \
                                         pj_log_wrapper_##level(arg); \
                                     } \
                                 } while (0)


LeonidGoltsblat:
@aymanalqadhi 

The idea is good, but it doesn't work, at least in the debug build in MS Visual Studio:
PJ_LOG(106, ...)
produces:
 warning C4013: 'pj_log_wrapper_106' undefined; assuming extern returning int

and then 
error LNK2019: unresolved external symbol _pj_log_wrapper_106 referenced in function ...
fatal error LNK1120: 1 unresolved externals

The compiler does not "always optimizes away the code generated by it", but only does so if optimization is enabled, which is not the case in a debug build.

aymanalqadhi:
@LeonidGoltsblat Of course the compiler wont optimize if optimizations are disabled. How is this a problem?

LeonidGoltsblat:
I looked more closely, you are right, apparently there is no problem.

PR #4582, opened 2025-09-05
Conf Slot changes and Few Enhancements for PJSIP 2.13

PesalaDeSilva:
This PR contains a few enhancements and bug fixes.

- **Preserve audio connectivity between calls even if the media stream is re-established** - Extending PJSIP library to automatically remember the states of the audio connections between calls independently of the lifetime of the actual media streams. If a media stream dynamically reinitializes within the progress of a call; then after reconnection, it will automatically restore the connection without the application layer’s involvement. 
Likewise, this enables the application to manage the connectivity between calls even when there are no media streams active (like at the very start of the call when early media is disabled on one end). This is primarily implemented with the addition of the following five new methods and updates to the final three:

- [ ] **pjmedia_conf_replace_port**
- [ ] **replace_conference_slot_to_null_port** (internal)
- [ ] **remove_conf_port** (internal)
- [ ] **allocate_conference_slot** (internal)
- [ ] **deallocate_conference_slot** (internal)
- [ ] **pjsua_aud_stop_stream** – replace stream with a null port before terminating
- [ ] **pjmedia_stream_create** – replace existing (null) stream with the new one
- [ ] **pjsua_call_get_conf_port** - modified to always return a usable conf slot if possible

- **Misc minor enhancements**

- [ ] Added new **pjmedia_get_queued_dtmf_digits**   and **pjsua_call_get_queued_dtmf_digits** methods to query the count of pending DTMF digits waiting in the transmit queue of a media stream.
- [ ] Added a new **pjsua_create2** method to allow the start-up logging emitted by the library to be captured by the application if required.
- [ ] Added a new **pjsua_make_call2** method to allow the source/from uri and the contact uri to be explicitly supplied by the application.
- [ ] Allow the maximum number of calls to be controlled at runtime rather than build time and added the **pjsua_call_subsys_destroy** method to free the memory allocated for the calls on shutdown.
- [ ] Extended **pjsua_acc_set_registration** to handle PJSIP_EBUSY error conditions by creating a new client and retrying.

- **A few bug fixes and optimizations**

- [ ] Added minor optimization in **pjmedia_silence_det_detect** to exit early if silence detection is disabled.
- [ ] Bug fix in **null_get_frame** - to avoid overflowing during long calls or high sample rates on 64-bit systems.
- [ ] Avoid crash in **pjsip_tsx_retransmit_no_state** if _tsx_ is null (which happens occasionally especially under load).


sauwming:
This PR is still too large to review. We'd appreciate it if you can break down to the smallest possible (even one-liner fix is okay if it's granular). Thanks, and will try to review them next week.

PesalaDeSilva:
Hi, Thank You for the Response and created 3 separate PR for the above
https://github.com/pjsip/pjproject/pull/4583
https://github.com/pjsip/pjproject/pull/4585
https://github.com/pjsip/pjproject/pull/4584

sauwming:
Thanks for the PRs. As you can check from the number of comments per PR, it's actually still too large because a single PR still contains quite a number of unrelated patches/commits. Ideally, each PR should address only one item.

But at least it's more manageable after the split, so I will close this and do the review in the newly created PRs.

PR #4583, opened 2025-09-08
Signed-off-by: Pesala Silva <pesala.silva@ifs.com> Conference Slot Changes

PesalaDeSilva:
**Preserve audio connectivity between calls even if the media stream is re-established** - Extending PJSIP library to automatically remember the states of the audio connections between calls independently of the lifetime of the actual media streams. If a media stream dynamically reinitializes within the progress of a call; then after reconnection, it will automatically restore the connection without the application layer’s involvement.
Likewise, this enables the application to manage the connectivity between calls even when there are no media streams active (like at the very start of the call when early media is disabled on one end). This is primarily implemented with the addition of the following five new methods and updates to the final three:

 

- [ ] pjmedia_conf_replace_port
- [ ]  replace_conference_slot_to_null_port (internal)
- [ ] remove_conf_port (internal)
- [ ] allocate_conference_slot (internal)
- [ ] deallocate_conference_slot (internal)
- [ ] pjsua_aud_stop_stream – replace stream with a null port before terminating
- [ ] pjmedia_stream_create – replace existing (null) stream with the new one
- [ ]  pjsua_call_get_conf_port - modified to always return a usable conf slot if possible


sauwming:
The idea sounds quite good and it can indeed be useful for others.
Unfortunately, the conference has undergone major change and the conference operations have now become async. Also worth noting that the conference now has another backend, the parallel conference: `conf_thread`, which means that you will also need to do the implementation there.

So unfortunately, we can't accept this at the moment. Not because there's an inherent flaw in the PR itself, but because this will need a reimplementation in order to be compatible with the latest version.

nanangizz:
Few notes:
- Haven't check the code in detail, but there may some corner cases that automatic reconnection may cause issues or not preferred, for example: when application tries to insert a new audio stream before the old/existing audio stream (so we may end-up needing a better stream identification or at least a setting to disable this).
- For consistency, the video better adapts the similar feature.

PR #4584, opened 2025-09-08
Signed-off-by: Pesala Silva <pesala.silva@ifs.com> Misc Minor Enhancements

PesalaDeSilva:
**Misc minor enhancements**

 

- [ ] Added new pjmedia_get_queued_dtmf_digits and pjsua_call_get_queued_dtmf_digits methods to query the count of pending DTMF digits waiting in the transmit queue of a media stream.
- [ ] Added a new pjsua_create2 method to allow the start-up logging emitted by the library to be captured by the application if required.
- [ ]  Added a new pjsua_make_call2 method to allow the source/from uri and the contact uri to be explicitly supplied by the application.
- [ ] Allow the maximum number of calls to be controlled at runtime rather than build time and added the pjsua_call_subsys_destroy method to free the memory allocated for the calls on shutdown.
- [ ] Extended pjsua_acc_set_registration to handle PJSIP_EBUSY error conditions by creating a new client and retrying.


pjmedia/include/pjmedia/stream.h
```diff
@@ -354,6 +354,18 @@ PJ_DECL(pj_status_t) pjmedia_stream_resume(pjmedia_stream *stream,
...
+/**
+ * Get the number of DTMF digits currently in the RFC 2833 transmit
+ * queue for this stream (valid only for audio streams).
+ *
+ * @param stream	The media stream.
+ * @param digits	Receives the number of queued digits.
+ *
+ * @return		PJ_SUCCESS on success.
+ */
+PJ_DECL(pj_status_t) pjmedia_get_queued_dtmf_digits(pjmedia_stream* stream,
```
sauwming: It looks okay to me, but I propose to modify this to `PJ_DECL(unsigned) pjmedia_get_queued_dtmf_digits(pjmedia_stream  *stream)` instead (similar to `pjmedia_stream_get_last_jb_frame_type()` or `pjmedia_stream_get_transport()`, we don't seem to need the return status here).

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -2736,6 +2736,17 @@ PJ_DECL(pjsua_msg_data*) pjsua_msg_data_clone(pj_pool_t *pool,
...
 
+/**
+ * Instantiate pjsua application. Application must call this function before
+ * calling any other functions, to make sure that the underlying libraries
+ * are properly initialized. Once this function has returned success,
+ * application must call pjsua_destroy() before quitting.
+ *
+ * @return		PJ_SUCCESS on success, or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsua_create2(const pjsua_logging_config* log_cfg);
```
sauwming: Why not just use `pjsua_init(..., const pjsua_logging_config *)` or `pjsua_reconfigure_logging()`? To make PJSIP set of APIs as clean as possible, we don't want to introduce new API unless really necessary.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -5955,6 +5966,33 @@ PJ_DECL(pj_status_t) pjsua_call_make_call(pjsua_acc_id acc_id,
...
+ * @return		PJ_SUCCESS on success, or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsua_call_make_call2(pjsua_acc_id acc_id,
+    const pj_str_t* dst_uri,
+    const pjsua_call_setting* opt,
+    void* user_data,
+    const pjsua_msg_data* msg_data,
+    pjsua_call_id* p_call_id,
+    const pj_str_t* src_uri,
+    const pj_str_t* contact_uri);
```
sauwming: We also don't seem to need this new API.

`src_uri` is already available in `msg_data->local_uri`.
As for the contact, you can modify the account's contact. But if you really need to use different custom contacts for many calls, then I suppose you can add a new field in `pjsua_msg_data`.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -3127,7 +3130,25 @@ PJ_DEF(pj_status_t) pjsua_acc_set_registration( pjsua_acc_id acc_id,
         
         //pjsua_process_msg_data(tdata, NULL);
         status = pjsip_regc_send( regc, tdata );
-        
+        if (status == PJSIP_EBUSY)
```
sauwming: If this is in the context of IP change, we have PR #3021 (included in 2.14) to solve the issue.
If not, can you please let us know the scenario?

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -272,6 +272,17 @@ pj_status_t pjsua_call_subsys_start(void)
 }
 
 
+/*
+* Destroy call subsystem.
+*/
```
sauwming: This function doesn't seem to be necessary. First, call is not a subsystem. Second, the function is used only once. And finally, `pjsua_var` will later be cleared by:
```
    /* Clear pjsua_var */
    pj_bzero(&pjsua_var, sizeof(pjsua_var));
```

So you can either put the code directly in `pjsua_destroy()`, or removing it altogether should be fine.

pjsip/include/pjsua-lib/pjsua.h
sauwming: everything else seems to have been addressed, except this one?

sauwming (commented):
We'd also appreciate if you can clean up the patch.
- avoid unnecessary changes
- fix accidental whitespace modifications

You can use the command:
- git diff [branch_you want_to_compare], for example: `git diff master` if you want to compare your patch against the master branch, or `git diff 2.13.1` to compare it with version 2.13.1
- for specific file or directory: `git diff 2.13.1 pjsip` or `git diff 2.13.1 pjsip/src/pjsua-lib/pjsua_call.c`

Make sure any changes that are not intended to be part of the patch do not show up in the diff results above.

sauwming:
We have merged #4585, thank you. Please clean up this patch so we can review it again.

sauwming:
There are still some more cleanup needed, so in order to assist you, I have asked Copilot AI to do it on your behalf. I have also broken it down to separate PRs for each added feature in #4645 and #4647. Therefore I will close this PR and continue it there. Feel free to join the review and submit your comments/feedbacks in those PRs. If there's anything missing, you can create a new PR (one PR per issue).

Thank you for the contributions.

PR #4585, opened 2025-09-08
Signed-off-by: Pesala Silva <pesala.silva@ifs.com>Bug_Fixes_and_optim…

PesalaDeSilva:
**A few bug fixes and optimizations**

- [ ] Added minor optimization in pjmedia_silence_det_detect to exit early if silence detection is disabled
- [ ] Bug fix in null_get_frame - to avoid overflowing during long calls or high sample rates on 64-bit systems
- [ ] Avoid crash in pjsip_tsx_retransmit_no_state if tsx is null (which happens occasionally especially under load)

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -3165,7 +3165,8 @@ static void stop_media_stream(pjsua_call *call, unsigned med_idx)
 static void stop_media_session(pjsua_call_id call_id)
 {
     pjsua_call *call = &pjsua_var.calls[call_id];
-    unsigned mi;   
+    unsigned mi;
+
     for (mi=0; mi<call->med_cnt; ++mi) {
         stop_media_stream(call, mi);
     }


pjmedia/src/pjmedia/conference.c
```diff
@@ -1302,7 +1302,10 @@ PJ_DEF(pj_status_t) pjmedia_conf_enum_ports( pjmedia_conf *conf,
         if (!conf->ports[i])
             continue;
 
-        ports[count++] = i;
+        if (ports)
+            ports[count++] = i;
+        else
+            count++;   // if port is NULL, only return the number of active slots.
```
sauwming: This proposal looks okay to me, but it requires modifications to be compatible with the current master. So I've created PR #4602 for this. 

You can revert this modification and the one in `pjsua_aud.c` and we'll continue the review process in the new PR.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -166,7 +165,7 @@ static void reset_call(pjsua_call_id id)
     call->last_text.ptr = call->last_text_buf_;
     call->cname.ptr = call->cname_buf;
     call->cname.slen = sizeof(call->cname_buf);
-    for (i=0; i<PJ_ARRAY_SIZE(call->media); ++i) {
+    for (unsigned i=0; i<PJ_ARRAY_SIZE(call->media); ++i) {
```
sauwming: Unfortunately, for compatibility issue, we should not do this.
https://docs.pjsip.org/en/latest/get-started/coding-style.html#no-declaration-after-statement

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3165,9 +3165,8 @@ static void stop_media_stream(pjsua_call *call, unsigned med_idx)
 static void stop_media_session(pjsua_call_id call_id)
 {
     pjsua_call *call = &pjsua_var.calls[call_id];
-    unsigned mi;
-
-    for (mi=0; mi<call->med_cnt; ++mi) {
+    
+    for (unsigned mi=0; mi<call->med_cnt; ++mi) {
```
sauwming: revert this.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -658,7 +658,7 @@ PJ_DEF(void) pjsua_call_setting_default(pjsua_call_setting *opt)
                                PJSUA_VID_REQ_KEYFRAME_RTCP_PLI;
 #endif
 
-    for (i = 0; i < PJMEDIA_MAX_SDP_MEDIA; i++) {
+    for (unsigned i = 0; i < PJMEDIA_MAX_SDP_MEDIA; i++) {
```
sauwming: revert this.

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -2788,7 +2788,7 @@ pj_status_t pjsua_media_channel_create_sdp(pjsua_call_id call_id,
                         const pjmedia_sdp_session *s_;
                         pjmedia_sdp_neg_get_active_local(call->inv->neg, &s_);
 
-                        if (mi < s_->media_count) {
+                        if (s_ && mi < s_->media_count) {
```
sauwming: Since the comment above states that we must be generating reoffer and in the assertion, the sdp neg state is DONE, I don't think `s_` can be NULL here. If yes, there's a bug somewhere else which must be fixed.
So, if you know the scenario, please inform us. Otherwise, for now, revert this since it can hide a more serious bug.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -2751,8 +2751,14 @@ PJ_DEF(pj_status_t) pjsua_call_answer2(pjsua_call_id call_id,
             answer->msg_data = pjsua_msg_data_clone(call->inv->pool_prov,
                                                     msg_data);
         }
-        pj_list_push_back(&call->async_call.call_var.inc_call.answers,
-                          answer);
+        if (call->async_call.call_var.inc_call.answers.prev) {
```
sauwming: This is not the appropriate check. `call->async_call.call_var.inc_call.answers` is the list of answers, so it's not indicative of whether the call has a media transport or not.
Perhaps you can tell us the scenario?

pjsip/src/pjsip/sip_transport.c
```diff
@@ -1593,7 +1593,7 @@ PJ_DEF(pj_status_t) pjsip_tpmgr_create( pj_pool_t *pool,
     if (!mgr->table)
         return PJ_ENOMEM;
 
-    status = pj_lock_create_recursive_mutex(mgr->pool, "tmgr%p", &mgr->lock);
+    status = pj_lock_create_recursive_mutex(pool, "tmgr%p", &mgr->lock);
```
sauwming: Why the change?
Allocating the mutex from transport manager's pool should be the correct way, instead of from the pool parameter.

pjsip/src/pjsip/sip_transaction.c
```diff
@@ -2378,19 +2378,27 @@ PJ_DEF(pj_status_t) pjsip_tsx_retransmit_no_state(pjsip_transaction *tsx,
 {
     pj_status_t status;
```
sauwming, line 2380: Just add `    PJ_ASSERT_RETURN(tsx != NULL, PJ_EINVAL);` here should be sufficient, I think (same check employed by similar APIs).

pjsip/src/pjsip/sip_util.c
```diff
@@ -1227,7 +1227,8 @@ static void stateless_send_transport_cb( void *token,
 
             if (tdata->via_addr.host.slen > 0 &&
                 (!tdata->via_tp ||
-                 tdata->via_tp == (void *)stateless_data->cur_transport))
+                    tdata->via_tp == (void*)stateless_data->cur_transport ||
+                    tdata->msg->line.req.method.id == PJSIP_CANCEL_METHOD))
```
sauwming, line 1231: This seems to contradict the code several lines above this:
```
        if (tdata->msg->line.req.method.id == PJSIP_CANCEL_METHOD) {
            if (via->sent_by.host.slen > 0) {
                /* Don't update Via header on a CANCEL request if the sent-by
                 * parameter is already set since it needs to match the 
                 * original request. */
                need_update_via = PJ_FALSE;
            }
        }
```

pjmedia/src/pjmedia/conference.c
sauwming: Update: as @nanangizz suggested, perhaps just change `pjsua_conf_get_active_ports()` to use `pjmedia_conf_get_port_count()` instead?

pjsip/src/pjsip/sip_transport.c
PesalaDeSilva: yes, it is better to use mgr->pool which is created specifically for the transport manager and therefore safe on managing resources

pjsip/src/pjsip/sip_util.c
PesalaDeSilva: Hi, thanks for the comment.
But in some situations a CANCEL request is created and it might not yet have a sent-by value (via->sent_by.host.slen == 0). 
In such cases, we still need to assign a sent-by,
 and for CANCEL the correct sent-by must match the original request’s tdata->via_addr, 
 not the current transport’s local name.

pjsip/src/pjsip/sip_transaction.c
```diff
@@ -2387,12 +2389,12 @@ PJ_DEF(pj_status_t) pjsip_tsx_retransmit_no_state(pjsip_transaction *tsx,
     pj_grp_lock_release(tsx->grp_lock);
 
     /* Only decrement reference counter when it returns success.
-     * (This is the specification from the .PDF design document).
-     */
+    * (This is the specification from the .PDF design document).
```
sauwming: revert this whitespace modifications, also below.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -2751,9 +2754,7 @@ PJ_DEF(pj_status_t) pjsua_call_answer2(pjsua_call_id call_id,
             answer->msg_data = pjsua_msg_data_clone(call->inv->pool_prov,
                                                     msg_data);
         }
-        pj_list_push_back(&call->async_call.call_var.inc_call.answers,
-                          answer);
-
+
```
sauwming, line 2761: accidental removal? put the original code back.

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3166,7 +3166,7 @@ static void stop_media_session(pjsua_call_id call_id)
 {
     pjsua_call *call = &pjsua_var.calls[call_id];
     unsigned mi;
-
+
```
sauwming: revert this.

sauwming (approved):

pjsip/include/pjsua-lib/pjsua_internal.h
```diff
@@ -530,7 +530,7 @@ struct pjsua_data
     /* Calls: */
     pjsua_config         ua_cfg;                /**< UA config.         */
     unsigned             call_cnt;              /**< Call counter.      */
-    pjsua_call           calls[PJSUA_MAX_CALLS];/**< Calls array.       */
+    pjsua_call           *calls;/**< Calls array.       */
```
nanangizz: Cosmetics:
- the field name alignment should be done on the name, not the `*` (see other pointer fields in the struct).
- please put back the spaces for docs/comment alignment.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -209,17 +209,20 @@ pj_status_t pjsua_call_subsys_init(const pjsua_config *cfg)
...
     /* Copy config */
     pjsua_config_dup(pjsua_var.pool, &pjsua_var.ua_cfg, cfg);
 
     /* Verify settings */
-    if (pjsua_var.ua_cfg.max_calls >= PJSUA_MAX_CALLS) {
+    if (pjsua_var.ua_cfg.max_calls > PJSUA_MAX_CALLS) 
         pjsua_var.ua_cfg.max_calls = PJSUA_MAX_CALLS;
-    }
+    
+    pjsua_var.calls = (pjsua_call*)pj_pool_zalloc(pjsua_var.pool, sizeof(pjsua_call) * pjsua_var.ua_cfg.max_calls);
```
nanangizz: Code style: [line length limit](https://docs.pjsip.org/en/latest/get-started/coding-style.html#limit-line-length) <= 80.

pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -3165,8 +3165,7 @@ static void stop_media_stream(pjsua_call *call, unsigned med_idx)
 static void stop_media_session(pjsua_call_id call_id)
 {
     pjsua_call *call = &pjsua_var.calls[call_id];
-    unsigned mi;
-
+    unsigned mi;
```
nanangizz: We put local variables in the beginning of function (FYI, we tried our best to follow ANSI C for portability, [see here](https://docs.pjsip.org/en/latest/get-started/coding-style.html#use-ansi-c-c90-standard)), not before its usage. So the original is actually better.

sauwming:
@copilot: make the modifications as per @nanangizz's review here: https://github.com/pjsip/pjproject/pull/4585#pullrequestreview-3287896603. Only those three modifications are needed and there must be no other changes.

Issue #3969, opened 2024-05-23
Building python swig module fails due to either bad version written in version.mak or a bug in setup.py 


frisky5:
### Describe the bug

upon cloning, and building with opus, and then trying to build python swig module i got this error

> Traceback (most recent call last):
>   File "/root/pjproject/pjsip-apps/src/swig/python/setup.py", line 106, in <module>
>     setup(name="pjsua2", 
>   File "/usr/local/lib/python3.10/dist-packages/setuptools/_distutils/core.py", line 146, in setup
>     _setup_distribution = dist = klass(attrs)
>   File "/usr/local/lib/python3.10/dist-packages/setuptools/dist.py", line 318, in __init__
>     self.metadata.version = self._normalize_version(self.metadata.version)
>   File "/usr/local/lib/python3.10/dist-packages/setuptools/dist.py", line 354, in _normalize_version
>     normalized = str(Version(version))
>   File "/usr/local/lib/python3.10/dist-packages/setuptools/_vendor/packaging/version.py", line 200, in __init__
>     raise InvalidVersion(f"Invalid version: '{version}'")
> setuptools.extern.packaging.version.InvalidVersion: Invalid version: '2.14--dev'
> make: *** [Makefile:42: _pjsua2.so] Error 1

so i modified the setup.py file that runs in the make process at line 59 to remove the concatenation of '-', then the building of swig worked

```
if pj_version_suffix:
     #pj_version += "-" + pj_version_suffix
     pj_version += pj_version_suffix 
```

not sure if the bug is in the setup.py or the version.mak file, but the version.mak file mentions to never be modified manually

### Steps to reproduce

1. clone master
2. build pjsip
3. build python swig module

### PJSIP version

2.14

### Context

building inside an ubuntu 20 container.

### Log, call stack, etc

```shell
N/A
```

HayashiNiizuki:
i got this error when build python swig module 

```
/home/max/miniconda3/envs/pjsua2/compiler_compat/ld: /mnt/d/Coding/Application/pjproject-2.14.1/pjsip/lib/libpjsua2-x86_64-unknown-linux-gnu.a(media.o): warning: relocation against `_ZN2pj16AudioMediaPlayer6eof_cbEP12pjmedia_portPv' in read-only section `.text'
[...]

PR #3976, opened 2024-05-28
Fix version string in Python setup

sauwming:
To fix #3969.

Our version suffix convention already includes '-' (such as "-svn" back when we still used SVN, or "-dev" for now), so we don't need the extra '-'.
Thanks to @frisky5 for the fix.

PR #4594, opened 2025-09-14
Fix version string for Python setup

Tugaytalha:
To fix https://github.com/pjsip/pjproject/issues/3969.

Our version suffix convention already includes '-' (such as "-svn" back when we still used SVN, or "-dev" for now), so we don't need the extra '-'.

"-" comes from another part of the code and causes version became "pjsua2.15--dev" instead of "pjsua2.15-dev" and fails to python build.

### DIFF
--- a/version.mak
+++ b/version.mak
@@ -2,7 +2,7 @@
 export PJ_VERSION_MAJOR  := 2
 export PJ_VERSION_MINOR  := 15
 export PJ_VERSION_REV    :=
-export PJ_VERSION_SUFFIX := -dev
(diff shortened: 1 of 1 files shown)


nanangizz:
The `PJ_VERSION_SUFFIX` format is already correct with "-", it must not be modified.
Moreover, the issue should have been fixed by #3976.

Tugaytalha:
I ran into the same error again while trying to set everything up and was able to fix it with this solution. However, I can't reproduce the error now, I probably checked out the wrong branch during installation or something like that. Thanks, and good luck with your work!

Issue #4596, opened 2025-09-16
Asterisk with pjsip short-dumps when OpenSSL is switched to FIPS mode


nathanjel:
### Describe the bug

When OpenSSL is switched to FIPS mode (e.g. on Ubuntu using pro support) MD5-digest based, asterisk will segfault trying to register when SIP trunk vendor allows only MD5 digest.

### Steps to reproduce

0. get ubuntu 24.04, connect to pro account, pro enable fips
1. install Asterisk connecting to a SIP trunk with user/pass authorization and vendor expects MD5 digest
2. start Asterisk, it will segfault when trying to REGISTER

### PJSIP version

2.15.1

### Context

- ubuntu 24.04, pro, FIPS enabled
- Asterisk 22.5.2
- supervoip.pl SIP trunk provider with user/pass authorization

relevant config:
[supervoip-auth]
type = auth
auth_type = digest
username = YOUR_USER
password = YOUR_PASS
supported_algorithms_uac = SHA-256, SHA-512-256, MD5

### Log, call stack, etc

```shell
[...]
Asterisk Ready. [New Thread 0x7fffa545d6c0 (LWP 542594)] *CLI> [New Thread 0x7fffa53e06c0 (LWP 542861)] [2025-09-16 12:13:03] ERROR[542559]: pjproject: <?>: tlsc0x7fff9c21bbc8 RFC 5922 (section 7.2) does not allow TLS wildcard certificates. Advise your SIP provider, please! [2025-09-16 12:13:03] NOTICE[542559]: res_pjsip/pjsip_transport_events.c:190 verify_log_result: Transport 'transport-tls' to remote 'sip.supervoip.pl' - 213.199.246.200:5061 - OK

[...]

PR #4597, opened 2025-09-16
Fallback to builtin MD5 when OpenSSL MD5 is unavailable (e.g., FIPS)

nathanjel:
Some OpenSSL/FIPS builds disable MD5, causing pjsip_auth_create_digest2() to fail when computing MD5 digests. This change adds a one‑time runtime check for OpenSSL MD5 availability and, for MD5 only, falls back to pjlib’s internal MD5 (pj_md5_*) when EVP MD5 is not usable. When OpenSSL MD5 is available, behavior is unchanged; SHA‑256 and other algorithms continue to use OpenSSL EVP.

The change is minimal and localized to sip_auth_client.c. We also allow MD5 to be reported as “supported” so the fallback path is exercised instead of returning an error. No extra logging is added. Tests exercising SIP auth (REGISTER with Digest) pass with both OpenSSL‑MD5 and FIPS‑style environments.

### DIFF
--- a/pjsip/src/pjsip/sip_auth_client.c
+++ b/pjsip/src/pjsip/sip_auth_client.c
@@ -77,6 +77,35 @@
 #define EVP_MD_CTX_free(mdctx)
 #endif
 
+/*
+ * When building with OpenSSL, MD5 may not be available (e.g., FIPS-only
+ * configurations). Detect MD5 availability once and, if missing, fall back
+ * to pjlib's internal MD5 implementation for MD5-based digests.
+ */
+#if defined(PJ_HAS_SSL_SOCK) && PJ_HAS_SSL_SOCK != 0 && \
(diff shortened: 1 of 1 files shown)


nathanjel:
#4596

sauwming:
Just wondering, whether it's okay to detect this during compile-time instead by checking `OPENSSL_NO_MD5`?

If yes, it will be much easier since we already have the fallback macro to use `pj_md5` here:
https://github.com/pjsip/pjproject/blob/master/pjsip/src/pjsip/sip_auth_client.c#L75

nathanjel:
From my perspective compile time is too early, if you consider distro packages builds, and then need to have another version of downstream - e.g. asterisk build, depending on different OpenSSL packages. Doable, but not sure if practical - considering my experience. At runtime this is once at init, I would say this a limited cost.

nanangizz (approved):

sauwming (approved):

Issue #4589, opened 2025-09-10
PSJUA2: Answer incoming SIP MESSAGE requests at a later point in time


wosrediinanatour:
### Describe the feature

Currently, an incoming SIP MESSAGE is answered by SIP 200 OK immediately.

A UAS could be created and Account::onInstantMessage() could deliver a transaction ID, which then could be used in a new Buddy::answer() for answering a request.

Most probably, an account/endpoint configuration would be needed for backward compatibility.

### Describe alternatives you've considered

Class Account is already able to send outgoing request. This could be used for SIP MESSAGE as well.
The alternative would be to make "presence" optional in PJSUA2, so that for incoming MESSAGE an own account could be used. But - a "link" to accounts would be needed as well, which would make the change similar complex than the primary approach.

sauwming:
I'm curious though, what's your use case here? i.e. why do you need to delay the MESSAGE response?

In `on_call_rx_reinvite()`, the use case was for third party media, so user receives re-INVITE, initializes third party media asynchronously, which means that they can't immediately respond until the initialization completes later.

wosrediinanatour:
The PJSUA(s)'s callbacks are not executed in the main thread. In a single-threaded application, posting to the main thread (business logic) and allowing to have a negative response is a very useful scenario. More complex examples incorporate messaging to other processes. A B2BUA is the most general scenario. 

Updated feature request.

PR #4598, opened 2025-09-17
Defer response for incoming SIP MESSAGE requests

wosrediinanatour:
For an incoming SIP MESSAGE request, the response is sent immediately. This is a backward-compatible minimal approach for processing incoming SIP MESSAGE transactions asynchronously (see #4589) by deferring the response.

A new account configuration disables immediately response, whereas a new function (PJSUA) can be used for send a response.

It is now possible to respond with a status code != 200.

Even with PJSUA2, deferring a response is possible:
```cpp
void onInstantMessage(pj::OnInstantMessageParam& parameter) override
     {
     pjsip_rx_data* received_data_clone = nullptr;
     // Life-time of msg_info ends at the end of this scope. Copies only msg_info deeply.
     pjsip_rx_data_clone(static_cast<pjsip_rx_data*>(parameter.rdata.pjRxData), 0,
                         &received_data_clone);
     // Life-time ends when PJSIP transaction ends, i.e. after calling pjsua_im_respond_messag(...).
     auto* tsx{pjsip_rdata_get_tsx(static_cast<pjsip_rx_data*>(parameter.rdata.pjRxData))};
     
    //...
     }

// Most probably called from the main thread.
void sendResponse(pjsip_rx_data *received_data_clone, pjsip_transaction *tsx)
    {
    auto text{to_pj_string("OK")};
    pjsua_im_respond_message(&received_data_clone, tsx, 200, &text, nullptr);
    // Life-time of tsx ends very soon.
    }
```

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -4782,6 +4782,21 @@ typedef struct pjsua_acc_config
      */
     pj_bool_t        use_shared_auth;
 
+    /**
+    * Specify how to process incoming SIP MESSAGE requests.
+    *
+    * When set to PJ_TRUE (asynchronous processing), incoming MESSAGE requests
+    * will create UAS transactions and defer response sending. Applications
+    * must use pjsua_im_respond_message() to send responses manually.
+    *
+    * When set to PJ_FALSE (immediate processing), incoming MESSAGE requests
+    * will be responded to immediately with SIP 200 OK (legacy behavior).
+    *
+    * Default: PJ_FALSE (immediate response for backward compatibility)
+    */
+    pj_bool_t        process_sip_message_async;
+
+
 } pjsua_acc_config;
 
 
@@ -7283,6 +7298,26 @@ PJ_DECL(pj_status_t) pjsua_im_typing(pjsua_acc_id acc_id,
                                      pj_bool_t is_typing,
                                      const pjsua_msg_data *msg_data);
 
+/**
+ * Send response to incoming MESSAGE request using the UAS transaction.
+ * This function should be used when the application wants to send a deferred
+ * response to an incoming MESSAGE request that was processed by a UAS transaction.
+ *
+ * @param rdata         The incoming MESSAGE request data (from on_pager2 callback).
+ * @param st_code       Status code to be sent (e.g., 200 for OK, 400 for
+ *                      Bad Request, etc.).
+ * @param st_text       Optional status text. If NULL, default status text
+ *                      for the status code will be used.
+ * @param hdr_list      Optional list of headers to be added to the response.
+ * @param body          Optional message body for the response.
+ *
+ * @return              PJ_SUCCESS on success, or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsua_im_respond_message(pjsip_rx_data *rdata,
+                                              pjsip_transaction *tsx,
+                                              int st_code,
+                                              const pj_str_t *st_text,
+                                              const pjsip_hdr *hdr_list);
 
 
 /**
--- a/pjsip/include/pjsua2/account.hpp
+++ b/pjsip/include/pjsua2/account.hpp
@@ -311,6 +311,17 @@ struct AccountSipConfig : public PersistentObject
      */
     bool        useSharedAuth;
 
+    /**
+     * Configure SIP MESSAGE processing behavior for this account.
+     * When set to true, incoming SIP MESSAGE requests will create UAS 
+     * transactions allowing for deferred responses. When set to false,
+     * incoming SIP MESSAGE requests will be responded to immediately
+     * with a 200 OK response (legacy behavior).
+     *
+     * Default: false (immediate response for backward compatibility)
+     */
+    bool        processSipMessageAsync;
+
 public:
     /**
      * Read this object from a container node.
--- a/pjsip/src/pjsua-lib/pjsua_im.c
+++ b/pjsip/src/pjsua-lib/pjsua_im.c
@@ -146,6 +146,21 @@ pj_bool_t pjsua_im_accept_pager(pjsip_rx_data *rdata,
     return PJ_TRUE;
 }
 
+static pj_bool_t is_typing_indication(pjsip_rx_data *rdata)
+{
+    PJ_ASSERT_RETURN(rdata, PJ_FALSE);
+    PJ_ASSERT_RETURN(rdata->msg_info.msg, PJ_FALSE);
+
+    pjsip_msg_body *body = rdata->msg_info.msg->body;
+
+    if (body && pj_stricmp(&body->content_type.type, &STR_MIME_APP)==0 &&
+        pj_stricmp(&body->content_type.subtype, &STR_MIME_ISCOMPOSING)==0)
+    {
+        return PJ_TRUE;
+    }
+    return PJ_FALSE;
+}
+
 /**
  * Private: process pager message.
  *          This may trigger pjsua_ui_on_pager() or pjsua_ui_on_typing().
@@ -178,8 +193,7 @@ void pjsua_im_process_pager(int call_id, const pj_str_t *from,
         contact.slen = 0;
     }
 
-    if (body && pj_stricmp(&body->content_type.type, &STR_MIME_APP)==0 &&
-        pj_stricmp(&body->content_type.subtype, &STR_MIME_ISCOMPOSING)==0)
+    if (is_typing_indication(rdata) == PJ_TRUE)
     {
         /* Expecting typing indication */
         pj_status_t status;
@@ -273,6 +287,7 @@ void pjsua_im_process_pager(int call_id, const pj_str_t *from,
  */
 static pj_bool_t im_on_rx_request(pjsip_rx_data *rdata)
 {
+    pjsua_acc_id acc_id = PJSUA_INVALID_ID;
     pj_str_t from, to;
     pjsip_accept_hdr *accept_hdr;
     pjsip_msg *msg;
@@ -304,13 +319,30 @@ static pj_bool_t im_on_rx_request(pjsip_rx_data *rdata)
         return PJ_TRUE;
     }
     
-    /* Respond with 200 first, so that remote doesn't retransmit in case
-     * the UI takes too long to process the message. 
-     */
-    pjsip_endpt_respond( pjsua_var.endpt, NULL, rdata, 200, NULL,
-                         NULL, NULL, NULL);
+    acc_id = pjsua_acc_find_for_incoming(rdata);
+    
+    if (acc_id != PJSUA_INVALID_ID && pjsua_var.acc[acc_id].cfg.process_sip_message_async == PJ_TRUE
+            && is_typing_indication(rdata) == PJ_FALSE) {
+        pjsip_transaction *tsx;
+        pj_status_t status;
 
-    /* For the source URI, we use Contact header if present, since
+        status = pjsip_tsx_create_uas(&mod_pjsua_im, rdata, &tsx);
+        if (status != PJ_SUCCESS) {
+            /* Fall back to immediate response if UAS creation fails */
+            pjsip_endpt_respond(pjsua_var.endpt, NULL, rdata, 500, NULL,
+                               NULL, NULL, NULL);
+            return PJ_TRUE;
+        }
+
+        pjsip_tsx_recv_msg(tsx, rdata);
+    }
+    else
+    {
+        /* Respond with 200 OK */
+        pjsip_endpt_respond( pjsua_var.endpt, NULL, rdata, 200, NULL,
+                             NULL, NULL, NULL);
+    }
+     /* For the source URI, we use Contact header if present, since
      * Contact header contains the port number information. If this is
      * not available, then use From header.
      */
@@ -331,7 +363,7 @@ static pj_bool_t im_on_rx_request(pjsip_rx_data *rdata)
         to = pj_str("<--URI is too long-->");
 
     /* Process pager. */
-    pjsua_im_process_pager(-1, &from, &to, rdata);
+    pjsua_im_process_pager(PJSUA_INVALID_ID, &from, &to, rdata);
 
     /* Done. */
     return PJ_TRUE;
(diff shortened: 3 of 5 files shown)


wosrediinanatour:
I added a test, where you can see how it is working (both PJSUA and PJSUA2).

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -4782,6 +4782,21 @@ typedef struct pjsua_acc_config
...
+    * When set to PJ_TRUE (asynchronous processing), incoming MESSAGE requests
+    * will create UAS transactions and defer response sending. Applications
+    * must use pjsua_im_respond_message() to send responses manually.
+    *
+    * When set to PJ_FALSE (immediate processing), incoming MESSAGE requests
+    * will be responded to immediately with SIP 200 OK (legacy behavior).
+    *
+    * Default: PJ_FALSE (immediate response for backward compatibility)
+    */
+    pj_bool_t        process_sip_message_async;
```
sauwming: I'm not sure if `async` is the right term here, since it typically refers to blocking/non-blocking. Perhaps `auto` will be more clear and understandable, such as the setting `PJSUA_DISABLE_AUTO_SEND_100`.

pjsip/include/pjsua2/account.hpp
```diff
@@ -1931,6 +1940,127 @@ class FindBuddyMatch
     virtual ~FindBuddyMatch() {}
 };
 
+/**
```
sauwming, line 1943: The usage of DeferredResponse seems rather lengthy and convoluted.

If possible, we should make it as simple as possible for the users. How about something like this (note that this is still a rough idea, so it may not work):
- Add `Token` inside `onInstantMessageParam`.
- Then when user wants to respond, just call `account->sendInstantMessageResponse(Token, statusCode , reason);`
- pjsua2 will manage the DeferredResponse under the hood (so it does not need to be visible to the user).

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -7283,6 +7298,26 @@ PJ_DECL(pj_status_t) pjsua_im_typing(pjsua_acc_id acc_id,
...
+ * @param st_code       Status code to be sent (e.g., 200 for OK, 400 for
+ *                      Bad Request, etc.).
+ * @param st_text       Optional status text. If NULL, default status text
+ *                      for the status code will be used.
+ * @param hdr_list      Optional list of headers to be added to the response.
+ * @param body          Optional message body for the response.
+ *
+ * @return              PJ_SUCCESS on success, or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsua_im_respond_message(pjsip_rx_data *rdata,
```
sauwming: Since it's account-tied API, we should still add `pjsua_acc_id` as the first parameter, even though the parameter later will not be used.

pjsip/include/pjsua-lib/pjsua.h
wosrediinanatour: I tried to be consistent with
```c
on_call_rx_reinvite(..., pj_bool_t *async, ...)
```

But I agree - `auto_send_200 = PJ_FALSE (default)` would be OK as well.
What do you think?

pjsip/include/pjsua2/account.hpp
wosrediinanatour: `DeferredResponse` is the token, although not explicit. `DeferredResponse ` is just the ["Rule of 5 (or 6)"](https://www.modernescpp.com/index.php/c-core-guidelines-constructors-assignments-and-desctructors/), so it is nothing special in C++.

The problem is that we need rdata for generating tdata. Or, even if we could make a tdata before (although we don't know the response code), we still would have the same issue:
* rdata has to be cloned, since it becomes destroyed after the fallback
* We need both: non-owning pointer to the transaction and owning pointer to the rdata
* We cannot get rid of the the `DeferredResponse` in the PJSUA2 layer (C++). Because, to be able to save "tokens" to a container, we need to have a move and copy constructors and assignment operators. Because the copy constructors/assignment operators still have to clone "rdata".
* Even if we move the cloning (deep copying) to the lower levels and copy depending on the configuration, we still need it in the user space, when we want to store it. So I would opt against it for locality.

The only improvement, I would come up, is making a similar "class" in C as well (PJUSA):
* A struct that combines pointers `pjsip_rx_data*` and `pjsip_transaction*`.
* Functions that shallow copies and deep copy
* Note that the "rather lengthy and convoluted" code would then be in the user space (as now).
* The struct could be put to the callbacks as you suggested. For C++ this would be OK. But for C (PJSUA) this would mean to add another version of the callback. Something I frankly don't like much because it is not backward compatible.

pjsip/include/pjsua-lib/pjsua.h
sauwming: Ah okay, I forgot about `on_call_rx_reinvite()` (even though I was the one that implemented it :) ). I believe the context that time was for async third party media.

`auto_process_sip_message` sounds fine to me (`auto_send_200_for_sip_message` seems too long).

pjsip/include/pjsua-lib/pjsua.h
wosrediinanatour: Good point!

This brings up a question.

Do we want to have explicit "instant messaging" in the function names, as I have now implemented?
(Since we have already dedicated callbacks I chose this naming)

Because, there is nothing really instant-messaging related at the moment. Sometimes in the future, it may be used for other responses as well (.e.g responsing to OPTIONS)...?

pjsip/include/pjsua2/account.hpp
sauwming: > * pjsua2 will manage the DeferredResponse under the hood (so it does not need to be visible to the user).

What I meant was that DeferredResponse to be moved to `account.cpp`, so all the things you mentioned will still be done, but will be transparent to the user.

pjsip/include/pjsua-lib/pjsua.h
wosrediinanatour: OK, I am going to change it.

pjsip/include/pjsua2/account.hpp
wosrediinanatour: OK! Of course! I will change it.

pjsip/include/pjsua-lib/pjsua.h
sauwming: If it's not necessary, then it doesn't have to be tied to "instant messaging".

For example, since we already have `pjsua_acc_send_request()`, adding `pjsua_acc_send_response()` seems to make sense. Or you have other ideas?

wosrediinanatour: Sounds good!

wosrediinanatour:
@sauwming: Applied the changes - PTAL for changes https://github.com/pjsip/pjproject/compare/4ec864123be3a04598b5bc7242ae223690371eff..0e6bbf65902801f0e5dbffacf7e9203cc7842fc6

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -5116,6 +5131,28 @@ PJ_DECL(pj_status_t) pjsua_acc_get_config(pjsua_acc_id acc_id,
 PJ_DECL(pj_status_t) pjsua_acc_modify(pjsua_acc_id acc_id,
                                       const pjsua_acc_config *acc_cfg);
 
+/**
+ * Send response to of an UAS transaction.
+ * This function should be used when the application wants to send a deferred
+ * response to an incoming request (e.g. SIP MESSAGE) that was processed by 
+ * a UAS transaction.
+ *
```
sauwming, line 5148: doc for @param acc_id and tsx.

pjsip/src/pjsua2-test/test_deferred_message_response.hpp
```diff
@@ -0,0 +1,245 @@
...
+ * @file test_deferred_message_response.cpp
+ * @brief Test program for PJSUA2 deferred SIP MESSAGE responses
+ *
+ * This test program demonstrates the new deferred SIP MESSAGE response 
+ * functionality in PJSUA2. It creates two accounts where:
+ * 1. First account receives SIP MESSAGE and defers the response
+ * 2. Second account sends SIP MESSAGE to the first account
+ * 3. First account later sends a deferred response with specific status code
+ */
+
```
sauwming: `samples` is supposed to be for sample apps, so would it be possible to put this inside `pjsua2_demo` instead? All the howto examples are currently put there.

Alternatively, maybe move to `pjsip/src/pjsua2-test`. This will be the first one though.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -5116,6 +5131,28 @@ PJ_DECL(pj_status_t) pjsua_acc_get_config(pjsua_acc_id acc_id,
...
+ * @param rdata         The received request data (from on_pager2 callback).
```
sauwming: Since the API is now generic, better remove the reference to `on_pager2` callback. It gives the impression that rdata can only comes from that callback.

pjsip/src/pjsua2-test/test_deferred_message_response.hpp
wosrediinanatour: I opt for the alternative. It tests the implementation, so could be e.g. used in CI pipeline for regression testing.

Btw: It also tests the PJSUA implementation.

wosrediinanatour:
Incorporated @sauwming's feedback: https://github.com/pjsip/pjproject/compare/715b3248b49dd729aae674c36ba50586545ede51..1b128f655ad055e50014c57a7b8ec5a9a9954268

pjsip/build/Makefile
```diff
@@ -171,6 +171,19 @@ endif
 export CC_OUT CC AR RANLIB HOST_MV HOST_RM HOST_RMDIR HOST_MKDIR OBJEXT LD LDOUT 
 
 
+###############################################################################
```
sauwming: you can put it into pjsua2-test, so you won't need a separate build section.

It will be something similar like pjsip-test, pjlib-test, ..., etc

wosrediinanatour: I have done this - see https://github.com/pjsip/pjproject/compare/1b128f655ad055e50014c57a7b8ec5a9a9954268..ccd3498652137c8592083d9453895cb5d85c7dd5.

pjsip-apps/build/Samples.mak
```diff
@@ -48,8 +48,7 @@ SAMPLES := auddemo \
 	   vid_codec_test \
 	   vid_streamutil
 
-PJSUA2_SAMPLES := pjsua2_demo \
```
sauwming, line 51: (minor) exclude this from the PR

pjsip/build/Makefile
```diff
@@ -170,7 +170,6 @@ export PJSUA2_TEST_EXE := pjsua2-test-$(TARGET_NAME)$(HOST_EXE)
 endif
 export CC_OUT CC AR RANLIB HOST_MV HOST_RM HOST_RMDIR HOST_MKDIR OBJEXT LD LDOUT 
 
-
```
sauwming, line 173: and also this

pjsip/src/pjsua2-test/main.cpp
```diff
@@ -16,23 +16,19 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
  */
 #include <pjsua2/endpoint.hpp>
+#include "test_deferred_message_response.hpp"
```
sauwming: cpp should be better since it's implementation file.

We do this in other parts of the library, for example `#include "ssl_sock_imp_common.c"`.

sauwming (approved):
Just a few minors. That's all from me.

wosrediinanatour:
I split unit tests into header and implementation. Since I had to change several lines, I also added a second unit test for testing the "immediate response". PTAL at the changes https://github.com/pjsip/pjproject/compare/ccd3498652137c8592083d9453895cb5d85c7dd5..0a0d875c27a310a378b70c143e36e8484bc95bec

wosrediinanatour:
Additional minor changes pushed: https://github.com/pjsip/pjproject/compare/0a0d875c27a310a378b70c143e36e8484bc95bec..af81002ad4ca2c1b871344d33377c7861e9a7d6c

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -5116,6 +5131,33 @@ PJ_DECL(pj_status_t) pjsua_acc_get_config(pjsua_acc_id acc_id,
...
+ *
+ * @return              PJ_SUCCESS on success, or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsua_acc_send_response(pjsua_acc_id acc_id,
+                                             pjsip_rx_data *rdata,
+                                             pjsip_transaction *tsx,
+                                             int st_code,
+                                             const pj_str_t *st_text,
+                                             const pjsip_hdr *hdr_list);
+
```
nanangizz, line 5168: I think it is better to use `const pjsua_msg_data *msg_data` instead of `hdr_list` if possible.

pjsip/src/pjsua-lib/pjsua_im.c
```diff
@@ -304,13 +319,30 @@ static pj_bool_t im_on_rx_request(pjsip_rx_data *rdata)
...
     
-    /* Respond with 200 first, so that remote doesn't retransmit in case
-     * the UI takes too long to process the message. 
-     */
-    pjsip_endpt_respond( pjsua_var.endpt, NULL, rdata, 200, NULL,
-                         NULL, NULL, NULL);
+    acc_id = pjsua_acc_find_for_incoming(rdata);
+    
+    if (acc_id != PJSUA_INVALID_ID && pjsua_var.acc[acc_id].cfg.auto_process_sip_message == PJ_FALSE
+            && is_typing_indication(rdata) == PJ_FALSE) {
```
nanangizz: Minor:
- Coding style for [brace in multiline statement](https://docs.pjsip.org/en/latest/get-started/coding-style.html#use-k-r-style-brace-placement), the brace should be in a separate line.
- While in coding style, please also recheck for [line length limit](https://docs.pjsip.org/en/latest/get-started/coding-style.html#limit-line-length) in the whole PR code.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -4782,6 +4782,21 @@ typedef struct pjsua_acc_config
...
 
+    /**
+    * Specify how to automatically process incoming SIP MESSAGE requests.
+    *
+    * When set to PJ_TRUE (automatic processing), incoming MESSAGE requests
+    * will be responded to immediately with SIP 200 OK (legacy behavior).
+    *
+    * When set to PJ_FALSE (manual processing), incoming MESSAGE requests
+    * will create UAS transactions and defer response sending. Applications
+    * must use pjsua_acc_send_response() to send responses manually.
```
nanangizz: How about also adding a brief hint about the required steps for the app in the docs, for example:

> app must clone the request data (using ...) and obtain the transaction instance (using ...) from the `on_pager()/on_pager2()` callback, which is then used to send the response using `pjsua_acc_send_response()`.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -4782,6 +4782,21 @@ typedef struct pjsua_acc_config
...
+    *
+    * Default: PJ_TRUE (automatic response for backward compatibility)
+    */
+    pj_bool_t        auto_process_sip_message;
```
nanangizz: IMHO `auto_process` sounds somewhat ambigous as the libray is merely sending a response automatically, not fully processing it, so how about `auto_respond` instead?

wosrediinanatour:
Incorporated @nanangizz feedback in https://github.com/pjsip/pjproject/compare/3aa2faf28cf0cc6734d0a7494f710ca0905e177f..2a859d0ed260d2a92fcc39c132fa9e857b55ca09.

wosrediinanatour:
I hope that https://github.com/pjsip/pjproject/compare/2a859d0ed260d2a92fcc39c132fa9e857b55ca09..3f81713293c421f5cf303b5de760470836934c03 now fixes all ill-formatted code.

PR #4599, opened 2025-09-18
Expose SIP and media ioqueue fds via pjsua2 Endpoint

IkerAriz:
This change lets python users integrate pjsua2 with their own event loops. Support is limited to platforms where the underlying PJ ioqueues are associated with a file descriptor (currently those that use epoll and kqueue).

### DIFF
--- a/pjlib/include/pj/ioqueue.h
+++ b/pjlib/include/pj/ioqueue.h
@@ -944,6 +944,17 @@ PJ_DECL(pj_status_t) pj_ioqueue_sendto( pj_ioqueue_key_t *key,
  */
 PJ_DECL(pj_oshandle_t) pj_ioqueue_get_os_handle( pj_ioqueue_t *ioqueue );
 
+/**
+ * Get the file descriptor associated with an ioqueue instance.
+ *
+ * @param ioqueue        The ioqueue instance.
+ *
+ * @return          The OS file descriptor associated with the instance.
+ *                  For epoll/kqueue this will be a value >= 0. For all
+ *                  other platforms, or if no handle is available, -1 will
+*                   be returned.
+ */
+PJ_DECL(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue );
 
 /**
  * @}
--- a/pjsip/include/pjsua2/endpoint.hpp
+++ b/pjsip/include/pjsua2/endpoint.hpp
@@ -1545,6 +1545,19 @@ public:
      */
     void libDestroy(unsigned prmFlags=0) PJSUA2_THROW(Error);
 
+    /**
+     * Get the file descriptor associated with pjsua's SIP IO queue, if any.
+     * Returns an integer >= 0 if the IO queue is implemented with epoll or
+     * kqueue, otherwise returns -1.
+     */
+    int libGetSipIoQueueFd();
+
+    /**
+     * Get the file descriptor associated with pjsua's media IO queue, if any.
+     * Returns an integer >= 0 if the IO queue is implemented with epoll or
+     * kqueue, otherwise returns -1.
+     */
+    int libGetMediaIoQueueFd();
 
     /*************************************************************************
      * Utilities
--- a/pjlib/src/pj/ioqueue_dummy.c
+++ b/pjlib/src/pj/ioqueue_dummy.c
@@ -202,4 +202,10 @@ PJ_DEF(pj_oshandle_t) pj_ioqueue_get_os_handle( pj_ioqueue_t *ioqueue )
     return NULL;
 }
 
+PJ_DEF(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue )
+{
+    PJ_UNUSED_ARG(ioqueue);
+    return -1;
+}
+
 #endif /* PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_NONE */
--- a/pjlib/src/pj/ioqueue_epoll.c
+++ b/pjlib/src/pj/ioqueue_epoll.c
@@ -1088,4 +1088,9 @@ PJ_DEF(pj_oshandle_t) pj_ioqueue_get_os_handle( pj_ioqueue_t *ioqueue )
     return ioqueue ? (pj_oshandle_t)&ioqueue->epfd : NULL;
 }
 
+PJ_DEF(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue )
+{
+    return ioqueue ? ioqueue->epfd : -1;
+}
+
 #endif /* PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_EPOLL */
--- a/pjlib/src/pj/ioqueue_kqueue.c
+++ b/pjlib/src/pj/ioqueue_kqueue.c
@@ -733,4 +733,9 @@ PJ_DEF(pj_oshandle_t) pj_ioqueue_get_os_handle( pj_ioqueue_t *ioqueue )
     return ioqueue ? (pj_oshandle_t)&ioqueue->kfd : NULL;
 }
 
+PJ_DEF(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue )
+{
+    return ioqueue ? ioqueue->kfd : -1;
+}
+
 #endif /* PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_KQUEUE */
--- a/pjlib/src/pj/ioqueue_select.c
+++ b/pjlib/src/pj/ioqueue_select.c
@@ -1148,4 +1148,10 @@ PJ_DEF(pj_oshandle_t) pj_ioqueue_get_os_handle( pj_ioqueue_t *ioqueue )
     return NULL;
 }
 
+PJ_DEF(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue )
+{
+    PJ_UNUSED_ARG(ioqueue);
+    return -1;
+}
+
 #endif /* PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_SELECT */
--- a/pjlib/src/pj/ioqueue_symbian.cpp
+++ b/pjlib/src/pj/ioqueue_symbian.cpp
@@ -875,4 +875,10 @@ PJ_DEF(pj_oshandle_t) pj_ioqueue_get_os_handle( pj_ioqueue_t *ioqueue )
     return NULL;
 }
 
+PJ_DEF(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue )
+{
+    PJ_UNUSED_ARG(ioqueue);
+    return -1;
+}
+
 #endif /* PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_SYMBIAN */
--- a/pjlib/src/pj/ioqueue_uwp.cpp
+++ b/pjlib/src/pj/ioqueue_uwp.cpp
@@ -372,4 +372,10 @@ PJ_DEF(pj_oshandle_t) pj_ioqueue_get_os_handle( pj_ioqueue_t *ioqueue )
     return NULL;
 }
 
+PJ_DEF(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue )
+{
+    PJ_UNUSED_ARG(ioqueue);
+    return -1;
+}
+
 #endif /* PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_UWP */
--- a/pjlib/src/pj/ioqueue_winnt.c
+++ b/pjlib/src/pj/ioqueue_winnt.c
@@ -1699,5 +1699,10 @@ PJ_DEF(pj_oshandle_t) pj_ioqueue_get_os_handle( pj_ioqueue_t *ioqueue )
     return ioqueue ? (pj_oshandle_t)ioqueue->iocp : NULL;
 }
 
+PJ_DEF(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue )
+{
+    PJ_UNUSED_ARG(ioqueue);
+    return -1;
+}
 
 #endif /* PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_IOCP */
--- a/pjsip/src/pjsua2/endpoint.cpp
+++ b/pjsip/src/pjsua2/endpoint.cpp
@@ -2244,6 +2244,20 @@ void Endpoint::libDestroy(unsigned flags) PJSUA2_THROW(Error)
     PJSUA2_CHECK_RAISE_ERROR(status);
 }
 
+int Endpoint::libGetSipIoQueueFd()
+{
+    pjsip_endpoint* endp = pjsua_get_pjsip_endpt();
+    pj_ioqueue_t* ioq = pjsip_endpt_get_ioqueue(endp);
+    return pj_ioqueue_get_os_fd(ioq);
+}
+
+int Endpoint::libGetMediaIoQueueFd()
+{
+    pjmedia_endpt* endp = pjsua_get_pjmedia_endpt();
+    pj_ioqueue_t* ioq = pjmedia_endpt_get_ioqueue(endp);
+    return pj_ioqueue_get_os_fd(ioq);
+}
+
 ///////////////////////////////////////////////////////////////////////////////
 /*
  * Endpoint Utilities


pjlib/include/pj/ioqueue.h
```diff
@@ -944,6 +944,17 @@ PJ_DECL(pj_status_t) pj_ioqueue_sendto( pj_ioqueue_key_t *key,
...
+ * Get the file descriptor associated with an ioqueue instance.
+ *
+ * @param ioqueue        The ioqueue instance.
+ *
+ * @return          The OS file descriptor associated with the instance.
+ *                  For epoll/kqueue this will be a value >= 0. For all
+ *                  other platforms, or if no handle is available, -1 will
+*                   be returned.
+ */
+PJ_DECL(int) pj_ioqueue_get_os_fd( pj_ioqueue_t *ioqueue );
```
sauwming: Why not use `pj_ioqueue_get_os_handle()` above?

IkerAriz: The downside of `pj_ioqueue_get_os_handle` is that it returns the opaque `pj_oshandle_t` type which can vary by platform, and it results in a more complex SWIG bindings. The `int` return type, on the other hand, allows client code to easily and portably check whether ioqueue fds are available and, if they are, to use the values as is (wherever standard fds are accepted).  The SWIG bindings are also straightforward.

IkerAriz: Here's a rough sketch of how I use the method introduced by the MR:
```
class MyEndpoint(pj.Endpoint):
    def handleEvents(self):
        ...

def main_loop(my_ep:MyEndpoint, my_thing:MyThing):
    ep_fd =  my_ep.libGetSipIoQueueFd()
    assert ep_fd >= 0, f'current platform not supported'
    thing_fd = my_thing.rfd
    while(True):
        r, _, _ = select.select([ep_fd, thing_fd], [], [], None)
        if ep_fd in r:
            my_ep.handleEvents()
        elif thing_fd in r:
            my_thing.handleEvents()
        ...
    ...
```

sauwming: We shouldn't add a duplicate API which does exactly the same thing with the existing one.
If you need to create an interface for `pj_oshandle_t`, you can do so in `pjsua2.i`.

IkerAriz: The new functions do indeed appear to do the same thing, but the return types of the new functions are considerably easier and safer to use for portable client code (especially scripting clients).  Consider the `main_loop` example above - suppose it's rewritten to use `pj_oshandle_t`:
```
import ctypes
c_int_p = ctypes.POINTER(c_int)

def main_loop(my_ep:MyEndpoint, my_thing:MyThing):
    ep_hnd =  my_ep.libGetSipIoQueueHandle()
    assert ???
    ep_fd = ctypes.cast(ep_hnd, c_int_p).contents.value 
    ...
``` 
To implement the `assert` one would have to write code to determine if the _opaque_ handle points to a 32-bit file descriptor on the current platform... this is doable but it involves both undermining the opaqueness of `pj_oshandle_t` and (faithfully) recreating information that `pj` already has cooked into it during compilation [1]. To finally convert the handle to an integer descriptor we also need language facilities to do that - these may not be available in all scripting languages and when they are the resulting code will typically be more delicate (eg, python). 

[1] The documentation for `pj_ioqueue_get_os_handle` says it returns non-null values on platforms where the queue is backed by epoll/kqueue _but also on other platforms_... which leaves callers unable to distinguish the two cases.

nanangizz: Perhaps we can skip the new PJLIB function `pj_ioqueue_get_os_fd()` to avoid API duplication and include only the new PJSUA2 functions: `libGetSipIoQueueFd() & libGetMediaIoQueueFd()` so it can be properly documented? These functions should handle the conversion from `pj_oshandle_t` to `int`.

Also few comments:
- For safety, add ioqueue type (compile-time) checks, so it only converts & returns fd `>=0` for epoll & kqueue.
- For consistency, consider renaming the functions to `libGetSipIoqueueFd() & libGetMediaIoqueueFd()`, as somewhere else in PJSUA2 already uses `Ioqueue` instead of `IoQueue`.

IkerAriz: Limiting the addition to the `Endpoint` class would indeed do the trick for clients of the SWIG bindings (myself included)... though it would leave C clients of `pjlib` in a potentially similar position [1]. If that's an issue, here's another possibility:
```
// given an ioqueue's handle returns its corresponding fd
PJ_DECL(int) pj_ioqueue_handle_to_fd( pj_oshandle_t );
```

[1] C clients might have easier access to the compiler defs that were used to build pjlib itself so maybe less of a problem for them.

nanangizz: Right, it sounds better to me. Note that there is also IO Completion Ports uses `HANDLE` or `void*` rather than `int`, so `libGetXyzIoqueueFd()` may not be portable, perhaps consider `libGetXyzIoqueueHandle()` + `utilIoqueueHandleToFd()`?

sauwming: You don't need to worry about `pjlib` clients. As compared to pjsua2 swig users, direct C/C++ users, including pjlib clients, have access to much more APIs and features. For example, they can easily check `pj_ioqueue_name()` to know which ioq backend they're using, so dealing with `pj_oshandle_t` should be no problem at all.

wosrediinanatour:
A general question: does it reallymake sense to use endpoint instead an own type?

pjlib/include/pj/ioqueue.h
IkerAriz: > Right, it sounds better to me. Note that there is also IO Completion Ports uses `HANDLE` or `void*` rather than `int`, so `libGetXyzIoqueueFd()` may not be portable, perhaps consider `libGetXyzIoqueueHandle()` + `utilIoqueueHandleToFd()`?

That would work as well, though it does involve somewhat hairier SWIG bindings (the `pj_oshandle_t` would need to be exposed to scripting clients).

IkerAriz: > You don't need to worry about `pjlib` clients. As compared to pjsua2 swig users, direct C/C++ users, including pjlib clients, have access to much more APIs and features. For example, they can easily check `pj_ioqueue_name()` to know which ioq backend they're using, so dealing with `pj_oshandle_t` should be no problem at all.

One downside of calling `pj_ioqueue_name` to explicitly check for backends with file descriptors is that it requires client code to maintain this mapping and update it as backend support changes. An API like `pj_ioqueue_handle_to_fd` where each backend directly signals its file descriptor support allows for portable client code that is relatively oblivious to the underlying backend (as long as it supports fds).

nanangizz: > That would work as well, though it does involve somewhat hairier SWIG bindings (the `pj_oshandle_t` would need to be exposed to scripting clients).

From the library's perspective, I think it's cleaner to provide a converter from the portable type (`pj_oshandle_t`) to the OS-specific type (e.g: `int` or `fd`), rather than duplicating functions for each platform or type?

IkerAriz: > > That would work as well, though it does involve somewhat hairier SWIG bindings (the `pj_oshandle_t` would need to be exposed to scripting clients).
> 
> From the library's perspective, I think it's cleaner to provide a converter from the portable type (`pj_oshandle_t`) to the OS-specific type (e.g: `int` or `fd`), rather than duplicating functions for each platform or type?

A single converter function would involve some duplication as well, although of a different type. For example, a function like the following duplicates the conditional compilation checks present in each of the `ioqueue_*` family of files. 

```
int utilIoqueueHandleToFd(pj_oshandle_t h) 
{
    #if PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_KQUEUE
        ...
    #elif PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_EPOLL
        ...
    #else
        ...
    #endif
}
```

Also note that this function wouldn't always convert from the portable `pjlib` type to an OS-specific type (eg, on Windows the OS-specific handle type is `HANDLE`). It instead converts from one portable type to another (`pjlib` handle to POSIX file descriptor).

nanangizz: > A single converter function would involve some duplication as well, although of a different type. For example, a function like the following duplicates the conditional compilation checks present in each of the `ioqueue_*` family of files.
> 
> ```
> int utilIoqueueHandleToFd(pj_oshandle_t h) 
> {
>     #if PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_KQUEUE
>         ...
>     #elif PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_EPOLL
>         ...
>     #else
>         ...
>     #endif
> }
> ```

Not really, the conversion function can be implemented once only: in PJSUA2. IMO, the conversion in `pjlib` is not so urgent (see the reason below), and if it is to be provided, it can be put in `ioqueue.h` only. Imagine the mess if every `ioqueue_*` file had to include `pj_ioqueue_get_os_fd()` & potentially some `pj_ioqueue_get_os_*()` variants.


> Also note that this function wouldn't always convert from the portable `pjlib` type to an OS-specific type (eg, on Windows the OS-specific handle type is `HANDLE`). It instead converts from one portable type to another (`pjlib` handle to POSIX file descriptor).

You're right, but note that `pjlib` doesn't base its abstraction on POSIX, in other words, from `pjlib` perspective, POSIX is simply one of the platforms to abstract.

Note that any app needing to retrieve the ioqueue handle should already be platform-aware (shouldn't it?), so IMO the conversion and any other platform-specific code are better handled within the app as much as possible, allowing the lib remain less platform-specific. Also as @sauwming previously described, C/C++ users should be capable of performing the conversion without a helper function, hence the conversion in `pjlib` is not so urgent. And in PJSUA2, the approaches are:
1.  `libGetSipIoqueueFd() + libGetMediaIoqueueFd()`, `libGetSipIoqueueWinHandle() + libGetMediaIoqueueWinHandle()`, so every additional ioqueue, we need to add variants as much as the supported platforms.
2.  `libGetSipIoqueueHandle() + libGetMediaIoqueueHandle()` & conversion helper functions, so for every additional ioqueue we only need to add one conversion function variant.
3. ...?

And so far, I think I'd vote for approach no 2, as IMO it is cleaner in the long run.

IkerAriz: Sorry for delayed reply (was away for a few days). 

Fleshing out approach 2 a bit more and assuming (for now) that we have to worry about only two native platform handle types - POSIX file descriptors and Windows handles - here's a potential implementation of the conversion helpers:
```
// approach 2
// pjsip/src/pjsua2/endpoint.cpp (or `ioqueue.h` if C support were needed)

#if PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_EPOLL
    int utilIoqueueHandleToFd(pj_oshandle_t h) { return h ? * h : -1; }
#elif PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_KQUEUE
    int utilIoqueueHandleToFd(pj_oshandle_t h) { return h ? * h : -1; }
#elif PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_IOCP
    int utilIoqueueHandleToFd(pj_oshandle_t h) { return -1; }
    HANDLE utilIoqueueHandleToWinHandle(pj_oshandle_t h) { h ? *h : INVALID_HANDLE_VALUE; }
#else
    ...
#endif
```
Backend authors just need to be careful to ensure this code is kept in sync whenever backends are added or updated.

IkerAriz: > Not really, the conversion function can be implemented once only: in PJSUA2.

Note that the duplication I was referring to is with respect to the conditional compilation check. For example, at the top of `ioqueue_epoll.c` we have this:
```
#if PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_EPOLL
   // main details of epoll backend
#endif
``` 
and in `endpoint.cpp` we would have:
```
#if PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_EPOLL
    // handle conversion details for epoll backend
#endif
``` 
> Imagine the mess if every ioqueue_* file had to include pj_ioqueue_get_os_fd() & potentially some pj_ioqueue_get_os_*() variants.

I suppose in practice the result would be that every backend implements `pj_ioqueue_get_os_fd` and only the Windows/IOCP backend implements (eg) `pj_ioqueue_get_os_winhnd` (see example above). However, this does give a special preference to the `fd` flavor which... doesn't feel right either.

IkerAriz: > Note that any app needing to retrieve the ioqueue handle should already be platform-aware (shouldn't it?)

You're right. Though the question is what degree of platform independence an app needs/wants. For apps only interested in `ioqueue` backends that work with fd-friendly selectors (epoll,kqueue,select,etc) an API like `libGetSipIoQueueFd` is all that's needed:
```
# the following works or gives a meaningful error otherwise 
ep_fd =  my_ep.libGetSipIoQueueFd()
assert ep_fd >= 0, f'current platform not supported'
```
But as noted above, this gives POSIX file descriptors preferential treatment over other handle types.

IkerAriz: To sidestep the issue of converting `pj_oshandle_t` to the various types of platform-specific handles I removed the interfaces added previously (`pj_ioqueue_get_os_fd` and associated `Endpoint` methods) and went with the following:

1. Add the proposed `Endpoint::libGetSipIoQueueHandle` that returns `pj_oshandle_t`.
2. Add a type map to `pjsua2.i` that maps `pj_oshandle_t` to an integral type that works uniformly across binding languages. 

Here's some sample usage in python:
```
plat = platform.system().lower()
assert plat in ('linux', 'darwin'), f'platform {plat} not supported'

ep_cfg = pj.EpConfig()
ep = pj.Endpoint()
ep.libCreate()
ep.libInit(ep_cfg) 
h = ep.libGetSipIoqueueHandle()
# interpret integral handle as pointer to integer
h_fd = ctypes.cast(h, ctypes.POINTER(ctypes.c_int)).contents.value
print(f'h: {h} | val: {h_fd}')
```

IkerAriz: Note that the python snippet assumes the backends on MacOs and Linux are always fd-based... which seems reasonable at present. But if that were to change the code above would have to instead check for specific ioqueue backend names.

pjsip/include/pjsua2/endpoint.hpp
```diff
@@ -1545,6 +1545,12 @@ class Endpoint
      */
     void libDestroy(unsigned prmFlags=0) PJSUA2_THROW(Error);
 
+    /**
+     * Get the ioqueue handle associated with pjsua's SIP IO queue, if any.
+     * Returns an integer >= 0 if the IO queue is implemented with epoll or
+     * kqueue, otherwise returns -1.
+     */
+    pj_oshandle_t libGetSipIoqueueHandle();
```
sauwming, line 1556: Missing the one for the media?

Also need to fix the doc since the return value is no longer integer. Check the formatting as well (such as use `@return`).

IkerAriz: Fixed documentation for `libGetSipIoqueueHandle`. 

Re. media. My original motivation for adding the media handle method was to support apps that used only one thread for both the app and `pjsua2` - driven entirely by the app's top-level event loop with calls to `libHandleEvents` as needed. But in the end I couldn't find a way to configure `pjsua2` to keep it from creating media threads... and without that it wasn't clear a `libGetMediaIoqueueHandle` would still add value.

sauwming (dismissed):

pjsip-apps/src/swig/pjsua2.i
```diff
@@ -29,6 +29,11 @@ using namespace pj;
 %template(ByteVector)            std::vector<unsigned char>;
 %template(StringToStringMap)     std::map<string, string>;
 
+// simplifies handling of void* pointer across
+// all language bindings
+%typemap(out) pj_oshandle_t {
+  $result = SWIG_From_unsigned_SS_long_SS_long(reinterpret_cast<uintptr_t>($1));
+}
```
nanangizz: Please note that `void *` has been remapped, to `long long` for Java or `IntPtr` for CSharp), which work fine so far:

https://github.com/pjsip/pjproject/blob/2776fe3ecec33a33f0630bf818e274fffe26698b/pjsip-apps/src/swig/pjsua2.i#L81-L96

https://github.com/pjsip/pjproject/blob/2776fe3ecec33a33f0630bf818e274fffe26698b/pjsip-apps/src/swig/pjsua2.i#L126-L127

So to avoid any possible risk to those languages, I think the map is better be specific for Python (i.e: protect with `#ifdef SWIGPYTHON`). Also optionally, add an additional helper function (will be undocumented though):
```
#ifdef SWIGPYTHON

  %typemap(out) pj_oshandle_t {
    $result = SWIG_From_unsigned_SS_long_SS_long(reinterpret_cast<uintptr_t>($1));
  }

  // Optionally also add helper function, no docs though
  #if PJ_IOQUEUE_IMP==PJ_IOQUEUE_IMP_SELECT || PJ_IOQUEUE_IMP==PJ_IOQUEUE_IMP_KQUEUE
    %extend pj::Endpoint {
      int libGetSipIoqueueFd(void) {
        pj_oshandle_t h = self->libGetSipIoqueueHandle();
        return (h? *(int*)h : -1);
      }
    }
  #endif /* PJ_IOQUEUE_IMP */

#endif
```

What do you think?

IkerAriz: So far we've discussed two ways to support integration with application event loops:

1. Expose raw handle, leave portable handle conversion to application.
2. Expose raw handle, supply helpers for portable handle conversion.

The latest commit adopts option 1 (which I think is what @sauwming had originally suggested). This may not be as convenient for apps but it makes for a small change and it has the virtue of neatly sidestepping the issues that arise with option 2 (providing helpers for every platform type, where to best maintain the helpers, should the helpers be limited to higher level languages, etc). Option 1 is also compatible with option 2 so the latter can always be pursued later if needed.

Re. the `pj_oshandle_t` type map. IIUC, SWIG's handling of `typedef`s (described [here](https://www.swig.org/Doc1.3/Typemaps.html#Typemaps_nn18)) should ensure that the `pj_oshandle_t` type map only applies to that type without affecting the existing handling of `void*`.

PR #4409, opened 2025-04-28
Fixed buffer overflow in Android video codec

sauwming:
To fix #4404.

As reported in https://github.com/pjsip/pjproject/issues/4404, PJSIP will crash when using Android video mediacodec and remote sends a video with resolution higher than supported.

Although the issue was caused by the remote sending video with resolution higher than signalled, PJSIP shouldn't crash. The crash is caused due to lack of checking of the frame buffer size that's supposed to hold the decoded frame.

PR #4600, opened 2025-09-19
fix and_vid_mediacodec.cpp bug 

Haha003:
fix write_yuv() when invoke pj_memcpy() sometimes  will overflow


sauwming:
Should have been fixed by #4409. Please let us know if this is not the case.

Haha003:
yes,it is, but the before fixed had containted on write_yuv() ,so add another one in it

wxiaoguang:
I think these unrelated changes (comments) should be cleaned up first?

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -1316,11 +1326,19 @@ static int write_yuv(pj_uint8_t *buf,
...
     req_size = (iWidth * iHeight) + (iWidth / 2 * iHeight / 2) +
                (iWidth / 2 * iHeight / 2);
-    if (dst_len < req_size)
-        return -1;
 
+    // PJ_LOG(4,(THIS_FILE, "write_yuv dst_len:[%d]，req_size:[%d],input_len:[%d]",
+        // dst_len,req_size,input_len));
+
+    
+    if (dst_len < req_size||input_len<req_size)
```
sauwming, line 1327: I don't quite understand though. Under what circumstances would `input_len` (which is `output_size`) be smaller than the size of the decoded image?

`AMediaCodec_getOutputBuffer(..., &output_size)` is supposed to return buffer of the correct size, or NULL if it fails to do so.
If it somehow can return buffer of incorrect size then perhaps the check should be put together with `if (output_buf == NULL)` (so it should be `if (output_buf == NULL || output_size < the_correct_decoded_image_size)` )

Haha003: actually,i don't know the exact reason either，in my case, it may happened when the video size adjusted according to  remote capability.

Haha003: i saw the note at document 
图片

Haha003:
> I think these unrelated changes (comments) should be cleaned up first?

ok😊

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
wxiaoguang: ( New to the code base, just interesting in learning it 😄 )

If I understand correctly, the `output` is from argument `and_media_decode( ... pjmedia_frame *output )`, and only `output->size` is used for overflow check.

However, the media data format might have changed `if (and_media_data->format_changed)`

So `AMediaCodec_getOutputBuffer` might return a buffer with different size (never checked) which doesn't match the old `pjmedia_frame *output`?


----

And I think it is always good to check the buffer size to ensure there is no overflow, instead of assuming that it won't overflow.

sauwming: I suppose there may be a race between decode and `and_med_on_format_changed()`.

So we need to make sure that `output_size` is exactly the same as the expected image size, i.e.:
```
    dec_size = and_media_data->prm->dec_fmt.det.vid.size.w *
               and_media_data->prm->dec_fmt.det.vid.size.h * 3 / 2;
    if (output_size == dec_size && output->size >= dec_size) {
        len = write_yuv((pj_uint8_t *)output->buf,
...
```

sauwming: Let me know whether the above fix could work

wxiaoguang: It looks right (I have no real environment to test)

But `dec_size` looks duplicate with the `req_size` in `write_yuv`

Maybe it's better to move all the checks into `write_yuv`, check all together at the same time.

```c
    if (output_buf == NULL) {
        am_status = AMediaCodec_releaseOutputBuffer(and_media_data->dec,
                                        buf_info.index, 0);
        PJ_LOG(4,(THIS_FILE, "Decoder getOutputBuffer failed"));
        return status;
    }

    // CHANGE: no need to check, just call write_yuv with the buffers and their sizes.

    len = write_yuv((pj_uint8_t *)output->buf,
                output->size,
                output_buf,
                output_size,
                and_media_data->dec_stride_len,
                and_media_data->prm->dec_fmt.det.vid.size.w,
                and_media_data->prm->dec_fmt.det.vid.size.h);   

...

static int write_yuv(pj_uint8_t *buf,
                     unsigned dst_len,
                     unsigned char* input,
                     int input_len,
                     int stride_len,
                     int iWidth,
                     int iHeight) {

    // or also simplify it to iWidth * iHeight * 3 / 2
    req_size = (iWidth * iHeight) + (iWidth / 2 * iHeight / 2) +
               (iWidth / 2 * iHeight / 2);

    // CHANGE: check all the sizes before use the buffers
    if (dst_len < req_size || input_len < req_size) {
        return -1;
    }
}
```

sauwming: okay

wxiaoguang: It seems that the check here is not right. "stride len" should be used for checking "input size"

-> Fix android video codec overflow #4617

sauwming:
Continued in #4617

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
Haha003: ok😊

Haha003:
> Continued in #4617

ok😊

wxiaoguang:
> > Continued in #4617
> 
> ok😊

@Haha003  "Fix android video codec overflow #4617" has been merged, can you try it to see whether it works for you? Thank you.

Haha003:
> > > Continued in #4617
> > 
> > 
> > ok😊
> 
> @Haha003 "Fix android video codec overflow #4617" has been merged, can you try it to see whether it works for you? Thank you.

yes ,it is work ,sorry for so late reply,and thank you for your work😊

PR #4602, opened 2025-09-21
Modify pjmedia_conf_enum_ports()'s ports parameter to be optional

sauwming:
Based on #4585.
For efficiency, allow the ports parameter in the API `pjmedia_conf_enum_ports()` to be optional.

Thanks to @PesalaDeSilva for the original patch.


nanangizz:
`pjmedia_conf_get_port_count()` offers a more efficient approach to get the port count as no ports iteration is involved?

sauwming:
Ah right, the naming and implementation of `pjsua_conf_get_active_ports()` gave me the impression that they would give two different numbers (i.e. all ports and only the active ones).

PR #4587, opened 2025-09-09
Correctly escaping IPv6 address used in contact URI

j4nnis:
If an IPv6 address is used in a contact URI it must be wrapped in square brackets. Otherwise the contact URI pjua builds is invalid. A validly escaped IPv6 address looks like this "[2001:0db8:0001:0000:0000:0ab9:C0A8:0102]:1234" where the port is added after the closing bracket. 
The existing code did not check if the address that was wrapped was an IPv6 address correctly so that during testing in an. IPv6/NAT64 network the IPv6 address was not correctly escaped. 
This resulted in pjsua being unable to register and make calls.

I also created an issue before this PR: #4586 This PR does solve that issue for me and seems to not cause issues.

PR #4609, opened 2025-09-25
Fixed SIP TLS to update local address only if IP version is correct

sauwming:
Re #4587,

Even though the SIP transport TLS is configured to be IPv4, Apple SSL backend may return IPv6 local address in a NAT64 network.

### DIFF
--- a/pjsip/src/pjsip/sip_transport_tls.c
+++ b/pjsip/src/pjsip/sip_transport_tls.c
@@ -1955,7 +1955,14 @@ static pj_bool_t on_connect_complete(pj_ssl_sock_t *ssock,
     tp_addr = &tls->base.local_addr;
     pj_sockaddr_cp((pj_sockaddr_t*)&addr, 
                    (pj_sockaddr_t*)&ssl_info.local_addr);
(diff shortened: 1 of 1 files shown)


nanangizz:
Can this approach also be applied to the SSL socket?

sauwming:
During my test, the SSL socket is initialized with `0.0.0.0`.

nanangizz:
So in the SSL socket, it sees `0.0.0.0` initially, then IPv6 address after connected?

IMHO ideally, the fix/workaround should be applied at the source of the problem, i.e: the SSL socket, rather than at each of the SSL socket apps.

sauwming:
> So in the SSL socket, it sees `0.0.0.0` initially, then IPv6 address after connected?

Yes.

> IMHO ideally, the fix/workaround should be applied at the source of the problem, i.e: the SSL socket, rather than at each of the SSL socket apps.

That was my initial attempt as well. I tried `pj_sockaddr_synthesize()` to convert the IPv6 address to IPv4, but it failed (note that `IN6_IS_ADDR_V4MAPPED(the_ipv6_local_address)` returns 0).

PR #4614, opened 2025-10-03
Make sure port and port range settings are applied when bound address is not set on STUN config

trengginas:
The https://github.com/pjsip/pjproject/pull/4552 PR changed the check on when to apply the bound address config.
Due to the patch, the `port` setting might not get applied when the `bound_address` is not set.

This PR will check if the bound_address has a port set to make sure that it is still applied.


nanangizz (approved):

sauwming (approved):

hrc-amilcar:
@trengginas
When `AF_INET6` is zero, this seems like it will still copy a blank address (with a valid port)
Is that the intent?

amubiera:
@trengginas
When AF_INET6 is zero, this seems like it will still copy a blank address (with a valid port)
Is that the intent?

If so, is this equivalent?
```C
if (pj_sockaddr_has_addr(&cfg->bound_addr.addr) ||
    pj_sockaddr_get_port(&cfg->bound_addr) != 0)
    pj_sockaddr_cp(&bound_addr, &cfg->bound_addr);
```

PR #4617, opened 2025-10-07
Fix android video codec overflow

wxiaoguang:
`write_yuv` should check both output buffer size and input buffer size, to avoid buffer overflow.

By pre-checking the buffer sizes, the `write_yuv` loops can also be simplified.


wxiaoguang:
It's easier to review by "split" diff: https://github.com/pjsip/pjproject/pull/4617/files?diff=split&w=0

Ref:

* #4600
* #4409

PR #4618, opened 2025-10-07
Add manual host candidates for ICE

nanangizz:
This is useful in systems that have both internal and public IP addresses and aim to use ICE without STUN, thereby simplifying the overall architecture.

This feature can be used via `pjsua_ice_config.ice_manual_host`:
```
    /**
     * Optional configuration for manually specifying host candidates.
     * Useful for applications that need to assign a specific IP address
     * as the host candidate. The candidate will use the same port as
     * the automatic/base host candidate.
     *
     * Note that this setting will be used only when STUN server is
     * not configured.
     */
    pj_sockaddr         ice_manual_host[PJ_ICE_MAX_STUN];
```

Thanks Hyun Chul Kim for the feedback.

### DIFF
--- a/pjnath/include/pjnath/ice_strans.h
+++ b/pjnath/include/pjnath/ice_strans.h
@@ -248,6 +248,17 @@ typedef struct pj_ice_strans_stun_cfg
     unsigned             max_host_cands;
 
     /**
+     * Optional configuration for manually specifying host candidates.
+     * Useful for applications that need to assign a specific IP address
+     * as the host candidate. The candidate will use the same port as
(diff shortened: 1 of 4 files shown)


hkim06:
Could you also update AccountNatConfig in pjsip/include/pjsua2/account.hpp?

nanangizz:
> Could you also update AccountNatConfig in pjsip/include/pjsua2/account.hpp?

Done, but not tested yet (Github CI may not test this either). Note that I also added field `manual_host_cnt`, changed the array size, and removed the requirement of disabling STUN.

PR #4622, opened 2025-10-08
Fix SDP negotiation for audio with asymmetric encoding parameters

### LINKED ISSUE #4615: SDP Negotiation failed due to optional encoding parameter in rtpmap attribute
### Describe the bug

Assume that the SDP offer contains `a=rtpmap:118 AMR/8000/1` whereas the answer contains `a=rtpmap:118 AMR/8000 `. Then the SDP negotiation fails.
 
Note the missing `/1` in the answer. This is valid, because a missing encoding parameter is equal to having 1, cf.  RFC 4566:
> For audio streams, <encoding parameters> indicates the number  of audio channels.  This parameter is OPTIONAL and may be omitted if the number of channels is one, provided that no additional parameters are needed.

### Steps to reproduce

Any SDP offer and answer, where in one a rtpmap attribute has a missing encoding parameter, whereas the other has coding parameter `1`.

### PJSIP version

2.15.1

### Context

* This can easily happen, if you have a 3rd-party media.
* Commit https://github.com/pjsip/pjproject/commit/b180dcab7469eb4b4e382a059fa64cce602463af introduced the whole fmtp matching, while commit https://github.com/pjsip/pjproject/ commit/1d9b9a4996cae7eeb0f32e97908931945928438b shows the change that have to be improved.

### Log, call stack, etc

```shell
Line of code which reject the codec: https://github.com/pjsip/pjproject/commit/1d9b9a4996cae7eeb0f32e97908931945928438b#diff-3981710a639e7fa1e83963a5dac4cb6a05cffb2da2179a4bd891c11f790de3beR733
```

nanangizz:
To fix #4615.

Currently the library assumes to always send offer without encoding param when channel count is 1. But alas, app may modify the SDP or 3rd party media may creates its own SDP.

Thanks @wosrediinanatour for the report.

### DIFF
--- a/pjmedia/src/pjmedia/sdp_neg.c
+++ b/pjmedia/src/pjmedia/sdp_neg.c
@@ -778,6 +778,7 @@ static pj_status_t process_m_answer( pj_pool_t *pool,
                                      pj_bool_t allow_asym)
 {
     unsigned i;
+    pj_bool_t is_audio;
 
     /* Check that the media type match our offer. */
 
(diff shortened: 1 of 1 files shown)


wosrediinanatour:
Used it in my use case. Worked! Thank you!

PR #4624, opened 2025-10-09
Fix build fail for Android due to incorrect AMediaCodec detection

trengginas:
Since #4105, android media codec (`AMediaCodec`) is using async method which requires `android-28`.
On #4364, on NDK >= r21 the APP_PLATFORM is picked by the minimum SDK version available. e.g.: on ndk r27 the minimum is `android-21` and it will fail the build:
```
      | 
```
This is because `AMediaCodec` will be enabled by checking an API (`AMediaCodec_createDecoderByType()`) available since `android-21`. 

This patch will detect for `AMediaCodec_setAsyncNotifyCallback()` instead when enabling `AMediaCodec` to avoid the error.

Note:
To enable `AMediaCodec`, you need to specify `APP_PLATFORM` to minimal `android-28`.
e.g.: 
```
APP_PLATFORM=android-28 ./configure-android
```

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -678,6 +678,7 @@ openssl_h_present
 ac_ssl_backend
 ac_ssl_has_aes_gcm
 ac_no_ssl
+ac_mediacodec_cflags
 ac_no_mediacodec
 ac_vpx_ldflags
 ac_vpx_cflags
@@ -9581,6 +9582,7 @@ fi
 
 
 
+
 # Check whether --enable-android-mediacodec was given.
 if test ${enable_android_mediacodec+y}
 then :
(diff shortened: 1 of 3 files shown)


nanangizz:
From discussion in #4367, there seem to be a better approach, i.e: use a run-time check instead of compile-time check and also avoid registering the codecs when android sdk < 28, not sure if this has been investigated and encountering an issue?

trengginas:
Tested the latest modification with Android NDK 27 (default using android-21), built successfully, and tested the app on android device 28++ codec initiated/run as expected. 
Tested on an older Android device (< android-28), the codec will not be registered.

PR #4630, opened 2025-10-12
Fix ffmpeg with libopenh264 encoder configuration parameter incompatibility(#4629)

### LINKED ISSUE #4629: PJSIP has incomplete support for FFmpeg with libopenh264: Incompatible profile type and NAL slice size configuration
### Describe the bug

PJSIP exhibits inconsistent configuration parameter handling between x264 and libopenh264 encoders when using FFmpeg:

Mismatched profile parameter type: x264 uses a string type, while libopenh264 requires an integer type.

Different NAL maximum slice size parameter names: x264 uses slice-max-size, while libopenh264 uses max_nal_size.

This prevents correct parameter configuration when using the libopenh264 encoder, impacting encoding quality and performance.

### Steps to reproduce

make video call.

### PJSIP version

2.15.1

### Context

windows
ffmpeg4.4.4 with libopenh264

### Log, call stack, etc

```shell
pjmeida_codec: ffmepg_vid_codecs.c: 537
```

MaybeNullPtr:
To fix #4629.

### DIFF
--- a/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
+++ b/pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
@@ -558,6 +558,18 @@ static pj_status_t h264_preopen(ffmpeg_private *ff)
                        ff->param->enc_mtu));
         }
 
+        /* openh264 settings */
+        if (!AV_OPT_SET_INT(ctx->priv_data, "profile", ctx->profile))
+        {
+            PJ_LOG(3, (THIS_FILE, "Failed to set openh264 H264 profile to '%d'",
+                ctx->profile));
+        }
+        if (!AV_OPT_SET_INT(ctx->priv_data, "max_nal_size", ff->param->enc_mtu))
+        {
+            PJ_LOG(3, (THIS_FILE, "Failed to set openh264 H264 max NAL size to %d",
+                ff->param->enc_mtu));
+        }
+
         /* Apply intra-refresh */
         if (!AV_OPT_SET_INT(ctx->priv_data, "intra-refresh", 1))
         {


pjmedia/src/pjmedia-codec/ffmpeg_vid_codecs.c
```diff
@@ -558,6 +558,18 @@ static pj_status_t h264_preopen(ffmpeg_private *ff)
                        ff->param->enc_mtu));
         }
 
+        /* openh264 settings */
+        if (!AV_OPT_SET_INT(ctx->priv_data, "profile", ctx->profile))
+        {
+            PJ_LOG(3, (THIS_FILE, "Failed to set openh264 H264 profile to '%d'",
+                ctx->profile));
+        }
+        if (!AV_OPT_SET_INT(ctx->priv_data, "max_nal_size", ff->param->enc_mtu))
```
sauwming, line 567: won't it be better if this is only done if setting "slice-max-size" fails? i.e.:
```
if (!AV_OPT_SET_INT(ctx->priv_data, "slice-max-size", ff->param->enc_mtu))
{
    if (!AV_OPT_SET_INT(ctx->priv_data, "max_nal_size", ff->param->enc_mtu))
    {
          PJ_LOG();
    }
}
```

And also the same for profile.

MaybeNullPtr:
I came across this in the source code. Does it mean the project team does not intend to support FFmpeg with OpenH264? My request for support might be unnecessary. The relevant code is located near line 657 in pjmedia_codec config.h, as shown below:

```
/**
 * Enable FFMPEG H264 codec (requires libx264).
 *
 * Default: disabled when OpenH264 is used, otherwise it is set to
 * PJMEDIA_HAS_FFMPEG_VID_CODEC
 */
#ifndef PJMEDIA_HAS_FFMPEG_CODEC_H264
#   if defined(PJMEDIA_HAS_OPENH264_CODEC) && PJMEDIA_HAS_OPENH264_CODEC != 0
#       define PJMEDIA_HAS_FFMPEG_CODEC_H264    0
#   else
#       define PJMEDIA_HAS_FFMPEG_CODEC_H264    PJMEDIA_HAS_FFMPEG_VID_CODEC
#   endif
#endif
```

sauwming:
`config.h` is the default/recommended setting, so yes, by default, only one H264 backend should be enabled since we usually only need one. So if you can work with only one, then true, the PR will be unnecessary and you can close it. But if you have custom requirements that require both and you can make it work without conflict, feel free to proceed.

Issue #4634, opened 2025-10-13
Build error with Xcode 26


saghul:
### Describe the bug

I'm getting the following build error with tip-of-tree: 64177d1aa

Since `make_unique` is part of C++ > 11, maybe the minimum should be bumped to 17?

### Steps to reproduce

Just run `make` with Xcode 26 configured as the compiler with `xcode-select`.

### PJSIP version

current master 64177d1aa

### Context

macOS 15, Xcode 26.

### Log, call stack, etc

```shell
[...]

PR #4638, opened 2025-10-14
Pj file io

LeonidGoltsblat:
This PR contains low-level support for asynchronous file reading (a high-level implementation is presented in ioqueue in the next PR).
IOQueue must differentiate between file backend implementations, so config.h and config_site.h now support the PJ_FILE_IO macro with two available values: PJ_FILE_IO_WIN32 and PJ_FILE_IO_ANSI.
Thus, both file_io_ansi.c and file_io_win32.c can be selected for Windows builds (for debugging), but only ANSI is supported for other platforms.

Both backends support the new pj_file_size_by_handle() API.

To support truly asynchronous (overlapped) I/O and optimize file caching strategies, new flags have been added to the pj_file_access enumeration. Currently, all these flags are ignored on all platforms except Windows, but can be safely used by applications on any platform.

Fixed a number of bugs in the Win32 file I/O backend, the main one being that pj_file_open() incorrectly treats dwCreationDisposition as a mask rather than as one of the specified values.

### DIFF
--- a/pjlib/src/pj/file_io_ansi.c
+++ b/pjlib/src/pj/file_io_ansi.c
@@ -28,7 +28,9 @@
 #if defined(PJ_HAS_SYS_TYPES_H) && PJ_HAS_SYS_TYPES_H != 0
 #include <sys/types.h>
 #endif
+#if defined(PJ_HAS_SYS_STAT_H) && PJ_HAS_SYS_STAT_H != 0
 #include <sys/stat.h>
+#endif
 
 #if PJ_FILE_IO == PJ_FILE_IO_ANSI
 


pjlib/src/pj/file_io_ansi.c
```diff
@@ -182,3 +201,53 @@ PJ_DEF(pj_status_t) pj_file_flush(pj_oshandle_t fd)
...
+
+#else
+
+#if    (defined(PJ_HAS_UNISTD_H) && PJ_HAS_UNISTD_H != 0)
+#   define _fileno(f) fileno(f)
+#endif
+    struct stat buf;
+    int fd, result;
+
+    if ((fd = _fileno((FILE*)fh)) == -1)
```
sauwming: Minor: The usage of `else` here is not needed and makes the code unnecessarily more difficult to read. And cramping an assignment statement with error checking in a single `if` makes it even harder.

So consider this instead (this is much easier to read and understand IMO, with more and less the same number of LOC).
```
fd = _fileno((FILE*)fh);
if (fd == -1)
    return -1;

result = fstat(fd, &buf);
if (result != 0)
    return -1;

return buf.st_size;
```

sauwming (approved):

pjlib/src/pj/file_io_ansi.c
LeonidGoltsblat: @sauwming! More serious question: stat.h and fstat() are not part of ANSI C. Can we assume they always exist and remove the #if 0 part that bothers the AI ​​so much?

sauwming: I think you can ignore AI's feedback about comments/dead code deletion if there is justification (such as for legacy code archiving).

For the sys/stat, perhaps better to add its detection in aconfigure.ac?

nanangizz:
Could you clarify how this feature will be used in the library (e.g: by the pjmedia WAV/AVI player or recorder)? Note that pjlib does not need to abstract most OS functionalities, only those that are actually used by the library.

LeonidGoltsblat:
I'm trying to gradually move toward a wav_player PR that supports:
- the concept of a data source (DS) - an abstraction for the data being played (file, database, etc.)
- asynchronous proactive data reading (proactive is performed on the player side, and asynchronous or not on the DS side)

(The same ideas can be used for AVI and recording devices, but are currently only implemented for wav_player.)

DS can use any mechanism for asynchronous I/O support, and may or may not use pjlib ioqueue. However, for asynchronous file reading, I implemented the ioqueue extension and am using it in the wav_player mediaport.

I try to keep PRs small, but each one tends to become large...
The next PR should include ioqueue, then wav_player, and then optimized udp_transport. I hope the verification process won't take as long as the parallel bridge did, but there's every chance... but there's no way to dedicate your whole life to it.

PS #4626 is for the same PR chain

pjlib/src/pj/file_io_ansi.c
LeonidGoltsblat: Since CL shows that the code compiles with all used compilers, does this mean the assumption is correct?
Detection is a good idea, but I'm not ready to implement it at this point. (See @nanangizz 's answer below.)

sauwming: The CI doesn't represent all compilers, it typically only uses the latest/recent version, while our users use all kinds of machines and compilers, so a build failure is possible. For example, a recent build failure: https://github.com/pjsip/pjproject/issues/4634.

LeonidGoltsblat: > The usage of else here is not needed

else removed

nanangizz (approved):

pjlib/src/pj/file_io_ansi.c
sauwming: The detection seems to be needed, since some OS, such as FreeRTOS doesn't inherently come with sys/stat.h.

LeonidGoltsblat: The PJ_HAS_SYS_STAT_H macro was manually added to two os_*.h files and to file_io_ansi.c (Al should be happy now!), but not to the autoconfiguration scripts.
Let someone who plans to use ansi file io add that to aconfigure.ac.

Note 1: Perhaps PJ_HAS_SYS_TYPES_H has the same meaning, and PJ_HAS_SYS_STAT_H should be equal to PJ_HAS_SYS_TYPES_H.

Note 2: The entire contents of file_access_unistd.c are cross-platform, except for the line #include <unistd.h>. There is no need for separate implementations.

PR #4640, opened 2025-10-17
pjsua_media: Exclude usage of potentially uninitialized mvididx array when video is disabled.

amubiera:
Fixes warning (as errors) with GCC compilers:
* gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
* gcc (Ubuntu 14.2.0-19ubuntu2) 14.2.0
* Other compilers (like clang and Visual Studio) do not report this error.

```C
In function ‘pj_memchr’,
    inlined from ‘pjsua_media_channel_init’ at pjsip/pjproject/pjsip/src/pjsua-lib/pjsua_media.c:2694:20:
pjsip/pjproject/pjlib/include/pj/string.h:897:12: error: ‘mvididx’ may be used uninitialized [-Werror=maybe-uninitialized]
  897 |     return (void*)memchr((void*)buf, c, size);

In function ‘pj_memchr’,
    inlined from ‘pjsua_media_channel_init’ at pjsip/pjproject/pjsip/src/pjsua-lib/pjsua_media.c:2697:17:
pjsip/pjproject/pjlib/include/pj/string.h:897:12: error: ‘mvididx’ may be used uninitialized [-Werror=maybe-uninitialized]
  897 |     return (void*)memchr((void*)buf, c, size);
```

After some bisecting, it was introduced indirectly by #4585 seemingly because the `pjsua_var.calls` array is no longer pre-allocated-- likely confusing the GCC optimizer.

Let me know if this is the correct fix.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -2427,13 +2427,13 @@ pj_status_t pjsua_media_channel_init(pjsua_call_id call_id,
     const pj_str_t STR_TEXT  = { "text", 4 };
     pjsua_call *call = &pjsua_var.calls[call_id];
     pjsua_acc *acc = &pjsua_var.acc[call->acc_id];
-    pj_uint8_t maudidx[PJSUA_MAX_CALL_MEDIA];
+    pj_uint8_t maudidx[PJSUA_MAX_CALL_MEDIA] = {0};
     unsigned maudcnt = PJ_ARRAY_SIZE(maudidx);
     unsigned mtotaudcnt = PJ_ARRAY_SIZE(maudidx);
-    pj_uint8_t mvididx[PJSUA_MAX_CALL_MEDIA];
+    pj_uint8_t mvididx[PJSUA_MAX_CALL_MEDIA] = {0};
     unsigned mvidcnt = PJ_ARRAY_SIZE(mvididx);
     unsigned mtotvidcnt = PJ_ARRAY_SIZE(mvididx);
-    pj_uint8_t mtxtidx[PJSUA_MAX_CALL_MEDIA];
+    pj_uint8_t mtxtidx[PJSUA_MAX_CALL_MEDIA] = {0};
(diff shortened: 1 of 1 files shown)


wxiaoguang:
There are other accesses without feature macro check, like https://github.com/pjsip/pjproject/pull/4640/files#diff-29def4cfd6b2e0bb7bc454130942b8d60a375d2fa5ac8aece3bef6639c569e83R2566-R2567 :

```
sort_media2(call->media_prov, sort_check_tp, call->med_prov_cnt,
PJMEDIA_TYPE_VIDEO, mvididx, &mvidcnt, &mtotvidcnt);
```

Maybe the complete fix is to skip the definition:

```
#if PJMEDIA_HAS_VIDEO
pj_uint8_t mvididx[PJSUA_MAX_CALL_MEDIA];
unsigned mvidcnt = PJ_ARRAY_SIZE(mvididx);
#end
```

Or leave them empty ahead, then `mvidcnt = mtotvidcnt = 0;` can be removed?

```
#if PJMEDIA_HAS_VIDEO
pj_uint8_t mvididx[PJSUA_MAX_CALL_MEDIA];
unsigned mvidcnt = PJ_ARRAY_SIZE(mvididx);
...
#else
pj_uint8_t mvididx[1] = {0}; // make it invalid?
unsigned mvidcnt = 0; // treat as empty array
...
#end

```

trengginas:
I tried in gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0, but the warning was not raised.

The other alternative would be to initialize the maudidx/mvididx/mtxtidx explicitly.
e.g:

```
    pj_uint8_t maudidx[PJSUA_MAX_CALL_MEDIA] = {0};
    pj_uint8_t mvididx[PJSUA_MAX_CALL_MEDIA] = {0};
    pj_uint8_t mtxtidx[PJSUA_MAX_CALL_MEDIA] = {0};
```

amubiera:
Updated as suggested by @trengginas

nanangizz (approved):

PR #4654, opened 2025-10-27
Try to use SCHED_RR when setting Posix's thread max priority

trengginas:
On `PJMEDIA` clock-thread, the thread priority will be set to the maximum if `PJMEDIA_CLOCK_NO_HIGHEST_PRIO` is used.
However, on platform's using POSIX thread, the priority will be using the policy currently used by the thread (inherited from the calling thread). 
```
Linux allows the static priority range 1 to 99 for the SCHED_FIFO
       and SCHED_RR policies, and the priority 0 for the remaining
       policies.
```
If `SCHED_OTHER` is used, the priority cannot be changed, so using other policy would be required.
The patch will try to get the maximum thread priority using `SCHED_RR` before falling back to original policy.
Note that `alsa_dev.c` is using `SCHED_RR` to raise its capture thread priority. This patch will also raise the playback thread policy.

### DIFF
--- a/pjlib/src/pj/os_core_unix.c
+++ b/pjlib/src/pj/os_core_unix.c
@@ -194,7 +194,7 @@ static unsigned atexit_count;
 static void (*atexit_func[32])(void);
 
 static pj_status_t init_mutex(pj_mutex_t *mutex, const char *name, int type);
-static int get_prio_max(pj_thread_t *thread, pj_bool_t max_policy);
+static int get_prio_max(pj_thread_t *thread);
 static pj_status_t set_prio(pj_thread_t *thread, int prio, 
                             pj_bool_t max_policy);
 /*
@@ -475,8 +475,11 @@ PJ_DEF(pj_status_t) pj_thread_set_prio(pj_thread_t *thread,  int prio)
     return set_android_thread_priority(prio);
 
 #  else
+    pj_status_t status = PJ_EUNKNOWN;
 
-    pj_status_t status = set_prio(thread, prio, PJ_TRUE);
+    if (prio > 0)
+        status = set_prio(thread, prio, PJ_TRUE);
+
     if (status != PJ_SUCCESS)
         status = set_prio(thread, prio, PJ_FALSE);
 
@@ -538,7 +541,7 @@ static pj_status_t set_prio(pj_thread_t *thread, int prio,
     return PJ_SUCCESS;    
 }
 
-static int get_prio_max(pj_thread_t *thread, pj_bool_t max_policy)
+static int get_prio_max(pj_thread_t *thread)
 {
     int ori_policy, policy;
     int prio;
@@ -549,7 +552,7 @@ static int get_prio_max(pj_thread_t *thread, pj_bool_t max_policy)
     if (rc != 0)
         return -1;
  
-    policy = max_policy?SCHED_RR:ori_policy;
+    policy = SCHED_RR;
     prio = sched_get_priority_max(policy);
     
     /* Make sure the priority is usable. */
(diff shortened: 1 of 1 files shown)


nanangizz:
Considering setting thread prio is not really used a lot in the library (only in pjmedia clock?), perhaps the proposed solution is a bit overkill with the new functions & settings, also the API feels leaning towards POSIX-specific (with `policy` setting), while it requires a specific/uncommon previlege (`root`) to use the capability fully.

So, perhaps we can try with a bit simpler approach first? An idea (not sure if this is feasible though, especially the probing):
- modify `pj_thread_get_prio_max()` implementation (for POSIX) to first probe whether try `SCHED_RR` is allowed, if yes, return the max prio for `SCHED_RR` policy.
- modify `pj_thread_set_prio()` (for POSIX) to use `SCHED_RR` policy if the specified prio is > 0, if it fails, fallback to the current implementation.

For ALSA sound dev, can be left as is (using platform specific API instead of pjlib), because ALSA is already platform specific. Also, perhaps apply higher thread prio to the playback thread?

nanangizz (approved):

pjlib/src/pj/os_core_unix.c
```diff
@@ -474,22 +476,11 @@ PJ_DEF(pj_status_t) pj_thread_set_prio(pj_thread_t *thread,  int prio)
...
-        return PJ_RETURN_OS_ERROR(rc);
-
-    param.sched_priority = prio;
-
-    rc = pthread_setschedparam(thread->thread, policy, &param);
-    if (rc != 0)
-        return PJ_RETURN_OS_ERROR(rc);
-
-    return PJ_SUCCESS;
+    pj_status_t status = set_prio(thread, prio, PJ_TRUE);
```
sauwming: as @nanangizz suggested, shouldn't we check for prio > 0? Three's no point of trying to set the policy to SCHED_RR if prio <= 0

pjlib/src/pj/os_core_unix.c
```diff
@@ -525,22 +516,66 @@ PJ_DEF(int) pj_thread_get_prio_min(pj_thread_t *thread)
...
+ 
+    policy = max_policy?SCHED_RR:ori_policy;
+    prio = sched_get_priority_max(policy);
+    
+    /* Make sure the priority is usable. */
+    param.sched_priority = prio;
+
+    rc = pthread_setschedparam(thread->thread, policy, &param);
     if (rc != 0)
         return -1;
```
sauwming: just return `sched_get_priority_max(ori_policy)` here, so you don't need to call get_prio_max() twice.

pjmedia/src/pjmedia/clock_thread.c
```diff
@@ -357,9 +357,17 @@ static int clock_thread(void *arg)
...
     if ((clock->options & PJMEDIA_CLOCK_NO_HIGHEST_PRIO) == 0) {
-        int max = pj_thread_get_prio_max(pj_thread_this());
-        if (max > 0)
-            pj_thread_set_prio(pj_thread_this(), max);
+        pj_thread_t *this_thread = pj_thread_this();
+        int max = pj_thread_get_prio_max(this_thread);
+        if (max > 0) {
+            pj_status_t status;
+
+            status = pj_thread_set_prio(this_thread, max);
```
sauwming, line 365: Verify that the thread priority is successfully set, such as by using `ps` in Linux.

trengginas: This is the output when priority using default policy:
```    
     PID COMMAND         SCH RTPRIO PRI  NI
   6685 pjsua-x86_64-pc   0      -  19   0
   6685 media             0      -  19   0
   6685 ev_thread         0      -  19   0
   6685 pjsua_0           0      -  19   0
   6685 clock             0      -  19   0
   6685 alsasound_playb   0      -  19   0
   6685 alsasound_captu   0      -  19   0
```

This is when pjsua has elevated permission/SCED_RR is use:
```
    PID COMMAND         SCH RTPRIO PRI  NI
   6406 pjsua-x86_64-pc   0      -  19   0
   6406 media             0      -  19   0
   6406 ev_thread         0      -  19   0
   6406 pjsua_0           0      -  19   0
   6406 clock             2     99 139   -
   6406 alsasound_playb   2     99 139   -
   6406 alsasound_captu   2     99 139   -
```
command:
```
ps -o pid,comm,sched,rtprio,pri,ni -T -p <PID>
```

Issue #2115, opened 2018-05-18
Deadlock between PJSUA LOCK and conference mutex


pjsipbot:
_2018-05-18 03:34:54: @mingteluu_ _created the issue on trac ticket [2115](https://trac.pjsip.org/repos/ticket/2115)_

The same issue as in ticket #1464 happens again.

The ticket doesn't entirely fix the problem, because PJSUA_LOCK() can be called by the upper functions instead, such as:
```
 	pjsua_call_on_state_changed(pjsip_inv_session * inv, pjsip_event * e) Line 4016	C
 	inv_set_state(pjsip_inv_session * inv, pjsip_inv_state state, pjsip_event * e) Line 317	C
 	inv_on_state_null(pjsip_inv_session * inv, pjsip_event * e) Line 3941	C
 	mod_inv_on_tsx_state(pjsip_transaction * tsx, pjsip_event * e) Line 717	C
 	pjsip_dlg_on_tsx_state(pjsip_dialog * dlg, pjsip_transaction * tsx, pjsip_event * e) Line 2069	C
 	mod_ua_on_tsx_state(pjsip_transaction * tsx, pjsip_event * e) Line 178	C
 	tsx_set_state(pjsip_transaction * tsx, pjsip_tsx_state_e state, pjsip_event_id_e event_src_type, void * event_src, int flag) Line 1268	C
 	tsx_on_state_null(pjsip_transaction * tsx, pjsip_event * event) Line 2483	C
 	pjsip_tsx_send_msg(pjsip_transaction * tsx, pjsip_tx_data * tdata) Line 1790	C
 	pjsip_dlg_send_request(pjsip_dialog * dlg, pjsip_tx_data * tdata, int mod_data_id, void * mod_data) Line 1288	C
 	pjsip_inv_send_msg(pjsip_inv_session * inv, pjsip_tx_data * tdata) Line 3282	C
 	on_make_call_med_tp_complete(int call_id, const pjsua_med_tp_state_info * info) Line 518	C
 	pjsua_call_make_call(int acc_id, const pj_str_t * dest_uri, const pjsua_call_setting * opt, void * user_data, const pjsua_msg_data * msg_data, int * p_call_id) Line 919	C
```

In the above stack trace, both pjsua_call_make_call() and on_make_call_med_tp_complete() both call PJSUA_LOCK().

[...]

Issue #1464, opened 2012-03-09
Deadlock between PJSUA LOCK and conference mutex (thanks Marcus Froeschl for the report)


pjsipbot:
_2012-03-09 12:51:47: @nanangizz_ _created the issue on trac ticket [1464](https://trac.pjsip.org/repos/ticket/1464)_

Deadlock scenario:

Thread 1 (audio playback thread):
 1. play_cb() calls conference get_frame()
 1. **conference mutex is acquired**, and conf starts collecting frames from ports
 1. wav player (a conf port member) triggers EOF callback to application
 1. application calls pjsua_conf_disconnect()
 1. check_snd_dev_idle() in pjsua_conf_disconnect() **tries to acquire PJSUA LOCK** that is being held by thread 2.

Thread 2 (pjsua worker thread):
 1. pjsua_call_on_state_changed() invoked
 1. **PJSUA LOCK is acquired**
 1. pjsua triggers on_call_state() callback to application
 1. application calls pjsua_conf_connect()
 1. pjsua_conf_connect() **waits for conference mutex** that is being held by thread 1.

___
_2012-03-15 09:37:20: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2012-03-15 09:37:20: @nanangizz_ _set **resolution** to **fixed**_

___
_2012-03-15 09:37:20: @nanangizz_ _commented_

(In [3977]) Fix #1464: Only use PJSUA lock in PJSUA implementations of INVITE session callbacks when updating 'pjsua_var' is needed, while updating 'pjsua_call' should be enough with call/dialog lock (which is actually being held by the INVITE session layer during invoking its callback).

PR #4664, opened 2025-11-05
Add synchronous conferencer back again

mike2307:
This pull request adds the synchronous (serial) conferencer back again as a selectable variant.
In addition to the original implementation this one contains two improvements:
- For multiple connections from the same source-to-sink a nesting counter is maintained
- For finding a new free slot a dedicated free list is used.

### DIFF
--- a/pjmedia/src/pjmedia/conf_sync.c
+++ b/pjmedia/src/pjmedia/conf_sync.c
@@ -766,6 +766,18 @@ PJ_DEF(pj_status_t) pjmedia_conf_destroy( pjmedia_conf *conf )
     return PJ_SUCCESS;
 }
 
+/*
+ * Register the callback to be called when a port operation has been
+ * completed.
+ */
+PJ_DEF(pj_status_t) pjmedia_conf_set_op_cb(pjmedia_conf *conf,
+                                           pjmedia_conf_op_cb cb)
+{
+    PJ_UNUSED_ARG(conf);
+    PJ_UNUSED_ARG(cb);
+
+    return PJ_ENOTSUP;
+}
 
 /*
  * Destroy the master port (will destroy the conference)


sauwming:
Theoretically, this is indeed good for users who only handle few calls, and hence do not require the performance boost and only runs a minimal risk of deadlock. Practically though, this means maintaining another backend. Recently, it's been quite troublesome to modify all the existing backends. So let's put this to a vote. My vote would unfortunately be a no (but note that majority is not required here, as long as there's one or two positive votes that think this is beneficial, we may still go ahead and merge it).

mike2307:
@sauwming
Thanks for the feedback but let me explain why I added the old synchronous implementation back again...

In the current asynchronous implementation the addition of a new port is done asynchronously as well.
If now I want to directly create a connection with the new port directly afterwards it fails the parameter checks (because the port addition is still pending and thus the argument checks inside `pjmedia_conf_connect_port()` fail). So basically I'd need to synchronize externally, which requires the asynchronous implementation to be somewhat synchronized externally yet again.
This would require a lot of additional work and is likely highly error prone, especially considering that it worked flawless with the old synchronous implementation.

On a long term perspective it probably would be the best to modify conference.c to implement both modes of operation and select it via a define or similar. But that's probably quite a challenge to implement.

nanangizz:
> In the current asynchronous implementation the addition of a new port is done asynchronously as well. If now I want to directly create a connection with the new port directly afterwards it fails the parameter checks (because the port addition is still pending and thus the argument checks inside `pjmedia_conf_connect_port()` fail). So basically I'd need to synchronize externally, which requires the asynchronous implementation to be somewhat synchronized externally yet again. This would require a lot of additional work and is likely highly error prone, especially considering that it worked flawless with the old synchronous implementation.

Which parameter/argument check is failing? The port is actually added synchronously, just the flag `is_new` is pending, so  `pjmedia_conf_connect_port()` should not fail, I've rechecked the lib code & tried it myself.

FYI, the main issue with the old synchronous implementation is a deadlock when media ports use callbacks, and it is a quite a major issue.

> On a long term perspective it probably would be the best to modify conference.c to implement both modes of operation and select it via a define or similar. But that's probably quite a challenge to implement.

Agree, if it is to be implemented, I'd vote modifying the `conference.c` via a define, e.g: invoke `handle_op_queue()` right after queuing the operation (under mutex) and put back mutex into the `get_frame()`.

andreas-wehrmann:
The lifetime issue is indeed extremely annoying and error prone.
So in cases where I really must handle things (like port destruction/removal) synchronously,
I use a wrapper port to decouple the lifetimes.

The wrapper port is the one being added/removed asynchronously to/from the confbridge and simply forwards put_frame()/get_frame() calls to a parent port. The wrapper port has a function with which I can tell it to simply forget the parent port,
so it won't be touched by the confbridge anymore and can then be destroyed immediately.

It feels hacky, but seems to be working fine.
[wrapper.zip](https://github.com/user-attachments/files/23397669/wrapper.zip)

mike2307:
> Which parameter/argument check is failing? The port is actually added synchronously, just the flag `is_new` is pending, so `pjmedia_conf_connect_port()` should not fail, I've rechecked the lib code & tried it myself.

I was testing the latest released version (2.15.1) and ran into a bunch of issues with an unmodified application code from an older version. When debugging I saw that the connect port failed here: https://github.com/pjsip/pjproject/blob/2.15.1/pjmedia/src/pjmedia/conference.c#L1229
Maybe this is not directly caused by the queued add but rather by a lifetime issue.

> FYI, the main issue with the old synchronous implementation is a deadlock when media ports use callbacks, and it is a quite a major issue.

Can you give an example where 2 locks can collide against another and their path? So I can verify if this is an actual potential issue on my side?

Thanks in advance.

andreas-wehrmann:
In the sync implementation, everytime the confbridge runs it holds the conf_mutex for the entirety of the iteration or "tick". So when it calls into a port (for example with `get_frame()`) and in there a callback is called where you then try to interact in any way with the confbridge, you'll deadlock. A usual cause of deadlocks in my apps was (going way back in time now) calling `pjsua_conf_connect()`, `pjsua_conf_disconnect()` in the callbacks. The hideous thing is, you might not be doing something bad in the callback directly but a few functions further down the callstack, so you won't immediately see this kind of issue when looking at just the callback.

nanangizz:
> Can you give an example where 2 locks can collide against another and their path? So I can verify if this is an actual potential issue on my side?

https://github.com/pjsip/pjproject/issues/1464
https://github.com/pjsip/pjproject/issues/2115
https://github.com/pjsip/pjproject/issues/2251

Generally, if your app implements any media port callback (e.g: incoming DTMF, WAV EOF) and calls library functions from within the callback, it may be prone to deadlocks. As @andreas-wehrmann described, the issue is not immediately apparent.

mike2307:
@nanangizz:
I'm not using pjsua but actually just pjmedia to connect some media resources (like RTP stream, sound, tonegen, etc.).
Also the control of the conferencer is strictly kept apart from the media entities, which means that an event/callback from a media port object will **not** have a 1:1 direct influence on the conferencer. Events are nicely decoupled and picked up asynchronously on the application side.

In the linked examples I think the root cause is actually more an issue in the application itself.

But I agree with the following:
> Agree, if it is to be implemented, I'd vote modifying the conference.c via a define, e.g: invoke handle_op_queue() right after queuing the operation (under mutex) and put back mutex into the get_frame().

It would be nice if a simple _ifdef_ could be used to make it synchronous on demand in the same source file. Maybe in addition with a comment of caution that the application shall not call functions in the conferencer when there's an event/callback from a media entity (unless it is nicely decoupled).

nanangizz:
> @nanangizz: I'm not using pjsua but actually just pjmedia to connect some media resources (like RTP stream, sound, tonegen, etc.). Also the control of the conferencer is strictly kept apart from the media entities, which means that an event/callback from a media port object will **not** have a 1:1 direct influence on the conferencer. Events are nicely decoupled and picked up asynchronously on the application side.
> 
> In the linked examples I think the root cause is actually more an issue in the application itself.

Well, I wouldn't say that is an issue in the app, decoupling the callback from the action introduces an additional, non-trivial step for the application.

FYI, some media ports have tried to adopt kind of similar approach by delegating the callback to a [media event framework](https://docs.pjsip.org/en/latest/api/generated/pjmedia/group/group__PJMEDIA__EVENT.html). However, this introduces its own issues, for example, the media event framework also uses mutexes in distributing an event, which theoretically can still lead to deadlocks.

I'm afraid my vote is a no on the sync conf bridge, for now.

sauwming:
We apologize we can't integrate this at the moment. But if you happen to have the time to implement a compile-time switch in the conference, please feel free to create a new PR.

PR #4673, opened 2025-11-11
Fix compile warnings on VS2022

nanangizz:


pjlib/src/pj/ssl_sock_ossl.c
```diff
@@ -1076,7 +1076,7 @@ static int xname_cmp(const X509_NAME **a, const X509_NAME **b) {
 #if !defined(OPENSSL_NO_DH)
 
 static void set_option(const pj_ssl_sock_t* ssock, SSL_CTX* ctx) {
-    unsigned long options = SSL_OP_CIPHER_SERVER_PREFERENCE |
+    unsigned long long options = SSL_OP_CIPHER_SERVER_PREFERENCE |
```
sauwming: From CI build output:
`D:\a\pjproject\pjproject\pjlib\src\pj\ssl_sock_ossl.c(1084,40): warning C4244: 'function': conversion from 'unsigned __int64' to 'unsigned long', possible loss of data [D:\a\pjproject\pjproject\pjlib\build\pjlib.vcxproj]
`

sauwming:
It's a bit difficult to pore over CI build output to see if these changes introduce new warnings in other platforms.

Previously, my intention in #4667 (commit https://github.com/pjsip/pjproject/pull/4667/commits/644ea6c9ef101461f9293334832415fa0d3dd619 - reverted here https://github.com/pjsip/pjproject/pull/4667#pullrequestreview-3441443564) is to suppress all the irrelevant TODO warnings so CI build can be cleaner and easier to check. Is it okay if we use very simple config_site.h instead for these vanilla build? such as `echo -e "#define PJ_TODO(x)" > config_site.h`?

pjlib/src/pj/ssl_sock_ossl.c
nanangizz: Ah, it turns out that the signature of `SSL_CTX_set_options()` differs between OpenSSL 1.x (used in CI test) and OpenSSL 3.x (in my machine):
```
OpenSSL 1.x: unsigned long SSL_CTX_set_options(SSL_CTX *ctx, unsigned long op);
OpenSSL 3.x: uint64_t SSL_CTX_set_options(SSL_CTX *ctx, uint64_t op);
```

nanangizz:
> It's a bit difficult to pore over CI build output to see if these changes introduce new warnings in other platforms.
> 
> Previously, my intention in #4667 (commit [644ea6c](https://github.com/pjsip/pjproject/commit/644ea6c9ef101461f9293334832415fa0d3dd619) - reverted here [#4667 (review)](https://github.com/pjsip/pjproject/pull/4667#pullrequestreview-3441443564)) is to suppress all the irrelevant TODO warnings so CI build can be cleaner and easier to check. Is it okay if we use very simple config_site.h instead for these vanilla build? such as `echo -e "#define PJ_TODO(x)" > config_site.h`?

Yes, I think it is okay.

sauwming:
> Yes, I think it is okay.

Great. I actually want to see if Copilot can join an existing PR (my previous attempt to do so didn't work, but it was a PR created by an external party).

@copilot: Add https://github.com/pjsip/pjproject/commit/644ea6c9ef101461f9293334832415fa0d3dd619 into this PR, but change the usage of `config_site_test.h` to `echo -e "#define PJ_TODO(x)" > config_site.h` instead.

PR #4678, opened 2025-11-13
Fix compile warnings on gcc/llvm

nanangizz:
Heads up: in `third_party/build/webrtc_aec3/Makefile`, will switching the order cause any issue? 

`$(_CXXFLAGS)` may contain `-Wall`, if it is put last, it will override warning suppressions specified in `$(WEBRTC_AEC3_CFLAGS)`.

### DIFF
--- a/pjmedia/src/pjmedia/echo_webrtc_aec3.cpp
+++ b/pjmedia/src/pjmedia/echo_webrtc_aec3.cpp
@@ -31,8 +31,7 @@
 #endif
 
 #if defined(__GNUC__)
-#  pragma GCC diagnostic ignored "-Wunknown-warning-option"
-#  pragma GCC diagnostic ignored "-Wdeprecated-builtins"
+#  pragma GCC diagnostic ignored "-Wdeprecated"
 #endif
 
 #include "modules/audio_processing/aec3/echo_canceller3.h"


sauwming:
Perhaps we can include webrtc aec3 in one CI job build for each platform, so we're able to check if there's any issue with the flag switching.

Any job should be fine, for example on Mac&Linux, we can add it into:
```
  default-pjlib-util-pjmedia-pjnath/default-full-bundle-1:
    # full bundle: enable all codecs + AEC + DTLS
```
since it already specifies full bundle including AEC.

sauwming:
On Linux:
```
../src/pjmedia/echo_webrtc_aec3.cpp:34:34: warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
   34 | #  pragma GCC diagnostic ignored "-Wunknown-warning-option"
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~
../src/pjmedia/echo_webrtc_aec3.cpp:35:34: warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
   35 | #  pragma GCC diagnostic ignored "-Wdeprecated-builtins"
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~
../src/pjmedia/echo_webrtc_aec3.cpp:35:34: note: did you mean ‘-Wdeprecated-declarations’?
```

nanangizz:
Unfortunately it seems to depend on gcc/clang version. They work fine with NDK r27c (clang 18.0.3) & my ubuntu (gcc 13.3.0). Tried something else in the last commit.

sauwming (approved):

PR #4683, opened 2025-11-19
More race condition during AEC settings modification when sound device is running

nanangizz:
Continuation of #4665.

For some audio devices, such as WMME, stopping the audio device is not synchronous. If the audio device callback is still running, accessing the EC state may cause a crash.

Thanks to Iswarya Arunan for the report.

### DIFF
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -7931,11 +7931,14 @@ struct pjsua_media_config
      * Optional callback for audio frame preview right before queued to
      * the speaker.
      * Notes:
-     * - application MUST NOT block or perform long operation in the callback
-     *   as the callback may be executed in sound device thread
-     * - when using software echo cancellation, application MUST NOT modify
+     * - Application MUST NOT block or perform long operation in the callback
+     *   as the callback may be executed in sound device thread.
+     * - When using software echo cancellation, application MUST NOT modify
      *   the audio data from within the callback, otherwise the echo canceller
      *   will not work properly.
+     * - Application MUST NOT stop or switch the audio device, or modify
+     *   audio device settings (e.g: by calling #pjsua_set_ec()) from
+     *   within this callback.
      */
     void (*on_aud_prev_play_frame)(pjmedia_frame *frame);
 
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia/sound_port.c
```diff
@@ -775,12 +798,28 @@ PJ_DEF(pj_status_t) pjmedia_snd_port_set_ec( pjmedia_snd_port *snd_port,
...
             snd_port->ec_state = NULL;
+
+            /* Stop the stream first to prevent race condition. */
+            if (snd_port->aud_started) {
+                restart_stream = PJ_TRUE;
+                pjmedia_aud_stream_stop(snd_port->aud_stream);
+                snd_port->aud_started = PJ_FALSE;
+
+                /* Callback may still be running, wait for it */
+                while (snd_port->in_aud_callback) {
```
sauwming, line 817: Unfortunately using a simple boolean variable will not completely solve the race condition, for example if the thread is already inside `play/rec_cb` but hasn't set the variable yet, or if it's still within the native audio device's callback.

nanangizz: If it hasn't set yet at this point, the `ec_state` should have been reset, so the callback won't access the EC state and the crash can be avoided?

sauwming: Ah okay, so now the only restriction is that app must not call set_ec() from within `on_aud_prev_rec/play_frame()`?

nanangizz: Ah right! Should mention that in the docs.

pjmedia/include/pjmedia/sound_port.h
```diff
@@ -113,12 +113,14 @@ typedef struct pjmedia_snd_port_param
...
-     *   as the callback may be executed in sound device thread
-     * - when using software echo cancellation, application MUST NOT modify
+     * - Application MUST NOT block or perform long operation in the callback
+     *   as the callback may be executed in sound device thread.
+     * - When using software echo cancellation, application MUST NOT modify
      *   the audio data from within the callback, otherwise the echo canceller
      *   will not work properly.
-     * - the return value of the callback will be ignored
+     * - The return value of the callback will be ignored.
+     * - Application MUST NOT call #pjmedia_snd_port_destroy() or
```
sauwming, line 122: Need to update pjsua&pjsua2 docs as well?

nanangizz: I can't find the callbacks in pjsua2, are there any?

sauwming: I forgot that pjsua2 uses the AudioMediaPort. That should be all then.

sauwming (approved):

PR #4699, opened 2025-12-08
Add pjmedia_conf/pjmedia_vid_conf reference to pjmedia_conf_op_info/pjmedia_vid_conf_op_info

trengginas:
This PR will add `pjmedia_conf`/`pjmedia_vid_conf` to `pjmedia_conf_op_info`/`pjmedia_vid_conf_op_info`

### DIFF
--- a/pjmedia/include/pjmedia/conference.h
+++ b/pjmedia/include/pjmedia/conference.h
@@ -171,6 +171,11 @@ typedef union pjmedia_conf_op_param
 typedef struct pjmedia_conf_op_info
 {
     /**
+     * The conference instance.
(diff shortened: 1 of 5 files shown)


sauwming:
I would prefer if we don't create a new API since the previous API has just been released in 2.16.

Adding the conference instance in `pjmedia_conf_op_info` should be sufficient even though it will give more work to the app (such as to maintain mapping between pjmedia conference instances and its high level objects). But considering that multiple conferences setup are quite rare and it should have been obsoleted by parallel conference bridge, it should be okay.

PR #4704, opened 2025-12-09
CMake improvements and macOS support

saghul:
See the individual commits for details.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -429,3 +429,23 @@ jobs:
       with:
         name: ${{ runner.os }}-${{ runner.arch }}-${{ github.job }}-${{ github.run_id }}
         path: artifacts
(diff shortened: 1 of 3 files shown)


saghul:
The tests currently expect a given directory layout, so they fail with stuff like:

```
12:12:52.005 Error loading CA list file '../build/cacert.pem': Unknown OpenSSL error 16777218
```

Thoughts on if-defing based on the build tool? Maybe make it coonfigurable at build time with a define?

saghul:
Some more work might be needed to make tests work with ctest.

Since this is marked as experimental, maybe it's ok not to run them yet?

sauwming:
Yes, I suppose it's ok not to run the tests for now.

saghul:
> Yes, I suppose it's ok not to run the tests for now.

Updated!

PR #4705, opened 2025-12-09
Fix use of external pool with video stream

kunkku:
Passing a non-null pool argument to pjmedia_vid_stream_create caused dereferencing a null pointer when setting up the send manager.


sauwming (approved):
I can reproduce it here. Thanks for the fix

PR #4712, opened 2025-12-14
pcap: Fix bad length read when swap is required

orgads:


orgads:
The entire swap logic is inaccurate. If the capture was taken on little-endian and read on big-endian it is unlikely to work, since `ntohl` and friends are no-op on big-endian. But that's unrelated to this particular fix.

nanangizz:
> The entire swap logic is inaccurate. If the capture was taken on little-endian and read on big-endian it is unlikely to work, since `ntohl` and friends are no-op on big-endian. But that's unrelated to this particular fix.

You're right.

If the capture was taken on little-endian, the magic number would be written as: `d4 c3 b2 a1`.
On a big-endian machine, it will be read as: `0xd4c3b2a1` (in `file->hdr.magic_number`), so `file->swap` will be correctly `PJ_TRUE`, but it swaps using `ntohl` which is a no-op on big-endian. It should use `pj_swap32()` instead.

@copilot please fix this.

nanangizz (approved):

sauwming:
I think you need to create a new issue for Copilot.
I also tried it before in https://github.com/pjsip/pjproject/pull/4585#issuecomment-3399858708 and it didn't work on externally created PR.

Update: Ah, just saw that you've already done so.

PR #4729, opened 2026-01-08
Add pjmedia_ice_parse_cand

saghul:
When PJMEDIA is used without SIP it's useful to be able to parse
candidate lines if they follow the same format as SDP. An example of
this is parsing candidates coming from WebRTC events.

### DIFF
--- a/pjmedia/include/pjmedia/transport_ice.h
+++ b/pjmedia/include/pjmedia/transport_ice.h
@@ -444,6 +444,22 @@ PJ_DECL(pj_status_t) pjmedia_ice_trickle_send_local_cand(
                                             pjmedia_sdp_session *sdp,
                                             pj_bool_t *p_end_of_cand);
 
+/**
+ * Parse an ICE candidate attribute string into a pj_ice_sess_cand structure.
+ *
+ * @param pool      Optional pool for string duplication. If NULL, the
+ *                  foundation string will point to the input string.
+ * @param cand_str  The candidate attribute string to parse (without
+ *                  "candidate:" prefix, just the value part).
+ * @param cand      Pointer to candidate structure to be filled.
+ *
+ * @return          PJ_SUCCESS on success, or PJNATH_EICEINCANDSDP on
+ *                  parsing error, or PJ_EINVAL if parameters are invalid.
+ */
+PJ_DECL(pj_status_t) pjmedia_ice_parse_cand(pj_pool_t *pool,
+                                            const pj_str_t *cand_str,
+                                            pj_ice_sess_cand *cand);
+
 
 PJ_END_DECL
 
--- a/pjmedia/src/pjmedia/transport_ice.c
+++ b/pjmedia/src/pjmedia/transport_ice.c
@@ -775,6 +775,19 @@ PJ_DEF(pj_status_t) pjmedia_ice_trickle_send_local_cand(
 }
 
 
+/*
+ * Public wrapper for parse_cand with parameter validation.
+ */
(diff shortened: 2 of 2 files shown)


pjmedia/src/pjmedia/transport_ice.c
```diff
@@ -775,6 +775,21 @@ PJ_DEF(pj_status_t) pjmedia_ice_trickle_send_local_cand(
...
 
+/*
+ * Public wrapper for parse_cand with parameter validation.
+ */
+PJ_DEF(pj_status_t) pjmedia_ice_parse_cand(const char *obj_name,
+                                           pj_pool_t *pool,
+                                           const pj_str_t *cand_str,
+                                           pj_ice_sess_cand *cand)
+{
+    if (!obj_name || !cand_str || !cand)
```
sauwming: Use PJ_ASSERT

pjmedia/include/pjmedia/transport_ice.h
```diff
@@ -444,6 +444,24 @@ PJ_DECL(pj_status_t) pjmedia_ice_trickle_send_local_cand(
...
+ * @param pool      Optional pool for string duplication. If NULL, the
+ *                  foundation string will point to the input string.
+ * @param cand_str  The candidate attribute string to parse (without
+ *                  "candidate:" prefix, just the value part).
+ * @param cand      Pointer to candidate structure to be filled.
+ *
+ * @return          PJ_SUCCESS on success, or PJNATH_EICEINCANDSDP on
+ *                  parsing error, or PJ_EINVAL if parameters are invalid.
+ */
+PJ_DECL(pj_status_t) pjmedia_ice_parse_cand(const char *obj_name,
```
sauwming: Usually pool is the first parameter.

Also, it will be better if obj_name is optional, or just remove the parameter altogether if it's not really necessary (you can generate it in the function instead, such as "icecnd%p")

saghul: I can do that, I was just keeping the status quo, but indeed that's better.

saghul:
@sauwming Done!

pjmedia/include/pjmedia/transport_ice.h
nanangizz: I am not sure if replacing `obj_name`, which originally contains `tp_ice->base.name`, with the candidate address is a good idea (haven't checked the details, but candidate address sounds less useful info than `tp_ice->base.name` in debugging?). So instead of removing `parse_cand()`, perhaps `pjmedia_ice_parse_cand()` can be implemented as a thin wrapper calling `parse_cand()` with `NULL` or self-generated `obj_name` (may be candidate address)?

saghul: Note that it onay logged the transport in a single place (or 2?) the other places pass null or a static string.

Since it's also at trace level, and not enabled by default, not sure if it's that useful anyway?

Thoughts?

sauwming: Even though it's only for tracing, I prefer your first commit because it's cleaner and doesn't change existing code.

If you don't think obj_name is that useful for the API users, you can pass empty string for now. So:
```
PJ_DEF pjmedia_ice_parse_cand()
{
    parse_cand("", ...);
}
```

saghul: Ok, I'll go back to that approach.

saghul: Done!

saghul:
Updated again!

sauwming (approved):

saghul:
Thanks guys!

PR #4730, opened 2026-01-08
Add Dockerfile for PJSIP with Python bindings

DevNull-IR:
This Dockerfile sets up an environment for building and installing PJSIP along with its Python bindings, ensuring all necessary dependencies are included.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,41 @@
+FROM ubuntu:22.04
+
+WORKDIR /pjsip
+
+COPY . /pjsip
+
+RUN apt-get update && apt-get install -y \
+    swig \
+    python3 \
+    python3-pip \
+    python3-venv \
+    build-essential \
+    gcc \
+    g++ \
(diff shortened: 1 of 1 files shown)


sauwming (approved):

sauwming:
I have no objection, let's see what others think

wxiaoguang:
It seems that it doesn't generate a clean image for production usage because all temp (intermediate) files and toolchains are left in the image filesystem?

wxiaoguang:
And ubuntu 22.04 is very old, it will reach EOL in 2027 (https://ubuntu.com/about/release-cycle), and its Python package is also pretty old 3.10, which will reach EOL in October 2026 (https://devguide.python.org/versions/)

So such image isn't likely to be a general-purpose or widely-used one?

sauwming:
Closing this due to lack of follow up regarding the modifications required. If there are more demands, we can consider creating our own.

PR #4731, opened 2026-01-08
Fix processing peer reflexive ICE candidates with empty checklist

saghul:
When we learn of a peer-reflexive ICE candidate through STUN binding
requests, try to match the local candidate to the list oof local
candidates rather than the checklist, since it might be empty.

The checklist can be empty when using trickle ICE for example, if remote
cancidates haven't been signalled yet.


sauwming (approved):

sauwming:
Looks okay to me, but @nanangizz  should be more expert in this area

PR #4736, opened 2026-01-11
Always use crypto library for GCM test for rstp when building with CMake

joelpelaez:
When building on Fedora CMake detects OpenSSL in the first attempt and the variable `OPENSSL_LIBRARIES` only references to `ssl` and the GCM test doesn't build and srtp will use internal GCM implementation.

This change is related to #4732 PR.

### DIFF
--- a/third_party/srtp/CMakeLists.txt
+++ b/third_party/srtp/CMakeLists.txt
@@ -14,7 +14,7 @@ if(SRTP_WITH_OPENSSL)
 
     cmake_push_check_state(RESET)
       set(CMAKE_REQUIRED_INCLUDES "${OPENSSL_INCLUDE_DIR}")
-      set(CMAKE_REQUIRED_LIBRARIES "${OPENSSL_LIBRARIES}")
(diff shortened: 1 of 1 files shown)


nanangizz (approved):
From [here](https://cmake.org/cmake/help/latest/module/FindOpenSSL.html#result-variables), `OPENSSL_LIBRARIES` should refer to all OpenSSL libraries including `crypto`. Anyway, SRTP GCM seems to requires `crypto` only.

joelpelaez:
> From [here](https://cmake.org/cmake/help/latest/module/FindOpenSSL.html#result-variables), `OPENSSL_LIBRARIES` should refer to all OpenSSL libraries including `crypto`. Anyway, SRTP GCM seems to requires `crypto` only.

Yes, OPENSSL_LIBRARIES must include crypto, but in Fedora it returns only libssl.so and the GCM test code doesn't link and mark the test as failed, when it would pass.

aconfigure do the same test using only libcrypt.so: https://github.com/pjsip/pjproject/blob/8b2189def13167e84237506de8cc72c51bb5daf6/aconfigure.ac#L2216-L2225

SRTP target_link_libraries remains untouched (links to both libssl and libcrypto).

Issue #4735, opened 2026-01-09
SIPREC module breaks existing SRS implementation


andreas-wehrmann:
### Describe the feature

I did implement a SRC as well as an SRS based on PJSUA but am running into issues with the SRS after updating to PJ 2.16.
This is because of the introduction of the SIPREC module in PJ itself (#4132).

In it's current state I don't know what the purpose of the SIPREC module exactly is to be honest.
That is because the module is overly restrictive and prevents valid use cases.
From what I could see it only checks whether an incoming INVITE indicates
if the call is supposed to be an RS and extracts the XML metadata document from the body.

The main issues I've got are:
- INVITEs without an rs-metadata document are rejected
- INVITEs that contain SDP media without a label attribute are rejected
- rs-metadata updates are not handled at all (I realise this has been mentioned in the original PR)

Since I was able to implement all of this with "standard PJSUA" I don't see why the module (in its current form)
was necessary in the first place.

I'm creating this issue in order to have a discussion on how to go forward with this.
Session recording is often considered critical and the SRS should therefore be resilient,
stable but also tolerant of what it receives meaning the SRS should do its best to make sense of what it receives,
even when there are inconsistencies.

I guess it boils down to this:
- What is the actual purpose of the SIPREC module?
- What should its reponsibilities be?
- How should PJSUA interact with it, if at all?

I'm sorry to bring up these basic questions but it's a complex subject.

For my personal use case the easiest option would be to get rid of the SIPREC module.
Unfortunately I don't have an easy way of "disabling" the module, so I'll have to do code patches.

Let me maybe elaborate a bit on how my implementation roughly works.
Deciding to go with PJSUA I realised that having the confbridge around (audio or video) is disadvantageous
because for the purpose of recording it only gets in the way and needlessly consumes processing power
(there is no point in doing decoding or resampling when all you want to do is to record the stream along with some metadata).
On the other hand though, PJSUA makes it easy to quickly write a UA and it provides many customisation points
which allow for implementing special use cases.

- Adding the "siprec" capability to stock PJSUA is easily done through calling pjsip_endpt_add_capability() without a module;
  this allows for receiving calls that require siprec.
- INVITE and metadata as well as update handling is done through the normal callbacks.
- Checking whether an INVITE indicates an RS is just checking the Require header
  and the existance of the +sip.src feature tag in the Contact header
  (NOT the SDP media labels and NOT the rs-metadata body!)
[...]

PR #4745, opened 2026-01-19
Remove `siprec` from Supported header when session recording is inactive

nanangizz:
When `use_siprec` or `siprecUse` is set to `PJSUA_SIP_SIPREC_INACTIVE`, the Supported header must not include the `siprec` tag. It may activate something in the server that should not be activated.

Thanks Sean Riley for the report.


sauwming (approved):

trengginas (approved):

sauwming:
Does this also fix #4735?

nanangizz:
> Does this also fix #4735?

I don't think so, that is a different issue.

PR #3837, opened 2024-01-22
Fixed deprecation warnings on Mac and iOS

sauwming:
Increase default minimum iOS version to 11 as XCode 14 doesn't support deployment to releases older than iOS 11 (note that latest XCode is now XCode 15).
https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes
`Building for deployment to OS releases older than macOS 10.13, iOS 11, tvOS 11, and watchOS 4 is no longer supported. (92834476)`

Also fixed various deprecation warnings on Mac and iOS:
```
../src/pjmedia-audiodev/coreaudio_dev.m:557:25: warning: 'kAudioObjectPropertyElementMaster' is
 deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
        addr.mElement = kAudioObjectPropertyElementMaster;
                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                        kAudioObjectPropertyElementMain
```

```
../src/pjmedia-videodev/darwin_dev.m:329:21: warning: 'AVCaptureDeviceTypeBuiltInDuoCamera' is
 deprecated: first deprecated in iOS 10.2 - Use AVCaptureDeviceTypeBuiltInDualCamera instead. [-Wdeprecated-declarations]
                  , AVCaptureDeviceTypeBuiltInDuoCamera
```

```
../src/pjmedia-videodev/darwin_dev.m:342:41: warning: 'devices' is deprecated: first deprecated in
 iOS 10.0 - Use AVCaptureDeviceDiscoverySession instead. [-Wdeprecated-declarations]
            dev_list = [AVCaptureDevice devices];
```

```
../src/pj/activesock.c:155:37: warning: 'kCFStreamNetworkServiceTypeVoIP' is deprecated: first 
deprecated in iOS 9.0 - use PushKit for VoIP control purposes [-Wdeprecated-declarations]
                                    kCFStreamNetworkServiceTypeVoIP)
```

PR #4751, opened 2026-01-21
Fix incorrect capture angle when setting orientation

trengginas:
Changes e15b9d498fe1ffed167cd271cc4fdd4e4888131b use `videoRotationAngle` to replace deprecated `videoOrientation`.

This can raise issue since different devices might have different camera sensor. 

|Device Category | Portrait Rotation Angle |	Reason |
------------------|-----------------------|-----------|
| Home Button Devices (SE, 8, etc.) | 90.0° | Sensors are natively landscape; 90° rotation is the standard correction for portrait. |
|Modern "Full Screen" Devices (13, 14, 15, 16) | 0.0° or 90.0° |	Depends on the specific camera (Front vs. Back). The Back camera often still reports 90°, while the Front camera might report 0°. |
| iPhone 17 / Square Sensor | 0.0° | The square sensor can capture vertical and horizontal frames natively without needing to "flip" the hardware buffer.|

Possible issue (On iPhone17, the image is doubled with multiple stripes):
|Iphone Se | Iphone 17|
|----------|-----------|
|angle=0, vid_orient=4|angle=270, vid_orient=4 | 
|angle=270, vid_orient=2 | angle=180, vid_orient=2 |
|angle=180, vid_orient=3|angle=90, vid_orient=3|
|angle=90, vid_orient=1|angle=0, vid_orient=1|

This patch will use `AVCaptureDeviceRotationCoordinator` to determine the correct angle regardless the device, specifically for devices using iOS 17 or newer.

### DIFF
--- a/pjmedia/src/pjmedia-videodev/darwin_dev.m
+++ b/pjmedia/src/pjmedia-videodev/darwin_dev.m
@@ -39,7 +39,9 @@
 #define DEFAULT_WIDTH           640
 #define DEFAULT_HEIGHT          480
 #define DEFAULT_FPS             15
-/* When camera orientation is changed, will level relative to gravity */
+/* When the camera orientation changes, maintain horizon-level capture
+ * relative to gravity.
+ */
 #define USE_HORIZON_LEVEL       1
 
 /* Define whether we should maintain the aspect ratio when rotating the image.
@@ -151,10 +153,7 @@ struct darwin_stream
     UIView              *render_view;
     AVCaptureVideoPreviewLayer  *prev_layer;
     UIView                      *prev_view;
-
-    void                *rotation_coord;
 #endif
-
     pj_timestamp         frame_ts;
     unsigned             ts_inc;
 };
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia-videodev/darwin_dev.m
```diff
@@ -39,6 +39,10 @@
 #define DEFAULT_WIDTH           640
 #define DEFAULT_HEIGHT          480
 #define DEFAULT_FPS             15
+/* When the camera orientation changes, maintain horizon-level capture
+ * relative to gravity.
+ */
+#define USE_HORIZON_LEVEL       1
```
nanangizz, line 45: Why is this configurable?

pjmedia/src/pjmedia-videodev/darwin_dev.m
```diff
@@ -1308,16 +1317,25 @@ static pj_status_t darwin_stream_set_cap(pjmedia_vid_dev_stream *s,
...
+
+#if (USE_HORIZON_LEVEL)
+                AVCaptureDeviceRotationCoordinator *coord =
+                    [[AVCaptureDeviceRotationCoordinator alloc]
+                     initWithDevice:strm->dev_input.device
+                     previewLayer:nil];
+                if (coord) {
+                    rotation = coord.videoRotationAngleForHorizonLevelCapture;
+                    [coord release];
+                }
```
nanangizz, line 1332: The PR desc says "regardless the device", but this is inside `__IPHONE_17_0`, how about other devices?

trengginas: Updated the PR comments. The changes related to #3837 which is applicable for iOS 17 or newer.
Older devices, will still use `videoOrientation`.

pjmedia/src/pjmedia-videodev/darwin_dev.m
trengginas: This because the patch will add behavior changes. The angle is by `videoRotationAngleForHorizonLevelCapture` [ref](https://developer.apple.com/documentation/avfoundation/avcapturedevice/rotationcoordinator/videorotationangleforhorizonlevelcapture?language=objc). Which means the capture image will always be level, depending on the current camera orientation (e.g.: the image cannot be upside-down). The alternative would be to use the previous logic by the orientation/angle mapping. But we need to get the "baseline" angle for the current device to the mapping equation. I haven't found the way to get the "baseline" angle in runtime.

sauwming (approved):

PR #3714, opened 2023-09-21
Fixed deadlock between SIP transaction and dialog

sauwming:
It is reported that deadlock can occur when a dialog receives a request and its retransmission at the same time. The stack trace is as follows:
```
      PjSipDll.dll!grp_lock_acquire(void * p=0x06ae9d3c) Line 297 C
      PjSipDll.dll!pjsip_tsx_recv_msg(pjsip_transaction * tsx=0x01dff41c, pjsip_rx_data * rdata=0x01e945ac) Line 1902   C
      PjSipDll.dll!pjsip_dlg_on_rx_request(pjsip_dialog * dlg=0x01dfc3ec, pjsip_rx_data * rdata=0x01e945ac) Line 1785   C
      PjSipDll.dll!mod_ua_on_rx_request(pjsip_rx_data * rdata=0x01e945ac) Line 748  C
      PjSipDll.dll!pjsip_endpt_process_rx_data(pjsip_endpoint * endpt=0x03fb7084, pjsip_rx_data * rdata=0x01e945ac, pjsip_process_rdata_param * p=0x04c2278c, int * p_handled=0x04c2272c) Line 929    C
      PjSipDll.dll!endpt_on_rx_msg(pjsip_endpoint * endpt=0x03fb7084, int status=66810160, pjsip_rx_data * rdata=0x01e945ac) Line 1079    C
      PjSipDll.dll!pjsip_tpmgr_receive_packet(pjsip_tpmgr * mgr=0x01e86d5c, pjsip_rx_data * rdata=0x01e945ac) Line 2199 C
      PjSipDll.dll!udp_on_read_complete(pj_ioqueue_key_t * key=0x0404c9f8, pj_ioqueue_op_key_t * op_key, long bytes_read=465) Line 202    C
      PjSipDll.dll!ioqueue_dispatch_read_event(pj_ioqueue_t * ioqueue=0x0402f27c, pj_ioqueue_key_t * h=0x0404c9f8) Line 609   C
```

```
      [Inline Frame] PjSipDll.dll!pjsip_dlg_inc_lock(pjsip_dialog *) Line 928 C
      PjSipDll.dll!pjsip_dlg_on_tsx_state(pjsip_dialog * dlg=0x01dfc3ec, pjsip_transaction * tsx=0x01dff41c, pjsip_event * e=0x04fe24dc) Line 2135    C
[...]

Issue #2001, opened 2017-03-20
Deadlock between dialog lock and transaction group lock


pjsipbot:
_2017-03-20 10:37:42: @mingteluu_ _created the issue on trac ticket [2001](https://trac.pjsip.org/repos/ticket/2001)_

Sample scenario:
- PJSIP receives a re-INVITE and send 200 OK.
- The ACK is delayed and received at the same time the timeout timer fires.

Thread 1 receives the delayed ACK:
```
- pjsip_dlg_on_rx_request() -> pjsip_dlg_inc_lock(dlg)
- pjsip_tsx_terminate() -> pj_grp_lock_acquire(tsx->grp_lock)
```

Thread 2 handles the timeout:
```
- tsx_timer_callback() -> pj_grp_lock_acquire(tsx->grp_lock) then (*tsx->state_handler)(tsx, &event);
- tsx_on_state_completed_uas()
- tsx_set_state() -> (*tsx->tsx_user->on_tsx_state)(tsx, &e),
- pjsip_dlg_on_tsx_state() -> pjsip_dlg_inc_lock(dlg);
```

Thread 1's stack trace:
```
#5  0x00007f4748aac5a2 in pj_grp_lock_acquire (grp_lock=<optimized out>) at ../src/pj/lock.c:478
#6  0x00007f4749be20cd in pjsip_tsx_terminate (tsx=0x7f466ee29bb8, code=408) at ../src/pjsip/sip_transaction.c:1636
#7  0x00007f474a021dc6 in mod_inv_on_rx_request (rdata=0x7f46c4074c38) at ../src/pjsip-ua/sip_inv.c:586
#8  0x00007f4749be4b01 in pjsip_dlg_on_rx_request (dlg=dlg@entry=0x7f466eacb138, rdata=rdata@entry=0x7f46c4074c38) at ../src/pjsip/sip_dialog.c:1694
#9  0x00007f4749be6147 in mod_ua_on_rx_request (rdata=0x7f46c4074c38) at ../src/pjsip/sip_ua_layer.c:699
#10 0x00007f4749bccf77 in pjsip_endpt_process_rx_data (endpt=<optimized out>, rdata=rdata@entry=0x7f46c4074c38, p=p@entry=0x7f46d0004c40 <param>, p_handled=p_handled@entry=0x7f466155ecf4) at ../src/pjsip/sip_endpoint.c:887
```

Thread 2's stack trace:
```
[...]

PR #4755, opened 2026-01-23
Lock ordering issue between sip dialog and sip transaction

sauwming:
This is a classic lock ordering inversion issue in outgoing message and incoming message flows between sip dialog and sip transaction. In outgoing flow, dlg lock will be locked first then tsx, while in incoming flow, tsx lock will be locked first.

Example stack traces:
* UAC dialog:
```
# Mutex M1 acquired here while holding mutex M0 in thread T2:
#4 pj_grp_lock_acquire ../src/pj/lock.c:486
#5 pjsip_tsx_set_transport ../src/pjsip/sip_transaction.c:1841
#6 pjsip_dlg_send_request ../src/pjsip/sip_dialog.c:1384
#7 pjsip_inv_send_msg ../src/pjsip-ua/sip_inv.c:3825
#8 perform_test ../src/test/inv_offer_answer_test.c:474

# Mutex M0 acquired here while holding mutex M1 in thread T2:
#8 pj_grp_lock_acquire ../src/pj/lock.c:486
#9 pjsip_dlg_inc_lock ../src/pjsip/sip_dialog.c:965
#10 pjsip_dlg_on_tsx_state ../src/pjsip/sip_dialog.c:2249
#11 mod_ua_on_tsx_state ../src/pjsip/sip_ua_layer.c:186
#12 tsx_set_state ../src/pjsip/sip_transaction.c:1460
#13 tsx_on_state_proceeding_uac ../src/pjsip/sip_transaction.c:3404
#14 pjsip_tsx_recv_msg ../src/pjsip/sip_transaction.c:2061
```
* UAS dialog:
```
Mutex M0 acquired here while holding mutex M1 in thread T4:
#4 pj_grp_lock_acquire ../src/pj/lock.c:486
#5 create_uas_dialog ../src/pjsip/sip_dialog.c:577
#6 pjsip_dlg_create_uas_and_inc_lock ../src/pjsip/sip_dialog.c:668
#7 on_rx_request ../src/test/inv_offer_answer_test.c:326
#8 pjsip_endpt_process_rx_data ../src/pjsip/sip_endpoint.c:961

Mutex M1 acquired here while holding mutex M0 in thread T4:
#11 pjsip_dlg_inc_lock ../src/pjsip/sip_dialog.c:965
#12 pjsip_dlg_on_tsx_state ../src/pjsip/sip_dialog.c:2252
#13 mod_ua_on_tsx_state ../src/pjsip/sip_ua_layer.c:186
#14 tsx_set_state ../src/pjsip/sip_transaction.c:1458
#15 tsx_on_state_null ../src/pjsip/sip_transaction.c:2842
#16 pjsip_tsx_recv_msg ../src/pjsip/sip_transaction.c:2059
#17 pjsip_dlg_on_rx_request ../src/pjsip/sip_dialog.c:1882
#18 mod_ua_on_rx_request ../src/pjsip/sip_ua_layer.c:745
#19 pjsip_endpt_process_rx_data ../src/pjsip/sip_endpoint.c:967
```

There doesn't seem to be any straightforward solution to fix the problem, so even though not ideal, we just use a TSan suppression filter for now.

A few ideas to solve the issue in the future:
* Using chain lock
Patch here: https://github.com/pjsip/pjproject/pull/4755/commits/87bfa3d6b71597c4499adf94a8ffdd60f4b76b82
Pro: This should be able to solve any deadlock should it occur
Con: Unfortunately, TSan will have no knowledge that the two locks have been chained together so it will still report it.
Before chain:
TX flow: `dlg_lock->tsx_lock`
RX flow: `tsx_lock->dlg_lock`
After chain:
TX flow: `dlg_lock->(dlg_lock->tsx_lock)`
RX flow: `(dlg_lock->tsx_lock)->dlg_lock`

  **Note**: One idea (by @nanangizz [below](https://github.com/pjsip/pjproject/pull/4755#issuecomment-3809568652)) to solve this is by allowing certain locks to skip locking if it is already held (note that info about the lock owner and nesting counter is available but only in debug mode).
* Using a single group lock between sip dlg and sip tsx
  Pro: Solve any potential deadlock issue and will also clear TSan
  Con:
  - Performance penalty (should not be much though (?) since a dialog should not have that many transactions happening simultaneously).
  - **Major**: Need to address sip tsx lifetime issue because by using a single group lock, all transactions will only be destroyed much later, together with dlg. So perhaps sip tsx needs a separate ref counter (by using a separate group lock?)

Note that we may be able to safely avoid locking in `pjsip_dlg_on_tsx_state()` (patch: https://github.com/pjsip/pjproject/pull/4755/commits/4e1979ff9be7861a78d33cf7c943d16798d39cae) since dialog usages shouldn't change midway. However, since removing the dlg lock here won't help much with the lock ordering issue (because we will still lock the dialog in another function anyway), there's not much benefit in committing the patch.

### DIFF
--- a/pjsip/src/pjsip/sip_dialog.c
+++ b/pjsip/src/pjsip/sip_dialog.c
@@ -2240,6 +2240,9 @@ void pjsip_dlg_on_tsx_state( pjsip_dialog *dlg,
               tsx->obj_name, pjsip_tsx_state_str(tsx->state)));
     pj_log_push_indent();
 
+    /* Lock the dialog and increment session. */
+    pjsip_dlg_inc_lock(dlg);
+
     /* Pass to dialog usages. */
     for (i=0; i<dlg->usage_cnt; ++i) {
 
@@ -2258,17 +2261,12 @@ void pjsip_dlg_on_tsx_state( pjsip_dialog *dlg,
         tsx->mod_data[dlg->ua->id] == dlg)
     {
         pj_assert(dlg->tsx_count>0);
-
-        /* Lock the dialog and increment session. */
-        pjsip_dlg_inc_lock(dlg);
-
         --dlg->tsx_count;
         tsx->mod_data[dlg->ua->id] = NULL;
-
-        /* Unlock dialog and dec session, may destroy dialog. */
-        pjsip_dlg_dec_lock(dlg);
     }
 
+    /* Unlock dialog and dec session, may destroy dialog. */
+    pjsip_dlg_dec_lock(dlg);
     pj_log_pop_indent();
 }
 


nanangizz:
From the sample call stack trace above, the first stack shows a request being sent, which typically occurs when a UAC tsx has just been created. It should be impossible for an incoming response to arrive concurrently before the request is sent?

nanangizz:
Btw, in the second stack, `pjsip_dlg_on_tsx_state()` should have already acquired the dialog lock, it should not be waiting until `pjsip_inv_create_ack()`, or something (temporarily) released the dialog lock in the middle?

Also, for suppression string, instead of using dialog function, perhaps better to use a more specific function, e.g: `inv_send_ack()`, to narrow the suppression scope?

sauwming:
> Btw, in the second stack, `pjsip_dlg_on_tsx_state()` should have already acquired the dialog lock, it should not be waiting until `pjsip_inv_create_ack()`, or something (temporarily) released the dialog lock in the middle?

Right, I temporarily removed the locking in `pjsip_dlg_on_tsx_state()`. I have updated the stack trace in the PR description.

> Also, for suppression string, instead of using dialog function, perhaps better to use a more specific function, e.g: `inv_send_ack()`, to narrow the suppression scope?

Unfortunately it's not limited to ACK sending.

> From the sample call stack trace above, the first stack shows a request being sent, which typically occurs when a UAC tsx has just been created. It should be impossible for an incoming response to arrive concurrently before the request is sent?

True, but TSan doesn't know that. All it knows is that during the lifetime of the app, there are two locks that are acquired in different ordering.
The stack trace is just one example (I have added another one in the PR desc for the UAS case), but there are many more of different varieties following similar patterns involving outgoing and incoming message flow. Just like you say with the sample trace, most of them are actually impossible to cause deadlock in real life scenario, but a very rare few could indeed happen, such as:
https://github.com/pjsip/pjproject/pull/3714
https://github.com/pjsip/pjproject/issues/2001

nanangizz:
> Con: Unfortunately, TSan will have no knowledge that the two locks have been chained together so it will still report it.

If TSan still reports it, perhaps the order is indeed still inversed because AFAIK the report should be based on code execution rather than code reading (TSan replaces/wraps the original SDK mutex functions with its own). This suggests there may be a bug in our code.

I tried to investigate and found something suspicious:
- `grp_lock_acquire()` locks `glock->lock_list.next` first
https://github.com/pjsip/pjproject/blob/96c16eb8665f4778bab015a46032843d5ed9dde0/pjlib/src/pj/lock.c#L295-L298
- But `pj_grp_lock_chain_lock()` does not modify the `glock->lock_list.next`
https://github.com/pjsip/pjproject/blob/96c16eb8665f4778bab015a46032843d5ed9dde0/pjlib/src/pj/lock.c#L682-L685

So `pj_grp_lock_chain_lock(tsx_lock, dlg->grp_lock_, 0)` will generate a chain of `{tsx_lock->dlg_lock}` instead of `{dlg_lock->tsx_lock}`?

sauwming:
> So `pj_grp_lock_chain_lock(tsx_lock, dlg->grp_lock_, 0)` will generate a chain of `{tsx_lock->dlg_lock}` instead of `{dlg_lock->tsx_lock}`?

I tested it here and it seems to be correct. `pj_list_insert_before()` will update the links of the previous and next nodes.

> If TSan still reports it, perhaps the order is indeed still inversed because AFAIK the report should be based on code execution rather than code reading (TSan replaces/wraps the original SDK mutex functions with its own). This suggests there may be a bug in our code.

TSan seems to still report the following pattern:
`L1->L1->L2`
`L1->L2->L1`
even though in the second flow, L1 is just reacquiring the same lock.

nanangizz:
> TSan seems to still report the following pattern: `L1->L1->L2` `L1->L2->L1` even though in the second flow, L1 is just reacquiring the same lock.

Ah okay. After a quick search, it looks like there are some reports of TSan producing false positive lock inversion warnings.

We can modify the `lock_acquire()` functions to skip locking if it is already held since info about the lock owner and nesting counter is available (in debug mode), but I think it is too far.

sauwming:
> We can modify the `lock_acquire()` functions to skip locking if it is already held since info about the lock owner and nesting counter is available (in debug mode), but I think it is too far.

This sounds like a good idea IMO. Wondering whether there's any downside of this approach though?

nanangizz:
> This sounds like a good idea IMO. Wondering whether there's any downside of this approach though?

- Behavioral change: all mutexes will become recursive, e.g: this may hide issues in code that locks a non-recursive mutex twice.
- Only applies in debug mode.
- Effort & risk taken solely to workaround known TSan limitations :) or is there another purpose?

sauwming:
> * Behavioral change: all mutexes will become recursive, e.g: this may hide issues in code that locks a non-recursive mutex twice.
> * Only applies in debug mode.

We can add `pj_lock/grp_lock_set_mode()` to make the "feature" applicable to particular mutexes only.

> * Effort & risk taken solely to workaround known TSan limitations :) or is there another purpose?

The grand goal is to make PJSIP deadlock-free :) If we can pass TSan without the need to use any suppression, then we should be able to make that claim.
So I'll take note of your idea in this PR desc as well as in our roadmap, in case that we need it for the future.

PR #4758, opened 2026-01-27
Avoid race when using the same op key in ioq send/recv/accept

sauwming:
There could be a race when app is trying to issue multiple calls at the same time to `pj_ioqueue_recv()/recvfrom()` or `pj_ioqueue_send()/sendto()` or `pj_ioqueue_accept()` using the same key and op key.
This could cause a crash, such as:
```
0x00007f02250f76cc in pj_list_insert_before (pos=pos@entry=0x7f02531fe7c0, node=node@entry=0x7f0253309408) at pjlib/include/pj/list_i.h:39
0x00007f0225109758 in pj_ioqueue_recvfrom (key=0x7f02531fe748, op_key=0x7f0253309408, buffer=<optimized out>, length=0x7f015aeeca88, flags=0, addr=0x0, addrlen=0x0) at pjlib/src/pj/ioqueue_common_abs.c:846
```

Note that serial calls using the same key and op key will be handled by the assertion `PJ_ASSERT_RETURN(read_op->op == PJ_IOQUEUE_OP_NONE` (or `write/accept_op` for sending/accept).


nanangizz:
Also in `pj_ioqueue_accept()`?

PR #4760, opened 2026-01-27
size of additional short rtp header has to be 8

lukas-nowak:
If pjmedia_rtp_add_hdr_short structure is based on pj_uint16_t than it is copied to payload buffer of RTP red packet as 2 bytes, the second one is just 0x00. But this is wrong according the standard, there is no any alignment at this point. The data of packets should start imeediately as in RFC 2198 : " The final header is followed, immediately, by the data blocks, stored
   in the same order as the headers. ". In attachment there is also how wireshark displays such packet, with additional wrong 0x00 at the beginning of data.
wrong-red-rtp-header

### DIFF
--- a/pjmedia/src/pjmedia/txt_stream.c
+++ b/pjmedia/src/pjmedia/txt_stream.c
@@ -499,7 +499,7 @@ static pj_status_t decode_red(pjmedia_txt_stream *stream,
             return PJ_ETOOBIG;
 
         pj_memcpy(&shdr, buf, sizeof(shdr));
-        if (shdr.pt != pt) {
+        if (shdr.pt != (pj_uint8_t)pt) {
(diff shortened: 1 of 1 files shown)


sauwming:
There are a couple of casts to `pj_uint16_t` in txt_stream that needs to be changed as well?
i.e. `shdr.pt = (pj_uint8_t)pt` and `hdr.pt = (pj_uint8_t)pt`

Also, we'd appreciate if you can sign the CLA, thank you.

lukas-nowak:
Unfortunately I commited with email I don't have access to. I suspect this is the reason the CLA process does not go through. 
I can commit these changes with new PR with proper email if you want to.

sauwming:
Yes, a new PR would be better, thanks.

PR #4764, opened 2026-01-29
Lock order inversion issue in sip transport

sauwming:
Lock order inversion in sip transport between pjsip_transport and pjsip_tpmgr.

Example stack trace:
```
# Mutex of pj_ioqueue acquired here while holding mutex of pjsip_tpmgr:
#5 pj_ioqueue_lock_key ../src/pj/ioqueue_common_abs.c:1478
#6 pj_ioqueue_connect ../src/pj/ioqueue_common_abs.c:1305
#7 pj_activesock_start_connect ../src/pj/activesock.c:961
#8 lis_create_transport ../src/pjsip/sip_transport_tcp.c:1063
#9 pjsip_tpmgr_acquire_transport2 ../src/pjsip/sip_transport.c:2800
#10 pjsip_tpmgr_acquire_transport ../src/pjsip/sip_transport.c:2493
#11 pjsip_endpt_acquire_transport ../src/pjsip/sip_endpoint.c:1234
#
# Mutex of pjsip_tpmgr acquired here while holding mutex of pj_ioqueue:
#2 pj_lock_acquire ../src/pj/lock.c:179
#3 tp_state_callback ../src/pjsip/sip_transport.c:2924
#4 on_connect_complete ../src/pjsip/sip_transport_tcp.c:1569
#5 ioqueue_on_connect_complete ../src/pj/activesock.c:976
#6 ioqueue_dispatch_write_event ../src/pj/ioqueue_common_abs.c:287
#7 pj_ioqueue_poll ../src/pj/ioqueue_select.c:1097
#
# Mutex of pjsip_tpmgr acquired here while holding mutex of pjsip_transport:
#2 pj_lock_acquire ../src/pj/lock.c:179
#3 destroy_transport ../src/pjsip/sip_transport.c:1415
#4 pjsip_transport_destroy ../src/pjsip/sip_transport.c:1583
```

This can cause issues such as https://github.com/pjsip/pjproject/pull/4307.

Proposed fix:
Obey a uniform lock order: always lock pjsip_tpmgr first, and then pjsip_transport
The reason for this order is that the natural usage is top-down, i.e. by using APIs such as: `pjsip_tpmgr_acquire_transport()`, `pjsip_tpmgr_shutdown_all()`, `pjsip_transport_shutdown()`, `pjsip_transport_destroy()`, while there doesn't seem to be any bottom-up flow from `pjsip_transport` into `pjsip_tpmgr`.

By following the above order, the modifications are as follows:
* Avoid locking in `pjsip_tpmgr_destroy()` as  it doesn't seem to be much useful anyway since the worker thread should have already been stopped by that time, and we're going to destroy the lock and the pool anyway at the end of the function . Only in the beginning we try to wait for the lock, just in case that there are still other tpmgr functions still in progress.
* Swap the lock order in `destroy_transport()` (called by `pjsip_transport_destroy()`) to lock tpmgr first and then pjsip_transport.
* Use the same lock order in `pjsip_transport_shutdown()` to lock tpmgr first and then pjsip_transport. Note that this will revert the lock usage removal in #4307, and will swap the order instead. This is because it will still cause lock order inversion in SIP UDP transport:
```
Mutex M1 acquired here while holding mutex M0 in main thread:
    #2 pj_lock_acquire ../src/pj/lock.c:179 (pjsua-x86_64-pc-linux-gnu+0x414a54) (BuildId: 713254ae3b173bf4d5c9cc4ec8d848fc756f6725)
    #3 pjsip_transport_dec_ref ../src/pjsip/sip_transport.c:1268 (pjsua-x86_64-pc-linux-gnu+0x165060) (BuildId: 713254ae3b173bf4d5c9cc4ec8d848fc756f6725)
    #4 udp_shutdown ../src/pjsip/sip_transport_udp.c:523 (pjsua-x86_64-pc-linux-gnu+0x16bec0) (BuildId: 713254ae3b173bf4d5c9cc4ec8d848fc756f6725)
    #5 pjsip_transport_shutdown2 ../src/pjsip/sip_transport.c:1534 (pjsua-x86_64-pc-linux-gnu+0x165ca3) (BuildId: 713254ae3b173bf4d5c9cc4ec8d848fc756f6725)
```

### DIFF
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -2015,8 +2015,19 @@ PJ_DEF(pj_status_t) pjsip_tpmgr_destroy( pjsip_tpmgr *mgr )
          itr = pj_hash_first(mgr->table, &itr_val))
     {
         transport *tp_ref;
+        pjsip_transport *tp;
+
         tp_ref = pj_hash_this(mgr->table, itr);
-        destroy_transport(mgr, tp_ref->tp);
+        tp = tp_ref->tp;
+
+        /* We need to release first the transport manager's lock here.
+         * Later in the beginning of destroy_transport(), we will reacquire
+         * the lock, but only after acquiring the transport lock,
+         * to prevent lock ordering violation and avoid deadlock.
+         */
+        pj_lock_release(mgr->lock);
+        destroy_transport(mgr, tp);
+        pj_lock_acquire(mgr->lock);
     }
 
     /*
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -2003,11 +2003,7 @@ PJ_DEF(pj_status_t) pjsip_tpmgr_destroy( pjsip_tpmgr *mgr )
 
     PJ_LOG(5, (THIS_FILE, "Destroying transport manager"));
 
-    /* Last chance to wait for other tpmgr functions to finish before
-     * we destroy everything.
-     */
     pj_lock_acquire(mgr->lock);
-    pj_lock_release(mgr->lock);
 
     /*
      * Destroy all transports in the hash table.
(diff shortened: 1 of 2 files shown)


jcolp:
@sauwming Greetings! Is this a regression, or has it always existed?

sauwming:
Hi @jcolp Yes, it's always been there, but as far as we know, it should be safe from deadlocking since the inversion occurs during destruction when worker thread has stopped.
Still, it would be great if we can get rid of the warning.

jcolp:
Thanks for the confirmation!

nanangizz (approved):

PR #4765, opened 2026-01-30
Prevent INVITE retransmissions from being treated as new requests

nanangizz:
[RFC 6026](https://www.rfc-editor.org/rfc/rfc6026.html) updates transaction handling to prevent INVITE retransmissions from being treated as new requests.

This PR modifies the INVITE session to:
- delay INVITE transaction termination to absorb retransmissions after the ACK is received
- stop retransmission of 200 responses

It also modifies the transaction to:
- handle request retransmissions (same branch) with different CSeq values

Notes:
- While waiting for the termination of the INVITE tsx, outgoing/incoming re-INVITE can still be handled properly.
- In general, implementing RFC 6026 is problematic (not only for us but in general) because it breaks compatibility with existing apps due to changes in the state machine (introduction of ACCEPTED state). Therefore we decided not to implement this RFC in full. After analyzing our compatibility with the specification, we identified that this particular behavior is the crucial part we can adopt for now.

### DIFF
--- a/pjsip/include/pjsip/sip_transaction.h
+++ b/pjsip/include/pjsip/sip_transaction.h
@@ -424,7 +424,7 @@ PJ_DECL(pj_status_t) pjsip_tsx_terminate( pjsip_transaction *tsx,
  *
  * @param tsx       The transaction.
  * @param code      The status code to report. This is applied only if
- *                  the current status code is not final.
+ *                  the current status code is not final (i.e. < 200).
  *
  * @return          PJ_SUCCESS or the appropriate error code.
  */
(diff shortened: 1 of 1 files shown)


pjsip/include/pjsip/sip_transaction.h
```diff
@@ -423,7 +423,8 @@ PJ_DECL(pj_status_t) pjsip_tsx_terminate( pjsip_transaction *tsx,
  * internal timer.
  *
  * @param tsx       The transaction.
- * @param code      The status code to report.
+ * @param code      The status code to report. This is applied only if
+ *                  the current status code is not final.
```
sauwming: add more detail about "not final". (i.e. < 200)

Issue #4774, opened 2026-02-04
rtcp_fb attribute is ignored


dmitmaloukas:
rtcp_fb attribute is ignored due a bug, the library never sends PLI request.

It happens in the function `pjmedia_stream_info_common_from_sdp`.
In the very begginnning, structure `pjmedia_stream_info_common` is filled with zeros.
```c
    /* Reset: */
    pj_bzero(si, sizeof(*si));
```

In the end of the function rtcp_fb are supposed to be added to the stream info structure.
The payload type (`si->rx_pt`) is always 0, as `pjmedia_stream_info_common` was zeroed above.
```c
    /* Get local RTCP-FB info */
    status = pjmedia_rtcp_fb_decode_sdp2(pool, endpt, NULL, local,
                                         stream_idx, si->rx_pt,
                                         &si->loc_rtcp_fb);
```

Inside function `pjmedia_rtcp_fb_decode_sdp2` there is a payload type check,
which does not allow rtcp_fb attribute to be used.
In my case, `pt == 0` and `pt_ == 105`.
        
```        
        if (pt_ == sci[j].pt && (pt < 0 || pt == (int)pt_)) {
            codec_id = sci[j].id;
            break;
        }
```

Passing negative value as payload type in `pjmedia_rtcp_fb_decode_sdp2` resolves the issue.

### PJSIP version

2.16 / master

PR #4775, opened 2026-02-04
fix parsing rtcp-fb attribute

dmitmaloukas:
Fix for the issue #4774


nanangizz:
The issue is valid and reproducible here. It appears to be a regression introduced by #4300.

The `si->rx_pt` and `si->tx_pt` values are parsed & set in `get_audio_codec_info_param()`, so the fix should move the call to `pjmedia_rtcp_fb_decode_sdp2()` to occur afterward. Attached is a draft patch, which seems to work here. If you agree, please update this PR accordingly. Thanks.

[parse-sdp-rtcp-fb.txt](https://github.com/user-attachments/files/25086895/parse-sdp-rtcp-fb.txt)

dmitmaloukas:
@nanangizz done

PR #4782, opened 2026-02-10
pcaputil: Convert rtp timestamp endianness earlier

orgads:
It was converted twice for each packet, as pkt0 and as pkt1.

Convert it once instead, and store in the packet data.

### DIFF
--- a/pjsip-apps/src/samples/pcaputil.c
+++ b/pjsip-apps/src/samples/pcaputil.c
@@ -192,6 +192,7 @@ struct rtp_packet
     pjmedia_rtp_hdr *rtp;
     pj_uint8_t      *payload;
     unsigned         payload_len;
+    pj_uint32_t      rtp_ts;
     pj_timestamp     packet_ts;
 };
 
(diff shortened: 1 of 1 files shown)


pjsip-apps/src/samples/pcaputil.c
```diff
@@ -279,6 +279,13 @@ static int read_rtp(pj_uint8_t *buf, pj_size_t bufsize,
 
 
         *rtp = (pjmedia_rtp_hdr*)r;
+#if PJ_IS_LITTLE_ENDIAN
+        /* seq and ssrc are unused here, so don't waste cpu cycles for them
+        (*rtp)->seq = ntohs((*rtp)->seq);
+        (*rtp)->ssrc = ntohl((*rtp)->ssrc);
+        */
+        (*rtp)->ts = ntohl((*rtp)->ts);
+#endif
```
nanangizz: We should keep `pjmedia_rtp_hdr` untouched and instead use a separate variable to handle the endianness normalization.

Btw, could you describe the PR a bit more, for example, what background issue you’re seeing or what kind of improvement you’re aiming for?

orgads: The `ts_gap` computation compares each packet to its immediate predecessor. As a result, every timestamp is converted twice - once when treated as the "current" packet and again when it becomes the "previous" one. This change ensures each timestamp is converted only once.

As for the header rewrite - why not modify it in place? The header is not transmitted; it is used solely as internal processing data, so an in-place update should be sufficient. We could add another field named `rtp_ts` to `pkt`, but it feels redundant.

orgads: Added a new field instead. Depends on #4794 now.

sauwming (approved):

PR #4794, opened 2026-02-12
pcaputil: Simplify read_rtp function

orgads:
Pass the packet directly instead of passing all its members.


nanangizz:
#4782 has integrated all modifications from this PR, right?

orgads:
Yes, thank you.

PR #4816, opened 2026-02-24
Asynchronous SIP client authentication

nanangizz:
### Summary

- Add async authentication API allowing applications to handle 401/407 challenges asynchronously — e.g. prompt the user for credentials, fetch from a vault, or perform OAuth token exchange — before retrying the request.
- Three-tier API: low-level PJSIP, mid-level PJSUA, and high-level PJSUA2 (C++/Java/Python via SWIG).
- Fully backward-compatible: if the callback is not set or does not handle the challenge, the library falls back to synchronous credential-based authentication.

### API Overview

| Layer | Entry Point | Async Mechanism |
|-------|-------------|-----------------|
| PJSIP | `pjsip_auth_clt_async_configure()` | Token-based callback, call `async_send_req()` or `async_abandon()` |
| PJSUA | `pjsua_callback.on_auth_challenge` | Set `param->handled = PJ_TRUE`, then call PJSIP async APIs |
| PJSUA2 | `Account::onAuthChallenge()` | Call `challenge.defer()` to get heap-allocated handle, then `respond()` or `abandon()` later |

### Changes

- **pjsip/sip_auth**: `pjsip_auth_clt_async_configure()`, `pjsip_auth_clt_async_send_req()`, `pjsip_auth_clt_async_abandon()` with per-challenge token management. Tokens allocated from `tsx->pool` with `grp_lock` lifetime to prevent use-after-free.
- **pjsip integration modules**: Async auth wired into `sip_reg.c` (REGISTER), `sip_inv.c` (INVITE/BYE/UPDATE/re-INVITE with per-site abandon callbacks), `sip_dialog.c` (in-dialog requests), `evsub.c` (SUBSCRIBE/NOTIFY), `publishc.c` (PUBLISH), and `pjsua_im.c` (MESSAGE/typing indication).
- **pjsua**: `on_auth_challenge` callback in `pjsua_callback`. Async auth auto-configured on `shared_auth_sess` for REGISTER, and wired with parent delegation (`pjsip_auth_clt_set_parent_sess`) for PUBLISH, IM, and MWI subscription dialog auth sessions.
- **pjsua2**: `AuthChallenge` class with `defer()`, `respond()`, `respond(creds)`, `abandon()`. Auto-abandon in destructor with GC-thread safety (schedules cleanup on pjlib event loop when destructor runs on unregistered thread). `OnAuthChallengeParam` + `Account::onAuthChallenge()` virtual method exposed via SWIG directors.
- **Credential propagation**: `pjsua_acc_modify()` now pushes updated credentials and auth preferences to `shared_auth_sess`, and `respond(creds)` allows per-challenge credential override.

### Sample: PJSUA (C)

```c
#include <pjsua-lib/pjsua.h>

/*
 * Callback: called when a 401/407 challenge is received.
 * Here we fetch credentials asynchronously (simulated with a timer).
 */

/* Timer callback — runs after credentials are available */
struct auth_ctx {
    pjsip_auth_clt_sess *auth_sess;
    void                *token;
    pjsip_tx_data       *tdata;
    pjsip_rx_data       *rdata_clone;
    pjsua_acc_id         acc_id;
};

static void on_creds_ready(pj_timer_heap_t *th, pj_timer_entry *te)
{
    struct auth_ctx *ctx = (struct auth_ctx *)te->user_data;
    pjsip_cred_info cred;
    pjsip_tx_data *new_req = NULL;
    pj_status_t status;

    PJ_UNUSED_ARG(th);

    /* Set credentials on the auth session */
    pj_bzero(&cred, sizeof(cred));
    cred.scheme   = pj_str("digest");
    cred.realm    = pj_str("*");
    cred.username = pj_str("alice");
    cred.data     = pj_str("secret123");
    cred.data_type = PJSIP_CRED_DATA_PLAIN_PASSWD;
    pjsip_auth_clt_set_credentials(ctx->auth_sess, 1, &cred);

    /* Build the authenticated request */
    status = pjsip_auth_clt_reinit_req(ctx->auth_sess, ctx->rdata_clone,
                                       ctx->tdata, &new_req);
    if (status == PJ_SUCCESS && new_req) {
        pjsip_auth_clt_async_send_req(ctx->auth_sess,
                                      ctx->token, new_req);
    } else {
        pjsip_auth_clt_async_abandon(ctx->auth_sess, ctx->token);
    }

    /* Free rdata clone, dec ref tdata, etc */
    free(ctx);
}

static void on_auth_challenge(pjsua_on_auth_challenge_param *prm)
{
    struct auth_ctx *ctx;
    pj_timer_entry *te;

    PJ_LOG(3, ("app", "Auth challenge for acc %d, call %d",
               prm->acc_id, prm->call_id));

    /* Allocate context for deferred handling */
    ctx = (struct auth_ctx *)malloc(sizeof(*ctx));
    ctx->auth_sess = prm->auth_sess;
    ctx->token     = prm->token;
    ctx->tdata     = prm->tdata;
    ctx->acc_id    = prm->acc_id;
    pjsip_rx_data_clone(prm->rdata, 0, &ctx->rdata_clone);
    pjsip_tx_data_add_ref(ctx->tdata);

    /* Signal we'll handle it asynchronously */
    prm->handled = PJ_TRUE;

    /* Simulate async credential fetch with a 1-second timer */
    te = (pj_timer_entry *)malloc(sizeof(*te));
    pj_timer_entry_init(te, 0, ctx, &on_creds_ready);

    {
        pj_time_val delay = {1, 0};
        pjsip_endpt_schedule_timer(pjsua_get_pjsip_endpt(), te, &delay);
    }
}

int main()
{
    pjsua_config cfg;
    pjsua_logging_config log_cfg;
    pjsua_acc_config acc_cfg;
    pjsua_acc_id acc_id;

    pjsua_create();

    pjsua_config_default(&cfg);
    cfg.cb.on_auth_challenge = &on_auth_challenge;

    pjsua_logging_config_default(&log_cfg);
    log_cfg.console_level = 4;

    pjsua_init(&cfg, &log_cfg, NULL);
    pjsua_transport_create(PJSIP_TRANSPORT_UDP, NULL, NULL);
    pjsua_start();

    /* Add account without credentials — they'll be supplied
     * dynamically in on_auth_challenge */
    pjsua_acc_config_default(&acc_cfg);
    acc_cfg.id = pj_str("sip:alice@example.com");
    acc_cfg.reg_uri = pj_str("sip:example.com");
    pjsua_acc_add(&acc_cfg, PJ_TRUE, &acc_id);

    /* ... run event loop ... */

    pjsua_destroy();
    return 0;
}
```

### Sample: PJSUA2 (C++)

```cpp
#include <pjsua2.hpp>
#include <iostream>

using namespace pj;

class MyAccount : public Account {
public:
    void onAuthChallenge(OnAuthChallengeParam &prm) override
    {
        std::cout << "Auth challenge received for acc "
                  << prm.accId << std::endl;

        /*
         * Option 1: Respond immediately with current credentials.
         * (Credentials were set via AccountConfig.sipConfig.authCreds)
         */
        // prm.challenge.respond();

        /*
         * Option 2: Respond with new credentials.
         */
        // AuthCredInfoVector creds;
        // creds.push_back(AuthCredInfo("digest", "*",
        //                              "alice", 0, "secret123"));
        // prm.challenge.respond(creds);

        /*
         * Option 3: Defer and respond asynchronously.
         * defer() returns a heap-allocated AuthChallenge* that you own.
         * You can pass it to another thread or timer, then call
         * respond() or abandon() later.
         */
        AuthChallenge *deferred = prm.challenge.defer();

        /* Simulate async: in real code, hand 'deferred' to your
         * credential-fetching logic. Here we respond inline. */
        AuthCredInfoVector creds;
        creds.push_back(AuthCredInfo("digest", "*",
                                     "alice", 0, "secret123"));
        deferred->respond(creds);
        delete deferred;
    }

    void onRegState(OnRegStateParam &prm) override
    {
        AccountInfo ai = getInfo();
        std::cout << (ai.regIsActive ? "Registered" : "Unregistered")
                  << " (" << prm.code << ")" << std::endl;
    }
};

int main()
{
    Endpoint ep;
    ep.libCreate();

    EpConfig epCfg;
    epCfg.logConfig.consoleLevel = 4;
    ep.libInit(epCfg);

    TransportConfig tcfg;
    tcfg.port = 5060;
    ep.transportCreate(PJSIP_TRANSPORT_UDP, tcfg);

    ep.libStart();

    /* Create account — no credentials needed upfront */
    AccountConfig acfg;
    acfg.idUri = "sip:alice@example.com";
    acfg.regConfig.registrarUri = "sip:example.com";

    MyAccount acc;
    acc.create(acfg);

    /* ... run event loop ... */
    pj_thread_sleep(30000);

    ep.libDestroy();
    return 0;
}
```

### Sample: PJSUA2 (Python via SWIG)

```python
import pjsua2 as pj

class MyAccount(pj.Account):
    def onAuthChallenge(self, prm):
        print(f"Auth challenge for acc {prm.accId}")

        # Defer for async handling
        deferred = prm.challenge.defer()

        # ... fetch credentials asynchronously ...

        # Respond with new credentials
        creds = pj.AuthCredInfoVector()
        creds.append(pj.AuthCredInfo("digest", "*",
                                     "alice", 0, "secret123"))
        deferred.respond(creds)

        # IMPORTANT: Always explicitly delete the deferred object when
        # done. The destructor calls pjsip APIs that require a
        # pjlib-registered thread. Python's GC may run the destructor
        # on its own finalizer thread (not registered with pjlib),
        # which will crash. See:
        # https://docs.pjsip.org/en/latest/pjsua2/general_concept.html#problems-with-garbage-collection
        del deferred

    def onRegState(self, prm):
        info = self.getInfo()
        print(f"Reg status: {prm.code}, active={info.regIsActive}")

ep = pj.Endpoint()
ep.libCreate()

ep_cfg = pj.EpConfig()
ep_cfg.logConfig.consoleLevel = 4
ep.libInit(ep_cfg)

tcfg = pj.TransportConfig()
tcfg.port = 5060
ep.transportCreate(pj.PJSIP_TRANSPORT_UDP, tcfg)
ep.libStart()

acfg = pj.AccountConfig()
acfg.idUri = "sip:alice@example.com"
acfg.regConfig.registrarUri = "sip:example.com"

acc = MyAccount()
acc.create(acfg)

import time
time.sleep(30)

ep.libDestroy()
```

### Test plan

- [x] Low-level `auth_async_test` — REGISTER with 5 scenarios: sync resend, deferred resend, double-send prevention, abandoned token, abandon API
- [x] pjsua-level `on_auth_challenge` callback: deferred respond, respond with credentials, abandon
- [x] pjsua2 `AuthChallenge`: immediate respond, `defer()` + deferred `respond()`, `respond(creds)`, `abandon()`
- [ ] Account deletion with outstanding deferred challenge — test exists but disabled due to pre-existing lock-order-inversion in `pjsua_acc_del()` unrelated to async auth
- [x] MWI subscription async auth wiring verification

### DIFF
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -2078,7 +2078,7 @@ PJ_DEF(pj_status_t) pjmedia_stream_create( pjmedia_endpt *endpt,
         ptime <<= 1;
 
         /* Allocate buffer */
-        stream->enc_buf_size = afd->clock_rate * ptime / 1000 / 1000;
+        stream->enc_buf_size = ptime / 1000 * afd->clock_rate / 1000;
         c_strm->enc_buf = (pj_int16_t*)
                           pj_pool_alloc(pool, stream->enc_buf_size * 2);
 
--- a/pjnath/src/pjnath/ice_session.c
+++ b/pjnath/src/pjnath/ice_session.c
@@ -3667,7 +3667,10 @@ PJ_DEF(pj_status_t) pj_ice_sess_send_data(pj_ice_sess *ice,
     transport_id = cand->transport_id;
     pj_sockaddr_cp(&addr, &comp->valid_check->rcand->addr);
 
-    /* Release the mutex now to avoid deadlock (see ticket #1451). */
+    /* Release the mutex now to avoid deadlock (see ticket #1451),
+     * but add ref first to avoid premature destruction in the cb.
+     */
+    pj_grp_lock_add_ref(ice->grp_lock);
     pj_grp_lock_release(ice->grp_lock);
 
     PJ_RACE_ME(5);
@@ -3676,6 +3679,8 @@ PJ_DEF(pj_status_t) pj_ice_sess_send_data(pj_ice_sess *ice,
                                   data, data_len, 
                                   &addr, 
                                   pj_sockaddr_get_len(&addr));
+
+    pj_grp_lock_dec_ref(ice->grp_lock);
 
 on_return:
     return status;
@@ -3743,7 +3748,9 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
     } else {
         /* Not a STUN packet. Call application's callback instead, but release
          * the mutex now or otherwise we may get deadlock.
+         * Add ref first to avoid race with session destruction.
          */
+        pj_grp_lock_add_ref(ice->grp_lock);
         pj_grp_lock_release(ice->grp_lock);
 
         PJ_RACE_ME(5);
@@ -3800,6 +3807,9 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
                          "component [%d] because source addr %s unrecognized "
                          "or unchecked",
                          comp_id, paddr));
+
+                pj_grp_lock_dec_ref(ice->grp_lock);
+
                 return PJ_SUCCESS;
             }
         } 
@@ -3807,6 +3817,8 @@ PJ_DEF(pj_status_t) pj_ice_sess_on_rx_pkt(pj_ice_sess *ice,
         (*ice->cb.on_rx_data)(ice, comp_id, transport_id, pkt, pkt_size, 
                               src_addr, src_addr_len);
         status = PJ_SUCCESS;
+
+        pj_grp_lock_dec_ref(ice->grp_lock);
     }
 
     return status;
--- a/pjnath/src/pjnath/stun_msg.c
+++ b/pjnath/src/pjnath/stun_msg.c
@@ -1789,7 +1789,9 @@ static pj_status_t encode_errcode_attr(const void *a, pj_uint8_t *buf,
     
     PJ_UNUSED_ARG(msghdr);
 
-    if (len < ATTR_HDR_LEN + 4 + (unsigned)ca->reason.slen) 
+    /* Calculated total attr_len (add padding if necessary) */
+    *printed = (ATTR_HDR_LEN + 4 + (unsigned)ca->reason.slen + 3) & (~3);
+    if (len < *printed)
         return PJ_ETOOSMALL;
 
     /* Copy and convert attribute to network byte order */
@@ -1802,8 +1804,15 @@ static pj_status_t encode_errcode_attr(const void *a, pj_uint8_t *buf,
     /* Copy error string */
     pj_memcpy(buf + ATTR_HDR_LEN + 4, ca->reason.ptr, ca->reason.slen);
 
+    /* Zero-fill padding bytes if reason string is not 4-bytes aligned */
+    if (ca->reason.slen & 0x03) {
+        pj_uint8_t pad[3];
+        pj_memset(pad, padding_char, sizeof(pad));
+        pj_memcpy(buf + ATTR_HDR_LEN + 4 + ca->reason.slen, pad,
+                  4 - (ca->reason.slen & 0x03));
+    }
+
     /* Done */
-    *printed = (ATTR_HDR_LEN + 4 + (unsigned)ca->reason.slen + 3) & (~3);
 
     return PJ_SUCCESS;
 }
@@ -1925,8 +1934,11 @@ static pj_status_t encode_unknown_attr(const void *a, pj_uint8_t *buf,
     
     PJ_UNUSED_ARG(msghdr);
 
+    /* Calculated total attr_len (add padding if necessary) */
+    *printed = (ATTR_HDR_LEN + (ca->attr_count << 1) + 3) & (~3);
+
     /* Check that buffer is enough */
-    if (len < ATTR_HDR_LEN + (ca->attr_count << 1))
+    if (len < *printed)
         return PJ_ETOOSMALL;
 
     PUTVAL16H(buf, 0, ca->hdr.type);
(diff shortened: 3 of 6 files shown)


pjsip/include/pjsip-ua/sip_inv.h
```diff
@@ -294,7 +294,22 @@ typedef struct pjsip_inv_callback
      *                  4xx/5xx response received for the INVITE sent to
      *                  subsequent targets, or NULL if this callback is
      *                  called from within #pjsip_inv_process_redirect()
-     *                  context.
+     *                  context, or when called via
```
sauwming: This is very long doc for a parameter. Consider moving it to the API doc above instead.
`If callback is called ..., then the event parameter e ...`

pjsip/include/pjsip/sip_auth.h
```diff
@@ -612,14 +624,191 @@ PJ_DECL(pj_status_t) pjsip_auth_srv_init( pj_pool_t *pool,
...
+ *                      The application MUST then eventually call
+ *                      pjsip_auth_clt_async_send_req() or
+ *                      pjsip_auth_clt_async_abandon(); failure to do so
+ *                      will leak the token's group lock reference.
+ *                      It is safe to call either function synchronously
+ *                      from within the callback itself.
+ *                      PJ_FALSE to let the library handle authentication
+ *                      via the synchronous path.
+ */
+typedef pj_bool_t pjsip_auth_clt_async_on_challenge(
```
sauwming, line 711: I think it's better to make the return value pj_status_t instead of pj_bool to allow more options in the future, for example PJ_ECANCELLED, PJ_EPENDING, etc.

nanangizz: Actually I was thinking the same (the core API design is by me, not AI). `pj_bool_t` emphasize who handles the challenge, while `pj_status_t` is more about how the challenge should be handled. And as it is now, I chose the former for simplicity, clarity, and lower risk. The library can just provide a fixed procedure (existing code, sync), any alternative procedures (abandon, async, etc) must be done by app itself.

Btw, I also asked the AI, it seems to agree with these reasons:

1. Simpler contract — callers return PJ_TRUE or PJ_FALSE, no ambiguity about which status codes are valid or what each one means.
2. Less error-prone — with pj_status_t, the library would need to define and document behavior for every
possible return value, and handle unexpected codes gracefully. That's added complexity for no real gain.
3. PJ_EPENDING doesn't apply — the async flow already implies pending; the callback just signals ownership
transfer.
4. PJ_ECANCELLED doesn't apply — that's what pjsip_auth_clt_async_abandon() is for, which is the explicit
cancellation mechanism.

If a future need did arise, you could always add a new field to pjsip_auth_clt_async_on_chal_param or
introduce a new callback, which is more extensible anyway.

Issue #890, opened 2009-06-18
pjlib-test failed on select() test Windows Vista multi-core


pjsipbot:
_2009-06-18 11:06:49: @bennylp_ _created the issue on trac ticket [890](https://trac.pjsip.org/repos/ticket/890)_

Error log:
```
pjlib-test error:
 19:24:28.914 Running select_test()...
 19:24:28.916 ...Testing simple UDP select()
 19:24:28.917 ..ERROR(-70)
```

The select() returns immediately with *only* write-fdset is set (while the test also expects that read-fdset is set). So perhaps it is just that the system hasn't got time/context to deliver the packet, since select(), with write-fdset included, is called directly after sendto().

Experiments:
 - inserting pj_thread_sleep(1) between sendto() & select(): OK
 - inserting pj_thread_sleep(0) between sendto() & select(): failed
 - inserting PJ_LOG() between sendto() & select(): OK

The corresponding ticket for 1.0.x is #908
___
_2009-06-18 11:14:24: @bennylp_ _changed **title** from **pjlib-test failed on Windows Vista 64bit** to **pjlib-test failed on select() test Windows Vista 64bit**_

___
_2009-06-18 19:18:38: @nanangizz_ _edited the issue description_

___
_2009-06-19 10:13:21: @bennylp_ _changed **title** from **pjlib-test failed on select() test Windows Vista 64bit** to **pjlib-test failed on select() test Windows Vista multi-core**_

___
_2009-06-19 10:13:21: @bennylp_ _commented_

I can reproduce this with Visual Studio 6 | Debug on Windows Vista 32-bit dual core. Changing the title.

___
_2009-06-19 10:21:53: @bennylp_ _commented_

As additional info, the error only happens about 50% of the time.

___
_2009-06-19 10:39:06: @bennylp_ _changed **status** from **new** to **closed**_

___
_2009-06-19 10:39:06: @bennylp_ _set **resolution** to **fixed**_

___
_2009-06-19 10:39:06: @bennylp_ _commented_

In [r2778](../commit/6a65a7da5f7922ff96cc76c109f08ed9c4fc6565):
 - Added pj_thread_sleep(10) before calling sleep(). Tried with other ways too but this seems to be appropriate for the test, as in the real application this shouldn't matter.

___
_2009-06-23 13:13:20: @bennylp_ _changed **milestone** from **release-1.3-QA** to **release-1.0.3**_

___
[...]

PR #4833, opened 2026-03-04
Fix flaky select_test on busy CI runners

nanangizz:
## Summary
- Increase pre-select sleep from 10 ms to 100 ms to give the OS more time to deliver the UDP loopback packet
- Increase select() timeout from 1 s to 5 s to tolerate slow CI runners

On heavily loaded GitHub Actions runners, the original tight margins cause spurious `-60` (`PJ_ETIMEDOUT`) failures in `select_test`. See https://github.com/pjsip/pjproject/issues/890 for prior context on this sleep.

## Test plan
- CI should no longer see intermittent `select_test` failures with error -60

### DIFF
--- a/pjlib/src/pjlib-test/select.c
+++ b/pjlib/src/pjlib-test/select.c
@@ -71,7 +71,7 @@ static int do_select( pj_sock_t sock1, pj_sock_t sock2,
         setcount[i] = 0;
     }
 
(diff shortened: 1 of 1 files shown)


nanangizz:
Oops, still failed: https://github.com/pjsip/pjproject/actions/runs/22701519107/job/65819402134

PR #4838, opened 2026-03-06
OSS-Fuzz: Add fuzzer for srtp sdes process

arthurscchan:
This PR adds a new fuzzer targeting the sdes process of the srtp protocol.


sauwming:
Conflict again after I just merged the other PR

sauwming:
Looks like the merge is not clean and causes build failure

PR #4835, opened 2026-03-05
Fix flaky CI tests: port conflicts, parallel scheduling, macOS ASan, TSan libdbus

nanangizz:
## Summary

A collection of fixes for flaky CI test failures across platforms:

- **Relax unittest parallel scheduling test in CI mode** — Only verify grouping constraints (exclusive/parallel ordering) without requiring exact completion order within parallel groups, which is unreliable on loaded CI runners
- **Fix select_test port conflict** — Use ephemeral port (bind to port 0 + getsockname) instead of hardcoded port 51232; also fix missing `pj_htons()` on port assignment
- **Fix inv_offer_answer_test port conflict** — Use ephemeral port instead of hardcoded port 50068 for the UDP transport; build contact URI dynamically from assigned port
- **Fix inv_offer_answer_test hang** — Fix mismatched contact URI user part ("inv_oa_test" vs expected "inv_offer_answer_test") causing incoming INVITE to be ignored
- **Work around ASan sigaltstack crash on macOS ARM64** — Set `ASAN_OPTIONS=use_sigaltstack=0` to avoid Apple clang ASan's failing `sigaltstack(SS_DISABLE)` call during thread exit on macOS Sequoia
- **Fix ASan longjmp crash on macOS ARM64** — Add `detect_stack_use_after_return=0` to `ASAN_OPTIONS` to prevent ASan's `PlatformUnpoisonStacks()` crash when longjmp is used in non-main threads (SIP parser error recovery)
- **Add TSan suppression for libdbus lock-order-inversion** — Suppress false positive from internal dbus mutexes during SDL video device initialization

## Changed files

- `.github/workflows/ci-mac.yml` — ASan options for macOS ARM64
- `pjlib/src/pjlib-test/select.c` — Ephemeral port for select test
- `pjlib/src/pjlib-test/unittest_test.c` — Relaxed parallel scheduling assertions in CI
- `pjsip/src/test/inv_offer_answer_test.c` — Ephemeral port + contact URI fix
- `tests/sanitizers/tsan.supp` — TSan suppression for libdbus

## Test plan

- [ ] CI jobs pass (macOS, Linux, Windows)

Co-Authored-By: Claude Code

PR #4846, opened 2026-03-10
Fix ASan asan_poisoning crash on macOS ARM64 worker threads

nanangizz:
## Summary
- Fixes the recurring `asan_poisoning.cpp` CHECK failure on macOS ARM64 CI that persists after the workaround in #4835
- Root cause: Apple Clang inserts `__asan_handle_no_return` before calls to `noreturn` functions. On worker threads, this triggers `PlatformUnpoisonStacks()` which crashes because it cannot determine the thread stack bounds (addr=0x0)
- Fix: conditionally disable `PJ_ATTR_NORETURN` (`__attribute__((noreturn))`) in `cc_clang.h` when ASan is active on Apple ARM64, preventing the compiler from inserting the problematic `__asan_handle_no_return` at `PJ_THROW` call sites

### Why `ASAN_OPTIONS=detect_stack_use_after_return=0` (PR #4835) is insufficient
That option only disables ASan's fake stack allocator. It does **not** prevent `__asan_handle_no_return` (compiler-inserted before `noreturn` calls) from invoking `PlatformUnpoisonStacks()`, which is the actual crash site.

### Crash stack trace
```
AddressSanitizer: CHECK failed: asan_poisoning.cpp:40
    "((AddrIsInMem(addr + size - (1ULL << 3)))) != (0)" (0x0, 0x0)
    #0 __asan::CheckUnwind()
    #1 __sanitizer::CheckFailed()
    #2 __asan::PoisonShadow() (.cold.4)
    #3 __asan::PoisonShadow()
    #4 __asan::PlatformUnpoisonStacks()
    #5 __asan_handle_no_return
    #6 on_syntax_error sip_parser.c:230       <-- PJ_THROW()
    #7 pj_scan_syntax_err scanner.c:46
    #8 pj_scan_peek scanner.c:225
    ...
    #16 run_test_case unittest.c:542          <-- worker thread
```

### CI failures
- https://github.com/pjsip/pjproject/actions/runs/22886918951/job/66401744251
- https://github.com/pjsip/pjproject/actions/runs/22886918951/job/66417527170

## Test plan
- [ ] macOS ARM64 CI pjsip-test passes (no more asan_poisoning crash)
- [ ] Non-ASan builds unaffected (PJ_ATTR_NORETURN still active)

Co-Authored-By: Claude Code

### DIFF
--- a/pjlib/include/pj/compat/cc_clang.h
+++ b/pjlib/include/pj/compat/cc_clang.h
@@ -48,9 +48,7 @@
   typedef long long             pj_int64_t;
   typedef unsigned long long    pj_uint64_t;
   #define PJ_INLINE_SPECIFIER   static inline
-  /* Disable noreturn when ASan is active on macOS ARM64 to avoid
-   * __asan_handle_no_return crash in PlatformUnpoisonStacks().
-   */
+  /* Apple Clang ASan crashes on longjmp/noreturn, see #4846 */
   #if defined(__has_feature) && __has_feature(address_sanitizer) && \
       defined(__APPLE__) && defined(__aarch64__)
     #define PJ_ATTR_NORETURN
--- a/pjlib/src/pjlib-test/test.h
+++ b/pjlib/src/pjlib-test/test.h
@@ -28,6 +28,17 @@
 #define GROUP_LIBC                  TEST_DEFAULT
 #define GROUP_OS                    TEST_DEFAULT
 #define GROUP_DATA_STRUCTURE        TEST_DEFAULT
+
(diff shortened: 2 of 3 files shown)


sauwming:
Previously, I have also tried various recommended solutions but none of them seems to work. That's why in the end, I disabled pjsip test msg_test here:
https://github.com/pjsip/pjproject/pull/4725#issuecomment-3717369562

So if this proves to work, we can re-enable it.

nanangizz:
Still failed. I guess we need to disable all tests involving exception (e.g: pjlib exception & pjsip msg test) on MacOS and leave them enabled on other platforms.

PR #4854, opened 2026-03-13
Fix Coverity issues from async auth feature

nanangizz:
## Summary
- Fix Coverity static analysis issues introduced by the async SIP client authentication feature (#4816)
- Zero-initialize `chal_param` struct before use across 10 call sites to silence uninitialized `user_data` field warnings
- Add return value checks for `pjsip_endpt_create()` and `pjsip_tsx_layer_init_module()` in test helper
- Fix potential null pointer dereference in `auth_async_test()` cleanup path
- Check `pjsip_rx_data_clone()` return value in `pjsua_call_on_incoming()` and reject the call with 500 on failure (previously a failed clone would silently skip the `on_incoming_call` callback)
- Use `std::move()` for `push_back` in C++ test code

### Coverity CIDs addressed

| CID | Type | Fix |
|-----|------|-----|
| 1645641, 1645645, 1645648, 1645649, 1645652, 1645655, 1645656, 1645658, 1645659 | UNINIT | `pj_bzero(&chal_param, sizeof(chal_param))` |
| 1645646, 1645653, 1645657 | CHECKED_RETURN / FORWARD_NULL | Check return values in `restore_endpt()` |
| 1645640 | REVERSE_INULL | Use `goto on_return` instead of early `return` |
| 1645647 | CHECKED_RETURN | Check `pjsip_rx_data_clone()`, reject call on failure |
| 1645651 | COPY_INSTEAD_OF_MOVE | `std::move(cred)` |

### Not addressed (false positives)

| CID | Type | Reason |
|-----|------|--------|
| 1645642, 1645643, 1645644, 1645650, 1645654 | WRAPPER_ESCAPE | `unique_ptr<TestAccount>` internal pointer escapes via `acc->create()` — intentional PJSUA2 account registration pattern in test code |

## Test plan
- [x] Build passes with zero warnings (`make -j3`)

Co-Authored-By: Claude Code

### DIFF
--- a/pjsip/src/pjsua2-test/auth_challenge.cpp
+++ b/pjsip/src/pjsua2-test/auth_challenge.cpp
@@ -217,7 +217,7 @@ static AccountConfig make_acc_config(int port, bool with_creds)
 
     if (with_creds) {
         AuthCredInfo cred("digest", "*", TEST_USER, 0, "secret");
-        cfg.sipConfig.authCreds.push_back(std::move(cred));
+        cfg.sipConfig.authCreds.push_back(PJSUA2_MOVE(cred));
     }
 
     cfg.natConfig.contactRewriteUse = 0;


pjsip/src/pjsua2-test/auth_challenge.cpp
```diff
@@ -217,7 +217,7 @@ static AccountConfig make_acc_config(int port, bool with_creds)
 
     if (with_creds) {
         AuthCredInfo cred("digest", "*", TEST_USER, 0, "secret");
-        cfg.sipConfig.authCreds.push_back(cred);
+        cfg.sipConfig.authCreds.push_back(std::move(cred));
```
wosrediinanatour: Isn't there a macro?

nanangizz: Good catch! Thanks for pointing it out.

PR #4856, opened 2026-03-13
Fix occasional crash on send manager startup

kunkku:
This PR ensures the video stream send manager's group lock has a positive reference count before the thread starts. Otherwise, an assertion may fail in `pj_grp_lock_acquire`, which is about the first thing called by the new thread.


nanangizz (approved):

sauwming (approved):
Minor, unrelated to the issue: the code seems to give the impression that attach can be called multiple times, which doesn't seem to be the case

nanangizz:
https://github.com/pjsip/pjproject/blob/7a9a81f5bfa43957b7ca6c25eff47c4287034c7c/pjmedia/src/pjmedia/vid_stream.c#L202

yes, was about to make send manager shareable among multiple vid streams, but then deferred it (can't remember why, seeing some issue perhaps).

PR #4862, opened 2026-03-17
Improve DNS packet parser robustness

nanangizz:
## Summary
- Add upper-bound validation (`PJ_DNS_MAX_SECTION_COUNT`, default 256) on DNS header section counts (`qdcount`/`anscount`/`nscount`/`arcount`) to avoid excessive pool allocation from malformed responses. The constant is configurable in `pjlib-util/include/pjlib-util/config.h`.
- Add explicit bounds check (`p < max`) in `get_name_len()` and `get_name()` while-loop conditions, making the existing safety guarantee more robust against future code changes.
- Reduce compression pointer recursion limit from 10 to 5, which is still well above the 2–3 levels seen in practice per RFC 1035.

## Test plan
- [ ] `make -j3` — builds with zero errors and zero warnings
- [ ] `make pjlib-util-test` — existing DNS parsing tests pass
- [ ] Verify crafted packet with `qdcount=65535` is rejected with `PJLIB_UTIL_EDNSINSIZE`
- [ ] Verify valid DNS responses with moderate section counts (< 256) still parse correctly

Co-Authored-By: Claude Code

### DIFF
--- a/pjlib-util/src/pjlib-util/dns.c
+++ b/pjlib-util/src/pjlib-util/dns.c
@@ -111,8 +111,13 @@ PJ_DEF(pj_status_t) pj_dns_make_query( void *packet,
 }
 
 
+/* Maximum recursion depth for DNS name compression pointers.
+ * RFC 1035 compression in practice never exceeds 2-3 levels.
+ */
+#define MAX_NAME_RECURSION  5
+
 /* Get a name length (note: name consists of multiple labels and
- * it may contain pointers when name compression is applied) 
+ * it may contain pointers when name compression is applied)
  */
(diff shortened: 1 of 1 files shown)


pjlib-util/src/pjlib-util/dns.c
```diff
@@ -122,7 +122,7 @@ static pj_status_t get_name_len(int rec_counter, const pj_uint8_t *pkt,
     pj_status_t status;
 
     /* Limit the number of recursion */
-    if (rec_counter > 10) {
+    if (rec_counter > 5) {
```
sauwming: Use local define to specify the limit

Issue #4474, opened 2025-06-24
Unable to create server subscription for BLF


alfedotov:
### Describe the bug

Error: **Unable to create server subscription: Bad Event [status=170489]**

pjsip_pres_create_uas is expecting for Subscribe with presence event, but there is dialog event in request:

### Steps to reproduce

PJSUA acting as a registration server for Dlink phone (number 6333). One of the buttons on the phone configured as BLF for number 6060.

### PJSIP version

master commit b7cfd79b1

### Context

My environment is the following:
1.  PJSUA client from master branch (commit b7cfd79b1)
2. D-link SIP phone DPH-150S

### Log, call stack, etc

```shell
16:08:50.060           pjsua_core.c  .RX 523 bytes Request msg REGISTER/cseq=1 (rdata000001EBECC78738) from UDP 192.168.100.33:5060:
REGISTER sip:192.168.100.212 SIP/2.0
Via: SIP/2.0/UDP 192.168.100.33:5060;branch=z9hG4bK2520046842027416748
From: Af <sip:6333@192.168.100.212:5060>;tag=181235747
To: Af <sip:6333@192.168.100.212:5060>
Call-ID: 32756280132733-100572225029760@192.168.100.33
CSeq: 1 REGISTER
Contact: <sip:6333@192.168.100.33:5060>
Max-Forwards: 70
Expires: 3600
Supported: path
User-Agent: DLINK DPH-150S
Allow: INVITE, ACK, OPTIONS, BYE, CANCEL, REFER, NOTIFY, INFO, PRACK, UPDATE, MESSAGE
Content-Length: 0


--end msg--
16:08:50.060           pjsua_core.c  .TX 402 bytes Response msg 200/REGISTER/cseq=1 (tdta000001EBECC5E3C0) to UDP 192.168.100.33:5060:
SIP/2.0 200 OK
Via: SIP/2.0/UDP 192.168.100.33:5060;received=192.168.100.33;branch=z9hG4bK2520046842027416748
Call-ID: 32756280132733-100572225029760@192.168.100.33
From: "Af" <sip:6333@192.168.100.212>;tag=181235747
To: "Af" <sip:6333@192.168.100.212>;tag=z9hG4bK2520046842027416748
CSeq: 1 REGISTER
Contact: <sip:6333@192.168.100.33:5060>
Server: pjsua simple registrar
Content-Length:  0


--end msg--
[...]

PR #4891, opened 2026-03-26
Support basic Busy Lamp Field (BLF) function as server (#4474)

alfedotov:


sauwming:
This doesn't seem like a complete server implementation. So perhaps you can describe the objective and scope of this PR in the PR description.

alfedotov:
Correct, this is not complete server implementation with subscribe to dialog states.

I've found that modern yealink phones has no Presence option in setup for programmable buttons, but only BLF. My PR simply adds minimal support for BLF needed for correct busy/available indication.

sauwming:
Thank you for the contribution, but at its current state, I am leaning towards rejecting this since the server support seems to be still basic, and I'm not sure how it can be useful for PJSIP users.

PR #4905, opened 2026-04-02
Make it easier to set alternative SONAME for pjproject libraries.

jkroonza:
From the contributor (@jkroonza):

> Since pjproject doesn't modify the so version on ABI changes, it's often difficult to detect breakage due to ABI changes. This makes it easier for packagers of pjproject to use an alternative SONAME resulting in distribution mechanisms like preserved-rebuild on Gentoo to maintain package consistency during upgrade processes.

A new Make variable `PJ_SONAME_VERSION` is added in `version.mak`, defaulting to `PJ_VERSION_MAJOR` (preserving current behavior). Packagers can override it at build time, e.g. `make PJ_SONAME_VERSION=2.16`, to produce distinct SONAMEs across ABI-incompatible releases. All component and third-party Makefiles are updated to use `PJ_SONAME_VERSION` in place of the previously hardcoded `PJ_VERSION_MAJOR`.


jkroonza:
Since pjproject doesn't modify the so version on ABI changes, it's often difficult to detect breakage due to ABI changes.  This makes it easier for packagers of pjproject to use an alternative SONAME resulting in distribution mechanisms like preserved-rebuild on Gentoo to maintain package consistency during upgrade processes.

sauwming (approved):

nanangizz (approved):

jkroonza:
Thanks!

nanangizz:
You're welcome, and thank you for the contribution!

PR #4969, opened 2026-05-06
OSS-Fuzz: Fix fuzz_sdp to include sdp negotiation

arthurscchan:
This PR fixes the fuzz_sdp fuzzer to cover the sdp neogtiation path.

PR #4971, opened 2026-05-06
sdp-negotiation: Add strict checking for invalid sdp message

arthurscchan:
The PR adds strict validation in pjmedia_sdp_neg_negotiate before create_answer is called, so any SDP that fails the stricter check is rejected cleanly as a negotiation error rather than hitting the assertion, which is an no-op in production.

This is discovered by the fuzz_sdp after fixes in #4969.

### DIFF
--- a/pjmedia/src/pjmedia/sdp_neg.c
+++ b/pjmedia/src/pjmedia/sdp_neg.c
@@ -1697,10 +1697,11 @@ static pj_status_t create_answer( pj_pool_t *pool,
     char media_used[PJMEDIA_MAX_SDP_MEDIA];
     unsigned i;
 
-    /* Validate remote offer. 
+    /* Validate remote offer.
      * This should have been validated before.
+     * Use lenient validation (allow missing c= for disabled media).
      */
-    status = pjmedia_sdp_validate(offer);
+    status = pjmedia_sdp_validate2(offer, PJ_FALSE);
     PJ_ASSERT_RETURN(status==PJ_SUCCESS, status);
 
     /* Create initial answer by duplicating initial SDP,
--- a/pjmedia/src/test/sdp_neg_test.c
+++ b/pjmedia/src/test/sdp_neg_test.c
@@ -1736,7 +1736,7 @@ static int sdp_neg_strict_validate_test(pj_pool_t *pool)
         return -1020;
 
     status = pjmedia_sdp_neg_negotiate(pool, neg, 0);
-    if (status == PJ_SUCCESS)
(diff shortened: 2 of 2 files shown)


pjmedia/src/pjmedia/sdp_neg.c
```diff
@@ -2146,10 +2146,14 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_negotiate( pj_pool_t *pool,
     } else {
         pjmedia_sdp_session *answer = NULL;
 
-        status = create_answer(pool, neg->prefer_remote_codec_order,
-                               neg->answer_with_multiple_codecs,
-                               neg->neg_local_sdp, neg->neg_remote_sdp,
-                               &answer);
+        status = pjmedia_sdp_validate(neg->neg_remote_sdp);
+        if (status == PJ_SUCCESS) {
+            status = create_answer(pool, neg->prefer_remote_codec_order,
```
nanangizz: There is already remote offer validation in `pjmedia_sdp_neg_create_w_remote_offer()`, lenient one (allowing no c=line for disabled media, using `pjmedia_sdp_validate2(sdp, PJ_FALSE)`). Then another validation in `create_answer()`, a strict one (must have c= line). Rather than adding another validation here, perhaps better to modify the validation in `create_answer()` to the lenient version?

pjmedia/src/test/sdp_neg_test.c
```diff
@@ -1701,6 +1701,50 @@ static int sdp_neg_static_pt_repr_test(pj_pool_t *pool)
...
+    if (pjmedia_sdp_parse(pool, b1, pj_ansi_strlen(b1), &offer) != PJ_SUCCESS)
+        return -1000;
+    if (pjmedia_sdp_parse(pool, b2, pj_ansi_strlen(b2), &local) != PJ_SUCCESS)
+        return -1010;
+    if (pjmedia_sdp_neg_create_w_remote_offer(pool, local, offer, &neg) != PJ_SUCCESS)
+        return -1020;
+
+    status = pjmedia_sdp_neg_negotiate(pool, neg, 0);
+    if (status == PJ_SUCCESS)
+        return -1030;
```
nanangizz: Perhaps this should be (`status != PJ_SUCCESS`) instead?

arthurscchan: Yes, indeed, sorry for the overlook.

sauwming (approved):

nanangizz (approved):

PR #4977, opened 2026-05-11
Account-scoped server affinity: UDP destination-pinning (#4964)

### LINKED ISSUE #4964: Proposal: account-scoped server affinity for outbound SIP
## Motivation

PJSIP re-resolves the destination on every outgoing request and re-elects a server on each resolution call. Two sequential requests from the same account against a DNS/SRV-balanced cluster can land on different backends — even when the DNS records are TTL-cached and the previously chosen server is still healthy.

The flip-flop comes from RFC 2782 weighted-random selection (`pjlib-util/src/pjlib-util/srv_resolver.c`): it runs fresh for every resolution call, picking a new entry from the same-priority group based on `pj_rand()`. The DNS layer caches *records*; the *picked entry* is rolled per resolve.

For a SIP account this is usually unwanted:

- **Session anchoring** — re-REGISTER, in-dialog INVITE, and follow-up MESSAGE/SUBSCRIBE benefit from landing on the same server that already holds prior context (registration binding, auth nonce, dialog state).
- **Connection reuse** — a different picked entry forces a new TCP/TLS connection even though the prior one is healthy.
- **TLS hostname check edge case** — proxy reached by IP literal, registrar by FQDN: today's CVE-2020-15260 reuse check (`pj_stricmp("198.51.100.10", "server.example.com")`) fails, a second handshake opens, also fails. PR #4954 patches this narrow case via peer-cert SAN inspection; account-scoped pinning removes the need entirely because trust is asserted once at handshake, not per request.

The common ask underneath all three: **"keep me on the same server as long as it's still in the DNS result set."**

## Proposal

Pin an account's **next-hop server** — `(resolved address, transport)` — when the account first registers, and reuse it for subsequent same-account requests via `PJSIP_TPSELECTOR_TRANSPORT`. Apps that don't register can pin explicitly via API.

The transport manager already honors explicit transport selection (`sip_transport.c:2531-2565`) by short-circuiting hash lookup, the destination match, and the CVE-2020-15260 hostname check — so no changes to pjsip core.

Applies to TLS and TCP in this version (pins the specific connected transport, avoiding redundant connection setup). UDP destination-pinning is tracked separately — see "Current PR scope vs deferred" below.

For TLS specifically: trust is established once at handshake (peer cert validated against the next-hop hostname per existing rules). After that, the account asserts ownership of the connection — see Trust model section.

## Current PR scope vs deferred

This is the scope split for the initial implementation in #4966:

**Landed in this PR:**
- TCP/TLS pinning end-to-end via `PJSIP_TPSELECTOR_TRANSPORT` — solves the SRV flip-flop and TLS connection-coalescing cases (the proposal's main motivations).
- Auto-capture from `regc_cb` on successful REGISTER.
- Explicit pin via `pjsua_acc_set_affinity_addr` (eager; uses next-hop URI hostname for SNI / cert validation on TLS).
- Lifecycle: clear on transport disconnect, IP change, `acc_modify` next-hop change, account destroy, and explicit `pjsua_acc_refresh_transport`.
- `--server-affinity[=on|strict|off]` CLI option.
- Basic E2E test (`tests/pjsua/scripts-run/210_server_affinity.py`).

**Deferred to follow-up commits / PRs:**
- **UDP destination-pinning.** For UDP the address is stored but the per-request destination is not yet constrained — selector forces use of the listener but `dst_addr` still comes from URI resolution. Likely implementation: route_set rewrite (replace next-hop URI host with the resolved IP) or hidden-proxy injection (`;lr;hide` Route entry). Both have lifecycle and TLS-cert-on-reconnect wrinkles that warrant their own commit.
- **DNS-driven auto-refresh.** Currently the pin holds until disconnect / IP change / explicit refresh — effectively `strict` is the default behavior. Implementing "verify cached addr against the resolved set on each outgoing request, refresh if missing" needs hooking into the resolver-callback paths in `sip_util.c`; will land alongside UDP destination-pinning since both share that hook.
- **SRV flip-flop integration test** (mock DNS + two server backends + assertion that two REGISTERs land on the same backend) — needs DNS-mocking infrastructure on top of the existing pjsua test framework.

## Why account scope is the right boundary

- The next-hop URI is configured per account (`reg_uri`, `proxy[0]`); affinity tracking the account boundary aligns with how the user already organizes routing.
- Same-account reuse is what TLS connection coalescing is for (RFC 5626 outbound, HTTP/2 connection coalescing follow this model).
- No cross-account leakage: each account maintains its own affinity independently. Multi-tenant TLS servers don't get aliased across accounts.
- No request-time cert SAN inspection: trust is at handshake, not per request — fewer security knobs, smaller attack surface.

## Account next-hop resolution

The affinity attaches to the account's effective next-hop URI, in this priority:

1. `acc_config.proxy[0]` — when set, the connected destination of REGISTER is `proxy[0]` (with `reg_uri` as Request-URI and `proxy[]` in the Route header per loose routing).
2. `acc_config.reg_uri` — when no proxy is configured.

The library doesn't choose between these; pjsip's existing routing already does. The affinity captures whichever URI REGISTER actually used.

For accounts without registration, the explicit pin API (`pjsua_acc_set_affinity_addr`) covers the same need without auto-capture.

## DNS re-resolution & auto-refresh

The affinity is not blindly sticky. On each outgoing request (or other re-validation point), resolve the next-hop URI again. The pjlib-util resolver caches by TTL (`pjlib-util/resolver.c:900`), so repeated resolution within TTL returns from cache — no DNS round-trip cost. Resolution returns multiple addresses (SRV+A/AAAA results); RFC 2782 selection swaps the elected entry to position 0 but leaves the others in the list, so verification compares against **all** entries in `pjsip_server_addresses.entry[]`.

- **Cached address still in the resolved set** → keep the affinity, reuse continues, ongoing dialogs unaffected.
- **Cached address no longer in the resolved set** (default behavior) → server-side has dropped this endpoint. Auto-refresh: dec_ref the old transport, pick a fresh address from the resolved set, establish a new transport.
- **Strict mode** → cached address stays even if DNS drops it. App manages refresh entirely via the explicit API.

This balances stability (don't churn the affinity while the server still says the address is good) against staleness (don't ride a dead address forever).

## Per-account state

```c
struct pjsua_acc {
    ...
    /* Account-scoped server affinity */
    pj_sockaddr      sa_next_hop_addr;   /* cached resolved address */
    pjsip_transport *sa_next_hop_tp;     /* ref'd transport (NULL until first send after explicit pin) */
    pj_bool_t        sa_enabled;         /* effective config */
    pj_bool_t        sa_strict;          /* effective config: no auto-refresh */
};
```

Next-hop-URI change detection in `acc_modify` works directly off existing config fields (`proxy[]` CRC, `reg_uri` compare, `transport_id`); no separately stored URI needed.

## Lifecycle

| Trigger | Effect |
|---|---|
| REGISTER (initial or refresh) | **Set** if empty; otherwise **use+verify** — re-resolve URI, check `sa_next_hop_addr` in result, reuse or auto-refresh |
| Subsequent outgoing request | **Use** — verify cached addr in resolved set; pass `PJSIP_TPSELECTOR_TRANSPORT` if matched |
| `pjsua_acc_set_affinity_addr(addr)` | **Set explicitly** — store addr, tp filled on first send; auto-refresh applies same as auto-pin (unless strict) |
| `pjsua_acc_refresh_transport()` | **Clear** — next REGISTER (or explicit set) re-pins |
| Transport disconnect (state cb) | **Clear** (matches existing `via_tp` clearing in `pjsua_acc.c:4569-4576`) |
| `pjsua_handle_ip_change()` | **Clear** (via disconnect path) |
| `pjsua_acc_modify` — next-hop URI **unchanged** | **No-op** (preserve affinity) |
| `pjsua_acc_modify` — `proxy[0]`/`reg_uri`/`transport_id` changed | **Clear** automatically |
| Account destroy | dec_ref tp, clear |

REGISTER capture is empty-only: explicit pins set by `pjsua_acc_set_affinity_addr` aren't overwritten by subsequent REGISTERs (they can still be cleared by disconnect, ip_change, or `pjsua_acc_refresh_transport`).

Disconnect and IP-change clear the affinity state and let the existing re-registration flow (`schedule_reregistration` at `pjsua_acc.c:4621`) re-establish it, matching today's `via_tp` lifecycle.

Existing dialogs hold their own transport refs — refreshing the account affinity doesn't touch them. In-flight calls always survive a refresh.

## API additions

```c
/**
 * Discard the account's cached next-hop affinity (address, transport).
 * The next REGISTER from this account will perform fresh resolution
 * and may pin to a different server.
 *
 * Existing dialogs/calls keep their own transport references and are
 * not affected.
 */
PJ_DECL(pj_status_t) pjsua_acc_refresh_transport(pjsua_acc_id acc_id);

/**
 * Pin the account's affinity to a specific remote address. Subsequent
 * outgoing requests resolve the account's next-hop URI, verify this
 * address is still in the result set, and reuse the matching transport
 * (created on first send if no existing one matches). Auto-refresh
 * applies the same as auto-pin unless the account is in strict mode.
 *
 * For accounts that don't register, this is the way to enable affinity.
 * For registering accounts, this overrides the address that REGISTER
 * would otherwise pick.
 */
PJ_DECL(pj_status_t) pjsua_acc_set_affinity_addr(pjsua_acc_id acc_id,
                                                 const pj_sockaddr *addr);
```

Typical app triggers for `pjsua_acc_refresh_transport()`:
- Network handover detected (Wi-Fi → cellular, IP change) — though `pjsua_handle_ip_change()` already covers the common case
- Long idle followed by activity, suspecting a half-open connection
- 503 Retry-After from registrar (server signalling rediscovery)
- App-level health check fails

The library doesn't try to detect these — app-policy territory. Specifically: **503 with `Retry-After` is NOT auto-handled**; apps wanting that behavior should call `pjsua_acc_refresh_transport()` from `on_reg_state` / `on_reg_state2`.

## Configurability

Three layers, increasing precedence:

1. **Compile-time** — `PJSUA_ACC_SERVER_AFFINITY_DEFAULT` in `config_site.h`, default `0`.
2. **Runtime global** — `pjsua_config.acc_server_affinity_default`, sets default for new accounts. Falls through to #1.
3. **Per-account** — `pjsua_acc_config.server_affinity` and `server_affinity_strict`, tristate (`UNSPECIFIED / TRUE / FALSE`) so `acc_modify` can leave inheritance intact. Same pattern as `ipv6_sip_use`.

**Default: off.** Preserves existing strict CVE-2020-15260 behavior. Opting in is an explicit acknowledgment of the trust model below.

When opted in, **`server_affinity_strict` defaults off** (auto-refresh on DNS change). Strict mode is for advanced users who want absolute pin stability.

## Trust model (TLS-specific consequence)

When affinity is enabled for a TLS account, the security posture changes — documented in cfg doxygen:

> "When enabled for a TLS account, the peer certificate is validated at the TLS handshake performed for the first outgoing request. Subsequent same-account requests are sent over the cached connection without re-validating cert subject/SAN against destination identity. Multi-tenant TLS servers that host unrelated security domains on the same connection require those domains to use separate accounts."
>
> "The affinity persists across re-REGISTERs as long as the cached address is still returned by DNS resolution. Set `server_affinity_strict` to disable automatic refresh on DNS change. Use `pjsua_acc_refresh_transport()` to invalidate explicitly. The affinity is automatically invalidated on transport disconnect, IP change, or when account configuration changes the next hop."

For TCP/UDP, the feature is purely a connection/DNS-cache optimization — same trust posture as today.

## Resolver interaction

Resolution still runs on each outgoing request; the pjlib-util resolver's TTL cache (`resolver.c:900`) makes this cheap for repeated lookups. The affinity contributes:

1. **Verification** — confirm cached `sa_next_hop_addr` is still present in `pjsip_server_addresses.entry[]`; either keep, auto-refresh, or hold (strict mode).
2. **Selector injection** — when address is verified, set `tp_sel.type = PJSIP_TPSELECTOR_TRANSPORT, tp_sel.u.transport = sa_next_hop_tp`. Transport mgr short-circuits at `sip_transport.c:2531`.
3. **Defensive shutdown check** — before injecting, verify `!sa_next_hop_tp->is_shutdown`; treat shutdown as a clear-and-refresh signal, since `acquire_transport2`'s short-circuit checks `is_destroying` but not `is_shutdown`.

A "skip resolver entirely" optimization is possible but unnecessary given the TTL cache.

## Compatibility / rollout

- No public API removed. `pjsua_acc_refresh_transport()`, `pjsua_acc_set_affinity_addr()`, and two config fields added.
- `pjsua_init_tpselector` (`pjsua_core.c:3386`) gains the affinity path; **affinity is skipped entirely when `cfg.transport_id` is set** (`transport_id` already expresses pinning, no need for two layers). Existing `transport_id` users unaffected.
- `acc->ka_transport` mechanism unchanged — UDP keep-alive keeps using its own state.
- `acc->via_tp` mechanism unchanged — Via tracking is a separate concern; both states clear on the same disconnect callback.
- PR #4954 becomes unnecessary; can be closed once this lands. The cert-SAN bypass and the new `pjsip_tls_transport_get_ssl_info` public API are not part of this proposal.

## Test coverage

- **Integration: SRV flip-flop** (motivation test) — `_sip._tcp.example.com` returns two equal-weight A entries; two REGISTERs in a row must land on the same backend (without affinity, `pj_rand()` randomizes; with affinity, both go to the same).
- Unit: affinity set on first REGISTER, reused on follow-up INVITE/MESSAGE, cleared on `pjsua_acc_refresh_transport()`.
- Unit: re-resolve with cached addr in result → reuse continues.
- Unit: re-resolve with cached addr gone → auto-refresh (and conversely, strict mode → no auto-refresh).
- Unit: explicit pin via `pjsua_acc_set_affinity_addr()` — survives REGISTER (not overwritten); transport created on first send.
- Integration: IP-proxy + FQDN-registrar scenario — single TLS connection across REGISTER + INVITE + BYE.
- Integration: `pjsua_handle_ip_change` clears affinity along with `via_tp`; re-registration repopulates.
- Negative: `acc_modify` changes `proxy[0]`/`reg_uri`/`transport_id` → next request opens new connection.
- Negative: `acc_modify` changes unrelated fields (display name, codec list) → affinity preserved.
- Negative: transport disconnect → next request resolves+reconnects without app intervention.
- Negative: cached transport `is_shutdown==TRUE` → defensive check fires; affinity cleared; new transport opened.
- Stability: in-flight call survives `pjsua_acc_refresh_transport()` and explicit-pin.

## Resolved decisions

- **Naming**: `server_affinity` (`server_affinity_strict` for the no-auto-refresh mode).
- **First-pin trigger**: REGISTER (initial or refresh) for auto-capture; `pjsua_acc_set_affinity_addr()` for explicit. Use applies to all outgoing requests on the account.
- **Auto-refresh default**: on (refresh when cached addr no longer in DNS result). Strict mode opt-in.
- **Resolution cadence**: rely on pjlib-util resolver's TTL cache; resolve on each outgoing request, no separate timer.
- **Disconnect/IP change**: clear affinity, matching existing `via_tp` behavior.
- **`acc_modify`**: clear only when next-hop URI (`proxy[0]` / `reg_uri` / `transport_id`) changes.
- **`transport_id` precedence**: when set, affinity is skipped entirely.

nanangizz:
## Summary

Follow-up to #4966 (TCP/TLS account-scoped affinity). Pins the UDP destination per-account so REGISTERs and subsequent same-account requests stay on the same backend across SRV/DNS reroll, matching the TCP/TLS behavior shipped earlier.

Issue: https://github.com/pjsip/pjproject/issues/4964

## Why a different mechanism for UDP

TCP/TLS pins via the transport selector (`PJSIP_TPSELECTOR_TRANSPORT`) — each connection is its own destination. UDP can't use that because the UDP listener is shared across all peers. This PR pins UDP destination by injecting a hidden Route header `<sip:host:port;lr;hide>` at the head of `acc->route_set`. The `;hide` URI param suppresses the Route from the wire (handled in `pjsip_routing_hdr_print`), but loose-routing still uses it as the request destination — so REGISTER/INVITE go to the pinned address even though the listener is shared.

## How

- **Sync helper**: `sa_sync_route_set(acc)` is the single point that adds/removes the hidden Route in `acc->route_set` based on the current pin state (UDP-only — TCP/TLS continue to use tp_sel), and pushes the result to regc only if it differs from a mirror of the last push (`sa_pushed_route`). Force-pushes when the hidden hdr was rebuilt for a new address, since `pjsip_hdr_cmp` would otherwise falsely match two different hidden Routes by their empty printed form.
- **Lifecycle**: detach/reattach the hidden Route around `acc_modify`'s route_set rebuild, `update_service_route`'s strip-tail, and `pjsua_regc_init`'s direct push (mirror seeded after to keep the diff invariant).
- **Pin recovery**: `auto_rereg_timer_cb` drops auto-captured pins just before the retry REGISTER fires, so fresh resolution can pick a different address from the resolved set. This fills the graceful-migration gap for UDP (no transport disconnect signal). Explicit pins set via `pjsua_acc_set_affinity_addr` are preserved across retries (tracked by `sa_pin_explicit`).
- **Mirror reset on regc destroy**: `destroy_regc` resets `sa_pushed_route` so the post-recreate push diffs against an empty mirror and re-pushes the hidden Route correctly.

## Behavior changes / caveats

- DNS auto-refresh was considered and dropped in favor of the auto_rereg pin reset — same calculus as the original `server_affinity_strict` drop: the actual graceful-decomm trigger is when REGISTER starts failing, and the retry path is exactly where we'd want to re-resolve.
- `reg_use_proxy = 0` limitation: when REGISTER is configured to bypass both outbound and account proxies, enabling UDP affinity may still push those proxy entries to regc alongside the hidden Route. Niche config — documented in the `server_affinity` doxygen with a recommendation to use the default `PJSUA_REG_USE_ALL_PROXY` with UDP affinity.

## Test plan

- [x] Build clean (`make -C pjsip/build`, `-Wall` no warnings)
- [x] `tests/pjsua/scripts-run/210_server_affinity.py` passes (basic E2E with simple_registrar — verifies the auto-capture log line fires)
- [x] `tests/pjsua/scripts-run/100_simple.py` passes (smoke)
- [ ] SRV flip-flop assertion test — needs mock-DNS infrastructure on top of pjsua test framework, tracked as a separate follow-up

## Review history

This PR went through two adversarial review passes that surfaced 6 real bugs and 1 latent invariant, all fixed across the commit history:
- Detach/reattach gaps in `acc_modify` early-exit and `update_service_route` (commit `0c777934b`)
- Mirror divergence after Service-Route changes (commit `0c777934b`)
- `pjsua_regc_init` direct push leaving mirror stale (commit `0c777934b`)
- `auto_rereg` dropping explicit (API-set) pins (commit `e6052243a`)
- `pjsip_hdr_cmp` empty-print false-match on hidden Routes (commit `e6052243a` — latent today, defended via force-push on rebuild)

Co-Authored-By: Claude Code

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -574,6 +574,7 @@ static pj_bool_t update_hdr_list(pj_pool_t *pool, pjsip_hdr *dst,
 static void sa_sync_route_set(pjsua_acc *acc)
 {
     pj_bool_t need_route;
+    pj_bool_t hdr_rebuilt = PJ_FALSE;
 
     if (acc->sa_route_hdr)
         pj_list_erase(acc->sa_route_hdr);
@@ -619,6 +620,7 @@ static void sa_sync_route_set(pjsua_acc *acc)
                     acc->sa_route_hdr = hdr;
                     pj_sockaddr_cp(&acc->sa_route_hdr_addr,
                                    &acc->sa_next_hop_addr);
+                    hdr_rebuilt = PJ_TRUE;
                 } else {
                     PJ_LOG(2,(THIS_FILE, "Account %d: failed to parse "
                                          "hidden Route '%.*s' (#4964)",
@@ -632,17 +634,20 @@ static void sa_sync_route_set(pjsua_acc *acc)
             pj_list_push_front(&acc->route_set, acc->sa_route_hdr);
     }
 
-    /* Push to regc only if it actually differs from what we pushed
-     * last time. update_hdr_list both detects the diff and keeps
-     * acc->sa_pushed_route in sync as our mirror of regc's view.
-     * Skipping avoids the regc->pool clone that
-     * pjsip_regc_set_route_set does unconditionally.
+    /* Push to regc when the route_set diffs from sa_pushed_route, OR
+     * when we rebuilt the hidden hdr. The rebuild case is force-pushed
+     * because pjsip_hdr_cmp (used by update_hdr_list) compares headers
+     * by printed form, and ;hide entries print empty — so a stale
+     * hidden Route clone in the mirror would falsely match a fresh
+     * hidden Route built for a different address. (#4964)
      */
-    if (acc->regc &&
-        update_hdr_list(acc->pool, (pjsip_hdr*)&acc->sa_pushed_route,
-                        (const pjsip_hdr*)&acc->route_set))
-    {
-        pjsip_regc_set_route_set(acc->regc, &acc->route_set);
+    if (acc->regc) {
+        pj_bool_t changed = update_hdr_list(
+                                acc->pool,
+                                (pjsip_hdr*)&acc->sa_pushed_route,
+                                (const pjsip_hdr*)&acc->route_set);
+        if (changed || hdr_rebuilt)
+            pjsip_regc_set_route_set(acc->regc, &acc->route_set);
     }
 }
 
@@ -655,6 +660,7 @@ static void clear_sa_pin(pjsua_acc *acc)
         acc->sa_next_hop_tp = NULL;
     }
     pj_bzero(&acc->sa_next_hop_addr, sizeof(acc->sa_next_hop_addr));
+    acc->sa_pin_explicit = PJ_FALSE;
     sa_sync_route_set(acc);
 }
 
@@ -2957,6 +2963,7 @@ static void regc_cb(struct pjsip_regc_cbparam *param)
                         }
                         acc->sa_next_hop_tp = param->rdata->tp_info.transport;
                         pjsip_transport_add_ref(acc->sa_next_hop_tp);
+                        acc->sa_pin_explicit = PJ_FALSE;
                         /* Mirror the pin into route_set (UDP only —
                          * for TCP/TLS tp_sel covers destination).
                          */
@@ -4769,6 +4776,7 @@ PJ_DEF(pj_status_t) pjsua_acc_set_affinity_addr(pjsua_acc_id acc_id,
         clear_sa_pin(acc);
         pj_memcpy(&acc->sa_next_hop_addr, addr, addr_len);
         acc->sa_next_hop_tp = tp;
+        acc->sa_pin_explicit = PJ_TRUE;
         /* Inject hidden Route for UDP destination-pinning. */
         sa_sync_route_set(acc);
         PJ_LOG(3,(THIS_FILE,
@@ -4853,12 +4861,12 @@ static void auto_rereg_timer_cb(pj_timer_heap_t *th, pj_timer_entry *te)
         pj_pool_release(pool);
     }
 
-    /* Drop affinity pin before retrying — if the pinned server is the
-     * one that misbehaved, fresh resolution gives the retry a chance
-     * at a different address from the resolved set. Re-pin happens
-     * naturally via regc_cb capture on a successful retry. (#4964)
+    /* Drop auto-captured affinity pin before retrying — if the pinned
+     * server is the one that misbehaved, fresh resolution gives the
+     * retry a chance at a different address. Explicit (API-set) pins
+     * are preserved per the set_affinity_addr contract. (#4964)
      */
-    if (acc->sa_enabled && acc->sa_next_hop_tp) {
+    if (acc->sa_enabled && acc->sa_next_hop_tp && !acc->sa_pin_explicit) {
         PJ_LOG(4,(THIS_FILE,
                   "Account %d: dropping server affinity pin before "
                   "re-registration retry", acc->index));
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -301,6 +301,10 @@ static pj_status_t destroy_regc(pjsua_acc *acc, pj_bool_t force)
     acc->reg_mapped_addr.slen = 0;
     acc->rfc5626_status = OUTBOUND_UNKNOWN;
     acc->rfc5626_flowtmr = 0;
+    /* New regc starts with an empty route_set; reset the mirror so the
+     * next sa_sync_route_set diff re-pushes the hidden Route. (#4964)
+     */
+    pj_list_init(&acc->sa_pushed_route);
 
     return PJ_SUCCESS;
 }
@@ -414,10 +418,11 @@ static pj_status_t initialize_acc(unsigned acc_id)
     //  acc_cfg->contact = acc_cfg->id;
     //}
 
-    /* Build account route-set from outbound proxies and route set from 
+    /* Build account route-set from outbound proxies and route set from
      * account configuration.
      */
     pj_list_init(&acc->route_set);
+    pj_list_init(&acc->sa_pushed_route);   /* mirror for diff (#4964) */
 
     if (!pj_list_empty(&pjsua_var.outbound_proxy)) {
         pjsip_route_hdr *r;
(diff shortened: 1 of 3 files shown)


andreas-wehrmann:
Just a suggestion, but wouldn't it be better to call the param `x-pj-hide` to make it absolutely clear this is an extension param. I think this would also make sure this hidden param never collides with anything defined by any RFC?
I believe this is also [what Asterisk does internally](https://github.com/search?q=repo%3Aasterisk%2Fasterisk%20x-ast-txp&type=code) to transport various infos through the SIP stack.

nanangizz:
The `;hide` param is already there for a long time (not introduced by this PR). I guess we'll leave it that way for now.

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -558,6 +563,127 @@ static void update_sa_effective_flags(pjsua_acc *acc)
...
+    /* Push to regc when the route_set diffs from sa_pushed_route, OR
+     * when we rebuilt the hidden hdr. The rebuild case is force-pushed
+     * because pjsip_hdr_cmp compares headers by printed form, and
+     * ;hide entries print empty — so a stale hidden Route clone in
+     * the mirror would falsely match a fresh hidden Route built for a
+     * different address. (#4964)
+     */
+    if (acc->regc) {
+        pj_bool_t changed = sa_sync_mirror(
+                                acc->pool,
```
sauwming: Using acc pool means that the memory can potentially keep growing with each route_set cloning. Previously, we use regc's pool to store the route.

nanangizz: Addressed in 78c5fa896: introduced a dedicated `sa_mirror_pool` sub-pool for the mirror clones, reset on every rebuild and on `destroy_regc`. Couldn't use `regc->pool` directly because resetting it would corrupt regc's own internal allocations.

sauwming (approved):

PR #4985, opened 2026-05-16
Implementation_of_replace_port_to_support_conf_async_port

PesalaDeSilva:
Signed-off-by: Pesala Silva <pesala.silva@ifs.com>

## Summary

Add optional early conference slot reservation for PJSUA calls (null-port
placeholder) and implement `pjmedia_conf_replace_port` to swap that slot to
the real stream after SDP/stream setup, without performing the heavy work on
arbitrary threads. Queued operations run on the conference media tick; old
port destruction is deferred for safety.

## Motivation

Some integrations need a stable `conf_slot` ID while the call is still being
set up (e.g. before media is fully active). PJSIP 2.15-era behaviour can defer
slot assignment until channel update; this patch preserves an early slot and
replaces the port in place when audio is ready.

## Changes (high level)

- pjmedia: `pjmedia_conf_replace_port` + implementation, op queue, deferred
  destroy list; small hardening in conf get/put frame paths.
- pjsua: `pjsua_call_preallocate_conf_port()`; `pjsua_aud_channel_update`
  uses replace when a placeholder slot exists, with add/remove fallback.


sauwming:
@PesalaDeSilva , we have replied via email for discussions and feedback regarding this PR, so please check it as well

PesalaDeSilva:
Hi @sauwming, 
thank you for the feedback. I have checked the email and addressed all the Copilot review comments on the PR. Here is a summary of what was fixed:

**Bug fixes:**

- _c_onf->op_mutex was never created — added pj_mutex_create_recursive() for it in pjmedia_conf_create()
- conf->mutex name was incorrectly renamed to "conf_op" — reverted to "conf"; the new conf->op_mutex now correctly uses "conf_op"
- Stray #endif after put_frame() — removed
- tx_buf was allocated from caller's pool instead of slot_pool — fixed to use slot_pool
- destroy_list was allocated without pj_list_init nearby — alloc and init now happen together
- PJSUA_LOCK held across pjmedia_conf_add_port / pjmedia_conf_replace_port (3 places) — fixed with unlock/work/relock_ pattern

**Design fix:**

- _Removed the unconditional null-port pre-allocation inside pjsua_aud_channel_update — placeholder is now only created when the caller explicitly opts in via pjsua_call_preallocate_conf_port()_

**Code quality:**

- Internal types (conf_op, port_destroy_item, enum, callback typedef) moved out of the public header into conference.c
- Static function declarations removed from the public header
- Debug trace logs (pj_thread_this, LOCKED/UNLOCKED/START messages) removed
- Doxygen /** block with @param and @return added for pjsua_call_preallocate_conf_port in pjsua.h

**Style:**

- Pointer style, brace placement, // → /* */ comments, and typos ("embeded", "detatching") all corrected

**Noted / intentionally deferred:**

- The async replace_port fallback limitation: on deferred failure, on_error_unlock safely restores the null placeholder to the slot (no crash, no dangling pointer — audio is silence). Full async recovery via op->cb is acknowledged as a future improvement and documented in a code comment.

Please let me know if anything else needs to be addressed.

_I wanted to confirm — I believe I have not missed any email regarding this PR. If there was a specific email with additional feedback or discussion points that I may have overlooked, could you please resend it or summarize the key points here on the PR? I want to make sure I have addressed everything correctly._

sauwming:
Here is @nanangizz's feedback sent via email dated 26 May:
"
There seem to be several issues with the PR as described in the
copilot/AI review and some are quite major (e.g: rewriting the
async-op mechanism, exposing internal conference types in the public
header). Some approaches could also be improved, for example,
using a null-port placeholder causes unnecessary recreation of
objects (resampler, null-port itself, etc) on every renegotiation.

We understand the feature is important for your integration, and we
think it can be useful to other apps as well. So we are considering
implementing the feature ourselves. The outline:

- Configurable in account/call settings, replacing
pjsua_call_preallocate_conf_port(), so no pre-created placeholder
port is needed. Slot ID and all per-slot state (connections, mute,
adj_level, etc.) survive renegotiation. Default off, to avoid
backward-compatibility issues.
- New pjmedia_conf_detach_port() which mainly sets conf_port->port
to NULL, and pjmedia_conf_replace_port() to re-attach the slot to a
new port.
- We'll cover both the conference and parallel-conference backends,
video conference can be a follow-up.
"

So yes, there are still significant missing gaps.

PesalaDeSilva:
Hi @nanangizz, thank you very much for the detailed feedback and for taking the time to review this PR.

We fully understand and agree with the concerns raised — particularly around the async-op mechanism design and the null-port placeholder causing unnecessary object recreation on renegotiation. These are valid architectural issues that we were not fully aware of when implementing the feature.

We are very glad to hear that the PJSIP team is considering implementing this natively. The outlined approach sounds much cleaner and more robust than our current implementation:

One question, is there an estimated timeline for when this implementation might be available?

Thank you again for considering this feature. We look forward to seeing the official implementation.

nanangizz:
@PesalaDeSilva We’ve started working on this. The usual steps are investigation, design, implementation, and review. We expect it to be available within a week or two.

nanangizz:
@PesalaDeSilva Please check+test #5125, it is the initial implementation. Let us know if it aligns with your needs, and feel free to join the review.

nanangizz:
Implemented & addressed issues in #5125.

PR #4993, opened 2026-05-27
Add pjsua-stress: a multi-threaded PJSUA stress test for CI

sauwming:
# Add `pjsua-stress`: a multi-threaded PJSUA stress test for CI

## Summary

Adds a new self-contained C test app, `pjsua-stress`, that exercises the
high-level PJSUA API under sustained concurrent load. It is intended to
surface deadlocks, races, and memory issues that only appear with many
calls and many random operations happening in parallel — the kind of
bugs unit tests rarely catch.

The app brings up **one** PJSUA instance acting as both caller and
callee (loopback via UDP on `127.0.0.1`), ramps to a configurable number
of concurrent call legs (some carrying video), then runs random call
operations from worker threads for a bounded duration. A weekly GitHub
Actions workflow runs it on Linux under both AddressSanitizer and
ThreadSanitizer.

## ThreadSanitizer reports and suppression

Note that it's beyond the scope of this PR to fix all Thread Sanitizer reports.
Currently, we suppress those TSan findings in the suppression file, to be
addressed separately in separate PRs in the future.

## How it works

Three phases:

1. **Init** — `pjsua_create / init / start`, single anonymous local
   account (no registrar), null sound device, colorbar capture
   device for video, no audio I/O.
2. **Ramp** — issue `pjsua_call_make_call` repeatedly until the tracked
   call-leg count reaches `-n N`. Up to `-v V` of those legs carry
   video (audio-only otherwise).
3. **Stress** — spawn `-t T` worker threads. Each worker loops, picking
   a random tracked call and a random op until the duration expires:

   | Op           | What it does |
   |--------------|-----|
   | `hangup`     | `pjsua_call_hangup` |
   | `make_audio` | new outgoing audio call |
   | `make_video` | new outgoing video call (subject to `V` cap) |
   | `conf_conn`  | `pjsua_conf_connect` between two random tracked legs |
   | `conf_disc`  | `pjsua_conf_disconnect` between two random tracked legs |
   | `hold`       | `pjsua_call_set_hold` |
   | `unhold`     | `pjsua_call_reinvite` with `PJSUA_CALL_UNHOLD` |

4. **Cleanup** — `pjsua_call_hangup_all`, brief drain wait,
   `pjsua_destroy`.

## Usage

```
pjsua-stress [options]
  -n N            Max audio call legs   (default 64)
  -v V            Max video call legs   (default 0)
  -t T            Worker thread count   (default 4)
  --duration S    Stress phase seconds  (default 300)
  --port P        Local UDP SIP port    (default 5060)
  --self-uri URI  Outgoing dial target  (default sip:stress@127.0.0.1:<port>)
  --seed S        RNG seed              (default time-based)
  --log-level L   PJ console log level  (default 3)
  -h, --help      Show usage
```

Each call **leg** is one `pjsua_call_id`. Because the same process is
both caller and callee, every outgoing `make_call` creates two legs, so
`-n 1024` means roughly 512 concurrent two-leg conversations.

`N` is bounded by `PJSUA_MAX_CALLS` at compile time. The committed
`config_site_stress.h` raises the limits used by CI; see below.

## Compile-time configuration (`pjlib/include/pj/config_site_stress.h`)

Wraps `config_site_test.h` and bumps a few limits so the stress test
can hold up to ~1024 simultaneous call legs:

| Macro                   | Default | Stress |
|-------------------------|---------|--------|
| `PJSUA_MAX_CALLS`       |    4    |  2048  |
| `PJSUA_MAX_CONF_PORTS`  |  254    |  4096  |
| `PJ_IOQUEUE_MAX_HANDLES`|   64    |  4096  |
| `PJ_IOQUEUE_HAS_SAFE_UNREG` |  -  |  1     |
| `PJMEDIA_HAS_VIDEO`     |    0    |  1     |

The CI workflow copies this file to `config_site.h` before
`./configure`, mirroring how regular CI uses `config_site_test.h`. On
Linux the ioqueue backend is epoll, so the larger `PJ_IOQUEUE_MAX_HANDLES`
is not bounded by `FD_SETSIZE`.

## CI workflow

`.github/workflows/stress-test.yml`:

* Triggers: weekly cron (`0 3 * * 0`, Sunday 03:00 UTC) and
  `workflow_dispatch` for on-demand runs.
* Matrix: `sanitizer: [address, thread]` (no `fail-fast`, so an ASan
  failure doesn't cancel the TSan job).
* Each job: checkout → copy `config_site_stress.h` → configure with
  `-fsanitize=$matrix.sanitizer` CFLAGS/LDFLAGS → `make -j3` →
  `pjsua-stress -n 1024 -v 32 -t 8 --duration 600`.
* `LSAN_OPTIONS` / `TSAN_OPTIONS` use the existing
  `tests/sanitizers/{lsan,tsan}.supp` suppression files. `halt_on_error=1`
  so the first race/leak terminates the run rather than producing
  cascading reports.
* `stress.log` is uploaded as an artifact unconditionally
  (`if: always()`).
* Gated on `github.repository == 'pjsip/pjproject'` so forks don't
  attempt to schedule the workflow.

`.github/workflows/ci-linux.yml`:

* Since the stress test is a very useful tool that has caught nearly a dozen issues so far,
   we add two short stress-test jobs to the standard ci-linux pipeline so every PR /
   push to master exercises the high-level PJSUA API under sanitizers.
* Run on a lighter load than the weekly stress test: `pjsua-stress -n 512 -v 16 -t 4 --duration 360`

## Build system

The binary is built as part of the standard `make` in
`pjsip/build/Makefile`. A new `pjsua-stress` target follows the same
pattern as `pjsua2-test`:

* `PJSUA_STRESS_*` defines (SRCDIR / OBJS / CFLAGS / LDFLAGS / EXE).
* Added to `TARGETS_EXE`.
* `clean` / `depend` / `realclean` entries.
* Output: `pjsip/bin/pjsua-stress-<target>`.

`make pjsua-stress` builds it standalone.

## Files

| Path | Purpose |
|------|------|
| `pjsip/src/test/pjsua_stress.c` | The test app (single file, ~900 lines, C89). |
| `pjsip/build/Makefile` | `pjsua-stress` build target. |
| `pjlib/include/pj/config_site_stress.h` | Compile-time tuning for stress runs. |
| `.github/workflows/stress-test.yml` | Weekly ASan + TSan workflow. |

## Test plan

- [x] `make pjsua-stress` builds cleanly.
- [x] `pjsua-stress -h` prints usage.
- [x] Small smoke run completes cleanly: `pjsua-stress -n 4 -v 0 -t 2 --duration 5`.
- [x] Larger run with video completes: `pjsua-stress -n 8 -v 4 -t 4 --duration 15`.
- [x] CI workflow YAML parses (`gh workflow run stress-test.yml`
      should succeed after merge).

Future weekly runs surface deadlocks/races via the artifact log.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -650,6 +650,8 @@ jobs:
         cd tests/pjsua
         python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uac-custom-sdp.xml
         python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uas-custom-sdp.xml
+        python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uas-amr-sdp.xml
+        python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uas-static-pt-no-rtpmap.xml
     - name: upload artifacts on failure
       if: ${{ failure() }}
       uses: actions/upload-artifact@v4
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -719,7 +719,7 @@ jobs:
         export ASAN_OPTIONS="halt_on_error=1"
         ARCH=$(make --no-print-directory -s infotarget)
         cd pjsip/bin
-        ./pjsua-stress-$ARCH -n 256 -v 16 -t 4 --duration 300 2>&1 | tee stress.log
+        ./pjsua-stress-$ARCH -n 256 -v 16 -t 4 --duration 300 --log-level 4 2>&1 | tee stress.log
     - name: upload log
       if: always()
       uses: actions/upload-artifact@v4
@@ -754,7 +754,7 @@ jobs:
         export TSAN_OPTIONS="suppressions=$GITHUB_WORKSPACE/tests/sanitizers/tsan.supp halt_on_error=1"
         ARCH=$(make --no-print-directory -s infotarget)
         cd pjsip/bin
-        ./pjsua-stress-$ARCH -n 256 -v 16 -t 4 --duration 300 2>&1 | tee stress.log
+        ./pjsua-stress-$ARCH -n 256 -v 16 -t 4 --duration 300 --log-level 4 2>&1 | tee stress.log
         # halt_on_error doesn't always abort on lock-order-inversion warnings,
         # so explicitly fail the job if TSan printed any report.
         if grep -qE "WARNING: ThreadSanitizer|ERROR: ThreadSanitizer" stress.log; then
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -715,11 +715,25 @@ jobs:
       run: $MAKE_FAST
     - name: run pjsua-stress
       run: |
+        set -eo pipefail
         export LSAN_OPTIONS="suppressions=$GITHUB_WORKSPACE/tests/sanitizers/lsan.supp"
         export ASAN_OPTIONS="halt_on_error=1"
         ARCH=$(make --no-print-directory -s infotarget)
         cd pjsip/bin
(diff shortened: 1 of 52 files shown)


nanangizz:
I doubt the CI runner is capable to run 1024 concurrent audio calls :)
Perhaps the codec and the resampling need to be adjusted?

sauwming:
The max calls is a runtime parameter so let's adjust it to the CI capability later. 

Noted about the codec and resampling.

nanangizz (approved):

PR #4999, opened 2026-05-28
[PROPOSAL] Add support to Mac and Windows at MAUI sample

weslleymurdock:
## Description
Added support to MacCatalyst and Windows in MAUI Sample

## Motivation and Context
Right now the sample has only android & ios targets

## How Has This Been Tested?
It has tested against the sip server used in c# sample

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:


sauwming:
The CI test Windows build fails.

weslleymurdock:
I've added the sample project to sln, causing this problem. I'll remove now the push, but i'm thinking, if maybe it would be nice a slnx only for samples, in order to add a job for build the samples, and if possible, split all samples to this new *;slnx and keep only the lib in the  main ,sln

sauwming:
IMO it should be fine to create separate sln for the MAUI sample app, but you don't need to remove the other sample apps from the existing pjproject solution.

weslleymurdock:
~~Okay, i'll do that~~. There is already a sln in `pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui.sln`. I'll migrate to slnx, I still have some questions: 

1. ~~Where to put the slnx for maui sample~~ Whats the right place to this sln? 
2. Is valid add a test project, even that project be a ui automated test for maui sample?
3. If the test project creation will be valid, is the creation of a new job to build the pjsip for each platform, in order to build and test the maui sample?

sauwming:
1. The current location seems fine pjsip-apps/src/swig/csharp/pjsua2maui. 
2. Sure, you can add a test project, automated also ok. 
3. If you mean the CI build job, it's optional whether you want to include the project in the CI.

sauwming:
There's still failure in the CI build

sauwming:
This now passes the CI tests, would appreciate if you can check the review above. Thank you.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/Platforms/MacCatalyst/Info.plist
```diff
@@ -0,0 +1,76 @@
...
+	<string>LaunchScreen</string>
+	<key>CFBundleName</key>
+	<string>pjsua2maui</string>
+	<key>NSCameraUsageDescription</key>
+	<string>Camera permission required</string>
+	<key>NSMicrophoneUsageDescription</key>
+	<string>Mic permission required</string>
+	<key>NSMicrophoneUsageDescription</key>
+	<string>This app needs access to the microphone to make voip calls.</string>
+	<key>NSCameraUsageDescription </string>
```
sauwming, line 41: **Blocking: malformed XML.** `<key>NSCameraUsageDescription </string>` opens a `<key>` and closes it with `</string>`.

```
$ plutil -lint Info.plist
Info.plist: Close tag on line 41 does not match open tag key
```

Any MacCatalyst build using this file fails outright. Should be `<key>NSCameraUsageDescription</key>` — though see the duplicate-key comment below, this whole entry is redundant with line 35.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/Platforms/MacCatalyst/Info.plist
```diff
@@ -0,0 +1,76 @@
...
+<dict>
+    <!-- The Mac App Store requires you specify if the app uses encryption. -->
+    <!-- Please consult https://developer.apple.com/documentation/bundleresources/information_property_list/itsappusesnonexemptencryption -->
+    <!-- <key>ITSAppUsesNonExemptEncryption</key> -->
+    <!-- Please indicate <true/> or <false/> here. -->
+	<!-- App Sandbox must be enabled to distribute a MacCatalyst app through the Mac App Store. -->
+	<key>com.apple.security.app-sandbox</key>
+	<true/>
+	<!-- When App Sandbox is enabled, this value is required to open outgoing network connections. -->
+	<key>com.apple.security.network.client</key>
```
sauwming, line 13: `com.apple.security.app-sandbox` and `com.apple.security.network.client` are *entitlements*, not Info.plist keys. They have no effect here, and they are already correctly declared in `Platforms/MacCatalyst/Entitlements.plist`. Please drop them from Info.plist.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/Platforms/MacCatalyst/Info.plist
```diff
@@ -0,0 +1,76 @@
...
+		<string>microphone</string>
+		<string>video-camera</string>
+		<string>opengles-1</string>
+		<string>wifi</string>
+	</array>
+	<key>UIRequiredDeviceCapabilities</key>
+	<array>
+		<string>arm64</string>
+		<string>x64</string>
+	</array>
```
sauwming, line 59: `UIRequiredDeviceCapabilities` is declared twice (48 and 55). Last one wins, so the first list — `microphone`, `video-camera`, `opengles-1`, `wifi` — is silently discarded, which is presumably the opposite of what was intended.

Also `x64` is not a valid `UIRequiredDeviceCapabilities` value (the documented ones are `armv7`, `arm64`, etc.).

Same duplicate-key problem for `NSMicrophoneUsageDescription` (37 and 39) and `NSCameraUsageDescription` (35 and 41). Please keep exactly one entry per key.

pjsip-apps/build/swig_maui_pjsua2.vcxproj
```diff
@@ -0,0 +1,363 @@
...
+        %40echo  - Make sure %27swig%27 is in your PATH environment vars.
+        %40echo  - Add the dependency library locations (e.g: OpenH264) to library search paths in the project settings.
+        %40echo  - Since we are building DLL, make sure the PJSIP library is built using "dynamic" configuration, e.g: Debug-Dynamic, Release-Dynamic.
+        %40echo **************************************************************
+
+        %40echo on
+        %40echo Building pjsua2.dll to $(OutDir)
+        swig -c++ -csharp  -w312 -namespace libpjsua2.maui -I$(ProjectDir)..\..\pjsip\include -I$(ProjectDir)..\..\pjlib\include -o $(OutDir)pjsua2_wrap.cpp %(FullPath)
+        xcopy /Y /I "$(OutDir)*.cs" "$(ProjectDir)pjsua2maui\Platforms\Windows\pjsua2\"
+        copy /Y "$(OutDir)pjsua2.dll" "$(ProjectDir)pjsua2maui\Platforms\Windows\Native\x64\pjsua2.dll"
```
sauwming, line 293: Two problems in these two lines (and repeated at 304-305, 333-334, 345-346):

**1. Wrong destination.** `$(ProjectDir)` here is `pjsip-apps\build\`, so this resolves to `pjsip-apps\build\pjsua2maui\Platforms\Windows\pjsua2\`, which has nothing to do with the app — the MAUI project lives at `pjsip-apps\src\swig\csharp\pjsua2maui\pjsua2maui\`. `xcopy /I` will happily create that bogus tree without complaining, so the generated bindings and `pjsua2.dll` never reach the app. It needs to be `$(ProjectDir)..\src\swig\csharp\pjsua2maui\pjsua2maui\Platforms\Windows\...`.

**2. The DLL doesn't exist yet.** This `copy /Y "$(OutDir)pjsua2.dll"` sits inside the `CustomBuild` step that *generates* `pjsua2_wrap.cpp`, i.e. it runs before compile and link. On a clean build the DLL isn't there yet and the copy fails; on an incremental build it deploys the previous build's DLL. The DLL deployment belongs in a post-build target, not in the SWIG custom build step.

pjsip-apps/build/swig_maui_pjsua2.vcxproj
```diff
@@ -0,0 +1,363 @@
...
+      </Command>
+      <Command Condition="'$(Configuration)|$(Platform)'=='Debug|ARM64'">
+        %40echo off
+        %40echo **************************************************************
+        %40echo  Make sure %27swig%27 is in your PATH environment vars.
+        %40echo **************************************************************
+
+        %40echo on
+        %40echo Building pjsua2.dll to $(OutDir)
+        swig -c++ -csharp  -I$(ProjectDir)..\..\pjsip\include -I$(ProjectDir)..\..\pjlib\include -outdir $(OutDir) -o $(OutDir)pjsua2_wrap.cpp %(FullPath)
```
sauwming, line 303: The ARM64 SWIG invocations (this line and 344) are missing `-namespace libpjsua2.maui`, unlike every x64/Win32 command in this file. The app sources do `using libpjsua2.maui;`, and `win-arm64` is an advertised RID in the csproj, so an ARM64 build generates the bindings into the global namespace and the app then fails to compile.

(For the record: the missing `-outdir` that the Copilot review flagged on these same lines is *not* a bug — SWIG 4.3 defaults the C# output directory to the directory of the `-o` file. Only `-namespace` is actually missing.)

pjsip-apps/build/swig_maui_pjsua2.vcxproj
```diff
@@ -0,0 +1,363 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
sauwming, line 1: This new project isn't referenced by `pjproject-vs14.sln` or any other solution — `git grep swig_maui` matches only this file. So it is never built by CI, and never built by anyone just opening the solution, which is why none of the issues above were caught.

Please either add it to `pjproject-vs14.sln` (which would also give us the CI coverage asked for in the review summary), or document in the README exactly how it's meant to be built.

pjproject-vs14.sln
```diff
@@ -2502,6 +2502,95 @@ Global
...
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|ARM64.ActiveCfg = Release|Any CPU
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|ARM64.Build.0 = Release|Any CPU
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|ARM64.Deploy.0 = Release|Any CPU
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|Win32.ActiveCfg = Release|Any CPU
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|Win32.Build.0 = Release|Any CPU
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|Win32.Deploy.0 = Release|Any CPU
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|x64.ActiveCfg = Release|Any CPU
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|x64.Build.0 = Release|Any CPU
+		{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}.Release-Static|x64.Deploy.0 = Release|Any CPU
+
```
sauwming, line 2593: This block adds 88 `ProjectConfigurationPlatforms` entries for `{C72F15C3-EF84-912D-5E0E-6DC3CCC33C45}`, but there is no matching `Project(...)` declaration anywhere in the file — 88 occurrences of the GUID, zero of them in a `Project` line. These look like leftovers from the reverted "add project to sln" commit and should be removed.

(If instead you go with adding `swig_maui_pjsua2.vcxproj` to this solution, then the `Project(...)` entry needs to be added and these configurations need to name the real platforms, not `Any CPU`.)

There's also a stray tab-only line at 2593, just before `EndGlobalSection`.

pjproject-vs14.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.11.35327.3
+# Visual Studio Version 18
+VisualStudioVersion = 18.6.11819.183
```
sauwming, line 4: Unrelated churn: this bumps `VisualStudioVersion` to whatever the contributor's local VS wrote, and leaves a trailing space at the end of the line. Please revert lines 3-4 to keep the diff focused.

pjsip-apps/src/swig/Makefile
```diff
@@ -4,10 +4,12 @@ ifneq ($(findstring android,$(TARGET_NAME)),)
     # no python for android
     LANG = java csharp
 else
-    ifneq ($(findstring ios,$(TARGET_NAME)),)
-  	LANG = csharp
+	ifneq ($(findstring macabi,$(APP_CFLAGS)),)
```
sauwming, line 7: This `macabi` branch is a no-op as written: when `TARGET_NAME` contains `ios` (which it does for a MacCatalyst target) both this branch and the next one set `LANG = csharp`. If it's here for clarity that's fine, but worth a comment saying so.

Also, the new conditionals are tab-indented while the surrounding ones use spaces — please match the existing style.

pjsip-apps/src/swig/Makefile
```diff
@@ -22,7 +24,16 @@ all: symbols.i
 	
 all clean dep depend distclean print realclean install uninstall: $(LANG)
 
-java python csharp:
+csharp:
+ifeq ($(CMD_is_lang),)
+	$(MAKE) $(MAKE_FLAGS) --no-print-directory -C $@ $(MAKECMDGOALS)
+	$(MAKE) $(MAKE_FLAGS) --no-print-directory -C $@/pjsua2maui $(MAKECMDGOALS)
```
sauwming, line 30: This now recurses into both `csharp` and `csharp/pjsua2maui`, but `pjsip-apps/src/swig/csharp/Makefile` still carries its own complete set of MAUI targets (`MAUI_LIBPJSUA2`, `MAUI_OUT_DIR`, `MAUI_SUP_CLASS_DIR` and their clean rules — unchanged by this PR, and `all:` still depends on `$(MAUI_LIBPJSUA2)`).

So the MAUI wrapper gets built twice, from two different makefiles, into overlapping trees. It's worse than just wasted work for MacCatalyst: the old `csharp/Makefile` classifies MacCatalyst as `OS=ios` and writes MacCatalyst-compiled artifacts into `Platforms/iOS/`, while the new makefile writes into `Platforms/MacCatalyst/`.

The MAUI targets should be removed from `csharp/Makefile` now that they live in `csharp/pjsua2maui/Makefile`.

pjsip-apps/src/swig/csharp/pjsua2maui/Makefile
```diff
@@ -0,0 +1,102 @@
...
+    LIB_NAME = libpjsua2.so
+else ifneq ($(findstring macabi,$(APP_CFLAGS)),)
+    MAUI_OS = maccatalyst
+    LIB_NAME = libpjsua2.a
+else ifneq ($(findstring ios,$(TARGET_NAME)),)
+    MAUI_OS = ios
+    LIB_NAME = libpjsua2.a
+else
+    MAUI_OS = desktop
+    LIB_NAME = libpjsua2.a
```
sauwming, line 15: The `desktop` fallback never gets a `PLATFORM_DIR` of its own: it isn't `android`, and it isn't `maccatalyst`, so it falls through the `else` at line 41 and ends up with `PLATFORM_DIR = iOS` plus `-dllimport "__Internal"`.

A plain `make csharp` on a desktop build therefore writes host-native artifacts into `Platforms/iOS/` and then runs `libtool -static` (line 83) — that's macOS `libtool`; on Linux that name is GNU libtool and the rule fails.

Either give `desktop` a real branch, or make it an explicit `$(error ...)` so it fails with a clear message instead of silently producing wrong output.

pjsip-apps/src/swig/csharp/pjsua2maui/Makefile
```diff
@@ -0,0 +1,102 @@
...
+	if [ -f "${STD_CPP_LIB}" ]; then cp -f ${STD_CPP_LIB} $(MAUI_LIBPJSUA2_DIR); fi
+	# Android specific audio and video helpers
+	if echo "$(MY_CFLAGS)" | grep -q "PJMEDIA_VIDEO_DEV_HAS_ANDROID=1"; then \
+		cp $(PJDIR)/pjmedia/src/pjmedia-videodev/android/PjCamera*.java $(MAUI_SUP_CLASS_DIR)/; \
+	fi
+	if echo "$(MY_CFLAGS)" | grep -q "PJMEDIA_AUDIO_DEV_HAS_OBOE=1"; then \
+		cp $(PJDIR)/pjmedia/src/pjmedia-audiodev/android/PjAudioDevInfo.java $(MAUI_SUP_CLASS_DIR)/; \
+	fi
+else
+	# CORREÇÃO DA APPLE (iOS/MacCatalyst): Uso do libtool nativo para unificar de forma limpa o wrapper e as libs do PJSIP
```
sauwming, line 82: Please keep comments in English — same for the Portuguese comments in `Platforms/MacCatalyst/VideoViewHandler.cs` (lines 21, 24) and `Platforms/Windows/VideoViewHandler.cs` (line 41).

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/Platforms/Windows/VideoViewHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+            nint windHandle = 0;
+            VideoWindowHandle vidWH = new VideoWindowHandle();
+
+            if (Show)
+            {
+                var gridView = PlatformView;
+
+                // Captura o ponteiro nativo (HWND/IUnknown) do elemento visual do WinUI 3
+                // O PJSIP no Windows espera receber o ponteiro correspondente ao controle ou à janela de desenho.
+                IntPtr winPtr = gridView.As<IInspectable>().ThisPtr;
```
sauwming, line 43: `gridView.As<IInspectable>().ThisPtr` is a WinRT `IInspectable` pointer, not an HWND, and PJMEDIA's Windows renderers require an HWND: `pjmedia_vid_dev_hwnd.info.win.hwnd` is documented as `HWND`, and `sdl_dev.c` passes `hwnd->info.window` straight into `SDL_CreateWindowFrom()`. Handing it an `IInspectable` will fail or crash.

Two further problems:

* The `IObjectReference` returned by `.As<>()` isn't stored anywhere, so the GC can collect it and release `ThisPtr` while PJSIP still holds the value.
* WinUI child elements have no HWND at all, so there is nothing to get from `PlatformView` here. The only usable handle is the top-level window's, via `WinRT.Interop.WindowNative.GetWindowHandle(...)`.

Has the video path actually been exercised on Windows? A CI or local build won't catch this one — it needs a run.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/Platforms/Windows/app.manifest
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="RemotePhone.WinUI.app"/>
```
sauwming, line 3: Copy-paste identity from another app: `name="RemotePhone.WinUI.app"`. Should be `pjsua2maui`.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/pjsua2maui.csproj
```diff
@@ -1,8 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net10.0-ios;net9.0-android;net9.0-ios</TargetFrameworks>
-    
+    <TargetFrameworks>net10.0-android</TargetFrameworks>
+    <TargetFrameworks Condition="!$([MSBuild]::IsOSPlatform('linux'))">$(TargetFrameworks);net10.0-ios;net10.0-maccatalyst</TargetFrameworks>
```
sauwming, line 5: `!$([MSBuild]::IsOSPlatform('linux'))` is true on Windows too, so this adds `net10.0-ios` and `net10.0-maccatalyst` on Windows hosts. Neither can be built there, so a plain `dotnet build` of the sample on Windows fails.

This should be `$([MSBuild]::IsOSPlatform('osx'))` for the ios/maccatalyst TFMs, matching the `IsOSPlatform('windows')` guard on the next line.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/pjsua2maui.csproj
```diff
@@ -20,108 +25,153 @@
...
 
+  <PropertyGroup>
+    <AppxPackageSigningEnabled Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">false</AppxPackageSigningEnabled>
+    <HasPackageAndPublishMenu Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">false</HasPackageAndPublishMenu>
+    <ProjectPriFileName Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">resources.pri</ProjectPriFileName>
+    <RuntimeIdentifiers Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">win-x64;win-arm64</RuntimeIdentifiers>
+  </PropertyGroup>
+
+  <PropertyGroup> 
+    <Platform Condition="'$(Platform)' == '' Or '$(Platform)' == 'AnyCPU'">x64</Platform>
```
sauwming, line 47: This `<Platform>` default isn't scoped to Windows, so it forces `Platform=x64` for the android, ios and maccatalyst builds as well.

It also makes the pre-existing `Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net10.0-ios|AnyCPU'"` PropertyGroup just below (line 52) permanently dead, since `$(Platform)` can no longer be `AnyCPU`.

Please add a `windows` platform-identifier condition to this `PropertyGroup`.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/pjsua2maui.csproj
```diff
@@ -20,108 +25,153 @@
...
+    <None Condition="$(RuntimeIdentifier.Contains('x64')) Or '$(Platform)' == 'x64' Or '$(PlatformTarget)' == 'x64'" Include="Platforms\Windows\Native\x64\pjsua2.dll">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <Visible>True</Visible>
+      <TargetPath>%(Filename)%(Extension)</TargetPath>
+    </None>
+    <None Condition="$(RuntimeIdentifier.Contains('arm64')) Or '$(Platform)' == 'ARM64' Or '$(PlatformTarget)' == 'ARM64'" Include="Platforms\Windows\Native\ARM64\pjsua2.dll">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <Visible>True</Visible>
+      <TargetPath>%(Filename)%(Extension)</TargetPath>
+    </None>
```
sauwming, line 166: Switching these two from `Update=` to `Include=` without an `Exists(...)` guard breaks a clean checkout: `pjsua2.dll` is gitignored and doesn't exist until the vcxproj has been built, and `Include=` on a missing file with `CopyToOutputDirectory=Always` fails the Windows TFM build.

Every other native item in this file (the Android `.so`, the iOS/MacCatalyst `.a`) is guarded with `Condition="Exists(...)"` — these two should be too.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/pjsua2maui.csproj.user
```diff
@@ -1,10 +1,14 @@
...
+    <IsFirstTimeProjectOpen>False</IsFirstTimeProjectOpen>
+    <ActiveDebugFramework>net10.0-ios</ActiveDebugFramework>
+    <ActiveDebugProfile>Simulador</ActiveDebugProfile>
+    <SelectedPlatformGroup>Simulator</SelectedPlatformGroup>
+    <DefaultDevice>pixel_7_-_api_29</DefaultDevice>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetPlatformIdentifier)'=='iOS'">
+    <RuntimeIdentifier>iossimulator-x64</RuntimeIdentifier>
+    <PlatformTarget>x64</PlatformTarget>
+  </PropertyGroup>
```
sauwming, line 13: `.csproj.user` is machine-local IDE state and shouldn't be in git at all — and this change makes it actively harmful: it pins `RuntimeIdentifier=iossimulator-x64` and `PlatformTarget=x64` for *every* iOS build. On an Apple Silicon Mac that makes the default iOS build target the x64 simulator, which won't link against the arm64 `libpjsua2.a` the Makefile produces.

It also hardcodes a localized debug profile name (`Simulador`, line 6) and rewrites the whole file's line endings.

Please delete this file from git and add `*.csproj.user` to `.gitignore` instead of updating it.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/.gitignore
```diff
@@ -3,4 +3,6 @@
 .class
 .jar
 .so
+.a
+.cs
```
sauwming, line 7: Minor: `.a` and `.cs` (like the pre-existing `.dll`/`.java`/`.so` above) match files *named* exactly `.a` and `.cs`, not files with those extensions — these need to be `*.a` / `*.cs` to do anything. Worth fixing the whole list while you're here.

Note `*.cs` in this directory would be wrong anyway — it would ignore the app's own sources. The generated bindings are already covered by the `**/Platforms/*/pjsua2/*` rules in the top-level `.gitignore`, so these two lines can probably just be dropped.

sauwming (changes requested):
Thanks for the proposal — extending the MAUI sample to Windows and MacCatalyst is genuinely useful, and the overall structure (a dedicated `csharp/pjsua2maui/Makefile`, per-platform `VideoViewHandler`, a native Windows vcxproj) is the right shape.

That said, I don't think this is mergeable yet. The main problem is that **nothing in this PR is actually compiled by CI**, so the all-green check list is not evidence that it builds:

* `pjsip-apps/build/swig_maui_pjsua2.vcxproj` is not referenced by `pjproject-vs14.sln` or any other solution (`git grep swig_maui` matches only the file itself), so `ci-win` never touches it.
* The MAUI app itself is never built anywhere.

Going through it by hand, there are several hard build breaks — the malformed `Info.plist`, the `$(ProjectDir)` copy destinations in the vcxproj, the missing `-namespace` on the ARM64 SWIG calls, and the unguarded `Include=` for `pjsua2.dll`. Details are in the inline comments.

### Could we get one CI build covering this?

Before the next review round, could you wire up **at least one CI job that actually compiles what this PR adds**? Even a single configuration would catch most of the findings below. Two options that would work:

1. Add `swig_maui_pjsua2.vcxproj` to `pjproject-vs14.sln` so the existing `ci-win.yml` builds it (this also fixes the "not in any solution" problem on its own), and/or
2. Add a job that runs `dotnet build` on `pjsua2maui.slnx` for one TFM.

If neither is practical in this repo's CI, please at least paste the output of a local clean build for each platform you're claiming support for (Windows x64 + ARM64, MacCatalyst) into the PR description, so it's on record that the build was verified.

### Other things to settle before merge

* **CLA is still unsigned** for one of the committers (`Weslley`, commits `8b1f9d3f` and `c26c5de4`, authored with an empty `=` e-mail). This blocks merge independently of the code.
* Dropping `net9.0-android` / `net9.0-ios` in favour of net10.0-only is a real requirement bump for anyone using this sample. That's fine if deliberate, but it should be a conscious decision and it needs a README note.
* Replacing `pjsua2maui.sln` with `pjsua2maui.slnx` requires VS 17.13+ / .NET SDK 9.0.200+. Also fine if intentional, same request: state it in the README.
* `pjsua2maui/README.md` is untouched — it still documents Android/iOS only and still references `net9.0-ios` output paths. It should cover the two new platforms and the new build flow.

One note in the other direction: the missing `-outdir` on the ARM64 SWIG commands that the Copilot review flagged is **not** a bug — SWIG 4.3 defaults the language output directory to the directory of the `-o` file. The missing `-namespace` on those same lines *is* a real bug though (see inline).

weslleymurdock:
Sorry by the late answer and implementation, i'll fix this as soon as possible, this past month i've run out of time because my graduation. 
I need a little help implementing the VideoWindow on Windows targets, i do not found a type at MAUI that is  compatible with the HWND implemented by pjsua, and currently i dont have a sip server to test the video calls.
I'll continue the research, but if there anyone to help me with that i'll be glad.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/Platforms/Windows/VideoViewHandler.cs
weslleymurdock: There is a way to do this in a control, or something like? I do not tested the video screen because i'm currently do not own an asterisk installation to test it, but i can add a audio/video/codecs page setup, that embbeds an control for preview the camera, to test the handle, is it valid?

weslleymurdock:
Some note about the CI job:
At the inital commits, i've put the projects to build using the existent jobs. But i've got build failures, because of conflicts with the current swig csharp (winforms) project, also the restore workloads needed by the maui project causes unexpected behaviours at the current workflows.
Given that, i consider the following option: 
- Instantiate a solution 'fixture' for contain all pj*.vcxprojs and the swig maui vsxproj only, to use into a separated workflow for build the native pj library for all maui platforms, and then build the maui sample, copying the assets built previously from this maui sample exclusive workflow, to avoid breaks all current ci's. 
I've already have an workflow like that, on a personal repo that i built for packaging pj as a nuget package. If there isn't a problem implement the build of maui sample (and swig maui) like that, I'll do it.

pjsip-apps/build/swig_maui_pjsua2.vcxproj
```diff
@@ -0,0 +1,363 @@
...
+      <WarningLevel>Level3</WarningLevel>
+    </ClCompile>
+    <Link>
+      <AdditionalDependencies>Iphlpapi.lib;dsound.lib;dxguid.lib;netapi32.lib;mswsock.lib;ws2_32.lib;odbc32.lib;odbccp32.lib;ole32.lib;user32.lib;gdi32.lib;advapi32.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <OutputFile>..\src\swig\csharp\output\maui\pjsua2.dll</OutputFile>
+      <!-- <IgnoreSpecificDefaultLibraries>msvcrt.lib;%(IgnoreSpecificDefaultLibraries)</IgnoreSpecificDefaultLibraries> -->
+      <GenerateDebugInformation>true</GenerateDebugInformation>
+      <TargetMachine>MachineX64</TargetMachine>
+    </Link>
+  </ItemDefinitionGroup>
```
sauwming, line 190: **Debug CRT mismatch.** Both of these Debug configurations import `pjproject-vs14-debug-static-defaults.props` (lines 86 and 101), which sets `<RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>` — the *static* debug CRT (`/MTd`).

The sibling `swig_csharp_pjsua2.vcxproj` imports exactly the same props, but overrides it back to `MultiThreadedDebugDLL` in its Debug|Win32 and Debug|x64 `ItemDefinitionGroup`s (lines 151 and 169 there) precisely because of this. Those two overrides were dropped when this project was derived from it, so this `DynamicLibrary` — which exports C++/STL objects across the DLL boundary to the MAUI app — ends up built against the static debug CRT while the PJSIP libs it links against are built dynamic.

`IgnoreSpecificDefaultLibraries` is also commented out here (169, 186) and in every other configuration in this file, where the sibling has it active.

Please restore `<RuntimeLibrary>MultiThreadedDebugDLL</RuntimeLibrary>` for Debug|Win32 and Debug|x64 and uncomment the `IgnoreSpecificDefaultLibraries` lines, so this project matches `swig_csharp_pjsua2.vcxproj`. (The Release configs and both ARM64 configs already match the sibling, so those are fine as-is.)

sauwming (commented):
No rush on this — congratulations on the graduation, and thanks for coming back to it.

Answering your two points, plus two findings I missed in the previous round (inline).

### 1. The CI "solution fixture" + separate workflow

Yes, please do it that way. A dedicated solution containing only the `pj*.vcxproj`s plus `swig_maui_pjsua2.vcxproj`, driven by its own workflow file, is exactly the right shape. It keeps the MAUI workload restore and the `net10.0-*` toolchain out of `ci-win.yml`, which is what was causing the conflicts you hit when you tried to reuse the existing jobs, and it gives us a job that actually compiles what this PR adds. That last part matters: right now `git grep swig_maui` matches only the vcxproj itself, so nothing in this PR is built by any CI job — which is why the check list is green despite the build breaks flagged in the previous review.

This supersedes the "add it to `pjproject-vs14.sln`" suggestion in my earlier comment on the vcxproj — your approach solves the same problem better, and it's consistent with what we agreed back in July about a separate solution for the MAUI sample.

Two requests on top:

- Gate the workflow with `paths:` so it only runs when `pjsip-apps/src/swig/**`, `pjsip-apps/build/swig_maui_pjsua2.vcxproj` or the new solution change — no need to spend CI minutes on unrelated PRs.
- A single configuration (Windows x64 Debug, say) is plenty for the first round. It will catch most of the outstanding findings on its own.

If the full matrix turns out to be too much work, the fallback stands: paste a local clean-build log for each platform you're claiming (Windows x64 + ARM64, MacCatalyst) into the PR description so the verification is on record.

While you're there, please also drop the 88 orphaned `ProjectConfigurationPlatforms` entries for `{C72F15C3-…}` from `pjproject-vs14.sln` — they're leftovers from the reverted "add project to sln" commit and reference a GUID that has no `Project(...)` line (and isn't even this vcxproj's `{28181FAC-…}`).

### 2. VideoWindow on Windows

You won't find a MAUI type that works, because there isn't one. In WinUI 3 there is no per-control HWND at all — every `UIElement`, `Grid` included, is composition-rendered inside the *single* top-level window HWND. So nothing you can pull off `PlatformView` will ever be a valid handle, which is what the inline comment on `VideoViewHandler.cs:43` is about.

Two ways forward:

**(a) Don't hand PJSIP a window at all** — this is what I'd suggest for a sample. Skip `setWindow()` on Windows and let pjmedia create and own its own SDL window, then just position it and show/hide it via `VideoWindow.setPos()` / `Show()`. You get a separate top-level video window rather than an embedded one, which is not pretty, but it works today with zero native interop and no lifetime hazards.

**(b) Embed it properly** — the only handle MAUI can give you is the top-level one, via `WinRT.Interop.WindowNative.GetWindowHandle(mauiWindow.Handler.PlatformView)`. To get a child region you have to create your own HWND: `CreateWindowEx` a plain child window parented to that top-level HWND, keep it alive in a field, and pass *that* to `setWindow()`. You then have to size and position it by hand to track the `Grid`'s layout on `SizeChanged`. That's real native interop work — more than a sample arguably needs, but it's the only route to a genuinely embedded view.

Either way, see the inline comment about `vidWH.type` — it's currently never set, which would break the handle even once you have a valid one.

**On testing without a SIP server:** your idea of an audio/video/codec settings page with a camera preview is a good one, and yes, it's a valid way to exercise this. `VidPreview.start()` on the capture device renders through the same `VideoWindow` / `setWindow()` path that an incoming call uses, so a working local preview proves the handle end to end. That would be a welcome addition to the sample in its own right.

---

Everything else from the previous review is still outstanding — the head commit here is just a merge of master, so those inline comments all still apply. Ping me when the next round is up and I'll take another look.

pjsip-apps/src/swig/csharp/pjsua2maui/pjsua2maui/Platforms/Windows/VideoViewHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+
+                windHandle = (nint)winPtr;
+            }
+
+            vidWH.handle.window = windHandle;
```
sauwming, line 48: `vidWH.type` is never set, so it stays at its default `PJMEDIA_VID_DEV_HWND_TYPE_NONE`. `pjsua2` copies it straight through to pjmedia — `pjsip/src/pjsua2/media.cpp:1629`:

```cpp
vhwnd.type = win.type;
vhwnd.info.window = win.handle.window;
```

so the renderer never learns what kind of handle it was given. You need:

```csharp
vidWH.type = pjmedia_vid_dev_hwnd_type.PJMEDIA_VID_DEV_HWND_TYPE_WINDOWS;
```

before `setWindow()`. This is independent of the handle problem above — both need fixing.

Same omission in `Platforms/MacCatalyst/VideoViewHandler.cs:37` (needs `PJMEDIA_VID_DEV_HWND_TYPE_COCOA`), and in the pre-existing `Platforms/Android/VideoViewHandler.cs:38` — worth fixing all three while you're in here.

sauwming: Correction to my own comment above: for MacCatalyst it should be `PJMEDIA_VID_DEV_HWND_TYPE_IOS`, **not** `_COCOA`.

Mac Catalyst is UIKit-on-Mac, and pjmedia treats it as iOS throughout: `pjmedia/src/pjmedia-videodev/darwin_dev.m:1099` and `:1106` set `hwnd->type = PJMEDIA_VID_DEV_HWND_TYPE_IOS` unconditionally, and `metal_dev.m:709` selects between the two with `TARGET_OS_IPHONE`, which is 1 under Mac Catalyst. `_COCOA` is for AppKit/`NSView`, which is not what `PlatformView` gives you there.

So:

- `Platforms/Windows/VideoViewHandler.cs` → `PJMEDIA_VID_DEV_HWND_TYPE_WINDOWS`
- `Platforms/MacCatalyst/VideoViewHandler.cs` → `PJMEDIA_VID_DEV_HWND_TYPE_IOS`
- `Platforms/Android/VideoViewHandler.cs` → `PJMEDIA_VID_DEV_HWND_TYPE_ANDROID`

Sorry for the noise — the rest of that comment stands.

.gitignore
```diff
@@ -96,5 +96,18 @@ __cmake_systeminformation/
...
+**/Platforms/Android/org/pjsip/*
+**/Platforms/iOS/pjsua2/*
+**/Platforms/iOS/lib/*
+**/Platforms/MacCatalyst/pjsua2/*
+**/Platforms/MacCatalyst/lib/*
+**/Platforms/Windows/Native/*
+**/Platforms/Windows/pjsua2/*
+workload-install.ps1
+
+**/problems-report.html
```
sauwming, line 113: Three things in this block:

**1. `**/Platforms/Windows/Native/*` makes an existing problem certain.** `pjsua2maui.csproj:157` and `:162` reference `Platforms\Windows\Native\{x64,ARM64}\pjsua2.dll` with `Include=` (not `Update=`) and `CopyToOutputDirectory=Always`, with no `Condition="Exists(...)"`. Ignoring the DLL here guarantees it is absent on a clean checkout, so the Windows TFM build now fails with MSB3030 every time until the vcxproj has been built. The `Exists(...)` guard on those two items — which every other native item in that csproj already has — is now mandatory, not just defensive.

**2. `workload-install.ps1` is unanchored**, so it matches a file of that name anywhere in the tree, not just where you intended. If it's needed, make it a rooted path.

**3. `workload-install.ps1` and `**/problems-report.html` look out of scope** for a PR adding Windows/MacCatalyst support to the MAUI sample. If they're build artifacts from your local Android/MAUI tooling, they'd be better as a separate PR (or in your global gitignore), so this one stays reviewable.

sauwming (commented):
Thanks for the `.sln` fix — confirmed: `pjproject-vs14.sln` is now byte-identical to master (`git diff <merge-base> HEAD -- pjproject-vs14.sln` is empty), so both the 88 orphaned `{C72F15C3-…}` entries and the `VisualStudioVersion` churn are resolved. The CLA also shows signed for all committers now. Those two are closed.

Everything else from the previous two reviews is still open — I re-verified each one against `391d02320` rather than going by the old comments. Quick status so you know what's left:

**Hard build breaks, unchanged:**

- `Platforms/MacCatalyst/Info.plist:41` — still `<key>NSCameraUsageDescription </string>`; `plutil -lint` still fails.
- `swig_maui_pjsua2.vcxproj` — all eight `$(ProjectDir)pjsua2maui\...` copy destinations still resolve under `pjsip-apps/build/`; the `pjsua2.dll` copy is still inside the pre-link `CustomBuild` step.
- `swig_maui_pjsua2.vcxproj:303` and `:344` — ARM64 SWIG still missing `-namespace libpjsua2.maui`.
- `swig_maui_pjsua2.vcxproj:160,174` — Debug|Win32/x64 still missing the `MultiThreadedDebugDLL` override.
- `pjsua2maui.csproj:5` — `!IsOSPlatform('linux')` still adds the ios/maccatalyst TFMs on Windows hosts.
- `pjsua2maui.csproj:157,162` — unguarded `Include=` for `pjsua2.dll`; see the inline comment, this one got worse with the new gitignore rule.

**Still no CI coverage.** `pjsua2maui.slnx` contains only the csproj, no solution references `swig_maui_pjsua2.vcxproj`, and every GitHub Actions job on this PR reports `skipping`. The solution-fixture + dedicated workflow we agreed on in the last round hasn't landed yet, so none of the above is being caught by a build. That's still the single most useful thing you could add — a Windows x64 Debug job alone would surface most of this list.

**One red check:** `ci/bitrise/e4b6aade20ea9eb3/pr` is now failing (30m). It passed on the earlier head, so it's worth a look — it may or may not be related to the iOS-side changes here.

**Also still open:** `vidWH.type` never set in either handler (see my correction in the thread on `Platforms/Windows/VideoViewHandler.cs` — it's `_IOS` for MacCatalyst, not `_COCOA` as I first wrote); the `IInspectable` handle; `pjsua2maui.csproj:47` forcing `Platform=x64` for all TFMs; the duplicated MAUI targets still present in `pjsip-apps/src/swig/csharp/Makefile`; the `desktop` fallback in `pjsua2maui/Makefile` landing in the iOS branch; the tracked `pjsua2maui.csproj.user`; `app.manifest`'s `RemotePhone.WinUI.app` identity; the Portuguese comments; and `pjsua2maui/README.md`, which still documents `net9.0-android` / `net9.0-ios` and the now-deleted `pjsua2maui.sln`.

No rush — just flagging that the state hasn't moved on those yet so you know what the next round needs to cover.

PR #5002, opened 2026-05-29
Fix heap-use-after-free in shared video preview hangup races and harden clock stop/destroy serialization

sauwming:
## Description
This PR fixes a heap-use-after-free that occurs when concurrent call
hangups race on a shared video preview window (notably with
`Colorbar-active`), and hardens the clock stop/destroy path against
concurrent callers.

Main updates in the final state of this PR:

- **Serialized video window refcount transitions and made teardown
  idempotent** for concurrent decrements. `inc_vid_win` /
  `dec_vid_win` now run under `PJSUA_LOCK`. A new `pj_bool_t
  is_destroying` field on `pjsua_vid_win` marks the slot dead before
  `free_vid_win()` releases the lock to destroy its vid_ports, so
  racing inc/dec/preview-lookup/enumeration on the same wid see the
  flag and bail. The flag is cleared (via `pj_bzero`) by
  `pjsua_vid_win_reset()` at the tail of `free_vid_win()`.
- **Prevented teardown races by snapshotting `vp_cap`/`vp_rend` under
  `PJSUA_LOCK`** in both `free_vid_win()` and `pjsua_vid_stop_stream()`
  before any destroy/unsubscribe work, so a racing free can't NULL
  them out mid-call. `free_vid_win()` also clears the window's slot
  pointers under the lock so concurrent readers see NULL and skip.
- **Updated stop-stream cleanup** to always unsubscribe the
  `call_media_on_event` subscription from the stream's decoding port
  even when the renderer port has already been torn down.
- **Serialized `pjmedia_clock_stop()` / `pjmedia_clock_destroy()` join
  paths** via a recursive `destroy_lock` on `pjmedia_clock`, and
  propagated join errors so callers don't free backing memory while
  the clock thread is still alive. Distinguishes `PJ_ECANCELLED`
  (self-join) from other OS errors — both early-return with the clock
  state preserved for a future retry.
- **Co-owned `destroy_lock` with `clock->pool`**: allocated from
  `clock->pool`, `pj_mutex_destroy()`-ed at the tail of
  `pjmedia_clock_destroy()` immediately before
  `pj_pool_safe_release(&clock->pool)`. Lock and the memory it guards
  share one lifetime, torn down together by the single owner. No
  OS-resource leak on any platform (the Windows `CRITICAL_SECTION`
  global-list entry is properly freed). Relies on the higher-layer
  `is_destroying` guarantee that no concurrent stop/destroy can race
  past a successful destroy.
- **Documented the `pjmedia_clock_destroy()` caller contract** in the
  public header: must not be called from within the clock callback
  (self-join returns `PJ_EBUSY`); the caller must also ensure no
  concurrent stop/destroy is in flight on the same clock when the
  destroy proceeds past its internal join (the lock that serializes
  stop/destroy is torn down at the tail of a successful destroy).

## Motivation and Context
Concurrent hangup and teardown paths could race across shared preview
windows and clock thread shutdown, leading to UAF crashes under
stress. The fix combines per-window serialization (pjsua_vid) with
clock-level serialization (pjmedia_clock); the two layers give
complementary guarantees — the higher layer prevents the duplicate
destroy, the lower layer makes single-owner stop/destroy safe and
leak-free.

## How Has This Been Tested?
- Built project successfully after configure/dependency generation:
  - `./configure && make dep && make -j2`
  - `make -C pjmedia/build all`
- Ran targeted tests:
  - `make pjmedia-test` (pass)
- Stress validation (ASan/macOS with `Colorbar-active`) clean across
  repeated `pjsua-stress` runs: 10/10 at `-n 8 -v 4` 30 s, 8–10/10 at
  `-n 32 -v 8` 30 s (the rare miss is a test-side cleanup-window
  timeout, no ASan/assert).

## Types of changes

## Checklist:

### DIFF
--- a/pjmedia/src/pjmedia/clock_thread.c
+++ b/pjmedia/src/pjmedia/clock_thread.c
@@ -511,17 +511,10 @@ PJ_DEF(pj_status_t) pjmedia_clock_destroy(pjmedia_clock *clock)
         clock->lock = NULL;
     }
 
-    /* Release destroy_lock and tear down its OS resources together
-     * with the pool that backs its memory. The lock and the memory
-     * it guards share one lifetime — both go away here, under one
-     * owner. Safe because the higher layer (e.g. pjsua_vid's
-     * is_destroying flag + dec_vid_win under PJSUA_LOCK) guarantees
-     * no concurrent destroy/stop on the same clock; any in-flight
-     * concurrent caller has already drained against destroy_lock
-     * above. */
+    /* Keep destroy_lock valid for the clock object's lifetime.
+     * This avoids late stop/destroy callers touching a destroyed/NULL
+     * mutex after another thread has already completed destroy(). */
     pj_mutex_unlock(clock->destroy_lock);
-    pj_mutex_destroy(clock->destroy_lock);
-    clock->destroy_lock = NULL;
 
     pj_pool_safe_release(&clock->pool);
 
--- a/pjmedia/src/pjmedia/clock_thread.c
+++ b/pjmedia/src/pjmedia/clock_thread.c
@@ -207,10 +207,17 @@ PJ_DEF(pj_status_t) pjmedia_clock_create2(pj_pool_t *pool,
      * rejects the second one with EINVAL, but on Windows the
      * underlying WaitForSingleObject() permits multiple waiters and
      * *both* return success, leading to double pj_thread_destroy()
-     * and double pj_pool_reset(). Allocated from the caller's pool so
-     * it outlives clock->pool's release in pjmedia_clock_destroy().
-     */
-    status = pj_mutex_create_recursive(pool, "clockdestroy",
+     * and double pj_pool_reset().
+     *
+     * Allocated from clock->pool — not the caller's pool — so the
+     * lock and the memory it guards share one lifetime and are torn
+     * down together in pjmedia_clock_destroy() by a single owner.
+     * pj_mutex_destroy(destroy_lock) runs immediately before
+     * pj_pool_safe_release(&clock->pool) there. This relies on the
+     * caller not issuing concurrent destroy on the same handle —
+     * which the higher-layer is_destroying flag + dec_vid_win under
+     * PJSUA_LOCK guarantees for the pjsua video path. */
+    status = pj_mutex_create_recursive(clock->pool, "clockdestroy",
                                        &clock->destroy_lock);
     if (status != PJ_SUCCESS) {
         pj_lock_destroy(clock->lock);
@@ -504,14 +511,19 @@ PJ_DEF(pj_status_t) pjmedia_clock_destroy(pjmedia_clock *clock)
         clock->lock = NULL;
     }
 
-    pj_pool_safe_release(&clock->pool);
-
-    /* Keep destroy_lock valid for the lifetime of this clock object.
-     * Unlocking and immediately destroying it can race with concurrent
-     * stop/destroy callers currently blocked on the same mutex.
-     */
+    /* Release destroy_lock and tear down its OS resources together
+     * with the pool that backs its memory. The lock and the memory
+     * it guards share one lifetime — both go away here, under one
+     * owner. Safe because the higher layer (e.g. pjsua_vid's
+     * is_destroying flag + dec_vid_win under PJSUA_LOCK) guarantees
+     * no concurrent destroy/stop on the same clock; any in-flight
+     * concurrent caller has already drained against destroy_lock
+     * above. */
     pj_mutex_unlock(clock->destroy_lock);
+    pj_mutex_destroy(clock->destroy_lock);
+    clock->destroy_lock = NULL;
+
+    pj_pool_safe_release(&clock->pool);
 
     return ret;
 }
-
--- a/pjmedia/src/pjmedia/clock_thread.c
+++ b/pjmedia/src/pjmedia/clock_thread.c
@@ -123,6 +123,12 @@ struct pjmedia_clock
     pj_bool_t                running;
     pj_bool_t                quitting;
     pj_lock_t               *lock;
+    /* Serializes pjmedia_clock_stop() and pjmedia_clock_destroy()
+     * across concurrent callers. Cannot be the same as `lock` above,
+     * because `lock` is held by the clock thread inside the callback
+     * (see clock_thread()), and stop/destroy must be allowed to run
+     * while the thread is mid-callback. */
+    pj_mutex_t              *destroy_lock;
 };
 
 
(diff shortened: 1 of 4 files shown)


pjmedia/src/pjmedia/clock_thread.c
```diff
@@ -186,12 +192,33 @@ PJ_DEF(pj_status_t) pjmedia_clock_create2(pj_pool_t *pool,
...
+     * vid_port handle_format_change and free_vid_win). Without this,
+     * two pthread_join() calls race the same descriptor — POSIX
+     * rejects the second one with EINVAL, but on Windows the
+     * underlying WaitForSingleObject() permits multiple waiters and
+     * *both* return success, leading to double pj_thread_destroy()
+     * and double pj_pool_reset(). Allocated from the caller's pool so
+     * it outlives clock->pool's release in pjmedia_clock_destroy().
+     */
+    status = pj_mutex_create_recursive(pool, "clockdestroy",
+                                       &clock->destroy_lock);
```
nanangizz: where is `pj_mutex_destroy()` for this mutex?
update: from copilot review, it seems to be covered by pool release(?), that is not sufficient, as the mutex handle may need to be released too.

sauwming: @copilot: add pj_mutex_destroy() for this mutex

pjmedia/src/pjmedia/clock_thread.c
```diff
@@ -245,28 +272,47 @@ PJ_DEF(pj_status_t) pjmedia_clock_start(pjmedia_clock *clock)
...
 
+    /* Serialize against concurrent stop/destroy on the same clock.
+     * The first caller does the join + descriptor destroy; subsequent
+     * callers acquire the mutex, find clock->thread == NULL, and
+     * return PJ_SUCCESS without touching the thread handle. This is
+     * portable across POSIX (where pthread_join would reject the
+     * second joiner with EINVAL) and Windows (where the underlying
+     * wait would otherwise allow multiple successful waiters).
+     */
+    pj_mutex_lock(clock->destroy_lock);
```
nanangizz, line 289: is it possible that the subsequent caller is `pjmedia_clock_stop()` and the destroy_lock is destroyed already?

sauwming: Good catch — yes, it is possible, and it would crash.

The scenario:
1. Thread T1 calls `pjmedia_clock_destroy()`, acquires `destroy_lock`,
   tears down `clock->pool`, releases `destroy_lock`, and (with the
   previous iteration of this code) `pj_mutex_destroy`s it and NULLs
   it out.
2. Thread T2 — which still holds the `pjmedia_clock *` — then calls
   `pjmedia_clock_stop()`. At line 287
   `pj_mutex_lock(clock->destroy_lock)` dereferences NULL → SIGSEGV
   (or hits a destroyed mutex if it lands between the unlock and the
   destroy).

This isn't purely theoretical. The exact pattern is what triggered
the original UAF this PR fixes: two `cbar_stream_destroy()`
invocations racing on the same stream. Each invocation calls
`pjmedia_clock_stop` → `pjmedia_clock_destroy`. With the previous
iteration of this PR, the second invocation's `pjmedia_clock_stop`
would land after the first's `pjmedia_clock_destroy` completed and
torn down the mutex. The higher-layer pjsua_vid fix (`is_destroying`
+ `dec_vid_win` under `PJSUA_LOCK`) prevents the double-invocation in
pjsua's video path, but the `pjmedia_clock` API itself shouldn't
rely on that.

Fix in `4cb82fe17`: drop the `pj_mutex_destroy(destroy_lock)` and
`destroy_lock = NULL` from `pjmedia_clock_destroy()`. A late
`stop`/`destroy` now finds `destroy_lock` still valid, locks it, sees
`clock->thread == NULL`, and returns cleanly.

Trade-off with the prior reviewer's "destroy_lock is leaked"
concern: the mutex memory lives in the caller's pool and is freed
when that pool is released. On all supported platforms (POSIX,
Windows) the OS-level resources of an *unlocked* mutex are
effectively nil — `pthread_mutex_destroy` is essentially bookkeeping,
and `CRITICAL_SECTION` has no kernel object. So skipping the explicit
destroy isn't an actual resource leak, only a contractual one. The
safety win against `stop`-after-`destroy` is clearly worth it,
especially since that's the exact bug class this PR was added to
defend against.

nanangizz: @copilot recheck further (on various OSes) that not destroying mutex here will not leak a mutex handle?

nanangizz: @copilot Note that the `destory_lock` does not seem to be destroyed anywhere. Are you saying that not calling `DeleteCriticalSection()` will not cause a CRITICAL_SECTION/event handle leak? Another AI says otherwise. A clean code should call `pj_mutex_destroy()` for the created mutex, shouldn't it?

nanangizz: Honestly, not fully convinced by the copilot answer :) It seems to ignore or trade a possible handle leak for concurrency issues. Note that clock is used by the sound port which normally created and destroyed on each call, so a handle leak can be fatal.

sauwming: Hahaha, Claude also said the same thing. If there's an alternative solution for the mutex destruction that doesn't introduce race, we can explore it here.

nanangizz: Claude Code says:

Per-platform leak behavior of skipping `pj_mutex_destroy()`, traced through pjlib:

**Windows (`os_core_win32.c`)** — leaks.
- Default path (`PJ_WIN32_WINNT >= 0x0400`): mutex is a `CRITICAL_SECTION`. `init_mutex()` → `InitializeCriticalSection()`, `pj_mutex_destroy()` → `DeleteCriticalSection()`. `InitializeCriticalSection()` heap-allocates an `RTL_CRITICAL_SECTION_DEBUG` block and links it into a process-global list; `DeleteCriticalSection()` is what unlinks/frees it. Skipping it leaks that block per clock. And because `destroy_lock` sits in the caller pool, releasing that pool frees the CS memory while the global list still points into it → dangling node (can fault later in an unrelated `Initialize/DeleteCriticalSection`). On older Windows a contended CS also lazily allocates an auto-reset event handle that `DeleteCriticalSection` closes.
- Legacy path (`< 0x0400`): `init_mutex()` → `CreateMutex()` (real kernel `HANDLE`), `pj_mutex_destroy()` → `CloseHandle()`. Unambiguous kernel-handle leak per clock.

**Linux (`os_core_unix.c`)** — does not leak. `pj_mutex_t` wraps `pthread_mutex_t`; glibc allocates no fd/kernel object at init, and `pthread_mutex_destroy()` on an unlocked normal mutex is just bookkeeping. Memory is reclaimed with the pool.

**iOS / macOS** — same as Linux. `os-darwinos.mak` compiles `os_core_unix.c`, so the mutex is a Darwin `pthread_mutex_t` — user-space, no mach port/handle allocated at init, `pthread_mutex_destroy()` is bookkeeping. No handle leak.

So: genuinely safe on POSIX (Linux/iOS/macOS), but on Windows skipping the destroy leaks (and risks global-list corruption via the caller-pool free). Since the clock is created per sound port repeatedly, that accumulates.

nanangizz: claude code says: 

Suggested fix: allocate `destroy_lock` from `clock->pool` (not the caller pool), and at the tail of `pjmedia_clock_destroy()` — after the thread is joined and `clock->thread` is cleared under the lock — call `pj_mutex_destroy(clock->destroy_lock)` right before `pj_pool_safe_release(&clock->pool)`. The lock and the memory it guards then share one lifetime and are torn down together by one owner. This restores the missing destroy (so Windows no longer leaks/corrupts the global CS list) while keeping the serialization that fixes the UAF. It relies on the caller not issuing concurrent destroy on the same handle — which the higher-layer `is_destroying` + dec under `PJSUA_LOCK` already guarantees for the pjsua path.

Alternatives, if the clock API must tolerate concurrent destroy internally:
- Use a `pj_grp_lock_t` (add_ref per in-flight stop/destroy, teardown in the grp_lock destroy handler) — it destroys its own mutex exactly once.
- Or a process-global/static serialization lock that is never destroyed — no leak, but coarser.

Key rule either way: `pj_mutex_destroy()` and the free of its backing memory must happen together, under one owner, after the join — never via a bare pool release.

sauwming: This should be addressed in the latest commit, please let us know if we miss anything

nanangizz (approved):

PR #5004, opened 2026-06-02
test: fuzz: Fix formatting in fuzz-audio.c comments

DavidKorczynski:


sauwming:
There are failures in the fuzzer.

sauwming (approved):

sauwming:
Ah, now I understand that there's currently a temporary workaround in the build script:
https://github.com/google/oss-fuzz/blob/master/projects/pjsip/build.sh#L21

nanangizz (approved):

DavidKorczynski:
> Ah, now I understand that there's currently a temporary workaround in the build script: https://github.com/google/oss-fuzz/blob/master/projects/pjsip/build.sh#L21

Ah yes, apologies -- should have mentioned. I deployed it to avoid hitting a build issue over night in OSS-Fuzz, and also to confirm its working with it. Will revert that commit now.

PR #5008, opened 2026-06-05
test: fuzz: add WAV/AVI container parser fuzz target

DavidKorczynski:
Add fuzz-wav, a fuzz target for the WAV and AVI container file format parsers. This harness is to be consumed by OSS-Fuzz and is written to ensure currently uncovered code gets covered.


nanangizz:
The failure the new `fuzz-wav` harness surfaced here is an **assertion abort**, not a memory-safety bug: a WAV `fmt` chunk with `sample_rate == 0` or `nchan == 0` passes the existing format checks and reaches `pjmedia_port_info_init()`, whose `PJ_ASSERT(clock_rate && channel_count)` aborts (ASan reports it as `ABRT ... in pthread_kill`).

#5030 fixes this specific case by validating the header in `pjmedia_wav_player_port_create()` (and applies the same guard to the AVI player's audio stream). That resolves this particular assertion; whether the harness runs fully clean is for fuzzing to confirm — it may well surface other malformed-input paths.

nanangizz:
I dug into the CIFuzz failure on this PR. The crash is a harness-init gap rather than a parser bug:

```
fuzz-wav: format.c:344: pjmedia_get_video_format_info: Assertion `mgr != NULL' failed.
  #10 pjmedia_get_video_format_info     format.c:344
  #11 pjmedia_avi_player_create_streams avi_player.c:505
```

`pjmedia_endpt_create2()` creates the video *codec* manager but **not** the video *format* manager (nor the converter/event managers). As soon as the fuzzer crafts an AVI with a recognizable video stream, `avi_player.c` calls `pjmedia_get_video_format_info()` with a NULL manager instance and aborts. A real app (e.g. the `aviplay` sample) creates these managers before parsing.

Suggested fix — initialize the video subsystem in the one-time init block:

```c
#include <pjmedia/converter.h>
#include <pjmedia/event.h>
#include <pjmedia/format.h>
```

```c
    /* (after pjmedia_endpt_create2 succeeds) */
    {
        pj_pool_t *vpool = pj_pool_create(&caching_pool.factory,
                                          "fuzz-wav-init", 4000, 4000, NULL);
        if (!pjmedia_video_format_mgr_instance())
            pjmedia_video_format_mgr_create(vpool, 64, 0, NULL);
        if (!pjmedia_converter_mgr_instance())
            pjmedia_converter_mgr_create(vpool, NULL);
        if (!pjmedia_event_mgr_instance())
            pjmedia_event_mgr_create(vpool, 0, NULL);
    }
```

With that in place the AVI video path is actually fuzzed, which then surfaces a real (separate) parser bug: a video `strl` header with `scale == 0` (or `rate == 0`) is passed as the fps denominator to `pjmedia_format_init_video()` and aborts / later divides by zero. I've opened #5034 to harden that (it complements #5030, which fixed the WAV header and the AVI audio stream but could not reach the video path from the WAV harness).

Verified locally under clang + ASan + libFuzzer with video enabled: after both changes, the original crash input and ~10M fuzz executions across WAV/AVI inputs run clean.

DavidKorczynski:
@nanangizz thanks for going through this. Just to confirm, are you currently waiting for something for me? I'm happy with the changes you suggest

nanangizz:
Yeah, waiting for your approval (or commit) on the suggestion. :) We're traveling right now and away from our dev tools, so we'll continue the review next week.

DavidKorczynski:
@nanangizz I have updated with your suggested changes. The current CIFuzz failures are both true positives as far as I can tell

sauwming (approved):

nanangizz (approved):

PR #5014, opened 2026-06-11
ipjsua/ios: apply and resync video orientation on startup/resume

trengginas:
**Summary**
This change improves orientation handling in the iOS ipjsua sample so camera capture orientation is correct when the app starts or returns to foreground while already in landscape.

### What changed
- Apply an initial orientation sync after pjsua startup:
schedule HANDLE_ORI_CHANGE right after orientation notifications are registered.
- Resync orientation on app resume:
schedule HANDLE_ORI_CHANGE in applicationDidBecomeActive.
- Improve startup/resume fallback:
when UIDeviceOrientation is not usable (Unknown/non-portrait-landscape), fallback to UIInterfaceOrientation (iOS 13+ windowScene.interfaceOrientation).
- Keep orientation application robust:
do not skip applying orientation just because the value appears unchanged; newly-created capture streams still need explicit orientation set.

### Why
On iOS, when the app starts in landscape, orientation-change callbacks may not fire immediately. That can leave capture orientation stale until the user rotates the device. This patch ensures the current orientation is applied proactively and again on foreground resume.

### Scope
File changed: ipjsuaAppDelegate.m
No API changes.
No behavior change outside iOS sample orientation handling.

### Validation
Launch app while device is already in landscape (including charging-port-right case).
Start call/video and verify camera image is upright (head-up).
Background and foreground the app, then verify orientation remains correct without requiring a manual rotate.

### DIFF
--- a/pjsip-apps/src/pjsua/ios/ipjsua/ipjsuaAppDelegate.m
+++ b/pjsip-apps/src/pjsua/ios/ipjsua/ipjsuaAppDelegate.m
@@ -188,9 +188,39 @@ static void pjsip_funcs(void *user_data)
         UIDeviceOrientation dev_ori = [[UIDevice currentDevice] orientation];
         int i;
 
-        if (dev_ori == prev_ori) return;
+        /* At startup or resume, device orientation may be unknown even when
+         * UI orientation is valid. Try UI orientation as fallback.
+         */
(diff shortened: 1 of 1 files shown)


nanangizz (approved):

nanangizz:
Covered by #5017

PR #5017, opened 2026-06-12
videodev/darwin: fix video orientation for non-standard camera sensors

trengginas:
### **Problem**
PR #4751 introduced `AVCaptureDeviceRotationCoordinator` support for iOS 17+ to handle iPhone 17's portrait-native camera sensor. However it had two issues:

1. API contract violation: It queried `videoRotationAngleForHorizonLevelCapture` on every `set_cap` call and used that angle directly, overriding the requested `pjmedia_orient` with the physical gravity-aligned angle. When the phone was tilted to landscape, calling set_cap(ROTATE_90DEG) produced landscape frames instead of portrait, breaking the `PJMEDIA_VID_DEV_CAP_ORIENTATION` documentation guarantee that the video size stays the same.

2. Stripe regression: If set_cap(ROTATE_90DEG) was first called while the phone was in landscape, the landscape coordinator angle was used as the portrait angle, causing a dimension mismatch (e.g. 640×480 landscape into a 480×640 portrait buffer) producing a striped image. Reproduced on both iPhone 17 and older iPhones.

### **Fix**
Derive a per-stream `cam_portrait_angle` by combining the coordinator angle with the physical device tilt from `UIDevice`:
```
cam_portrait_angle = (coord_angle + device_tilt_CW) % 360
```

device_tilt_CW: 0=portrait, 90=LandscapeRight, 180=upsidedown, 270=LandscapeLeft. This cancels out the physical orientation component, leaving a hardware constant that is the same regardless of when set_cap is first called.

The actual videoRotationAngle for any requested pjmedia_orient:
```
rotation = (cam_portrait_angle - 90 + cap_ori[idx] + 360) % 360
```

Results:

- Older iPhones (landscape-native sensor, cam_portrait_angle=90): conventional 0/90/180/270 angles, identical to before.
- iPhone 17 (portrait-native sensor, cam_portrait_angle=0): correct device-specific angles (0° for portrait, 90° for landscape, etc.).

The USE_HORIZON_LEVEL compile-time macro is removed; the `@available(iOS 17.0, *)` guard is the only version gate needed.

### DIFF
--- a/pjmedia/src/pjmedia-videodev/darwin_dev.m
+++ b/pjmedia/src/pjmedia-videodev/darwin_dev.m
@@ -39,10 +39,6 @@
 #define DEFAULT_WIDTH           640
 #define DEFAULT_HEIGHT          480
 #define DEFAULT_FPS             15
-/* When the camera orientation changes, maintain horizon-level capture
- * relative to gravity.
- */
-#define USE_HORIZON_LEVEL       1
 
 /* Define whether we should maintain the aspect ratio when rotating the image.
  * For more details, please refer to util.h.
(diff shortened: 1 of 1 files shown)


sauwming:
Is this still draft or ready for review? Is this supposed to be replacement or on top of #5014?

trengginas:
> Is this still draft or ready for review? Is this supposed to be replacement or on top of #5014?

Yes it is ready. With this patch, #5014 is no longer needed.

nanangizz (approved):

sauwming (approved):

PR #5021, opened 2026-06-15
CI Windows: pin to windows-2022 and resolve VsDevCmd via vswhere

nanangizz:
## Problem

The `windows-latest` (and `windows-2025`) GitHub-hosted runner image now serves the **`windows-2025-vs2026`** image (Visual Studio 2026, GA June 2026). VS 2026 installs under `C:\Program Files\Microsoft Visual Studio\18\Enterprise` (folder is the major version `18`, not the year).

The CI Windows workflow hardcoded the VS 2022 path:

```
%PROGRAMFILES%\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat
```

so every job failed at the `check VsDevCmd.bat` / `MSBuild` steps with *"The system cannot find the path specified."*

Example failing run: https://github.com/pjsip/pjproject/actions/runs/27520938661

## Fix

- **Pin all Windows jobs to `windows-2022`** to keep the known-good VS 2022 / `v143` toolset environment. VS 2026 build compatibility for the bundled `.sln`/`.vcxproj` files is unverified, so this avoids betting on it while restoring green CI.
- **Resolve `VsDevCmd.bat` dynamically via `vswhere`** instead of hardcoding the edition/year path. On `windows-2022`, `vswhere -latest` resolves the only VS present, so behavior is identical to before — but a future forced migration to VS 2026+ won't reintroduce this exact breakage.

## Follow-up (not in this PR)

`windows-2022` has no retirement notice today but will eventually be removed. Worth validating a VS 2026 build (toolset, `-Wall`-level warnings) separately so the migration is proactive. Once green, the `windows-2022` pin can be dropped back to `windows-latest` — the `vswhere` logic already handles the path.

Co-Authored-By Claude Code

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -54,7 +54,7 @@ jobs:
             - '.github/**'
 
   default:
-    runs-on: windows-latest
+    runs-on: windows-2022
     name: Default / build only
     steps:
     - uses: actions/checkout@master
@@ -81,7 +81,7 @@ jobs:
       shell: cmd
     - name: MSBuild
       run: |
-        call "%PROGRAMFILES%\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat"
+        for /f "usebackq tokens=*" %%i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath`) do call "%%i\Common7\Tools\VsDevCmd.bat"
         msbuild pjproject-vs14.sln /p:PlatformToolset=v143 /p:Configuration=Debug /p:Platform=win32
       shell: cmd
     - name: build swig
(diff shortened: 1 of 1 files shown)


nanangizz:
### Re: Copilot review — not actioning the fail-fast suggestions (by design)

Copilot's comments all reduce to one point: the `for /f ... do call/dir` construct is a no-op if `vswhere` returns no `installationPath`, so a step could exit 0 without the VS environment being set.

This is technically correct but intentionally left as-is, because an empty `vswhere` result still produces a **hard, traceable failure** rather than a misleading green build:

- The `check VsDevCmd.bat` steps are diagnostics only — they gate nothing.
- The real backstop is the `MSBuild` step: if `vswhere` returns nothing, `VsDevCmd.bat` is never called, so `msbuild` is not on PATH (the `windows-2022` image doesn't add it by default) and the step fails with `'msbuild' is not recognized` (errorlevel 9009). For the `UseEnv=true` jobs there's a second backstop — base `INCLUDE`/`LIB` would be unset and the compile fails on missing standard headers.
- The empty-result case effectively can't occur on the pinned `windows-2022` image: `vswhere` is at a fixed Microsoft-owned path and VS 2022 is installed, so `-latest` resolves it. If a future image removes VS, that's surfaced loudly at the MSBuild step anyway.

So adding `if not defined / if not exist ... exit 1` guards to ~10 steps would only change the failure *location* (check step vs. build step), not turn a silent pass into a failure — there is no silent-pass path. Keeping the steps minimal for now.

sauwming (approved):

PR #5026, opened 2026-06-17
Add buffer bounds check in print_media_type()

nanangizz:
print_media_type() in pjsip/src/pjsip/sip_msg.c writes "type/subtype" into the output buffer without checking that it fits, relying entirely on callers to provide a large enough buffer.

The internal callers — pjsip_ctype_hdr_print() and pjsip_msg_print() — validate the buffer beforehand, so SIP message serialization is unaffected. However, the public pjsip_media_type_print() API forwards the caller's buffer straight through with no check, so an undersized buffer there would overflow.

This adds an explicit length check before the writes, consistent with the other header printers in this file. The trailing media parameters are already bounds-checked by pjsip_param_print_on().

Not a remotely-triggerable issue (no attacker-controlled path through the library); this is defensive hardening of the public API.

Co-Authored-By Claude Code

### DIFF
--- a/pjsip/src/pjsip/sip_msg.c
+++ b/pjsip/src/pjsip/sip_msg.c
@@ -1402,7 +1402,11 @@ static int print_media_type(char *buf, unsigned len,
     const pjsip_parser_const_t *pc;
 
     /* Check buffer for "type/subtype"; params are checked below. */
-    if ((pj_ssize_t)len < media->type.slen + 1 + media->subtype.slen)
+    if (media->type.slen < 0 || media->subtype.slen < 0)
+        return -1;
+    if ((pj_size_t)len <= (pj_size_t)media->type.slen)
+        return -1;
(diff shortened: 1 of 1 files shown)


pjsip/src/pjsip/sip_msg.c
```diff
@@ -1401,6 +1401,14 @@ static int print_media_type(char *buf, unsigned len,
     pj_ssize_t printed;
     const pjsip_parser_const_t *pc;
 
+    /* Check buffer for "type/subtype"; params are checked below. */
```
sauwming, line 1408: Since `print_media_type()` can now return -1, any usage of `p += print_media_type(p, ...)` may potentially move the pointer backward instead

nanangizz: Theoretically, it should never return -1 for internal callers because they validate the buffer length beforehand. Anyway, added the checks.

sauwming (approved):

PR #5046, opened 2026-06-29
Fix video stripes after camera switch on iOS with portrait angle fix

trengginas:
Continuing #5017.

When switching cameras via `PJMEDIA_VID_DEV_CAP_SWITCH`, `cam_portrait_angle` was not reset, causing the new camera's rotation to be computed using the stale angle calibrated for the previous camera. Reset it to -1 so it is recalibrated against the new device on the next orientation set.


nanangizz:
Check log (and add one if none) about camera switching? In pjsua and/or pjmedia layer perhaps.

nanangizz (approved):

sauwming (approved):

PR #5070, opened 2026-07-12
pjsua: Return PJ_ETOOMANY instead of asserting when account table full

### LINKED ISSUE #5069: pjsua: Return PJ_ETOOMANY instead of asserting when account table full
### Summary

`pjsua_acc_add()` guards the account-table-full condition with an assert:

```c
PJ_ASSERT_RETURN(pjsua_var.acc_cnt < PJ_ARRAY_SIZE(pjsua_var.acc),
                 PJ_ETOOMANY);
```

(`pjsip/src/pjsua-lib/pjsua_acc.c`, ~line 788 at `c1ea7648`.)

In release builds this returns `PJ_ETOOMANY`; in debug builds `pj_assert()` **aborts the
process**. Whether the table is full is a **runtime condition driven by application input**
(how many accounts the app adds), not a programmer invariant — multi-account applications hit
it organically, especially under `PJ_CONFIG_IPHONE`, whose sample config pins
`PJSUA_MAX_ACC` to 4. Asserts are conventionally reserved for invariants the programmer
controls; capacity exhaustion from input should be an error return in every build type.

### Inconsistency with the calls limit

The codebase already treats the analogous `PJSUA_MAX_CALLS` capacity condition as a plain
error path, in both directions:

- `pjsua_call_make_call()` at the limit: plain `if` → returns `PJ_ETOOMANY`;
- an incoming INVITE at the limit: stateless **`486 Busy Here`** on the wire.

Neither asserts in any build type. `pjsua_acc_add()` is the outlier.

### Suggested fix

Replace the capacity assert with a plain `if (...) return PJ_ETOOMANY;`, mirroring
`pjsua_call_make_call()`. The neighbouring `PJ_ASSERT_RETURN(cfg, PJ_EINVAL)` (a genuine
null-pointer invariant) and the later free-slot-scan `PJ_ASSERT_ON_FAIL` stay as asserts.
Release behaviour is unchanged; only the debug-build abort is removed.

A PR implementing exactly this follows shortly (already built it — on my fork:
https://github.com/laconicman/pjproject/pull/2).

### Repro

Debug build with a small `PJSUA_MAX_ACC` (e.g. `PJ_CONFIG_IPHONE`): add
`PJSUA_MAX_ACC + 1` accounts →

```
Assertion failed: (pjsua_var.acc_cnt < (sizeof(pjsua_var.acc)/sizeof(pjsua_var.acc[0]))),
function pjsua_acc_add, file pjsua_acc.c
```

→ abort. A release build returns `PJ_ETOOMANY` for the same sequence.

### PJSIP version

2.17

### Context

Seen on iOS, but it is definitely platform agnostic issue.

### Log, call stack, etc

```shell
PJ_ASSERT
```

laconicman:
Fixes #5069

### Problem

`pjsua_acc_add()` guards the account-table-full condition with
`PJ_ASSERT_RETURN(pjsua_var.acc_cnt < PJ_ARRAY_SIZE(pjsua_var.acc), PJ_ETOOMANY)`.
In debug builds `pj_assert()` aborts the process; in release it returns `PJ_ETOOMANY`.
The table being full is a runtime condition driven by application input (easy to hit under
`PJ_CONFIG_IPHONE`, where `PJSUA_MAX_ACC` is 4), not a programmer invariant — see the linked
issue for the full rationale.

### Fix

Replace the capacity assert with a plain `if (...) return PJ_ETOOMANY;`, mirroring how
`pjsua_call_make_call()` already handles the analogous `PJSUA_MAX_CALLS` limit (plain `if` →
`PJ_ETOOMANY`; incoming INVITE at the limit → stateless `486 Busy Here` — neither asserts).
The neighbouring `PJ_ASSERT_RETURN(cfg, PJ_EINVAL)` (genuine null-pointer invariant) and the
later free-slot-scan `PJ_ASSERT_ON_FAIL` are deliberately left as asserts.

Release behaviour is unchanged (`PJ_ASSERT_RETURN` already returned `PJ_ETOOMANY` there);
only the debug-build abort is removed. No networking/RFC-visible behaviour is affected.

### Validation

- Manual repro: debug build, add `PJSUA_MAX_ACC + 1` accounts — previously aborted, now
  returns `PJ_ETOOMANY`; release behaviour identical before/after.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -785,8 +785,15 @@ PJ_DEF(pj_status_t) pjsua_acc_add( const pjsua_acc_config *cfg,
     pj_status_t status = PJ_SUCCESS;
 
     PJ_ASSERT_RETURN(cfg, PJ_EINVAL);
-    PJ_ASSERT_RETURN(pjsua_var.acc_cnt < PJ_ARRAY_SIZE(pjsua_var.acc),
-                     PJ_ETOOMANY);
+
+    /* The account table being full is a runtime condition driven by how many
+     * accounts the application adds, not a programmer invariant, so return an
+     * error in all build configurations instead of asserting (which aborts in
+     * debug builds). This mirrors how pjsua_call_make_call() handles the
+     * analogous PJSUA_MAX_CALLS limit.
+     */
+    if (pjsua_var.acc_cnt >= PJ_ARRAY_SIZE(pjsua_var.acc))
+        return PJ_ETOOMANY;
 
 #if !PJ_HAS_IPV6
     PJ_ASSERT_RETURN(cfg->ipv6_sip_use == PJSUA_IPV6_DISABLED, PJ_EINVAL);
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -786,12 +786,6 @@ PJ_DEF(pj_status_t) pjsua_acc_add( const pjsua_acc_config *cfg,
 
     PJ_ASSERT_RETURN(cfg, PJ_EINVAL);
 
-    /* The account table being full is a runtime condition driven by how many
-     * accounts the application adds, not a programmer invariant, so return an
-     * error in all build configurations instead of asserting (which aborts in
-     * debug builds). This mirrors how pjsua_call_make_call() handles the
-     * analogous PJSUA_MAX_CALLS limit.
-     */
     if (pjsua_var.acc_cnt >= PJ_ARRAY_SIZE(pjsua_var.acc))
         return PJ_ETOOMANY;
 


sauwming (approved):

pjsip/src/pjsua-lib/pjsua_acc.c
```diff
@@ -785,8 +785,15 @@ PJ_DEF(pj_status_t) pjsua_acc_add( const pjsua_acc_config *cfg,
     pj_status_t status = PJ_SUCCESS;
 
     PJ_ASSERT_RETURN(cfg, PJ_EINVAL);
-    PJ_ASSERT_RETURN(pjsua_var.acc_cnt < PJ_ARRAY_SIZE(pjsua_var.acc),
-                     PJ_ETOOMANY);
+
+    /* The account table being full is a runtime condition driven by how many
```
sauwming: just minor suggestion: no need to compose an entire paragraph of comments for such very obvious two lines of code

nanangizz (approved):

pjsip/src/pjsua-lib/pjsua_acc.c
laconicman: Agree

laconicman: Deleted that verbose comment.

PR #5072, opened 2026-07-13
Add TLS 1.3 session resumption to OpenSSL backend

sauwming:
## Summary

Adds TLS session resumption to the OpenSSL secure-socket backend, controlled by a new opt-in param `pj_ssl_sock_param.enable_session_reuse` (default off).

Previously every outgoing TLS connection performed a full handshake and the server side actively disabled session tickets. TLS 1.3 delivers the resumption ticket via a `NewSessionTicket` message **after** the handshake, and each outgoing connection uses its own `SSL_CTX`, so captured sessions are held in a **process-global, mutex-protected cache keyed by server name** and re-injected on the next connection to the same host.

When `enable_session_reuse` is set:
- **Client** caches sessions per `server_name` (via `SSL_CTX_sess_set_new_cb`) and resumes them (`SSL_set_session`), avoiding a full handshake.
- **Server** issues session tickets (including TLS 1.3 tickets), which were previously always disabled.

## Changes

- **`ssl_sock.h` / `config.h`**: new `enable_session_reuse` param, `session_reused` info field, and `PJ_SSL_SOCK_OSSL_SESS_CACHE_SIZE` (default 64).
- **`ssl_sock_ossl.c`**: session cache + helpers (correct `SSL_SESSION` ref-counting, LRU eviction, `pj_atexit` cleanup for ASan), new-session callback, session injection in `ssl_create`, and gating of server ticket disabling on the flag. Guarded to OpenSSL ≥ 1.1.0.
- **`ssl_sock_imp_common.c`**: populate `info.session_reused` via `SSL_session_reused()`.
- **`pjlib-test/ssl_sock.c`**: `session_reuse_test()` opens a ticket-issuing server and makes two client connections, asserting the first is a full handshake and the second resumes.

## Testing

- OpenSSL backend and test compile warning-free (`-Wall`) against OpenSSL 3.4.1.
- `pjlib` + `pjlib-test` build and link cleanly.
- The new `session_reuse_test` runs as part of `pjlib-test ssl_sock` in OpenSSL-enabled builds (e.g. CI).

## Notes

- Opt-in; default behavior is unchanged.
- Empty `server_name` (e.g. connecting by IP) → resumption is skipped.
- Not wired into `pjsip_tls_setting` in this change; lives at the pjlib `pj_ssl_sock` level.

### DIFF
--- a/pjlib/include/pj/config.h
+++ b/pjlib/include/pj/config.h
@@ -1275,6 +1275,18 @@
 
 
 /**
+ * Maximum number of TLS client sessions cached for resumption by the OpenSSL
+ * backend when pj_ssl_sock_param.enable_session_reuse is set. Sessions are
+ * cached per server name; the oldest entry is evicted when the cache is full.
+ *
+ * Default: 64
+ */
+#ifndef PJ_SSL_SOCK_OSSL_SESS_CACHE_SIZE
+#   define PJ_SSL_SOCK_OSSL_SESS_CACHE_SIZE   64
+#endif
+
+
+/**
  * Disable WSAECONNRESET error for UDP sockets on Win32 platforms. See
  * https://github.com/pjsip/pjproject/issues/1197.
  *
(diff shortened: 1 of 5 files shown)


nanangizz (approved):

sauwming:
@copilot: resolve conflicts

PR #5079, opened 2026-07-15
Add best effort to obtain stack trace in stress test deadlock

sauwming:
Follow-up to the `stress / asan` hang on PR #5074 (run 29330620040): the job hit the 45-minute timeout with no ASan report, no assertion, and no crash — the stress binary deadlocked ~160 s into the 360 s stress phase and wrote nothing for its final 40 minutes (three `Timed-out trying to acquire PJSUA mutex` warnings from `pjsua_call_hangup()`/`pjsua_call_set_hold()`, then silence). The same hang hit `stress / tsan` on an unrelated PR (#5070) the day before, so it is a pre-existing intermittent deadlock. Two changes to make the next occurrence diagnosable:

## 1. Fix global log decor race in `log_call_dump()` (pjsua_media.c)

`log_call_dump()` saved, cleared, and restored the **process-global** log decor around the chunked call-stats dump. With concurrent call teardowns, one thread can save the decor already cleared by another and restore that — permanently stripping timestamps and newlines from all subsequent output. In the failing run this hit ~85 s in: the rest of the 142 MB log has no timestamps and ends in a single 661 KB line, which made the hang forensics much harder.

The dump is now printed in chunks split at line boundaries (same ~3 log calls per dump as before), so each chunk is a normal decorated log message and no global state is touched.

## 2. Add a hang watchdog to the stress CI jobs (ci-linux.yml, stress-test.yml)

The stress binary now runs in the background (log streamed to the console via `tail --pid`) with a watchdog alongside: if the binary is still alive after 900 s (PR jobs; healthy runs take ~8 min) / 1500 s (weekly; ~20 min), the watchdog appends `gdb -p <pid> -batch -ex "thread apply all bt"` output to `stress.log`, kills the process, and fails the step with exit 124 and an error annotation. `gdb` added to the apt installs. The app's real exit code is still propagated (via `wait` instead of `PIPESTATUS`); crash-backtrace decoding and the sanitizer log checks are unchanged.

A deadlock now costs 15 minutes instead of 45 and uploads thread stacks instead of nothing.

## Validation

- `make` clean (zero warnings) with the pjsua_media.c change; chunking loop additionally exercised by a standalone ASan-built test (content preserved across chunk boundaries, oversize-single-line hard-split fallback, no-trailing-newline and lone-newline edge cases).
- Both workflows YAML-parse; all three embedded scripts pass `bash -n`; the watchdog control flow was exercised with a scaled-down harness for all three paths (normal exit → pass, crash → exit code propagated, hang → stacks in `stress.log`, exit 124).

Co-Authored-By Claude Code

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -730,7 +730,7 @@ jobs:
         # at 5 min by the binary, stress phase is 6 min), so 15 is generous.
         ./pjsua-stress-$ARCH -n 512 -v 16 -t 4 --duration 360 --log-level 4 > stress.log 2>&1 &
         app_pid=$!
-        tail --pid=$app_pid -f stress.log &
+        tail --pid=$app_pid -n +1 -f stress.log &
         (
           sleep 900
           if kill -0 $app_pid 2>/dev/null; then
@@ -821,7 +821,7 @@ jobs:
         # at 5 min by the binary, stress phase is 6 min), so 15 is generous.
         ./pjsua-stress-$ARCH -n 512 -v 16 -t 4 --duration 360 --log-level 4 > stress.log 2>&1 &
         app_pid=$!
-        tail --pid=$app_pid -f stress.log &
+        tail --pid=$app_pid -n +1 -f stress.log &
         (
           sleep 900
           if kill -0 $app_pid 2>/dev/null; then
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -701,7 +701,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: install dependencies
-      run: sudo apt-get update && sudo apt-get install -y nasm
+      run: sudo apt-get update && sudo apt-get install -y nasm gdb
     - name: get openh264
       run: git clone --depth 1 --single-branch --branch openh264v2.1.0 https://github.com/cisco/openh264.git
     - name: build openh264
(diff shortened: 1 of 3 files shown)


nanangizz (approved):

arch7tect:
We observed this exact failure mode in production with PJSIP 2.17 on Linux, using PJSUA2 file logging at level 4.

At the transition, two calls entered media deinitialization and emitted detailed call dumps concurrently:

```text
22:30:03.212 pjsua_call.c   Call 184 hanging up
22:30:03.212 pjsua_media.c  Call 184: deinitializing media
22:30:03.212 pjsua_media.c  Call 186: deinitializing media
[overlapping call dumps]
22:30:03.213 conference.c   Remove port 8 queued
```

Immediately after the second dump, every subsequent `PJ_LOG` event in the same process lost both its timestamp prefix and automatic newline. Ordinary events were concatenated into long physical lines, while SIP message dumps remained multiline only because the SIP payloads contained their own CRLFs. The process continued handling calls for more than a day and the log decor never recovered.

The observed transition matches the race described in this PR exactly: dump A saves the normal decor and clears it; dump B starts while it is cleared and saves zero; A restores the normal value; B restores zero permanently.

Besides damaging the raw log, this caused timestamp-based downstream normalization to silently skip all subsequent calls.

Thank you for fixing this. Since 2.17 is currently the latest release and still contains the old `log_call_dump()` implementation, could you please consider backporting this change to a 2.17.x maintenance release, or explicitly include it in the next release notes?

sauwming:
This will be included in 2.18 but won't be backported to 2.17.x, so you can just cherry pick this PR and apply it against your version.

PR #5086, opened 2026-07-16
tests: add pjsua TLS, SIPS, and UDP->TCP MTU switch-over tests

sauwming:
## Summary
- Adds `tests/pjsua/scripts-call/220_tls.py` — basic call over TLS transport, using new self-signed cert/key fixtures in `tests/pjsua/certs/` (the tracked `pjlib/build/privkey.pem` is password-protected and fails to load on some OpenSSL builds).
- Adds `tests/pjsua/scripts-call/230_sips.py` — basic call using the `sips:` URI scheme. Required generalizing `mod_call.py`'s `"INVITE sip:"`/`"UPDATE sip:"` checks to also match `"sips:"`, and making `mod_call.py` only install its generic `test_func` when a config script hasn't already supplied its own via `TestParam(..., func=...)`.
- Adds `tests/pjsua/scripts-call/210_udp_to_tcp_mtu.py` — verifies RFC 3261 §18.1.1's automatic UDP→TCP switch-over when an outgoing request exceeds the UDP size threshold. Needed its own `test_func` since pjsua's sample app binds each auto-created local account to one specific transport, so the test explicitly does `acc next` to force the call to originate on the UDP account.
- The two TLS-dependent tests are skipped (not failed) when the pjsua build under test doesn't have TLS compiled in, via a new `inc_util.has_ssl_sock()` helper that inspects pjlib's generated `os_auto.h` (honoring a `config_site.h` override if present).

## Test plan
- [x] Ran all 5 `scripts-call` tests (`100_simplecall`, `200_tcp`, `210_udp_to_tcp_mtu`, `220_tls`, `230_sips`) via `run.py` against a locally built `pjsua` — all pass.
- [x] Verified the TLS/SIPS tests correctly report "is skipped" when simulating a `--disable-ssl` build (temporarily flipping `PJ_HAS_SSL_SOCK` in the generated `os_auto.h`), then confirmed they pass again on a normal build.
- [x] Re-ran `100_simplecall.py` and `200_tcp.py` as a regression check after the `mod_call.py` changes.

Co-Authored-By Claude Code


sauwming:
Stress test failure unrelated to the PR, triggered by the new deadlock detection in #5079. It will be addressed separately.

tests/pjsua/inc_util.py
```diff
@@ -1,5 +1,53 @@
...
+   val = find_value(site_header)
+   if val is not None:
+      return val
+   val = find_value(auto_header)
+   if val is not None:
+      return val
+   # Headers not found (e.g. unusual working directory): assume SSL is
+   # available rather than silently skipping the TLS/SIPS tests.
+   return True
+
```
nanangizz, line 32: Just to make sure, does this work with cmake?

Another approach is to parse the output of `pj_dump_config()` (unfortunately not exported to pjsua2/python) or `./pjsua --version`.

sauwming: Good catch — you're right, it didn't work with CMake, since CMake writes its own generated `os_auto.h` under the build directory (via `os_auto.h.cm`), not into the source tree, so this script had no reliable way to find it.

Switched to your second suggestion in f5b955774: `has_ssl_sock()` now runs `<exe> --version`, which calls `pj_dump_config()` and prints `PJ_HAS_SSL_SOCK : 0|1` straight from the compiled binary — works the same regardless of build system. `run.py` now exposes its resolved executable path via `inc_cfg.G_EXE` so the config scripts can pass it in.

nanangizz (approved):

PR #5104, opened 2026-07-21
pjmedia: Recreate Android MediaCodec video decoder on fatal error

trengginas:
When the hardware decoder hits an AMediaCodec error that leaves it in Error state (any codec using this Android MediaCodec wrapper - H264, VP8, VP9), the decoder is left permanently unable to dequeue input/output buffers with no path to recovery, even after the remote resends a keyframe.

Per the NDK, only a transient error may simply be retried later. Any other action code leaves the codec in Error state, recoverable either by reconfiguring (recoverable) or by deleting and recreating (fatal) the instance. Since delete + create + configure + start is a superset of the recoverable recovery sequence, both are handled the same way here: classify via `AMediaCodecActionCode_isTransient()` and recreate the decoder instance before the next decode attempt.

The error callback runs on MediaCodec's own internal thread, so it only raises a flag and the actual AMediaCodec calls are deferred to the decode path.

Also fixes a related use-after-free in `and_media_dealloc_codec()`, which deleted the codec instances and destroyed the buffer queues with the async callback still registered, then released the pool holding the codec data. The unregister + stop + delete sequence is now shared by the deallocation and the decoder recreation.

### DIFF
--- a/pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
+++ b/pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
@@ -230,6 +230,10 @@ typedef struct and_media_codec_data
     pj_bool_t                    format_changed;
     pjmedia_rect_size            new_size;
     int                          new_stride;
+
+    pj_bool_t                    dec_fatal_error;/**< Decoder hit a fatal,
+                                                       unrecoverable error and
+                                                       must be recreated.    */
 } and_media_codec_data;
 
 /* Custom callbacks. */
@@ -271,11 +275,19 @@ static void and_med_on_input_avail(AMediaCodec *codec,
     and_med_buf_info buf_info;
     pj_atomic_queue_t *buf_queue;
 
+    /* Match against the currently live enc/dec instance explicitly, rather
+     * than assuming "not enc => dec". The decoder can be recreated
+     * mid-stream (see recreate_decoder()), so a stale callback still in
+     * flight from an already-deleted decoder instance must be dropped here
+     * instead of being misfiled into the new instance's buffer queue.
+     */
     pj_bzero(&buf_info, sizeof(buf_info));
     if (codec == and_media_data->enc) {
         buf_queue = and_media_data->enc_avail_input_buf;
-    } else {
+    } else if (codec == and_media_data->dec) {
         buf_queue = and_media_data->dec_avail_input_buf;
+    } else {
+        return;
     }
     buf_info.index = index;
     pj_atomic_queue_put(buf_queue, &buf_info);
@@ -293,11 +305,14 @@ static void and_med_on_output_avail(AMediaCodec *codec,
     and_med_buf_info buf_info;
     pj_atomic_queue_t *buf_queue;
 
+    /* See and_med_on_input_avail() for why this must match explicitly. */
     pj_bzero(&buf_info, sizeof(buf_info));
     if (codec == and_media_data->enc) {
         buf_queue = and_media_data->enc_avail_output_buf;
-    } else {
+    } else if (codec == and_media_data->dec) {
         buf_queue = and_media_data->dec_avail_output_buf;
+    } else {
+        return;
     }
     buf_info.index = index;
     buf_info.size = bufferInfo->size;
@@ -345,6 +360,25 @@ static void and_med_on_error(AMediaCodec *codec,
                         "[%s] On Media error : err[%d] code[%d] msg[%s]\r\n",
                         (codec==and_media_data->enc)?"encoder":"decoder", error,
                         actionCode, detail);
+
+    if (codec != and_media_data->dec)
+        return;
+
+    /* Per NDK docs, if the action code is neither recoverable nor transient,
+     * the error is fatal and the codec instance must be deleted and
+     * recreated. Just raise a flag here; the actual AMediaCodec calls are
+     * deferred to the decode path since this callback runs on MediaCodec's
+     * own internal thread.
+     */
+#if __ANDROID_API__ >= 28
+    if (!AMediaCodecActionCode_isRecoverable(actionCode) &&
+        !AMediaCodecActionCode_isTransient(actionCode))
+    {
+        and_media_data->dec_fatal_error = PJ_TRUE;
+    }
+#else
+    and_media_data->dec_fatal_error = PJ_TRUE;
+#endif
 }
 
 /* Custom callback implementation. */
@@ -566,6 +600,63 @@ static pj_status_t configure_decoder(and_media_codec_data *and_media_data) {
         return PJMEDIA_CODEC_EFAILED;
     }
     return PJ_SUCCESS;
+}
+
+/* Recreate the decoder from scratch after a fatal, unrecoverable
+ * AMediaCodec error (see and_med_on_error()). A plain flush/restart is not
+ * enough for a fatal error per the NDK docs, the codec instance itself has
+ * to be deleted and recreated.
+ */
+static pj_status_t recreate_decoder(and_media_codec_data *and_media_data)
+{
+    and_med_buf_info buf_info;
+    pj_str_t *dec_name;
+    pj_status_t status;
+
+    PJ_LOG(3, (THIS_FILE, "Recreating decoder after fatal codec error"));
+
+    if (and_media_data->dec) {
+        if (API_AT_LEAST(28)) {
+            AMediaCodecOnAsyncNotifyCallback null_cb = {NULL, NULL, NULL, NULL};
+
+            AMediaCodec_setAsyncNotifyCallback(and_media_data->dec, null_cb,
+                                               NULL);
+        }
+        AMediaCodec_stop(and_media_data->dec);
+        AMediaCodec_delete(and_media_data->dec);
+        and_media_data->dec = NULL;
+    }
+
+    /* Discard any buffer indices left over from the old codec instance. */
+    while (pj_atomic_queue_get(and_media_data->dec_avail_input_buf,
+                               &buf_info) == PJ_SUCCESS)
+    {}
+    while (pj_atomic_queue_get(and_media_data->dec_avail_output_buf,
+                               &buf_info) == PJ_SUCCESS)
+    {}
+
+    dec_name = and_media_codec[and_media_data->codec_idx].decoder_name;
+    and_media_data->dec = AMediaCodec_createCodecByName(dec_name->ptr);
+    if (!and_media_data->dec) {
+        PJ_LOG(4, (THIS_FILE, "Failed recreating decoder: %s", dec_name->ptr));
+        return PJMEDIA_CODEC_EFAILED;
+    }
+
+    if (API_AT_LEAST(28)) {
+        AMediaCodecOnAsyncNotifyCallback async_cb = {&and_med_on_input_avail,
+                                                     &and_med_on_output_avail,
+                                                     &and_med_on_format_changed,
+                                                     &and_med_on_error};
+
+        AMediaCodec_setAsyncNotifyCallback(and_media_data->dec, async_cb,
+                                           and_media_data);
+    }
+
+    status = configure_decoder(and_media_data);
+    if (status == PJ_SUCCESS)
+        and_media_data->dec_fatal_error = PJ_FALSE;
+
+    return status;
 }
 
 PJ_DEF(pj_status_t) pjmedia_codec_and_media_vid_init(
@@ -1543,6 +1634,13 @@ static pj_status_t and_media_codec_decode(pjmedia_vid_codec *codec,
     PJ_ASSERT_RETURN(output->buf, PJ_EINVAL);
 
     and_media_data = (and_media_codec_data*) codec->codec_data;
+
+    if (and_media_data->dec_fatal_error) {
+        status = recreate_decoder(and_media_data);
+        if (status != PJ_SUCCESS)
+            return status;
+    }
+
     and_media_data->dec_has_output_frame = PJ_FALSE;
     and_media_data->dec_input_buf = NULL;
     and_media_data->dec_input_buf_len = 0;
--- a/pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
+++ b/pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
@@ -69,6 +69,14 @@
 #define MAX_RX_WIDTH            1280
 #define MAX_RX_HEIGHT           800
 
+/* Minimum interval between two decoder recreation attempts, and the maximum
+ * number of consecutive failed attempts before giving up. Re-acquiring a
+ * reclaimed codec can keep failing for as long as the reclaiming app holds
+ * it, so the attempts must not be made once per decoded frame.
+ */
+#define RECREATE_DEC_INTERVAL   1000
+#define RECREATE_DEC_MAX_RETRY  30
+
 /* Maximum duration from one key frame to the next (in seconds). */
 #define KEYFRAME_INTERVAL       1
 
@@ -202,7 +210,7 @@ typedef struct and_media_codec_data
     void                        *ex_data;
 
     /* Encoder state */
-    AMediaCodec                 *enc;
+    AMediaCodec         * volatile enc;
     unsigned                     enc_input_size;
     pj_uint8_t                  *enc_frame_whole;
     unsigned                     enc_frame_size;
@@ -211,7 +219,7 @@ typedef struct and_media_codec_data
     int                          enc_output_buf_idx;
 
     /* Decoder state */
-    AMediaCodec                 *dec;
+    AMediaCodec         * volatile dec;
     pj_uint8_t                  *dec_buf;
     pj_uint8_t                  *dec_input_buf;
     unsigned                     dec_input_buf_len;
@@ -234,6 +242,10 @@ typedef struct and_media_codec_data
     volatile pj_bool_t           dec_fatal_error;/**< Decoder hit an error
                                                        that leaves it unusable
                                                        and must be recreated.*/
+    pj_timestamp                 dec_recreate_time;/**< Last recreation
+                                                        attempt.            */
+    unsigned                     dec_recreate_cnt;/**< Consecutive failed
+                                                       recreation attempts. */
 } and_media_codec_data;
 
 /* Custom callbacks. */
@@ -277,9 +289,12 @@ static void and_med_on_input_avail(AMediaCodec *codec,
 
     /* Match against the currently live enc/dec instance explicitly, rather
      * than assuming "not enc => dec". The decoder can be recreated
-     * mid-stream (see recreate_decoder()), so a stale callback still in
-     * flight from an already-deleted decoder instance must be dropped here
-     * instead of being misfiled into the new instance's buffer queue.
+     * mid-stream (see recreate_decoder()), so ->dec may be NULL here, and a
+     * callback for the encoder must not be misfiled into the decoder queue.
+     * Note that this cannot by itself reject a callback of an already
+     * deleted instance, as the address may be reused. What guarantees that
+     * is destroy_codec_instance(), which unregisters the callback before
+     * deleting the instance.
      */
     pj_bzero(&buf_info, sizeof(buf_info));
     if (codec == and_media_data->enc) {
@@ -601,11 +616,16 @@ static pj_status_t configure_decoder(and_media_codec_data *and_media_data) {
     return PJ_SUCCESS;
 }
 
-/* Stop and delete a codec instance. The async callback is unregistered first,
- * otherwise a callback already dispatched by MediaCodec's own thread may still
- * reference the codec data after this returns.
+/* Delete a codec instance. The async callback is unregistered first, otherwise
+ * a callback already dispatched by MediaCodec's own thread may still reference
+ * the codec data after this returns.
+ *
+ * Set stop to PJ_FALSE if the instance has not been started or is already in
+ * the Error state, where stop() would just return an error after a needless
+ * round trip to the codec's looper.
  */
-static void destroy_codec_instance(AMediaCodec **codec)
+static void destroy_codec_instance(AMediaCodec * volatile *codec,
+                                   pj_bool_t stop)
 {
     if (!*codec)
         return;
@@ -613,9 +633,14 @@ static void destroy_codec_instance(AMediaCodec **codec)
     if (API_AT_LEAST(28)) {
         AMediaCodecOnAsyncNotifyCallback null_cb = {NULL, NULL, NULL, NULL};
 
+        /* The NULL userdata is required, not just tidiness: the dispatcher
+         * only skips the callback if both it and the userdata are NULL.
+         */
         AMediaCodec_setAsyncNotifyCallback(*codec, null_cb, NULL);
     }
-    AMediaCodec_stop(*codec);
+    if (stop)
+        AMediaCodec_stop(*codec);
+
     AMediaCodec_delete(*codec);
     *codec = NULL;
 }
@@ -624,7 +649,7 @@ static void destroy_codec_instance(AMediaCodec **codec)
  * in Error state (see and_med_on_error()). A plain flush/restart is not enough
  * per the NDK docs, the codec instance itself has to be deleted and recreated.
  */
-static pj_status_t recreate_decoder(and_media_codec_data *and_media_data)
+static pj_status_t do_recreate_decoder(and_media_codec_data *and_media_data)
 {
     and_med_buf_info buf_info;
     pj_str_t *dec_name;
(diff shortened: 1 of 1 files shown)


pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -345,6 +360,25 @@ static void and_med_on_error(AMediaCodec *codec,
...
+     */
+#if __ANDROID_API__ >= 28
+    if (!AMediaCodecActionCode_isRecoverable(actionCode) &&
+        !AMediaCodecActionCode_isTransient(actionCode))
+    {
+        and_media_data->dec_fatal_error = PJ_TRUE;
+    }
+#else
+    and_media_data->dec_fatal_error = PJ_TRUE;
+#endif
```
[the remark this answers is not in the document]
sauwming: Agreed, fixed in 03adf81. Switched to the runtime `API_AT_LEAST(28)` guard already used elsewhere in this file.

This was the most impactful one: our Android build targets `android23`, so `__ANDROID_API__` is 23 and the `#else` was what actually compiled, marking every decoder error fatal. A transient `AMEDIACODEC_ERROR_INSUFFICIENT_RESOURCE` (another app briefly taking a hardware decoder) would have torn down and recreated the decoder, blacking out video until the peer sent a new keyframe, where the old code self-healed.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -568,6 +602,63 @@ static pj_status_t configure_decoder(and_media_codec_data *and_media_data) {
...
+                                                     &and_med_on_format_changed,
+                                                     &and_med_on_error};
+
+        AMediaCodec_setAsyncNotifyCallback(and_media_data->dec, async_cb,
+                                           and_media_data);
+    }
+
+    status = configure_decoder(and_media_data);
+    if (status == PJ_SUCCESS)
+        and_media_data->dec_fatal_error = PJ_FALSE;
```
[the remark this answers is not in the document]
sauwming: Agreed on both points, fixed in 03adf81.

The clearing order was the more serious half: the async callbacks are installed on the new instance *before* `AMediaCodec_configure()`/`start()` run inside `configure_decoder()`, so a fatal error raised by the replacement during setup was set on the callback thread and then unconditionally cleared afterwards, permanently marking a dead decoder healthy. The flag is now cleared before `configure_decoder()` and re-set only on failure.

For the race, the flag is now `volatile`. It is an aligned single word so it will not tear on ARM/x86; the practical risk was the read in the decode loop being cached across iterations at -O2. `volatile` is proportionate here, `pj_atomic_t` would be heavier than this needs.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -568,6 +602,63 @@ static pj_status_t configure_decoder(and_media_codec_data *and_media_data) {
...
+    }
+
+    if (API_AT_LEAST(28)) {
+        AMediaCodecOnAsyncNotifyCallback async_cb = {&and_med_on_input_avail,
+                                                     &and_med_on_output_avail,
+                                                     &and_med_on_format_changed,
+                                                     &and_med_on_error};
+
+        AMediaCodec_setAsyncNotifyCallback(and_media_data->dec, async_cb,
+                                           and_media_data);
```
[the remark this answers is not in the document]
sauwming: Agreed, fixed in 03adf81. `recreate_decoder()` now returns `PJMEDIA_CODEC_EFAILED` if registration fails.

Worth noting the asymmetry with the identical call in `and_media_codec_open()`: there, ignoring it merely fails the open. Here it would have left a permanently mute decoder with the fatal flag cleared, so no retry would ever be attempted.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -271,11 +275,19 @@ static void and_med_on_input_avail(AMediaCodec *codec,
...
+     * instead of being misfiled into the new instance's buffer queue.
+     */
     pj_bzero(&buf_info, sizeof(buf_info));
     if (codec == and_media_data->enc) {
         buf_queue = and_media_data->enc_avail_input_buf;
-    } else {
+    } else if (codec == and_media_data->dec) {
         buf_queue = and_media_data->dec_avail_input_buf;
+    } else {
+        return;
```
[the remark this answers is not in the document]
sauwming: Not acting on this one.

The null-out just above quiesces the old instance synchronously: AOSP `NdkMediaCodec.cpp` dispatches under `mAsyncCallbackLock` and skips dispatch entirely when both the callback function and the userdata are NULL, which is exactly what is passed here. So a callback surviving past `AMediaCodec_setAsyncNotifyCallback(dec, null_cb, NULL)`, which the address-reuse scenario requires, cannot happen, and generation-tagged userdata would add machinery for a case the NDK already prevents.

The `codec != and_media_data->dec` checks are still worth keeping as a defensive guard, so leaving them as they are.

sauwming:
Reviewed. The approach is right - the diagnosis is correct and the delete + recreate strategy matches the NDK recovery contract. The stale-callback guards in `and_med_on_input_avail()`/`and_med_on_output_avail()` are sound too. Three of the four Copilot comments were valid, and I pushed 03adf81 with those fixes plus two related issues found while going through it. Replies are inline on each thread.

Addressed from the review:

1. `#if __ANDROID_API__ >= 28` is a compile-time check. Our Android build targets `android23`, so the `#else` was what compiled and every decoder error was treated as fatal, even though `pjmedia_codec_and_media_vid_init()` already refuses to init below device API 28. Now uses the runtime `API_AT_LEAST(28)` guard.
2. `dec_fatal_error` was cleared after `configure_decoder()`, but the async callbacks are installed before it runs, so a fatal error raised by the replacement instance during configure/start was silently swallowed and a dead decoder was marked healthy. Cleared before now, re-set only on failure. Also marked `volatile`, it is written by the MediaCodec callback thread.
3. `AMediaCodec_setAsyncNotifyCallback()` return value ignored in `recreate_decoder()`. The decode path is fed exclusively by these callbacks, so a failed registration meant a permanently mute decoder with no retry. Now fails the recreation.

Declined the fourth (pointer-identity race on `dec`) - the null-out already quiesces the old instance synchronously, details in that thread.

Two more found on top:

4. Recoverable errors had no recovery path at all. The NDK contract for `AMediaCodecActionCode_isRecoverable()` is "the codec is in Error state and can only be recovered by `stop()` + `configure()` + `start()`", not "the codec is fine". Skipping those left the decoder wedged in Error state with permanently black video, which is the same symptom this PR sets out to fix. Since `recreate_decoder()` is a strict superset of that sequence, the condition is now just `!isTransient()`. This is a small scope increase over the original PR, flagging it explicitly.
5. `recreate_decoder()` drained the buffer queues but left `dec_input_buf_idx`/`dec_input_buf_max_size` pointing at the destroyed instance. `and_media_codec_decode()` resets only `dec_input_buf` and `dec_input_buf_len`, so in whole-frame packing mode a packet exceeding the stale max size would reach `AMediaCodec_queueInputBuffer()` with an index never dequeued from the new codec. Both are reset now.

Builds clean with `-Wall -Werror=unguarded-availability` against `aarch64-none-linux-android23` on NDK r29. That flag is itself the confirmation for (1): the `#if` form would not have satisfied it on an API 28 symbol.

Unrelated and pre-existing, not for this PR: `and_media_dealloc_codec()` deletes both codecs without nulling the async callback first, then releases the pool holding `and_media_data`. Same stale-callback hazard the comments here describe, but with a use-after-free as the consequence rather than a misfiled buffer index. Worth a separate fix.

sauwming:
Folded the `and_media_dealloc_codec()` use-after-free into this PR as well (84dcab8), since it is the same stale-callback hazard as the one the guards here already handle.

The function deleted both codec instances and destroyed the four buffer queues with the async callback still registered, then released the pool holding `and_media_codec_data`. A callback already dispatched on MediaCodec's own thread would reference freed memory, either the destroyed `pj_atomic_queue_t` in `and_med_on_input_avail()`/`and_med_on_output_avail()` or the pool-allocated codec data itself. Unlike the misfiled buffer index this PR was originally about, that is a genuine use-after-free.

The unregister + stop + delete sequence is now in `destroy_codec_instance()`, used for the encoder and the decoder in `and_media_dealloc_codec()` and for the decoder in `recreate_decoder()`, so both paths quiesce the instance the same way. Left the throwaway probe instance in `codec_exists()` alone, it never has a callback registered.

Builds clean, same toolchain as before.

sauwming (approved):

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -1028,19 +1147,15 @@ static pj_status_t and_media_dealloc_codec(pjmedia_vid_codec_factory *factory,
...
         and_media_data->enc_avail_input_buf = NULL;
         pj_atomic_queue_destroy(and_media_data->enc_avail_output_buf);
         and_media_data->enc_avail_output_buf = NULL;
     }
 
     if (and_media_data->dec) {
-        AMediaCodec_stop(and_media_data->dec);
-        AMediaCodec_delete(and_media_data->dec);
-        and_media_data->dec = NULL;
+        destroy_codec_instance(&and_media_data->dec);
```
nanangizz: **Medium — the atomic queues leak when `recreate_decoder()` leaves `->dec == NULL`.**

This teardown is guarded by `if (and_media_data->dec)` on the line above. Before this PR that guard could only be false on the `and_media_alloc_codec()` -> `on_error` path. Patch 1 makes `->dec == NULL` reachable at steady state: if `AMediaCodec_createCodecByName()` fails inside `recreate_decoder()` (line 646), `->dec` stays NULL for the rest of the codec's life.

Deallocating that codec then skips both `pj_atomic_queue_destroy()` calls, and that function does `delete atomic_queue->aQ` (`pjlib/src/pj/atomic_queue.cpp:157`) on a C++-heap `AtomicQueue` created with `new` (`atomic_queue.cpp:149`) — not pool memory, so `pj_pool_release()` below does not reclaim it. Two `AtomicQueue` objects plus their ring buffers leak per affected call.

This is not a corner case: `ERROR_RECLAIMED` is the fatal action code that motivates the PR, and it is precisely when re-acquiring the codec fails.

Suggest keying the teardown off the queue pointers rather than off `->dec`/`->enc`, e.g.:

```c
    destroy_codec_instance(&and_media_data->dec);
    if (and_media_data->dec_avail_input_buf) {
        pj_atomic_queue_destroy(and_media_data->dec_avail_input_buf);
        and_media_data->dec_avail_input_buf = NULL;
    }
    ...
```
(`destroy_codec_instance()` already no-ops on NULL.) Same shape for the encoder block above.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -1543,6 +1658,13 @@ static pj_status_t and_media_codec_decode(pjmedia_vid_codec *codec,
     PJ_ASSERT_RETURN(output->buf, PJ_EINVAL);
 
     and_media_data = (and_media_codec_data*) codec->codec_data;
+
+    if (and_media_data->dec_fatal_error) {
+        status = recreate_decoder(and_media_data);
+        if (status != PJ_SUCCESS)
+            return status;
+    }
```
nanangizz, line 1738: **Medium — unbounded per-frame recreation attempts, no backoff and no give-up.**

`dec_fatal_error` is only cleared at line 681, which is reached only after both `createCodecByName` and `setAsyncNotifyCallback` succeed, and it is re-set on `configure_decoder()` failure at line 685. So while recreation itself keeps failing, every single `and_media_codec_decode()` call re-runs the whole `stop + delete + createCodecByName + configure + start` sequence on the video decoding thread.

At ~30 fps that is 30 hardware-codec acquisition attempts per second, each a binder round-trip to mediaserver costing tens of ms, plus a `PJ_LOG(3)` "Recreating decoder after codec error" (line 633), a `PJ_LOG(4)`, and a `LOGERR_` from `vid_stream.c:1206` per frame. The decode thread stalls and the jitter buffer backs up.

Again `ERROR_RECLAIMED` is the motivating case and exactly the one where re-acquisition keeps failing until the reclaiming app releases the codec — so the retry storm is the expected path, not the unlucky one.

Suggest an attempt counter or a time-based backoff (e.g. don't retry within N ms of the last failed attempt, and give up after M attempts, leaving the flag set so the stream just reports no output).

The early return itself is correct — no NULL-deref of `->dec` downstream, and `vid_stream.c:1205-1209` sets `PJMEDIA_FRAME_TYPE_NONE` for us.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -271,11 +275,19 @@ static void and_med_on_input_avail(AMediaCodec *codec,
     and_med_buf_info buf_info;
     pj_atomic_queue_t *buf_queue;
 
+    /* Match against the currently live enc/dec instance explicitly, rather
+     * than assuming "not enc => dec". The decoder can be recreated
+     * mid-stream (see recreate_decoder()), so a stale callback still in
+     * flight from an already-deleted decoder instance must be dropped here
+     * instead of being misfiled into the new instance's buffer queue.
+     */
```
nanangizz, line 298: **Low — this comment overclaims, and `->dec` is now a cross-thread gate that was left non-`volatile`.**

Two related points on the same code:

1. The comment says the explicit match drops "a stale callback still in flight from an already-deleted decoder instance". Pointer identity cannot give that guarantee: `AMediaCodec_delete()` followed by `AMediaCodec_createCodecByName()` can hand back the same heap address, so a genuinely stale `codec` pointer may compare *equal* to the new instance and be misfiled anyway. The real protection is the unregister-before-delete you added in `destroy_codec_instance()` — the NDK holds `mAsyncCallbackLock` across the user-callback invocation, so `setAsyncNotifyCallback()` blocks until an in-flight dispatch returns. Suggest saying that instead; the explicit enc/dec match is still worth keeping, just for the reason it actually serves (not assuming "not enc => dec" when `->dec` is NULL).

2. Patch 2 correctly marked `dec_fatal_error` `volatile` because the MediaCodec callback thread writes it. But `->dec` — which that same thread now reads here (line 287), in `and_med_on_output_avail()` (line 312) and in `and_med_on_error()` (line 364) — is a plain field written by the decode thread in `recreate_decoder()`. In practice the NDK mutex taken by `setAsyncNotifyCallback()` right after the store supplies the barrier, so I don't think this is a live failure; flagging it for consistency with the `dec_fatal_error` treatment.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -568,6 +601,92 @@ static pj_status_t configure_decoder(and_media_codec_data *and_media_data) {
...
+{
+    if (!*codec)
+        return;
+
+    if (API_AT_LEAST(28)) {
+        AMediaCodecOnAsyncNotifyCallback null_cb = {NULL, NULL, NULL, NULL};
+
+        AMediaCodec_setAsyncNotifyCallback(*codec, null_cb, NULL);
+    }
+    AMediaCodec_stop(*codec);
```
nanangizz: **Low — `AMediaCodec_stop()` on an instance that is already in the Error state.**

For the fatal path this helper was added to serve, the documented MediaCodec contract is to release the instance directly: `stop()` in the Error/Uninitialized state returns `INVALID_OPERATION`, so this is dead weight — and it is a synchronous round-trip to the codec's looper on the video decoding thread, on the path that is already the slow one.

Same applies to the other caller: `and_media_alloc_codec()`'s `on_error` reaches here with an instance that was never configured or started, where `stop()` is likewise a no-op error.

Harmless today (it is dispatched and replied to without entering the component), but skipping `stop()` when tearing down after an error would match the NDK contract and drop a needless round-trip.

Separately on this hunk: unsetting with `{NULL,NULL,NULL,NULL}` plus **NULL** userdata is safe, but only because Android's dispatcher guards on `userdata != NULL || fn != NULL` and would call a NULL fn if userdata were non-NULL. The NULL userdata is load-bearing — worth a one-line comment so it does not get "tidied" later.

nanangizz (commented):
Reviewed the full diff. The core design is right, and I verified the two things it hinges on:

- Deferring the AMediaCodec calls out of the error callback onto the decode thread is the correct shape.
- Unregistering the async callback before `AMediaCodec_delete()` genuinely closes the teardown use-after-free — the NDK holds `mAsyncCallbackLock` across the user-callback invocation, so `setAsyncNotifyCallback()` blocks until an in-flight dispatch returns.

Also checked and happy with: the create -> assign `->dec` -> register-callback ordering in `recreate_decoder()` matches `create_codec()` + `and_media_codec_open()`, so the new explicit `codec == ->enc / == ->dec` match cannot drop legitimate callbacks; the `#if __ANDROID_API__ >= 28` -> `API_AT_LEAST(28)` change is correct and required (it expands to `__builtin_available(android x, *)`, the real clang availability guard for the `__INTRODUCED_IN(28)` call); keyframe recovery after a successful recreate works, since the first decode produces no output and the existing `PJMEDIA_EVENT_KEYFRAME_MISSING` publish fires the FIR/PLI; and no source files were added or removed, so the three build systems need no updates.

Five inline comments below: 2 Medium (an `AtomicQueue` leak, and an unbounded retry storm — both triggered by the `ERROR_RECLAIMED` case the PR targets) and 3 Low.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
```diff
@@ -568,6 +601,92 @@ static pj_status_t configure_decoder(and_media_codec_data *and_media_data) {
...
+                       "status=%d", am_status));
+            return PJMEDIA_CODEC_EFAILED;
+        }
+    }
+
+    /* Buffer indices of the deleted instance are meaningless now. */
+    and_media_data->dec_input_buf = NULL;
+    and_media_data->dec_input_buf_len = 0;
+    and_media_data->dec_input_buf_idx = 0;
+    and_media_data->dec_input_buf_max_size = 0;
```
nanangizz: **Low — these resets sit after the two early-return failure paths.**

On the `createCodecByName` failure (line 649) and the `setAsyncNotifyCallback` failure (line 668) we return before reaching here, so `dec_input_buf` still points into the deleted instance's buffer and `dec_input_buf_max_size` still describes it.

Nothing reads them on those paths today — `and_media_codec_decode()` returns at line 1665 before its own reset, and every other reader is downstream of that — so this is latent rather than live. Still, the resets belong immediately after `destroy_codec_instance()` at line 635, where they are unconditionally correct regardless of which failure path is taken.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
sauwming: Good catch, fixed in 2c4a353. The teardown is now keyed off the queue pointers rather than the codec instance, for both the encoder and the decoder, as suggested. `destroy_codec_instance()` already no-ops on NULL so the `if (->enc)` / `if (->dec)` wrappers are gone entirely.

Confirmed the leak path: `pj_atomic_queue_destroy()` does `delete atomic_queue->aQ` on a C++-heap object, so `pj_pool_release()` cannot reclaim it.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
sauwming: Agreed, fixed in 2c4a353. Rate limited to one attempt per second (`RECREATE_DEC_INTERVAL`), giving up after 30 consecutive failures (`RECREATE_DEC_MAX_RETRY`).

The throttling lives in a thin `recreate_decoder()` wrapper around the previous body, now `do_recreate_decoder()`, so the several early-return failure paths inside it all go through the same accounting. The counter resets on success, so a decoder that recovers and later fails again gets a full budget rather than a stale one.

On the give-up limit: I went with 30 rather than something tighter deliberately, since `ERROR_RECLAIMED` clears as soon as the reclaiming app releases the codec, and giving up means no video for the rest of the call. 30 s of failed re-acquisition is a reasonable point to stop hammering mediaserver, and it is logged at level 2 so it is visible in a normal log.

Note the first attempt after a fatal error is not delayed, only the retries.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
sauwming: Both taken, fixed in 2c4a353.

1. Rewrote the comment. It now says what the check actually buys (not assuming "not enc => dec", which matters precisely because `->dec` can be NULL) and points at `destroy_codec_instance()` as what actually rejects a stale instance.

2. Made `enc` and `dec` `AMediaCodec * volatile`. Qualifying the pointer rather than the pointee keeps every use site unchanged; only `destroy_codec_instance()` needed its parameter adjusted. Did both rather than just `dec` so the `destroy_codec_instance()` calls stay type-compatible without a cast, and both are read from the callback thread anyway.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
sauwming: Both taken, fixed in 2c4a353.

`destroy_codec_instance()` got a `stop` parameter. `recreate_decoder()` passes `PJ_FALSE`, since that instance is in the Error state. `and_media_dealloc_codec()` passes `PJ_TRUE`, since `and_media_codec_close()` is a no-op so that is the only real stop on the normal teardown path.

For the `and_media_alloc_codec()` `on_error` case you also mention: rather than teaching dealloc to tell the two apart, that path now deletes the two instances itself with `stop = PJ_FALSE` before calling dealloc, which then no-ops on them and goes on to destroy the queues.

Added the note that the NULL userdata is load-bearing.

pjmedia/src/pjmedia-codec/and_vid_mediacodec.cpp
sauwming: Agreed, fixed in 2c4a353. Moved the four resets up to immediately after `destroy_codec_instance()`, where they hold on every path out of the function.

sauwming:
Thanks for the review, all five are valid and all five are addressed in 2c4a353. Replies inline; summary:

**Medium — `AtomicQueue` leak (#1).** Correct, and the reachability argument is right: `recreate_decoder()` makes `->dec == NULL` a steady state. Teardown in `and_media_dealloc_codec()` is now keyed off the queue pointers for both the encoder and the decoder, and since `destroy_codec_instance()` no-ops on NULL the outer `if (->enc)` / `if (->dec)` guards are gone.

**Medium — retry storm (#2).** Correct, and it is the expected path for `ERROR_RECLAIMED` rather than the unlucky one. Now one attempt per second, giving up after 30 consecutive failures. The throttling is a thin `recreate_decoder()` wrapper around the old body (`do_recreate_decoder()`) so all its early-return failure paths share the accounting; the counter resets on success. First attempt after a fatal error is still immediate.

**Low — overclaiming comment and non-`volatile` `->dec` (#3).** Both fixed. The comment now states what the check actually buys and credits `destroy_codec_instance()` for the stale-instance guarantee. `enc`/`dec` are `AMediaCodec * volatile`, qualifying the pointer not the pointee, so no use site changed.

**Low — `stop()` on an Error-state instance (#4).** `destroy_codec_instance()` takes a `stop` flag. `PJ_FALSE` from the recreation path; `PJ_TRUE` from `and_media_dealloc_codec()`, which is the only real stop given `and_media_codec_close()` is a no-op. The `and_media_alloc_codec()` `on_error` case now deletes the instances itself with `stop = PJ_FALSE` first, so dealloc no-ops on them. Added the note that the NULL userdata is load-bearing.

**Low — resets after the early returns (#5).** Moved up to right after `destroy_codec_instance()`.

Also glad the two load-bearing points checked out independently — the `mAsyncCallbackLock` behaviour and the `__builtin_available` expansion were what the earlier round hinged on.

Builds clean, `aarch64-none-linux-android23`, NDK r29, `-Wall -Werror=unguarded-availability`.

nanangizz (approved):
Approving on 2c4a353. Read the follow-up diff rather than just the replies — all five points are addressed as described:

- Queue teardown keyed off the queue pointers, `destroy_codec_instance()` called unconditionally (no-ops on NULL). `and_media_data` is zalloc'd, so the pointer checks are safe on the alloc `on_error` path too.
- Throttle wrapper logic is right: the `cnt >= MAX` early return precedes any timestamp work, the interval check is gated on `cnt > 0` so the zeroed `dec_recreate_time` is never read, and the give-up log fires exactly once (30th failure; 31st+ hit the early return).
- `AMediaCodec * volatile` on the pointer rather than the pointee keeps every NDK call site unchanged; `destroy_codec_instance()`'s `AMediaCodec * volatile *` parameter matches the `&and_media_data->enc/dec` call sites.
- `stop` flag: `AMediaCodec_start()` is only reachable via `configure_encoder()`/`configure_decoder()`, which `create_codec()` never calls, so `PJ_FALSE` on the alloc `on_error` path is correct.
- Resets now sit right after `destroy_codec_instance()` and hold on every exit path.

Two residual nits, neither blocking, take or leave:

1. Every throttled/given-up decode returns `PJMEDIA_CODEC_EFAILED`, which `vid_stream.c:1206` logs per frame via `LOGERR_` — that is `PJ_PERROR(4, ...)`, so hidden at typical production levels and there is no binder work behind it, hence cosmetic. Returning `PJ_SUCCESS` with `output->type = PJMEDIA_FRAME_TYPE_NONE; output->size = 0` on the throttled path (like the existing "couldn't produce output" branch) would keep the log quiet.
2. Dealloc passes `stop = PJ_TRUE` for the decoder even when `dec_fatal_error` is set (i.e. the last failure was in `configure_decoder()`, so `->dec` is non-NULL but in Error state) — one harmless failing `stop()` round-trip at teardown. Could pass `!and_media_data->dec_fatal_error` for the decoder.

PR #5106, opened 2026-07-21
pjlib: fix race in pj_timer_heap_earliest_time and pj_timer_heap_count

trengginas:
## Summary
- Fix a race in `pj_timer_heap_earliest_time()`: the emptiness check (`cur_size == 0`) ran *before* acquiring the timer heap lock, then the function unconditionally read `heap[0]` after locking. A concurrent cancellation of the last scheduled timer between the check and the lock could shrink `cur_size` to 0 without clearing `heap[0]`, so the function could return `PJ_SUCCESS` with a stale/dangling timer value instead of `PJ_ENOTFOUND`.
- Move the emptiness check inside the lock, and drop the racy `pj_assert(ht->cur_size != 0)` (callers no longer need to call `pj_timer_heap_count()` first — the function safely reports `PJ_ENOTFOUND` on its own).
- Harden `pj_timer_heap_count()` to read `cur_size` under the lock instead of unlocked, for the same reason.
- Update the `pj_timer_heap_earliest_time()` doc comment in `timer.h` to reflect the new, safe-to-call-without-pre-checking contract.

## Why
`remove_node()` only decrements `cur_size` when removing the last heap entry — it doesn't clear `heap[0]`. So the old unlocked pre-check + locked-but-unconditional read was a genuine TOCTOU bug under multi-threaded use.

### DIFF
--- a/pjlib/include/pj/timer.h
+++ b/pjlib/include/pj/timer.h
@@ -338,16 +338,17 @@ PJ_DECL(int) pj_timer_heap_cancel_if_active(pj_timer_heap_t *ht,
 PJ_DECL(pj_size_t) pj_timer_heap_count( pj_timer_heap_t *ht );
 
 /**
- * Get the earliest time registered in the timer heap. The timer heap
- * MUST have at least one timer being scheduled (application should use
- * #pj_timer_heap_count() before calling this function).
+ * Get the earliest time registered in the timer heap. If no entry is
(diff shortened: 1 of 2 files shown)


nanangizz:
@trengginas
Ping!
Please review the Copilot comments, the first point appears to be a real bug.

nanangizz (approved):

sauwming (approved):

PR #5107, opened 2026-07-22
pjmedia: honor negotiated H264 packetization mode and enforce max bitrate

trengginas:
## Summary

- `oh264_codec_open()` previously ignored the SDP-negotiated `packetization-mode` and always forced single-NAL packetization ("better for compatibility"), even when the peer negotiated mode 1 (non-interleaved, FU-A capable). The packetizer mode, NAL size cap, and slice mode now follow what was actually negotiated:
  - `packetization-mode=0` → single-NAL packetization, size-limited slices so every NAL/slice fits the MTU.
  - `packetization-mode=1` → non-interleaved packetization, one slice per frame (better compression), relying on FU-A fragmentation for delivery over RTP.
- Seed the encoder's max bitrate ceiling (`iMaxBitrate` / `iMaxSpatialBitrate`) from the negotiated `max_bps` at codec-open time, instead of leaving it at OpenH264's library default until the first `oh264_codec_modify()` call. Clamp it to never fall below the target bitrate, per OpenH264's requirement.

## Why

Packetization mode 1 requires the receiving peer to support single-NAL, STAP-A, and FU-A packets — it doesn't mandate always fragmenting. But by always forcing single-NAL regardless of what was negotiated, we left compression efficiency on the table: many small MTU-sized slices per frame instead of one larger slice, even when the peer had already told us it could handle FU-A.

Separately, the max-bitrate ceiling was only ever wired up in the bitrate-modify path, not at initial codec setup — so any call that never explicitly triggers a bitrate change ran under OpenH264's internal default ceiling rather than the one actually agreed in SDP, for its entire duration.

## Trade-off to note for reviewers

Using one slice per frame under mode 1 improves compression, but reduces loss resilience: losing a single RTP/FU-A fragment now likely invalidates the whole frame, whereas the old size-limited-slice approach confined a lost packet's damage to one small slice. This is an intentional trade favoring bandwidth efficiency on the (common) case of a clean network; on lossy networks it may show more visible frame drops until the next keyframe/LTR recovery.

### DIFF
--- a/pjmedia/src/pjmedia-codec/openh264.cpp
+++ b/pjmedia/src/pjmedia-codec/openh264.cpp
@@ -459,25 +459,20 @@ static pj_status_t oh264_codec_open(pjmedia_vid_codec *codec,
 
     pj_bzero(&pktz_cfg, sizeof(pktz_cfg));
     pktz_cfg.mtu = param->enc_mtu;
-    /* Packetization mode */
-#if 0
+    /* Packetization mode: follow the mode negotiated in fmtp instead of
+     * always forcing single NAL mode. In mode 0 every NAL unit must fit
+     * in one RTP packet, so the encoder has to emit MTU-sized slices.
+     * In mode 1 the packetizer fragments large NAL units (FU-A), so the
+     * encoder can code one slice per frame, avoiding the compression
+     * penalty of many small slices. A peer that negotiated mode 1 is
+     * required to support FU-A, so this is safe.
+     */
     if (h264_fmtp.packetization_mode == 0)
         pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_SINGLE_NAL;
     else if (h264_fmtp.packetization_mode == 1)
         pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_NON_INTERLEAVED;
     else
         return PJ_ENOTSUP;
-#else
-    if (h264_fmtp.packetization_mode!=
-                                PJMEDIA_H264_PACKETIZER_MODE_SINGLE_NAL &&
-        h264_fmtp.packetization_mode!=
-                                PJMEDIA_H264_PACKETIZER_MODE_NON_INTERLEAVED)
-    {
-        return PJ_ENOTSUP;
-    }
-    /* Better always send in single NAL mode for better compatibility */
-    pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_SINGLE_NAL;
-#endif
 
     status = pjmedia_h264_packetizer_create(oh264_data->pool, &pktz_cfg,
                                             &oh264_data->pktz);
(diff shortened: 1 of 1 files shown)


jeffreym-emt:
This same behavior also exists in ffmpeg_vid_codecs.c::h264_preopen() 

```
static pj_status_t h264_preopen(ffmpeg_private *ff)
{
    h264_data *data;
    pjmedia_h264_packetizer_cfg pktz_cfg;
    pj_status_t status;

    data = PJ_POOL_ZALLOC_T(ff->pool, h264_data);
    ff->data = data;

    /* Parse remote fmtp */
    status = pjmedia_vid_codec_h264_parse_fmtp(&ff->param->enc_fmtp,
                                               &data->fmtp);
    if (status != PJ_SUCCESS)
        return status;

    /* Create packetizer */
    pktz_cfg.mtu = ff->param->enc_mtu;
    pktz_cfg.unpack_nal_start = 0;
#if 0
    if (data->fmtp.packetization_mode == 0)
        pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_SINGLE_NAL;
    else if (data->fmtp.packetization_mode == 1)
        pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_NON_INTERLEAVED;
    else
        return PJ_ENOTSUP;
#else
    if (data->fmtp.packetization_mode!=
                                PJMEDIA_H264_PACKETIZER_MODE_SINGLE_NAL &&
        data->fmtp.packetization_mode!=
                                PJMEDIA_H264_PACKETIZER_MODE_NON_INTERLEAVED)
    {
        return PJ_ENOTSUP;
    }
    /* Better always send in single NAL mode for better compatibility */
    pktz_cfg.mode = PJMEDIA_H264_PACKETIZER_MODE_SINGLE_NAL;
#endif

    status = pjmedia_h264_packetizer_create(ff->pool, &pktz_cfg, &data->pktz);
    if (status != PJ_SUCCESS)
        return status;
```

sauwming (commented):
Reviewed the two changes bundled here separately. The bitrate one is a real gap and worth fixing; the packetization mode one is not a bug and won't be addressed.

### `iMaxBitrate` at codec open — real gap

`GetDefaultParams()` leaves `iMaxBitrate` at `UNSPECIFIED_BIT_RATE`, so the negotiated `max_bps` is ignored for the entire call unless `oh264_codec_modify()` happens to be called. Worth fixing, and it matters more now that #5121 derives the send pacing budget from `max_bps`.

As written, though, this regresses the default path. `max_bps` is only ever clamped *downward* (`vid_stream_info.c:181`, `vid_codec_util.c:722`) from the codec default, and here that default is `DEFAULT_MAX_BITRATE` = 256000 == `DEFAULT_AVG_BITRATE` = 256000 (lines 52-53). So unless the app or the remote SDP says otherwise, this sets a ceiling exactly equal to the target — zero burst headroom, with `bEnableFrameSkip = 1` set a few lines below. Keyframes and scene changes are where that shows.

The clamp direction is also wrong: raising `iMaxBitrate` up to `iTargetBitrate` means transmitting above a ceiling the remote negotiated via `b=TIAS`/`max-br`. Lowering the target to the ceiling respects it and satisfies OpenH264 equally well.

Both points are handled in #5236, which takes this half with the headroom fixed.

### Packetization mode — not a bug, won't be addressed

Forcing single-NAL is **not** a defect. RFC 6184 mode 1 *permits* single NAL unit, STAP-A and FU-A packets; it does not require a sender to fragment. Sending only single-NAL packets to a mode-1 peer is fully compliant, so there is nothing incorrect to fix here.

Nor is the `#if 0` a stale workaround someone forgot to remove — it has been there since the original wrapper commit (b3034d498) as a deliberate interop choice, and no issue has been filed against it since.

The only argument for changing it is compression efficiency: `SM_SIZELIMITED_SLICE` resets prediction and entropy contexts at every slice boundary, which costs something on large keyframes. That cost is unquantified here, and against it sits a concrete regression risk — `dec_fmtp` advertises `packetization-mode=1` (line ~295) and negotiation takes min(offer, answer) (`vid_codec_util.c:531`), so this flips the default path for every mode-1 peer, and with `SM_SINGLE_SLICE` one lost fragment invalidates the whole frame instead of one small slice. `ffmpeg_vid_codecs.c::h264_preopen()` also carries the identical block (as @jeffreym-emt noted), so the two H.264 encoders would diverge.

Trading a deliberate interop default for an unmeasured bitrate saving isn't a good deal, so this half is being dropped rather than reworked.

sauwming:
no follow up by author, review addressed in #5236

PR #5124, opened 2026-07-29
pjsua: report media stream index in incoming text and DTMF callbacks

### LINKED ISSUE #5099: [PJSUA] realtime text and DTMF callback don't say which stream received it
### Describe the bug

I'm currently playing around with realtime text (on master) and noticed and asymmetry in PJSUA.

When I have a call with two (or more) text streams and I want to send text,
I can tell [pjsua_call_send_text()](https://github.com/pjsip/pjproject/blob/2bfce5d463a3b9d993d488caab6f180c089415e0/pjsip/src/pjsua-lib/pjsua_txt.c#L29) what stream to send the text on explicitly
by setting the media index in the ["send text parameter"](https://github.com/pjsip/pjproject/blob/2bfce5d463a3b9d993d488caab6f180c089415e0/pjsip/include/pjsua-lib/pjsua.h#L6334).

On the receiving side however, the receive callback [doesn't carry the media index](https://github.com/pjsip/pjproject/blob/2bfce5d463a3b9d993d488caab6f180c089415e0/pjsip/include/pjsua-lib/pjsua.h#L1109)
and so the application can't tell on what stream the text was actually received.

Having two or more (text) streams is not far fetched.
In recording or conference calls, you can have multiple audio and text streams.

While looking at this, I noticed that the DTMF part lacks the same info.

### Steps to reproduce

- Use PJSUA to receive a call with two or more text streams.
- Have the remote endpoint send text on all streams
- Try to tell the received text fragments apart by stream

### PJSIP version

master

### Context

Like stated above, I was playing around with the API and noticed this discrepancy.
I believe this to be an oversight.

### Log, call stack, etc

```shell
I didn't collect any logs.
```

andreas-wehrmann:
I had most changes be generated after a back and forth with Claude, reviewed the changes and tested them in my testapp.
The call ID and media index are now packed into a scalar and provided as "user data" to the callback functions.
New helper macros have been introduced for packing/unpacking the parameters.

Fixes #5099.

Everything below this line was generated by Claude

---

pjsua_call_send_text() lets the sender select the text stream via med_idx,
but the receive callbacks could not tell the application which stream incoming text (or DTMF) arrived on
which is ambiguous for calls with multiple streams of the same type.

- Add med_idx to pjsua_txt_stream_data (corresponding to pjsua_call_send_text_param.med_idx on the send side), and to pjsua_dtmf_info / pjsua_dtmf_event (-1 when the DTMF was not stream-borne, i.e. SIP INFO).
- The call id and media index are packed into the stream callback user_data — call id in the lower 16 bits, media index in the upper 16. Internal helper macros (PJSUA_MED_UDATA_PACK,
PJSUA_MED_UDATA_CALL_ID, PJSUA_MED_UDATA_MED_IDX) do the packing/unpacking. The callbacks stay lock-free on the media thread and dereference no shared state, so the reported index stays correct across re-INVITE stream recreation and teardown; 32 packed bits fit the 
pointer on all supported platforms.
- PJSUA2: mirror as medIdx in OnCallRxTextParam, OnDtmfDigitParam, OnDtmfEventParam
- PJSUA2: fix Call::sendText() dropping CallSendTextParam.medIdx.
- pjsua app: log the stream index for received text/DTMF.

### DIFF
--- a/pjsip-apps/src/pjsua/pjsua_app.c
+++ b/pjsip-apps/src/pjsua/pjsua_app.c
@@ -747,8 +747,9 @@ static void call_on_dtmf_callback2(pjsua_call_id call_id,
                          info->duration);
         break;
     };    
-    PJ_LOG(3,(THIS_FILE, "Incoming DTMF on call %d: %c%s, using %s method", 
-           call_id, info->digit, duration, method));
+    PJ_LOG(3,(THIS_FILE, "Incoming DTMF on call %d: %c%s, using %s method, "
+           "stream %d", call_id, info->digit, duration, method,
+           info->med_idx));
 }
 
 /* Incoming text stream callback. */
--- a/pjsip-apps/src/pjsua/pjsua_app.c
+++ b/pjsip-apps/src/pjsua/pjsua_app.c
@@ -757,10 +757,10 @@ static void call_on_rx_text(pjsua_call_id call_id,
                             const pjsua_txt_stream_data *data)
 {
     if (data->text.slen == 0) {
-        PJ_LOG(4, (THIS_FILE, "Received empty T140 block on stream %u "
+        PJ_LOG(4, (THIS_FILE, "Received empty T140 block on stream %d "
                               "with seq %d", data->med_idx, data->seq));
     } else {
-        PJ_LOG(3, (THIS_FILE, "Incoming text on call %d stream %u, "
+        PJ_LOG(3, (THIS_FILE, "Incoming text on call %d stream %d, "
                               "seq %d: %.*s (%d bytes)",
                               call_id, data->med_idx, data->seq,
                               (int)data->text.slen, data->text.ptr,
--- a/pjsip-apps/src/pjsua/pjsua_app.c
+++ b/pjsip-apps/src/pjsua/pjsua_app.c
@@ -756,11 +756,12 @@ static void call_on_rx_text(pjsua_call_id call_id,
                             const pjsua_txt_stream_data *data)
 {
     if (data->text.slen == 0) {
-        PJ_LOG(4, (THIS_FILE, "Received empty T140 block with seq %d",
-                              data->seq));
+        PJ_LOG(4, (THIS_FILE, "Received empty T140 block on stream %u "
+                              "with seq %d", data->med_idx, data->seq));
     } else {
-        PJ_LOG(3, (THIS_FILE, "Incoming text on call %d, seq %d: %.*s "
-                              "(%d bytes)", call_id, data->seq,
+        PJ_LOG(3, (THIS_FILE, "Incoming text on call %d stream %u, "
+                              "seq %d: %.*s (%d bytes)",
+                              call_id, data->med_idx, data->seq,
                               (int)data->text.slen, data->text.ptr,
                               (int)data->text.slen));
     }
--- a/pjsip/include/pjsua-lib/pjsua.h
+++ b/pjsip/include/pjsua-lib/pjsua.h
@@ -1051,6 +1051,12 @@ typedef struct pjsua_dtmf_info {
      */
     unsigned duration;
 
+    /**
+     * The media index of the audio stream that received the DTMF, or -1
+     * if the DTMF was not received via a media stream (e.g. SIP INFO).
+     */
+    int med_idx;
+
 } pjsua_dtmf_info;
 
 
@@ -1100,6 +1106,12 @@ typedef struct pjsua_dtmf_event {
      * an event with PJMEDIA_STREAM_DTMF_IS_END for every event.
      */
     unsigned flags;
+
+    /**
+     * The media index of the audio stream that received the DTMF, or -1
+     * if the DTMF was not received via a media stream (e.g. SIP INFO).
+     */
+    int med_idx;
 } pjsua_dtmf_event;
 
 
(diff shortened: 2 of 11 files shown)


nanangizz:
Packing the call ID and media index in` user_data` appears to be a simpler approach. The readability issue can be addressed via a helper function or macro with a descriptive name :)

That said, IMO this approach also works, it could be useful for future improvements when more information needs to be packed into `user_data`.

Here are some comments from Claude:

--

1. **Please make the new field `int` in all three structs.** Currently
   `pjsua_txt_stream_data.med_idx` is `unsigned` while `pjsua_dtmf_info` /
   `pjsua_dtmf_event` use `int` with `-1`, and the split propagates into pjsua2
   (`OnCallRxTextParam::medIdx` vs `OnDtmfDigitParam::medIdx`). Since the doc pairs the
   field with `pjsua_call_send_text_param.med_idx`, which is `int`, `int` everywhere is
   the consistent choice. This is public API, so it is the one item I would not defer.

3. **Please soften the comment in `pjsua_internal.h`.** Immutability does prevent
   reading torn or zeroed identity data, but it does not prevent a late callback from a
   destroyed stream being attributed to a recycled call slot — the `med_idx` would be
   right while `call_id` refers to a different call. That exposure is pre-existing and
   equally true of the old encoded `call->index`, so it is out of scope here, but "the
   reported index stays correct across re-INVITE stream recreation and teardown" reads
   as a stronger claim than the change makes.

4. **Minor:** `Endpoint::on_dtmf_digit()` and `on_dtmf_digit2()` are never registered —
   both are commented out at `endpoint.cpp:2354-2355`, only `on_dtmf_event` is wired,
   and the live path is `PendingOnDtmfEventCallback` → `prmBasic.medIdx`, which you got
   right. The hardcoded `medIdx = -1` in `on_dtmf_digit()` is therefore unreachable
   today, but would report stream-borne DTMF as "not from a stream" if anyone re-enabled
   it. A short comment there would avoid the trap.

5. **A test would be good to have.** #5098 recently added a DTMF-over-SIP-INFO case to
   `tests/pjsua` that could assert `med_idx == -1`.

andreas-wehrmann:
Ok done, the indices are now packed into the user data arg, the media index is int and I've updated the test.
Regarding pjsua2: Since the registration of on_dtmf_digit and on_dtmf_digit2 are commented out anyway, why not remove them?

sauwming (approved):

nanangizz:
Ah, you changed it to packing indices approach. Looks simpler indeed. Thanks.
Better also briefly mention this in the PR desc?

Asked Claude to review, it says `PJSUA_MED_UDATA_MED_IDX` silently converts -1 to 65535, but currently there is no usage/call with input `-1`, so this is minor (at least for now).

When I asked it to improve the macros, and here is its suggestion:
```
/* Pack/extract call id (lower 16 bits) and media index (upper 16 bits) in
 * media stream callback user data. Both are packed as 16-bit signed, so the
 * -1 sentinel of pjsua_dtmf_info.med_idx round-trips. Callers pack
 * pjsua_call.index and pjsua_call_media.idx, bounded by PJSUA_MAX_CALLS and
 * PJSUA_MAX_CALL_MEDIA, so the range is never exceeded.
 */
#define PJSUA_MED_UDATA_PACK(call_id, med_idx) \
            (pj_assert((int)(call_id) <= 0x7FFF && (int)(med_idx) <= 0x7FFF), \
             (void*)(((pj_size_t)(pj_uint16_t)(med_idx) << 16) | \
                     (pj_size_t)(pj_uint16_t)(call_id)))
#define PJSUA_MED_UDATA_CALL_ID(udata) \
            ((pjsua_call_id)(pj_int16_t)((pj_size_t)(udata) & 0xFFFF))
#define PJSUA_MED_UDATA_MED_IDX(udata) \
            ((int)(pj_int16_t)(((pj_size_t)(udata) >> 16) & 0xFFFF))
```

Fixes / enhancements:

- `-1` round-trips. `pj_int16_t` sign-extension on extract; `0xFFFF` no longer reads as `65535`.
- `MED_IDX` returns `int`, not `unsigned`. Matches the new public `med_idx` field's type; no sign flip hidden behind a cast.
- Casts (e.g: `(int)PJSUA_MED_UDATA_MED_IDX(user_data)`) dropped at 3 call sites. Macro type now matches the local.
- `CALL_ID` sign-extends too. Symmetric; `PJSUA_INVALID_ID` survives a round-trip.
- Bounds asserted on pack. Catches a caller exceeding 16 bits in debug builds.
- Contract documented. Comment states the signed-16-bit encoding and which bounds keep it valid.
- Pure expression preserved. Comma operator keeps it usable as a function argument; `pj_assert` compiles out under `NDEBUG`.

andreas-wehrmann:
You were right, the macros make the scalar approach easier overall which is why I went with it.
I added the assert but am also asserting on `>= 0` because negative values should never occur in the places where the macro is used which is also why I dismissed this when I was working on it.
Negative values would survive in release mode though, where asserts are disabled.

andreas-wehrmann:
I made this a draft again because of the failing builds.
Some builds seem to be failing because if type mismatches; I didn't get those on my Linux build.
I will build this on Windows and see what I can find and fix.

andreas-wehrmann:
The type mismatches were actually not related to the changes in this PR.
As stated in the commit message, the pack macro broke when compiling with PJ_DEBUG == 0 in which case
`pj_assert` became the  `do{}while(0)` construct which no longer compiled with the pack macro.
So instead of macros I now used little inline helper functions which I hope is okay.
I could have kept the unpack macros but changed them to inline funcs just for consistency.

nanangizz (approved):

PR #5130, opened 2026-08-03
Fix NaN/Inf percent loss in siprtp print_avg_stat (#5110)

### LINKED ISSUE #5110: siprtp print_avg_stat() can print NaN/Inf percent loss when packet counters are zero
### Describe the bug

In `pjsip-apps/src/samples/siprtp.c`, `print_avg_stat()` only checks that at least one call is confirmed/active. It does not ensure RX/TX packet counters are non-zero before computing loss percentages:
```c
min_stat.rx.loss*100.0/(min_stat.rx.pkt+min_stat.rx.loss)
...
min_stat.tx.loss*100.0/(min_stat.tx.pkt)
```
For a newly confirmed call with no RTP yet, denominators can be zero, so the printed "percent loss" becomes NaN (0.0/0.0) or Inf (non-zero/0).
This is a sample-app statistics correctness bug, not a memory-safety issue.

### Steps to reproduce

1. Build pjproject samples (includes `siprtp`).
2. Run two `siprtp` instances and establish a call.
3. Print average stats while the call is confirmed but RX/TX packet counters are still zero (or otherwise force zero pkt/loss counters).
4. Observe NaN/Inf in the "percent loss" fields.
Equivalent formula check (same expressions as in `print_avg_stat`):
rx: loss*100.0/(pkt+loss) with pkt=0, loss=0 -> NaN
tx: loss*100.0/pkt with pkt=0, loss=0 -> NaN
tx: loss*100.0/pkt with pkt=0, loss>0 -> Inf

### PJSIP version

2.17-dev

### Context

- Platform: macOS aarch64, clang 17
- Component: sample app `pjsip-apps/src/samples/siprtp.c`, function `print_avg_stat`
- Trigger: confirmed call with zero RTP/RTCP packet counters

### Log, call stack, etc

```shell
rx_percent_loss = NaN   (0/0)
tx_percent_loss = NaN   (0/0)
tx_percent_loss = Inf   (loss>0, pkt==0)
```

sauwming:
## Description

Fixes #5110.

In `pjsip-apps/src/samples/siprtp.c`, `print_avg_stat()` computed RX/TX percent loss without ensuring the packet counters were non-zero:

```c
min_stat.rx.loss*100.0/(min_stat.rx.pkt+min_stat.rx.loss)
...
min_stat.tx.loss*100.0/(min_stat.tx.pkt)
```

For a newly confirmed call with no RTP yet, the denominators can be zero, so the printed "percent loss" becomes `NaN` (0/0) or `Inf` (loss>0, pkt==0).

## Fix

Guard the divisions with a `PERCENT_LOSS()` helper that returns `0.0` when the denominator is zero, and precompute the six values before the `printf`.

This is a sample-app statistics correctness fix, not a memory-safety issue.

## Testing

Built the `siprtp` sample cleanly (no errors/warnings).

Co-Authored-By Claude Code

### DIFF
--- a/pjsip-apps/src/samples/siprtp.c
+++ b/pjsip-apps/src/samples/siprtp.c
@@ -1579,8 +1579,16 @@ static void print_avg_stat(void)
     char stx_min[16], stx_avg[16], stx_max[16];
     char btx_min[16], btx_avg[16], btx_max[16];
 
+    double rx_loss_min, rx_loss_avg, rx_loss_max;
+    double tx_loss_min, tx_loss_avg, tx_loss_max;
 
     unsigned i, count;
+
+    /* Compute percent loss, guarding against zero packet counters (which
+     * would otherwise produce NaN or Inf when no RTP has been received/sent
+     * yet on a confirmed call).
+     */
+#define PERCENT_LOSS(loss, denom)  ((denom) > 0 ? (loss)*100.0/(denom) : 0.0)
 
     pj_bzero(&call_dur, sizeof(call_dur)); 
     call_dur.min = BIGVAL;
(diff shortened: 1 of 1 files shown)


nanangizz (approved):

nanangizz:
IMHO, NaN/Inf is correct and honest as long as it does not cause crash :)

sauwming:
Yes, it shouldn't crash so it's harmless. However, NaN/Inf from div-by-zero is only guaranteed under IEEE 754, and `-ffast-math` / `-ffinite-math-only` (or non-IEEE targets) can change or optimize away that behavior.

PR #5133, opened 2026-08-03
docs: add minimal-comments guidance to coding conventions

sauwming:
Reading the code co-authored by AI agents starts giving me headache, due to their tendency to compose lengthy multi-line paragraphs of comments for every block of code. Compact should be preferable, to ensure code is still readable by humans.


andreas-wehrmann:
I just wanted to add: I've instructed Claude to produce small self-contained commits.
I find the code changes easier to understand and follow that way when I review the code changes.
The downside of this approach is of course, that there can be many commits; depending on the task.

nanangizz (approved):

PR #5148, opened 2026-08-05
pjlib-util: harden pj_scan_get_quotes() buffer bound checks

sauwming:
## Background

`pj_scan_get_quotes()` was not among the functions covered by the scanner hardening in c4d34984e (GHSA-fq45-m3f7-3mhj / CVE-2022-39244), so it still dereferences the scanner pointer without an explicit EOF check, and it has an off-by-one in the loop that counts escaping backslashes.

This applies the same guards the sibling scanner functions already use.

## Changes

- Bail out with a syntax error if the scanner is already at EOF, before probing for the `begin_quote`.
- Check `PJ_SCAN_CHECK_EOF()` before testing for the `end_quote`, both inside the scan loop and in the final "eat the end quote" check.
- Count the preceding backslashes with a counter walking down from `s`, bounded by the opening quote at `scanner->curptr`, instead of forming `s-2` and walking down to `scanner->begin`.

## Is this a security issue?

**No, and no advisory is planned.** Worth stating explicitly since this touches the same function family as a previous CVE.

The scanner API requires a NUL terminated buffer of at least `buflen+1` bytes:

> Note that the input string buffer MUST be NULL terminated and have length at least buflen+1 (buflen MUST NOT include the NULL terminator).
> — `pjlib-util/include/pjlib-util/scanner.h`

The SIP receive path guarantees this by NUL terminating each message fragment before parsing and restoring the byte afterwards (`pjsip/src/pjsip/sip_transport.c`), so a dereference at `scanner->end` reads the in-bounds NUL sentinel and correctly falls through to the syntax error path. No remote input reaches this function with a non-terminated buffer, and all in-tree callers (`sip_parser.c`, `sip_auth_parser.c`, `xml.c`, `json.c`, `cli.c`, `http_client.c`) inherit the same contract.

The changes make the bounds explicit rather than relying on the sentinel, and correct the backslash-counting bound. The latter is not reachable from any in-tree caller, since it needs the backslash itself to be passed as a `begin_quote` character. Nothing here is remotely reachable.

## Testing

No behaviour change for valid input. Verified against SIP, XML, JSON and CLI style quote pairs that well-formed tokens parse identically before and after, and that malformed ones still raise a syntax error — covering simple, empty, escaped-quote, escaped-backslash, unterminated, embedded-newline, missing-begin-quote, empty-buffer, already-at-EOF and mismatched-pair inputs, plus backslash-delimiter and mid-buffer (`scanner->curptr > scanner->begin`) cases for the backslash-counting bound.

The one intentional behaviour change is for `begin_quote == end_quote == '\'`: the two-byte buffer `\\` is now read as an empty quoted token, the analogue of `""`, rather than raising a syntax error. Previously the opening delimiter was miscounted as an escape. No in-tree caller uses `\` as a delimiter.

- `pjlib-util-test xml_test json_test` — pass
- `pjsip-test multipart_test txdata_test` — pass (`uri_test`/`msg_test` are compiled out on Apple + ASan, see `pjsip/src/test/test.h`)
- Clean build, no new warnings

Reported by Kanishka De Silva ([@hexer365](https://github.com/hexer365)).

Co-Authored-By Claude Code

### DIFF
--- a/pjlib-util/src/pjlib-util/scanner.c
+++ b/pjlib-util/src/pjlib-util/scanner.c
@@ -366,10 +366,17 @@ PJ_DEF(void) pj_scan_get_quotes(pj_scanner *scanner,
                                 int qsize, pj_str_t *out)
 {
     register char *s = scanner->curptr;
+    char *r;
+    unsigned nbackslash;
     int qpair = -1;
     int i;
 
     pj_assert(qsize > 0);
+
+    if (pj_scan_is_eof(scanner)) {
+        pj_scan_syntax_err(scanner);
+        return;
+    }
 
     /* Check and eat the begin_quote. */
     for (i = 0; i < qsize; ++i) {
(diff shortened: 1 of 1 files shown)


sauwming:
Thanks — both points were valid, and the second one turned out to be a real correctness bug, not just a style concern. Reworked the backslash counting.

`r > scanner->begin` stopped the *dereference* below the buffer start but not the out-of-range pointer *formation* at `s-2`, which is UB in its own right (C89 6.3.6). ASan doesn't flag pointer formation, only the read, which is why the patched build looked clean.

The bound is now the opening delimiter rather than the buffer start, using a counter so no pointer outside `[scanner->curptr + 1, s]` is ever formed:

```c
nbackslash = 0;
for (r = s; r > scanner->curptr + 1 && *(r-1) == '\\'; --r) {
    ++nbackslash;
}
if ((nbackslash & 0x01) == 0) {
    break;
}
++s;
```

`*(r-1)` is only evaluated when `r >= scanner->curptr + 2`, so it never reads the opening delimiter and never crosses into an already-parsed token when `scanner->curptr > scanner->begin`. Counting the total run and testing its parity also collapses the nested `if (*(s-1) == '\\') ... else` into one branch.

Your second observation has a visible consequence, so calling it out: with `begin_quote == end_quote == '\'`, the two-byte buffer `\\` is now read as an empty quoted token — the analogue of `""` — instead of raising a syntax error, because the opening `\` is no longer miscounted as an escape. That is the correct reading. No in-tree caller uses `\` as a delimiter.

Verified all real-delimiter cases (SIP `"`, XML `"`/`'`, CLI `{`) parse byte-identically before and after, including your exact example (`begin_quote='\'`, two-byte buffer `\"`) and a mid-buffer case with `scanner->curptr > scanner->begin`. `pjlib-util-test xml_test json_test` and `pjsip-test multipart_test txdata_test` pass.

nanangizz (approved):

PR #5149, opened 2026-08-05
pjsua: validate account Contact before storing it in acc_check_nat_addr()

sauwming:
## Problem

A field crash on Android arm64 (release build) with a NULL dereference in `acc_check_nat_addr()`:

```
#00  acc_check_nat_addr            pjsua_acc.c  (contact_hdr->uri)
#01  regc_cb                       pjsua_acc.c
#02  call_callback                 sip_reg.c:845
#03  regc_tsx_callback             sip_reg.c:1589
#04  mod_util_on_tsx_state         sip_util_statefull.c:80
```

`pjsip_parse_hdr()` returns NULL for the account's *own* stored Contact and the result is dereferenced under a `pj_assert()` that is compiled out in release.

The account's Contact becomes unparseable because `acc_check_nat_addr()` writes it itself, rebuilt from the Via `received`/`rport` of the registrar's response, with no validation:

* Via params are scanned with `pjsip_VIA_PARAM_SPEC` = `TOKEN_SPEC` + `"[:]"`, i.e. ``-.!%*_`'~+[:]`` (`sip_parser.c:388-390`, `TOKEN` at line 47), but a SIP URI host is scanned with `pjsip_HOST_SPEC` = alnum + `_-.` (`sip_parser.c:399-401`). Any of the extra characters in `received=` produces a Contact that cannot be parsed back.
* Square brackets for IPv6 were chosen from `tp->key.type & PJSIP_TRANSPORT_IPV6` rather than from the address itself. Note `int_parse_host()` strips brackets from `sent_by.host` and `pj_sockaddr_print(..., 0)` prints IPv6 unbracketed, so `via_addr` is essentially never bracketed on entry. An IPv6 `received` on a transport not typed IPv6 (dual-stack SBC, NAT64) was stored as `<sip:user@2001:db8::1:5060>`.

The poisoning pass does not crash — it parses the old, good Contact at the top and returns `PJ_TRUE`, which makes `regc_cb` return early with `contact_rewritten = PJ_TRUE`. The next response short-circuits the check and resets the flag. The crash lands two to three REGISTER round trips later, with nothing logged in between, which is consistent with an unreproducible field report.

`update_regc_contact()` has the identical unchecked pattern on the same string and is called from `acc_check_nat_addr()` *before* the crash site, so with `reg_contact_uri_params` configured it faults one round trip earlier.

### `pjsua_acc.c` — stop producing the bad Contact

* Bracket the address only when `pj_inet_pton(PJ_AF_INET6, ...)` accepts it. A looser "contains a colon" test is not enough: `received=foo:bar` is a legal Via param token, and `int_parse_host()` accepts any non-empty bracket contents, so `[foo:bar]` would parse and be stored as a meaningless Contact instead of the previous one being kept.
* Parse the rebuilt Contact back and decline the rewrite unless it is a valid SIP/SIPS Contact, keeping the previous good value instead of poisoning the account.
* Move the `PJSUA_CONTACT_REWRITE_UNREGISTER` step below the validation, so a declined rewrite cannot leave the account unregistered with a destroyed regc. It still runs before `acc->contact` is replaced, as that branch requires.
* Replace the `pj_assert()`s on the parse result in both `acc_check_nat_addr()` and `update_regc_contact()` with real checks.
* Reject a non-SIP URI **in `acc_check_nat_addr()`**, which casts the result to `pjsip_sip_uri*` and then reads `uri->port` / `uri->host` as struct fields. A `Contact: *` parses to a header with a NULL `uri`, and a `tel:` URI (reachable via `pjsua_acc_config.force_contact`) parses to a `pjsip_other_uri` whose `->host` would be read out of a mismatched struct. `update_regc_contact()` deliberately does **not** get this guard — it only passes the URI to `pjsip_uri_print()`, which dispatches on the URI's own vptr, so no SIP-specific field is read there. The local is declared `pjsip_uri*` instead, to stop the inert cast reading as type confusion.
* NULL terminate the `reg_contact` buffer in `update_regc_contact()`. It is built with `pj_strcpy()`/`pj_strcat()` from `pj_pool_alloc()` memory and then passed to `pjsip_parse_hdr()`, whose scanner requires a NULL terminated input (`scanner.h:219-224`); byte `[slen]` was uninitialized. The `pj_ansi_snprintf()` result is also checked for truncation before being used as `slen` by the subsequent `pj_strcat()`s.

### `sip_reg.c` — two NULL/state bugs the new test exposed

Both are pre-existing and independent of the crash above; the second is widened by the first.

* `set_contact()` checked only for a NULL parse result, then dereferenced `hdr->uri` — NULL for `Contact: *` — when `add_xuid_param` is enabled. `calculate_response_expiration()` dereferences it too. Reject a URI-less Contact: registering one is meaningless, and `pjsip_regc_unregister_all()` builds its starred Contact directly on the tdata rather than through `set_contact()`. This reproduces only in a combined test run, because `regc_test` toggles that global and leaves it set, which is presumably why it went unnoticed.
* `set_contact()` also moved the current Contacts to `removed_contact_hdr_list` with `expires=0` *before* parsing the new ones, so any error return partway left the regc with no active Contact and the old bindings queued for removal — the next REGISTER unregisters them even though the update was refused. All new Contacts are now parsed and validated into a local list first, and the regc lists are only touched once nothing can fail.

### `tsan.supp` — one suppression, for a pre-existing inversion

The new test is the first thing in `pjsip-test` to drive a successful contact rewrite, which exposes a lock-order inversion that makes the TSan CI job exit 66 with all tests passing:

```
M0 -> M1  regc_cb() holds PJSUA_LOCK -> acc_check_nat_addr() ->
          pjsua_acc_set_registration() -> pjsip_regc_send() ->
          pjsip_tsx_set_transport() takes the tsx grp_lock
M1 -> M0  regc_cb() / regc_tsx_cb() are called from tsx_set_state()
          with the tsx grp_lock held, and take PJSUA_LOCK
```

`pjsua_acc_set_registration()` does `PJSUA_UNLOCK()` before `pjsip_regc_send()` to avoid exactly this, but `PJSUA_LOCK` is a recursive mutex (`pjsua_core.c:1021`) and `regc_cb` already holds it, so the unlock only decrements the recursion count. Same problem as the TODO on the disabled `acc_delete_deferred_test` in `pjsua_auth_test.c`.

**This is not a test artifact and this PR does not fix it.** `acc_check_nat_addr()` re-registers for any `contact_rewrite_method` below `ALWAYS_UPDATE`, so every real NAT contact rewrite takes this path. Suppressed the way the two existing PJSUA lock-topology inversions in that file are, keyed on `acc_check_nat_addr` since that frame is unique to the cycle; `regc_cb` and `regc_tsx_cb` are left unsuppressed so other registration-path deadlocks still surface. Fixing it properly means deferring the re-registration out of the callback or making the unlock recursion-aware — a PJSUA lock topology change. Happy to open a separate issue.

## Testing

New `pjsua_acc_test` (registered in all three build systems). A mock registrar overwrites the Via `received` param of the REGISTER, which `pjsip_endpt_create_response()` clones into the 200 OK, letting the test drive `acc_check_nat_addr()` with an arbitrary address as seen by the registrar. Each case runs three REGISTER round trips, so a Contact stored on the first response is parsed back on a later one — the sequence that crashes.

| Case | Guards | Fails without the fix |
|---|---|---|
| IPv6 `received` over IPv4 transport | `acc_check_nat_addr` bracketing | **yes** — aborts on `contact_hdr != NULL` |
| plain IPv4 `received` | ordinary rewrite still happens | no |
| `received=1.2.3.4!` | `acc_check_nat_addr` validation | **yes** — same abort |
| `received=foo:bar` | `pj_inet_pton` bracket gate | **yes** — same abort |
| declined rewrite, `UNREGISTER` method | the reorder; regc not torn down | no |
| accepted rewrite + `reg_contact_uri_params` | `update_regc_contact` regeneration | no |
| `force_contact = "*"` | `update_regc_contact` fallback, `set_contact` NULL uri | **yes** — aborts on `pj_assert(uri != NULL)` |
| `force_contact = "<tel:...>"` + uri params | `acc_check_nat_addr` scheme guard | **yes** — aborts in `pj_inet_pton()` on a `pj_str_t` read out of a `pjsip_other_uri` |
| rejected `update_contact` keeps the binding | `set_contact` validate-before-mutate | **yes** — binding re-sent with `;expires=0` |

Every row was checked by reverting the relevant fix and re-running, not inferred. Being explicit about the three "no" rows: they pin correct behaviour rather than reproduce a failure. Pre-fix, the UNREGISTER path happens to recover because the failed `regc_init` causes the Contact to be rebuilt on the next cycle. Likewise the `reg_contact.ptr[slen] == '\0'` assertion is correct but not a dependable detector of the NUL-termination bug, since the uninitialized pool byte is usually already zero.

Parser behaviour also verified directly against the built libraries:

```
<sip:u@2001:db8::1:5060>    -> hdr=NULL                  <- the crash input
<sip:u@1.2.3.4!:5060>       -> hdr=NULL                  <- token char legal in received=
<sip:u@[2001:db8::1]:5060>  -> ok, host='2001:db8::1'    <- what the fix emits
*                           -> hdr=ok, uri=NULL          <- covered by the uri check
<tel:+15551234>             -> hdr=ok, uri=ok, sip=0     <- covered by the scheme check
```

Locally: `regc_test auth_async_test pjsua_auth_test pjsua_call_test pjsua_acc_test` 5/5 under ASan, plus `transport_loop_test transport_udp_test tsx_destroy_test`. Build is warning-free. CI is green on f8b7316 across Ubuntu (including the TSan job), MacOS, Windows, CodeQL and CIFuzz.

Co-Authored-By Claude Code

### DIFF
--- a/pjsip/CMakeLists.txt
+++ b/pjsip/CMakeLists.txt
@@ -318,6 +318,7 @@ if(BUILD_TESTING)
     src/test/auth_async_test.c
     src/test/pjsua_auth_test.c
     src/test/pjsua_call_test.c
+    src/test/pjsua_acc_test.c
     src/test/${PJ_TEST_MAIN}
   )
 
--- a/pjsip/build/Makefile
+++ b/pjsip/build/Makefile
@@ -205,7 +205,8 @@ export TEST_OBJS += dlg_core_test.o dns_test.o msg_err_test.o \
 		    tsx_basic_test.o tsx_bench.o tsx_uac_test.o \
 		    tsx_uas_test.o txdata_test.o uri_test.o \
 		    inv_offer_answer_test.o auth_async_test.o \
-		    pjsua_auth_test.o pjsua_call_test.o
+		    pjsua_auth_test.o pjsua_call_test.o \
+		    pjsua_acc_test.o
 export TEST_CFLAGS += $(_CFLAGS) $(PJ_VIDEO_CFLAGS)
 export TEST_CXXFLAGS += $(_CXXFLAGS)
 export TEST_LDFLAGS += $(PJSUA_LIB_LDLIB) \
--- a/pjsip/build/pjsip_test.vcxproj
+++ b/pjsip/build/pjsip_test.vcxproj
@@ -550,6 +550,7 @@
     <ClCompile Include="..\src\test\auth_async_test.c" />
     <ClCompile Include="..\src\test\pjsua_auth_test.c" />
     <ClCompile Include="..\src\test\pjsua_call_test.c" />
+    <ClCompile Include="..\src\test\pjsua_acc_test.c" />
     <ClCompile Include="..\src\test\pjsua_stress.c">
       <ExcludedFromBuild>true</ExcludedFromBuild>
     </ClCompile>
--- a/pjsip/src/pjsip-ua/sip_reg.c
+++ b/pjsip/src/pjsip-ua/sip_reg.c
@@ -287,27 +287,19 @@ static pj_status_t set_contact( pjsip_regc *regc,
                                 const pj_str_t contact[] )
 {
     const pj_str_t CONTACT = { "Contact", 7 };
-    pjsip_contact_hdr *h;
+    pjsip_contact_hdr new_list;
+    pjsip_contact_hdr *h, *hdr;
     int i;
-    
-    /* Save existing contact list to removed_contact_hdr_list and
-     * clear contact_hdr_list.
-     */
-    pj_list_merge_last(&regc->removed_contact_hdr_list, 
-                       &regc->contact_hdr_list);
 
-    /* Set the expiration of Contacts in to removed_contact_hdr_list 
-     * zero.
+    /* Parse and validate every new Contact up front. Nothing below this
+     * point may fail, because from here on the current registration
+     * Contacts are moved to removed_contact_hdr_list with their
+     * expiration set to zero: bailing out midway would leave the
+     * registration with no active Contact and the old bindings queued
+     * for removal by the next REGISTER.
      */
-    h = regc->removed_contact_hdr_list.next;
-    while (h != &regc->removed_contact_hdr_list) {
-        h->expires = 0;
-        h = h->next;
-    }
-
-    /* Process new contacts */
+    pj_list_init(&new_list);
     for (i=0; i<contact_cnt; ++i) {
-        pjsip_contact_hdr *hdr;
         pj_str_t tmp;
 
         pj_strdup_with_null(regc->pool, &tmp, &contact[i]);
@@ -324,6 +316,29 @@ static pj_status_t set_contact( pjsip_regc *regc,
                      (int)tmp.slen, tmp.ptr));
             return PJSIP_EINVALIDURI;
         }
+
+        pj_list_push_back(&new_list, hdr);
+    }
+
+    /* Save existing contact list to removed_contact_hdr_list and
+     * clear contact_hdr_list.
+     */
+    pj_list_merge_last(&regc->removed_contact_hdr_list,
+                       &regc->contact_hdr_list);
+
+    /* Set the expiration of Contacts in to removed_contact_hdr_list
+     * zero.
+     */
+    h = regc->removed_contact_hdr_list.next;
+    while (h != &regc->removed_contact_hdr_list) {
+        h->expires = 0;
+        h = h->next;
+    }
+
+    /* Process new contacts */
+    while (!pj_list_empty(&new_list)) {
+        hdr = new_list.next;
+        pj_list_erase(hdr);
 
         /* Find the new contact in old contact list. If found, remove
          * the old header from the old header list.
--- a/pjsip/src/pjsip-ua/sip_reg.c
+++ b/pjsip/src/pjsip-ua/sip_reg.c
@@ -313,8 +313,14 @@ static pj_status_t set_contact( pjsip_regc *regc,
         pj_strdup_with_null(regc->pool, &tmp, &contact[i]);
         hdr = (pjsip_contact_hdr*)
               pjsip_parse_hdr(regc->pool, &CONTACT, tmp.ptr, tmp.slen, NULL);
-        if (hdr == NULL) {
-            PJ_LOG(4,(THIS_FILE, "Invalid Contact: \"%.*s\"", 
+        /* A "Contact: *" header parses fine but carries no URI, and the
+         * registration Contacts are dereferenced as URIs below and in
+         * calculate_response_expiration(). Unregister-all is built
+         * separately by pjsip_regc_unregister_all(), so a starred Contact
+         * has no business being registered here.
+         */
+        if (hdr == NULL || hdr->uri == NULL) {
+            PJ_LOG(4,(THIS_FILE, "Invalid Contact: \"%.*s\"",
                      (int)tmp.slen, tmp.ptr));
             return PJSIP_EINVALIDURI;
         }
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -2141,17 +2141,20 @@ done:
         if (len > acc->contact.slen) {
             reg_contact.ptr = (char*) pj_pool_alloc(acc->pool, len);
 
-            pj_strcpy(&reg_contact, &acc->contact);
-        
+            /* Must be NULL terminated: the buffer is parsed below and by
+             * pjsip_regc, and the parser requires a NULL terminated input.
+             */
+            pj_strncpy_with_null(&reg_contact, &acc->contact, len);
+
             /* Contact URI params */
             if (acc->cfg.reg_contact_uri_params.slen) {
                 pj_pool_t *pool;
                 pjsip_contact_hdr *contact_hdr;
-                pjsip_sip_uri *uri;
+                pjsip_sip_uri *uri = NULL;
                 pj_str_t uri_param = acc->cfg.reg_contact_uri_params;
                 const pj_str_t STR_CONTACT = { "Contact", 7 };
                 char tmp_uri[PJSIP_MAX_URL_SIZE];
-                pj_ssize_t tmp_len;
+                pj_ssize_t tmp_len = -1;
 
                 /* Get the URI string */
                 pool = pjsua_pool_create("tmp", 512, 512);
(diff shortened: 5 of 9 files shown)


sauwming:
Both review points addressed in 9c062d1.

**Inline comment (bracket heuristic too loose)** — correct, and the fix was incomplete. `pjsip_VIA_PARAM_SPEC` accepts `:` in a token, so `received=foo:bar` was wrapped as `[foo:bar]`; `int_parse_host()` accepts any non-empty bracket contents, so the validation passed and a meaningless Contact was stored instead of the previous one being kept. Now brackets are added only when `pj_inet_pton(PJ_AF_INET6, ...)` accepts the value, so every other colon-bearing token stays unbracketed and is rejected by the validation.

**Suppressed comment (automated regression)** — added `pjsua_acc_test`, a pjsua-level test registered in all three build systems. A mock registrar overwrites the Via `received` param of the REGISTER; `pjsip_endpt_create_response()` clones it into the 200 OK, which lets the test drive `acc_check_nat_addr()` with an arbitrary address as seen by the registrar. Each case runs three REGISTER round trips, so a Contact stored on the first response is parsed back on a later one — the sequence that crashes.

Cases: IPv6 `received` over an IPv4 transport (must be bracketed), plain IPv4 (ordinary rewrite must still happen), a token character invalid in a host, and a colon-bearing non-address token (both must be declined with the previous Contact retained). The declined cases assert that `acc->contact` is byte-for-byte unchanged, which covers the point about the moved unregister/regc teardown not running on a rejected rewrite.

Verified the test fails without the fix — the first case aborts in `acc_check_nat_addr()` on the stored, unparseable Contact:

```
pjsua_acc.c  IP address change detected for account 0
             (192.168.18.89:52968 --> [2001:db8::1]:52968). Updating registration
Assertion failed: (contact_hdr != NULL), function acc_check_nat_addr,
                  file pjsua_acc.c, line 2362.
```

With the fix, `regc_test auth_async_test pjsua_auth_test pjsua_call_test pjsua_acc_test` all pass under ASan (5/5).

sauwming:
Second review round addressed in 9924481. No new inline comments; all three suppressed ones were coverage gaps in paths this branch changes, so I acted on each.

**`reg_contact_uri_params` untested** — added an accepted-rewrite case with the param set, driving `update_regc_contact()` through its parse-and-regenerate path, plus a `force_contact = "*"` case that drives its NULL-uri fallback.

**Entry guards untested** — added `force_contact` cases for `*` and `<tel:+15551234>`.

**UNREGISTER path untested** — added a declined rewrite with `contact_rewrite_method = PJSUA_CONTACT_REWRITE_UNREGISTER`, asserting both that the Contact is retained and that `acc->regc` survives.

Two of the four fail without the fix. `*` aborts in `update_regc_contact()` on `pj_assert(uri != NULL)`; the `tel:` URI aborts inside `pj_inet_pton()` on a `pj_str_t` read out of a `pjsip_other_uri` — the type confusion the scheme check prevents. To be straight about the other two: the UNREGISTER and `reg_contact_uri_params` cases pass against unfixed code as well. They pin correct behaviour rather than reproduce a failure — pre-fix, the UNREGISTER path happens to recover because the failed `regc_init` causes the Contact to be rebuilt on the next cycle. Likewise the `reg_contact.ptr[slen] == '\0'` assertion is correct but not a dependable detector of the NUL-termination bug, since the uninitialized pool byte is usually already zero.

### The new tests found a separate NULL dereference in pjsip_regc

`pjsua_acc_test` passed standalone but SEGV'd in a combined run:

```
SUMMARY: AddressSanitizer: SEGV sip_uri.h:260 in pjsip_uri_get_scheme
    #0  pjsip_uri_get_scheme   sip_uri.h:260
    #1  set_contact            sip_reg.c:344
    #2  pjsip_regc_init        sip_reg.c:414
    #3  pjsua_regc_init        pjsua_acc.c:3344
```

`set_contact()` checks only `hdr == NULL`, then dereferences `hdr->uri` — NULL for `Contact: *` — when `add_xuid_param` is enabled. `calculate_response_expiration()` dereferences it too. It reproduces only in a combined run because `regc_test` toggles that global and leaves it set, which is why it has gone unnoticed. Fixed by rejecting a URI-less Contact: registering one is meaningless, and `pjsip_regc_unregister_all()` builds its starred Contact directly on the tdata rather than through `set_contact()`. Consequence: `force_contact = "*"` is now refused with `PJSIP_EINVALIDURI` instead of crashing, which the test asserts.

### Testing

`regc_test auth_async_test pjsua_auth_test pjsua_call_test pjsua_acc_test` — 5/5 under ASan, plus `transport_loop_test transport_udp_test tsx_destroy_test` and the first 8 of the full run, all green. I did not get a complete 27-test run through in one go; the unrun tests (tsx_uac/uas, tcp transport) have no call path into `set_contact()`.

sauwming:
Fixed in 57dc9d7 — valid finding, and one I had noted in the original analysis without acting on.

`set_contact()` moves the current Contacts to `removed_contact_hdr_list` with `expires=0` *before* it parses the new ones, so any error return partway leaves the regc with no active Contact and the old bindings queued for removal. This predates the branch (an unparseable Contact string already triggered it); rejecting a URI-less Contact in the previous commit widened what reaches it. Now every new Contact is parsed and validated into a local list first, and the regc lists are only touched once nothing can fail.

Added a regression case: on an established registration, call `pjsip_regc_update_contact()` with `"*"`, expect `PJSIP_EINVALIDURI`, then re-register and require the binding on the wire to be unchanged.

Worth recording that my first version of that test passed against the unfixed code — it compared only the Contact URI, and the broken path re-sends the *same* URI with `;expires=0` appended. Capturing the whole header via `pjsip_hdr_print_on()` makes it discriminate:

```
error: rejected update changed the registered Contact:
  'Contact: <sip:pjsua-acc-test@1.2.3.4:51413;ob>'
-> 'Contact: <sip:pjsua-acc-test@1.2.3.4:51413;ob>;expires=0'
```

That is the reported failure mode demonstrated on the wire: the binding refused by the update goes out marked for removal.

`regc_test auth_async_test pjsua_auth_test pjsua_call_test pjsua_acc_test` — 5/5 under ASan, `regc_test` being the one with broad direct coverage of `set_contact()`.

sauwming:
Addressed in d7cecee, though not the way the comment proposed — the coverage gap was real, the defect wasn't.

**The gap was real.** The non-SIP case left `reg_contact_uri_params` NULL, so `update_regc_contact()`'s regeneration path never saw a non-SIP URI. That case now sets them.

**The defect isn't there.** `update_regc_contact()` is not the same situation as `acc_check_nat_addr()`. The value is only ever passed to `pjsip_uri_print()`, which dispatches on the URI's own vptr; no SIP-specific field is read, so the cast never reinterprets the object. `acc_check_nat_addr()` was different — it read `uri->port` and `uri->host` as struct fields, which is why the scheme guard belongs there and not here.

Verified with exactly the combination the comment expected to break — `force_contact="<tel:+15551234>"` plus `reg_contact_uri_params=";acc-test=1"`:

```
pjsua_acc.c  Contact for acc 0 updated: <tel:+15551234;acc-test=1>
```

Correct tel-URI output, params appended (valid tel URI syntax), no crash, clean under ASan.

So I did not add the scheme guard: it would be a behaviour change that silently drops configured URI params from a non-SIP Contact, with no safety gained. Instead the variable is now declared `pjsip_uri*`, which is what it actually holds, with a comment explaining why it must not be narrowed — so the next reader doesn't have to re-derive this.

One thing the comment got right about my own wording: the first commit message said the scheme rejection was added to `acc_check_nat_addr()` **and** `update_regc_contact()`. Only the former got it, deliberately. That was loose phrasing on my part.

`regc_test auth_async_test pjsua_auth_test pjsua_call_test pjsua_acc_test` — 5/5 under ASan.

sauwming:
CI Ubuntu is green on f8b7316 — `OpenH264+VPX / pjmedia,pjsip` now passes, along with the other two Ubuntu jobs, Windows, CodeQL and CIFuzz.

### What the failure actually was

Not a test failure: all 30 pjsip tests passed. The job exited 66 because ThreadSanitizer reported four lock-order-inversions, all the same cycle, all with `acc_check_nat_addr` in the stack:

```
M0 -> M1  regc_cb() holds PJSUA_LOCK -> acc_check_nat_addr() ->
          pjsua_acc_set_registration() -> pjsip_regc_send() ->
          pjsip_tsx_set_transport() takes the tsx grp_lock
M1 -> M0  regc_cb() / regc_tsx_cb() are called from tsx_set_state()
          with the tsx grp_lock held, and take PJSUA_LOCK
```

`pjsua_acc_set_registration()` does `PJSUA_UNLOCK()` before `pjsip_regc_send()` to avoid exactly this, but `PJSUA_LOCK` is a recursive mutex (`pjsua_core.c:1021`) and `regc_cb` already holds it, so the unlock only decrements the recursion count and the send runs with the lock still held.

### It is pre-existing, and it is not only a test artifact

The same recursive-lock problem is already recorded in-tree, in the TODO on the disabled `acc_delete_deferred_test` in `pjsua_auth_test.c`. And `acc_check_nat_addr()` re-registers for any `contact_rewrite_method` below `ALWAYS_UPDATE`, so **every real NAT contact rewrite takes this path** — `pjsua_acc_test` is simply the first thing in `pjsip-test` to drive a successful rewrite. I checked whether the test could avoid it: it cannot, every method value leads to the same internal re-registration.

### What f8b7316 does

Suppresses it the way the two existing PJSUA lock-topology inversions in `tests/sanitizers/tsan.supp` are handled, with a comparable write-up. Keyed on `acc_check_nat_addr` because that frame is unique to this cycle; `regc_cb` and `regc_tsx_cb` are deliberately left unsuppressed so other registration-path deadlocks still surface.

**This is suppression, not repair.** A real deadlock risk remains on the contact-rewrite path in shipping code. Fixing it means deferring the re-registration out of the callback, or making the unlock recursion-aware — a PJSUA lock topology change that does not belong in a crash-fix PR. Happy to open a separate issue for it if you'd like.

For the record: I could not verify the suppression locally, since TSan's deadlock detector is Linux-only and does not fire on macOS/arm64 (confirmed with a synthetic probe reproducing the same inversion). I only checked that the file parses; this CI run is what actually confirmed the match.

nanangizz (approved):

PR #5175, opened 2026-08-11
pjmedia: serialize access to the stream RTCP session

sauwming:
## Problem

The stream's `pjmedia_rtcp_session` (`c_strm->rtcp`) is updated from three threads with no mutual exclusion:

- the **RTP receive callback** (ioqueue worker) — `pjmedia_rtcp_rx_rtp2()`, `pjmedia_rtcp_rx_rtcp()`
- the **RTP transmit path** (media clock) — `pjmedia_rtcp_tx_rtp()`
- the **RTCP build path** — `pjmedia_stream_send_rtcp()`, which reads the whole session to build SR/RR

`pjmedia_stream_send_rtcp()` takes the transport group lock, and its comment says that is there to "prevent data race" — but that lock only serializes its own callers. Nothing in the rx/tx paths takes it. With `PJ_IOQUEUE_CALLBACK_NO_LOCK` enabled (the default in `config.h`), the receive callback no longer runs under the ioqueue key lock either, so the last implicit serialization is gone.

TSan on the `pjsua-stress` workload reports this directly. The clearest instance is the RTCP BYE that `pjmedia_stream_destroy()` sends while an rx callback is still in flight on the same stream:

```
Read of size 2 by thread T639 (mutexes: write M0, write M1):
  pjmedia_rtcp_build_rtcp      rtcp.c:1020
  pjmedia_stream_send_rtcp     stream_common.c:652
  send_rtcp                    stream_imp_common.c:257
  pjmedia_stream_destroy       stream.c:2566        <- the RTCP BYE
  pjsua_aud_stop_stream / stop_media_stream / pjsua_call_hangup

Previous write of size 2 by thread T1 (no mutexes):
  pjmedia_rtp_seq_update       rtp.c:364
  pjmedia_rtcp_rx_rtp2         rtcp.c:340
  on_rx_rtp                    stream_imp_common.c:599
  srtp_rtp_cb / call_rtp_cb / ioqueue_dispatch_read_event

Location is heap block allocated by pjmedia_stream_create
```

Across one stress run there were 18 such rx-vs-teardown reports, and every racing address in the 164 reports involving `pjmedia_stream_send_rtcp` was inside the stream's own pool block. By contrast there were **zero** races between transport teardown and the rx callback, so the transport-side synchronization is holding — this is specifically the stream's RTCP session.

These reports are normally invisible: `tests/sanitizers/tsan.supp` starts with a blanket `race:*`.

## Change

Add a per-stream mutex (`pjmedia_stream_common::rtcp_mutex`) for the RTCP session and take it in every path that touches it:

- `on_rx_rtp()` / `on_rx_rtcp()` / `send_keep_alive_packet()` (`stream_imp_common.c`)
- `put_frame_imp()` (`stream.c`) and the video encode paths (`vid_stream.c`)
- the build phase of `pjmedia_stream_send_rtcp()` (`stream_common.c`)
- `pjmedia_stream_common_get_stat()` / `_reset_stat()`, `pjmedia_stream_get_stat_xr()`

`pjmedia_vid_stream_get_stat()` / `_reset_stat()` duplicated the stat copy/reset inline rather than using the common accessors; they now delegate, so they pick up the locking too.

The mutex is a **leaf**: it is only ever acquired innermost and nothing else is acquired while holding it, so it adds no new lock ordering and cannot invert against the transport group lock. In `pjmedia_stream_send_rtcp()` it is held across the send too, because when no SDES/BYE/XR/FB is appended the packet sent is the session's own SR/RR buffer.

`on_rx_rtp()` now snapshots `rtcp.received` under the mutex and uses the snapshot for the two `>= 10` checks, so the RTCP-FB NACK and initial-RR sends happen outside the mutex (`send_rtcp()` takes it itself).

## Validation

- `make` on pjmedia and pjsip: zero warnings (`-Wall`).
- Local `pjsua-stress -n 32 -v 8 -t 4 --duration 45` under ASan (macOS/kqueue): exit 0, no ASan report, ~8.6k streams created/destroyed, timer heap drains to 0.
- CI `stress / tsan` with the `race:*` suppression temporarily lifted, same workload before vs after the first commit:

  | | before | after |
  |---|---|---|
  | stress.log size | 170 MB | 4 MB |
  | data race reports | 9782 | 4039 |
  | `rtcp.c` / `rtp.c` session races | ~90 | ~9 |
  | rx callback vs `pjmedia_stream_destroy` | 18 | **0** |

  The residual `rtcp.c` reports in that run are what prompted the second commit — they were accessors still bypassing the mutex, e.g. `pjsua_call_dump()` reading the stat via `pjmedia_vid_stream_get_stat()` from the hangup thread while `pjmedia_rtcp_rx_rtp2()` wrote it from the ioqueue worker.

  After that second commit, a re-verification run (31456658582) reports **zero** races in `rtcp.c` and `rtp.c`. The only two reports left in those files' call paths are on unrelated fields — `c_strm->dec->paused` in `pjmedia_stream_common_start()` and `c_strm->av_sync_media` in `on_rx_rtcp()` — both pre-existing and out of scope here.

The temporary commit that lifted `race:*` for these measurements has been reverted; this branch now touches pjmedia sources only.

## Relationship to the stress SIGSEGV

The recurring `stress / tsan` crash at `pjmedia_stream_send_rtcp+0x83` is what led here, and it reproduced on this branch — but that is expected: this branch is based on master and does **not** contain #5174.

That crash has since been root-caused, with TSan evidence from this branch's own diagnostic run, to the transport-publication ordering in `pjmedia_vid_stream_create()` that #5174 fixes: `c_strm->transport` is read by an in-flight `on_rx_rtp()` while `pjmedia_vid_stream_create()` is still about to assign it. See #5174 for both sides of the race.

So the two PRs are independent fixes for independent defects, and it is #5174, not this one, that addresses the crash.

## Notes

This is a hot-path locking change in core media (one extra mutex per RTP packet on rx and tx), so it deserves a careful look. The alternative — reusing the transport group lock in the rx/tx paths — was rejected because the rx callback's prologue also publishes events and logs, and holding the transport group lock across those risks deadlock. Reordering `detach()` before the BYE in destroy was also rejected: `transport_send_rtcp2()` returns early on `!udp->started` and detach clears `started`, so the BYE would be silently dropped.

Co-Authored-By Claude Code

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -831,7 +831,9 @@ jobs:
     - name: run pjsua-stress
       run: |
         set -eo pipefail
-        export TSAN_OPTIONS="suppressions=$GITHUB_WORKSPACE/tests/sanitizers/tsan.supp halt_on_error=1"
+        # TEMPORARY - DO NOT MERGE: report every race instead of aborting on the
+        # first one, so the fixed race can be confirmed absent.
+        export TSAN_OPTIONS="suppressions=$GITHUB_WORKSPACE/tests/sanitizers/tsan.supp halt_on_error=0 history_size=4"
         ARCH=$(make --no-print-directory -s infotarget)
         cd pjsip/bin
         # Run in the background (log streamed to the console by tail) so a
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -831,9 +831,7 @@ jobs:
     - name: run pjsua-stress
       run: |
         set -eo pipefail
-        # TEMPORARY - DO NOT MERGE: report every race instead of aborting on the
-        # first one, so the fixed race can be confirmed absent.
-        export TSAN_OPTIONS="suppressions=$GITHUB_WORKSPACE/tests/sanitizers/tsan.supp halt_on_error=0 history_size=4"
+        export TSAN_OPTIONS="suppressions=$GITHUB_WORKSPACE/tests/sanitizers/tsan.supp halt_on_error=1"
         ARCH=$(make --no-print-directory -s infotarget)
         cd pjsip/bin
         # Run in the background (log streamed to the console by tail) so a
--- a/pjmedia/include/pjmedia/stream_common.h
+++ b/pjmedia/include/pjmedia/stream_common.h
@@ -87,6 +87,10 @@ typedef struct pjmedia_stream_common
     unsigned                 jb_last_frm_cnt;/**< Last JB frame type counter*/
 
     pjmedia_rtcp_session     rtcp;          /**< RTCP for incoming RTP.     */
+    pj_mutex_t              *rtcp_mutex;    /**< Protects rtcp session, which
+                                                 is updated from the RTP tx,
+                                                 RTP/RTCP rx, and RTCP build
+                                                 paths concurrently.        */
 
     pj_uint32_t              rtcp_last_tx;  /**< RTCP tx time in timestamp  */
     pj_timestamp             rtcp_fb_last_tx;/**< Last RTCP-FB tx time.     */
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -2726,7 +2726,10 @@ PJ_DEF(pj_status_t) pjmedia_stream_get_stat_xr( const pjmedia_stream *stream,
     PJ_ASSERT_RETURN(stream && stat, PJ_EINVAL);
 
     if (c_strm->rtcp.xr_enabled) {
+        pj_mutex_t *mutex = ((pjmedia_stream_common*)c_strm)->rtcp_mutex;
+        pj_mutex_lock(mutex);
         pj_memcpy(stat, &c_strm->rtcp.xr_session.stat, sizeof(pjmedia_rtcp_xr_stat));
+        pj_mutex_unlock(mutex);
         return PJ_SUCCESS;
     }
     return PJ_ENOTFOUND;
--- a/pjmedia/src/pjmedia/stream.c
+++ b/pjmedia/src/pjmedia/stream.c
@@ -1305,9 +1305,11 @@ static pj_status_t put_frame_imp( pjmedia_port *port,
     }
 
     /* Update stat */
+    pj_mutex_lock(c_strm->rtcp_mutex);
     pjmedia_rtcp_tx_rtp(&c_strm->rtcp, (unsigned)frame_out.size);
     c_strm->rtcp.stat.rtp_tx_last_ts = pj_ntohl(c_strm->enc->rtp.out_hdr.ts);
     c_strm->rtcp.stat.rtp_tx_last_seq = pj_ntohs(c_strm->enc->rtp.out_hdr.seq);
+    pj_mutex_unlock(c_strm->rtcp_mutex);
 
 #if defined(PJMEDIA_STREAM_ENABLE_KA) && PJMEDIA_STREAM_ENABLE_KA!=0
     /* Update time of last sending packet. */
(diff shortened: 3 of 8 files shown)


sauwming:
Closing this. It fixes a real race, but not one worth the cost — and not the one it was opened to chase.

## Why it was opened

The recurring `stress / tsan` SIGSEGV at `pjmedia_stream_send_rtcp+0x83` led to a TSan run with the blanket `race:*` suppression lifted. That surfaced 18 reports of `pjmedia_stream_destroy()` building the RTCP BYE while an in-flight `on_rx_rtp()` mutated the same `pjmedia_rtcp_session`, with no mutual exclusion. This PR serialized that session behind a per-stream mutex, and it did work: those reports went to zero, and `rtcp.c` / `rtp.c` races went from ~90 to 0.

## Why it is being closed

**1. It does not fix the crash.** That has since been root-caused to a different defect, fixed in #5174: `pjmedia_vid_stream_create()` publishes `c_strm->transport` *after* `pjmedia_transport_attach2()`, so an rx callback delivered in that window reads a NULL transport. TSan caught both accesses on the same address — the write at `vid_stream.c:1937` and the read at `stream_common.c:655`, the latter being the faulting instruction, with the crash stack frame for frame. The crash also reproduced twice on this branch, which is based on master and lacks #5174.

**2. It regressed CI.** The mutex was held across `pjmedia_transport_send_rtcp()`, and the loop transport delivers to its attached users inline — so the send re-entered the stream's rx callbacks on the same thread and self-deadlocked on a non-recursive mutex. Three pjmedia jobs timed out at one hour. The same mistake applied to `on_rx_rtcp()`, which held the mutex across `pjmedia_rtcp_rx_rtcp()`; that publishes `PJMEDIA_EVENT_RX_RTCP_FB` synchronously from `parse_rtcp_fb()`, and a subscriber can send RTCP from the callback. Both violate the project's own rule about not invoking callbacks while holding a mutex.

**3. Fixing that shrank what it delivers.** Making the mutex a genuine leaf meant dropping the `rx_rtcp` coverage entirely, so RTCP receive updates are no longer serialized against the build path. What is left is a mutex on the per-RTP-packet rx and tx paths, buying consistency on statistics counters — fields that can be torn but cannot fault, since no pointers are involved.

## What remains true, for the record

The race is real and still present on master:

```
Read of size 2 by thread T639:
  pjmedia_rtcp_build_rtcp      rtcp.c:1020
  pjmedia_stream_send_rtcp     stream_common.c:652
  send_rtcp                    stream_imp_common.c:257
  pjmedia_stream_destroy       stream.c:2566        <- the RTCP BYE
  pjsua_aud_stop_stream / stop_media_stream / pjsua_call_hangup

Previous write of size 2 by thread T1 (no mutexes):
  pjmedia_rtp_seq_update       rtp.c:364
  pjmedia_rtcp_rx_rtp2         rtcp.c:340
  on_rx_rtp                    stream_imp_common.c:599
  srtp_rtp_cb / call_rtp_cb / ioqueue_dispatch_read_event
```

The comment in `pjmedia_stream_send_rtcp()` says the transport group lock is taken there to "prevent data race", but that lock only serializes that function's own callers; nothing in the rx/tx paths takes it. With `PJ_IOQUEUE_CALLBACK_NO_LOCK` defaulting to 1, the rx callback no longer runs under the ioqueue key lock either, so the last implicit serialization is gone.

Doing this properly means deferring the event publish out of `pjmedia_rtcp_rx_rtcp()` first, then serializing the whole session — a change to `rtcp.c`'s contract, worth making deliberately rather than bolting a lock onto the callers.

Also worth noting separately: `tests/sanitizers/tsan.supp` begins with `race:*`, which hides this entire class of report. Three defects found in this investigation were invisible under it; the job only ever surfaced a bare fault address.

PR #5178, opened 2026-08-11
pjmedia: make VoiceProcessingIO other-audio ducking configurable on Apple platforms

### LINKED ISSUE #5177: coreaudio: VPIO ducks other audio for the whole call, with no way to configure it
### Summary

The coreaudio backend opens the VoiceProcessingIO (VPIO) audio unit whenever echo cancellation is enabled. VPIO ducks "other audio" — every audio stream on the system that is not the voice chat — to improve the intelligibility of the call. pjproject never configures that ducking, so it gets the system default: a **fixed duck, held for the entire call**.

The user-visible effect: someone listening to music who accepts a call has that music attenuated from answer to hangup, including the long stretches when nobody is talking. There is currently no way for an application on this backend to change that, short of turning off echo cancellation.

### What macOS 14 / iOS 17 added

`kAUVoiceIOProperty_OtherAudioDuckingConfiguration` makes the behaviour configurable through two independent controls (per WWDC23 session 10235, *What's new in voice processing*):

- `mEnableAdvancedDucking` — the *style* of ducking. When enabled the duck follows the voice activity of the local and remote participants: more ducking while someone is talking, less when nobody is. Apple compares it to FaceTime SharePlay, where media volume returns between utterances. Disabled by default, which is what pjproject gets today.
- `mDuckingLevel` — the *amount* of ducking: `Default`, `Min`, `Mid`, `Max`. `Default` is the value Apple tuned for a typical voice chat, and is what pjproject gets today.

`AudioUnitProperties.h` states: "If not set, the default ducking configuration is to disable advanced ducking, with a ducking level set to `kAUVoiceIOOtherAudioDuckingLevelDefault`."

### Suggested fix

Set the property in `create_audio_unit()` when EC is enabled, driven by two `#ifndef`-guarded `config.h` macros so downstream projects can override them from `config_site.h` — advanced ducking on by default (it addresses the actual complaint and is better for any VoIP app), and the ducking depth left at Apple's default (a far more opinionated axis where deployments genuinely differ, so the library should not pick a side).

Gating would follow the macOS 14 / iOS 17 pattern already used in `pjmedia/src/pjmedia-videodev/darwin_dev.m` — an SDK check plus a runtime `@available` — with the property probed via `AudioUnitGetPropertyInfo` and any failure logged as a warning rather than treated as an error, so it can never break audio.

I have a patch ready and will open a PR against this issue; happy to make the new behaviour opt-in instead if you would rather ship it as strictly additive.

laconicman:
Fixes #5177

### Motivation

The coreaudio backend opens the VoiceProcessingIO (VPIO) audio unit whenever echo cancellation
is enabled. VPIO ducks "other audio" — every audio stream on the system that is not the voice
chat — to improve the intelligibility of the call.

With the system defaults that duck is **fixed and held for the entire call**. A user listening
to music who accepts a call has that music attenuated from answer to hangup, even during the
long stretches when nobody is talking.

macOS 14 / iOS 17 added `kAUVoiceIOProperty_OtherAudioDuckingConfiguration`, which exposes two
independent controls over that behaviour:

- `mEnableAdvancedDucking` — the *style* of ducking. When enabled, the duck follows the voice
  activity of the local and remote chat participants: more ducking while someone is talking,
  less when nobody is. Apple compares it to FaceTime SharePlay, where media volume comes back
  up between utterances.
- `mDuckingLevel` — the *amount* of ducking: `Default`, `Min`, `Mid`, `Max`.

pjproject currently sets neither, so it gets Apple's implicit default: advanced ducking off,
level `Default`.

### Change

`create_audio_unit()` sets the property right after the audio unit is instantiated, only when
EC is enabled (i.e. only when the unit actually is VPIO). Gating follows the existing in-tree
style for macOS 14 / iOS 17 API in `pjmedia/src/pjmedia-videodev/darwin_dev.m`: an SDK check
(`__IPHONE_17_0` / `__MAC_14_0`) plus a runtime `if (@available(macOS 14.0, iOS 17.0, *))`.

The property is queried with `AudioUnitGetPropertyInfo` first and skipped unless it is present
and writable. Every failure path logs a `PJ_LOG(4, …)` warning and returns — configuring the
duck must never be able to break audio.

Two `config.h` macros, `#ifndef`-guarded so they can be overridden from `config_site.h`:

| Macro | Default | Effect |
|---|---|---|
| `PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING` | `1` | Enable voice-activity-driven ducking |
| `PJMEDIA_AUDIO_DEV_COREAUDIO_DUCKING_LEVEL` | `0` (`kAUVoiceIOOtherAudioDuckingLevelDefault`) | Ducking depth |

No new `pjmedia_aud_dev_cap`: this has no independent runtime axis — it is a refinement of
existing EC/VPIO behaviour, not something an application would toggle per call through
`pjmedia_aud_stream_set_cap()` — and adding one would push `PJMEDIA_AUD_DEV_CAP_MAX` past its
current `16384`.

### Why these defaults

**Advanced ducking on.** The complaint this addresses is precisely "other audio is attenuated
for the whole call". Following voice activity fixes exactly that, and it is the part of the
change that is unambiguously better for any VoIP application.

**Depth left at Apple's default.** Ducking depth is a much more opinionated axis, and real
deployments differ: a dispatch or contact-centre application may want a deep duck for
intelligibility, a consumer softphone a light one. Apple describes `Default` as the tuned value
for a typical voice chat, and it is the level pjproject already gets today. A library should
not pick a side there, so it is exposed as its own macro and left where it is. Keeping the
depth unchanged also holds the behavioural delta to one clearly-motivated axis.

### Three things worth flagging

**1. This changes existing behaviour.** Anyone building against a macOS 14+ / iOS 17+ SDK with
EC enabled gets voice-activity-driven ducking instead of a constant duck. Setting
`PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING` to `0` restores the previous behaviour exactly:
with both macros at `0` the property is set to `{advanced off, level Default}`, which
`AudioUnitProperties.h` documents as identical to never setting the property at all — "If not
set, the default ducking configuration is to disable advanced ducking, with a ducking level set
to `kAUVoiceIOOtherAudioDuckingLevelDefault`." Happy to flip the default to opt-in (`0`) if you
would rather ship this as strictly additive.

**2. `mDuckingLevel` is meaningful whether or not advanced ducking is on.** I checked rather
than assumed. WWDC23 session 10235 states the struct "provides controls of two independent
aspects of ducking — the style of ducking, that is `mEnableAdvancedDucking`, and the amount of
ducking, that is `mDuckingLevel`", and that "the two controls can be used independently".
That is why they are two macros and not one enum.

**3. Platform scope.** `kAUVoiceIOProperty_OtherAudioDuckingConfiguration` is declared
`API_AVAILABLE(ios(17.0), macos(14.0)) API_UNAVAILABLE(watchos, tvos)`. That covers every
platform this backend targets — macOS, iOS, and Mac Catalyst, which inherits the iOS
availability — and I verified Catalyst compiles. It does **not** cover tvOS or watchOS, where
the symbol is declared but unavailable and would fail to compile. The guard I used matches the
file's convention that `TARGET_OS_IPHONE` means iOS, which is true for pjproject today (there
is no tvOS or watchOS build support anywhere in the tree). If you would rather harden it
against a future port, `&& !TARGET_OS_TV && !TARGET_OS_WATCH` on the `TARGET_OS_IPHONE` arm is
the one-line change — say the word and I will add it.

### Validation

- `make` in `pjmedia/build` on `aarch64-apple-darwin` — clean, zero warnings, and the new code
  is confirmed present in the built `coreaudio_dev.o`.
- `coreaudio_dev.m` compiles clean (`-Wall -Wextra`, no output) for: macOS at the host
  deployment target and at `-mmacosx-version-min=10.15`; iOS `arm64` at deployment targets 15.0
  and 12.0 — i.e. both the compiled-in and the runtime-`@available`-false paths; Mac Catalyst
  `arm64`; with `PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING=0`; with
  `PJMEDIA_AUDIO_DEV_COREAUDIO_DUCKING_LEVEL=kAUVoiceIOOtherAudioDuckingLevelMin` (confirming
  the macro accepts the Apple constants by name from `config_site.h`); and with
  `PJMEDIA_AUDIO_DEV_HAS_COREAUDIO=0`.
- Non-Apple platforms are unaffected: the two new macros are referenced only in
  `coreaudio_dev.m`, and `config.h` gains only integer literals, so it still compiles with no
  Apple headers in scope (verified by compiling `audiodev.c` against it).
- **Not verified: the listening test.** I have not played music, placed a call with EC enabled,
  and confirmed by ear that the music is no longer flattened for the call's duration and that
  echo cancellation still works. That check is human-only and has not been done.

### DIFF
--- a/pjlib/include/pj/config_site_sample.h
+++ b/pjlib/include/pj/config_site_sample.h
@@ -322,6 +322,11 @@
     /* The CoreAudio backend has built-in echo canceller! */
     #define PJMEDIA_HAS_SPEEX_AEC    0
 
+    /* Duck other audio based on voice activity, instead of for the whole
+     * call (requires iOS 17).
+     */
+    #define PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING 1
+
     /* Disable some codecs */
     #define PJMEDIA_HAS_L16_CODEC               0
     //#define PJMEDIA_HAS_G722_CODEC            0
--- a/pjmedia/include/pjmedia-audiodev/config.h
+++ b/pjmedia/include/pjmedia-audiodev/config.h
@@ -137,6 +137,43 @@ PJ_BEGIN_DECL
 #endif
 
 
+/**
+ * Specify whether the coreaudio backend should enable advanced ducking of
+ * other (i.e. non-voice) audio while the VoiceProcessingIO audio unit is in
+ * use, i.e. while echo cancellation is enabled. Advanced ducking follows the
+ * voice activity of the chat participants, so other audio is only attenuated
+ * while somebody is talking, instead of being attenuated for the whole call.
+ *
+ * Requires macOS 14 or iOS 17, it is ignored on older versions.
+ *
+ * Default: 0 (no, i.e. the fixed ducking applied by the system). Enabled in
+ * config_site_sample.h for iOS.
+ */
+#ifndef PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING
+#   define PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING 0
+#endif
+
+
+/**
+ * Specify how much the coreaudio backend should duck other (i.e. non-voice)
+ * audio while the VoiceProcessingIO audio unit is in use, i.e. while echo
+ * cancellation is enabled. Valid values are the AUVoiceIOOtherAudioDuckingLevel
+ * constants: kAUVoiceIOOtherAudioDuckingLevelDefault (0),
+ * kAUVoiceIOOtherAudioDuckingLevelMin (10),
+ * kAUVoiceIOOtherAudioDuckingLevelMid (20), or
+ * kAUVoiceIOOtherAudioDuckingLevelMax (30). This setting is independent of
+ * PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING.
+ *
+ * Requires macOS 14 or iOS 17, it is ignored on older versions.
+ *
+ * Default: 0 (kAUVoiceIOOtherAudioDuckingLevelDefault, the level Apple applies
+ * for a typical voice chat)
+ */
+#ifndef PJMEDIA_AUDIO_DEV_COREAUDIO_DUCKING_LEVEL
+#   define PJMEDIA_AUDIO_DEV_COREAUDIO_DUCKING_LEVEL 0
+#endif
+
+
  /**
   * This setting controls whether WMME support should be included.
   */
--- a/pjmedia/src/pjmedia-audiodev/coreaudio_dev.m
+++ b/pjmedia/src/pjmedia-audiodev/coreaudio_dev.m
@@ -1346,6 +1346,46 @@ static pj_status_t create_audio_resample(struct coreaudio_stream     *strm,
 }
 #endif
 
+#if (TARGET_OS_IPHONE && !TARGET_OS_TV && !TARGET_OS_WATCH && \
+     defined(__IPHONE_17_0)) || \
+    (TARGET_OS_OSX && defined(__MAC_14_0))
+/* Internal: configure how the VoiceProcessingIO unit ducks other audio */
+static void set_other_audio_ducking(AudioUnit io_unit)
+    API_AVAILABLE(macos(14.0), ios(17.0))
+{
+    AUVoiceIOOtherAudioDuckingConfiguration ducking;
+    Boolean writable = false;
+    OSStatus ostatus;
+
+    ostatus = AudioUnitGetPropertyInfo(io_unit,
+                  kAUVoiceIOProperty_OtherAudioDuckingConfiguration,
+                  kAudioUnitScope_Global, 0, NULL, &writable);
+    if (ostatus != noErr || !writable) {
+        PJ_LOG(4, (THIS_FILE, "Warning: other audio ducking is not "
+                              "configurable, error: %d", (int)ostatus));
+        return;
+    }
+
+    ducking.mEnableAdvancedDucking =
+        PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING;
+    ducking.mDuckingLevel = (AUVoiceIOOtherAudioDuckingLevel)
+        PJMEDIA_AUDIO_DEV_COREAUDIO_DUCKING_LEVEL;
+
+    ostatus = AudioUnitSetProperty(io_unit,
+                  kAUVoiceIOProperty_OtherAudioDuckingConfiguration,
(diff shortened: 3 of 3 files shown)


sauwming:
> **1. This changes existing behaviour.**  Happy to flip the default to opt-in (`0`) if you would rather ship this as strictly additive.

Yes, I would prefer to leave the existing behaviour unchanged by setting the default as 0.
What I can offer you is to enable PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING to 1 inside `config_site_sample.h` instead (within `PJ_CONFIG_IPHONE`).

> **3. Platform scope.** ` If you would rather harden it against a future port, `&& !TARGET_OS_TV && !TARGET_OS_WATCH` on the `TARGET_OS_IPHONE` arm is the one-line change — say the word and I will add it.

This also sounds good. Yes, please add it.

laconicman:
Both done, thanks — pushed as an amend to the same commit.

**Default.** `PJMEDIA_AUDIO_DEV_COREAUDIO_ADVANCED_DUCKING` now defaults to `0`, so existing behaviour is unchanged, and I took your suggestion: `config_site_sample.h` sets it to `1` inside `PJ_CONFIG_IPHONE`.

One thing worth checking — `PJ_CONFIG_IPHONE` only covers iOS, so macOS keeps the fixed whole-call duck. macOS enables the backend through `aconfigure`'s `*darwin*` branch rather than `config_site_sample.h`, so there's no equivalent hook to put it in. Is leaving macOS on the old behaviour what you intend, or would you like it enabled there too?

**Platform guard.** Added `&& !TARGET_OS_TV && !TARGET_OS_WATCH`. This was load-bearing rather than defensive — a tvOS build against the previous commit fails with `'kAUVoiceIOProperty_OtherAudioDuckingConfiguration' is unavailable: not available on tvOS`, and compiles clean with the guard. (watchOS has no `AudioUnit/AudioUnit.h` at all, so it's redundant there, but harmless and symmetric.)

**One thought for later, not for this PR.** Flipping this default is API- and ABI-neutral: it's a compile-time macro with no struct layout change, no new enum member, and no new symbol — purely behavioural. That makes a major release a natural place to revisit it, since semver already signals "behaviour you relied on may differ" there, and applications handling semver correctly wouldn't be caught out. Recording the thought so it isn't lost: keeping `0` through the upcoming minor release seems right to me, and I'd like to revisit it at the next major rather than treat it as settled.

sauwming:
I suppose we can leave MacOS as it is right now.

sauwming (approved):

nanangizz (approved):

Issue #1665, opened 2013-05-03
Assertion on retransmitting a pending message


pjsipbot:
_2013-05-03 14:33:52: @nanangizz_ _created the issue on trac ticket [1665](https://trac.pjsip.org/repos/ticket/1665)_

This is reported to happen on congested network, so that the retransmit timer kicks in before the previous transmission finished. The assertion itself occurs in pjsip_transport_send() due to the attempt of sending a pending tx_data.
___
_2013-05-06 10:10:58: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2013-05-06 10:10:58: @nanangizz_ _set **resolution** to **fixed**_

___
_2013-05-06 10:10:58: @nanangizz_ _commented_

In [r4516](../commit/9731c28d04c245b674fba275900618f0690ecfca):
Fix #1665: Added check for tx_data pending state in tsx_send_msg().

PR #5180, opened 2026-08-12
Fix race when txdata is reused while a send is still in progress

### LINKED ISSUE #5176: Possible race in send_msg_callback() when reusing pjsip_tx_data after INVITE 401 authentication retry
### Describe the bug

While analyzing a core dump file from our application, ChatGPT said the issue was caused by a bug in PJSIP and the issue was still existing in latest code, it suggested to create the issue. 
Although I am not intimately familiar with the pjproject source code, I wanted to report this to ensure a potential bug is not overlooked. If this turns out to be unrelated to pjproject, please feel free to close this issue.
Below is the report from ChatGPT:
#######################################

# Possible race in `send_msg_callback()` when reusing `pjsip_tx_data` after INVITE 401 authentication retry

## Description

We encountered a very rare assertion failure in pjproject 2.15.1:

```text
pjsip/src/pjsip/sip_transaction.c:2160

pj_assert(sent != 0);
```

The issue occurred only once during approximately five days of continuous operation.

Analysis of the core dump suggests a possible race between:

* an old INVITE transaction that still has an asynchronous transport/DNS send operation pending, and
* a new INVITE transaction created after a 401 authentication retry.

The core dump appears to show that the same `pjsip_tx_data` was reused for the authenticated INVITE while an asynchronous send callback associated with the previous transaction was still active.

---

## Environment

```text
PJPROJECT version: 2.15.1
OS: Debian Linux x86_64
Compiler: gcc 12.2.0
PJSUA SIP worker threads: 2
SIP transport involved: UDP
DNS resolver: PJSIP asynchronous resolver
```

pjproject source is vendored into our application repository, so the repository Git commit does not directly correspond to an upstream pjproject commit. The version macros report 2.15.1.

---

## Crash backtrace

```text
#0  libc
#1  raise
#2  abort
#3  libc
#4  __assert_fail

#5  send_msg_callback(
        send_state=0x5d4c4926a610,
        sent=0,
        cont=0x7f3088ff86b0)
    at pjsip/sip_transaction.c:2160

#6  stateless_send_transport_cb(
        token=0x5d4c4926a610,
        tdata=0x5d4c492f8e58,
        sent=0)
    at pjsip/sip_util.c:1162

#7  stateless_send_transport_cb(
        token=0x5d4c4926a610,
        tdata=0x5d4c492f8e58,
        sent=0)
    at pjsip/sip_util.c:1313

#8  stateless_send_resolver_callback(
        status=0,
        token=0x5d4c4926a610,
        addr=0x5d4c4926a850)
    at pjsip/sip_util.c:1439

#9  dns_a_callback(
        user_data=0x5d4c4926a660,
        status=0,
        pkt=0x5d4c4917c080)
    at pjsip/sip_resolve.c:597

#10 on_read_complete(...)
    at pjlib-util/resolver.c:1763

#11 ioqueue_dispatch_read_event(...)
    at pj/ioqueue_common_abs.c:609

#12 pj_ioqueue_poll(...)
    at pj/ioqueue_epoll.c:1001

#13 pjsip_endpt_handle_events2(...)
    at pjsip/sip_endpoint.c:745

#14 pjsua_handle_events(...)
    at pjsua-lib/pjsua_core.c:2256

#15 worker_thread(...)
    at pjsua-lib/pjsua_core.c:814
```

The direct assertion failure was:

```c
pj_assert(sent != 0);
```

with:

```text
sent = 0
```

---

## Core dump findings

There are two relevant INVITE transactions.

### Old transaction (`tsx1`)

```text
tsx1 = 0x5d4c49208da8

role        = PJSIP_ROLE_UAC
method      = INVITE
cseq        = 4801
state       = PJSIP_TSX_STATE_COMPLETED
status_code = 401
```

Its transport flags were:

```text
transport_flag = 0x5
```

which in this version means:

```text
TSX_HAS_PENDING_TRANSPORT | TSX_HAS_PENDING_SEND
```

So although the transaction had already received the final 401 response and entered `COMPLETED`, it still had pending transport/send state.

The callback that crashed was associated with this transaction:

```text
send_state->token = tsx1
```

---

### Current transaction (`tsx2`)

The INVITE session already pointed to another transaction:

```text
tsx2 = 0x7f3068074988
```

Its state was:

```text
role        = PJSIP_ROLE_UAC
method      = INVITE
cseq        = 4802
state       = PJSIP_TSX_STATE_CALLING
status_code = 0
transport_flag = 0
```

Most importantly:

```text
tsx2->last_tx = 0x5d4c492f8e58
```

The INVITE session also contained:

```text
inv->state      = PJSIP_INV_STATE_CALLING
inv->invite_tsx = tsx2
inv->invite_req = 0x5d4c492f8e58
```

Therefore:

```text
inv->invite_req == tsx2->last_tx
```

Both point to the same `pjsip_tx_data`:

```text
tdata = 0x5d4c492f8e58
```

---

## Reused `pjsip_tx_data`

The same `tdata` seen by the crashing callback contained:

```text
Request msg INVITE/cseq=4802
auth_retry = 1
```

The txdata was still alive:

```text
ref_cnt = 3
```

Its pool and SIP message/header list also appeared valid in the core dump.

This therefore does not appear to be a simple use-after-free of the txdata.

However, its encoded buffer was invalid at the time of the crash:

```text
tdata->buf.cur == tdata->buf.start
```

so:

```text
tdata->buf.cur - tdata->buf.start == 0
```

The old serialized SIP text was still present in the buffer memory, but the current valid encoded length was zero.

---

## Authentication retry reuses the same txdata

The INVITE 401 handling path eventually calls:

```c
status = pjsip_auth_clt_reinit_req(
             &inv->dlg->auth_sess,
             e->body.tsx_state.src.rdata,
             tsx->last_tx,
             &tdata);
```

`pjsip_auth_clt_reinit_req()` modifies the supplied request and then does:

```c
pjsip_tx_data_invalidate_msg(tdata);

tdata->auth_retry = PJ_TRUE;

pjsip_tx_data_add_ref(tdata);

*new_request = tdata;
```

Therefore the same `pjsip_tx_data` is reused rather than creating a new txdata object.

For the initial INVITE authentication failure, `sip_inv.c` then does:

```c
pjsip_inv_uac_restart(inv, PJ_FALSE);
pjsip_inv_send_msg(inv, tdata);
```

`pjsip_inv_uac_restart()` resets:

```c
inv->state = PJSIP_INV_STATE_NULL;
inv->invite_tsx = NULL;
```

but does not appear to explicitly cancel an outstanding asynchronous transport/send operation belonging to the old INVITE transaction.

---

## Transaction send ownership

When a transaction begins the stateless transport send, `tsx_send_msg()` does:

```c
pjsip_tx_data_add_ref(tdata);

tdata->mod_data[mod_tsx_layer.mod.id] = tsx;
tsx->pending_tx = tdata;

pj_grp_lock_add_ref(tsx->grp_lock);

status = pjsip_endpt_send_request_stateless(
             tsx->endpt,
             tdata,
             tsx,
             &send_msg_callback);
```

Thus the callback token identifies the transaction that initiated the send.

However, `send_msg_callback()` checks only:

```c
if (mod_tsx_layer.mod.id < 0 ||
    tdata->mod_data[mod_tsx_layer.mod.id] == NULL)
{
    *cont = PJ_FALSE;
    return;
}
```

It does not appear to check whether the transaction currently stored in the txdata is still the same transaction as the callback token, e.g.:

```c
tdata->mod_data[mod_tsx_layer.mod.id] == tsx
```

The callback later clears:

```c
tdata->mod_data[mod_tsx_layer.mod.id] = NULL;
tsx->pending_tx = NULL;
```

---

## Suspected race sequence

Based on the core dump and source code, the following sequence appears possible:

```text
tsx1
INVITE CSeq 4801
    |
    | asynchronous DNS/transport operation still pending
    |
    v
401 Unauthorized received
    |
    v
tsx1 -> COMPLETED
    |
    | transport_flag still contains
    | PENDING_TRANSPORT | PENDING_SEND
    |
    v
pjsip_auth_clt_reinit_req(tsx1->last_tx)
    |
    | same pjsip_tx_data is reused
    |
    +--> Authorization header added
    +--> Via branch reset
    +--> txdata invalidated
    +--> auth_retry = TRUE
    |
    v
same txdata becomes INVITE CSeq 4802
    |
    v
pjsip_inv_uac_restart()
    |
    v
pjsip_inv_send_msg(inv, same tdata)
    |
    v
tsx2 created
INVITE CSeq 4802
state = CALLING
tsx2->last_tx = same tdata

Meanwhile, the old asynchronous callback for tsx1 returns:

send_msg_callback(
    token = tsx1,
    tdata = same txdata now used by tsx2
)
```

At this point the callback transaction is still `tsx1`, while the txdata has already become the active request of `tsx2`.

If `tdata->mod_data[mod_tsx_layer.mod.id]` has been updated for `tsx2`, the current callback check only sees a non-NULL value and may continue processing the stale `tsx1` callback.

The old callback may then clear:

```c
tdata->mod_data[mod_tsx_layer.mod.id] = NULL;
```

even though the same txdata is already in use by `tsx2`.

Eventually the send path reaches:

```text
sent = 0
```

and triggers:

```c
pj_assert(sent != 0);
```

---

## UDP/send observation

The send path was using UDP.

`udp_send_msg()` calculates:

```c
size = tdata->buf.cur - tdata->buf.start;

status = pj_ioqueue_sendto(..., &size, ...);
```

The core shows that the surrounding `stateless_send_transport_cb()` received:

```text
status = PJ_SUCCESS
sent   = 0
```

and then recursively called:

```c
stateless_send_transport_cb(stateless_data, tdata, sent);
```

This ultimately delivered `sent == 0` to `send_msg_callback()`.

`pjsip_tx_data_encode()` normally guarantees a non-zero encoded message when successful:

```c
size = pjsip_msg_print(...);
pj_assert(size != 0);
tdata->buf.cur += size;
```

so the zero-length state appears to have occurred after/relevant to txdata invalidation/reuse rather than because the SIP message itself was empty.

---

## Thread state

The application uses:

```text
PJSUA SIP worker threads = 2
```

The crashing SIP worker was processing:

```text
DNS resolver callback
→ stateless_send_resolver_callback
→ stateless_send_transport_cb
→ send_msg_callback
```

The second PJSUA SIP worker was blocked in:

```text
epoll_wait
→ pj_ioqueue_poll
→ pjsip_endpt_handle_events2
→ pjsua_handle_events
→ worker_thread
```

At the instant represented by the core dump, no application thread was inside:

```text
pjsua_acc_del()
pjsua_call_hangup()
pjsua_call_hangup_all()
```

or similar account/call teardown APIs.

The INVITE session itself was still:

```text
state          = PJSIP_INV_STATE_CALLING
cancelling     = 0
pending_cancel = 0
pending_bye    = NULL
cause          = PJSIP_SC_NULL
```

so there is no obvious evidence in the core of an application-triggered cancel/hangup occurring at the crash point.

---

## Possible ownership issue

We are wondering whether this check:

```c
if (mod_tsx_layer.mod.id < 0 ||
    tdata->mod_data[mod_tsx_layer.mod.id] == NULL)
```

is sufficient when the same txdata can be reused by an authentication retry before an older asynchronous send callback has completed.

Would it be necessary to distinguish a stale callback from the current txdata owner, for example conceptually:

```c
if (mod_tsx_layer.mod.id < 0 ||
    tdata->mod_data[mod_tsx_layer.mod.id] != tsx)
{
    *cont = PJ_FALSE;
    return;
}
```

We have not applied such a patch because we are not familiar enough with all transaction/group-lock reference-counting expectations in this path to know whether this would be correct or whether additional cleanup would be required.

---

## Questions

1. Is it valid for `pjsip_auth_clt_reinit_req()` to reuse the same `pjsip_tx_data` while an asynchronous transport/send callback associated with the previous transaction may still be outstanding?

2. Can `send_msg_callback()` from an old transaction legitimately execute after the same txdata has been assigned to a new transaction?

3. Should `send_msg_callback()` verify that:

```c
tdata->mod_data[mod_tsx_layer.mod.id] == tsx
```

rather than only verifying that the value is non-NULL?

4. Is `sent == 0` expected in any legitimate execution of this callback path?

5. Is there an existing fix or newer change related to INVITE 401/407 authentication retry combined with pending asynchronous DNS/transport send?

6. Would reducing `pjsua_config.thread_cnt` from 2 to 1 be an appropriate temporary workaround for this type of race?

The issue has occurred only once in approximately five days, so we currently do not have a reliable reproduction case.

We still have the core dump and can provide additional GDB inspection results if useful.

### Steps to reproduce

We currently do not have a reliable reproduction case. There are several similar applications, but I observed this issue only once in past 5 days

### PJSIP version

2.15.1

### Context

PJPROJECT version: 2.15.1
OS: Debian Linux x86_64
Compiler: gcc 12.2.0
PJSUA SIP worker threads: 2
SIP transport involved: UDP
DNS resolver: PJSIP asynchronous resolver

### Log, call stack, etc

```shell
There is no PJSIP log, below is the call stack trace:
##############################################

(gdb) thread apply all bt

Thread 13 (Thread 0x7f30825426c0 (LWP 51)):
#0  0x00007f30919b019f in g711_decode (codec=0x7f3074026808, input=0x7f3082541880, output_buf_len=320, output=0x7f30825418b0) at ../src/pjmedia/g711.c:583
#1  0x00007f30919cba0f in pjmedia_codec_decode (codec=0x7f3074026808, input=0x7f3082541880, out_size=320, output=0x7f30825418b0) at ../include/pjmedia/codec.h:1119
#2  0x00007f30919cc4cc in get_frame (port=0x7f306806dbd0, frame=0x5d4c492e9770) at ../src/pjmedia/stream.c:773
#3  0x00007f30919b52d0 in pjmedia_port_get_frame (port=0x7f306806dbd0, frame=0x5d4c492e9770) at ../src/pjmedia/port.c:99
#4  0x00007f30919a6b4e in get_frame (this_port=0x5d4c492e9d90, frame=0x7f308255a7f0) at ../src/pjmedia/conf_switch.c:1282
#5  0x00007f30919b52d0 in pjmedia_port_get_frame (port=0x5d4c492e9d90, frame=0x7f308255a7f0) at ../src/pjmedia/port.c:99
#6  0x00005d4c421f452d in CallFramer::processCall (this=0x7f3082ffd240, pcd=0x7f308255a688) at callFramer.cpp:179
#7  0x00005d4c421f4463 in CallFramer::process (this=0x7f3082ffd240) at callFramer.cpp:162
#8  0x00005d4c421f400c in CallFramer::conf_thread_proc (arg=0x5d4c4902b320) at callFramer.cpp:109
#9  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#10 0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 12 (Thread 0x7f3089ffb6c0 (LWP 45)):
#0  0x00007f3091766f36 in epoll_wait () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3091931984 in pj_ioqueue_poll (ioqueue=0x5d4c490dc718, timeout=0x7f3089ffae70) at ../src/pj/ioqueue_epoll.c:844
#2  0x00007f30919acacf in worker_proc (arg=0x5d4c490db8e8) at ../src/pjmedia/endpoint.c:348
#3  0x00007f30919330ef in thread_main (param=0x5d4c4916c240) at ../src/pj/os_core_unix.c:701
#4  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 11 (Thread 0x7f308b7fe6c0 (LWP 42)):
#0  0x00007f30916e3f16 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f30916e65d8 in pthread_cond_wait () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f309190546b in consume_socket () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#3  0x00007f3091905a91 in worker_thread_run () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#4  0x00007f3091905bab in worker_thread () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#5  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#6  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 10 (Thread 0x7f3090c0a6c0 (LWP 40)):
#0  0x00007f309172d545 in clock_nanosleep () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3091731e53 in nanosleep () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f309175d295 in usleep () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00005d4c421a7e3e in PromeMetric::metric_thread_func (arg=0x5d4c43e858c0 <PromeMetric::s_tdata>) at promeMetric.cpp:220
#4  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 9 (Thread 0x7f309140eb80 (LWP 38)):
#0  0x00007f309172d545 in clock_nanosleep () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3091731e53 in nanosleep () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f309175d295 in usleep () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007f3091933531 in pj_thread_sleep (msec=20) at ../src/pj/os_core_unix.c:948
--Type <RET> for more, q to quit, c to continue without paging--
#4  0x00005d4c4216c6f7 in caller_proc (genCfg=..., msgId2Child=1, msgId2Parent=0, enableHeartBeat=true, jobRestartTimes=1) at caller.cpp:94
#5  0x00005d4c421eb6d3 in start_job (genCfg=..., msgId2Child=1, msgId2Parent=0, jobRestartTimes=1, func=0x5d4c4216c416 <caller_proc(GeneralConfig const&, int, int, bool, int)>) at jobManager.cpp:89
#6  0x00005d4c421ed1da in run_normal (func=0x5d4c4216c416 <caller_proc(GeneralConfig const&, int, int, bool, int)>) at jobManager.cpp:197
#7  0x00005d4c4216cd75 in main (argc=1, argv=0x7ffd4ed107e8) at caller.cpp:137

Thread 8 (Thread 0x7f308bfff6c0 (LWP 41)):
#0  0x00007f30916e3f16 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f30916e65d8 in pthread_cond_wait () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f309190546b in consume_socket () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#3  0x00007f3091905a91 in worker_thread_run () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#4  0x00007f3091905bab in worker_thread () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#5  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#6  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 7 (Thread 0x7f30897fa6c0 (LWP 46)):
#0  0x00007f30916e3f16 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f30916eed90 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f309193446f in pj_sem_wait (sem=0x5d4c49174598) at ../src/pj/os_core_unix.c:1824
#3  0x00007f30919ae819 in event_worker_thread (arg=0x5d4c491778e8) at ../src/pjmedia/event.c:163
#4  0x00007f30919330ef in thread_main (param=0x5d4c491745e0) at ../src/pj/os_core_unix.c:701
#5  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#6  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 6 (Thread 0x7f309140b6c0 (LWP 39)):
#0  0x00007f309172d545 in clock_nanosleep () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3091731e53 in nanosleep () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f309175d295 in usleep () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00005d4c421edcd1 in heartbeat_func (arg=0x5d4c43e85fe8 <JobManager::s_hbTdata>) at jobManager.cpp:270
#4  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 5 (Thread 0x7f308a7fc6c0 (LWP 44)):
#0  0x00007f3091767c1a in epoll_ctl () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3091931636 in update_epoll_event_set (ioqueue=0x5d4c490dc718, key=0x5d4c490e2d18, events=268435457) at ../src/pj/ioqueue_epoll.c:714
#2  0x00007f30919317ea in ioqueue_add_to_set2 (ioqueue=0x5d4c490dc718, key=0x5d4c490e2d18, event_types=1) at ../src/pj/ioqueue_epoll.c:790
#3  0x00007f3091931778 in ioqueue_add_to_set (ioqueue=0x5d4c490dc718, key=0x5d4c490e2d18, event_type=READABLE_EVENT) at ../src/pj/ioqueue_epoll.c:773
#4  0x00007f309192f990 in pj_ioqueue_recvfrom (key=0x5d4c490e2d18, op_key=0x5d4c49319e48, buffer=0x5d4c4931bb60, length=0x7f308a7fbb58, flags=0, addr=0x5d4c4931bb40, addrlen=0x5d4c4931bb5c) at ../src/pj/ioqueue_common_abs.c:847
#5  0x00007f30919e3579 in on_rx_rtp (key=0x5d4c490e2d18, op_key=0x5d4c49319e48, bytes_read=2000) at ../src/pjmedia/transport_udp.c:633
#6  0x00007f309192f3e7 in ioqueue_dispatch_read_event (ioqueue=0x5d4c490dc718, h=0x5d4c490e2d18) at ../src/pj/ioqueue_common_abs.c:609
#7  0x00007f3091931e83 in pj_ioqueue_poll (ioqueue=0x5d4c490dc718, timeout=0x7f308a7fbe70) at ../src/pj/ioqueue_epoll.c:1001
#8  0x00007f30919acacf in worker_proc (arg=0x5d4c490db8e8) at ../src/pjmedia/endpoint.c:348
#9  0x00007f30919330ef in thread_main (param=0x5d4c4916c1f8) at ../src/pj/os_core_unix.c:701
--Type <RET> for more, q to quit, c to continue without paging--
#10 0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#11 0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 4 (Thread 0x7f3083fff6c0 (LWP 48)):
#0  0x00007f3091766f36 in epoll_wait () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3091931984 in pj_ioqueue_poll (ioqueue=0x5d4c49046d60, timeout=0x7f3083ffedf0) at ../src/pj/ioqueue_epoll.c:844
#2  0x00007f3091a33f8f in pjsip_endpt_handle_events2 (endpt=0x5d4c490361f8, max_timeout=0x7f3083ffee40, p_count=0x7f3083ffee58) at ../src/pjsip/sip_endpoint.c:745
#3  0x00007f3091aab0f8 in pjsua_handle_events (msec_timeout=10) at ../src/pjsua-lib/pjsua_core.c:2256
#4  0x00007f3091aa7e4c in worker_thread (arg=0x0) at ../src/pjsua-lib/pjsua_core.c:814
#5  0x00007f30919330ef in thread_main (param=0x5d4c49178158) at ../src/pj/os_core_unix.c:701
#6  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#7  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 3 (Thread 0x7f3081d416c0 (LWP 52)):
#0  0x00007f309172d545 in clock_nanosleep () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3091731e53 in nanosleep () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f309175d295 in usleep () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007f3091933531 in pj_thread_sleep (msec=0) at ../src/pj/os_core_unix.c:948
#4  0x00005d4c421f4470 in CallFramer::process (this=0x7f3082ffd240) at callFramer.cpp:165
#5  0x00005d4c421f400c in CallFramer::conf_thread_proc (arg=0x5d4c4902b338) at callFramer.cpp:109
#6  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#7  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 2 (Thread 0x7f308affd6c0 (LWP 43)):
#0  0x00007f309175a26f in poll () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f30918f4ff7 in mg_poll () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#2  0x00007f3091905fb9 in master_thread_run () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#3  0x00007f3091906170 in master_thread () from /usr/local/lib/libprometheus-cpp-pull.so.1.1
#4  0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6

Thread 1 (Thread 0x7f3088ff96c0 (LWP 47)):
#0  0x00007f30916e8eec in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3091699fb2 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f3091684472 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007f3091684395 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007f3091692ec2 in __assert_fail () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00007f3091a51df2 in send_msg_callback (send_state=0x5d4c4926a610, sent=0, cont=0x7f3088ff86b0) at ../src/pjsip/sip_transaction.c:2160
#6  0x00007f3091a375db in stateless_send_transport_cb (token=0x5d4c4926a610, tdata=0x5d4c492f8e58, sent=0) at ../src/pjsip/sip_util.c:1162
#7  0x00007f3091a37b0a in stateless_send_transport_cb (token=0x5d4c4926a610, tdata=0x5d4c492f8e58, sent=0) at ../src/pjsip/sip_util.c:1313
#8  0x00007f3091a37fba in stateless_send_resolver_callback (status=0, token=0x5d4c4926a610, addr=0x5d4c4926a850) at ../src/pjsip/sip_util.c:1439
#9  0x00007f3091a3acb6 in dns_a_callback (user_data=0x5d4c4926a660, status=0, pkt=0x5d4c4917c080) at ../src/pjsip/sip_resolve.c:597
#10 0x00007f309197ce27 in on_read_complete (key=0x5d4c49047468, op_key=0x5d4c4917d388, bytes_read=101) at ../src/pjlib-util/resolver.c:1763
#11 0x00007f309192f3e7 in ioqueue_dispatch_read_event (ioqueue=0x5d4c49046d60, h=0x5d4c49047468) at ../src/pj/ioqueue_common_abs.c:609
--Type <RET> for more, q to quit, c to continue without paging--
#12 0x00007f3091931e83 in pj_ioqueue_poll (ioqueue=0x5d4c49046d60, timeout=0x7f3088ff8df0) at ../src/pj/ioqueue_epoll.c:1001
#13 0x00007f3091a33f8f in pjsip_endpt_handle_events2 (endpt=0x5d4c490361f8, max_timeout=0x7f3088ff8e40, p_count=0x7f3088ff8e58) at ../src/pjsip/sip_endpoint.c:745
#14 0x00007f3091aab0f8 in pjsua_handle_events (msec_timeout=10) at ../src/pjsua-lib/pjsua_core.c:2256
#15 0x00007f3091aa7e4c in worker_thread (arg=0x0) at ../src/pjsua-lib/pjsua_core.c:814
#16 0x00007f30919330ef in thread_main (param=0x5d4c49178110) at ../src/pj/os_core_unix.c:701
#17 0x00007f30916e71f5 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#18 0x00007f30917678ec in ?? () from /lib/x86_64-linux-gnu/libc.so.6
(gdb)
```

sauwming:
Fixes #5176.

## Root cause

By the time `pjsip_transport_send()` returns, the message has already left the socket, so the peer's response can already be in flight. With more than one worker thread, the response can be processed — and the txdata reused — before the sending thread executes its next statement.

The INVITE 401/407 retry path does exactly that: `pjsip_auth_clt_reinit_req()` deliberately reuses the same `pjsip_tx_data` and calls `pjsip_tx_data_invalidate_msg()` on it, which resets `buf.cur` to `buf.start`.

`stateless_send_transport_cb()` then computed the number of bytes sent *after* the send:

```c
status = pjsip_transport_send(...);
if (status == PJ_SUCCESS) {
    sent = tdata->buf.cur - tdata->buf.start;   /* may already be 0 */
    stateless_send_transport_cb(stateless_data, tdata, sent);
```

reading zero and hitting `pj_assert(sent != 0)` in `send_msg_callback()` (`sip_transaction.c`). This matches the reported core dump: the old serialized text still present in the buffer but with zero valid length, no use-after-free, and the new transaction stuck in `CALLING` with `transport_flag == 0` because its `tsx_send_msg()` had hit the `tdata->is_pending` guard.

Note that the check suggested in the issue (`tdata->mod_data[...] != tsx`) would not have prevented the assertion, since the zero length is computed in `sip_util.c` before `send_msg_callback()` inspects `mod_data`.

## Changes

- **`pjsip_transport_send2()`** (new): same as `pjsip_transport_send()` plus an optional `pj_ssize_t *sent` out-param. The size is captured from the buffer before the txdata is handed to the transport, i.e. while it is still exclusively ours. `pjsip_transport_send()` becomes a thin wrapper, so existing callers are unaffected.
- **`sip_util.c`**: the three places that derived the byte count from `tdata->buf` after a synchronous send now take it from the out-param — `stateless_send_transport_cb()`, `send_response_resolver_cb()` and `pjsip_endpt_send_response()`.
- **`sip_transaction.c`**: `send_msg_callback()` only clears `tdata->mod_data[mod_tsx_layer.mod.id]` when it still points at this transaction. Previously a late callback could wipe the marker belonging to the transaction which had taken over the txdata; that transaction's own send callback would then bail out early, leaking a group lock reference (transaction never destroyed) and skipping its retransmission scheduling.

## Testing

`pjsip-test` built with ASan, all passing:

- `txdata_test`, `tsx_basic_test`, `tsx_uac_test`, `tsx_uas_test`, `tsx_destroy_test`, `transport_loop_test`, `transport_udp_test`, `transport_tcp_test`, `resolve_test` — 15 tests, 0 failed
- `auth_async_test`, `pjsua_auth_test`, `regc_test`, `dlg_core_test`, `inv_offer_answer_test` — 6 tests, 0 failed

## Not addressed here

The same race can still make the retried INVITE hit the `tdata->is_pending` guard in `tsx_send_msg()` and be skipped (logged as "message is pending"). Timer A recovers it on UDP; on a reliable transport it would stall until timer B. That guard is deliberate (#1665) and addressing it properly needs send queueing, so it is left for a separate change.

Co-Authored-By Claude Code

### DIFF
--- a/pjsip/include/pjsip/sip_transport.h
+++ b/pjsip/include/pjsip/sip_transport.h
@@ -1481,12 +1481,42 @@ typedef void (*pjsip_tp_send_callback)(void *token, pjsip_tx_data *tdata,
  *                  indicates immediate failure, and in this case the 
  *                  callback will not be called.
  */
-PJ_DECL(pj_status_t) pjsip_transport_send( pjsip_transport *tr, 
+PJ_DECL(pj_status_t) pjsip_transport_send( pjsip_transport *tr,
                                            pjsip_tx_data *tdata,
                                            const pj_sockaddr_t *addr,
                                            int addr_len,
                                            void *token,
                                            pjsip_tp_send_callback cb);
+
+
+/**
+ * Variant of #pjsip_transport_send() which also reports the number of bytes
+ * sent when the message is sent immediately.
+ *
+ * Callers must use this instead of inspecting the transmit data buffer after
+ * the send, since the message may already be answered by the peer, and the
+ * transmit data invalidated and reused by another thread, before this
+ * function returns.
+ *
+ * @param tr        The SIP transport to be used.
+ * @param tdata     Transmit data buffer containing SIP message.
+ * @param addr      Destination address.
+ * @param addr_len  Length of destination address.
+ * @param token     Arbitrary token to be returned back to callback.
+ * @param cb        Optional callback to be called to notify caller about
+ *                  the completion status of the pending send operation.
+ * @param sent      Optional pointer to receive the number of bytes sent.
+ *                  It is only set when this function returns PJ_SUCCESS.
+ *
+ * @return          See #pjsip_transport_send().
+ */
+PJ_DECL(pj_status_t) pjsip_transport_send2(pjsip_transport *tr,
+                                           pjsip_tx_data *tdata,
+                                           const pj_sockaddr_t *addr,
+                                           int addr_len,
+                                           void *token,
+                                           pjsip_tp_send_callback cb,
+                                           pj_ssize_t *sent);
 
 
 /**
--- a/pjsip/include/pjsip/sip_transport.h
+++ b/pjsip/include/pjsip/sip_transport.h
@@ -1463,7 +1463,13 @@ typedef void (*pjsip_tp_send_callback)(void *token, pjsip_tx_data *tdata,
 /**
  * This is a low-level function to send a SIP message using the specified
  * transport to the specified destination.
- * 
+ *
+ * Once this function returns, the message is already on the wire, so the peer
+ * may have answered it and the application may have reused the transmit data
+ * for the next request from another thread. No field of the transmit data may
+ * be assumed to be stable afterwards, in particular its buffer, which is why
+ * #pjsip_transport_send2() exists to report the number of bytes sent.
+ *
  * @param tr        The SIP transport to be used.
  * @param tdata     Transmit data buffer containing SIP message.
  * @param addr      Destination address.
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -2150,6 +2150,7 @@ static void send_msg_callback( pjsip_send_state *send_state,
 {
     pjsip_transaction *tsx = (pjsip_transaction*) send_state->token;
     pjsip_tx_data *tdata = send_state->tdata;
+    pj_bool_t tdata_is_ours;
 
     /* Check if transaction has cancelled itself from this transmit
      * notification (https://github.com/pjsip/pjproject/issues/1033).
@@ -2177,19 +2178,30 @@ static void send_msg_callback( pjsip_send_state *send_state,
     /* Decrease pending send counter */
     pj_grp_lock_dec_ref(tsx->grp_lock);
 
-    /* Reset. Only release the tdata if it is still ours, it may already have
-     * been taken over by another transaction, e.g. reused for an
-     * authentication retry (see #5176).
+    /* The tdata may already have been taken over by another transaction,
+     * e.g. reused for an authentication retry (see #5176). If so, it is
+     * neither ours to release nor to inspect or send again, since the other
+     * transaction may be updating it concurrently.
      */
-    if (tdata->mod_data[mod_tsx_layer.mod.id] == tsx)
+    tdata_is_ours = (tdata->mod_data[mod_tsx_layer.mod.id] == tsx);
+
+    /* Reset */
+    if (tdata_is_ours)
         tdata->mod_data[mod_tsx_layer.mod.id] = NULL;
+    else
+        *cont = PJ_FALSE;
     tsx->pending_tx = NULL;
 
     if (sent > 0) {
         /* Successfully sent! */
         pj_assert(send_state->cur_transport != NULL);
 
-        if (tsx->transport != send_state->cur_transport) {
+        /* Adopt the transport and remote address only from a tdata which is
+         * still ours, the dest_info of a tdata taken over by another
+         * transaction may be updated concurrently. Leaving them untouched
+         * makes our next transmission resolve the destination again.
+         */
+        if (tdata_is_ours && tsx->transport != send_state->cur_transport) {
             /* Update transport. */
             tsx_update_transport(tsx, send_state->cur_transport);
 
@@ -2208,8 +2220,13 @@ static void send_msg_callback( pjsip_send_state *send_state,
         /* Clear pending transport flag. */
         tsx->transport_flag &= ~(TSX_HAS_PENDING_TRANSPORT);
 
-        /* Mark that we have resolved the addresses. */
-        tsx->transport_flag |= TSX_HAS_RESOLVED_SERVER;
+        /* Mark that we have resolved the addresses. Not when the tdata is no
+         * longer ours, since we then have no transport to go with it, and a
+         * UAC without transport but with resolved server is terminated by
+         * tsx_send_msg().
+         */
+        if (tdata_is_ours)
+            tsx->transport_flag |= TSX_HAS_RESOLVED_SERVER;
 
         /* Pending destroy? */
         if (tsx->transport_flag & TSX_HAS_PENDING_DESTROY) {
@@ -2342,7 +2359,9 @@ static void send_msg_callback( pjsip_send_state *send_state,
                 unlock_timer(tsx);
             }
 
-            /* Put again pending tdata */
+            /* Put again pending tdata. The tdata is still ours here, a tdata
+             * taken over by another transaction has cleared *cont above.
+             */
             tdata->mod_data[mod_tsx_layer.mod.id] = tsx;
             tsx->pending_tx = tdata;
 
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -2177,8 +2177,12 @@ static void send_msg_callback( pjsip_send_state *send_state,
     /* Decrease pending send counter */
     pj_grp_lock_dec_ref(tsx->grp_lock);
 
-    /* Reset */
-    tdata->mod_data[mod_tsx_layer.mod.id] = NULL;
+    /* Reset. Only release the tdata if it is still ours, it may already have
+     * been taken over by another transaction, e.g. reused for an
+     * authentication retry (see #5176).
+     */
+    if (tdata->mod_data[mod_tsx_layer.mod.id] == tsx)
+        tdata->mod_data[mod_tsx_layer.mod.id] = NULL;
     tsx->pending_tx = NULL;
 
     if (sent > 0) {
--- a/pjsip/src/pjsip/sip_transaction.c
+++ b/pjsip/src/pjsip/sip_transaction.c
@@ -2178,10 +2178,12 @@ static void send_msg_callback( pjsip_send_state *send_state,
     /* Decrease pending send counter */
     pj_grp_lock_dec_ref(tsx->grp_lock);
 
-    /* The tdata may already have been taken over by another transaction,
-     * e.g. reused for an authentication retry (see #5176). If so, it is
-     * neither ours to release nor to inspect or send again, since the other
-     * transaction may be updating it concurrently.
+    /* The tdata may already have been taken over by another transaction, e.g.
+     * reused for an authentication retry (see #5176). If so, it is neither
+     * ours to release nor to inspect or send again. The takeover is done
+     * under the other transaction's group lock, so this only narrows the
+     * window in general, but it is exact for the authentication retry, which
+     * takes over from within our own state callback, i.e. under this lock.
      */
     tdata_is_ours = (tdata->mod_data[mod_tsx_layer.mod.id] == tsx);
 
@@ -2291,8 +2293,11 @@ static void send_msg_callback( pjsip_send_state *send_state,
             /* Clear pending transport flag. */
             tsx->transport_flag &= ~(TSX_HAS_PENDING_TRANSPORT);
 
-            /* Mark that we have resolved the addresses. */
-            tsx->transport_flag |= TSX_HAS_RESOLVED_SERVER;
+            /* Mark that we have resolved the addresses, see the same
+             * assignment in the success branch above.
+             */
+            if (tdata_is_ours)
+                tsx->transport_flag |= TSX_HAS_RESOLVED_SERVER;
 
             /* Server resolution error is now mapped to 502 instead of 503,
              * since with 503 normally client should try again.
--- a/pjsip/src/pjsip/sip_transport.c
+++ b/pjsip/src/pjsip/sip_transport.c
@@ -967,13 +967,26 @@ static pj_status_t mod_on_tx_msg(pjsip_tx_data *tdata)
 /*
  * Send a SIP message using the specified transport.
  */
-PJ_DEF(pj_status_t) pjsip_transport_send(  pjsip_transport *tr, 
+PJ_DEF(pj_status_t) pjsip_transport_send(  pjsip_transport *tr,
                                            pjsip_tx_data *tdata,
                                            const pj_sockaddr_t *addr,
                                            int addr_len,
                                            void *token,
                                            pjsip_tp_send_callback cb)
 {
+    return pjsip_transport_send2(tr, tdata, addr, addr_len, token, cb, NULL);
+}
+
+
+PJ_DEF(pj_status_t) pjsip_transport_send2( pjsip_transport *tr,
+                                           pjsip_tx_data *tdata,
+                                           const pj_sockaddr_t *addr,
+                                           int addr_len,
+                                           void *token,
+                                           pjsip_tp_send_callback cb,
+                                           pj_ssize_t *sent)
+{
+    pj_ssize_t msg_size;
     pj_status_t status;
 
     PJ_ASSERT_RETURN(tr && tdata && addr, PJ_EINVAL);
@@ -1024,8 +1037,14 @@ PJ_DEF(pj_status_t) pjsip_transport_send(  pjsip_transport *tr,
     /* Mark as pending. */
     tdata->is_pending = 1;
 
+    /* Remember the message size while the buffer is still ours. Once the
+     * message is on the wire, the peer may already answer it and another
+     * thread may invalidate and reuse this tdata (see #5176).
+     */
+    msg_size = tdata->buf.cur - tdata->buf.start;
+
     /* Send to transport. */
-    status = (*tr->send_msg)(tr, tdata,  addr, addr_len, (void*)tdata, 
+    status = (*tr->send_msg)(tr, tdata,  addr, addr_len, (void*)tdata,
                              &transport_send_callback);
 
     if (status != PJ_EPENDING) {
@@ -1033,6 +1052,9 @@ PJ_DEF(pj_status_t) pjsip_transport_send(  pjsip_transport *tr,
 
         /* Print log on successful sending */
         if (status == PJ_SUCCESS) {
+            if (sent)
+                *sent = msg_size;
+
             PJ_LOG(5,(tr->obj_name,
                       "%s sent successfully", pjsip_tx_data_get_info(tdata)));
         }
--- a/pjsip/src/pjsip/sip_util.c
+++ b/pjsip/src/pjsip/sip_util.c
@@ -1318,6 +1318,9 @@ static void stateless_send_transport_cb( void *token,
                                         &stateless_send_transport_cb,
                                         &sent);
         if (status == PJ_SUCCESS) {
+            /* A sent message is never empty, don't derive this from tdata. */
+            pj_assert(sent > 0);
+
             /* Recursively call this function. */
             stateless_send_transport_cb( stateless_data, tdata, sent );
             return;
@@ -1914,6 +1917,9 @@ static void send_response_resolver_cb( pj_status_t status, void *token,
                                     &send_response_transport_cb,
                                     &sent);
     if (status == PJ_SUCCESS) {
+        /* A sent message is never empty, don't derive this from tdata. */
+        pj_assert(sent > 0);
+
         send_response_transport_cb(send_state, send_state->tdata, sent);
 
     } else if (status == PJ_EPENDING) {
@@ -1957,6 +1963,9 @@ PJ_DEF(pj_status_t) pjsip_endpt_send_response( pjsip_endpoint *endpt,
                                         &send_response_transport_cb,
                                         &sent );
         if (status == PJ_SUCCESS) {
+            /* A sent message is never empty, don't derive this from tdata. */
+            pj_assert(sent > 0);
+
             send_response_transport_cb(send_state, tdata, sent);
             return PJ_SUCCESS;
         } else if (status == PJ_EPENDING) {
--- a/pjsip/src/pjsip/sip_util.c
+++ b/pjsip/src/pjsip/sip_util.c
@@ -1310,15 +1310,15 @@ static void stateless_send_transport_cb( void *token,
         pjsip_tx_data_invalidate_msg(tdata);
 
         /* Send message using this transport. */
-        status = pjsip_transport_send( stateless_data->cur_transport,
-                                       tdata,
-                                       cur_addr,
-                                       cur_addr_len,
-                                       stateless_data,
-                                       &stateless_send_transport_cb);
+        status = pjsip_transport_send2( stateless_data->cur_transport,
+                                        tdata,
+                                        cur_addr,
+                                        cur_addr_len,
+                                        stateless_data,
+                                        &stateless_send_transport_cb,
+                                        &sent);
         if (status == PJ_SUCCESS) {
             /* Recursively call this function. */
-            sent = tdata->buf.cur - tdata->buf.start;
             stateless_send_transport_cb( stateless_data, tdata, sent );
             return;
         } else if (status == PJ_EPENDING) {
@@ -1866,6 +1866,7 @@ static void send_response_resolver_cb( pj_status_t status, void *token,
                                        const pjsip_server_addresses *addr )
 {
     pjsip_send_state *send_state = (pjsip_send_state*) token;
+    pj_ssize_t sent;
     unsigned i;
 
     if (status != PJ_SUCCESS) {
@@ -1905,15 +1906,14 @@ static void send_response_resolver_cb( pj_status_t status, void *token,
     }
 
     /* Send response using the transoprt. */
-    status = pjsip_transport_send( send_state->cur_transport, 
-                                   send_state->tdata,
-                                   &addr->entry[0].addr,
-                                   addr->entry[0].addr_len,
-                                   send_state,
-                                   &send_response_transport_cb);
+    status = pjsip_transport_send2( send_state->cur_transport,
+                                    send_state->tdata,
+                                    &addr->entry[0].addr,
+                                    addr->entry[0].addr_len,
+                                    send_state,
+                                    &send_response_transport_cb,
+                                    &sent);
     if (status == PJ_SUCCESS) {
-        pj_ssize_t sent = send_state->tdata->buf.cur - 
-                          send_state->tdata->buf.start;
         send_response_transport_cb(send_state, send_state->tdata, sent);
 
     } else if (status == PJ_EPENDING) {
@@ -1936,6 +1936,7 @@ PJ_DEF(pj_status_t) pjsip_endpt_send_response( pjsip_endpoint *endpt,
      * based on Section 18.2.2 of RFC 3261.
      */
     pjsip_send_state *send_state;
+    pj_ssize_t sent;
     pj_status_t status;
 
     /* Create structure to keep the sending state. */
@@ -1949,13 +1950,13 @@ PJ_DEF(pj_status_t) pjsip_endpt_send_response( pjsip_endpoint *endpt,
         send_state->cur_transport = res_addr->transport;
         pjsip_transport_add_ref(send_state->cur_transport);
 
-        status = pjsip_transport_send( send_state->cur_transport, tdata, 
-                                       &res_addr->addr,
-                                       res_addr->addr_len,
-                                       send_state,
-                                       &send_response_transport_cb );
+        status = pjsip_transport_send2( send_state->cur_transport, tdata,
+                                        &res_addr->addr,
+                                        res_addr->addr_len,
+                                        send_state,
+                                        &send_response_transport_cb,
+                                        &sent );
         if (status == PJ_SUCCESS) {
-            pj_ssize_t sent = tdata->buf.cur - tdata->buf.start;
             send_response_transport_cb(send_state, tdata, sent);
             return PJ_SUCCESS;
         } else if (status == PJ_EPENDING) {


sauwming:
## Follow-up on code review

### Review comments

> **1. Residual race: the late callback still runs the full success path on tdata that's no longer ours.** In the takeover case the guard skips the `mod_data` clear but tsx A's callback proceeds to read `tdata->dest_info` (`sip_transaction.c:2193-2195`) — fields tsx B is concurrently re-resolving. Consider treating "`mod_data != tsx`" like the existing cancelled-tsx check (line 2159): do the tsx A bookkeeping (`dec_ref`, `pending_tx = NULL`) and bail, instead of running transport update / pending-send / resched logic on shared tdata.
>
> **2. Asymmetric guard in the temporary-failure path.** `send_msg_callback()` still unconditionally re-arms `tdata->mod_data[...] = tsx; tsx->pending_tx = tdata;` at line 2341-2343. If tsx B has taken over by then, tsx A clobbers tsx B's marker and tsx B leaks. It's pre-existing and only on the failure path, but since the PR is specifically about guarding this marker, the same `== tsx` (or NULL) check would be more consistent. If you deliberately leave it, a short comment noting the asymmetry would help.

Both are valid. Addressed in 4b1b8823c.

### Analysis

**Point 1.** The race is real. `pjsip_endpt_send_request_stateless()` skips resolution when `dest_info.addr.count != 0`, so the taking-over transaction reuses tsx A's resolved address array and then mutates it — `cur_addr++` on retry in `stateless_send_transport_cb()`, and the UDP→TCP threshold rewrite in `stateless_send_resolver_callback()` — while tsx A reads `entry[cur_addr]`. Worst case tsx A copies the wrong remote address into `tsx->addr` and retransmits there.

However, bailing out like the cancelled-tsx check would leave `TSX_HAS_PENDING_TRANSPORT` set, which wedges tsx A permanently: every subsequent `tsx_send_msg()` would just re-set `TSX_HAS_PENDING_SEND`. In the exact 401 scenario that means the ACK deferred at `sip_transaction.c:3603` is never sent and the server keeps retransmitting the 401 until timer D. That is a worse outcome than the misread, so the bail is not taken as-is.

Only the parts that derive from the shared txdata need to be skipped. Note that `TSX_HAS_RESOLVED_SERVER` has to be skipped together with the transport: a UAC with `transport == NULL` and that flag set is terminated outright by `tsx_send_msg()`.

**Point 2.** Correct, and guarding the assignment alone would not be enough: continuing the retry loop would have tsx A transmit tsx B's re-purposed message — different CSeq, branch and Authorization — to the next address. The retry has to stop, not just avoid touching the marker.

### Fix

- `send_msg_callback()` computes `tdata_is_ours` once, right after acquiring the group lock.
- Success path: the transport adoption, the `tsx->addr` copy and `TSX_HAS_RESOLVED_SERVER` are skipped when the txdata is no longer ours, so `tdata->dest_info` is never read. The flag clearing, pending-destroy, pending-send and resched bookkeeping are kept, so the transaction is not left with a pending transport and the deferred ACK still goes out. Not adopting the transport simply makes the next transmission resolve the destination again.
- Failure path: `*cont` is cleared when the txdata is not ours, which routes into the existing `if ((!*cont) || ...)` branch and makes the re-arm at the end of the temporary-failure path unreachable in the takeover case. A one-line comment there notes why no guard is needed.

Both guards are no-ops on the normal path, where `mod_data == tsx` always holds.

### Testing

Rebuilt clean, `pjsip-test` with ASan — 20 tests, 0 failed: `txdata_test`, `dlg_core_test`, `resolve_test`, `inv_offer_answer_test`, `tsx_basic_test`/`tsx_uac_test`/`tsx_uas_test` (x3 each), `transport_loop_test`, `transport_udp_test`, `transport_tcp_test`, `regc_test`, `auth_async_test`, `pjsua_auth_test`, `tsx_destroy_test`.

nanangizz (commented):
Verified the mechanic: the message is encoded *inside* `pjsip_transport_send()` by `on_tx_msg` (`sip_transport.c:1005-1015`), and `msg_size` is captured after that and before the send — so the value is complete, and the new API is necessary rather than a convenience. Callers can't measure before the call because there's nothing encoded yet; `stateless_send_transport_cb()` makes that concrete by invoking `pjsip_tx_data_invalidate_msg()` at line 1310 immediately prior.

Coverage looks complete. Of the sites deriving a byte count from the tdata buffer, the three post-send reads in `sip_util.c` are all fixed; `:1415` (UDP→TCP size check) and `:1636` (`send_raw`) measure before sending, and `tsx_send_msg()` never touches the buffer. The `*cont = PJ_FALSE` reasoning also holds — the failure branch tests `(!*cont)` locally at line 2254, so the "put again pending tdata" block is genuinely unreachable when the tdata isn't ours.

**One thing to resolve.** The `sent > 0` branch now guards `TSX_HAS_RESOLVED_SERVER` with `tdata_is_ours`, but setting `*cont = PJ_FALSE` also creates a *new* path into the failure branch, where the identical assignment at line 2274 stays unguarded — reached possibly right after `tsx_update_transport(tsx, NULL)` at 2248. That branch's unconditional set is pre-existing, so this isn't a regression, but the two are now inconsistent for the same condition the PR is reasoning about. A UAC is terminated shortly after; a UAS on the 503 path at 2314 survives with transport NULL and the flag set. Worth either extending the guard or noting why the failure branch is safe.

**Minor.** `tdata_is_ours` is read under `tsx->grp_lock` while the field is written by the other transaction under a *different* group lock, so this narrows the window rather than closing it. Worth saying so in the comment — the obvious alternative, `tdata->lock`, isn't cheaply available, since it's created as a **null mutex** at `sip_transport.c:516` (the real one commented out at 515), so promoting it would be its own change with a per-message cost and new lock-ordering constraints.

**Suggestions.** The PR has to patch two readers — `buf` via the new API, `dest_info` via `tdata_is_ours` — and that pattern will keep growing. I'd add a line to `pjsip_transport_send()`'s docs stating the invariant: once it returns, no field of `tdata` may be assumed stable. Since `pjsip_tx_data_encode()` guarantees a non-zero size for a valid message, a `pj_assert`/warning on `sent == 0` at the old read sites would also make any future reintroduction loud rather than silent.

On alternatives: cloning the tdata in `pjsip_auth_clt_reinit_req()` rather than reusing it is the actual root fix — it would make both halves of this PR unnecessary. But `pjsip_tx_data_clone()` is response-only (hardcodes `PJSIP_RESPONSE_MSG`, copies only headers and body), so it'd need request-line support plus carry-over of `tp_sel`, `dest_info`, `via_addr`/`via_tp` and `mod_data[]` — a public API semantics change worth its own issue, not something to fold into a crash fix.

sauwming:
Thanks @nanangizz — all four addressed in da61212b7.

**Failure branch `TSX_HAS_RESOLVED_SERVER`.** You're right that clearing `*cont` creates a new path into that branch. Traced the consequence: the flag is only read at `sip_transaction.c:2652`, which is UAC-only, and that branch terminates the UAC at 2297 — so the UAS-survives-with-503 case you flagged never reads it, and the inconsistency is latent rather than live. Still guarded it, since it costs nothing and the asymmetry is exactly what a future reader would trip over.

**Lock scope.** Correct, and I'd overstated it. The comment now says the check only narrows the window in general, and is exact for the authentication retry specifically because the takeover happens inside our own state callback: `tsx_set_state()` at 3624 passes flag `0`, so neither `RELEASE_LOCK` nor the timer condition applies and the group lock is held across `on_tsx_state()`. `send_msg_callback()` therefore blocks on `pj_grp_lock_acquire()` until the takeover has completed. Agreed on `tdata->lock` — a null mutex at `sip_transport.c:516` isn't a usable alternative here.

**Documented invariant.** Added to `pjsip_transport_send()`: once it returns the message is on the wire, so no field of the tdata may be assumed stable, with a pointer to `pjsip_transport_send2()` for the byte count.

**Loud on reintroduction.** `pj_assert(sent > 0)` at all three former read sites, with a one-line note on why it isn't derived from the tdata.

**On cloning.** Agreed it's the actual root fix and agreed it doesn't belong here — `pjsip_tx_data_clone()` hardcodes `PJSIP_RESPONSE_MSG` and copies only the status line, headers and body, so requests would need the request line plus carry-over of `tp_sel`, `dest_info`, `via_addr`/`via_tp` and `mod_data[]`. Happy to open a separate issue for it if you want it tracked.

Retested: `pjsip-test` with ASan, 20 tests, 0 failed.

nanangizz (approved):

Issue #1809, opened 2014-12-24
Premature transaction destroy caused by DNS SRV resolver callback being called twice


pjsipbot:
_2014-12-24 05:40:54: @nanangizz_ _created the issue on trac ticket [1809](https://trac.pjsip.org/repos/ticket/1809)_

Scenario:
 1. run pjsua with one nameserver configured, the library uses the nameserver fine for sometime and it successfully stores some entries in the cache, until somehow the nameserver is down and the library marks it as bad
 1. a registration refresh is about to be done, DNS SRV resolver is used for resolving the SIP registrar address and the DNS server returns two entries from the cache, say server1 and server2
 1. the resolver then tries to resolve the A record of those two entries, which eventually the server1 cannot be resolved due to no working nameserver and the server2 is successfully resolved from the cache
 1. then assertion happens with the following call stack trace:
     ```
#4  0x006edea6 in __assert_fail () from /lib/libc.so.6
#5  0x082d7281 in pj_mutex_unlock (mutex=0xf0e3ee70) at ../src/pj/os_core_unix.c:1287
#6  0x082d6c7e in pj_atomic_inc_and_get (atomic_var=0xf0e3ee68) at ../src/pj/os_core_unix.c:918
#7  0x082d6c94 in pj_atomic_inc (atomic_var=0xf0e3ee68) at ../src/pj/os_core_unix.c:928
#8  0x082ab7c0 in pjsip_tx_data_add_ref (tdata=0xf0e3eb0c) at ../src/pjsip/sip_transport.c:463
#9  0x082ac370 in pjsip_transport_send (tr=0xa57fad4, tdata=0xf0e3eb0c, addr=0xf0e3fb98, addr_len=16, token=0xf0e3fb14, cb=0x82ba0cf <transport_callback>) at ../src/pjsip/sip_transport.c:827
#10 0x082ba40c in tsx_send_msg (tsx=0xf0e3fb14, tdata=0xf0e3eb0c) at ../src/pjsip/sip_transaction.c:2037
#11 0x082baab9 in tsx_retransmit (tsx=0xf0e3fb14, resched=1) at ../src/pjsip/sip_transaction.c:2270
#12 0x082baebf in tsx_on_state_calling (tsx=0xf0e3fb14, event=0xf4bfe188) at ../src/pjsip/sip_transaction.c:2401
[...]

Issue #1108, opened 2010-08-11
DNS deadlock (thanks Bogdan Krakowski for the report)


pjsipbot:
_2010-08-11 08:23:32: @bennylp_ _created the issue on trac ticket [1108](https://trac.pjsip.org/repos/ticket/1108)_

Stack trace:

```
Thread 0:
0   libSystem.B.dylib    b4 semaphore_wait_signal_trap + 8
1   libSystem.B.dylib    ed92 semaphore_wait_signal + 2
2   libSystem.B.dylib    a4 pthread_mutex_lock + 248
3   AbcdVoip             e8 pj_mutex_lock (os_core_unix.c:1195)
4   AbcdVoip             ba8 pj_dns_resolver_start_query (resolver.c:709)
5   AbcdVoip              pj_dns_srv_resolve (srv_resolver.c:148)
6   AbcdVoip             b6398 pjsip_resolve (sip_resolve.c:364)
7   AbcdVoip             c109c pjsip_endpt_resolve (sip_endpoint.c:1043)
8   AbcdVoip             c4168 pjsip_endpt_send_request_stateless (sip_util.c:1360)
9   AbcdVoip             b1db4 tsx_send_msg (sip_transaction.c:1913)
10  AbcdVoip             b2008 tsx_on_state_null (sip_transaction.c:2097)
11  AbcdVoip             b3d84 pjsip_tsx_send_msg (sip_transaction.c:1590)
12  AbcdVoip             c27a4 pjsip_endpt_send_request (sip_util_statefull.c:114)
13  AbcdVoip             d4864 pjsip_regc_send (sip_reg.c:1210)
14  AbcdVoip             e9a04 pjsua_acc_set_registration (pjsua_acc.c:1272)
15  AbcdVoip             eb294 pjsua_acc_add (pjsua_acc.c:330)
16  AbcdVoip             a0 -[CpcAccountImpl registerWithSipServerImpl]
17  Foundation           c86 __NSFireDelayedPerform + 362
18  CoreFoundation       a54 __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__ + 8
19  CoreFoundation       ede __CFRunLoopDoTimer + 854
20  CoreFoundation       e __CFRunLoopRun + 1082
21  CoreFoundation       d8e4 CFRunLoopRunSpecific + 224
22  CoreFoundation       d7ec CFRunLoopRunInMode + 52
[...]

Issue #1565, opened 2012-08-07
DNS deadlock


pjsipbot:
_2012-08-07 04:28:58: @nanangizz_ _created the issue on trac ticket [1565](https://trac.pjsip.org/repos/ticket/1565)_

Reported that there is DNS deadlock issue similar to #1108, the affected function is `on_timeout()`, also reported that the similar fix (as in #1108) seems to work.

Thanks Hermant KK for the report.
___
_2012-08-07 04:32:36: @nanangizz_ _changed **status** from **new** to **closed**_

___
_2012-08-07 04:32:36: @nanangizz_ _set **resolution** to **fixed**_

___
_2012-08-07 04:32:36: @nanangizz_ _commented_

In [r4220](../commit/911260c3cb2d4250500862f0fa96f7628e65d71f):
Fixed #1565: deadlock in DNS resolver.

___
_2012-08-07 04:33:43: @nanangizz_ _edited the issue description_

___
_2013-02-27 11:37:03: @mingteluu_ _commented_

In [r4390](../commit/864e7129e3ffd4104930a0ac4dde81e9e78469d6):
Re #1565: Backported to 1.x

PR #5181, opened 2026-08-12
resolver: capture and clear the query callback under the group lock in the timeout and read-complete paths

vlaushkin:
A DNS query callback can be invoked twice or called through a NULL pointer when
`pj_dns_resolver_cancel_query()` runs on another thread concurrently with the delivery
of that query's response or timeout.

Follow-up to #5167 (the two remaining call sites @nanangizz asked to track separately).

## Problem

`pj_dns_resolver_cancel_query()` captures and clears `query->cb` under the group lock
before invoking it. `on_timeout()` and `on_read_complete()` do the opposite: to avoid
the #1108/#1565 deadlock they release the group lock first, then read and invoke `q->cb`
(and each child `cq->cb`) with no lock held:

```c
/* on_read_complete(), before */
pj_grp_lock_release(resolver->grp_lock);
...
if (q->cb)
    (*q->cb)(q->user_data, status, dns_pkt);
```

A `cancel_query()` running on any other thread — an application thread holding the query
handle, or a second worker — races that unlocked read:

- it clears `q->cb` between the `if (q->cb)` check and the `(*q->cb)` deref, turning the
  call into a jump through a NULL pointer, or
- the notify path never clears `q->cb`, so the callback fires here and again from
  `cancel_query(notify=TRUE)` for the same query (double delivery, cf. #1809).

#5167 fixed this class for `pj_dns_resolver_destroy()`; `on_timeout()` and
`on_read_complete()` are the remaining call sites.

## Details

Both paths now mirror the #5167 destroy path: capture and clear each callback under the
group lock, then invoke the saved pointer after releasing it, for the parent query and
each child (`cancel_query()` already does the capture-and-clear half under the lock):

```c
/* on_read_complete(), after */
pj_grp_lock_acquire(resolver->grp_lock);
cb = q->cb;
q->cb = NULL;
pj_grp_lock_release(resolver->grp_lock);

if (cb)
    (*cb)(q->user_data, status, dns_pkt);
```

The child loop reads `next = child_q->next` before the release window; child nodes live
on the parent's `child_head` only (they are never in `hquerybyid`/`hquerybyres`), and the
parent was already removed from both hashes under the lock before the notify loops, so no
other thread can recycle the query or its children while the callbacks run. The
callback-before-cache ordering in `on_read_complete()` (notify first, then
`update_res_cache()`) is unchanged. No callback is invoked while holding the lock, so the
#1108/#1565 deadlock avoidance is preserved.

This also fixes a user-visible re-entrancy case on a single thread: because the callback pointer was previously read but not cleared before being invoked, an application that cancels a query (with `notify`) from inside that query's own completion callback received a second, `PJ_ECANCELLED`, delivery. This applies to child queries too, since a child is handed to the application as a query handle. Capturing and clearing the callback under the lock before invoking makes each query's callback fire exactly once.

This relies on the `if (notify && cb)` guard `cancel_query()` gained in #5167: once the
timeout/read-complete paths clear `q->cb`, a losing `cancel_query(notify=TRUE)` must skip
the now-NULL pointer.

## Testing

`pjlib-util-test resolver_test` passes on macOS (arm64) and Linux (gcc 13,
aarch64-unknown-linux-gnu), 0 failed, including under an ASan build. Three new deterministic tests pin the exactly-once contract, all red on master and green here, no thread interleaving: `dns_cancel_in_callback_test` (cancel from inside the completion callback, response path), `dns_cancel_in_timeout_callback_test` (same through `on_timeout()`, servers ignore the query), and `dns_cancel_child_in_callback_test` (a coalesced child cancels itself from its callback, covering the per-child capture loop). Each test also asserts its cancel actually ran, so a vacuous pass fails at the guard. The race is a few instructions wide; it was
reproduced deterministically on stock master by widening the window (a `pj_thread_sleep`
between the `if (q->cb)` check and the deref) and driving
`pj_dns_resolver_cancel_query()` from another thread, which faults through a NULL
callback. With this change the pointer is captured and cleared under the lock before
that window exists, so the cancel either runs first (delivery here is skipped) or reads
NULL and skips.

### DIFF
--- a/pjlib-util/src/pjlib-util-test/resolver_test.c
+++ b/pjlib-util/src/pjlib-util-test/resolver_test.c
@@ -76,6 +76,16 @@ static pj_status_t start_during_destroy_status;
 static pj_dns_async_query *start_during_destroy_query;
 static int cancel_in_cb_count;
 static pj_dns_async_query *cancel_in_cb_query;
+static pj_bool_t cancel_in_cb_cancelled;
+static int cancel_in_tocb_count;
+static pj_dns_async_query *cancel_in_tocb_query;
+static pj_bool_t cancel_in_tocb_cancelled;
+static pj_status_t cancel_in_tocb_status;
+static int cancel_child_parent_count;
+static int cancel_child_child_count;
+static pj_dns_async_query *cancel_child_query;
+static pj_bool_t cancel_child_cancelled;
+static pj_status_t cancel_child_status;
 
 #define MAX_LABEL   32
 
@@ -1227,6 +1237,7 @@ static void dns_callback_cancel_in_cb(void *user_data,
     if (cancel_in_cb_count == 1 && cancel_in_cb_query) {
         pj_dns_async_query *q = cancel_in_cb_query;
         cancel_in_cb_query = NULL;
+        cancel_in_cb_cancelled = PJ_TRUE;
         pj_dns_resolver_cancel_query(q, PJ_TRUE);
     }
 
@@ -1246,6 +1257,7 @@ static int dns_cancel_in_callback_test(void)
 
     cancel_in_cb_count = 0;
     cancel_in_cb_query = NULL;
+    cancel_in_cb_cancelled = PJ_FALSE;
 
     /* Servers answer so the query completes and the callback fires. */
     g_server[0].action = PJ_DNS_RCODE_NXDOMAIN;
@@ -1254,14 +1266,204 @@ static int dns_cancel_in_callback_test(void)
     PJ_TEST_SUCCESS(pj_dns_resolver_start_query(
                         resolver, &name, PJ_DNS_TYPE_A, 0,
                         &dns_callback_cancel_in_cb, NULL, &cancel_in_cb_query),
-                    NULL, return -600);
+                    NULL, return -670);
 
     pj_sem_wait(sem);
 
-    /* Give any erroneous second (PJ_ECANCELLED) delivery time to arrive. */
-    pj_thread_sleep(1000);
+    /* Guard against a vacuous pass: the assertions below only pin the
+     * behavior if the cancel actually ran inside the callback.
+     */
+    PJ_TEST_EQ(cancel_in_cb_cancelled, PJ_TRUE, NULL, return -672);
+    PJ_TEST_EQ(cancel_in_cb_count, 1, NULL, return -674);
 
-    PJ_TEST_EQ(cancel_in_cb_count, 1, NULL, return -610);
+    return 0;
+}
+
+
+/* Same as dns_callback_cancel_in_cb, but for the timeout variant below. */
+static void dns_callback_cancel_in_tocb(void *user_data,
+                                        pj_status_t status,
+                                        pj_dns_parsed_packet *resp)
+{
+    PJ_UNUSED_ARG(user_data);
+    PJ_UNUSED_ARG(resp);
+
+    cancel_in_tocb_count++;
+
+    if (cancel_in_tocb_count == 1) {
+        cancel_in_tocb_status = status;
+        if (cancel_in_tocb_query) {
+            pj_dns_async_query *q = cancel_in_tocb_query;
+            cancel_in_tocb_query = NULL;
+            cancel_in_tocb_cancelled = PJ_TRUE;
+            pj_dns_resolver_cancel_query(q, PJ_TRUE);
+        }
+    }
+
+    pj_sem_post(sem);
+}
+
+
+/* Same as dns_cancel_in_callback_test, but through the timeout path: the
+ * servers do not respond, so the callback is delivered by on_timeout()
+ * rather than on_read_complete(). Uses a dedicated resolver with short
+ * retransmission settings to keep the timeout fast.
+ */
+static int dns_cancel_in_timeout_callback_test(void)
+{
+    pj_str_t name = pj_str("name_cancel_in_tocb");
+    pj_str_t nameservers[2];
+    pj_uint16_t ports[2];
+    pj_dns_settings lset;
+    pj_dns_resolver *res;
+
+    PJ_LOG(3,(THIS_FILE, "  cancel query from within its timeout callback test"));
+
+    cancel_in_tocb_count = 0;
+    cancel_in_tocb_query = NULL;
+    cancel_in_tocb_cancelled = PJ_FALSE;
+    cancel_in_tocb_status = PJ_SUCCESS;
+
+    /* Servers ignore the query so it times out. */
+    g_server[0].action = ACTION_IGNORE;
+    g_server[1].action = ACTION_IGNORE;
+
+    nameservers[0] = nameservers[1] = pj_str("127.0.0.1");
+    ports[0] = g_server[0].port;
+    ports[1] = g_server[1].port;
+
+    PJ_TEST_SUCCESS(pj_dns_resolver_create(mem, NULL, 0, timer_heap, ioqueue,
+                                           &res),
+                    NULL, return -650);
+    PJ_TEST_SUCCESS(pj_dns_resolver_set_ns(res, 2, nameservers, ports),
+                    NULL, return -651);
+
+    pj_dns_resolver_get_settings(res, &lset);
+    lset.qretr_delay = 200;
+    lset.qretr_count = 1;
+    PJ_TEST_SUCCESS(pj_dns_resolver_set_settings(res, &lset),
+                    NULL, return -652);
+
+    PJ_TEST_SUCCESS(pj_dns_resolver_start_query(
+                        res, &name, PJ_DNS_TYPE_A, 0,
+                        &dns_callback_cancel_in_tocb, NULL,
+                        &cancel_in_tocb_query),
+                    NULL, return -653);
+
+    pj_sem_wait(sem);
+
+    PJ_TEST_EQ(cancel_in_tocb_status, PJ_ETIMEDOUT, NULL, return -654);
+    PJ_TEST_EQ(cancel_in_tocb_cancelled, PJ_TRUE, NULL, return -655);
+    PJ_TEST_EQ(cancel_in_tocb_count, 1, NULL, return -656);
+
+    pj_dns_resolver_destroy(res, PJ_FALSE);
+
+    return 0;
+}
+
+
+static void dns_callback_cancel_child_parent(void *user_data,
+                                             pj_status_t status,
+                                             pj_dns_parsed_packet *resp)
+{
+    PJ_UNUSED_ARG(user_data);
+    PJ_UNUSED_ARG(status);
+    PJ_UNUSED_ARG(resp);
+
+    cancel_child_parent_count++;
+
+    pj_sem_post(sem);
+}
+
+/* Callback of the coalesced child query; cancels its own query on the first
+ * delivery, like dns_callback_cancel_in_cb.
+ */
+static void dns_callback_cancel_child(void *user_data,
+                                      pj_status_t status,
+                                      pj_dns_parsed_packet *resp)
+{
+    PJ_UNUSED_ARG(user_data);
+    PJ_UNUSED_ARG(resp);
+
+    cancel_child_child_count++;
+
+    if (cancel_child_child_count == 1) {
+        cancel_child_status = status;
+        if (cancel_child_query) {
+            pj_dns_async_query *q = cancel_child_query;
+            cancel_child_query = NULL;
+            cancel_child_cancelled = PJ_TRUE;
+            pj_dns_resolver_cancel_query(q, PJ_TRUE);
+        }
+    }
+
+    pj_sem_post(sem);
+}
+
+
+/* Same again, but cancelling a coalesced CHILD query from inside its own
+ * callback, to cover the child (ccb) capture loop. The servers ignore the
+ * query, so the parent stays pending in the resolver's hash tables until
+ * its timeout fires; the second start_query() below is therefore guaranteed
+ * to join it as a child, and both callbacks are delivered by the same
+ * on_timeout() pass.
+ */
+static int dns_cancel_child_in_callback_test(void)
+{
+    pj_str_t name = pj_str("name_cancel_child_tocb");
+    pj_str_t nameservers[2];
+    pj_uint16_t ports[2];
+    pj_dns_settings lset;
+    pj_dns_resolver *res;
+
+    PJ_LOG(3,(THIS_FILE, "  cancel child query from within its callback test"));
+
+    cancel_child_parent_count = 0;
+    cancel_child_child_count = 0;
+    cancel_child_query = NULL;
+    cancel_child_cancelled = PJ_FALSE;
+    cancel_child_status = PJ_SUCCESS;
+
+    /* Servers ignore the query so it times out. */
+    g_server[0].action = ACTION_IGNORE;
+    g_server[1].action = ACTION_IGNORE;
+
+    nameservers[0] = nameservers[1] = pj_str("127.0.0.1");
+    ports[0] = g_server[0].port;
+    ports[1] = g_server[1].port;
+
+    PJ_TEST_SUCCESS(pj_dns_resolver_create(mem, NULL, 0, timer_heap, ioqueue,
+                                           &res),
+                    NULL, return -660);
+    PJ_TEST_SUCCESS(pj_dns_resolver_set_ns(res, 2, nameservers, ports),
+                    NULL, return -661);
+
+    pj_dns_resolver_get_settings(res, &lset);
+    lset.qretr_delay = 200;
+    lset.qretr_count = 1;
+    PJ_TEST_SUCCESS(pj_dns_resolver_set_settings(res, &lset),
+                    NULL, return -662);
+
+    PJ_TEST_SUCCESS(pj_dns_resolver_start_query(
+                        res, &name, PJ_DNS_TYPE_A, 0,
+                        &dns_callback_cancel_child_parent, NULL, NULL),
+                    NULL, return -663);
+    PJ_TEST_SUCCESS(pj_dns_resolver_start_query(
+                        res, &name, PJ_DNS_TYPE_A, 0,
+                        &dns_callback_cancel_child, NULL,
+                        &cancel_child_query),
+                    NULL, return -664);
+
+    /* One delivery for the parent, one for the child. */
+    pj_sem_wait(sem);
+    pj_sem_wait(sem);
+
+    PJ_TEST_EQ(cancel_child_status, PJ_ETIMEDOUT, NULL, return -665);
+    PJ_TEST_EQ(cancel_child_cancelled, PJ_TRUE, NULL, return -666);
+    PJ_TEST_EQ(cancel_child_child_count, 1, NULL, return -667);
+    PJ_TEST_EQ(cancel_child_parent_count, 1, NULL, return -668);
+
+    pj_dns_resolver_destroy(res, PJ_FALSE);
 
     return 0;
 }
@@ -2150,6 +2352,16 @@ int resolver_test(void)
 
     PJ_LOG(3,(THIS_FILE, "dns_cancel_in_callback_test"));
     rc = dns_cancel_in_callback_test();
+    if (rc != 0)
+        goto on_error;
+
+    PJ_LOG(3,(THIS_FILE, "dns_cancel_in_timeout_callback_test"));
+    rc = dns_cancel_in_timeout_callback_test();
+    if (rc != 0)
+        goto on_error;
+
+    PJ_LOG(3,(THIS_FILE, "dns_cancel_child_in_callback_test"));
+    rc = dns_cancel_child_in_callback_test();
     if (rc != 0)
         goto on_error;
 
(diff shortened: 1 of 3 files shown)


pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1665,16 +1666,32 @@ static void on_timeout( pj_timer_heap_t *timer_heap,
...
-    if (q->cb)
-        (*q->cb)(q->user_data, PJ_ETIMEDOUT, NULL);
+    /* Capture and clear each callback under the lock, but invoke it after
+     * releasing, so it neither runs under the lock nor races a concurrent
+     * cancel into a duplicate call.
+     */
+    pj_grp_lock_acquire(resolver->grp_lock);
+    cb = q->cb;
+    q->cb = NULL;
+    pj_grp_lock_release(resolver->grp_lock);
```
sauwming: The lock is released on the line above and re-acquired here with nothing in between. The parent's capture-and-clear is already reachable inside the held lock — please move `cb = q->cb; q->cb = NULL;` up to just after the hash removals, before the existing release.

Two reasons beyond saving a lock round-trip:

1. It makes clearing `q->cb` **atomic with removal from `hquerybyid`/`hquerybyres`**, which is the property that actually makes the query "committed".
2. As written there is still a window between the release above and this acquire where a concurrent `pj_dns_resolver_cancel_query(q, PJ_TRUE)` sees a non-NULL `q->cb` and delivers `PJ_ECANCELLED` — even though the timer is already cancelled and the query is out of both hash tables. Not a crash, and cancel-wins is defensible, but it costs nothing to close.

The per-child acquire/release below has to stay as it is — there's no way to stash N child callbacks under a single lock hold without allocating.

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1812,21 +1830,38 @@ static void on_read_complete(pj_ioqueue_key_t *key,
...
+     * Capture and clear each callback under the lock, but invoke it after
+     * releasing, so it neither runs under the lock nor races a concurrent
+     * cancel into a duplicate call.
      */
-    if (q->cb)
-        (*q->cb)(q->user_data, status, dns_pkt);
+    pj_grp_lock_acquire(resolver->grp_lock);
+    cb = q->cb;
+    q->cb = NULL;
+    pj_grp_lock_release(resolver->grp_lock);
```
sauwming: Same as in `on_timeout()`: this acquire immediately follows the release on line 1831 with nothing in between. Please hoist `cb = q->cb; q->cb = NULL;` to just after the hash removals on lines 1827-1828, so the clear is atomic with removal from both hash tables and the cancel window between release and re-acquire disappears.

The per-child capture in the loop below still needs its own acquire/release.

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1665,16 +1666,32 @@ static void on_timeout( pj_timer_heap_t *timer_heap,
     /* Workaround for deadlock problem in #1565 (similar to #1108) */
     pj_grp_lock_release(resolver->grp_lock);
 
-    /* Call application callback, if any. */
-    if (q->cb)
-        (*q->cb)(q->user_data, PJ_ETIMEDOUT, NULL);
+    /* Capture and clear each callback under the lock, but invoke it after
+     * releasing, so it neither runs under the lock nor races a concurrent
+     * cancel into a duplicate call.
+     */
```
sauwming: Minor: this paragraph now appears three times in the file (here, in `on_read_complete()`, and in `pj_dns_resolver_destroy()` from #5167). Project convention is minimal comments — suggest trimming the two new copies to a single line and letting the #5167 comment carry the full rationale:

```c
/* Capture and clear the callback under the lock; invoke it unlocked. */
```

sauwming (commented):
Thanks for splitting this out from #5167 — the analysis in the description is accurate and the fix is the right shape. I applied it on top of `12306eaf9` and built pjlib-util with ASan on aarch64-apple-darwin: zero warnings, `resolver_test` passes (1/1, 0 failed).

A few things I verified while reading, and a few requests inline.

### Confirmed sound

- Query lifetime across the unlocked window is safe: `on_timeout()` holds a group-lock reference via `pj_timer_heap_schedule_w_grp_lock()` (resolver.c:748) and `on_read_complete()` via `pj_ioqueue_register_sock2(..., resv->grp_lock, ...)` (resolver.c:288, 335), so a concurrent `pj_dns_resolver_destroy()` cannot release the pool underneath either callback.
- Saving `next` before the release window is necessary and correctly done in both child loops.
- Callback ordering relative to `update_res_cache()` is unchanged, and no callback runs under the lock, so the #1108/#1565 avoidance is preserved.

### Worth adding to the description

This also fixes a re-entrancy bug that isn't called out. Child queries are handed to applications as query handles (resolver.c:1035, when a duplicate query is already pending), so an application can legitimately call `pj_dns_resolver_cancel_query()` on a child from inside the *parent's* callback. Before this change the child's callback fired anyway with the result; now it is correctly suppressed. That's a user-visible fix, not just hardening.

### Follow-ups (not for this PR)

- `pj_dns_resolver_cancel_query()` still invokes the callback while holding the group lock (resolver.c:1099-1100) — precisely what #1108/#1565 exist to prevent. It is pre-existing, but this PR makes the losing thread defer to `cancel_query()`, so that under-lock delivery path will be exercised more often than before. Worth a tracking issue.
- `on_timeout()`'s recycle loop (resolver.c:1698-1710 on this branch) pushes children onto `query_free_nodes` without `pj_list_erase()` first, unlike `on_read_complete()`. Benign, since `alloc_qnode()` re-inits `child_head` on reuse, but the asymmetry is confusing.

### Test coverage

`resolver_test.c` has no `pj_dns_resolver_cancel_query()` coverage at all today. A deterministic race test isn't practical, but a non-racy regression test is cheap and would pin the contract this PR establishes: start a query, cancel it with `notify=PJ_TRUE`, then let the response or timeout fire, and assert the callback ran **exactly once**. That fails on current master (double delivery) and passes with this change, with no thread interleaving required. Would you be up for adding one?

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1812,21 +1830,38 @@ static void on_read_complete(pj_ioqueue_key_t *key,
...
     pj_grp_lock_release(resolver->grp_lock);
 
-    /* Notify applications first, to allow application to modify the 
+    /* Notify applications first, to allow application to modify the
      * record before it is saved to the hash table.
+     *
+     * Capture and clear each callback under the lock, but invoke it after
+     * releasing, so it neither runs under the lock nor races a concurrent
+     * cancel into a duplicate call.
      */
```
sauwming: Same comment-duplication nit as in `on_timeout()` — suggest collapsing the added paragraph to one line here too, keeping the original "Notify applications first" text intact:

```c
    /* Notify applications first, to allow application to modify the
     * record before it is saved to the hash table. Capture and clear
     * the callback under the lock; invoke it unlocked.
     */
```

(The trailing-whitespace cleanup on line 1833 is a welcome drive-by.)

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1665,16 +1666,32 @@ static void on_timeout( pj_timer_heap_t *timer_heap,
...
     /* Call application callback for child queries. */
     cq = q->child_head.next;
     while (cq != (void*)&q->child_head) {
-        if (cq->cb)
-            (*cq->cb)(cq->user_data, PJ_ETIMEDOUT, NULL);
-        cq = cq->next;
+        pj_dns_async_query *next = cq->next;
+
+        pj_grp_lock_acquire(resolver->grp_lock);
+        cb = cq->cb;
```
sauwming: Nit for consistency with the #5167 shape this mirrors: `pj_dns_resolver_destroy()` declares `cb`/`ccb` in inner block scope rather than reusing one function-scope variable for both the parent and the children. This block already opens with a declaration, so `pj_dns_callback *cb;` could live here and the function-scope declaration on line 1624 would go away. Both are valid C89; matching `destroy()` just reads better given the PR explicitly follows it.

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1812,21 +1830,38 @@ static void on_read_complete(pj_ioqueue_key_t *key,
...
     /* If query has subqueries, notify subqueries's application callback */
     if (!pj_list_empty(&q->child_head)) {
         pj_dns_async_query *child_q;
 
         child_q = q->child_head.next;
         while (child_q != (pj_dns_async_query*)&q->child_head) {
-            if (child_q->cb)
-                (*child_q->cb)(child_q->user_data, status, dns_pkt);
-            child_q = child_q->next;
+            pj_dns_async_query *next = child_q->next;
```
sauwming, line 1862: Capturing `next` before the unlocked callback is the right call and worth keeping — an application callback here can legitimately cancel a sibling query, and this ordering keeps the traversal valid. Same holds in the `on_timeout()` loop.

vlaushkin:
Thanks — all addressed in the follow-up commit, and thanks for building it with ASan.

**Confirmed-sound points** — agreed on all three (group-lock reference keeps the pool alive across the unlocked window; `next` saved before each callback; ordering vs `update_res_cache()` unchanged). Thanks for verifying those.

**Inline fixes:**
- Hoisted the parent capture under the held lock in both callbacks — `cb = q->cb; q->cb = NULL;` now runs right after the hash removals, before the existing release, so clearing the callback is atomic with removal from both hash tables and the cancel window is gone. The per-child capture keeps its own acquire/release.
- Collapsed the comments to your suggested one-liner in `on_timeout()` and the merged form keeping "Notify applications first" in `on_read_complete()`; left the `#1565`/`#1108` release comments intact.
- Gave the child loop its own `ccb` so parent and child no longer share a variable. I kept the parent's `cb` at function scope: unlike `destroy()`'s loop body, the parent capture spans the lock release, so a block to scope it would have to enclose the release and reads worse. Happy to switch if you'd prefer it match `destroy()` exactly.
- Kept saving `next` before the callback, as you confirmed.

**Test** — added `dns_cancel_in_callback_test`: it starts a query and, from its completion callback, cancels the query with `notify=PJ_TRUE`, asserting the callback runs exactly once. It fails on master (`cancel_in_cb_count == 2`, the extra `PJ_ECANCELLED`) and passes here, no thread interleaving. Verified on macOS and Linux.

**Re-entrancy in the description** — added, but framed around a query cancelled from inside its *own* callback, which is what the test reproduces. On the specific "cancel a child from inside the parent's callback" case: on current master the child loop re-reads `child_q->cb` before invoking (`if (child_q->cb)`), so a child cancelled during the parent's callback is already skipped there and receives only the `PJ_ECANCELLED` — it doesn't fire with the result, and this PR doesn't change that path. So I kept the wording to the self-cancel case to stay precise. Let me know if you'd rather word it differently, or if there's a variant I'm missing.

**Follow-ups:**
- `cancel_query()` under-lock delivery — opened #5187 to track it; pre-existing and out of scope here.
- `on_timeout()` recycle asymmetry — left out to keep this PR scoped, as you suggested. Happy to add the `pj_list_erase()` here if you'd rather close it in passing, otherwise I'll leave it noted.

pjlib-util/src/pjlib-util-test/resolver_test.c
```diff
@@ -1207,6 +1209,64 @@ static int dns_start_during_destroy_test(void)
...
+                        resolver, &name, PJ_DNS_TYPE_A, 0,
+                        &dns_callback_cancel_in_cb, NULL, &cancel_in_cb_query),
+                    NULL, return -600);
+
+    pj_sem_wait(sem);
+
+    /* Give any erroneous second (PJ_ECANCELLED) delivery time to arrive. */
+    pj_thread_sleep(1000);
+
+    PJ_TEST_EQ(cancel_in_cb_count, 1, NULL, return -610);
```
sauwming: Empirically, this assertion alone doesn't pin the behavior: it passes whether or not the cancel actually ran. I neutered the cancel entirely on top of your branch (`if (0 && cancel_in_cb_count == 1 && ...)`) and the test still reports `[OK]` — `cancel_in_cb_count` is 1 in both the "cancel correctly suppressed the second delivery" case and the "cancel never happened" case.

The guard on line 1227 is satisfied today in the async path, since `pj_dns_resolver_start_query()` assigns `*p_query` while still holding the group lock and `on_read_complete()` needs that lock to deliver. But the cache-hit path sets `*p_query = NULL` and invokes the callback synchronously (resolver.c:995-1004), so if this name ever resolves from cache — a cached negative entry, a future change to how the test servers are primed, or someone reusing the name — the guard silently goes false and this test degrades to a no-op that still passes.

Cheap to close: record that the cancel actually ran and assert it.

```c
static pj_bool_t cancel_in_cb_cancelled;
...
    if (cancel_in_cb_count == 1 && cancel_in_cb_query) {
        pj_dns_async_query *q = cancel_in_cb_query;
        cancel_in_cb_query = NULL;
        cancel_in_cb_cancelled = PJ_TRUE;
        pj_dns_resolver_cancel_query(q, PJ_TRUE);
    }
...
    PJ_TEST_EQ(cancel_in_cb_cancelled, PJ_TRUE, NULL, return -605);
    PJ_TEST_EQ(cancel_in_cb_count, 1, NULL, return -610);
```

Then a vacuous run fails at -605 instead of passing green.

pjlib-util/src/pjlib-util-test/resolver_test.c
```diff
@@ -1207,6 +1209,64 @@ static int dns_start_during_destroy_test(void)
...
+
+    PJ_TEST_SUCCESS(pj_dns_resolver_start_query(
+                        resolver, &name, PJ_DNS_TYPE_A, 0,
+                        &dns_callback_cancel_in_cb, NULL, &cancel_in_cb_query),
+                    NULL, return -600);
+
+    pj_sem_wait(sem);
+
+    /* Give any erroneous second (PJ_ECANCELLED) delivery time to arrive. */
+    pj_thread_sleep(1000);
```
sauwming: Optional: this second is dead time. The double delivery you're guarding against is synchronous — on master it happens inside `pj_dns_resolver_cancel_query()` before the callback returns, so `cancel_in_cb_count` is already 2 by the time `pj_sem_wait()` on line 1259 returns. My master run confirms it: the failure is reported ~1.0s after the test starts, i.e. the count was wrong before the sleep, not because of it.

So this only insures against a hypothetical future *asynchronous* second delivery. A second per CI run is cheap enough that I don't feel strongly, but it isn't buying coverage for the bug in question — worth dropping or shortening if you agree.

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1679,19 +1680,31 @@ static void on_timeout( pj_timer_heap_t *timer_heap,
     pj_hash_set(NULL, resolver->hquerybyid, &q->id, sizeof(q->id), 0, NULL);
     pj_hash_set(NULL, resolver->hquerybyres, &q->key, sizeof(q->key), 0, NULL);
 
+    /* Capture and clear the callback under the lock; invoke it unlocked. */
+    cb = q->cb;
+    q->cb = NULL;
```
sauwming, line 1685: This is what I was after — the clear is now atomic with the removal from both hash tables on lines 1680-1681, so there's no longer a window where a concurrent `cancel_query(notify=PJ_TRUE)` sees a live `cb` on a query that's already committed. Same in `on_read_complete()`. Thanks.

pjlib-util/src/pjlib-util/resolver.c
```diff
@@ -1679,19 +1680,31 @@ static void on_timeout( pj_timer_heap_t *timer_heap,
...
+        (*cb)(q->user_data, PJ_ETIMEDOUT, NULL);
 
     /* Call application callback for child queries. */
     cq = q->child_head.next;
     while (cq != (void*)&q->child_head) {
-        if (cq->cb)
-            (*cq->cb)(cq->user_data, PJ_ETIMEDOUT, NULL);
-        cq = cq->next;
+        pj_dns_async_query *next = cq->next;
+        pj_dns_callback *ccb;
```
sauwming, line 1697: Giving the child loop its own `ccb` reads much better. Worth noting for the record that the unlocked `cq->next` read on the first line here is safe: children are only appended under the lock via the `hquerybyres` lookup in `start_query()` (resolver.c:1041), and the parent was removed from that hash before this loop, so the list can't be mutated by another thread while it's walked.

sauwming (commented):
All five inline points are addressed, and the test does what you say it does. I rebuilt on the new head with ASan on aarch64-apple-darwin (both files are byte-identical to the PR base `5b0894b8`, so this is a clean measurement):

- **Test fails on master**, exactly as described: `Test "cancel_in_cb_count" (value=2) == "1" (value=1) fails in resolver_test.c:1264`.
- **Full PR passes**: `resolver_test [OK]`, 0 failed, clean build, no ASan reports.

That's a genuine regression test for the bug this PR fixes — thanks for adding it.

### You're right about the child case, I was wrong

I checked master again and your reading is correct: the child loop re-reads `if (child_q->cb)` at loop time, which is after the parent's callback has returned, so a child cancelled from inside the parent's callback is already skipped on master and receives only the `PJ_ECANCELLED`. There is no second delivery there and this PR doesn't change that path. Your narrower framing around the self-cancel case is the accurate one — keep the description as you have it.

### The header doc now overpromises

This is the one thing I'd still like changed. `pj_dns_resolver_cancel_query()`'s doxygen (resolver.h:431-432) says:

> `@param notify` If non-zero, the callback will be called with failure status to notify that the query has been cancelled.

After this PR that is conditionally false, which is the whole point of the fix: once the timeout/read-complete path has captured and cleared the callback, `notify=PJ_TRUE` delivers nothing. Your own test asserts exactly that. The behavior is right, but the contract should say so — otherwise the next person to hit it reads the header and files a bug. Something like:

```c
 * @param notify    If non-zero, the callback will be called with failure
 *                  status to notify that the query has been cancelled,
 *                  unless the query has already completed and its callback
 *                  has already been invoked.
```

### Follow-ups

- Thanks for opening #5187 for the `cancel_query()` under-lock delivery.
- Agreed on leaving the `on_timeout()` recycle asymmetry out of scope.
- Coverage gaps, both optional: the `on_timeout()` hoist and the child (`ccb`) loops are still untested. The timeout path is reachable by letting the servers not respond and cancelling from the timeout callback, if you think it's worth the extra wall-clock in CI.

On the `cb` scope question — function scope for the parent is fine, your reasoning about the capture spanning the release is right. Not worth churning.

Once the doc note is in, this is good to merge from my side.

vlaushkin:
Thanks for the re-review, for the ASan rebuild, and for double-checking the child case — I appreciate the confirmation, and the description stays as is.

All three points addressed in the follow-up commit:

- **Header doc** — updated `pj_dns_resolver_cancel_query()`'s `@param notify` with exactly your wording. I also checked `pj_dns_resolver_destroy()`'s `notify` doc for the same drift: it's accurate as is, since it promises the callback only for *pending* queries, and a query whose callback already ran is no longer pending.
- **Vacuous-pass guard** — you're right, and your `if (0 && ...)` experiment was a fair catch. Added the `cancel_in_cb_cancelled` flag with its assert ahead of the count check. On error codes: your suggested `-605` is already taken by `srv_resolver_fallback_test` — and so were the `-600`/`-610` this test picked in the first round, so the test now uses a free `-670..-674` range. Verified the guard bites the same way you did: with the cancel neutered on this branch, the test fails at the flag assert instead of passing.
- **The 1-second sleep** — dropped. Your trace matches mine: the second delivery is synchronous inside `cancel_query()`, so the count is already wrong (or right) before `pj_sem_wait()` returns; the sleep bought nothing.

On the coverage gaps, both halves turned out to be doable:

- **Timeout path** — added `dns_cancel_in_timeout_callback_test`: same shape, but the servers ignore the query so delivery goes through `on_timeout()`. It uses a dedicated resolver with `qretr_delay=200, qretr_count=1`, so it costs well under a second rather than the ~10s the default settings would take — which seemed to answer the wall-clock concern. It asserts the first delivery is `PJ_ETIMEDOUT` (so it keeps pinning the timeout path even if the harness changes), the cancel ran, and the callback fired exactly once. On master it fails with the count at 2 through the timeout path — the `on_timeout()` hoist doing the work.
- **Child (`ccb`) loop** — I initially wrote this off as timing-dependent, but there is a deterministic construction: with the servers ignoring the query, the parent stays pending in the hash tables until its timeout fires, so a second `start_query()` for the same name is *guaranteed* to coalesce as a child — no race against a response. `dns_cancel_child_in_callback_test` does that and has the child cancel itself from inside its own callback, which exercises the per-child capture in `on_timeout()`'s loop. On master it fails with `cancel_child_child_count == 2` (the second `PJ_ECANCELLED` delivered through the still-set child callback); with this PR the child fires exactly once, and the parent's single delivery is asserted alongside.

And thanks for writing down the `cq->next` safety argument — matches my reading (children are only appended under the lock, and the parent is out of `hquerybyres` before the loop).

All three tests were also checked the way you checked the first one — neutering each cancel makes its test fail at the corresponding guard assert rather than pass. `resolver_test` passes on macOS and Linux, and an ASan build of the full branch runs it clean.

sauwming (approved):

nanangizz (approved):

PR #5182, opened 2026-08-12
resolver: reject new queries once destroy has started

vlaushkin:
A DNS query started while `pj_dns_resolver_destroy()` is already running slips past the
teardown, leaving a retransmit timer armed in a heap the teardown destroys, or firing
later against the closed socket.

Follow-up to #5167 (second of the two remaining items @nanangizz asked to track separately).

## Problem

`pj_dns_resolver_destroy()` collects the pending queries and cancels their timers under
the group lock (#5167). A `pj_dns_resolver_start_query()` that acquires the lock after
that snapshot creates a query the snapshot never saw: its retransmit timer is armed in a
timer heap the teardown below is about to destroy (`own_timer`), or fires later against
the closed socket. The query also lands in hash tables destroy has already swept.

The common vector is a second thread: a resolution in flight (`srv_resolver.c`'s
fallback, `sip_resolve.c`) calls `start_query()` on a worker thread while another thread
destroys the resolver — pjsip itself tears it down this way (`sip_resolve.c` destroys
with `notify=FALSE`). With `destroy(notify=TRUE)` no second thread is even needed:
`sip_resolve.c` always sets the `PJ_DNS_SRV_FALLBACK_*` flags, and on an unsuccessful
SRV status — including the `PJ_ECANCELLED` delivered by destroy itself —
`srv_resolver.c`'s `dns_callback` falls back to starting new A/AAAA queries; destroy
invokes the callbacks after releasing the lock, so those `start_query()` calls run in
the middle of destroy, on the destroying thread.

## Details

The resolver now sets a `shutting_down` flag as the first statement of destroy's locked
snapshot, and `start_query()` checks it right after acquiring the group lock:

```c
if (resolver->shutting_down) {
    if (p_query)
        *p_query = NULL;
    pj_grp_lock_release(resolver->grp_lock);
    return PJ_EGONE;
}
```

Since both sides run under the group lock, every query either starts before the snapshot
(and is cancelled by it) or is rejected; no ordering arms a timer the teardown misses.

Writing `*p_query = NULL` is required: `srv_resolver.c` seeds `srv->q_aaaa` with a dummy
pointer and relies on `start_query()` overwriting it, and its completion logic
(`srv_completed = (srv->q_aaaa == NULL)`) would otherwise wait forever once a rejected
AAAA start follows an already-started A query. Both in-tree
callers (`srv_resolver.c`, `sip_resolve.c`) already handle a synchronous failure status
generically, so the new return introduces no leak or double report. `PJ_EGONE` follows
the existing convention for rejecting work on an object being destroyed
(`turn_sock.c`, `sip_reg.c`, `sip_inv.c`).

`on_read_complete()` gets the same check where it re-acquires the lock to update the
response cache: a response that came in while destroy is running must not allocate a new
cache entry, since destroy sweeps the cache and a late entry would never be freed.

Left out as pre-existing and separate: destroy's cache sweep does not honor
`cache->ref_cnt` against a concurrent cache-hit callback in `start_query()`, and
`pj_dns_resolver_add_entry()` / `pj_dns_resolver_handle_events()` are not gated by the
new flag.

## Testing

Adds `dns_start_during_destroy_test`, which starts a query from inside destroy's
`PJ_ECANCELLED` notification (the SRV-fallback pattern), seeds the out-pointer with a
dummy the way the SRV resolver does, and asserts the start is rejected with `PJ_EGONE`,
the out-pointer is cleared, and no timer fires past the teardown. It fails on stock
master (`start_query()` returns `PJ_SUCCESS` and arms the stray timer) and passes with
this change. `pjlib-util-test resolver_test` passes on macOS (arm64) and Linux (gcc 13),
0 failed, including this test and `dns_destroy_pending_test` from #5167.

### DIFF
--- a/pjlib-util/include/pjlib-util/resolver.h
+++ b/pjlib-util/include/pjlib-util/resolver.h
@@ -407,11 +407,14 @@ PJ_DECL(pj_status_t) pj_dns_resolver_destroy(pj_dns_resolver *resolver,
  *                  and which will be given back in the callback.
  * @param p_query   Optional pointer to receive the query object, if one
  *                  was started. If this pointer is specified, a NULL will
- *                  be returned if response cache is available immediately.
+ *                  be returned if response cache is available immediately,
+ *                  or if the query is rejected because the resolver is
+ *                  being destroyed.
  *
- * @return          PJ_SUCCESS if either an asynchronous query has been 
+ * @return          PJ_SUCCESS if either an asynchronous query has been
  *                  started successfully or response cache is available and
- *                  the user callback has been called.
+ *                  the user callback has been called; PJ_EGONE if the
+ *                  resolver is being destroyed.
  */
 PJ_DECL(pj_status_t) pj_dns_resolver_start_query(pj_dns_resolver *resolver,
                                                  const pj_str_t *name,
--- a/pjlib-util/src/pjlib-util-test/resolver_test.c
+++ b/pjlib-util/src/pjlib-util-test/resolver_test.c
@@ -71,6 +71,9 @@ static pj_sem_t *sem;
 static pj_dns_settings set;
 static volatile pj_bool_t destroy_done;
 static volatile pj_bool_t cb_after_destroy;
+static pj_dns_resolver *start_during_destroy_res;
+static pj_status_t start_during_destroy_status;
+static pj_dns_async_query *start_during_destroy_query;
 
 #define MAX_LABEL   32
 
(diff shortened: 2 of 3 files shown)


sauwming (approved):
Reviewed the diff and traced the race through `srv_resolver.c`/`sip_resolve.c` — the reentrant `start_query()` scenario described in the PR body checks out exactly against the current code (SRV-fallback `dns_callback()` calling `resolve_hostnames()` on any non-`PJ_SUCCESS` status, which includes the `PJ_ECANCELLED` destroy delivers).

Correctness:
- `shutting_down` is only read/written under `resolver->grp_lock`, set as the first statement of `destroy()`'s snapshot — no ordering lets a query slip past that arms a timer the teardown misses.
- `*p_query = NULL` on rejection is correctly required for `srv_resolver.c`'s `q_aaaa` dummy-pointer/`srv_completed` logic.
- Both in-tree callers already handle non-`PJ_SUCCESS` generically, so the new `PJ_EGONE` introduces no leak/double-report.
- `PJ_EGONE` follows the existing "reject work on an object being destroyed" convention (`turn_sock.c`, `sip_reg.c`, `sip_inv.c`).
- The `on_read_complete()` cache-update guard closes the matching late-response window (destroy's cache-destroy sweep runs unlocked, so this also narrows exposure to that pre-existing gap for the paths this PR touches).

Test coverage looks good: `dns_start_during_destroy_test` reproduces the exact SRV-fallback reentrancy pattern (same `0x1` sentinel, nested `start_query()` from inside the `PJ_ECANCELLED` callback) and checks rejection status, cleared out-pointer, and no stray timer firing past teardown.

No blocking issues. Minor non-blocking notes: the sleep-based "no timer fired" assertion is inherently a little racy under slow CI, but it matches the pattern already accepted in `dns_destroy_pending_test` from #5167; the pre-existing gaps called out in the PR body (unlocked cache sweep vs. `ref_cnt`, `add_entry()`/`handle_events()` not gated) are reasonable to leave as follow-ups.

nanangizz (approved):

Issue #1, opened 2016-11-22
Is this repo a pure mirror? do you accept github pull request?


jiulongw:
Thanks for setting up pjproject mirror in github!  Do you have plan to accept contribution here in github?

ismangil:
It is just a pure mirror. Please submit to public mailing list.

PR #5183, opened 2026-08-12
pjmedia: send unconditional start-of-stream keep-alive for text streams

trengginas:
An idle text stream never emits an RTP packet until the user types, since `send_text_locked()` returns early on the idle check. Unlike audio, which punches a NAT hole via its initial VAD-suspended frames, a text stream behind NAT therefore has no RTP binding at call setup and cannot receive the peer's text until it has sent some itself.

This adds a start-of-stream keep-alive burst for text streams, driven from the stream clock alongside the periodic RTCP check:

- Sends a few empty T140 RTP packets (plus RTCP) spaced `PJMEDIA_STREAM_START_KA_INTERVAL_MSEC` apart, up to `PJMEDIA_STREAM_START_KA_CNT`, so the hole punch survives a lost packet.
- Runs regardless of `PJMEDIA_STREAM_ENABLE_KA` and of media direction (including `inactive`); when `PJMEDIA_STREAM_ENABLE_KA` is enabled, the stream's runtime `use_ka`/`ka_cfg` apply instead.
- Skipped when ICE is active, as connectivity checks already create the bindings.
- Kept out of `send_text_locked()` so the first-packet (BOM) logic is untouched.

### DIFF
--- a/pjmedia/src/pjmedia/txt_stream.c
+++ b/pjmedia/src/pjmedia/txt_stream.c
@@ -477,11 +477,11 @@ PJ_DEF(pj_status_t) pjmedia_txt_stream_start(pjmedia_txt_stream *stream)
     pj_status_t status;
 
     pjmedia_stream_common_start(c_strm);
-    if (!c_strm->enc->paused || !c_strm->dec->paused) {
-        status = pjmedia_clock_start(stream->clock);
-    } else {
-        status = pjmedia_clock_stop(stream->clock);
-    }
+
+    /* Run the clock even for inactive direction, RTCP and start-of-stream
+     * keep-alives are still needed (RFC 3264 5.1, RFC 6263 1).
+     */
+    status = pjmedia_clock_start(stream->clock);
 
     return status;
 }
--- a/pjmedia/src/pjmedia/txt_stream.c
+++ b/pjmedia/src/pjmedia/txt_stream.c
@@ -100,10 +100,10 @@ typedef struct pjmedia_txt_stream {
     pj_timestamp rtcp_last_tx;   /**< Last RTCP tx time.     */
     pj_timestamp tx_last_ts;     /**< Timestamp of last tx.  */
 
-    unsigned start_ka_cnt;         /**< Remaining start-of-stream
+    unsigned start_ka_left;        /**< Remaining start-of-stream
                                         keep-alives to send.      */
-    unsigned start_ka_interval;    /**< Start keep-alive interval
-                                        in msec.                  */
+    unsigned start_ka_msec;        /**< Effective start keep-alive
+                                        interval, in msec.        */
     pj_timestamp start_ka_last_tx; /**< Last start keep-alive tx. */
     red_buf tx_buf[NUM_BUFFERS]; /**< Tx buffer.         */
     int tx_nred;                 /**< Num of redundant data. */
@@ -124,28 +124,28 @@ static void clock_cb(const pj_timestamp *ts, void *user_data);
  * Set up the start-of-stream keep-alive burst, see check_start_ka().
  * Must be called after the transport is attached.
  */
-static void init_start_ka(pjmedia_txt_stream *stream,
-                          const pjmedia_txt_stream_info *info)
+static void init_start_ka(pjmedia_txt_stream *stream)
 {
     pjmedia_stream_common *c_strm = &stream->base;
     pjmedia_transport_info tpinfo;
     pjmedia_ice_transport_info *ice_info;
 
 #if defined(PJMEDIA_STREAM_ENABLE_KA) && PJMEDIA_STREAM_ENABLE_KA != 0
-    stream->start_ka_cnt = info->use_ka ? info->ka_cfg.start_count : 0;
-    stream->start_ka_interval = info->ka_cfg.start_interval;
+    /* Take the runtime settings, already set from info->ka_cfg above. */
+    stream->start_ka_left = c_strm->use_ka ? c_strm->start_ka_count : 0;
+    stream->start_ka_msec = c_strm->start_ka_interval;
 #else
-    PJ_UNUSED_ARG(info);
-    stream->start_ka_cnt = PJMEDIA_STREAM_START_KA_CNT;
-    stream->start_ka_interval = PJMEDIA_STREAM_START_KA_INTERVAL_MSEC;
+    /* No runtime settings in this build, they are compiled out. */
+    stream->start_ka_left = PJMEDIA_STREAM_START_KA_CNT;
+    stream->start_ka_msec = PJMEDIA_STREAM_START_KA_INTERVAL_MSEC;
 #endif
 
-    if (stream->start_ka_cnt > START_KA_MAX_CNT)
-        stream->start_ka_cnt = START_KA_MAX_CNT;
-    if (stream->start_ka_interval < START_KA_MIN_INTERVAL_MSEC)
-        stream->start_ka_interval = START_KA_MIN_INTERVAL_MSEC;
+    if (stream->start_ka_left > START_KA_MAX_CNT)
+        stream->start_ka_left = START_KA_MAX_CNT;
+    if (stream->start_ka_msec < START_KA_MIN_INTERVAL_MSEC)
+        stream->start_ka_msec = START_KA_MIN_INTERVAL_MSEC;
 
-    if (stream->start_ka_cnt == 0)
+    if (stream->start_ka_left == 0)
         return;
 
     /* ICE connectivity checks already create the NAT bindings. */
@@ -156,7 +156,7 @@ static void init_start_ka(pjmedia_txt_stream *stream,
                    pjmedia_transport_info_get_spc_info(
                        &tpinfo, PJMEDIA_TRANSPORT_TYPE_ICE);
         if (ice_info && ice_info->active)
-            stream->start_ka_cnt = 0;
+            stream->start_ka_left = 0;
     }
 }
 
@@ -450,7 +450,7 @@ pjmedia_txt_stream_create(pjmedia_endpt *endpt, pj_pool_t *pool,
         pjmedia_stream_common_send_rtcp_sdes(c_strm);
     }
 
-    init_start_ka(stream, info);
+    init_start_ka(stream);
 
 #if defined(PJMEDIA_STREAM_ENABLE_KA) && PJMEDIA_STREAM_ENABLE_KA != 0
     /* NAT hole punching by sending KA packet via RTP transport. */
@@ -1040,7 +1040,7 @@ static void check_tx_rtcp(pjmedia_txt_stream *stream)
 /*
  * check_start_ka()
  * Punch NAT holes at stream start by sending a few empty RTP packets (plus
- * RTCP) spaced start_ka_interval apart, so the bindings exist before any
+ * RTCP) spaced start_ka_msec apart, so the bindings exist before any
  * text is typed and survive a lost packet. Kept out of send_text_locked()
  * so it doesn't disturb the first-packet (BOM) logic there.
  */
@@ -1054,13 +1054,13 @@ static void check_start_ka(pjmedia_txt_stream *stream)
     pj_timestamp now;
     pj_status_t status;
 
-    if (stream->start_ka_cnt == 0)
+    if (stream->start_ka_left == 0)
         return;
 
     pj_get_timestamp(&now);
     if (stream->start_ka_last_tx.u64 != 0 &&
         pj_elapsed_msec(&stream->start_ka_last_tx, &now) <
-            stream->start_ka_interval)
+            stream->start_ka_msec)
     {
         return;
     }
@@ -1083,10 +1083,14 @@ static void check_start_ka(pjmedia_txt_stream *stream)
     if (status != PJ_SUCCESS)
         return;
 
-    send_rtcp(c_strm, !c_strm->rtcp_sdes_bye_disabled, PJ_FALSE, PJ_FALSE,
-              PJ_FALSE, PJ_FALSE, PJ_FALSE);
+    status = send_rtcp(c_strm, !c_strm->rtcp_sdes_bye_disabled, PJ_FALSE,
+                       PJ_FALSE, PJ_FALSE, PJ_FALSE, PJ_FALSE);
+    if (status != PJ_SUCCESS) {
+        PJ_PERROR(4, (c_strm->port.info.name.ptr, status,
+                      "Error sending start keep-alive RTCP"));
+    }
 
-    stream->start_ka_cnt--;
+    stream->start_ka_left--;
 }
 
 /* Clock callback */
--- a/pjmedia/src/pjmedia/txt_stream.c
+++ b/pjmedia/src/pjmedia/txt_stream.c
@@ -91,6 +91,9 @@ typedef struct pjmedia_txt_stream {
     pjmedia_clock *clock; /**< Clock.                 */
     unsigned buf_time;    /**< Buffering time.        */
     pj_bool_t is_idle;    /**< Is idle?               */
+    unsigned start_ka_left; /**< Remaining number of automatic
+                                 start-of-stream keep-alives to
+                                 send (for NAT hole punching).  */
 
     pj_timestamp rtcp_last_tx;   /**< Last RTCP tx time.     */
     pj_timestamp tx_last_ts;     /**< Timestamp of last tx.  */
@@ -241,6 +244,13 @@ pjmedia_txt_stream_create(pjmedia_endpt *endpt, pj_pool_t *pool,
     c_strm->rtcp_fb_nack.pid = -1;
     stream->rx_last_seq = -1;
     stream->is_idle = PJ_TRUE;
+
+    /* Unconditionally (regardless of PJMEDIA_STREAM_ENABLE_KA) send a
+     * short burst of empty packets right after stream start, to help with
+     * NAT hole punching even if the user hasn't typed anything yet. This
+     * reuses the existing start-keep-alive count/interval settings.
+     */
+    stream->start_ka_left = PJMEDIA_STREAM_START_KA_CNT;
 
 #if defined(PJMEDIA_STREAM_ENABLE_KA) && PJMEDIA_STREAM_ENABLE_KA != 0
     c_strm->use_ka = info->use_ka;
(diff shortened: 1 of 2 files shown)


nanangizz (commented):
Thanks for tackling this — the underlying gap is real: an idle text stream currently never emits a first RTP packet, because `send_text_locked()` returns early at the idle check before ever reaching the send path. Sending empty-but-valid RTP is also the right packet format for both goals here (address latching and NAT binding creation); a non-RTP keepalive would fail latching outright, since the receiver's switch decision in `stream_imp_common.c:478-510` is driven by parsing the RTP SSRC. So the intent and the packet choice are both sound. My concerns are with where the trigger is hooked.

## Issues with the current approach

**1. The T.140 BOM is no longer sent (blocking).**

`is_first_packet` (line 822) gates two unrelated things: the new hole-punch trigger (line 851) and the BOM-prepend block (line 871). On any text call, the first `clock_cb` fires on an idle stream, takes the `is_first_packet` branch, sends an empty packet, and sets `tx_last_ts`. From then on `is_first_packet` is permanently false, so when the user types their first character the BOM block is skipped and `EF BB BF` is never transmitted. Peers relying on the start-of-text marker lose it on every call. It's a quiet failure, since the receive path at line 837 strips the BOM anyway — nothing errors, the marker just stops going out.

The hole punch needs its own state rather than borrowing the flag that the transmission path depends on.

**2. This shouldn't be sent when ICE is in use.**

ICE connectivity checks already create the binding and perform address discovery, so the burst is pure redundant traffic on every ICE call. RFC 6263 §1 explicitly scopes ICE agents out of its keepalive mechanisms for this reason. Suggest gating on `pjmedia_transport_info_get_spc_info(&info, PJMEDIA_TRANSPORT_TYPE_ICE) == NULL`.

This also argues against removing the runtime gate entirely (see #4) — ICE is exactly the case where an application needs to turn this off.

**3. The `tx_last_ts` assignment is dead code, and the stated rationale doesn't hold.**

`stream->tx_last_ts = now;` already exists at line 955, inside the "Network Transmission" block, same `now`, same pass. The new assignment 12 lines later is a no-op. The commit message says `tx_last_ts` "was never updated after creation" — it isn't updated only because the idle check returns before the send path is reached, not because the assignment is missing. Worth correcting, since that premise is what led to hooking `is_first_packet`.

**4. No way to disable it.**

Reading `PJMEDIA_STREAM_START_KA_CNT` / `..._INTERVAL_MSEC` directly means `info->ka_cfg` has no effect and there's no runtime control at all. Combined with #2, there needs to be some gate. I realize `ka_cfg` and `c_strm->start_ka_count` are both compiled out when `PJMEDIA_STREAM_ENABLE_KA == 0` (`stream_common.h:111` and `:330`), so reusing them for an unconditional path would mean lifting those out of the `#if` — a public header change. That may be the right call if this is meant to be always-on, but it should be a deliberate decision rather than a side effect of hardcoding the macros.

**5. Robustness of the macro values.**

- `START_KA_INTERVAL_MSEC == 0` makes `idle_msec >= 0` always true, so the remaining budget drains at the 300ms clock rate instead of 1s apart — which defeats the loss-resilience rationale, since the packets leave through the same binding almost simultaneously.
- `START_KA_INTERVAL_MSEC > 10000` lets the ordinary `idle_msec >= 10000` branch fire first and, because the decrement is shared, silently consume the start budget.
- `start_ka_left` is `unsigned`, so a negative or oversized `CNT` wraps into an effectively unbounded burst with no runtime off switch.

Some clamping at the point of use would help. (pjlib has no `PJ_STATIC_ASSERT`, so a compile-time check isn't available.)

**6. Off-by-one between comment and code.**

The comment says the first packet is always sent and "on top of that, up to `PJMEDIA_STREAM_START_KA_CNT` more empty packets". But the `is_first_packet` send also decrements `start_ka_left` (lines 854-855), so the total is `CNT` packets — 2 by default, not 3.

Unrelated to this PR, but noticed while checking: `config.h:1461` says `start_count = 0` "will disable it" while `stream_common.h:334-338` says it means "keepalive will be sent once for NAT hole punching". Those contradict each other and could use a follow-up.

## Suggested approach

Keep the hole punch out of `send_text_locked()` entirely — give it a dedicated counter and a dedicated timestamp, and drive it from `clock_cb` alongside `check_tx_rtcp()`, calling `send_keep_alive_packet()` directly. That resolves #1 structurally (no contention with `is_first_packet`), fits the existing pattern (`check_tx_rtcp()` already does exactly this against `rtcp_last_tx`), and needs no timer heap — which matters, since pjmedia has none.

Sending `send_keep_alive_packet()` rather than a bare RTP packet also punches the RTCP binding in the same step, which the current RTP-only path doesn't.

On **where the first send belongs**: the create-time hook at line 402-405 is already valid — it's after `pjmedia_transport_attach2()` (386) with `rem_addr` populated (352-359) — and would fire ~300ms earlier than the first clock tick. But I'd lean toward the **first clock callback** anyway: it's guaranteed to be after `pjmedia_txt_stream_start()`, which avoids sending before the transport is fully started, and sidesteps any race with DTLS-SRTP keys still being negotiated at create time. 300ms is negligible against call-setup timescales, and it keeps all the burst logic in one place instead of splitting the first packet from the rest. Worth confirming the DTLS timing concern is real before treating it as decisive, though.

andreas-wehrmann:
I'm also currently playing around with RTT; from what I can tell the text stream module doesn't really consider the negotiated media direction. I'm not entirely there yet, but an endpoint accepting a stream recvonly shouldn't transmit anything I would presume.

sauwming:
@nanangizz , please help continue this one

nanangizz:
@andreas-wehrmann Thanks, agreed on the direction handling. RFC 3264 §5.1 restricts `recvonly` to media, so a `recvonly` text stream should not be transmitting T.140 text, and the text stream currently doesn't check `enc->paused` on its send path the way the audio/video streams do. That's a separate fix, so I'd rather track it outside this PR.

The start-of-stream keep-alive here is intentionally direction-agnostic. RFC 3264 states that RTCP is still sent for sendonly, recvonly and inactive streams, and RFC 6263 §1 spells out the `recvonly` case: the receiving agent behind a NAT "will never receive RTP packets from the public peer if the NAT mapping has not been created", so an agent needs to send keepalive "regardless of whether the media stream is currently inactive, sendonly, recvonly, or sendrecv." The packet we send is an empty T140block, which RFC 4103 §5.2 already defines as what a T.140 sender emits when it has nothing to say, so it's valid media-plane traffic rather than a non-RTP hack. This matches what the audio stream does today, where its start keep-alives go out before the paused check.

The burst runs regardless of direction, including `a=inactive` (see the latest commit, which keeps the stream clock running in that case). It's only skipped when ICE is active, which RFC 6263 explicitly scopes out, or when the media is disabled (port zero), where no stream exists at all.

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -109,6 +120,46 @@ static void clock_cb(const pj_timestamp *ts, void *user_data);
...
+static void init_start_ka(pjmedia_txt_stream *stream,
+                          const pjmedia_txt_stream_info *info)
+{
+    pjmedia_stream_common *c_strm = &stream->base;
+    pjmedia_transport_info tpinfo;
+    pjmedia_ice_transport_info *ice_info;
+
+#if defined(PJMEDIA_STREAM_ENABLE_KA) && PJMEDIA_STREAM_ENABLE_KA != 0
+    stream->start_ka_cnt = info->use_ka ? info->ka_cfg.start_count : 0;
+    stream->start_ka_interval = info->ka_cfg.start_interval;
```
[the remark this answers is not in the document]
nanangizz: Intended. When `PJMEDIA_STREAM_ENABLE_KA` is compiled in, `use_ka` is the runtime switch for the non-codec keep-alive/NAT hole punch mechanism across audio, video and text, and pjsua defaults it to on in that build. Having the burst ignore it would leave text with no runtime off switch, which was one of the review points. The always-on path is for the default build, where no runtime knob exists. I've reworded the config.h note to make this explicit.

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -109,6 +120,46 @@ static void clock_cb(const pj_timestamp *ts, void *user_data);
...
+#else
+    PJ_UNUSED_ARG(info);
+    stream->start_ka_cnt = PJMEDIA_STREAM_START_KA_CNT;
+    stream->start_ka_interval = PJMEDIA_STREAM_START_KA_INTERVAL_MSEC;
+#endif
+
+    if (stream->start_ka_cnt > START_KA_MAX_CNT)
+        stream->start_ka_cnt = START_KA_MAX_CNT;
+    if (stream->start_ka_interval < START_KA_MIN_INTERVAL_MSEC)
+        stream->start_ka_interval = START_KA_MIN_INTERVAL_MSEC;
```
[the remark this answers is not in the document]
nanangizz: The clamps are deliberate (review point 5): a zero interval would drain the whole budget at the clock rate and a negative or oversized macro value wraps the unsigned count into an effectively unbounded burst, with no runtime off switch in the default build. Since audio/video consume these values from a different path with their own gating, I'd rather keep the guard local to the text stream. Documented the bounds in the `PJMEDIA_STREAM_START_KA_CNT` note in 058e42e30.

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -405,6 +456,8 @@ pjmedia_txt_stream_create(pjmedia_endpt *endpt, pj_pool_t *pool,
         send_keep_alive_packet(c_strm);
 #endif
 
+    init_start_ka(stream, info);
```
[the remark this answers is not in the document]
nanangizz: Fair point on the timing, fixed in 058e42e30: the create-time keep-alive now records `start_ka_last_tx`, so the first burst packet waits for `start_ka_interval`. The create-time send itself stays, since that is existing behavior for text under `PJMEDIA_STREAM_ENABLE_KA` and matches audio/video (one keep-alive at creation plus `start_count` more afterwards). Reviewer preference was for the burst to begin at the first clock callback rather than at create time, to be safely after `pjmedia_txt_stream_start()`.

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -982,6 +1035,58 @@ static void check_tx_rtcp(pjmedia_txt_stream *stream)
...
+            stream->start_ka_interval)
+    {
+        return;
+    }
+    stream->start_ka_last_tx = now;
+
+    /* The RTP session is shared with send_text_locked(). */
+    pj_mutex_lock(c_strm->jb_mutex);
+    status = pjmedia_rtp_encode_rtp(&channel->rtp, channel->pt, 0, 1, 0,
+                                    &rtphdr, &hdrlen);
```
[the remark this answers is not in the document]
nanangizz: This is inherent to T.140 over RTP without redundancy: the receiver cannot tell a lost empty T140block from a lost text block, so RFC 4103 has it mark any gap as missing text, and recommends redundancy for exactly this reason (pjsua defaults to RED level 2, where the RED decoder recovers the gap). The existing 10-second idle keep-alives in `send_text_locked()` have the same property. Reusing a sequence number would make the second packet a duplicate: it is dropped by the sequence check in `stream_imp_common.c` and would skew RTCP loss/jitter statistics, so I'd rather not.

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -982,6 +1037,58 @@ static void check_tx_rtcp(pjmedia_txt_stream *stream)
     }
 }
 
+/*
+ * check_start_ka()
+ * Punch NAT holes at stream start by sending a few empty RTP packets (plus
+ * RTCP) spaced start_ka_interval apart, so the bindings exist before any
+ * text is typed and survive a lost packet. Kept out of send_text_locked()
+ * so it doesn't disturb the first-packet (BOM) logic there.
```
sauwming, line 1045: check_start_ka() sends outbound RTP keep-alive packets without checking the negotiated media direction, so it fires even when the text stream is `recvonly` or `inactive`.

`pjmedia_txt_stream_start()` now unconditionally calls `pjmedia_clock_start()` instead of skipping it when both `enc->paused` and `dec->paused` are true (set when `c_strm->dir` excludes ENCODING/DECODING — e.g. SDP `inactive` on hold, or `recvonly`). `init_start_ka()`/`check_start_ka()` never consult `c_strm->dir` or `enc->paused`, so a text stream negotiated recvonly or put on hold still transmits several real RTP packets to the peer, violating the RFC 3264 direction contract, and keeps a clock thread alive for the whole hold duration just to run this burst.

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -982,6 +1037,58 @@ static void check_tx_rtcp(pjmedia_txt_stream *stream)
...
+    pj_get_timestamp(&now);
+    if (stream->start_ka_last_tx.u64 != 0 &&
+        pj_elapsed_msec(&stream->start_ka_last_tx, &now) <
+            stream->start_ka_interval)
+    {
+        return;
+    }
+    stream->start_ka_last_tx = now;
+
+    /* The RTP session is shared with send_text_locked(). */
```
sauwming, line 1069: The start-of-stream keep-alive shares the same RTP sequence-number space as real text, so a dropped keep-alive packet is indistinguishable from lost text on the receiver.

check_start_ka() calls `pjmedia_rtp_encode_rtp()` on the same session used by `send_text_locked()`, incrementing `ses->out_extseq` for every KA packet. If one of the start keep-alive packets is lost in transit (realistic for NAT-punch traffic), the peers `decode_red()`/`on_stream_rx_rtp()` sees a sequence gap and injects a U+FFFD replacement character into the jitter buffer, delivered to the application as corrupted text even though nothing typed was actually lost. With RED redundancy in use, the KA packet sits outside the RED history buffer so this gap can never be recovered.

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -982,6 +1037,58 @@ static void check_tx_rtcp(pjmedia_txt_stream *stream)
...
+        pj_memcpy(&hdr, rtphdr, sizeof(hdr));
+    pj_mutex_unlock(c_strm->jb_mutex);
+    if (status != PJ_SUCCESS)
+        return;
+
+    TRC_((c_strm->port.info.name.ptr, "Sending start keep-alive"));
+
+    /* On failure (e.g. SRTP keys not ready yet), retry at next interval. */
+    status = pjmedia_transport_send_rtp(c_strm->transport, &hdr, sizeof(hdr));
+    if (status != PJ_SUCCESS)
```
sauwming, line 1083: check_start_ka() discards the pj_status_t returned by send_rtcp(), unlike the sibling check_tx_rtcp() in the same file which checks it before updating state.

The return value of `send_rtcp()` during the start-of-stream burst is never captured, so a transport failure is silently swallowed with no log or error propagation, whereas `check_tx_rtcp()` a few lines above does `status = send_rtcp(...); if (status == PJ_SUCCESS) ...`. This is the same class of defect the immediately preceding commit on this branch ("Fix Coverity CHECKED_RETURN defects") was fixing.

pjmedia/src/pjmedia/txt_stream.c
```diff
@@ -109,6 +120,46 @@ static void clock_cb(const pj_timestamp *ts, void *user_data);
 
 #include "stream_imp_common.c"
 
+/*
+ * Set up the start-of-stream keep-alive burst, see check_start_ka().
+ * Must be called after the transport is attached.
```
sauwming, line 125: New fields start_ka_cnt/start_ka_interval/start_ka_last_tx duplicate the existing pjmedia_stream_common fields start_ka_count/start_ka_interval/last_start_ka_tx that this file already populates from info->ka_cfg.

When PJMEDIA_STREAM_ENABLE_KA is enabled, init_start_ka() re-derives the same values from info->use_ka/info->ka_cfg that were already used to set c_strm->start_ka_count/start_ka_interval earlier in pjmedia_txt_stream_create(). The two copies can drift if a future change updates only one path, and readers must track two parallel variables for one concept.

pjmedia/src/pjmedia/txt_stream.c
nanangizz: Two separate cases here.

For `recvonly`, the clock behavior is unchanged by this PR: only `enc->paused` is set, so `!enc->paused || !dec->paused` was already true and the clock already ran. What changed is the both-paused case, i.e. direction `inactive` (or `PJMEDIA_DIR_NONE`).

Sending the burst regardless of direction is deliberate. RFC 6263 §1 requires keepalive "regardless of whether the media stream is currently inactive, sendonly, recvonly, or sendrecv", and its motivating example is exactly the `recvonly` agent behind a NAT that "will never receive RTP packets from the public peer if the NAT mapping has not been created". For `inactive` it also notes that letting the binding time out means "media won't be able to come off hold". RFC 3264 §5.1 adds that RTCP is still sent and received for sendonly, recvonly and inactive streams, so stopping the clock was also suppressing the periodic RTCP that direction is not supposed to affect. The packets themselves are empty T140blocks, which RFC 4103 §5.2 defines as what a T.140 sender emits during an idle period, so this isn't T.140 data being transmitted against the direction contract.

The genuine direction gap is a different one, also raised by @andreas-wehrmann above: `send_text_locked()` never checks `enc->paused`, so a `recvonly` or on-hold text stream would transmit real text if the app called `pjmedia_txt_stream_send_text()`. I left that out of this PR as a separate fix, but happy to fold it in here if you prefer.

pjmedia/src/pjmedia/txt_stream.c
nanangizz: I traced this through both paths and I don't think the burst should reuse a sequence number.

With RED in use (pjsua defaults to level 2), the gap is covered rather than unrecoverable. `encode_red()` writes a redundant header for each of the previous `level` sequence numbers, and for a slot with nothing in it that header carries `hist_len == 0`. On the peer, `decode_red()` maps it to `block_seq = seq - (level - i)`, sees `diff == 1`, so it skips the U+FFFD loop and does `put_frame()` with a zero-length block, and `call_cb()` drops zero-length frames before reaching the application. Nothing needs recovering because the lost block was empty by definition.

Without RED, yes, a lost burst packet can surface one U+FFFD before the first real text. That is inherent to plain T.140 over RTP: the receiver cannot distinguish a lost empty T140block from lost text, which is why RFC 4103 recommends redundancy. The existing 10-second idle keep-alives in `send_text_locked()` have exactly the same property today.

Reusing one sequence number would make things worse. `MIN_SEQUENTIAL` is 2 in `rtp.c`, so a duplicate arriving during source validation takes the `else` branch in `pjmedia_rtp_seq_update()`, gets flagged `bad` plus `dup`, is discarded by the rx path, and resets `probation`. Distinct sequence numbers do the opposite: the two empty packets absorb RFC 3550 probation, so the first real text packet lands on an already-validated source. Before this PR probation was consumed by the first two real text packets, and losing the first one made the second one get discarded as `bad`, which is actual text loss. So the burst is a net improvement on that front.

pjmedia/src/pjmedia/txt_stream.c
nanangizz: Agreed, fixed in fc6ecbd29. The status is now captured and logged with `PJ_PERROR`. The keep-alive budget is still consumed in that case, since the RTP hole punch has already succeeded at that point and resending it would only multiply the RTP packets.

pjmedia/src/pjmedia/txt_stream.c
nanangizz: Agreed on the drift risk, fixed in fc6ecbd29. `init_start_ka()` now reads `c_strm->use_ka`, `c_strm->start_ka_count` and `c_strm->start_ka_interval` instead of re-deriving them from `info`, so `info->ka_cfg` is consumed in one place only and the `info` argument is gone.

The local working fields stay, because `pjmedia_stream_common`'s keep-alive block is compiled out when `PJMEDIA_STREAM_ENABLE_KA == 0`, which is the default and the case this feature mainly exists for. Lifting those fields out of the `#if` would be a public header change, and it seemed better to keep that a deliberate decision rather than a side effect of this fix. To make the two levels distinguishable I renamed them to `start_ka_left` and `start_ka_msec`, i.e. the remaining count and the effective clamped interval, rather than shadowing the common names.

sauwming (approved):

PR #5184, opened 2026-08-12
swig: catch C++ exceptions in the Java binding instead of aborting the JVM

vlaushkin:
A pjsua2 method or constructor that is not declared `PJSUA2_THROW(Error)` but throws a
C++ exception aborts the whole JVM instead of raising a catchable Java exception.

## Problem

`PJSUA2_THROW(x)` expands to `throw(x)` only under `#ifdef SWIG` (config.hpp), and the
Java `%typemap(throws) pj::Error` fires only for methods carrying that specification. Any
pjsua2 method or **constructor** without it that throws — e.g. the `Endpoint`
constructor, which raises `PJ_EEXISTS` on double construction (endpoint.cpp:666) — gets
no wrapper try/catch. The C++ exception unwinds across the JNI boundary, finds no C++
handler in the JVM frames, and `std::terminate()` aborts the process (`SIGABRT`). This is
uncatchable and unrecoverable from Java.

Only the `PJSUA2_THROW`-annotated functions are protected today (about 320 of the ~3000
generated JNI wrappers); every constructor and the rest are not. The Python target has a
`%feature("director:except")` safety net (pjsua2.i, `#ifdef SWIGPYTHON`); the Java target
has neither that nor a global `%exception`.

The uncaught-exception abort at the JNI boundary looks like:

```
first Endpoint created
libc++abi: terminating due to uncaught exception of type pj::Error   (exit 134)
```

## Details

The change adds, inside `#ifdef SWIGJAVA` only, a global `%exception` that translates a
C++ exception escaping any wrapper into a Java `RuntimeException`, and a
`%feature("director:except")` that clears (after logging) a Java exception thrown inside
a director callback instead of letting it unwind into pjsip C code.

It composes cleanly with the existing `throws` typemap. For a `PJSUA2_THROW`-declared
method the generated wrapper nests the typemap's catch **inside** `$action`, so it still
fires first and is unchanged; the new outer handler is never reached for those:

```c
/* Endpoint_libCreate (declared) — unchanged behavior */
try {                                   // new outer %exception
  try {                                 // existing throws typemap
    (arg1)->libCreate();
  } catch(pj::Error &_e) {              // -> java.lang.Exception (checked), as before
    ...; return ;
  }
} catch (Swig::DirectorException &e){ ... }
  catch (pj::Error &e){ ... }           // never reached for declared methods
```

```c
/* new_Endpoint (undeclared ctor) — previously unprotected */
try {
  result = (pj::Endpoint *)new SwigDirector_Endpoint(jenv);
} catch (pj::Error &e) {
  ... ThrowNew(RuntimeException, e.info(true)); return 0;   // was: unwind -> abort
}
```

Declared methods keep their `throws java.lang.Exception` signature; the new net on
undeclared methods raises an **unchecked** `RuntimeException`, so no existing Java
signature changes and no caller must be updated. All non-Java targets are unaffected
(the whole block is under `#ifdef SWIGJAVA`).

## Design notes

The `director:except` behavior is a deliberate choice; happy to adjust to your preference.

- This mirrors the existing Python handler: describe + clear
  + continue (Python does `PyErr_Print` with `Py_Exit(1)` commented out). Only three
  director callbacks are non-void — `Endpoint::onCredAuth` (`pj_status_t`),
  `Call::onCallRedirected` (`pjsip_redirect_op`), `FindBuddyMatch::match` (`bool`). For
  those, a throwing override makes the JNI upcall return 0: `PJSIP_REDIRECT_REJECT` for
  the redirect (the C++ default is `STOP`, so it tries the next target instead of
  stopping), `false` for the match (no match), and `PJ_SUCCESS` for `onCredAuth` — the
  last sends an empty digest, so authentication fails rather than the library recomputing
  it. All are contained, never a crash. If you prefer, `onCredAuth` could special-case
  `PJ_ENOTSUP` (library computes the digest) on a throwing override, at the cost of
  diverging from strict Python parity. `%catches(Swig::DirectorException)`
  is not sufficient here, since most pjsua2 callbacks fire from pjsip's own C threads with
  no Java frame to catch at.

## Testing

`cd pjsip-apps/src/swig && make java` builds cleanly (SWIG 4.3.1, g++, JDK 21): wrapper
generation, JNI library, and `javac` of the package and samples all succeed.

The generated wrapper shows the fix: `new_Endpoint` now has the try/catch, while the
declared `Endpoint_libCreate` keeps its inner `pj::Error -> java.lang.Exception` typemap
unchanged (the new outer handler nests around it).

### DIFF
--- a/pjsip-apps/src/swig/pjsua2.i
+++ b/pjsip-apps/src/swig/pjsua2.i
@@ -14,6 +14,43 @@ using namespace std;
 using namespace pj;
 %}
 
+#ifdef SWIGJAVA
+// Keep C++ exceptions from crossing the JNI boundary, which aborts the process
+// (there is no C++ unwind handler in the intervening JVM frames). Exceptions
+// from methods that are not declared PJSUA2_THROW, including constructors,
+// become a Java RuntimeException; a Java exception thrown inside a director
+// callback is logged and cleared instead of unwinding into pjsip C code.
+%exception {
+  try {
+    $action
+  } catch (Swig::DirectorException &e) {
+    e.throwException(jenv);
+    return $null;
+  } catch (pj::Error &e) {
+    jclass excep = jenv->FindClass("java/lang/RuntimeException");
+    if (excep)
+      jenv->ThrowNew(excep, e.info(true).c_str());
+    return $null;
+  } catch (std::exception &e) {
+    jclass excep = jenv->FindClass("java/lang/RuntimeException");
+    if (excep)
+      jenv->ThrowNew(excep, e.what());
+    return $null;
+  } catch (...) {
+    jclass excep = jenv->FindClass("java/lang/RuntimeException");
+    if (excep)
+      jenv->ThrowNew(excep, "Unknown C++ exception");
+    return $null;
+  }
+}
+%feature("director:except") {
+  if (jenv->ExceptionCheck()) {
+    jenv->ExceptionDescribe();
+    jenv->ExceptionClear();
+  }
+}
+#endif
+
 //
 // STL stuff.
 //
--- a/pjsip-apps/src/swig/pjsua2.i
+++ b/pjsip-apps/src/swig/pjsua2.i
@@ -15,11 +15,21 @@ using namespace pj;
 %}
 
 #ifdef SWIGJAVA
+%{
+static void pjsua2_throw_java_runtime(JNIEnv *jenv, const char *msg)
+{
+  jclass excep = jenv->FindClass("java/lang/RuntimeException");
+  if (excep)
+    jenv->ThrowNew(excep, msg);
+}
+%}
 // Keep C++ exceptions from crossing the JNI boundary, which aborts the process
 // (there is no C++ unwind handler in the intervening JVM frames). Exceptions
 // from methods that are not declared PJSUA2_THROW, including constructors,
 // become a Java RuntimeException; a Java exception thrown inside a director
 // callback is logged and cleared instead of unwinding into pjsip C code.
+// Placed before the STL %template instantiations below so the handler also
+// wraps their generated wrappers (e.g. std::vector::at() out-of-range).
 %exception {
   try {
     $action
@@ -27,19 +37,13 @@ using namespace pj;
     e.throwException(jenv);
     return $null;
   } catch (pj::Error &e) {
-    jclass excep = jenv->FindClass("java/lang/RuntimeException");
-    if (excep)
-      jenv->ThrowNew(excep, e.info(true).c_str());
+    pjsua2_throw_java_runtime(jenv, e.info(true).c_str());
     return $null;
   } catch (std::exception &e) {
-    jclass excep = jenv->FindClass("java/lang/RuntimeException");
-    if (excep)
-      jenv->ThrowNew(excep, e.what());
+    pjsua2_throw_java_runtime(jenv, e.what());
     return $null;
   } catch (...) {
-    jclass excep = jenv->FindClass("java/lang/RuntimeException");
-    if (excep)
-      jenv->ThrowNew(excep, "Unknown C++ exception");
+    pjsua2_throw_java_runtime(jenv, "Unknown C++ exception");
     return $null;
   }
 }


pjsip-apps/src/swig/pjsua2.i
```diff
@@ -14,6 +14,43 @@ using namespace std;
 using namespace pj;
 %}
 
+#ifdef SWIGJAVA
+// Keep C++ exceptions from crossing the JNI boundary, which aborts the process
```
sauwming, line 26: Placement nit: this block sits before the `std_string.i`/`std_vector.i`/`std_map.i` includes and `%template()` instantiations, and there's already a separate `#ifdef SWIGJAVA` section ~90 lines down (with the `pj::Error` throws-typemap, `javabase`, `javacode`, etc.). Two consequences:

1. The global `%exception` also wraps the STL container wrappers instantiated below (`StringVector`, `IntVector`, ...) for Java. Probably fine/desirable (e.g. `std::vector::at()` throwing `std::out_of_range` would previously hit the same abort bug), but it's a wider blast radius than "pjsua2 methods" — worth confirming intentional.
2. Having two separate `#ifdef SWIGJAVA ... #endif` blocks fragments the Java-specific config. Consider moving this next to the other SWIGJAVA section for discoverability, unless there's an ordering reason it needs to be this early (e.g. needing to precede the STL templates).

sauwming (commented):
## Overview

Fixes a real bug: pjsua2 methods/constructors without `PJSUA2_THROW(Error)` (most of ~3000 generated JNI wrappers, including every constructor) currently abort the JVM if they throw a C++ exception, since nothing catches it at the JNI boundary. This adds a Java-only global `%exception` that converts any escaping C++ exception into an unchecked `RuntimeException`, plus a `director:except` handler that clears Java exceptions thrown inside director callbacks (mirrors the existing Python handler). Declared (`PJSUA2_THROW`) methods are unaffected — their throws-typemap still catches first, confirmed via `pjsua2/types.hpp` (`pj::Error` isn't a `std::exception`, so catch ordering is fine). Low-risk, well-scoped, and CI already compiles the Java target.

See inline notes, plus one open question worth explicit sign-off: `director:except` silently changes call-handling behavior (not just crash-safety) for three non-void director callbacks.

pjsip-apps/src/swig/pjsua2.i
```diff
@@ -14,6 +14,43 @@ using namespace std;
...
+// (there is no C++ unwind handler in the intervening JVM frames). Exceptions
+// from methods that are not declared PJSUA2_THROW, including constructors,
+// become a Java RuntimeException; a Java exception thrown inside a director
+// callback is logged and cleared instead of unwinding into pjsip C code.
+%exception {
+  try {
```
sauwming, line 34: The `FindClass("java/lang/RuntimeException")` + null-check + `ThrowNew` boilerplate is duplicated three times across the `pj::Error`, `std::exception`, and `catch (...)` branches. Could factor into a small helper, though this mirrors the existing style of the `pj::Error` throws-typemap elsewhere in this file, so it's consistent with local convention rather than a new problem — not a blocker.

pjsip-apps/src/swig/pjsua2.i
```diff
@@ -14,6 +14,43 @@ using namespace std;
...
+    return $null;
+  } catch (...) {
+    jclass excep = jenv->FindClass("java/lang/RuntimeException");
+    if (excep)
+      jenv->ThrowNew(excep, "Unknown C++ exception");
+    return $null;
+  }
+}
+%feature("director:except") {
+  if (jenv->ExceptionCheck()) {
```
sauwming, line 51: Worth explicit maintainer sign-off: this changes runtime semantics, not just crash safety, for the three non-void director callbacks (`Endpoint::onCredAuth`, `Call::onCallRedirected`, `FindBuddyMatch::match`) per the PR description — a Java exception in an override now silently resolves to a default return (auth fails / redirect rejected / no match) logged only via `ExceptionDescribe()`. Matches the existing Python `director:except` handler's describe+clear+continue pattern, but since it's user-visible call-handling behavior it seems worth a maintainer decision rather than defaulting silently.

sauwming: LGTM

vlaushkin:
Thanks for the detailed review.

**Placement (:18)** — intentional, and it's the ordering reason you flagged. SWIG's `%exception` only applies to wrappers generated after it, so placing it before the `// STL stuff` `%template` instantiations is what extends the crash-safety to the container wrappers (`StringVector`, `IntVector`, `ByteVector`, `StringToStringMap`, and the later `SockOptVector`/… batch). `std::vector::at()` out-of-range or `std::bad_alloc` from those would previously hit the same JVM abort. Moving the block down to the other `#ifdef SWIGJAVA` section (which sits after the first `%template` batch) would drop coverage for those early containers, so I added a comment noting the ordering dependency so it isn't moved by mistake. The wider blast radius is deliberate.

**`std::exception` branch** — kept intentionally: it's the C++ STL that pjsua2 exposes to Java (the container wrappers above, `bad_alloc`) that throws these, not pjsip's C core. Even without the branch `catch (...)` stays crash-safe, but the branch yields a real message (`e.what()`) instead of "Unknown C++ exception". And since `pj::Error` isn't a `std::exception` (as you confirmed), the two branches are disjoint and both needed.

**Helper (:24)** — you're right, factored the three `FindClass`/`ThrowNew` branches into a small static `pjsua2_throw_java_runtime()` helper (`8e7f5442e`). I'd kept it inline to follow this file's existing convention (the `pj::Error` throws-typemap does the same) — I try to stay consistent with the project's local style — but the helper genuinely reads cleaner here, so it was a good call. Verified SWIG regenerates and the Java wrapper compiles clean.

**`director:except` (:47)** — agreed it's a user-visible semantics change for the three non-void callbacks; it mirrors the existing Python `director:except` (describe + clear + continue), falling back to the default return rather than unwinding a Java exception into pjsip C code. Thanks for the sign-off.

sauwming (approved):

nanangizz (approved):

PR #5185, opened 2026-08-13
sip_msg: auto-add 'lr' param instead of asserting when 'hide' is set without it

sauwming:
## Description

`pjsip_routing_hdr_print()` in `sip_msg.c` asserted that a Route/Record-Route URI's proprietary `hide` param is always paired with `lr`:

```c
if (pj_stricmp(&p->name, &st_hide) == 0) {
    /* Check if param 'hide' is specified without 'lr'. */
    pj_assert(sip_uri->lr_param != 0);
    return 0;
}
```

Since `hide` can appear in an externally supplied/malformed Route header (e.g. `Route: <sip:host;hide>`), this assertion is reachable with external input. `pj_assert()` aborts the process in debug builds, and is a no-op in release builds (i.e. provides no real protection there either).

This follows up on #4991, which fixed the crash by simply deleting the assertion. Per [review feedback on that PR](https://github.com/pjsip/pjproject/pull/4991#issuecomment-4539562352), rather than dropping the check, this PR normalizes `'hide'` to imply `'lr'`.

**Where the normalization actually belongs.** An earlier revision of this PR did the normalization inside `pjsip_routing_hdr_print()` itself. [Review feedback here](https://github.com/pjsip/pjproject/pull/5185#pullrequestreview-4923772118) pointed out that's ineffective: the very next statement suppresses the header (`return 0`), so mutating `lr_param` there changes nothing on the wire. Worse, by the time a message is ever printed, `pjsip_process_route_set()` has already run and, for a route without `lr`, already erased it as a strict route — rewriting the Request-URI to the very URI `hide` was meant to keep off the wire. A print function mutating the message is also worth avoiding on its own, since messages get printed more than once (re-encode after `pjsip_tx_data_invalidate_msg()`, retransmissions, auth retries).

The fix now lives in `pjsip_process_route_set()` (`sip_util.c`), where `has_lr_param` is computed and is the only place the value is actually acted upon:

```c
has_lr_param = url->lr_param;

if (!has_lr_param) {
    const pj_str_t st_hide = {"hide", 4};
    const pjsip_param *prm = url->other_param.next;

    for (; prm != &url->other_param; prm = prm->next) {
        if (pj_stricmp(&prm->name, &st_hide) == 0) {
            has_lr_param = PJ_TRUE;
            break;
        }
    }
}
```

`sip_msg.c` no longer mutates the URI; it just logs the anomaly (at level 5, since this is reachable from a remote-supplied route set and doesn't dedupe — each message has its own URI object):

```c
if (pj_stricmp(&p->name, &st_hide) == 0) {
    if (sip_uri->lr_param == 0) {
        PJ_LOG(5, ("sip_msg", "Route/Record-Route URI has 'hide' "
                              "param without 'lr'"));
    }
    return 0;
}
```

Behavior is unchanged for well-formed input (`hide` + `lr`). For malformed input (`hide` without `lr`): no crash, the route is still suppressed from the wire, and — the actual behavioral fix — it's now correctly treated as a loose route (Request-URI left alone, Route header kept) instead of a strict route.

`sip_util_proxy.c:191`, the other reader of `lr_param`, was also reviewed but left untouched since that whole code path is `#if 0`.

## Motivation and Context

Fixes the DoS vector described in #4991 (crash via `Route: <sip:host;hide>`), and — per review on this PR — fixes the actual routing-decision bug rather than a print-time side effect that never took effect.

## How Has This Been Tested?

- Full `pjsip` build completes with zero errors/warnings.
- Added `route_hide_lr_test()` to `pjsip/src/test/msg_test.c` (real CI coverage; `INCLUDE_MSG_TEST` only excludes Apple+ASan, so this runs under normal CI). It builds a request via `pjsip_endpt_create_request()`, adds a parsed `Route: <sip:proxy.example.com;hide>` header, calls `pjsip_process_route_set()` first (matching real message-flow order: routing decisions happen before transmission-time printing), and checks:
  - the Route header is kept (not erased as a strict route),
  - the Request-URI is left untouched (not rewritten to the hidden route),
  - printing the header afterwards doesn't crash.
- Verified locally (temporarily flipping the `APPLE_ASAN` test guard off) that the test actually catches both failure modes:
  - reverting to the original `pj_assert()` aborts the process on this test input,
  - reverting just the `sip_util.c` fix (keeping only the old print-function mutation) makes the test fail with "Request-URI was rewritten to the hidden route URI",
  - with both fixes in place, the test passes.

## Types of changes

## Checklist:
- [x] My code follows the CODING STYLE of this project.

### DIFF
--- a/pjsip/src/pjsip/sip_msg.c
+++ b/pjsip/src/pjsip/sip_msg.c
@@ -1800,8 +1800,18 @@ static int pjsip_routing_hdr_print( pjsip_routing_hdr *hdr,
             const pj_str_t st_hide = {"hide", 4};
 
             if (pj_stricmp(&p->name, &st_hide) == 0) {
-                /* Check if param 'hide' is specified without 'lr'. */
-                pj_assert(sip_uri->lr_param != 0);
+                /* By design, the proprietary 'hide' param is always paired
+                 * with 'lr' when PJSIP generates the route itself. However,
+                 * this URI may come from external/malformed input, so
+                 * rather than asserting (which is a no-op in release builds
+                 * and could be used as a DoS vector), just auto-add the
+                 * missing 'lr' and log it.
+                 */
+                if (sip_uri->lr_param == 0) {
+                    PJ_LOG(4, ("sip_msg", "Route/Record-Route URI has 'hide' "
+                                          "param without 'lr', adding 'lr'"));
+                    sip_uri->lr_param = 1;
+                }
                 return 0;
             }
             p = p->next;
--- a/pjsip/src/pjsip/sip_msg.c
+++ b/pjsip/src/pjsip/sip_msg.c
@@ -1801,16 +1801,16 @@ static int pjsip_routing_hdr_print( pjsip_routing_hdr *hdr,
 
             if (pj_stricmp(&p->name, &st_hide) == 0) {
                 /* By design, the proprietary 'hide' param is always paired
-                 * with 'lr' when PJSIP generates the route itself. However,
-                 * this URI may come from external/malformed input, so
-                 * rather than asserting (which is a no-op in release builds
-                 * and could be used as a DoS vector), just auto-add the
-                 * missing 'lr' and log it.
+                 * with 'lr' when PJSIP generates the route itself. This URI
+                 * may come from external/malformed input though, so rather
+                 * than asserting (a no-op in release builds, and reachable
+                 * with external input), just log it. This header is
+                 * suppressed either way; pjsip_process_route_set() is
+                 * where 'hide' without 'lr' actually gets normalized.
                  */
                 if (sip_uri->lr_param == 0) {
-                    PJ_LOG(4, ("sip_msg", "Route/Record-Route URI has 'hide' "
-                                          "param without 'lr', adding 'lr'"));
-                    sip_uri->lr_param = 1;
+                    PJ_LOG(5, ("sip_msg", "Route/Record-Route URI has 'hide' "
+                                          "param without 'lr'"));
                 }
                 return 0;
             }
--- a/pjsip/src/pjsip/sip_util.c
+++ b/pjsip/src/pjsip/sip_util.c
@@ -989,6 +989,26 @@ PJ_DEF(pj_status_t) pjsip_process_route_set(pjsip_tx_data *tdata,
             const pjsip_sip_uri *url = (const pjsip_sip_uri*)
                 pjsip_uri_get_uri((const void*)topmost_route_uri);
             has_lr_param = url->lr_param;
+
+            /* By design, PJSIP's proprietary 'hide' param is always
+             * paired with 'lr'. This URI may come from external/
+             * malformed input though, so don't rely on that invariant:
+             * treat 'hide' as implying 'lr' here too, otherwise a
+             * hidden route would incorrectly be treated as a strict
+             * route (Request-URI rewritten to a URI that was meant to
+             * stay hidden).
+             */
+            if (!has_lr_param) {
+                const pj_str_t st_hide = {"hide", 4};
+                const pjsip_param *prm = url->other_param.next;
+
+                for (; prm != &url->other_param; prm = prm->next) {
+                    if (pj_stricmp(&prm->name, &st_hide) == 0) {
+                        has_lr_param = PJ_TRUE;
+                        break;
+                    }
+                }
+            }
         } else {
             has_lr_param = 0;
         }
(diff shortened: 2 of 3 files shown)


nanangizz (commented):
Agreed the `;lr` normalization is right by design — my comments are only about where it happens. The crash fix itself is correct.

**The normalization can't take effect in the print function.** Two reasons:

1. The next statement is `return 0`, suppressing the whole header, so `lr_param = 1` changes nothing on the wire. The log text "adding 'lr'" reads as though it does.

2. `lr_param` drives exactly one behavioural decision — `pjsip_process_route_set()` at `sip_util.c:983-996` — and in the strict branch that function erases the topmost Route from the message before serialization:

```c
} else {
    new_request_uri = pjsip_uri_get_uri((pjsip_uri*)topmost_route_uri);
    pj_list_erase(first_route_hdr);
    tdata->saved_strict_route = first_route_hdr;
```

(`sip_util.c:1007-1011`). So a topmost `hide`-without-`lr` route is already gone by print time. What still reaches the print is either a topmost route that already has `lr`, or a non-topmost one — and only the topmost is ever consulted. Net: the URIs that reach the fix don't need it, and the one that needs it doesn't reach the fix.

**Where it belongs.** `pjsip_process_route_set()`, where `has_lr_param` is computed (`sip_util.c:986-994`) — that's the only place the value is acted on, and normalizing there directly gives the intended outcome (`hide` implies `lr`, hence loose routing, Request-URI not rewritten). Alternatively at entry, the way `force_lr` already does it in `pjsua_core.c:3719-3728`. Also worth a glance at `sip_util_proxy.c:191`, the other reader.

Separately, a print function mutating the message is worth avoiding regardless, since messages get printed more than once (re-encode after `pjsip_tx_data_invalidate_msg()`, retransmissions, auth retries).

**Minor:**
- `PJ_LOG(4, ...)` is reachable from a remote-supplied route set, and each message has its own URI object so it doesn't dedupe — one attacker-driven log line per message. Level 5 would be safer for a DoS-hardening change.
- `INCLUDE_MSG_TEST` is disabled only for Apple + ASan (`test.h:81`), so a regression case in `msg_test` would still run in normal CI — worth adding rather than a standalone harness.
- The `("sip_msg", ...)` literal is correct, matching `sip_msg.c:521`; this file has no `THIS_FILE`.

sauwming:
Thanks for the detailed review — addressed in 636ebe722:

- Moved the normalization out of `pjsip_routing_hdr_print()` into `pjsip_process_route_set()` (`sip_util.c:989-1008`), where `has_lr_param` is actually computed and acted on. `'hide'` now implies `'lr'` there directly, without mutating the shared URI object.
- `sip_msg.c` no longer mutates `lr_param`; it only logs the anomaly, dropped to level 5 per your point about remote-triggerable, non-deduped log spam.
- Added `route_hide_lr_test()` to `msg_test.c` instead of a standalone harness, since `INCLUDE_MSG_TEST` only excludes Apple+ASan. It builds a request via `pjsip_endpt_create_request()`, adds a parsed `Route: <sip:proxy.example.com;hide>` header, calls `pjsip_process_route_set()` first (matching real message-flow order) to check the Route header is kept and the Request-URI is untouched, then prints the header to confirm no crash.
- Verified locally (temporarily flipping `APPLE_ASAN` off) that: (1) the test fails with the pre-fix `pj_assert()` (process aborts), (2) it fails with `sip_uri->lr_param = 1` only in the print function reverted (Request-URI gets rewritten to the hidden route), and (3) it passes with both fixes in place.
- Left `sip_util_proxy.c:191` alone since that whole branch is `#if 0`.

nanangizz (commented):
LGTM.

A few nits on the test:

- `pool` is created and released but never used — every allocation goes through `tdata->pool`.
- The print check accepts any `len >= 0`, but a `hide` route should print exactly 0 bytes. As written, a regression in the suppression behaviour would still pass — suggest asserting `len == 0`.
- The Request-URI check compares only `len` bytes, so a truncated URI would pass; worth also checking `len` matches the expected length.
- Consider adding the well-formed `hide;lr` case — cheap, and it locks in that suppression is unchanged for correct input.

sauwming:
Addressed in 211e9d264:

- Removed the unused \`pool\` — everything already allocates through \`tdata->pool\`.
- The print check now asserts \`len == 0\` (fully suppressed), not just \`len >= 0\`.
- The Request-URI check now uses \`pj_strcmp2()\` against the exact printed length instead of a length-bounded \`strncmp\`, so a truncated/garbled URI can't slip through.
- Added the well-formed \`hide;lr\` case alongside the malformed one (now a small table-driven loop over both cases) to lock in that suppression is unchanged for correct input.

Re-verified locally (flipping \`APPLE_ASAN\` off temporarily): both cases pass with the fix in place, and reverting the \`sip_util.c\` normalization still correctly fails the malformed case with "Request-URI was rewritten to the hidden route URI".

nanangizz (approved):

sauwming:
Stress test failure will be addressed separately.

PR #5188, opened 2026-08-13
pjsua: fix answer direction and stream restart in apply_med_update()

andreas-wehrmann:
I'm currently working on an SRS (recorder) and wanted to make use of the [media_dir](https://github.com/andreas-wehrmann/pjproject/blob/5b0894b8a72a67e4931f5da20aacc7fcabc6ba21/pjsip/include/pjsua-lib/pjsua.h#L1230) setting in [pjsua_call_setting](https://github.com/andreas-wehrmann/pjproject/blob/5b0894b8a72a67e4931f5da20aacc7fcabc6ba21/pjsip/include/pjsua-lib/pjsua.h#L1166) so that media cannot "exceed" recvonly and to make it so streams offered sendrecv are accepted recvonly.

I told Claude to scout the code and do a test impl for me and it came back with two issues:
recvonly is never signalled to the other end and that pjsua would sometimes rebuild media even though there was no change.

I did confirm the issue with the media direction myself.
When setting all media_dir to PJMEDIA_DECODING and calling the SRS with one audio media with sendrecv,
the SDP answer will **NOT** say recvonly but with the fix it will.

Everything below this line was created by Claude Code

---

### 1. The media direction we ask for never reaches the other side

An application can answer a call receive-only by setting
`pjsua_call_setting.media_dir` to `PJMEDIA_DIR_DECODING`. pjsua does the right
thing internally — the stream is receive-only and sends no RTP — but the SDP it
puts in the 200 OK still says `a=sendrecv`.

The peer is therefore told it may expect media from us, and none ever arrives.
Only the signalling is wrong, which is why nothing looking at RTP would notice.

When several streams are set receive-only, only the first one is affected and
the rest come out correct. A call with a single audio stream loses it entirely.

### 2. The decision to rebuild a media stream is made on stale values

After each renegotiation pjsua compares the running stream against what was
just negotiated, and rebuilds the stream only if something actually changed.
The values it compares against are collected too early — before three of them
are finalised (RTCP-XR, RTCP multiplexing, and the media direction). The stream
itself is built from the final values, so the comparison is made against
numbers that were never used.

This goes wrong in both directions.

**It misses a rebuild it should do.** When an application changes a running
stream to receive-only, the comparison still sees the old direction and reports
the media as unchanged. The stream is left alone and keeps transmitting in the
old direction, even though pjsua now reports it as receive-only and — once the
first defect is fixed — signals it as receive-only as well.

**It forces a rebuild it should not do.** RTCP multiplexing is recorded from
whatever the peer advertised and then cleared again if the local account has
not enabled it, but the stale copy keeps the peer's value. So on any call where
the peer offers `a=rtcp-mux` and the account has not enabled multiplexing —
which is the default — the comparison always reports a change, and the audio
stream is torn down and rebuilt on every renegotiation: session timer refresh,
hold, unhold, ICE restart. Each rebuild means a new SSRC, a reset jitter buffer
and a gap in the audio. Nothing is logged as an error, so it just sounds like a
network glitch.

This half needs no special configuration, no `media_dir`, and affects the
offering side as well as the answering side.

### Seeing the second one

Two stock pjsua instances, no patch required:

    # Terminal 1 - answerer, default settings (this is the side that misbehaves)
    pjsua --null-audio --no-tcp --local-port 5090 --auto-answer 200 --app-log-level 4

    # Terminal 2 - caller, advertises a=rtcp-mux
    pjsua --null-audio --no-tcp --local-port 5080 --rtcp-mux --auto-answer 200 --app-log-level 4

In terminal 2, call `sip:127.0.0.1:5090`, wait a few seconds, then press `v` to
send a re-INVITE. Nothing in the offer changes. Terminal 1 destroys and recreates
the audio stream; it should log `stream #0 (audio) unchanged.` instead. Drop
`--rtcp-mux` and the same binary reports it unchanged, which shows multiplexing
is the trigger.

### When this started

The values were collected after the adjustments until #4300 reordered them. The
direction defect arrived with #2705 and moved into this function in that same
refactoring.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -4349,7 +4349,6 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
                 const pjmedia_sdp_session *local_sdp_renego =
                     pjmedia_sdp_session_clone(tmp_pool, local_sdp);
                 local_sdp = local_sdp_renego;
-                /* replace the local SDP with the new clone */
                 *local_sdp_ref = local_sdp_renego;
                 *need_renego_sdp = PJ_TRUE;
             }
@@ -4376,11 +4375,7 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
         }
     }
 
-    /* Take the copy of the stream info only now, i.e. after all the
-     * adjustments above (RTCP XR, RTCP mux, media direction) have been
-     * applied via 'si'. Copying it any earlier would hide those changes
-     * from check_srtp_roc() and is_media_changed() below.
-     */
+    /* Copy stream info only now, so is_media_changed() sees the adjustments above. */
     if (call_med->type == PJMEDIA_TYPE_AUDIO) {
         stream_info.info.aud = asi;
 #if defined(PJMEDIA_HAS_VIDEO) && (PJMEDIA_HAS_VIDEO != 0)
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -4216,7 +4216,7 @@ static pj_bool_t is_media_changed(const pjsua_call *call,
 
 /* Apply media update. */
 static pj_status_t apply_med_update(pjsua_call_media *call_med,
-                                    const pjmedia_sdp_session *local_sdp,
+                                    const pjmedia_sdp_session **local_sdp_ref,
                                     const pjmedia_sdp_session *remote_sdp,
                                     pj_bool_t *need_renego_sdp)
 {
@@ -4225,6 +4225,7 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
     const char *STR_RECVONLY = "recvonly";
     const char *STR_INACTIVE = "inactive";
 
+    const pjmedia_sdp_session *local_sdp = *local_sdp_ref;
     pjsua_call *call = call_med->call;
     pjsua_acc *acc = &pjsua_var.acc[call->acc_id];
     pjsua_call_id call_id = call->index;
@@ -4241,7 +4242,7 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
 #if defined(PJMEDIA_HAS_VIDEO) && (PJMEDIA_HAS_VIDEO != 0)
     pjmedia_vid_stream_info vsi = {0};
 #endif
-    pjmedia_txt_stream_info tsi;
+    pjmedia_txt_stream_info tsi = {0};
     pjmedia_stream_info_common *si = NULL;
     pjsua_stream_info stream_info = {0};
     pj_str_t *enc_name = NULL;
@@ -4251,7 +4252,6 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
         status = pjmedia_stream_info_from_sdp(
                                     &asi, tmp_pool, pjsua_var.med_endpt,
                                     local_sdp, remote_sdp, mi);
-        stream_info.info.aud = asi;
         enc_name = &asi.fmt.encoding_name;
 
 #if defined(PJMEDIA_HAS_VIDEO) && (PJMEDIA_HAS_VIDEO != 0)
@@ -4260,7 +4260,6 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
         status = pjmedia_vid_stream_info_from_sdp(
                                     &vsi, tmp_pool, pjsua_var.med_endpt,
                                     local_sdp, remote_sdp, mi);
-        stream_info.info.vid = vsi;
         enc_name = &vsi.codec_info.encoding_name;
 #endif
     } else if (call_med->type == PJMEDIA_TYPE_TEXT) {
@@ -4268,7 +4267,6 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
         status = pjmedia_txt_stream_info_from_sdp(
                                     &tsi, tmp_pool, pjsua_var.med_endpt,
                                     local_sdp, remote_sdp, mi);
-        stream_info.info.txt = tsi;
         enc_name = &tsi.fmt.encoding_name;
     }
     else {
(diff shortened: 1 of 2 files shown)


pjsip/src/pjsua-lib/pjsua_media.c
```diff
@@ -4348,10 +4346,11 @@ static pj_status_t apply_med_update(pjsua_call_media *call_med,
             pjmedia_sdp_media *m;
 
             if (!*need_renego_sdp) {
-                pjmedia_sdp_session *local_sdp_renego;
-                local_sdp_renego =
+                const pjmedia_sdp_session *local_sdp_renego =
                     pjmedia_sdp_session_clone(tmp_pool, local_sdp);
                 local_sdp = local_sdp_renego;
+                /* replace the local SDP with the new clone */
```
sauwming: Nit on comment volume — `CLAUDE.md` in this repo asks for minimal comments ("only if the code itself is not self explanatory", "as brief as possible ... avoid overbloating").

This one restates the line directly below it verbatim, so I'd just drop it.

Same thought for the 5-line block at line 4379 (`/* Take the copy of the stream info only now, ... */`) — the *why* is worth one line, but not a paragraph. Something like:

```c
/* Copy stream info only now, so is_media_changed() sees the adjustments above. */
```

Purely cosmetic, no functional impact.

pjsip/src/test/pjsua_call_test.c
```diff
@@ -1078,6 +1239,364 @@ static int test_rfc2543_hold_without_conn(void)
...
+
+    rc = establish_self_call(&caller, &callee, -1700);
+    if (rc != 0)
+        goto on_return;
+
+    /* Both legs are now running audio in ENCODING_DECODING. Arm the callee
+     * leg and poke it with an ordinary no-op re-INVITE from the caller.
+     */
+    g_ctx.narrow_dir_call_id = callee;
+    g_ctx.narrow_dir_armed = PJ_TRUE;
```
sauwming, line 1458: `g_ctx.narrow_dir_armed` stays `PJ_TRUE` from here all the way to `on_return` (line 1590), i.e. across `pjsua_call_answer_with_sdp()` **and** both `wait_until(&media_state_reached, ..., 8000)` calls.

If a *second* re-INVITE lands on the callee inside that window — lock-codec re-INVITE, session-timer refresh, ICE restart — `on_call_rx_reinvite()` fires again, sets `*async = PJ_TRUE` and returns, but nothing ever answers it. The leg then sits at `100 Trying`, `med_state_target` is never reached, and the failure surfaces as a misleading `-1715` ("re-INVITE media update did not complete") rather than as the real cause.

Not reachable with the current setup — the answer carries a single codec so `check_lock_codec()` declines, and the session timer is nowhere near expiry — so this is robustness rather than a live bug. Disarming right after the manual answer, or gating the callback on `!g_ctx.narrow_dir_reinvite_seen`, would close it.

sauwming (commented):
Reviewed the change and ran it locally (macOS/arm64, aarch64-apple-darwin, OpenSSL from MacPorts). The fix looks correct to me and the two new sub-tests are load-bearing.

**What I verified**

* Built the PR head clean with `-Wall` (no new warnings) and ran `pjsip-test pjsua_call_test` → `[OK]`.
* Reverted only `pjsua_media.c` back to `origin/master`, kept the new tests, and forced a relink → `pjsua_call_test` **fails** with `audio stream restarted on a no-op re-INVITE: caller created 1->2 destroyed 0->1, callee created 1->2 destroyed 0->1`. Temporary instrumentation in `is_media_changed()` prints `old_mux=0 new_mux=1` on that path, which is exactly the stale copy this PR removes. So the tests genuinely fail without the fix.
  * One build gotcha worth flagging for anyone repeating this: `pjsip/build` does **not** relink `pjsip-test` when only a library object changes, so a naive revert-and-`make` silently keeps the old binary and the tests appear to pass either way. `rm pjsip/bin/pjsip-test-*` before re-running.
* Compile-checked `pjsua_media.c` with `PJMEDIA_HAS_VIDEO 0` — the new `#if`-interleaved `if / else if` copy block is valid and warning-free in both configurations.
* Spot-checked for regressions on the PR head: `uri_test`, `msg_test`, `msg_err_test`, `multipart_test`, `txdata_test`, `dlg_core_test`, `tsx_bench`, `transport_loop_multi_test`, `resolve_test`, **`inv_offer_answer_test`**, `tsx_basic_test`, `tsx_uac_test` all `[OK]`.

**On the reasoning**

The `*local_sdp_ref` write-back preserves the function's implicit invariant: `*need_renego_sdp == PJ_TRUE` now always implies the caller's `local_sdp` already points at a `tmp_pool` clone (set either by the rtcp-fb / media-count block earlier in `pjsua_media_channel_update()` or by a previous `apply_med_update()` call). As a bonus that also removes the case where a second narrowed media line mutated the negotiator's *active local* SDP in place.

Moving the `stream_info` copy is likewise safe: `si` genuinely aliases `asi`/`vsi`/`tsi`, the only two consumers are `check_srtp_roc()` (reads addresses only) and `is_media_changed()` (the intended beneficiary), and `pjsua_{aud,vid,txt}_channel_update()` use `&asi`/`&vsi`/`&tsi` directly. The only `return` between the old and new copy sites is the `stream_info_from_sdp()` failure check, so no path leaves `stream_info` stale.

I also convinced myself the missing `remove_all_attr(m, STR_INACTIVE)` before adding the new direction attribute is not a gap: `stream_common.c` derives `si->dir` from the local m-line and yields `PJMEDIA_DIR_NONE` whenever `a=inactive` is present, so the `si->dir != PJMEDIA_DIR_NONE` guard makes a stale `a=inactive` unreachable.

**Scope note (so the fix doesn't get mis-described later)**

The description reads as though any app setting `pjsua_call_setting.media_dir` gets a `sendrecv` answer. In the ordinary path it doesn't: `pjsua_media_channel_create_sdp()` already passes `param.dir = call_med->def_dir` into `pjmedia_endpt_create_audio_sdp()`, and `sdp_neg.c`'s `update_media_direction()` preserves a `recvonly` local answer against a `sendrecv` offer. The narrowing in `apply_med_update()` — and hence defect #1 — only bites when the local SDP's direction doesn't already match `def_dir`, e.g. when the app supplies its own answer via `pjsua_call_answer_with_sdp()`, which is exactly (and only) what the new test exercises. Defect #2 (the rtcp-mux restart) is unconditional and needs no special configuration, so it's worth keeping the two clearly separated in the commit message.

Two minor comments inline, neither blocking.

pjsip/src/pjsua-lib/pjsua_media.c
andreas-wehrmann: The first comment was actually mine because I deemed it important to make it clear we're replacing what the _caller_ of the function provided, but okay; will drop it.
Agreement on the second one; dunno why Claude didn't pick up on it itself.

sauwming (approved):

nanangizz (approved):

PR #5190, opened 2026-08-14
pjmedia: increase RTP/RTCP receive buffer for DTLS-SRTP to avoid silent handshake truncation

average-enjoyer:
## Description
Adds a new `PJMEDIA_MAX_MRU_DTLS` config option (default 4000) and uses it for
`RTP_LEN`/`RTCP_LEN` in `transport_udp.c` whenever `PJMEDIA_SRTP_HAS_DTLS` is
enabled. Plain RTP/RTCP (no DTLS-SRTP) is unaffected — those paths still use
`PJMEDIA_MAX_MRU` (2000) and 600 respectively. `PJMEDIA_MAX_MRU` itself is
left untouched, since it's also used by `vid_stream.c` for unrelated video
RTP chunk-size calculations, and bumping it directly would have changed that
behavior as a side effect.

## Motivation and Context
DTLS-SRTP keying traffic (RFC 5764) is multiplexed onto the same RTP/RTCP UDP
sockets as media. Per RFC 6347 §4.2.3, a DTLS server is free to bundle
multiple handshake messages from the same flight into a single datagram:

> "multiple handshake messages may be placed in the same DTLS record,
> provided that there is room and that they are part of the same flight"

RFC 6347 §4.1.1 only recommends (SHOULD, not MUST) sizing records to fit the
path MTU, so implementations that skip PMTU probing will commonly emit the
full `ServerHello + Certificate + ServerKeyExchange + CertificateRequest +
ServerHelloDone` flight as one datagram, relying on IP fragmentation for
delivery.

Once a real certificate is in play, that flight routinely exceeds the
existing 2000-byte `PJMEDIA_MAX_MRU`. In one reproduction against a peer
server, the flight measured ~2505-2530 bytes (the certificate alone was
~2435 bytes PEM / ~1600+ bytes DER).

`pj_sock_recvfrom()` (`pjlib/src/pj/sock_bsd.c`) is a plain `recvfrom()`
wrapper with no `MSG_TRUNC`, so on Linux/BSD a datagram larger than the
supplied buffer is truncated by the kernel with **no indication anything was
lost**. The truncated bytes are handed straight to OpenSSL's `BIO_write()`
in `ssl_on_recv_packet()` (`transport_srtp_dtls.c`), corrupting the DTLS
record layer. Instead of failing loudly, both peers just keep retransmitting
their last flight — the handshake never completes and looks like a hang
rather than a clear error.

## How Has This Been Tested?
- `./configure && make -j3` — clean build, zero errors/warnings, with
  OpenSSL SRTP enabled (`PJMEDIA_SRTP_HAS_DTLS` active), exercising the
  changed branch.
- Compiled `pjsua` (`pjsip-apps/bin/pjsua-x86_64-pc-linux-gnu`) end-to-end
  against the change to confirm it links and builds cleanly through the
  full call stack, not just the `pjmedia` library in isolation.

## Screenshots (if appropriate):
N/A

## Types of changes

## Checklist:
- [ ] I have read the **CONTRIBUTING** document.

## Note for maintainers
This is a compile-time change: since `RTP_LEN`/`RTCP_LEN` size fixed struct
fields, any build with `PJMEDIA_SRTP_HAS_DTLS` enabled gets the larger
buffer on **every** RTP/RTCP UDP socket, not only transports that actually
negotiate DTLS-SRTP for a given call — `transport_udp.c` has no visibility
into what's layered above it. The memory cost is small per call (~1.4-3.4 KB
extra per media transport) but scales with concurrent call count, so flagging
it explicitly in case a runtime-configurable size is preferred instead.
`PJMEDIA_MAX_MRU_DTLS` can also be overridden via `config_site.h` like the
existing `PJMEDIA_MAX_MRU`, so deployments that want the old, smaller
footprint can opt back down.

### DIFF
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -499,24 +499,27 @@
 
 
 /**
- * Max packet size for receiving direction.
+ * Max packet size for receiving direction. This also bounds the RTP/RTCP
+ * UDP transport buffer and the STUN/TURN transport buffers used by ICE
+ * (see pjsua_media.c), since DTLS-SRTP keying traffic (RFC 5764) is
+ * multiplexed onto the same sockets as media.
+ *
+ * A DTLS-SRTP peer that reuses its full TLS server certificate (plus any
+ * intermediates) for the handshake, rather than a single self-signed
+ * leaf cert, can produce a ServerHello+Certificate+... flight larger than
+ * the default below. Since recvfrom() truncates an oversized datagram
+ * silently (see the truncation warning logged in transport_udp.c), such
+ * a peer's handshake would otherwise hang instead of failing loudly.
+ *
+ * Default: 4000 when DTLS-SRTP is enabled (#PJMEDIA_SRTP_HAS_DTLS),
+ *          2000 otherwise.
  */
 #ifndef PJMEDIA_MAX_MRU
-#  define PJMEDIA_MAX_MRU                       2000
-#endif
-
-
-/**
- * Max packet size for receiving direction when DTLS-SRTP keying is used
- * (see #PJMEDIA_SRTP_HAS_DTLS). A DTLS handshake flight carrying a
- * certificate can exceed #PJMEDIA_MAX_MRU; since recvfrom() truncates
- * oversized datagrams silently, that corrupts the handshake instead of
- * failing loudly.
- *
- * Default: 4000
- */
-#ifndef PJMEDIA_MAX_MRU_DTLS
-#  define PJMEDIA_MAX_MRU_DTLS                  4000
+#  if defined(PJMEDIA_SRTP_HAS_DTLS) && (PJMEDIA_SRTP_HAS_DTLS != 0)
+#    define PJMEDIA_MAX_MRU                     4000
+#  else
+#    define PJMEDIA_MAX_MRU                     2000
+#  endif
 #endif
 
 
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -501,8 +501,22 @@
 /**
  * Max packet size for receiving direction.
  */
-#ifndef PJMEDIA_MAX_MRU                 
+#ifndef PJMEDIA_MAX_MRU
 #  define PJMEDIA_MAX_MRU                       2000
+#endif
+
+
+/**
+ * Max packet size for receiving direction when DTLS-SRTP keying is used
+ * (see #PJMEDIA_SRTP_HAS_DTLS). A DTLS handshake flight carrying a
+ * certificate can exceed #PJMEDIA_MAX_MRU; since recvfrom() truncates
+ * oversized datagrams silently, that corrupts the handshake instead of
+ * failing loudly.
+ *
+ * Default: 4000
+ */
+#ifndef PJMEDIA_MAX_MRU_DTLS
+#  define PJMEDIA_MAX_MRU_DTLS                  4000
 #endif
 
 
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -499,31 +499,6 @@
 
 
 /**
- * Max packet size for receiving direction. This also bounds the RTP/RTCP
- * UDP transport buffer and the STUN/TURN transport buffers used by ICE
- * (see pjsua_media.c), since DTLS-SRTP keying traffic (RFC 5764) is
- * multiplexed onto the same sockets as media.
- *
- * A DTLS-SRTP peer that reuses its full TLS server certificate (plus any
- * intermediates) for the handshake, rather than a single self-signed
- * leaf cert, can produce a ServerHello+Certificate+... flight larger than
- * the default below. Since recvfrom() truncates an oversized datagram
- * silently (see the truncation warning logged in transport_udp.c), such
- * a peer's handshake would otherwise hang instead of failing loudly.
- *
- * Default: 4000 when DTLS-SRTP is enabled (#PJMEDIA_SRTP_HAS_DTLS),
- *          2000 otherwise.
- */
-#ifndef PJMEDIA_MAX_MRU
-#  if defined(PJMEDIA_SRTP_HAS_DTLS) && (PJMEDIA_SRTP_HAS_DTLS != 0)
-#    define PJMEDIA_MAX_MRU                     4000
-#  else
-#    define PJMEDIA_MAX_MRU                     2000
-#  endif
-#endif
-
-
-/**
  * DTMF/telephone-event duration, in timestamp. To specify the duration in
  * milliseconds, use the setting PJMEDIA_DTMF_DURATION_MSEC instead.
  *
@@ -1214,6 +1189,41 @@
  */
 #ifndef PJMEDIA_SRTP_HAS_DTLS
 #   define PJMEDIA_SRTP_HAS_DTLS                    0
+#endif
+
+
+/**
+ * Max packet size for receiving direction. This also bounds the RTP/RTCP
+ * UDP transport buffer and the STUN/TURN transport buffers used by ICE
+ * (see pjsua_media.c), since DTLS-SRTP keying traffic (RFC 5764) is
+ * multiplexed onto the same sockets as media. This block is placed after
+ * #PJMEDIA_SRTP_HAS_DTLS on purpose, so the dependency below is positional
+ * rather than coincidental.
+ *
+ * A DTLS-SRTP peer that reuses its full TLS server certificate (plus any
+ * intermediates) for the handshake, rather than a single self-signed
+ * leaf cert, can produce a ServerHello+Certificate+... flight larger than
+ * the default below. Since recvfrom() truncates an oversized datagram
+ * silently (see the truncation warning logged in transport_udp.c), such
+ * a peer's handshake would otherwise hang instead of failing loudly.
+ *
+ * Raising this value also halves the video jitter buffer/reassembly
+ * capacity computed in vid_stream.c relative to what the same
+ * MIN_CHUNKS_PER_FRM floor would give at 2000, since that capacity is
+ * derived by dividing a fixed max frame size by this macro. That file
+ * raises its floor to compensate, at the cost of roughly 2x the jitter
+ * buffer memory per video stream (since the jitter buffer's slot size is
+ * this macro's value): about 2 extra MB per stream when DTLS-SRTP is on.
+ *
+ * Default: 4000 when DTLS-SRTP is enabled (#PJMEDIA_SRTP_HAS_DTLS),
+ *          2000 otherwise.
+ */
+#ifndef PJMEDIA_MAX_MRU
+#  if defined(PJMEDIA_SRTP_HAS_DTLS) && (PJMEDIA_SRTP_HAS_DTLS != 0)
+#    define PJMEDIA_MAX_MRU                     4000
+#  else
+#    define PJMEDIA_MAX_MRU                     2000
+#  endif
 #endif
 
 
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -1220,24 +1220,33 @@
  * intermediates) for the handshake, rather than a single self-signed
  * leaf cert, can produce a ServerHello+Certificate+... flight larger than
  * the default below, even with #PJMEDIA_SRTP_DTLS_USE_EC_KEY keeping our
- * own certificate small. Since recvfrom() truncates an oversized datagram
- * silently (see the truncation warning logged in transport_udp.c), such
- * a peer's handshake would otherwise hang instead of failing loudly.
+ * own certificate small (observed ~2500 bytes for such a flight). Since
+ * recvfrom() truncates an oversized datagram silently (see the truncation
+ * warning logged in transport_udp.c), such a peer's handshake would
+ * otherwise hang instead of failing loudly.
  *
- * Raising this value also halves the video jitter buffer/reassembly
- * capacity computed in vid_stream.c relative to what the same
- * MIN_CHUNKS_PER_FRM floor would give at 2000, since that capacity is
- * derived by dividing a fixed max frame size by this macro. That file
- * raises its floor to compensate, at the cost of roughly 2x the jitter
- * buffer memory per video stream (since the jitter buffer's slot size is
- * this macro's value): about 2 extra MB per stream when DTLS-SRTP is on.
+ * Raising this value below 2672 is free: vid_stream.c's video jitter
+ * buffer byte size stays flat, since its slot count shrinks by the same
+ * factor this macro grows. Above that, the reassembly capacity it derives
+ * would drop below one max-size video frame, so vid_stream.c raises its
+ * MIN_CHUNKS_PER_FRM floor to compensate once the shrunk capacity would no
+ * longer hold a full frame, which pins the count and makes jitter buffer
+ * memory scale with this macro from then on. This default (3000) sits
+ * just above that free threshold, keeping the added cost to ~15% of a
+ * video stream's jitter buffer instead of the 2x an unconstrained 4000
+ * would cost. See vid_stream.c's MIN_CHUNKS_PER_FRM comment for the exact
+ * capacity margin this leaves.
  *
- * Default: 4000 when DTLS-SRTP is enabled (#PJMEDIA_SRTP_HAS_DTLS),
+ * 3000 also matches pjnath's PJ_TURN_MAX_PKT_LEN default, so it stops
+ * pjsua_media.c's ICE TURN transport buffer from being silently lowered
+ * from that default down to this macro's value.
+ *
+ * Default: 3000 when DTLS-SRTP is enabled (#PJMEDIA_SRTP_HAS_DTLS),
  *          2000 otherwise.
  */
 #ifndef PJMEDIA_MAX_MRU
 #  if defined(PJMEDIA_SRTP_HAS_DTLS) && (PJMEDIA_SRTP_HAS_DTLS != 0)
-#    define PJMEDIA_MAX_MRU                     4000
+#    define PJMEDIA_MAX_MRU                     3000
 #  else
 #    define PJMEDIA_MAX_MRU                     2000
 #  endif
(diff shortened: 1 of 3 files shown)


nanangizz:
Thanks for the detailed analysis, the silent truncation is a real problem and the write-up made it easy to verify. A few concerns before we take this direction, though.

### 1. ICE is not covered

`transport_udp.c` is only one of the receive paths. With ICE, media (including the multiplexed DTLS traffic) is received through pjnath, sized by `max_pkt_size`, which pjsua sets from `PJMEDIA_MAX_MRU`:

- `pjsip/src/pjsua-lib/pjsua_media.c:1269` (STUN/host candidates)
- `pjsip/src/pjsua-lib/pjsua_media.c:1347` (TURN)

`transport_ice.c` has no receive buffer of its own. So after this PR, ICE + DTLS-SRTP would still truncate at 2000 exactly as before, and ICE is arguably the more likely companion to DTLS-SRTP. Applications using `ice_strans` directly get pjnath's defaults instead, `PJ_STUN_SOCK_PKT_LEN` 2000 and `PJ_TURN_MAX_PKT_LEN` 3000 (`pjnath/include/pjnath/config.h:153`, `:181`). Note the pjsua line above also lowers TURN's buffer from 3000 to 2000.

### 2. The new macro is less complete than reusing the existing one

The stated reason for not touching `PJMEDIA_MAX_MRU`, that `vid_stream.c` uses it for video RTP chunk-size calculations, does not hold on inspection. Both uses are receive-side sizing:

- `vid_stream.c:1799`, `chunks_per_frm = c_strm->frame_size / PJMEDIA_MAX_MRU`, sizes the incoming frame reassembly array, with `MIN_CHUNKS_PER_FRM` as a floor.
- `vid_stream.c:1850`, `pjmedia_jbuf_create(..., PJMEDIA_MAX_MRU, ...)`, the jitter buffer per frame size.

Outgoing packet size comes from `PJMEDIA_MAX_VID_PAYLOAD_SIZE`, derived from `PJMEDIA_MAX_MTU` (`config.h:496`, `:1739`), a separate macro. So `PJMEDIA_MAX_MRU` means what its doc comment says, max packet size for the receiving direction. Raising it costs jitter buffer memory, it does not change what we put on the wire.

That matters because all the affected paths already derive from that one macro: `RTP_LEN`/`RTCP_LEN` here, plus both `max_pkt_size` assignments in `pjsua_media.c`. A deployment hitting this peer can therefore already fix it today with `#define PJMEDIA_MAX_MRU 4000` in `config_site.h`, covering UDP and ICE both, which is more than this PR delivers. The only thing the PR adds is a different default for DTLS builds, and for that the contained form is a conditional default on the existing macro rather than a second knob:

```c
#ifndef PJMEDIA_MAX_MRU
#  if defined(PJMEDIA_SRTP_HAS_DTLS) && (PJMEDIA_SRTP_HAS_DTLS != 0)
#    define PJMEDIA_MAX_MRU                     4000
#  else
#    define PJMEDIA_MAX_MRU                     2000
#  endif
#endif
```

### 3. How common is the triggering case

DTLS-SRTP authenticates by matching the certificate fingerprint carried in SDP (RFC 5763), not by chain validation, so a certificate chain adds nothing and costs every handshake. The near universal practice is a single self-signed certificate with no chain: WebRTC endpoints generate an ephemeral self-signed ECDSA P-256 cert, and pjsip does the same with RSA-2048 (`transport_srtp_dtls.c:418`), with the verify callback deliberately skipping validation at `:383`.

A flight large enough to overflow 2000 bytes therefore comes from a peer reusing its TLS server certificate plus intermediates for DTLS-SRTP. Some SBCs and media servers do this, so the report is valid, but it is a minority configuration rather than the common case. That argues against raising a buffer for every DTLS build to accommodate it.

### 4. What is worth fixing regardless of the size

The core defect in your report is not that 2000 is too small, it is that exceeding it fails silently: `pj_sock_recvfrom()` has no `MSG_TRUNC`, so the truncated bytes go straight into `BIO_write()` and both peers retransmit forever with nothing logged. No constant is provably sufficient (a longer chain or RSA-4096 defeats 4000 as well), so the durable improvement is to make it loud: when a received datagram exactly fills the buffer in `transport_udp.c`, log a warning that the packet was likely truncated and that `PJMEDIA_MAX_MRU` may need raising. That turns an undiagnosable hang into a one line clue, and it helps the ICE paths too once they carry the same check.

So the shape we would prefer: drop `PJMEDIA_MAX_MRU_DTLS`, add the truncation warning, and document the chain case in `PJMEDIA_MAX_MRU`'s doc comment. Raising the DTLS build default via the conditional above is optional on top of that. Would you be willing to rework it along those lines?

### Unrelated observation

Our own generated cert is RSA-2048. Switching `ssl_generate_cert()` to ECDSA P-256 cuts it from 694 to 297 bytes DER, and the key generation in `dtls_init()` from ~145ms to ~0.07ms (measured, OpenSSL 3.0.13). Out of scope here, now #5194.

Correcting the flight size I estimated above: measured, it is ~1.2 to 1.4 KB, not 1.5 to 1.9 KB.

average-enjoyer:
Thanks for the thorough review!
You're right on all counts, and point 2 ("the new macro is less complete than reusing the existing one") in particular caught a mistake in my own reasoning. I claimed `vid_stream.c` used `PJMEDIA_MAX_MRU` for outgoing chunk sizing, which is wrong: `chunks_per_frm` (`vid_stream.c:1799`) and the `pjmedia_jbuf_create()` call (`:1850`) are both receive-side sizing, and the actual outgoing wire size (`PJMEDIA_MAX_VID_PAYLOAD_SIZE`) is derived from `PJMEDIA_MAX_MTU`, a separate macro. So there was never a real reason to introduce a second knob.

I've reworked this along the lines you suggested:

- Dropped `PJMEDIA_MAX_MRU_DTLS`. `RTP_LEN`/`RTCP_LEN` in `transport_udp.c` are back to using `PJMEDIA_MAX_MRU` directly, unchanged from before this PR.
- `PJMEDIA_MAX_MRU` itself now defaults to 4000 under `PJMEDIA_SRTP_HAS_DTLS` and 2000 otherwise, using the conditional-default form you proposed. This also covers the ICE/STUN/TURN `max_pkt_size` paths in `pjsua_media.c` for free, since they already derive from this same macro.
- Updated `PJMEDIA_MAX_MRU`'s doc comment to describe the chain-cert scenario and note that it also bounds the ICE transport buffers.
- Added a `PJ_LOG` warning (level 2) in `transport_udp.c`'s `call_rtp_cb()`/`call_rtcp_cb()` — the two functions every receive path funnels through — firing whenever a received packet exactly fills the buffer, pointing at `PJMEDIA_MAX_MRU` as the likely fix. That should turn a silent hang into a one-line diagnostic regardless of what buffer size someone's running with.

On ICE: raising the shared macro fixes the STUN/TURN `max_pkt_size` sizing in `pjsua_media.c`, but I haven't touched `transport_ice.c`/pjnath itself or added the truncation warning there — happy to extend the warning to that path too if you'd like it in scope for this PR, otherwise I'll leave it as a known follow-up.

On the TURN buffer regression you flagged (pjsua lowering it from pjnath's own 3000 default to 2000) — I left that alone since it's a separate, pre-existing issue independent of this fix; let me know if you'd rather it get folded in here.

Also noted your point on certificate size (RSA-2048 self-signed putting our own flight uncomfortably close to 2000) — glad to hear that's being tracked separately.

One more thing worth flagging on myself: my original "How Has This Been Tested?" claimed the build exercised the DTLS branch with "OpenSSL SRTP enabled." That was inaccurate — `PJMEDIA_SRTP_HAS_DTLS` isn't auto-enabled by OpenSSL being present, it's a separate opt-in flag, and my `config_site.h` never set it, so that earlier build had silently compiled the non-DTLS path the whole time. Re-verified just now with `PJMEDIA_SRTP_HAS_DTLS=1` explicitly set, confirmed `PJMEDIA_MAX_MRU` actually resolves to 4000 under it, and the build is still clean. Updated PR description reflects this properly now.

Updated diff and PR description incoming.

nanangizz:
Thanks for the rework, and for the note correcting your own test claim. The shape is right now: one macro, ICE covered for free via the `pjsua_media.c` derivations, and the truncation warning is the part that keeps paying off whatever buffer size someone runs.

Three things, the last one I'd want fixed before merge.

### 1. The conditional default depends on macro ordering

`PJMEDIA_MAX_MRU` resolves at `config.h:505`, but `PJMEDIA_SRTP_HAS_DTLS` is only defaulted at `config.h:1199`. It works today in both directions, since the only way it is defined that early is `config_site.h` or `-D`, and the later default is 0. But it is silent if that ever changes: a default of 1, or the macro arriving from a generated header included after line 505, would leave the MRU at 2000 and this fix would vanish with no diagnostic. Please move the `PJMEDIA_MAX_MRU` block below the SRTP section so the dependency is positional rather than coincidental.

### 2. Raising the MRU changes video reassembly and jitter buffer depth

Your original instinct that `vid_stream.c` is coupled to this macro was right, just not through the outgoing size:

```c
chunks_per_frm  = c_strm->frame_size / PJMEDIA_MAX_MRU;   /* vid_stream.c:1799 */
```

`frame_size` is `enc_w * enc_h * 4` capped at `PJMEDIA_MAX_VIDEO_ENC_FRAME_SIZE` (`:1697`, `:1711-1715`), so for any real resolution it is simply 131072 — the `max_bps` estimate above it is dead code inside `#if 0`. It then feeds `jb_max` (`:1805`, `:1809`) and `stream->rx_frame_cnt` (`:1844`), so doubling the divisor halves both:

- **Reassembly:** capacity drops from 130 to 64 packets, while a maximum size frame is 131072 / 1336 = 98 packets at SRTP payload size. Frames beyond roughly 85 kB now hit `:1156`, which discards the excess and logs `"Discarding %u frames because array is full!"`, yielding an incomplete frame. HD keyframes reach that size.
- **Jitter buffer:** `jb_max` is `500ms * chunks_per_frm / frm_ptime`, i.e. milliseconds converted to packets with this same estimate, so its depth halves from about 500ms to 250ms. More late packet drops on jittery links.

Neither is memory-unsafe, but both are real video regressions in DTLS-enabled builds, and neither is visible from this diff.

**Suggested fix:** raise `MIN_CHUNKS_PER_FRM` (`vid_stream.c:68`), which already exists to guarantee a minimum packets-per-frame estimate and is used nowhere else:

```c
#define MIN_CHUNKS_PER_FRM      65
```

That restores today's counts exactly and holds them for any larger MRU:

| | `rx_frame_cnt` | `jb_max` | jbuf bytes |
|---|---|---|---|
| today (MRU 2000, floor 30) | 130 | 984 | 1.97 MB |
| this PR (MRU 4000, floor 30) | 64 | 484 | 1.94 MB |
| MRU 4000, floor 65 | 130 | 984 | 3.94 MB |
| MRU 4000, floor 98 | 196 | 1484 | 5.94 MB |

Note the cost, since it is unavoidable: any variant that preserves the packet counts doubles the jitter buffer allocation, because the slot size is the MRU (`jbuf.c:201`) while the count is pinned. The current diff is the only variant that keeps memory flat, and it does that by halving capacity. So the choice is roughly 2 MB per video stream against a 64 packet reassembly limit facing 98 packet frames.

Two caveats on the value. 65 equals today's number only by the coincidence that 131072 / 2000 is 65; the derived floor is 98, the frame size cap over the 1336 byte max payload, which also repairs a pre-existing thin margin, today's 130 against 98 being only 1.3x. And a floor of 65 will over-allocate slightly for configurations with a smaller `PJMEDIA_MAX_VIDEO_ENC_FRAME_SIZE`, where the floor starts to bind.

Deriving the estimate from `PJMEDIA_MAX_VID_PAYLOAD_SIZE` instead would be the more principled fix, but it belongs in a follow-up: the surrounding sizing already mixes directions, with `frame_size` coming from the encoder format while it sizes decode side buffers (`:1697`), and `frm_ptime` taken from `vfd_dec->fps` (`:1798`) while the jitter buffer is created with `vfd_enc->fps` (`:1852`). Untangling that deserves its own change and its own testing.

Whichever way you go, please state the effect in the PR description and the `PJMEDIA_MAX_MRU` doc comment, so it is recorded rather than a surprise for video users.

### 3. The truncation warning is a remote, unauthenticated log flood

`call_rtp_cb()` does no validation — it snapshots the callbacks and hands the packet up (`transport_udp.c:507-580`). The source address check happens above this layer via `param.rem_switch`, and SRTP authentication above that. The new warning sits at the top of the function, so it fires before any address check, authentication, or RTP header sanity check. `call_rtcp_cb()` (`:596`) is the same.

So it takes only reaching the media port and sending datagrams at or above the buffer size — the kernel truncates any oversized datagram to exactly that size, making the condition a guaranteed hit. One packet, one log line, with no call state or keys, and the source may as well be spoofed. Each line is a synchronous level 2 `PJ_LOG` on the ioqueue thread servicing RTP receive, under the process-wide log mutex, so at 1 Gbps (~60k pps) it is both ~15 MB/s of log and a stall on the media path for every stream in the process. That makes us the amplifier, in a patch meant to harden.

A single counter keeps the diagnostic and caps the rate, logging the first and then every hundredth:

```c
    if (bytes_read > 0 && (pj_size_t)bytes_read == sizeof(udp->rtp_pkt) &&
        udp->rtp_trunc_cnt++ % 100 == 0)
    {
        PJ_LOG(2,(udp->base.name,
                  "Received RTP packet filled the %d-byte buffer and was "
                  "likely truncated (%u so far), consider raising "
                  "PJMEDIA_MAX_MRU",
                  (int)sizeof(udp->rtp_pkt), udp->rtp_trunc_cnt));
    }
```

One `unsigned` per direction, no teardown hook, and a persistent condition still reports itself. If you want a harder cap, `(udp->rtp_trunc_cnt & (udp->rtp_trunc_cnt - 1)) == 0` logs on powers of two with the same single variable, about 17 lines per 100k packets instead of 1000.

### Minor

- The two warning blocks are identical apart from the buffer; worth a shared helper or format string.

average-enjoyer:
Thanks for digging this deep!
All three land, and point 3 especially: you're right that this was an unauthenticated, remotely-triggerable log point sitting below any address/auth check. Pushed a new commit addressing all three.

**Macro ordering** — moved the `PJMEDIA_MAX_MRU` block to right after `PJMEDIA_SRTP_HAS_DTLS`'s own definition, so the dependency is now positional instead of relying on nothing else defining it earlier.

**Video reassembly/jitter buffer** — raised `MIN_CHUNKS_PER_FRM` from 30 to 65, restoring today's exact `rx_frame_cnt`/`jb_max` counts under the larger MRU, per your table. Went with 65 rather than the more principled 98 since you called that out as its own follow-up with its own testing. Documented the ~2 extra MB/stream cost in both the `PJMEDIA_MAX_MRU` doc comment and next to `MIN_CHUNKS_PER_FRM`, so it's recorded rather than a surprise.

**Log flood** — added a per-direction counter (`rtp_trunc_cnt`/`rtcp_trunc_cnt`) to `struct transport_udp`, rate-limited to the first hit and then every 100th, and folded the duplicated RTP/RTCP warning blocks into a shared `warn_if_truncated()` helper per your minor note. Went with the modulo-100 form rather than the power-of-two variant since it keeps giving periodic feedback under a sustained condition instead of tailing off almost entirely — happy to switch if you'd rather have the harder cap.

Rebuilt with `PJMEDIA_SRTP_HAS_DTLS=1` genuinely set (not just OpenSSL present) and confirmed clean, including that `PJMEDIA_MAX_MRU` still resolves to 4000 after the reorder.

nanangizz:
Rebase looks good — both config blocks kept and the ordering fix intact. The rate-limited truncation warning and the shared helper are what I wanted, and the video floor restores the counts correctly. One remaining concern, on the size itself.

I've been reconsidering the 4000. Since #5194 landed our own certificate is much smaller, so the only case 4000 still buys is a peer reusing a full TLS chain. Your own measurement puts that at ~2500 bytes, which is already an edge case — and the cost of covering it lands on every DTLS build with video, met or not. I'd propose 3000 instead.

Worth being precise about where that cost comes from. Raising MRU alone is free: `jb_max` shrinks by exactly the factor the slot grows, so jbuf bytes stay flat. What it does instead is halve reassembly capacity. Raising the floor to restore that capacity pins the count, and from then on memory scales linearly with MRU. The two are only expensive together. The limit is `rx_frame_cnt * 1336 >= 131072`, which holds up to MRU 2672 — below that the floor never binds and the increase is genuinely free.

| MRU | rx_cnt | jbuf | vs today |
|---|---|---|---|
| 2000 today | 130 | 1.97 MB | — |
| 3000 + floor 50 | 100 | 2.27 MB | +15% |
| 4000 + floor 65 | 130 | 3.94 MB | +100% |

So please use 3000 with `MIN_CHUNKS_PER_FRM` 50:

- Still clears your ~2500-byte flight with room to spare.
- 50 is derived, not coincidental: `ceil(131072/1336)/2`, the smallest value keeping `rx_frame_cnt` above one max-size frame. My earlier 65 only matched today's number by the accident of `131072/2000`. Non-DTLS builds are unaffected, `max(65, 50)` is still 65.
- 3000 matches `PJ_TURN_MAX_PKT_LEN`, so `pjsua_media.c:1356` stops lowering TURN's buffer from 3000 to 2000 — fixes the regression from my first review for free.

Worth recording in the doc comment: `rx_frame_cnt` goes 130 to 100 against a 99-packet worst case. Today's 130 was never a designed margin, but it should be written down.

average-enjoyer:
Thanks for the review!
Applied both exactly as specified:

- `PJMEDIA_MAX_MRU` under `PJMEDIA_SRTP_HAS_DTLS` is now 3000, not 4000.
- `MIN_CHUNKS_PER_FRM` is now 50, derived from `ceil(131072/1336)/2` as you laid out, not the coincidental 65.
- Updated both doc comments with the free-below-2672 reasoning, the ~15% vs 2x cost comparison, the TURN buffer alignment, and the 100-vs-99-packet margin — written down as you asked, not left as a surprise.

Rebuilt clean (had to `find . -name "*.depend" -delete` first — this repo's dependency tracking wasn't picking up header-only changes for me during the earlier merge, worth knowing if you hit the same locally). Confirmed via preprocessor dump that `PJMEDIA_MAX_MRU` now resolves to 3000 under the flag.

nanangizz:
LGTM. I appreciate the contribution and responsive review‑fix cycle.
Will wait for CI and review from other (if any).

sauwming (approved):

nanangizz (approved):

Issue #5159, opened 2026-08-07
pjsua: update_rfc5626_status() records outbound as unused but the regc keeps advertising it


laconicman:
### Describe the bug

`update_rfc5626_status()` records that SIP outbound is not in use, but never tells the client registration, so the account keeps advertising outbound on every refresh.

On a 2xx that does not confirm outbound, the function sets `acc->rfc5626_status = OUTBOUND_NA` and strips the outbound parameters from the account's Contact:

```c
on_return:
    if (acc->rfc5626_status != OUTBOUND_ACTIVE) {
        acc->reg_contact = acc->contact;
    }
```
<sub>`pjsip/src/pjsua-lib/pjsua_acc.c`, `update_rfc5626_status()`</sub>

It does not call `pjsip_regc_update_contact()`, and neither does the 2xx path in `regc_cb()` that invokes it, unless `acc_check_nat_addr()` happens to rewrite the Contact for unrelated reasons. `pjsip_regc_update_contact()` is reached only from `acc_check_nat_addr()`, `auto_rereg_timer_cb()` and `pjsua_acc_update_contact_on_ip_change()`.

The regc therefore keeps sending the Contact and the `Supported` header it was initialised with. Account state and wire disagree until something re-inits the regc.

Two consequences:

1. **`rfc5626_status` and `acc->reg_contact` stop describing what is being sent.** Anything keyed on them is wrong about the actual request. This is not hypothetical — it invalidated an otherwise reasonable fix in #5154, where gating on `rfc5626_status` silently left accounts unregistered.
2. **The account keeps presenting the 439 trigger.** RFC 5626 §6 has a registrar answer 439 (`PJSIP_SC_FIRST_HOP_LACKS_OUTBOUND_SUPPORT`) when a REGISTER carries `reg-id` plus the `outbound` option tag and the first hop is not outbound-aware. Every refresh still carries both, so a later change of first hop — failover, roaming, a proxy reconfiguration — draws a 439 for a registration where outbound had already been determined unavailable.

Note this is the *normal* path for a registrar without outbound support: such a registrar simply answers 200 with no `Require: outbound`, so an account in this state is the steady state rather than an edge case.

### Steps to reproduce

1. Configure an account over TCP or TLS with defaults, so `use_rfc5626 == PJ_TRUE`.
2. Register to a registrar that answers 200 **without** `Require: outbound`. `update_rfc5626_status()` logs `SIP outbound status for acc N is not active`.
[...]

PR #5191, opened 2026-08-16
pjsua: sync regc contact when outbound turns out unsupported

arccat-114:
Ref #5159

## Problem

`update_rfc5626_status()` records `OUTBOUND_NA` on a 2xx that does not confirm outbound, and rewrites `acc->reg_contact` to the outbound-free contact — but it never tells the client registration. The regc keeps the Contact from the initial registration, so every refresh REGISTER keeps advertising outbound (`reg-id`, `+sip.instance`, `;ob`, `outbound` option-tag) while `acc->rfc5626_status` says `OUTBOUND_NA` and `acc->reg_contact` no longer carries those parameters. State and wire disagree.

The regular refresh path goes through the regc's own timer with the regc's stored contact; the contact-regeneration in `auto_rereg_timer_cb()` only runs on the auto-rereg retry path, so the divergence persists across refreshes.

## Fix

In `update_rfc5626_status()`, after rewriting `acc->reg_contact`, sync the regc with `pjsip_regc_update_contact()` — the same pattern already used by the NAT-bound-address rewrite paths (`acc_check_nat_addr()` and the IP-change recovery path). The next refresh then registers the outbound-free Contact, matching the recorded status.

## Validation

- `./configure && make dep && make` (Linux, pjsua/pjsua-lib included): builds clean, no warnings in the touched file.
- Code-path review: `pjsip_regc_update_contact()` updates the Contact used for the *next* REGISTER; the call site here mirrors the existing NAT-rewrite call sites. The regc is non-null on this path (registration just succeeded via it).
- Full SIPp-based scenario tests are not run in this environment (no SIPp); the change is confined to the same code region the existing contact-rewrite paths exercise.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_acc.c
+++ b/pjsip/src/pjsua-lib/pjsua_acc.c
@@ -2996,6 +2996,16 @@ static void update_rfc5626_status(pjsua_acc *acc, pjsip_rx_data *rdata)
 on_return:
     if (acc->rfc5626_status != OUTBOUND_ACTIVE) {
         acc->reg_contact = acc->contact;
+        /* The regc still carries the outbound-parameter Contact from the
+         * initial registration, so refresh REGISTERs keep advertising
+         * outbound even though we just learned the server does not
+         * support it. Sync the regc the same way the NAT rewrite paths
+         * do (see pjsip_regc_update_contact() call sites), so state and
+         * wire agree from the next refresh on.
+         */
+        if (acc->regc) {
+            pjsip_regc_update_contact(acc->regc, 1, &acc->reg_contact);
+        }
     }
     PJ_LOG(4,(THIS_FILE, "SIP outbound status for acc %d is %s",
                          acc->index, (acc->rfc5626_status==OUTBOUND_ACTIVE?


nanangizz:
Thanks for the analysis, the described divergence is real: `pjsip_regc_init()` is seeded with `&acc->reg_contact` (`pjsua_acc.c:3275`), refreshes reuse the regc's stored Contact, and nothing pushes an update when the status flips to `OUTBOUND_NA`. Your reading of `auto_rereg_timer_cb()` (`:5061`) being the only regeneration path is correct too.

That said, I don't think we should merge this, for two reasons: the patch would drop configured Contact params on the wire, and the divergence it removes is actually load-bearing.

### 1. The pushed Contact has the configured params stripped

`update_rfc5626_status()` does `acc->reg_contact = acc->contact` (`:2865`), a raw assignment that bypasses `update_regc_contact()` (`:2098`) which is what appends `reg_contact_uri_params`, `contact_params`, outbound params, and `reg_contact_params`.

Today that stripped value is latent, because every path that pushes to the regc rebuilds first (`:5153` in the auto-rereg path, `:3268` via `pjsua_regc_init()`); only the IP-change retry (`:5398`) can leak it. This patch pushes it directly, so any account configured with `reg_contact_params` / `contact_params` / `reg_contact_uri_params`, push notification params being the common case, would lose them from the next refresh REGISTER onward.

If we ever do sync here, it has to rebuild first, the way `auto_rereg_timer_cb()` does:

```c
    update_regc_contact(acc);
    if (acc->regc)
        pjsip_regc_update_contact(acc->regc, 1, &acc->reg_contact);
```

### 2. It would fire on every successful REGISTER

`OUTBOUND_NA` is the steady state for every account that does not use outbound (`update_regc_contact()` sets it at `:2190`/`:2210`), and the `!= OUTBOUND_ACTIVE` condition also covers `OUTBOUND_UNKNOWN`. So the new call would run on every 2xx for the majority of accounts, not only on the one transition it targets. It would need to key off an actual `WANTED -> NA` change.

### 3. The divergence preserves a recovery path, which is the bigger point

`OUTBOUND_NA` is sticky: `update_regc_contact()` short-circuits on it at `:2110`, so the outbound params are never re-added, and nothing resets `NA` back to `WANTED` for the life of the regc. The reset at `:5330` only downgrades `ACTIVE -> WANTED` on IP change, and `:302` only clears the status when the regc is destroyed.

But `update_rfc5626_status()` *can* still flip `NA -> ACTIVE`: only `OUTBOUND_UNKNOWN` short-circuits at `:2843`, so if a later 2xx carries `Require: outbound`, the account transparently upgrades. That is exactly what happens when the REGISTER lands on a capable node during a rolling upgrade of a registrar cluster, or when the server gains outbound support.

That upgrade only works because the regc keeps advertising `reg-id` and `+sip.instance`: a registrar includes `Require: outbound` only in response to a REGISTER that requested outbound. Sync the regc to the outbound-free Contact and the request stops asking, the server therefore stops answering, and the account is locked out of outbound permanently, with no path back short of recreating the regc.

Keeping the params is also provably safe for the server in question, since the initial REGISTER already carried them and was answered with a 2xx. So the cost of the current behavior is a few dozen bytes per refresh that a non-supporting server ignores, and the benefit is automatic recovery. That matches Benny's assessment that it is harmless.

### Suggestion

I'd rather close this and keep the current behavior. If it is worth any change at all, it would be a comment at `:2864` recording that the regc is intentionally not synced, so the next reader does not see it as an oversight, and possibly dropping the misleading `acc->reg_contact = acc->contact` assignment since nothing consumes the stripped value except the IP-change retry path.

If we later decide state/wire consistency is worth more than the upgrade path, then the correct shape is the rebuild above plus a transition-only guard, and we would want a note in the docs that outbound will not be retried until re-registration from scratch.

For the record, the mechanics of your patch are otherwise sound: `regc->lock` is a recursive mutex (`sip_reg.c:159`), so calling `pjsip_regc_update_contact()` from inside the regc callback is safe, and the ordering against `acc_check_nat_addr()` (`:3023` before `:3027`) is fine.

arccat-114:
Thanks for the detailed review. You're right that the direct regc update would drop configured Contact parameters and remove the outbound recovery path. Given those trade-offs, this patch isn't safe to merge, so I'll close it rather than pursue the current approach. The analysis was still useful in documenting the state divergence.

arccat-114:
Closing this based on the maintainer review; the current behavior preserves the outbound recovery path.

PR #5192, opened 2026-08-17
pjmedia: validate bundle-only media as active media

nanangizz:
- A bundle-only media (RFC 9143) has zero port but is not disabled, it is negotiated as any active media since #5067. However, SDP validation still relaxes the format, connection info, and rtpmap checks for any zero port media.
- Consequently, an offer containing a bundle-only media without rtpmap for its dynamic payload type passes the validation and then hits the assertion in `match_offer()`, i.e. "Bug! Offer should have been validated".
- Now `pjmedia_sdp_validate2()` validates bundle-only media as if it had non-zero port, so such offer is rejected with `PJMEDIA_SDP_EMISSINGRTPMAP` or `PJMEDIA_SDP_ENOFMT`, instead of being negotiated as if it was validated.
- The bundle-only check is moved from sdp_neg.c to sdp.c and shared via a new private header, so nothing bundle related is published.

Note that in non-assert build, the offer was not negotiated either, `match_offer()` simply returned an error and the media got rejected in the answer, so there is no memory safety issue.

Credits to OSS-Fuzz and libFuzzer for finding this bug via the fuzz-sdp target.

## Test plan

- New `sdp_neg_bundle_only_invalid_test()` in pjmedia sdp_neg test, covering bundle-only media without rtpmap and without format.
- Verified that the assertion is reproducible before the fix and gone after it.
- `pjmedia-test sdp_neg_test` passed.

Co-Authored-By Claude Code

### DIFF
--- a/pjmedia/include/pjmedia/sdp.h
+++ b/pjmedia/include/pjmedia/sdp.h
@@ -783,7 +783,10 @@ PJ_DECL(pj_status_t) pjmedia_sdp_validate(const pjmedia_sdp_session *sdp);
  *
  * @param sdp       The SDP session descriptor to validate.
  * @param strict    Flag whether the check should be strict, i.e: allow
- *                  media without connection line when port is zero.
+ *                  media without connection line when port is zero. Note
+ *                  that this relaxation, along with the format and rtpmap
+ *                  checks, is not applied to bundle-only media, as such
+ *                  media is not disabled despite having zero port.
  *
  * @return          PJ_SUCCESS on success.
  */
--- a/pjmedia/src/pjmedia/sdp.c
+++ b/pjmedia/src/pjmedia/sdp.c
@@ -18,6 +18,7 @@
  */
 #include <pjmedia/sdp.h>
 #include <pjmedia/errno.h>
+#include "sdp_internal.h"
 #include <pjlib-util/scanner.h>
 #include <pj/array.h>
 #include <pj/except.h>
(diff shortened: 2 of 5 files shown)


sauwming:
- pjmedia_sdp_media_is_bundle_only is called twice per media in validate2.

is_active is assigned at the top of the media loop by calling pjmedia_sdp_media_is_bundle_only(sdp, m), which itself calls pjmedia_sdp_media_find_attr2 twice (once for bundle-only, once for mid) and then walks all session-level attributes in bundle_group_contains_mid. For typical SDPs with few attributes this is immaterial, but it's O(media × session_attrs) per call. Since the early CHECK already gates on port == 0, you could short-circuit: only call pjmedia_sdp_media_is_bundle_only when port == 0.

- Missing test: bundle-only without a mid attribute.

pjmedia_sdp_media_is_bundle_only returns PJ_FALSE when the media has no a=mid, so such a section would be treated as ordinary disabled-zero-port media (not bundle-only), and the relaxations would apply to it. That's arguably correct per RFC 9143, but it might be worth adding a test case to document the expected behavior explicitly — e.g. a section with port=0, a=bundle-only, but no a=mid. Right now that falls through quietly as non-bundle-only.

- Missing test: bundle-only section whose MID is not listed in group:BUNDLE.

Similarly, a section with all the right attributes (port=0, a=bundle-only, a=mid:X) but where X is not in the session-level group:BUNDLE would be treated as ordinary zero-port disabled media. A test case documenting this edge would complete the picture.

nanangizz:
Thanks for the review!

**1. `pjmedia_sdp_media_is_bundle_only()` cost**

The suggested short-circuit is already in place at both call sites, so the two `find_attr2()` walks and the session level attribute walk only ever run for zero port media:

- `sdp.c`, in `validate2()`: `is_active = (m->desc.port != 0) || pjmedia_sdp_media_is_bundle_only(sdp, m);` — the `||` short-circuits, so the helper is not called at all when the port is non-zero.
- `sdp_neg.c`, in `create_answer()`: the call is unconditional, but the helper's own first check is `if (media->desc.port != 0 || !find_attr2(media, "bundle-only", NULL)) return PJ_FALSE;`, i.e. one integer compare for ordinary media.

What remains is that a zero port media goes through the helper twice, once when validating and once when answering. That seems fine to keep as is: `pjmedia_sdp_validate2()` is a public API that may be called standalone, so caching the result would mean carrying bundle state across the API boundary, and zero port sections carry few attributes anyway.

**2 & 3. Missing tests for non bundle-only zero port media**

Good catch, both cases were indeed undocumented. Added `sdp_neg_bundle_only_relaxed_test()` in 7f334e5, covering:

- zero port media with `a=bundle-only` but no `a=mid`
- zero port media with `a=bundle-only` and `a=mid:1`, but the MID is not listed in the session level `a=group:BUNDLE`

Both offers deliberately omit the rtpmap for their dynamic payload type, so the test would fail if these were ever treated as bundle-only. It asserts that `pjmedia_sdp_validate2()` returns `PJ_SUCCESS`, i.e. the relaxations still apply, and that after negotiation the media is rejected in the answer (zero port), just like any other disabled media.

`pjmedia-test sdp_neg_test` passed.

sauwming (approved):

PR #5194, opened 2026-08-17
pjmedia: use EC P-256 key for DTLS-SRTP self-signed certificate

nanangizz:
- The DTLS-SRTP certificate we generate is self-signed and authenticated by its fingerprint in the SDP, not by PKI, so the key type is purely a local choice. The RSA 2048 bit key dates back to the initial DTLS-SRTP implementation (edcedb569) with no interop rationale recorded.
- Switch it to an EC key on the P-256 curve, which is what peers commonly present, and which RFC 8827 requires for WebRTC: *"All implementations MUST support DTLS 1.2 with the TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 cipher suite and the P-256 curve"*. RSA cipher suites carry no such requirement.
- Measured with OpenSSL 3.0.13 on the exact certificate this code builds: certificate DER **694 to 297 bytes**, and the one time key generation in `dtls_init()` **~145ms to ~0.07ms**. The ServerKeyExchange signature also shrinks, from 256 bytes with RSA 2048 to about 72 with ECDSA P-256. A smaller flight is welcome given how easily an oversized one gets silently truncated on receive (see #5190).
- The curve is encoded by name (`OPENSSL_EC_NAMED_CURVE`), as some peers reject certificates carrying explicit curve parameters.
- `EC_KEY_new_by_curve_name()` is used rather than the `EVP_PKEY_CTX` form, since it behaves the same across the OpenSSL versions supported here and matches existing usage in `pjlib/src/pj/ssl_sock_ossl.c`.

**Behavior change and escape hatch.** A peer that cannot negotiate an ECDSA cipher suite would now fail the handshake outright rather than degrade. Compliant WebRTC peers cannot be in that set, but legacy SIP equipment conceivably could, and so could a deployment that has narrowed `PJMEDIA_SRTP_DTLS_OSSL_CIPHERS` to an RSA only list. `PJMEDIA_SRTP_DTLS_USE_EC_KEY` is therefore added, enabled by default, so such deployments can restore the previous RSA key from `config_site.h` without patching. Worth a release note regardless.

Key generation moves into its own function so the certificate code stays free of the conditional. The error path also no longer leaks the key on the (unreachable in practice) `EVP_PKEY_assign_*` failure.

## Test plan

With `PJMEDIA_SRTP_HAS_DTLS 1`, both branches of the new option were built and exercised:

| | build | `160_srtp_dtls` | `320_ice_dtls_srtp` |
|---|---|---|---|
| `PJMEDIA_SRTP_DTLS_USE_EC_KEY 1` (default) | clean, no new warnings | passed | passed |
| `PJMEDIA_SRTP_DTLS_USE_EC_KEY 0` (RSA) | clean, no new warnings | passed | passed |

Both tests complete a real DTLS-SRTP handshake between two pjsua instances, so the generated certificate is exercised end to end in each configuration.

Co-Authored-By Claude Code

### DIFF
--- a/pjmedia/include/pjmedia/config.h
+++ b/pjmedia/include/pjmedia/config.h
@@ -1201,6 +1201,22 @@
 
 
 /**
+ * Use EC key on the P-256 curve for the DTLS-SRTP self-signed certificate,
+ * which is the key type commonly used by DTLS-SRTP peers and the one required
+ * by RFC 8827 for WebRTC. It also makes the certificate, and hence the
+ * handshake, significantly smaller than RSA.
+ *
+ * Disable this to use RSA 2048 bit key instead, e.g: for interoperability
+ * with peers that do not support ECDSA cipher suites.
+ *
+ * Default value: 1 (enabled)
+ */
+#ifndef PJMEDIA_SRTP_DTLS_USE_EC_KEY
+#   define PJMEDIA_SRTP_DTLS_USE_EC_KEY             1
+#endif
+
+
+/**
  * Set OpenSSL ciphers for DTLS-SRTP.
  *
  * Default value: "DEFAULT"
--- a/pjmedia/src/pjmedia/transport_srtp_dtls.c
+++ b/pjmedia/src/pjmedia/transport_srtp_dtls.c
@@ -27,7 +27,9 @@
  * Include OpenSSL headers
  */
 #include <openssl/bn.h>
+#include <openssl/ec.h>
 #include <openssl/err.h>
+#include <openssl/obj_mac.h>
 #include <openssl/rsa.h>
 #include <openssl/ssl.h>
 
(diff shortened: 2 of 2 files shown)


sauwming (approved):

dev31337:
Your commit has a lot of opensssl 3.0 deprecated warnings:

Is it possible to migrate it to non-deprecated functions ?

In file included from ../src/pjmedia/transport_srtp.c:501:
../src/pjmedia/transport_srtp_dtls.c: In function ‘ssl_generate_key’:
../src/pjmedia/transport_srtp_dtls.c:426:5: warning: ‘EC_KEY_new_by_curve_name’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declaration ]
  426 |     ec_key = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1);
      |     ^~~~~~
In file included from ../src/pjmedia/transport_srtp_dtls.c:30:
/usr/include/openssl/ec.h:998:31: note: declared here
  998 | OSSL_DEPRECATEDIN_3_0 EC_KEY *EC_KEY_new_by_curve_name(int nid);
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~
../src/pjmedia/transport_srtp_dtls.c:432:5: warning: ‘EC_KEY_set_asn1_flag’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  432 |     EC_KEY_set_asn1_flag(ec_key, OPENSSL_EC_NAMED_CURVE);
      |     ^~~~~~~~~~~~~~~~~~~~
/usr/include/openssl/ec.h:1087:28: note: declared here
 1087 | OSSL_DEPRECATEDIN_3_0 void EC_KEY_set_asn1_flag(EC_KEY *eckey, int asn1_flag);
      |                            ^~~~~~~~~~~~~~~~~~~~
../src/pjmedia/transport_srtp_dtls.c:434:5: warning: ‘EC_KEY_generate_key’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  434 |     if (!EC_KEY_generate_key(ec_key))
      |     ^~
/usr/include/openssl/ec.h:1101:27: note: declared here
 1101 | OSSL_DEPRECATEDIN_3_0 int EC_KEY_generate_key(EC_KEY *key);
      |                           ^~~~~~~~~~~~~~~~~~~
../src/pjmedia/transport_srtp_dtls.c:441:5: warning: ‘EVP_PKEY_assign’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  441 |     if (!EVP_PKEY_assign_EC_KEY(priv_key, ec_key))
      |     ^~
In file included from /usr/include/openssl/rand.h:23,
                 from ../src/pjmedia/transport_srtp.c:42:
/usr/include/openssl/evp.h:1330:5: note: declared here
 1330 | int EVP_PKEY_assign(EVP_PKEY *pkey, int type, void *key);
      |     ^~~~~~~~~~~~~~~
../src/pjmedia/transport_srtp_dtls.c:448:5: warning: ‘EC_KEY_free’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  448 |     EC_KEY_free(ec_key);
      |     ^~~~~~~~~~~
/usr/include/openssl/ec.h:1003:28: note: declared here
 1003 | OSSL_DEPRECATEDIN_3_0 void EC_KEY_free(EC_KEY *key);
      |                            ^~~~~~~~~~~

dev31337:
I replaced all your code in "ssl_generate_key" by a few (non-deprecated) lines to generate the key:

static pj_status_t ssl_generate_key(EVP_PKEY **p_priv_key)
{
priv_key = EVP_EC_gen("P-256");
if (!priv_key)
    return PJ_EUNKNOWN;

*p_priv_key = priv_key;
return PJ_SUCCESS;
}

I tested it and it works. (and it's OPENSSL_EC_NAMED_CURVE by default)

One last thing the certificate is signed with SHA1, I switched to SHA256 (not in your commit but later in the sources) for better compatibility since the extra bytes required are no longer a problem thanks to your RSA->ECDSA byte economy.

nanangizz:
Thanks for the report and for testing a replacement — you're right, and `EVP_EC_gen()` is the better call. I've opened #5231 with it, guarded by `OPENSSL_VERSION_NUMBER >= 0x30000000L` so older OpenSSL and forks that report an earlier version (LibreSSL is `0x20000000L`) keep the current code. It also converts the RSA fallback to `EVP_RSA_gen()`, which had the same deprecated calls behind `PJMEDIA_SRTP_DTLS_USE_EC_KEY 0`.

I checked that nothing about the certificate changes: `EVP_EC_gen()` and the old `EC_KEY_*` path produce the same key id, size, SPKI length and AlgorithmIdentifier, so the named-curve encoding that `EC_KEY_set_asn1_flag()` was there to force is preserved. Both DTLS tests pass in either key configuration, warning-free.

On the SHA-1 point, could you tell me which line you changed and which version you're building? On master `X509_sign()` already uses `EVP_sha256()` — that was changed in 9e6a63e42 (#3876), which first shipped in 2.15 — and the fingerprint we put in our own SDP is hardcoded SHA-256. The only remaining `EVP_sha1()` is in `ssl_get_fingerprint()`, and it's used just to verify a *remote* certificate when the peer itself advertised `a=fingerprint:SHA-1`, so we can't drop it unilaterally. If you're seeing SHA-1 signing somewhere else I've missed, I'd like to fix it.

dev31337:
Sorry I'm using 2.14.1 and did not know about this commit which changed the X509_sign() to SHA 256.

Good to know it has been added since.

PR #5195, opened 2026-08-17
pjsua: fix stack buffer overflow when saving settings with long contact params (follow-up to #5163)

laconicman:
Follow-up to #5163, which was merged before this correction landed. Root cause tracked in #5196.

#5163 appends the account's `pj_str_t` values directly:

```c
        pj_strcat2(result, "--reg-contact-params ");
        pj_strcat(result, &acc_cfg->reg_contact_params);
```

`pj_strcat()` performs no capacity check, `write_settings()` ignores its `max`, and all four callers pass `char settings[2000]` — so a long value writes past the caller's frame when settings are saved. My fault: I changed it to that form during review of #5163 without checking the destination.

## Changes

Bounded `pj_ansi_snprintf` as the surrounding fields use, plus:

- **Newline appended separately.** Truncating through `line[128]` also drops the trailing `\n`, so the next option lands on the same physical line and the file reloads as one merged token:
  ```
  --reg-contact-params ;pn-prid=PtYg...kv--reg-contact-uri-params ;pn-provider=apns
  ```
- **Truncation marked and reported** — `PJ_CHECK_TRUNC_STR()` plus a warning naming the account and option, since for these two options truncation is the expected case and the result otherwise looks like a valid config.

## Verified

163-character token, current master: no crash, line bounded at 127 and terminated with `..`, the warning emitted, the following option intact on its own line, file reads back through `--config-file` without error.

## Not covered

Long values are still truncated. Refusing and reporting instead needs `write_settings()` to honour `max` — #5196. The merged-line flaw affects the neighbouring fields too; left for that work rather than widening this fix.

### DIFF
--- a/pjsip-apps/src/pjsua/pjsua_app_config.c
+++ b/pjsip-apps/src/pjsua/pjsua_app_config.c
@@ -1966,6 +1966,7 @@ pj_status_t load_config(int argc,
 static void write_account_settings(int acc_index, pj_str_t *result)
 {
     unsigned i;
+    int len;
     char line[128];
     pjsua_acc_config *acc_cfg = &app_config.acc_cfg[acc_index];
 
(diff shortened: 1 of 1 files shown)


nanangizz:
Verified the bug and the fix. The overflow fix is correct and should land, with one addition below.

The overflow was real: `write_settings()` discards its bound (`PJ_UNUSED_ARG(max)`, `pjsua_app_config.c:2240`) and points `cfg.ptr` at the caller's buffer (`:2242`), while all four callers pass `char settings[2000]`. With `pj_strcat()` doing no capacity check, an arbitrary length `reg_contact_params` wrote past the caller's frame. The bounded `snprintf` with `%.*s` matches the surrounding fields, and appending the newline separately is an improvement on them.

### 1. Please make the truncation visible here

Truncation is not a regression, every other field truncates through the same `line[128]`. But for these two options it is the expected case rather than an edge case: 128 less the 21 character prefix leaves 106 characters, and a realistic `;pn-provider=apns;pn-prid=<64+ chars>;pn-param=<bundle id>` exceeds that. Since the file is read back through `--config-file`, silent truncation yields a config that looks valid and misconfigures the account on next start. One line each:

```c
        PJ_CHECK_TRUNC_STR(len, line, sizeof(line));
```

(`string.h:80`, already used at `pjsua_app.c:1095`; it also covers platforms where snprintf returns -1.) A `PJ_LOG(2,(THIS_FILE, ...))` beside it is better still, since whoever saves the settings needs to know the token did not survive.

Marking is a stopgap, not the destination. The proper handling is refuse and report, and that contract already exists: every caller does `if (len < 1) PJ_LOG(1,(THIS_FILE, "Error: not enough buffer"))` (`pjsua_app_legacy.c:1780`, `:1793`), so honouring `max` needs no caller change. Mechanically: bounded appends instead of `pj_strcat2` into a fixed buffer, since `pj_ansi_strxcat()` returns -1 on truncation (`string.h:973`) and `pj_ansi_strxcpy2()` does the same for `pj_str_t` (`:954`), with long values formatted straight into the output rather than through a scratch. Worth recording as the target in #5196.

Also worth resisting the obvious follow-up of enlarging `line[]`: while `max` is ignored, a bigger scratch only lets one field consume more of the caller's 2000 bytes. Note the destination is already overflowable without any long value, since `write_account_settings()` has 42 appends and runs once per account up to `PJSUA_MAX_ACC`, on top of the global ones. That makes #5196 reachable in ordinary use rather than only with a long token, which is worth stating there.

### 2. The merged line flaw in the neighbouring fields

Worse than truncation, since the reloaded file yields a *different* value with nothing indicating a loss. Either here or immediately after, next to #5196.

Please also link #5196 from the description.

laconicman:
Thanks — all three done in e64a1583c.

`PJ_CHECK_TRUNC_STR()` plus a warning naming the account and option. Verified with a 163-character token: line bounded at 127 and marked `..`, warning emitted, following option intact, file reads back cleanly. Description now links #5196.

Left the merged-line flaw in the neighbouring fields for #5196 rather than widening this one, since it is the same change as honouring `max`. Recorded your target design there — refuse and report via `pj_ansi_strxcat()`/`pj_ansi_strxcpy2()`, no caller change needed, and not enlarging `line[]`.

sauwming (approved):

nanangizz (approved):

PR #5197, opened 2026-08-18
pjsip: add pjsip_endpt_send_request2() returning the UAC transaction

nanangizz:
`pjsip_endpt_send_request()` does not return any handle to the transaction it creates, so an application has no way to abandon an out-of-dialog request that it no longer cares about.

The use case comes from a user sending an OPTIONS to every address resolved for an endpoint: once one target replies, the endpoint is already known to be reachable, but the requests to the non-responding targets keep being retransmitted until they time out. There is no way to stop them, as CANCEL is not applicable to non-INVITE requests (RFC 3261 section 9.1), so the request can only be abandoned locally, which requires the transaction. Replicating `pjsip_endpt_send_request()` in the application just to get the transaction from `pjsip_tsx_create_uac()` is not always practical.

- Add `pjsip_endpt_send_request2()`, which optionally returns the transaction created to send the request, using the same `pjsip_transaction **p_tsx` output convention as the existing `pjsip_endpt_respond()`. Unlike `pjsip_endpt_respond()`, which hands back the pointer without adding any reference, the transaction is returned with its reference counter incremented, so the application must release it using `pj_grp_lock_dec_ref(tsx->grp_lock)` once it no longer needs the transaction. That reference is what makes the handle safe to keep across the asynchronous completion, which is the point of the API, and the header documents the difference.
- No new API is needed for the termination itself, application can simply use the existing `pjsip_tsx_terminate_async()`, which is safe to be called from any thread, including from within the callback.
- `pjsip_endpt_send_request()` is now a thin wrapper of the new function, its behavior is unchanged.

Note that the transaction is handed over only *after* the request has been sent successfully. The reference the function holds across the send keeps the transaction alive until then, even when the completion callback has already run, e.g. on receiving a response or an asynchronous transport error from another thread.

This ordering matters because `pjsip_tsx_send_msg()` can invoke the callback and only *then* return an error, via the synchronous-send paths in `sip_transaction.c` that end with `if (status==PJ_SUCCESS && tsx->state == PJSIP_TSX_STATE_TERMINATED) status = tsx->transport_err;`. Setting the output before the send would let the application observe a reference in its callback that the failure path then releases, so an application releasing it there would release it twice. The callback does not need the argument in any case, as the event already carries the transaction.

On every failure path, including the argument checks, `*p_tsx` is set to NULL and no reference is handed out.

## Test plan

- New `send_request2_test()` in pjsip test `tsx_basic_test`: the loop transport is set to discard the outgoing packet so the request will never be answered, then an OPTIONS is sent using the new API and terminated using `pjsip_tsx_terminate_async()`. The test verifies that the transaction is returned, that it does not complete by itself, that the callback is then called with `PJSIP_SC_REQUEST_TERMINATED`, and that the transaction is unregistered from the transaction layer while the returned reference keeps the instance alive.
- `pjsip-test`: `tsx_basic_test` (loop, udp, tcp), `regc_test`, `auth_async_test`, `pjsua_auth_test`, `pjsua_call_test`, `dlg_core_test`, `inv_offer_answer_test` and `transport_loop_test` passed. `regc_test` and the pjsua tests cover the wrapped `pjsip_endpt_send_request()` path (`sip_reg.c`, `pjsua_im.c`, `pjsua_acc.c`).

Co-Authored-By Claude Code

### DIFF
--- a/pjsip/include/pjsip/sip_util.h
+++ b/pjsip/include/pjsip/sip_util.h
@@ -794,6 +794,54 @@ PJ_DECL(pj_status_t) pjsip_endpt_send_request( pjsip_endpoint *endpt,
                                                pjsip_endpt_send_callback cb);
 
 /**
+ * Variant of #pjsip_endpt_send_request() which can also return the
+ * transaction created to send the request. Application may use the
+ * transaction, for example, to terminate the request before any response
+ * is received, e.g. to stop the retransmissions of an out-of-dialog request
+ * such as OPTIONS. Note that terminating the transaction only abandons the
+ * request locally, nothing is sent to the network, as CANCEL is not
+ * applicable to non-INVITE requests (see RFC 3261 section 9.1).
+ *
+ * To terminate the transaction, application should use
+ * #pjsip_tsx_terminate_async() or #pjsip_tsx_terminate_async2(), which are
+ * safe to be called from any thread, including from within the callback
+ * \a cb itself. Terminating an already completed transaction is harmless.
+ *
+ * @param endpt     The endpoint instance.
+ * @param tdata     The transmit data to be sent.
+ * @param timeout   Optional timeout for final response to be received, or -1
+ *                  if the transaction should not have a timeout restriction.
+ *                  The value is in miliseconds. Note that this is not
+ *                  implemented yet, so application needs to use its own timer
+ *                  to handle timeout.
+ * @param token     Optional token to be associated with the transaction, and
+ *                  to be passed to the callback.
+ * @param cb        Optional callback to be called when the transaction has
+ *                  received a final response. The callback will be called with
+ *                  the previously registered token and the event that triggers
+ *                  the completion of the transaction.
+ * @param p_tsx     Optional pointer to receive the transaction which was
+ *                  created to send the request. If it is not NULL, on
+ *                  success it will be set to the transaction with its
+ *                  reference counter incremented, so application must
+ *                  release it using pj_grp_lock_dec_ref(tsx->grp_lock) once
+ *                  it no longer needs the transaction, e.g. after the
+ *                  callback \a cb is called. On failure, it will be set to
+ *                  NULL. Note that the callback \a cb may already be called
+ *                  before this function returns, so application must be
+ *                  ready for the transaction to be already completed by the
+ *                  time it inspects this argument.
+ *
+ * @return          PJ_SUCCESS, or the appropriate error code.
+ */
+PJ_DECL(pj_status_t) pjsip_endpt_send_request2(pjsip_endpoint *endpt,
+                                               pjsip_tx_data *tdata,
+                                               pj_int32_t timeout,
+                                               void *token,
+                                               pjsip_endpt_send_callback cb,
+                                               pjsip_transaction **p_tsx);
+
+/**
  * @}
  */
 
--- a/pjsip/include/pjsip/sip_util.h
+++ b/pjsip/include/pjsip/sip_util.h
@@ -821,16 +821,17 @@ PJ_DECL(pj_status_t) pjsip_endpt_send_request( pjsip_endpoint *endpt,
  *                  the previously registered token and the event that triggers
  *                  the completion of the transaction.
  * @param p_tsx     Optional pointer to receive the transaction which was
- *                  created to send the request. If it is not NULL, on
- *                  success it will be set to the transaction with its
+ *                  created to send the request. On failure it will be set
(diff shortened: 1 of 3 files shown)


nanangizz:
Duplicate of #5198.

pjsip/src/pjsip/sip_util_statefull.c
```diff
@@ -100,6 +111,8 @@ PJ_DEF(pj_status_t) pjsip_endpt_send_request(  pjsip_endpoint *endpt,
 
     PJ_UNUSED_ARG(timeout);
 
+    if (p_tsx) *p_tsx = NULL;
```
sauwming: `pjsip_endpt_send_request2()` can fail its `PJ_ASSERT_RETURN` checks (invalid `endpt`/`tdata`/timeout, or module not yet registered) before this `*p_tsx = NULL;` runs. `PJ_ASSERT_RETURN` returns unconditionally on failure, and `pj_assert()` is a no-op in release builds, so `*p_tsx` is left uninitialized on those failure paths — contradicting the new header doc's "On failure it will be set to NULL" contract. Consider moving the `if (p_tsx) *p_tsx = NULL;` to the top of the function, before the argument checks.

pjsip/include/pjsip/sip_util.h
```diff
@@ -793,6 +793,55 @@ PJ_DECL(pj_status_t) pjsip_endpt_send_request( pjsip_endpoint *endpt,
...
+ *                  to be passed to the callback.
+ * @param cb        Optional callback to be called when the transaction has
+ *                  received a final response. The callback will be called with
+ *                  the previously registered token and the event that triggers
+ *                  the completion of the transaction.
+ * @param p_tsx     Optional pointer to receive the transaction which was
+ *                  created to send the request. On failure it will be set
+ *                  to NULL, on success to the transaction with its
+ *                  reference counter incremented, so application must
+ *                  release it using pj_grp_lock_dec_ref(tsx->grp_lock) once
```
sauwming, line 832: This documents that `p_tsx` hands back an extra grp_lock reference the caller must `pj_grp_lock_dec_ref()`. That's a different contract from the structurally identical `p_tsx` out-param on `pjsip_endpt_respond()` (same file), which assigns `*p_tsx` with no added reference and must never be dec_ref'd. The PR description says this function "follows the pattern of the existing pjsip_endpt_respond()", but the ref-counting behavior diverges — worth calling out explicitly (or aligning the two) so callers familiar with `pjsip_endpt_respond()` don't leak the transaction's grp_lock ref.

sauwming (commented):
Two merge-readiness issues found, see inline comments.

pjsip/src/pjsip/sip_util_statefull.c
nanangizz: Good catch, fixed in 4653d3c19 by moving the reset above both `PJ_ASSERT_RETURN()` checks, so `*p_tsx` is NULL on every failure path as documented.

For the record, `pjsip_endpt_respond()` has the same ordering (`PJ_ASSERT_RETURN(endpt && rdata, ...)` before `if (p_tsx) *p_tsx = NULL;`), it just does not promise anything about `p_tsx` in its doc, so nothing there contradicts itself. I left it alone.

pjsip/include/pjsip/sip_util.h
nanangizz: You are right that the two contracts differ, and the PR description overstated the similarity. `pjsip_endpt_respond()` adds a reference before its send and drops it before returning, so `*p_tsx` comes back with no reference and must never be dec_ref'd.

I kept the added reference here rather than aligning with `pjsip_endpt_respond()`: the whole point of this API is to hold the transaction across the asynchronous completion so it can be terminated later, and an unreferenced pointer would be exactly the dangling handle we are trying to avoid. `pjsip_endpt_respond()` gets away with it because callers use the pointer immediately.

So I documented the difference instead, in 4653d3c19, and reworded the PR description. The header now says the reference is added unlike in `pjsip_endpt_respond()`, and that application must release it.

Note that 54e3c78c5 (pushed before your review, in case you were looking at the first commit) did align the part of `pjsip_endpt_respond()` worth matching: `*p_tsx` is now assigned only after a successful send, as `pjsip_endpt_respond()` does in its `} else if (p_tsx) { *p_tsx = tsx; }`. Previously the output was set before the send, but `pjsip_tsx_send_msg()` can invoke the callback and only then return an error, so an application releasing the reference from within the callback could have released it twice.

sauwming (approved):

PR #5198, opened 2026-08-18
pjsip: add pjsip_endpt_send_request2()/pjsip_endpt_cancel_request()

trengginas:
## Summary

- Applications sending an out-of-dialog non-INVITE request (e.g. OPTIONS) via `pjsip_endpt_send_request()` had no way to locally abandon the request before a response arrives — the function returns no transaction handle, and per RFC 3261 §9.1 a wire-level CANCEL doesn't apply to non-INVITE requests anyway, so there was no supported mechanism at all.
- Adds `pjsip_endpt_send_request2()`, a variant of `pjsip_endpt_send_request()` that optionally returns the transaction's *key* (not the transaction object itself) into a caller-supplied buffer. The caller passes the buffer's actual capacity in via `p_tsx_key->slen`, and the function bounds-checks against that real value rather than assuming a fixed size — consistent with the existing `pjsip_uri_print()`/`pjsip_hdr_print_on()`/`pjsip_msg_print()` convention for output buffers in this codebase.
- Returning a key instead of the transaction pointer avoids any reference-counting or use-after-free hazard for the caller — the key is an inert value with no lifetime/ownership contract beyond "keep the buffer valid until your callback fires."
- Adds `pjsip_endpt_cancel_request(endpt, tsx_key, code)`, which looks the transaction up by that key (`pjsip_tsx_layer_find_tsx2()`) and terminates it via `pjsip_tsx_terminate_async()` (not the synchronous `pjsip_tsx_terminate()`), so it is safe to call from anywhere — including from within the transaction's own completion callback — without risking reentrant re-entry into the transaction's state-transition notification. If the transaction has already completed, it simply returns `PJ_ENOTFOUND`.
- `pjsip_endpt_send_request()` is now a thin wrapper delegating to `pjsip_endpt_send_request2()` with no key requested, so all existing callers are unaffected.
- Adds `PJSIP_TSX_KEY_MAX_LEN` as a suggested (not required/assumed) buffer size for callers who want a size that comfortably fits any standard SIP method's key.

### DIFF
--- a/pjsip/include/pjsip/sip_util.h
+++ b/pjsip/include/pjsip/sip_util.h
@@ -794,6 +794,102 @@ PJ_DECL(pj_status_t) pjsip_endpt_send_request( pjsip_endpoint *endpt,
                                                pjsip_endpt_send_callback cb);
 
 /**
+ * Suggested buffer size, in bytes, comfortably covering the key returned
+ * by #pjsip_endpt_send_request2() for any standard SIP method: 1 byte for
+ * the role character, 1 byte for the separator, up to 31 bytes for the
+ * method name, plus #PJSIP_MAX_BRANCH_LEN for the branch parameter. This
+ * is only a suggestion for sizing the caller's buffer -- the function
+ * itself bounds-checks against whatever capacity the caller actually
+ * supplies (see \a p_tsx_key below), it does not assume this size.
+ */
+#define PJSIP_TSX_KEY_MAX_LEN   (PJSIP_MAX_BRANCH_LEN + 32)
+
+/**
+ * Variant of #pjsip_endpt_send_request() that also returns the key of the
+ * transaction created to send the request. The application can save this
+ * key and use it later, at any time, to call #pjsip_endpt_cancel_request()
+ * to locally abandon the request (e.g. to stop retransmissions of an
+ * out-of-dialog request such as OPTIONS) before a response arrives.
+ *
+ * Unlike returning the transaction object itself, this key is a plain,
+ * inert value with no reference-counting or use-after-free hazard: the
+ * application does not need to add or release any reference. It only
+ * needs to keep the buffer backing \a p_tsx_key valid for as long as it
+ * may still want to call #pjsip_endpt_cancel_request() -- in practice,
+ * until its \a cb has been called for this transaction. Calling
+ * #pjsip_endpt_cancel_request() with a key whose transaction has already
+ * completed is safe and simply returns PJ_ENOTFOUND.
+ *
+ * @param endpt     The endpoint instance.
+ * @param tdata     The transmit data to be sent.
+ * @param timeout   Optional timeout for final response to be received, or -1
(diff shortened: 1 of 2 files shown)


nanangizz:
Oops. I did the code review on the original patch (from email attachment), found some issues, e.g: working with tsx key string directly is not a common pattern in pjsip API, etc. So I modified the patch and created #5197. My proposal is to merge the update here and do the review here (it is handled by the original author here), i will close PR #5197.

nanangizz (commented):
Thanks Riza. A few points, mostly on the key output argument.

**1. `PJSIP_TSX_KEY_MAX_LEN` is not a compile-time constant.** It expands through `PJSIP_MAX_BRANCH_LEN` (`sip_config.h:1181`) to `PJSIP_RFC3261_BRANCH_LEN + pj_GUID_STRING_LENGTH() + 2`, and `pj_GUID_STRING_LENGTH()` is a function (`pj/guid.h:65`). So `char key_buf[PJSIP_TSX_KEY_MAX_LEN];` as "a field embedded in the caller's own session/call struct" — what the doc recommends — does not compile. A `PJSIP_TSX_*` constant would also fit better in `sip_transaction.h`.

**2. `pj_str_t` as buffer+capacity.** `pj_str_t` is (ptr, len) everywhere else, and the convention cited in the PR description is actually `char *buf, pj_size_t size` (`pjsip_msg_print()`, `pjsip_hdr_print_on()`). Suggest matching that literally: `char *key_buf, pj_size_t key_size`.

**3. The key is copied after the send, which races the callback.** Once `pjsip_tsx_send_msg()` succeeds, the completion callback may already be running on another thread (response, or async transport error). If the application frees the struct holding the buffer there, the `pj_memcpy()` writes into freed memory. The key is ready right after `pjsip_tsx_create_uac()` (`create_tsx_key_3261()`, `sip_transaction.c:1660`), so it can be copied before sending and cleared if the send fails.

**4. Silent no-key path.** The key is `role + separator + method + '$' + branch`, and method name length is unbounded for extension methods. If it does not fit, the request is still sent but the caller gets `slen = 0` and can never cancel it — the exact situation this API exists to prevent. Better to fail the call than to send an uncancellable request.

**5. Minor.** `pjsip_endpt_cancel_request()` reads like it sends a CANCEL; something like `pjsip_endpt_terminate_request()` would be clearer, and `endpt` is unused (`PJ_UNUSED_ARG`). Also no test — one option is sending an OPTIONS over the loop transport with `pjsip_loop_set_discard()` so nothing can answer it, then cancelling and asserting the callback fires with 487.

On the shape itself: #5197 returned the transaction (ref-counted) instead of a key, matching `pjsip_endpt_respond()` in this same header, and dropped `pjsip_endpt_cancel_request()` entirely since `pjsip_tsx_terminate_async()` is already public. No public PJSIP API takes or returns a transaction key today, so I'd rather not start here.

sauwming:
Abandoned by author.

@nanangizz, just use #5197 then, so we can proceed.

PR #5205, opened 2026-08-19
Fix orphaned recycle-list node in call_media_on_event()

### LINKED ISSUE #5204: [pjsua-lib] call_media_on_event() orphans a recycle-list node when the call is hanging up
### Describe the bug

`call_media_on_event()` in `pjsip/src/pjsua-lib/pjsua_media.c` takes a `pjsua_event_list` node off
the recycle free list, and *then* returns early when the call is hanging up — without putting the
node back:

```c
    if (pjsua_var.ua_cfg.cb.on_call_media_event) {
        pjsua_event_list *eve = NULL;

        pj_mutex_lock(pjsua_var.timer_mutex);

        if (pj_list_empty(&pjsua_var.event_list)) {
            eve = PJ_POOL_ALLOC_T(pjsua_var.timer_pool, pjsua_event_list);
        } else {
            eve = pjsua_var.event_list.next;
            pj_list_erase(eve);                     /* <-- taken off the free list */
        }

        pj_mutex_unlock(pjsua_var.timer_mutex);

        if (call->hanging_up)
            return status;                          /* <-- and never put back */

        eve->call_id = call->index;
        eve->med_idx = call_med->idx;
        pj_memcpy(&eve->event, event, sizeof(pjmedia_event));
        pjsua_schedule_timer2(&call_med_event_cb, eve, 1);
    }
```

A node is returned to `pjsua_var.event_list` in exactly one place — `call_med_event_cb()`:

```c
    pj_mutex_lock(pjsua_var.timer_mutex);
    pj_list_push_back(&pjsua_var.event_list, eve);
    pj_mutex_unlock(pjsua_var.timer_mutex);
```

…and that callback only runs if `pjsua_schedule_timer2()` was reached. Since the early return sits
*after* the acquisition and *before* the scheduling, every media event dropped during hangup costs
one node permanently:

- free list non-empty → the node is erased and orphaned; the list shrinks by one;
- free list empty → a node is allocated from `timer_pool` and immediately orphaned; the pool grows.

This is not a heap leak — `timer_pool` is released at shutdown — but it is unbounded growth over the
lifetime of a long-running process, proportional to the number of media events that arrive while
calls are tearing down.

### Suggested fix

Hoist the guard above the acquisition, which also matches how the other `hanging_up` checks in
pjsua guard the callback itself rather than its bookkeeping:

```diff
-    if (pjsua_var.ua_cfg.cb.on_call_media_event) {
+    if (pjsua_var.ua_cfg.cb.on_call_media_event && !call->hanging_up) {
         pjsua_event_list *eve = NULL;
@@
         pj_mutex_unlock(pjsua_var.timer_mutex);
 
-        if (call->hanging_up)
-            return status;
-
         eve->call_id = call->index;
```

### Steps to reproduce

Found by inspection rather than by symptom, and worth saying so plainly: the effect is slow,
bounded by process lifetime, and invisible unless you are watching the free list. To observe it,
install `on_call_media_event`, drive calls that raise media events around teardown, and instrument
the length of `pjsua_var.event_list` (or the size of `timer_pool`) across many call cycles — the
list drains monotonically and never recovers.

### PJSIP version

current master (`288de6142`)

### Context

Noticed while wiring `on_call_media_event` into a Swift pjsua1 wrapper. No local workaround is
needed or possible from the application side — the bookkeeping is entirely internal.

laconicman:
Fixes #5204.

`call_media_on_event()` takes a `pjsua_event_list` node off the recycle free list and then returns
early when the call is hanging up, without putting the node back. Nodes are returned in exactly one
place — `call_med_event_cb()` — which only runs if `pjsua_schedule_timer2()` was reached, so every
media event dropped during hangup orphans one node: the free list drains, or `timer_pool` grows
when the list was already empty.

The fix moves the `hanging_up` check into the enclosing condition, so the node is never acquired
when it will not be used:

```diff
-    if (pjsua_var.ua_cfg.cb.on_call_media_event) {
+    if (pjsua_var.ua_cfg.cb.on_call_media_event && !call->hanging_up) {
@@
-        if (call->hanging_up)
-            return status;
-
```

This also matches the other `hanging_up` guards in pjsua, which gate the callback itself rather
than its bookkeeping.

**Safety of the moved read:** `call` is dereferenced unconditionally at the top of the function
(`call->index` in the entry `PJ_LOG`), so reading `call->hanging_up` earlier cannot introduce a
null dereference. `&&` short-circuits, so it is still not read when no callback is installed.

**Behaviour:** unchanged for callers — the event was already being dropped in this case; only the
bookkeeping differs.

### DIFF
--- a/pjsip/src/pjsua-lib/pjsua_media.c
+++ b/pjsip/src/pjsua-lib/pjsua_media.c
@@ -1919,7 +1919,7 @@ pj_status_t call_media_on_event(pjmedia_event *event,
             break;
     }
 
-    if (pjsua_var.ua_cfg.cb.on_call_media_event) {
+    if (pjsua_var.ua_cfg.cb.on_call_media_event && !call->hanging_up) {
         pjsua_event_list *eve = NULL;
  
         pj_mutex_lock(pjsua_var.timer_mutex);
@@ -1932,9 +1932,6 @@ pj_status_t call_media_on_event(pjmedia_event *event,
         }
 
         pj_mutex_unlock(pjsua_var.timer_mutex);
-
-        if (call->hanging_up)
-            return status;
 
         eve->call_id = call->index;
         eve->med_idx = call_med->idx;


sauwming (approved):

nanangizz (approved):

laconicman:
One correction to the impact, in the direction of "worse than I wrote": this is not conditional on
an application opting in.

`Endpoint::libInit()` installs the callback unconditionally — `pjsip/src/pjsua2/endpoint.cpp:2374`:

```cpp
    ua_cfg.cb.on_call_media_event       = &Endpoint::on_call_media_event;
```

There is no conditional between the start of `libInit()` (`:2311`) and that assignment, so
`pjsua_var.ua_cfg.cb.on_call_media_event` is non-NULL for **every** pjsua2 application from
`libInit()` onward. The guard in `call_media_on_event()` is therefore always taken, and the node is
always acquired — and orphaned whenever the call is hanging up.

Nor does it depend on the application overriding anything: `Endpoint::on_call_media_event`
(`:2140`) unconditionally allocates a `PendingOnMediaEventCallback` and queues it, whether or not
`Call::onCallMediaEvent()` has been overridden.

So for C++ applications there is no opt-out and nothing to configure: every media event that
arrives during teardown costs one node, permanently. For C applications it remains conditional on
installing the callback, which is how I found it.

Doesn't change the patch — just moves it from latent to live for the whole pjsua2 surface.

PR #5208, opened 2026-08-22
pjsua: honour the settings buffer bound in write_settings()

### LINKED ISSUE #5196: pjsua: write_settings() ignores its max argument — unbounded appends into the caller's fixed buffer
### Describe the bug

`write_settings()` (`pjsip-apps/src/pjsua/pjsua_app_config.c`) accepts a destination capacity and discards it:

```c
int write_settings(pjsua_app_config *config, char *buf, pj_size_t max)
{
    ...
    PJ_UNUSED_ARG(max);

    cfg.ptr = buf;
    cfg.slen = 0;
```

It then performs 139 `pj_strcat()` / `pj_strcat2()` appends into `cfg`, none of which check remaining capacity, and finishes with

```c
    *(cfg.ptr + cfg.slen) = '\0';
```

which writes one byte past whatever was accumulated, also unchecked.

Every caller passes a fixed stack buffer of 2000 bytes — `pjsua_app_legacy.c` (twice) and `pjsua_app_cli.c` (twice). Nothing anywhere bounds the total.

The per-field `line[128]` limits each field's *contribution*, not the sum, so the protection is incidental. With `PJSUA_MAX_ACC` accounts each emitting id, registrar, proxies, credentials and Contact parameters, 2000 bytes is reachable with ordinary configuration, and the write runs off the end of the caller's stack frame.

There is also at least one field that bypasses `line[]` entirely and appends a user-supplied value directly:

```c
        pj_strcat(&cfg, &escaped);      /* --custom-sdp */
```

### Steps to reproduce

Configure several accounts with long-ish values (registrar URI, proxy, credentials, Contact parameters) and save the settings — `f` in the legacy UI, or the CLI equivalent. The output is written into `char settings[2000]` with no bound.

### PJSIP version

Current `master` (`7e95d9f70`).

### Context

macOS, GNU make build. Found after introducing an unbounded append of my own in #5163; that instance is fixed in #5195, but the surrounding function has the same shape throughout, so the class of bug is not specific to those two options.

### Log, call stack, etc

```shell
Not reproduced as a crash here: the trigger is configuration size rather than a
single value, and I did not want to publish a tuned overflow recipe. The
evidence is the discarded max argument, the 139 unchecked appends and the
unchecked terminator, all quoted above.
```

---

### Suggested fix

Thread the capacity through: a small bounded append helper that takes remaining space and truncates, used in place of the direct `pj_strcat`/`pj_strcat2` calls, with `max` passed into `write_account_settings()` (static, so the signature is free to change) and the terminating write guarded. Mechanical but broad.

Happy to take it — say the word, or leave it to me if you would rather shape it differently.

laconicman:
## Description

Makes `write_settings()` respect its `max` argument, as agreed in #5196.

Two bounded helpers replace the 139 direct appends. Both stop at `max` and mark the buffer full by setting `slen` negative, so a partial line is never emitted and the function returns an error the callers already handle. Values are formatted straight into the output instead of through a scratch buffer, so a long one no longer depends on `line[128]` — which stays at 128 and is now unused in these two functions. `--custom-sdp`'s `pj_strcat()` gets the same treatment, the terminator is inside the bound, and capacity is threaded through `write_account_settings()`.

The merged-line flaw is fixed uniformly rather than field by field as in #5195: a value too long for the scratch buffer used to swallow its own newline, running the next option onto the same physical line, so the file reloaded as a different configuration with nothing indicating a loss.

Honouring `max` on its own would turn a save that currently runs off the caller's frame into a refusal, so the callers now take the buffer from a temporary pool created and released around each call. Not a larger fixed buffer — sizing for the worst case means tens of KB of stack at four sites, and any fixed number goes stale as options are added. Not `app_config.pool` either, since it is released only at teardown.

## Motivation and Context

Fixes #5196. The per-field `line[128]` bounded each contribution but never the total, and `write_account_settings()` has 42 appends running once per account up to `PJSUA_MAX_ACC`, so the account section alone can reach tens of KB against `char settings[2000]`.

Sample application only — no library links `pjsua_app_config.o`, and there is no remote vector.

## How Has This Been Tested?

macOS, GNU make, clean build with no new warnings.

- 163-character `--reg-contact-params` value: written in full (172 characters) and byte-identical to the input, where it was previously truncated at the scratch buffer. Saved file reloads through `--config-file` with no argument errors.
- No line in the saved file carries two options.
- Bound exercised by temporarily setting `PJSUA_APP_SETTINGS_SIZE` to 300: the save is refused with "Error: not enough buffer", no partial file is written and nothing crashes.
- `scripts-sipp/uas-auth` passes.

## Types of changes

## Checklist:

### DIFF
--- a/pjsip-apps/src/pjsua/pjsua_app_legacy.c
+++ b/pjsip-apps/src/pjsua/pjsua_app_legacy.c
@@ -1773,23 +1773,45 @@ static void ui_dump_call_quality()
 
 static void ui_dump_configuration()
 {
-    char settings[2000];
+    pj_pool_t *pool;
+    char *settings;
     int len;
 
-    len = write_settings(&app_config, settings, sizeof(settings));
+    pool = pjsua_pool_create("settings", PJSUA_APP_SETTINGS_SIZE + 1024,
+                             1024);
+    if (!pool) {
+        PJ_LOG(1,(THIS_FILE, "Error: unable to allocate settings buffer"));
+        pj_pool_release(pool);
+        return;
+    }
+    settings = (char*)pj_pool_alloc(pool, PJSUA_APP_SETTINGS_SIZE);
+
+    len = write_settings(&app_config, settings, PJSUA_APP_SETTINGS_SIZE);
     if (len < 1)
         PJ_LOG(1,(THIS_FILE, "Error: not enough buffer"));
     else
         PJ_LOG(3,(THIS_FILE, "Dumping configuration (%d bytes):\n%s\n",
                   len, settings));
+
+    pj_pool_release(pool);
 }
 
 static void ui_write_settings(const char *filename)
 {
-    char settings[2000];
+    pj_pool_t *pool;
+    char *settings;
     int len;
 
-    len = write_settings(&app_config, settings, sizeof(settings));
+    pool = pjsua_pool_create("settings", PJSUA_APP_SETTINGS_SIZE + 1024,
+                             1024);
+    if (!pool) {
+        PJ_LOG(1,(THIS_FILE, "Error: unable to allocate settings buffer"));
+        pj_pool_release(pool);
+        return;
+    }
+    settings = (char*)pj_pool_alloc(pool, PJSUA_APP_SETTINGS_SIZE);
+
+    len = write_settings(&app_config, settings, PJSUA_APP_SETTINGS_SIZE);
     if (len < 1)
         PJ_LOG(1,(THIS_FILE, "Error: not enough buffer"));
     else {
@@ -1807,6 +1829,8 @@ static void ui_write_settings(const char *filename)
             printf("Settings successfully written to '%s'\n", filename);
         }
     }
+
+    pj_pool_release(pool);
 }
 
 /*
--- a/pjsip-apps/src/pjsua/pjsua_app_legacy.c
+++ b/pjsip-apps/src/pjsua/pjsua_app_legacy.c
@@ -1792,7 +1792,7 @@ static void ui_dump_configuration()
         PJ_LOG(3,(THIS_FILE, "Dumping configuration (%d bytes):\n%s\n",
                   len, settings));
 
-    pj_pool_release(pool);
+    pj_pool_secure_release(&pool);
 }
 
 static void ui_write_settings(const char *filename)
@@ -1828,7 +1828,7 @@ static void ui_write_settings(const char *filename)
         }
     }
 
-    pj_pool_release(pool);
+    pj_pool_secure_release(&pool);
 }
 
 /*
--- a/pjsip-apps/src/pjsua/pjsua_app_legacy.c
+++ b/pjsip-apps/src/pjsua/pjsua_app_legacy.c
@@ -1781,7 +1781,6 @@ static void ui_dump_configuration()
                              1024);
     if (!pool) {
         PJ_LOG(1,(THIS_FILE, "Error: unable to allocate settings buffer"));
-        pj_pool_release(pool);
         return;
     }
     settings = (char*)pj_pool_alloc(pool, PJSUA_APP_SETTINGS_SIZE);
@@ -1806,7 +1805,6 @@ static void ui_write_settings(const char *filename)
                              1024);
     if (!pool) {
         PJ_LOG(1,(THIS_FILE, "Error: unable to allocate settings buffer"));
-        pj_pool_release(pool);
         return;
     }
     settings = (char*)pj_pool_alloc(pool, PJSUA_APP_SETTINGS_SIZE);
--- a/pjsip-apps/src/pjsua/pjsua_app_cli.c
+++ b/pjsip-apps/src/pjsua/pjsua_app_cli.c
@@ -1858,7 +1858,7 @@ static pj_status_t cmd_show_config()
                   "Dumping configuration (%d bytes):\n%s\n",
                   len, settings));
 
-    pj_pool_release(pool);
+    pj_pool_secure_release(&pool);
     return PJ_SUCCESS;
 }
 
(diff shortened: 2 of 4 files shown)


pjsip-apps/src/pjsua/pjsua_app_legacy.c
```diff
@@ -1773,23 +1773,45 @@ static void ui_dump_call_quality()
...
+    pj_pool_t *pool;
+    char *settings;
     int len;
 
-    len = write_settings(&app_config, settings, sizeof(settings));
+    pool = pjsua_pool_create("settings", PJSUA_APP_SETTINGS_SIZE + 1024,
+                             1024);
+    if (!pool) {
+        PJ_LOG(1,(THIS_FILE, "Error: unable to allocate settings buffer"));
+        pj_pool_release(pool);
```
sauwming: `pool` is NULL here (the `!pool` branch above). `pj_pool_release()` unconditionally dereferences `pool->factory`, so this crashes instead of failing gracefully. `cmd_show_config()` in pjsua_app_cli.c handles the same case correctly by just returning — this looks like a copy-paste regression. Suggest dropping this line.

sauwming (commented):
## Review

Traced the new `cfg_add`/`cfg_addf` bounded-append helpers against `pj_ansi_strxcat`/`vsnprintf` semantics, and checked `pj_pool_release()`'s implementation. Overall the fix for the buffer-bound issue is solid, but there's one crash bug that should be fixed before merge.

### Blocking

- `pjsua_app_legacy.c`: `ui_dump_configuration()` and `ui_write_settings()` call `pj_pool_release(pool)` on the OOM path where `pool` is `NULL` (see inline comments). `pj_pool_release()` unconditionally dereferences `pool->factory` (and `pool->block_list` if `PJ_POOL_RELEASE_WIPE_DATA` is set), so this crashes instead of failing gracefully. The sibling call sites in `pjsua_app_cli.c` (`cmd_show_config`, `cmd_write_config`) got this right by just returning on `!pool` — these two look like a copy-paste regression. Simplest fix: drop the `pj_pool_release(pool);` line in both spots.

### Non-blocking (worth a look, not required for merge)

- `pj_pool_alloc(pool, PJSUA_APP_SETTINGS_SIZE)`'s return value is never checked at any of the 4 call sites before being passed to `write_settings()`. Unlikely to fail given the margin, but there's no guard.
- The 4 pool-create/alloc/error-check blocks are duplicated near-verbatim across `pjsua_app_cli.c` and `pjsua_app_legacy.c` — a shared helper would remove the duplication that let the NULL-release bug diverge between copies in the first place.
- `cfg_add()` re-derives the append offset via `pj_ansi_strxcat()`'s internal `pj_ansi_strlen(dst)` on every call instead of using the already-tracked `cfg->slen`, making the ~100+ calls per `write_settings()` invocation effectively quadratic in buffer size. `cfg_addf()` right next to it already appends at `cfg->ptr + cfg->slen` directly.
- `--custom-sdp` is now appended via `cfg_addf("%.*s", ...)` (vsnprintf-based) instead of the previous `pj_strcat()`. If the escaped SDP string ever contains an embedded NUL, `%.*s` stops at it (per C `%s` semantics) while the old length-based `pj_strcat()` copied all bytes — silent truncation with `write_settings()` still reporting success.
- A few literal, no-argument strings (`"--stereo\n"`, `"--auto-play-avi\n"`, `"--rec-avi-audio\n"`, `"--auto-rec-avi\n"`) go through `cfg_addf()` instead of `cfg_add()`, inconsistent with every other literal in the file and paying for an unnecessary vsnprintf parse.

pjsip-apps/src/pjsua/pjsua_app_legacy.c
```diff
@@ -1773,23 +1773,45 @@ static void ui_dump_call_quality()
...
```
sauwming: Same issue as ui_dump_configuration() above: `pool` is NULL here, and `pj_pool_release(NULL)` will crash. Suggest dropping this line.

laconicman:
Fixed in 8ef00da7d — both lines dropped.

Also switched the success-path releases to `pj_pool_secure_release()` in 4dd6756a2. `write_settings()` emits `--password` and `--tls-password`, `PJ_POOL_RELEASE_WIPE_DATA` defaults to 0, and `cpool_release_pool()` recycles via `pj_pool_reset()`, which only resets the allocation pointers — so the credentials would sit in the free list for a later `pj_pool_alloc()`. Moving the buffer from the stack into a pool is what made that reachable, so it seemed worth closing here.

On the `!pool` guard: `cpool_create_pool()` has a `return NULL`, but the OOM path calls the policy callback, which throws, so I could not convince myself the guard is reachable in a default build. `pj_pool_secure_release()` tolerates NULL anyway. Drop the guard, or keep it as cheap insurance for a custom factory policy?

sauwming:
Keep the guard — it is reachable in a default build.

`pjsua_pool_create()` → `pj_pool_create()` → (caching pool factory) `cpool_create_pool()` → `pj_pool_create_int()`. Look at `pj_pool_create_int()` itself (pjlib/src/pj/pool.c:224-226):

```c
buffer = (pj_uint8_t*) (*f->policy.block_alloc)(f, initial_size);
if (!buffer)
    return NULL;
```

That first block allocation is a raw `malloc()` (`default_block_alloc`) called *before* the pool object exists, so there's no `pj_pool_t` yet to hold a callback on. The exception-throwing callback (`default_pool_callback` → `PJ_THROW(PJ_NO_MEMORY_EXCEPTION)`) only fires for allocations *after* the pool is constructed — i.e. later `pj_pool_alloc()` growth. So a plain `malloc()` failure for the initial ~settings-buffer-sized block under memory pressure returns `NULL` straight through, with no exception in the mix. The guard is real insurance, not dead code — please keep it.

sauwming (approved):

laconicman:
Keeping the guard — thanks for the `pj_pool_create_int()` trace, that settles it.

The five non-blocking items are in #5213 rather than pushed here, so this approval stands. One note from that work: the length-based helper for `--custom-sdp` has to sit under `#if !PJSUA_MEDIA_HAS_PJMEDIA` with its caller, otherwise it is an unused-function warning in a default `-Wall` build.

nanangizz (approved):

PR #5211, opened 2026-08-25
fix(pjsip): answer an oversized request with 513 Message Too Large

jwes:
A stream message is assembled in rdata->pkt_info.packet, a fixed PJSIP_MAX_PKT_LEN buffer. A request larger than that can never be completed, so pjsip_tpmgr_receive_packet() reports PJSIP_ERXOVERFLOW and discards it. endpt_on_rx_msg() only logged that, so nothing went back on the wire and the sender could not tell "too large" from a lost connection. RFC 3261 section 21.5.7 defines 513 for this case, yet the status code had no sender anywhere in the stack.

Re-parse the truncated buffer up to the blank line ending the header block, restore the received/rport Via params the transport layer would have added, and reply via pjsip_endpt_respond_stateless(). A Warning: 399 header names the byte limit so the peer learns what to stay under.

Only an overflowing body is recoverable: a request whose header block alone exceeds the buffer has no Via to address a response to and is still dropped, as is an oversized response. The overflow branch is stream-only, so this adds no reflection vector for datagram transports.

transport_rx_overflow_test covers it, writing the request from a plain socket because a pjsip_tx_data buffer is itself capped at PJSIP_MAX_PKT_LEN.

### DIFF
--- a/pjsip/src/pjsip/sip_endpoint.c
+++ b/pjsip/src/pjsip/sip_endpoint.c
@@ -1003,10 +1003,10 @@ on_return:
 static void endpt_respond_msg_too_large( pjsip_endpoint *endpt,
                                          pjsip_rx_data *rdata )
 {
-    char *buf = rdata->msg_info.msg_buf;
-    pj_size_t len = (pj_size_t)rdata->msg_info.len;
-    pj_size_t hdr_len = 0;
-    pj_size_t i;
+    const pj_str_t end_hdr = { "\n\r\n", 3 };
+    pj_str_t received;
+    pj_size_t hdr_len;
+    char *pos;
     char *hdr_buf;
     pjsip_msg *msg;
     pjsip_warning_hdr *warn;
@@ -1015,32 +1015,25 @@ static void endpt_respond_msg_too_large( pjsip_endpoint *endpt,
     pj_str_t warn_str;
     pj_status_t status;
 
-    if (buf == NULL || len == 0)
+    received.ptr = rdata->msg_info.msg_buf;
+    received.slen = rdata->msg_info.len;
+    if (received.ptr == NULL || received.slen <= 0)
         return;
 
-    /* Locate the empty line that ends the header block, and keep it: the
-     * parser needs it to see that the headers are complete.
+    /* Locate the empty line that ends the header block, with the sentinel
+     * pjsip_find_msg() uses, so the two cannot disagree on where the headers
+     * end. Keep the line: the parser needs it to see that they are complete.
      */
-    for (i = 0; i + 1 < len; ++i) {
-        if (buf[i] == '\r' && i + 3 < len && buf[i+1] == '\n' &&
-            buf[i+2] == '\r' && buf[i+3] == '\n')
-        {
-            hdr_len = i + 4;
-            break;
-        }
-        if (buf[i] == '\n' && buf[i+1] == '\n') {
-            hdr_len = i + 2;
-            break;
-        }
-    }
-    if (hdr_len == 0)
+    pos = pj_strstr(&received, &end_hdr);
+    if (pos == NULL)
         return;
+    hdr_len = (pj_size_t)(pos + 3 - received.ptr);
 
     /* Parse from a copy: the scanner writes into the buffer it is given, and
      * the transport still hands the original on to its dropped-data callback.
      */
     hdr_buf = (char*) pj_pool_alloc(rdata->tp_info.pool, hdr_len + 1);
-    pj_memcpy(hdr_buf, buf, hdr_len);
+    pj_memcpy(hdr_buf, received.ptr, hdr_len);
     hdr_buf[hdr_len] = '\0';
 
     pj_bzero(&rdata->msg_info, sizeof(rdata->msg_info));
--- a/pjsip/src/pjsip/sip_endpoint.c
+++ b/pjsip/src/pjsip/sip_endpoint.c
@@ -996,7 +996,7 @@ on_return:
  * msg_info and the sender is left waiting for a response that never comes.
  * It is the body that overflows in practice while the header block still fits
  * the buffer, so parse the headers back into rdata and answer through the
- * normal stateless path - RFC 3261 section 21.4.11 defines 513 for exactly
+ * normal stateless path - RFC 3261 section 21.5.7 defines 513 for exactly
  * this. If not even the headers arrived complete there is nobody we can
  * address a response to, and the request stays silently dropped.
  */
--- a/pjsip/src/pjsip/sip_endpoint.c
+++ b/pjsip/src/pjsip/sip_endpoint.c
@@ -1069,8 +1069,7 @@ static void endpt_respond_msg_too_large( pjsip_endpoint *endpt,
 
     pj_list_init(&hdr_list);
     warn = pjsip_warning_hdr_create(rdata->tp_info.pool, 399,
-                                    &rdata->tp_info.transport->local_name.host,
-                                    &warn_str);
+                                    pjsip_endpt_name(endpt), &warn_str);
     if (warn)
         pj_list_push_back(&hdr_list, (pjsip_hdr*)warn);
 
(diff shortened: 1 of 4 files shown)


sauwming (commented):
Reviewed the diff and traced it against the actual transport/parser code (not just reading in isolation):

- Confirmed `rdata->msg_info.msg_buf`/`len` are in fact populated by `pjsip_tpmgr_receive_packet()` before `on_rx_msg()` is invoked with `PJSIP_ERXOVERFLOW` (sip_transport.c:2229-2232), so `endpt_respond_msg_too_large()` isn't reading stale/empty data despite the function comment's "empty msg_info" phrasing being a bit imprecise.
- The manual header-block scan (`\r\n\r\n` / `\n\n`) has no OOB read — the `i+3 < len` bound is checked before the corresponding byte accesses.
- The `recvd_param`/`rport_param` backfill exactly mirrors what the transport manager does for every normal request (sip_transport.c:2421-2433) — needed because `pjsip_get_response_addr()` asserts `recvd_param.slen != 0`, and this path bypasses the normal transport flow.
- `rdata->tp_info.pool` is reset right after the read callback returns (sip_transport_tcp.c:1580) regardless of this new branch, so no pool growth/leak on repeated oversized messages on a long-lived connection.
- `pjsip_warning_hdr_create()` / `pjsip_endpt_respond_stateless()` call signatures match current headers.
- Compiled `sip_endpoint.c` and `transport_tcp_test.c` standalone with `-Wall -Wextra`: zero warnings, zero errors. Full link/run wasn't possible in my sandbox due to an unrelated local ffmpeg/libswscale version mismatch in `pjmedia`, not caused by this PR.
- CI: all checks that have completed so far are green (including asan/tsan stress and fuzzing); remaining jobs are still pending/queued, none failing.

No remaining issues found. Looks ready to merge once the rest of CI finishes green.

sauwming (approved):

laconicman (commented):
Built this and ran the new test locally. The approach looks right and the response is well formed on the wire. Four findings, one non-blocking note.

(On my setup: the test as written is unreachable here — a VPN interface owns the address `pj_gethostip()` returns, which fails the two pre-existing TCP tests too. I re-ran with the listener bound to loopback. Not a problem with the PR.)

### 1. The header-boundary scan disagrees with the parser

`pjsip_find_msg()` locates the end of the header block with a single sentinel:

```c
const pj_str_t end_hdr = { "\n\r\n", 3};
pos = pj_strstr(&cur_msg, &end_hdr);
```

The scan here matches `\r\n\r\n` and `\n\n`, but not `\n\r\n` — so an oversized request whose last header line ends with a bare LF gets no 513, silently back to the old behaviour. Both cases, rebuilt from clean each time:

| header block ends | this PR | `pj_strstr` |
|---|---|---|
| `\r\n\r\n` | 513 | 513 |
| `\n\r\n` | no response | 513 |

Reusing the parser's own sentinel is also 12 lines shorter, and keeps the two from drifting apart later:

```c
    const pj_str_t end_hdr = { "\n\r\n", 3 };
    pj_str_t received;
    pj_size_t hdr_len;
    char *pos;
    ...
    received.ptr = rdata->msg_info.msg_buf;
    received.slen = rdata->msg_info.len;
    if (received.ptr == NULL || received.slen <= 0)
        return;

    pos = pj_strstr(&received, &end_hdr);
    if (pos == NULL)
        return;
    hdr_len = (pj_size_t)(pos + 3 - received.ptr);
```

`pj_strstr()` rather than `strstr()` for the same reason the parser uses it: the buffer may contain NULs.

### 2. `parse_err` is not checked

The normal receive path gates on it before anything else:

```c
if (msg==NULL || !pj_list_empty(&rdata->msg_info.parse_err)) { ... }
```

Here only `msg == NULL` is tested, so the overflow path answers requests the normal path drops. Adding `Expires: notanumber` to the test request gives a non-empty `parse_err` and the 513 still goes out. One line:

```c
    if (msg == NULL || !pj_list_empty(&rdata->msg_info.parse_err) ||
        msg->type != PJSIP_REQUEST_MSG ||
```

### 3. Warning agent

The four existing `pjsip_warning_hdr_create()` callers pass `pjsip_endpt_name(endpt)`. `local_name.host` is stored unbracketed — printers add the brackets via `pj_addr_str_print(..., 1)` — so on an IPv6 transport this emits `Warning: 399 ::1 "..."`, which is not a valid `hostport`.

```c
    warn = pjsip_warning_hdr_create(rdata->tp_info.pool, 399,
                                    pjsip_endpt_name(endpt), &warn_str);
```

### 4. 513 is RFC 3261 section 21.5.7

Section 21.4.11 is 413 Request Entity Too Large. The citation appears three times: the description above, the function comment, and the test comment.

### Non-blocking

The test only covers an oversized request as the *first* message on a connection. The production case is an established one, which takes a different path — and with `pjsip_cfg()->tcp.initial_timeout` set, skips the `pjsip_transport_shutdown()` that follows the response here.

Related from the other side: #5076 covers an oversized request going out over UDP anyway, because the section 18.1.1 TCP switch found no TCP transport to acquire — the sending mirror of what this fixes on receive.

The four changes are verified together here: builds warning-free, the test passes with either terminator, and the malformed-header request is correctly refused. Happy to send them as a patch, or as a PR against your branch if that is easier.

laconicman:
Materialized the four points as https://github.com/Auerswald-GmbH/pjproject/pull/1 against `fix/respond-513-on-rx-overflow` — one commit each, so any can be dropped. Merging it updates this PR in place. Take or leave whichever is useful.

laconicman:
Also opened #5214 against master with the same commits stacked on this one, so the whole change is mergeable from this repo without a round trip through my fork. `13d0a4c5d` keeps its authorship there. Whichever route is easier — merging Auerswald-GmbH/pjproject#1 here, or #5214 — the other can just be closed.

nanangizz (approved):

sauwming:
A couple of minor notes before merging:
- PR description still says "RFC 3261 section 21.4.11.", worth correcting before merging.
- The two low-severity findings I just posted on #5214 apply equally here: the 513 can be lost to an RST when an initial timeout is configured (sip_endpoint.c:1145), and the single-read wait loop in the test (transport_tcp_test.c:741).

Nothing major, but worth addressing, either in this PR or in #5214.

andreas-wehrmann:
With this change, will the connection never be dropped when a message `> PJSIP_MAX_PKT_LEN ` is announced?
I think dropping the connection is fine if the announced Content-Length is above a certain threshold, otherwise a malicious client could use this to saturate a network link.

sauwming:
@andreas-wehrmann Good question, but the connection lifetime is unchanged by this PR.

Dropping the connection was never tied to the message size. `pjsip_tpmgr_receive_packet()` calls `pjsip_transport_shutdown()` only when `idle_timer.id == INITIAL_IDLE_TIMER_ID`, i.e. when an initial timeout is configured (`PJSIP_TCP_INITIAL_TIMEOUT`, `tls_opt.initial_timeout` or `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST`, all 0 by default) and this is the first message on the connection. That is exactly as it was before: an established connection already survived an oversized message, it just got no response. This PR touches only the response, not that logic.

On saturation: the exchange runs against the attacker, not for them. They have to push the full oversized request (>= `PJSIP_MAX_PKT_LEN`, 4000 bytes by default) to elicit a ~400-byte 513, so the ratio is roughly 1:0.1 -- no amplification, and cheaper for us than for them. It is also stream-only (`PJSIP_ERXOVERFLOW` is produced solely in the non-datagram branch), so there is no reflection vector: the response goes back over the attacker's own TCP connection, to an address they had to complete a handshake from.

A Content-Length threshold would not help either, since the value is attacker-controlled and need not match what is actually sent.

laconicman:
One addition to @sauwming's answer, on the word "announced": the 513 is not driven by `Content-Length` at all. `pjsip_find_msg()` returns `PJSIP_EPARTIALMSG` and the overflow branch additionally requires `remaining_len == PJSIP_MAX_PKT_LEN` — the buffer actually being full. A request that announces a huge `Content-Length` but sends little never reaches this path: it stays a partial message and is held until the idle timeout, exactly as before this PR.

So the announced length is not just attacker-controlled, it is never consulted — which is the other reason a threshold on it would not be the lever.

PR #5213, opened 2026-08-26
pjsua: cfg_add offset and a shared settings buffer allocation

laconicman:
## Description

The non-blocking points from the #5208 review, plus one regression that review turned up. Rebased onto master now that #5208 has merged, so this is just the three commits.

- **`cfg_add()` appends at the tracked `cfg->slen`** instead of letting `pj_ansi_strxcat()` rescan the destination on every call — the ~140 appends per `write_settings()` were quadratic in buffer size. `cfg_addf()` alongside it already did this. Four argument-less literals move off `cfg_addf()` while here.
- **`--custom-sdp` goes through a length-based `cfg_add_str()`.** #5208 routed it through `cfg_addf("%.*s", ...)`, which stops at an embedded NUL where the `pj_strcat()` it replaced copied all bytes. The helper sits under `#if !PJSUA_MEDIA_HAS_PJMEDIA` with its only caller — outside that guard it is an unused-function warning in a default `-Wall` build.
- **`alloc_settings()`** holds the pool-create/alloc/write/error-check block that was duplicated near-verbatim at four call sites. The `!pool` guard is kept, per the reasoning in #5208.
- **`dump_settings()` emits the dump in log-sized chunks.** `PJ_LOG` truncates a message at `PJ_LOG_MAX_SIZE` (4000). Before #5208 the buffer was `char settings[2000]` and always fit; at 64 KB it does not, so `dc` reported the full byte count while logging only the first part. This is a regression currently on master. Chunks break at option boundaries so no option is split across log lines, and the helper removes the last copy-paste pair between the legacy and CLI front ends.

Corrected from the earlier description: `alloc_settings()` checking the `pj_pool_alloc()` return is **not** the OOM protection I claimed. Under the default pool policy `pj_pool_allocate_find()` calls `(*pool->callback)()`, which throws, before it can return NULL — so that guard is unreachable. It is `pjsua_pool_create()` that genuinely returns NULL, via `pj_pool_create_int()` (`pool.c:224`), which is why the `!pool` guard stays.

## Motivation and Context

Follow-up to #5208 (issue #5196).

## How Has This Been Tested?

macOS/arm64.

- A config with four 1400-character option values, dumped with `dc` and written with `f`. On master the log shows `Dumping configuration (6228 bytes)` followed by ~4229 characters; with this change all 24 distinct options and all four full-length values reach the log, and no option is split across lines. The file path was already correct and is unchanged.
- A config with three 163-character values round-trips byte-exact, and re-reading the output reproduces it identically.
- With `PJSUA_APP_SETTINGS_SIZE` temporarily lowered to 300, both `dc` and `f` refuse — "not enough buffer", no file created, no partial dump.
- `pjsua_app_config.c`, `pjsua_app_legacy.c` and `pjsua_app_cli.c` compile warning-free under `-Wall` with `PJSUA_MEDIA_HAS_PJMEDIA` set to both 1 and 0.

Not run here: `tests/pjsua/run.py` imports `telnetlib`, removed in Python 3.13, and this host only has 3.14.

One behavioural note: the dump now logs with `THIS_FILE` of `pjsua_app_config.c` rather than of the calling front end, since the helper lives there.

## Types of changes

## Checklist:

### DIFF
--- a/pjsip-apps/src/pjsua/pjsua_app_config.c
+++ b/pjsip-apps/src/pjsua/pjsua_app_config.c
@@ -1974,9 +1974,31 @@ static void cfg_add(pj_str_t *cfg, pj_size_t max, const char *src)
     if (cfg->slen < 0)
         return;
 
-    len = pj_ansi_strxcat(cfg->ptr, src, max);
-    cfg->slen = (len < 0)? -1 : len;
+    len = pj_ansi_strxcpy(cfg->ptr + cfg->slen, src,
+                          max - (pj_size_t)cfg->slen);
+    if (len < 0)
+        cfg->slen = -1;
+    else
+        cfg->slen += len;
 }
+
+#if !PJSUA_MEDIA_HAS_PJMEDIA
+/* Length based, so a value containing a NUL is copied whole. */
+static void cfg_add_str(pj_str_t *cfg, pj_size_t max, const pj_str_t *src)
+{
+    if (cfg->slen < 0)
+        return;
+
+    if ((pj_size_t)cfg->slen + (pj_size_t)src->slen + 1 > max) {
+        cfg->slen = -1;
+        return;
+    }
+
+    pj_memcpy(cfg->ptr + cfg->slen, src->ptr, src->slen);
+    cfg->slen += src->slen;
+    cfg->ptr[cfg->slen] = '\0';
+}
+#endif
 
 static void cfg_addf(pj_str_t *cfg, pj_size_t max, const char *fmt, ...)
 {
@@ -2503,7 +2525,7 @@ int write_settings(pjsua_app_config *config, char *buf, pj_size_t max)
 
     /* Stereo mode. */
     if (config->media_cfg.channel_count == 2) {
-        cfg_addf(&cfg, max, "--stereo\n");
+        cfg_add(&cfg, max, "--stereo\n");
     }
 
     /* quality */
@@ -2528,7 +2550,7 @@ int write_settings(pjsua_app_config *config, char *buf, pj_size_t max)
                          config->avi[i].path.ptr);
     }
     if (config->avi_auto_play) {
-        cfg_addf(&cfg, max, "--auto-play-avi\n");
+        cfg_add(&cfg, max, "--auto-play-avi\n");
     }
     if (config->avi_rec.slen) {
         cfg_addf(&cfg, max, "--rec-avi %s\n",
@@ -2539,10 +2561,10 @@ int write_settings(pjsua_app_config *config, char *buf, pj_size_t max)
                          config->avi_rec_size);
     }
     if (config->avi_rec_audio) {
-        cfg_addf(&cfg, max, "--rec-avi-audio\n");
+        cfg_add(&cfg, max, "--rec-avi-audio\n");
     }
     if (config->avi_auto_rec) {
-        cfg_addf(&cfg, max, "--auto-rec-avi\n");
+        cfg_add(&cfg, max, "--auto-rec-avi\n");
     }
 
     /* ptime */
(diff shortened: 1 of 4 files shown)


sauwming (commented):
Two notes from reviewing the buffer handling. The bounds logic in `cfg_add`/`cfg_add_str`/`cfg_addf` itself checks out — I verified the append boundaries and that all emitted option strings are unchanged from master apart from the two intended `--reg-contact-*-params` newline merges.

pjsip-apps/src/pjsua/pjsua_app_legacy.c
```diff
@@ -1773,40 +1773,44 @@ static void ui_dump_call_quality()
...
-    if (len < 1)
-        PJ_LOG(1,(THIS_FILE, "Error: not enough buffer"));
-    else
-        PJ_LOG(3,(THIS_FILE, "Dumping configuration (%d bytes):\n%s\n",
-                  len, settings));
+    settings = alloc_settings(&app_config, &pool, &len);
+    if (!settings)
+        return;
+
+    PJ_LOG(3,(THIS_FILE, "Dumping configuration (%d bytes):\n%s\n",
```
sauwming: `PJ_LOG` truncates its formatted output at `PJ_LOG_MAX_SIZE` (default 4000). With the buffer grown from 2000 to 64 KB, `dc` now reports the full size but prints only the first ~4 KB, e.g. one account with a 20000-char `--contact-params` logs `Dumping configuration (20485 bytes):` followed by ~4 KB of text. The old 2000-byte buffer guaranteed the dump fit; the file write path is unaffected. Worth writing the dump in `PJ_LOG_MAX_SIZE`-sized chunks (or via `puts`/`pj_cli_sess_write_msg`). Same at `pjsua_app_cli.c:1849`.

pjsip-apps/src/pjsua/pjsua_app_config.c
```diff
@@ -2129,313 +2149,318 @@ static void write_account_settings(int acc_index, pj_str_t *result)
...
+                     int *p_len)
+{
+    pj_pool_t *pool;
+    char *buf;
+    int len;
+
+    *p_pool = NULL;
+
+    pool = pjsua_pool_create("settings", PJSUA_APP_SETTINGS_SIZE + 1024, 1024);
+    if (!pool) {
```
sauwming, line 2270: These two NULL guards don't catch OOM: under the default pool policy both `pjsua_pool_create()` and `pj_pool_alloc()` raise `PJ_NO_MEMORY_EXCEPTION` rather than returning NULL, and with no handler installed that aborts. Harmless as written, but the PR description lists the `pj_pool_alloc()` return check as a fix and it doesn't provide that protection.

pjsip-apps/src/pjsua/pjsua_app_legacy.c
laconicman: Confirmed and fixed in 9832fde97. Reproduced with four 1400-character option values: master logs `Dumping configuration (6228 bytes)` followed by ~4229 characters, while `f` writes all 6228. The pre-#5208 `char settings[2000]` always fit under `PJ_LOG_MAX_SIZE`; 64 KB does not.

Both call sites now go through a `dump_settings()` helper that chunks at option boundaries, so no option is split across log lines — all 24 options and all four full-length values now reach the log. That also removes the last copy-paste pair between the two front ends. One side effect worth naming: the dump now logs with `THIS_FILE` of `pjsua_app_config.c` rather than of the calling front end.

pjsip-apps/src/pjsua/pjsua_app_config.c
laconicman: You are right, and the description was overstating it — corrected.

The two cases differ, which is why your #5208 reasoning still holds for the other guard. `pj_pool_allocate_find()` calls `(*pool->callback)()` before its `return NULL`, and the default callback throws, so the `pj_pool_alloc()` check is indeed unreachable in a default build. `pj_pool_create_int()` returns NULL outright at `pool.c:224` with no callback in the path — exactly the trace you gave on #5208 — so the `!pool` guard is live and stays.

I have dropped the OOM-protection claim for the `pj_pool_alloc()` return from the description.

sauwming (approved):

nanangizz (approved):

PR #5214, opened 2026-08-26
pjsip: review follow-ups for the 513 Message Too Large path

laconicman:
## Description

Review follow-ups to #5211, stacked on it so the whole change is mergeable from this repo alone. Until #5211 merges the diff here carries its commit as well; after it merges this reduces to the five commits below. `13d0a4c5d` keeps its original authorship either way.

- **Locate the header block with the parser's own sentinel.** `pjsip_find_msg()` ends the headers at `"\n\r\n"`, so a last header line closed by a bare LF is a complete header block. The hand-rolled scan in `endpt_respond_msg_too_large()` matched `\r\n\r\n` and `\n\n` but not that, so such an oversized request still got no response. `pj_strstr()` also tolerates NULs in the buffer, which is why the parser uses it rather than `strstr()`. 12 lines shorter.
- **Do not answer a request the parser reported errors on.** `pjsip_tpmgr_receive_packet()` drops a message whose `parse_err` list is non-empty; testing only for a NULL `msg` let this path answer requests the normal path rejects.
- **Name the endpoint as the Warning agent.** Every other `pjsip_warning_hdr_create()` caller passes `pjsip_endpt_name()`. `local_name.host` is stored unbracketed — printers add the brackets through `pj_addr_str_print()` — so an IPv6 transport produced an unbracketed literal where `warn-agent` expects a `hostport`.
- **Correct the RFC reference.** 513 Message Too Large is RFC 3261 section 21.5.7; section 21.4.11 is 413 Request Entity Too Large.
- **Cover both header block terminators in the test.**

## Motivation and Context

Follow-up to #5211. The first two are correctness: an oversized request with an LF-terminated last header line gets no 513 at all, and a request the parser flagged gets one when it should be dropped.

The same commits are offered to the author directly as Auerswald-GmbH/pjproject#1, in case that route is preferred — that PR needs no action here, and this one does not depend on it.

## How Has This Been Tested?

macOS/arm64, on top of `13d0a4c5d`.

- `transport_rx_overflow_test` now runs the scenario for both terminators and passes for both. With only the first commit reverted, the `\n\r\n` case times out with no response — so the new case does catch the bug it is there for.
- With `Expires: notanumber` added to the request, the message is refused instead of answered, matching what `pjsip_tpmgr_receive_packet()` does with a non-empty `parse_err`.
- `sip_endpoint.c` and `transport_tcp_test.c` compile warning-free under `-Wall`; the pjsip build is clean.

One caveat on this host: the test's listener is unreachable here because a VPN interface owns the address `pj_gethostip()` returns — the two pre-existing TCP tests fail the same way. The runs above used a loopback-bound listener; that change is local only and is not in this branch.

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/pjsip/src/pjsip/sip_endpoint.c
+++ b/pjsip/src/pjsip/sip_endpoint.c
@@ -1003,10 +1003,10 @@ on_return:
 static void endpt_respond_msg_too_large( pjsip_endpoint *endpt,
                                          pjsip_rx_data *rdata )
 {
-    char *buf = rdata->msg_info.msg_buf;
-    pj_size_t len = (pj_size_t)rdata->msg_info.len;
-    pj_size_t hdr_len = 0;
-    pj_size_t i;
+    const pj_str_t end_hdr = { "\n\r\n", 3 };
+    pj_str_t received;
+    pj_size_t hdr_len;
+    char *pos;
     char *hdr_buf;
     pjsip_msg *msg;
     pjsip_warning_hdr *warn;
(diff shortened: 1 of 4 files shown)


pjsip/src/pjsip/sip_endpoint.c
```diff
@@ -1040,6 +1140,10 @@ static void endpt_on_rx_msg( pjsip_endpoint *endpt,
                   status,
                   (int)rdata->msg_info.len,     
                   rdata->msg_info.msg_buf));
+
+        if (status == PJSIP_ERXOVERFLOW)
+            endpt_respond_msg_too_large(endpt, rdata);
```
sauwming, line 1145: Note on reachability of this response: right after `endpt_on_rx_msg()` returns, `pjsip_tpmgr_receive_packet()` (sip_transport.c:2298) calls `pjsip_transport_shutdown()` on the same connection whenever `idle_timer.id == INITIAL_IDLE_TIMER_ID`, i.e. when `PJSIP_TCP_INITIAL_TIMEOUT` / `tls_opt.initial_timeout` / `PJSIP_TRANSPORT_SERVER_IDLE_TIME_FIRST` is non-zero and the oversized request is the first message on the connection.

The remainder of the oversized body is still unread in the socket, so the resulting `close()` emits an RST rather than a FIN, and on BSD/macOS an RST discards data already sitting in the peer's receive queue -- the 513 can be dropped before the peer reads it.

Not reachable with the shipped defaults (both timeouts default to 0), which is why the new test passes, but a deployment that configures an initial timeout is exactly the one where a first-message overflow happens. A brief comment noting that the response is best-effort in that configuration would be enough.

pjsip/src/test/transport_tcp_test.c
```diff
@@ -606,6 +606,199 @@ int transport_tcp_keep_alive_test(void)
...
+
+        flush_events(10);
+
+        PJ_FD_ZERO(&rset);
+        PJ_FD_SET(sock, &rset);
+        if (pj_sock_select((int)sock + 1, &rset, NULL, NULL, &zero) > 0 &&
+            PJ_FD_ISSET(sock, &rset))
+        {
+            received = sizeof(reply) - 1;
+            status = pj_sock_recv(sock, reply, &received, 0);
```
sauwming, line 741: This breaks out of the wait loop after a single `pj_sock_recv()`, and the check below then `strncmp`s against the full 11-byte `"SIP/2.0 513"`. A short read -- the response split across TCP segments -- would fail the test with `-545` instead of accumulating and retrying.

Practically impossible on loopback, but CI runners behind virtualized/proxied network stacks with a small MSS are where this would bite. Looping until at least `strlen(EXPECTED)` bytes have accumulated (or until the deadline expires) would make it deterministic.

laconicman:
jwes merged Auerswald-GmbH/pjproject#1, so #5211 now carries all five commits directly and this is redundant. Closing in favour of #5211.

The two findings from the review here apply to #5211 as it stands; I will send them there.

Issue #5209, opened 2026-08-23
TLS listener reports successful startup before certificate and private key are loaded


kazuki-ma:
### Describe the bug

With the OpenSSL backend, a TLS server transport can report successful startup before the configured certificate and private key files are parsed, validated, or loaded into an `SSL_CTX`.

`pj_ssl_cert_load_from_files2()` stores the file paths, while the actual OpenSSL context is initialized when the first connection is accepted. With server-side context reuse enabled, that context is then cached and shared by subsequent connections.

This creates an unexpected lifecycle:

- The TLS listener can be reported as ready before its credentials are usable.
- The first real client connection becomes the credential validation point.
- Files replaced between listener startup and the first accepted connection determine which certificate is served.
- Files replaced after the first accepted connection are not reflected until the transport is restarted.

Expected behavior:

The TLS listener startup should eagerly load and validate the server certificate and private key, and fail if they cannot be loaded or do not match. Alternatively, PJPROJECT could provide an explicit API or option for eager TLS context initialization and validation.

<details>
<summary>日本語訳</summary>

OpenSSL backend を使用する TLS server transport では、設定された証明書と秘密鍵が解析・検証され、`SSL_CTX` にロードされる前に listener の起動が成功として扱われます。

`pj_ssl_cert_load_from_files2()` はファイルパスを保持するだけで、実際の OpenSSL context は最初の接続を accept した時点で初期化され、その後キャッシュされます。

このため、証明書と秘密鍵が使用可能か確認される前に listener が ready になり、最初の実クライアント接続が credential の検証処理になります。また、初回接続前のファイル交換は提示証明書に反映されますが、初回接続後の交換は transport restart まで反映されません。

TLS listener の起動時に証明書と秘密鍵をロード・検証し、ロード不能または不一致の場合は起動を失敗させることを期待します。互換性上難しい場合は、TLS context を明示的に事前初期化・検証する API またはオプションが必要だと考えます。

</details>

### Steps to reproduce

1. Generate two valid certificate/key pairs:
   - Pair A: `CN=cert-before-start`
   - Pair B: `CN=cert-before-first-accept`
2. Configure a TLS transport to use pair A.
3. Start the TLS listener, but do not make any TCP/TLS connection to it.
4. Replace the configured files with pair B.
5. Make the first TLS connection. Pair B is presented.
6. Replace the files with pair A again, without restarting the transport.
7. Make a second TLS connection. Pair B is still presented.

Mismatch variant:

[...]

Issue #5215, opened 2026-08-26
pjlib: ssl_sock_darwin.c and ssl_sock_apple.m duplicate their certificate-loading code, and the copies have diverged


laconicman:
### Describe the bug

`pjlib/src/pj/ssl_sock_darwin.c` (Secure Transport) and `pjlib/src/pj/ssl_sock_apple.m` (Network.framework) carry copy-pasted Security.framework / CoreFoundation certificate code. The copies have drifted, and at least one crash fix landed in only one of them.

Measured on master (b7317ffca):

| code | `ssl_sock_darwin.c` | `ssl_sock_apple.m` | difference |
|---|---|---|---|
| `create_data_from_file()` | L158-224 | L411-477 | **byte-identical**, 67 lines |
| `get_cert_info()` | L893-1132 | L1866-2106 | one comment |
| `get_info_and_cn()` + `get_cert_oid()` | L827-892 | L1793-1865 | a NULL guard — see 3 below |
| identity-extraction loop | in `set_cert()`, L312-350 | in `create_identity_from_cert()`, L566-608 | a `CFRetain` — see 2 below |

Three defects follow from it.

**1. `create_data_from_file()` truncates at 8192 bytes — in both backends.**

```c
UInt8 data_buf[8192];
...
nbytes = CFReadStreamRead(read_stream, data_buf, sizeof(data_buf));
if (nbytes > 0)
    *data = CFDataCreate(NULL, data_buf, nbytes);
```

One read into a fixed buffer, no loop, no comparison against the file size. A certificate or PKCS#12 bundle over 8 KB is silently truncated and then fails inside `SecItemImport`/`SecPKCS12Import` as an unparsable file rather than as a too-large one. `CFReadStreamRead()` is also not obliged to fill the buffer in a single call, so a short read can truncate a smaller file too. This is the only file-reading path for both `cert_file` and `CA_file` in both backends (`ssl_sock_darwin.c:250` and `:1234`; `ssl_sock_apple.m:504` and `:625`).

Worth noting that every other backend delegates this to its TLS library instead of hand-rolling it — `SSL_CTX_use_certificate_chain_file()` (`ssl_sock_ossl.c:1627`), `gnutls_certificate_set_x509_*_file()` (`ssl_sock_gtls.c:783`), `mbedtls_x509_crt_parse_file()` (`ssl_sock_mbedtls.c:528`). The hand-rolled reader exists only in the two Apple backends, where it is duplicated.

[...]

PR #5216, opened 2026-08-26
pjlib: eagerly load/validate TLS server cert on listener start (Darwin)

laconicman:
> **Stacked on #5220.** That PR fixes the certificate over-release in this backend, without which the eager check added here crashes at listener startup on iOS. Review the top commit only; the diff collapses to one file once #5220 merges.

## Summary

- Extends #5210's eager server-cert validation to the Darwin (Secure Transport) backend, which has the same half of #5209. `set_cert()` is reachable only from `ssl_create()`, which never runs on the listener socket, so `pj_ssl_sock_start_accept()` returns `PJ_SUCCESS` for a certificate that is missing, unparsable, or has no matching private key, and every inbound connection then fails its handshake instead. Reproduction table in my comment on #5209.
- Splits the identity loading out of `set_cert()` into `create_identity_from_cert()`, mirroring how `ssl_sock_apple.m` already factors the same code out of `network_create_params()`, and calls it from `pj_ssl_sock_start_accept2()`. `set_cert()` keeps only the `SSLSetCertificate()` call.
- Now that #5210 has merged, the two backends share one call site: the guard is `PJ_SSL_SOCK_IMP_OPENSSL || PJ_SSL_SOCK_IMP_DARWIN` rather than a second `#if` block, and the header declares `ssl_init_server_ctx()` once. Same hook, same placement.
- **Caches nothing on the listener.** Unlike OpenSSL this backend has no shared server context; every accepted connection still loads the certificate in its own `ssl_create()`, so a certificate file replaced while the listener is up continues to be picked up per connection. Preserved deliberately — this validates, it does not pin.
- Carries #5210's `!ssock->cert` guard, so a listener with no certificate configured at all still starts, exactly as on the OpenSSL side.

## Scope of the check

The eager path exercises `create_identity_from_cert()` but not the `SSLSetCertificate()` that `set_cert()` goes on to make against a per-connection context, so a listener can still start for a configuration that every connection then rejects. The check fails open, never closed. Noted in the function comment.

## Verified on both platforms

Same repro, five certificate configurations, one real TCP connection driven into each listener that starts. Built with `PJ_SSL_SOCK_IMP_DARWIN` forced via `config_site.h`, since `./configure` cannot select this backend on a current SDK (#5217).

**iOS Simulator, arm64** — with #5220 applied:

| certificate configuration | master: `start_accept` | master: first connection | this PR: `start_accept` |
|---|---|---|---|
| no certificate configured | `PJ_SUCCESS` | timeout[^1] | `PJ_SUCCESS` |
| valid .p12, correct password | `PJ_SUCCESS` | timeout[^1] | `PJ_SUCCESS` |
| valid .p12, wrong password | `PJ_SUCCESS` | OSStatus -25293 | OSStatus -25293 |
| unparsable file | `PJ_SUCCESS` | OSStatus -26275 | OSStatus -26275 |
| nonexistent file | `PJ_SUCCESS` | `PJ_EINVAL` | `PJ_EINVAL` |

The valid `.p12` row is the positive control: it loads, the listener starts, and this PR does not reject it.

**macOS, arm64** — same shape, with the valid `.p12` row failing on both master and this PR because the macOS import path resolves the private key through the keychain rather than from the file (`errSecItemNotFound`), and I did not install one. Pre-existing platform behaviour, not a change from this PR:

| certificate configuration | master: `start_accept` | master: first connection | this PR: `start_accept` |
|---|---|---|---|
| no certificate configured | `PJ_SUCCESS` | timeout[^1] | `PJ_SUCCESS` |
| valid .p12, correct password | `PJ_SUCCESS` | `PJ_EINVAL` | `PJ_EINVAL` |
| valid .p12, wrong password | `PJ_SUCCESS` | OSStatus -25256 | OSStatus -25256 |
| unparsable file | `PJ_SUCCESS` | OSStatus -25256 | OSStatus -25256 |
| nonexistent file | `PJ_SUCCESS` | `PJ_EINVAL` | `PJ_EINVAL` |

In every row that fails, the status is identical to what master produced at first connection; only *when* it is reported changes.

[^1]: my client opens a TCP connection but sends no ClientHello, so a healthy listener times out waiting. That is the "listener is fine" outcome.

## Also

- Compile-checked for `PJ_SSL_SOCK_IMP_APPLE` and `PJ_SSL_SOCK_IMP_OPENSSL` — unaffected, as the guard implies.
- `pjlib-test ssl_sock_test` not run: it does not link in this configuration on a current SDK (configure omits `-framework Security` because the probe in #5217 fails), and its server path expects a keychain identity. No existing test starts a listener with a deliberately bad certificate.

## Possible follow-up, not in this PR

Caching the `SecIdentityRef` on the listener and having accepted sockets borrow it via `ssock->parent` would also remove a file read and a PKCS#12 import from every inbound connection; `ssl_sock_ossl.c` already has the ownership pattern (`own_ctx`, child reads the parent's context, child holds a group-lock reference on the parent). It would pin the certificate for the listener's lifetime, matching OpenSSL and the Apple backend. Left out because it is a behaviour change rather than a fix.

### DIFF
--- a/pjlib/src/pj/ssl_sock_darwin.c
+++ b/pjlib/src/pj/ssl_sock_darwin.c
@@ -222,7 +222,18 @@ static pj_status_t create_data_from_file(CFDataRef *data,
     return (*data? PJ_SUCCESS: PJ_EINVAL);
 }
 
-static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
+/* Load the certificate configured in cert into a SecIdentityRef. The private
+ * key is not read from cert: it must come from the PKCS#12 bundle itself or
+ * already be present in the keychain.
+ *
+ * On success *p_identity is set to NULL when no certificate is configured at
+ * all, and otherwise to an identity the caller owns and must CFRelease() --
+ * in both import branches, since the dictionary branch takes its own
+ * reference to what CFDictionaryGetValue() only lends it.
+ */
+static pj_status_t create_identity_from_cert(darwinssl_sock_t *dssock,
+                                             pj_ssl_cert_t *cert,
+                                             SecIdentityRef *p_identity)
 {
     CFStringRef password = NULL;
     CFDataRef cert_data = NULL;
@@ -232,11 +243,10 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
     CFArrayRef items;
     CFIndex i, count;
     SecIdentityRef identity = NULL;
-    CFTypeRef cert_arr[1];
-    CFArrayRef cert_refs;
     OSStatus err;
     pj_status_t status;
 
+    *p_identity = NULL;
 
     if (cert->privkey_file.slen || cert->privkey_buf.slen ||
         cert->privkey_pass.slen)
@@ -357,7 +367,26 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
                                   "the cert file"));
             return PJ_EINVAL;
         }
-        
+
+        *p_identity = identity;
+    }
+
+    return PJ_SUCCESS;
+}
+
+static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
+{
+    SecIdentityRef identity = NULL;
+    CFTypeRef cert_arr[1];
+    CFArrayRef cert_refs;
+    OSStatus err;
+    pj_status_t status;
+
+    status = create_identity_from_cert(dssock, cert, &identity);
+    if (status != PJ_SUCCESS)
+        return status;
+
+    if (identity) {
         cert_arr[0] = identity;
         cert_refs = CFArrayCreate(NULL, (const void **)cert_arr, 1,
                               &kCFTypeArrayCallBacks);
@@ -375,6 +404,41 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
     }
 
     return PJ_SUCCESS;
+}
+
+/* Eagerly load and validate the certificate on the listener socket, so that a
+ * certificate that is missing, unparsable, or has no matching private key
+ * fails pj_ssl_sock_start_accept() up front. Without this, set_cert() (via
+ * ssl_create() below) does not run until the first connection is accepted, so
+ * the listener reports success and it is every inbound connection that fails
+ * its handshake instead.
+ *
+ * The identity is released again here rather than cached: unlike the OpenSSL
+ * backend, this backend builds no shared server context, and every accepted
+ * connection loads the certificate itself in its own ssl_create(). This
+ * validates the configuration, it does not pin it.
+ *
+ * Note this covers loading only, not the SSLSetCertificate() that set_cert()
+ * goes on to make against a per-connection context. A listener can therefore
+ * still start for a configuration that every connection subsequently
+ * rejects -- the check fails open, never closed.
+ */
+static pj_status_t ssl_init_server_ctx(pj_ssl_sock_t *ssock)
+{
+    darwinssl_sock_t *dssock = (darwinssl_sock_t *)ssock;
+    SecIdentityRef identity = NULL;
+    pj_status_t status;
+
+    pj_assert(ssock->is_server && !ssock->parent);
+
+    if (!ssock->cert)
+        return PJ_SUCCESS;
+
+    status = create_identity_from_cert(dssock, ssock->cert, &identity);
+    if (identity)
+        CFRelease(identity);
+
+    return status;
 }
 
 /* Create and initialize new Darwin SSL context and instance */
--- a/pjlib/src/pj/ssl_sock_darwin.c
+++ b/pjlib/src/pj/ssl_sock_darwin.c
@@ -273,8 +273,12 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
         options = CFDictionaryCreate(NULL, (const void **)keys,
                                      (const void **)values,
                                      (password? 1: 0), NULL, NULL);
-        if (!options)
+        if (!options) {
+            if (password)
+                CFRelease(password);
+            CFRelease(cert_data);
             return PJ_ENOMEM;
+        }
         
 #if TARGET_OS_IPHONE
         err = SecPKCS12Import(cert_data, options, &items);
@@ -322,6 +326,11 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
                 identity = (SecIdentityRef)
                            CFDictionaryGetValue((CFDictionaryRef) item,
                                                 kSecImportItemIdentity);
+                /* Get rule: this reference is owned by the dictionary, which
+                 * is released along with items below, so take our own.
+                 */
+                if (identity)
+                    CFRetain(identity);
                 break;
             }
 #if !TARGET_OS_IPHONE
@@ -352,8 +361,10 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
         cert_arr[0] = identity;
         cert_refs = CFArrayCreate(NULL, (const void **)cert_arr, 1,
                               &kCFTypeArrayCallBacks);
-        if (!cert_refs)
+        if (!cert_refs) {
+            CFRelease(identity);
             return PJ_ENOMEM;
+        }
     
         err = SSLSetCertificate(dssock->ssl_ctx, cert_refs);
     
@@ -878,6 +889,12 @@ static CFDictionaryRef get_cert_oid(SecCertificateRef cert, CFStringRef oid,
 
     vals = SecCertificateCopyValues(cert, key_arr, NULL);
     dict = CFDictionaryGetValue(vals, key[0]);
+    if (!dict) {
+        CFRelease(key_arr);
+        CFRelease(vals);
+        return NULL;
+    }
+
     *value = CFDictionaryGetValue(dict, kSecPropertyKeyValue);
 
     CFRelease(key_arr);
--- a/pjlib/src/pj/ssl_sock_imp_common.c
+++ b/pjlib/src/pj/ssl_sock_imp_common.c
@@ -2233,7 +2233,8 @@ pj_ssl_sock_start_accept2(pj_ssl_sock_t *ssock,
 
     ssock->is_server = PJ_TRUE;
 
-#if (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_OPENSSL)
+#if (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_OPENSSL) || \
+    (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_DARWIN)
     /* Eagerly load and validate the certificate/private key now, instead
      * of leaving it to the first accepted connection, so that a bad
      * credential fails the listener startup itself.
--- a/pjlib/src/pj/ssl_sock_imp_common.h
+++ b/pjlib/src/pj/ssl_sock_imp_common.h
@@ -293,9 +293,12 @@ static pj_ssl_cipher ssl_get_cipher(pj_ssl_sock_t *ssock);
 static void ssl_update_certs_info(pj_ssl_sock_t *ssock);
 #if (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_OPENSSL)
 static void ssl_free_cert(pj_ssl_cert_t *cert);
-/* Eagerly create/validate the server SSL_CTX (incl. certificate and
- * private key loading) on a listener socket, instead of deferring it
- * to the first accepted connection.
+#endif
+#if (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_OPENSSL) || \
+    (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_DARWIN)
+/* Eagerly create/validate the server credentials -- for OpenSSL the whole
+ * SSL_CTX, including certificate and private key loading -- on a listener
+ * socket, instead of deferring it to the first accepted connection.
  */
 static pj_status_t ssl_init_server_ctx(pj_ssl_sock_t *ssock);
 #endif


sauwming (commented):
Thanks for the patch. The refactor itself looks faithful: `create_identity_from_cert()` is a clean lift of the identity-loading half of `set_cert()`, the residual `set_cert()` is behaviour-identical, the `#if PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_DARWIN` guard is placed correctly, and `goto on_error` at that point is safe (`ssl_reset_sock_state()` only touches state already created by `pj_ssl_sock_create()`). Outbound TLS is unaffected, as you say.

Two issues below need resolving before this can go in, plus a few smaller notes.

One more, on a line the diff does not touch but that this PR makes reachable from listener startup: `ssl_sock_darwin.c:283`, `if (!options) return PJ_ENOMEM;` returns without releasing `cert_data` or `password`. Pre-existing, but worth folding in while the function is being moved.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -366,6 +392,36 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
...
+    pj_status_t status;
+
+    pj_assert(ssock->is_server && !ssock->parent);
+
+    if (!ssock->cert)
+        return PJ_SUCCESS;
+
+    status = create_identity_from_cert(dssock, ssock->cert, &identity);
+    if (identity)
+        CFRelease(identity);
```
sauwming, line 439: This `CFRelease()` is applied to a reference that may be borrowed. In the `CFDictionaryGetValue(item, kSecImportItemIdentity)` branch at line 330 the identity follows the CoreFoundation Get rule -- it is not owned -- and `CFRelease(items)` at line 351 has already dropped the only owning reference before the function returns. That is the over-release tracked in #5215.

The problem here is that this PR adds a *second* consumer of it, on a code path that previously did no CoreFoundation work at all. On iOS, where `SecPKCS12Import()` is the only import path and always yields dictionaries, a perfectly valid `.p12` would now hit the over-release inside `pj_ssl_sock_start_accept()` -- i.e. a crash at listener startup, where master merely deferred the same over-release to `ssl_create()`.

Please either land the #5215 fix first (`CFRetain()` the identity in the dictionary branch before `CFRelease(items)`, mirroring what af82c9085 did for `ssl_sock_apple.m`), or have `ssl_init_server_ctx()` not release at all.

pjlib/src/pj/ssl_sock_imp_common.c
```diff
@@ -2233,6 +2233,16 @@ pj_ssl_sock_start_accept2(pj_ssl_sock_t *ssock,
 
     ssock->is_server = PJ_TRUE;
 
+#if (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_DARWIN)
+    /* Eagerly load and validate the certificate now, instead of leaving it
+     * to the first accepted connection, so that a bad credential fails the
+     * listener startup itself.
+     */
+    status = ssl_init_server_ctx(ssock);
```
sauwming, line 2242: Hard-failing `pj_ssl_sock_start_accept2()` here contradicts this backend's deliberately non-pinning design, which the PR description itself preserves ("a certificate file replaced while the listener is up continues to be picked up per connection").

Concretely: a service started before the macOS keychain is unlocked, or with the identity provisioned shortly after launch, gets `errSecItemNotFound` from `SecIdentityCreateWithCertificate()`. On master the listener comes up and starts serving as soon as the identity becomes available, precisely because every accepted connection reloads it. With this change `pjsip_tls_transport_start()` fails outright at `sip_transport_tls.c:546`, the transport is never created, and there is no recovery once the cert *does* become valid.

So the eager check is only equivalent to the per-connection check for configurations that are statically broken; for transiently-unavailable ones it is strictly worse. Could we log a warning at listener start instead of failing, or else gate the hard failure behind the same caching change that would make the certificate genuinely pinned?

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -222,7 +222,15 @@ static pj_status_t create_data_from_file(CFDataRef *data,
     return (*data? PJ_SUCCESS: PJ_EINVAL);
 }
 
-static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
+/* Load the certificate configured in cert into a SecIdentityRef. The private
+ * key is not read from cert: it must come from the PKCS#12 bundle itself or
+ * already be present in the keychain. On success *p_identity is set to the
+ * identity, which the caller owns, or to NULL when no certificate is
```
sauwming: This contract is not accurate: the caller owns `*p_identity` only in the `SecIdentityCreateWithCertificate()` branch. In the `CFDictionaryGetValue()` branch at line 330 the reference is borrowed and already dead by the time the function returns.

Since this comment is what justifies the unconditional `CFRelease()` at both call sites, it will keep the bug alive for the next reader. Please either document the split ownership or make the contract true by retaining.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -366,6 +392,36 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
...
+ * certificate that is missing, unparsable, or has no matching private key
+ * fails pj_ssl_sock_start_accept() up front. Without this, set_cert() (via
+ * ssl_create() below) does not run until the first connection is accepted, so
+ * the listener reports success and it is every inbound connection that fails
+ * its handshake instead.
+ *
+ * The identity is released again here rather than cached: unlike the OpenSSL
+ * backend, this backend builds no shared server context, and every accepted
+ * connection loads the certificate itself in its own ssl_create(). This
+ * validates the configuration, it does not pin it.
```
sauwming, line 419: Worth being precise here: the eager check exercises `create_identity_from_cert()` but never `SSLSetCertificate()` (line 384), so the validation is one-directional -- the listener can still start for a configuration that every connection will then reject. Not a regression, failing open is the safe direction, but it does narrow the "fails exactly when every connection would have failed" claim in the commit message.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Confirmed, and it is worse than "a second consumer" — I reproduced it on the iOS Simulator and master already crashes without this PR.

Master, valid `.p12`, correct password, one TCP connection into the listener:

```
EXC_BAD_ACCESS, KERN_INVALID_ADDRESS at 0x0000beadde8d66a0
  libobjc.A.dylib   objc_retain
  CoreFoundation    __NSSingleObjectArrayI_new
  repro             ssl_create
  repro             asock_on_accept_complete2
```

With this PR the same crash simply moves earlier, exactly as you predicted:

```
EXC_BAD_ACCESS, KERN_INVALID_ADDRESS at 0x0000beadde8d0160
  libobjc.A.dylib   objc_release
  repro             pj_ssl_sock_start_accept2
```

So the Darwin backend segfaults on iOS on any valid PKCS#12 today. I have sent the `CFRetain()` backport plus two smaller fixes as #5220, and rebased this PR on top of it; with #5220 applied, both platforms are clean and a valid `.p12` loads and serves. Test matrix is in #5220 and in the updated description here.

My macOS-only testing is why I missed it: `SecItemImport()` returns a `SecCertificateRef`, so macOS takes the `SecIdentityCreateWithCertificate()` branch, which is correctly owned. The borrowed branch is iOS-only in practice. Thanks for catching it.

pjlib/src/pj/ssl_sock_imp_common.c
laconicman: This is the one I want to push back on, though I will change it if you still disagree after the following.

**The self-heal being protected is macOS-only, and on iOS it is a crash.** Before #5220, the deferred path on iOS does not recover when the identity appears — it segfaults at the first accepted connection (backtrace in the other thread). So "the listener comes up and starts serving as soon as the identity becomes available" holds on macOS and not on iOS, where the per-connection reload is precisely what dereferences the freed identity.

**The transient argument applies equally to OpenSSL, and #5210 already accepted it.** `ssl_init_server_ctx()` returns early on `!ssock->cert`, so OpenSSL hard-fails only when a certificate was configured but cannot be loaded. A cert file that appears shortly after startup — a slow-mounting volume, a provisioning script, cert-manager writing a secret — is the same class of transient failure as a late keychain unlock, and nothing in either backend distinguishes "not yet available" from "broken". My patch carries the identical `!ssock->cert` guard. I do not see a principled reason for this backend to be more permissive than OpenSSL for the same class of failure.

**Recovery is first-class API here, not a gap.** `pjsip_tls_transport_restart()` and `restart2()` (`sip_transport_tls.h:635-659`) close and recreate the listener from the stored config, and `pjsua_transport_lis_restart()` wraps them. pjsua already runs a scheduled-retry loop of exactly this shape: `restart_listener()` in `pjsua_core.c` calls the per-type restart and, when it fails, reschedules itself via `pjsua_schedule_timer2(&restart_listener_cb, ..., restart_lis_delay)`. Failure leaves nothing poisoned either — `pjsip_tls_transport_start()` calls `lis_destroy()` on the error path, and `pjsua_transport_create()` returns without touching `pjsua_var.tpdata[id]`, so a later retry gets a clean slot.

**And what is the listener for in the meantime?** Nothing in the tree wires up anonymous cipher suites, so a server with no usable identity cannot complete a handshake. A listener that reports ready and then fails every connection is not degraded service, it is the same failure reported later and more quietly — which is what #5209 was filed about.

Where I think you are straightforwardly right: for a keychain that unlocks after launch, master on macOS recovers with no application involvement and this PR requires an explicit restart. That is a real behaviour change for that one case. My argument is that an explicit restart is the house convention for a listener that failed to start, that it is already implemented and already used for IP-change handling, and that the alternative leaves this backend inconsistent with OpenSSL while #5209 stays open for it.

If you would still rather have the softer form, warn-only is a two-line change — drop the `goto on_error` and `PJ_PERROR` the status — and I will push it. It would mean this PR stops fixing #5209 for this backend and becomes a diagnostic, so I would want that stated in the commit message rather than left implicit.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Fixed. #5220 adds the `CFRetain()`, which makes the contract true rather than merely documented, and the comment now says so explicitly instead of glossing over the split:

```c
 * On success *p_identity is set to NULL when no certificate is configured at
 * all, and otherwise to an identity the caller owns and must CFRelease() --
 * in both import branches, since the dictionary branch takes its own
 * reference to what CFDictionaryGetValue() only lends it.
```

You were right that the old wording would have kept the bug alive for the next reader; it very nearly kept it alive for me.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Agreed, and the claim is now narrowed in three places: the commit message, the PR description, and the function comment, which reads:

```c
 * Note this covers loading only, not the SSLSetCertificate() that set_cert()
 * goes on to make against a per-connection context. A listener can therefore
 * still start for a configuration that every connection subsequently
 * rejects -- the check fails open, never closed.
```

The measured table still shows identical statuses in every row, but you are right that this is an observation about these five configurations rather than a guarantee the code structure provides.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -348,7 +367,26 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
...
+    pj_status_t status;
+
+    status = create_identity_from_cert(dssock, cert, &identity);
+    if (status != PJ_SUCCESS)
+        return status;
+
+    if (identity) {
         cert_arr[0] = identity;
         cert_refs = CFArrayCreate(NULL, (const void **)cert_arr, 1,
                               &kCFTypeArrayCallBacks);
```
sauwming, line 392: New minor issue, courtesy of #5220: the `return PJ_ENOMEM` just below leaves `identity` unreleased.

That was harmless before -- on iOS the reference was borrowed and already dead, so *not* releasing it was accidentally correct -- but now that #5220 retains in both branches, `set_cert()` owns the identity and this path leaks a `SecIdentityRef`. Needs a `CFRelease(identity)` before the return.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -366,6 +404,41 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
...
+    darwinssl_sock_t *dssock = (darwinssl_sock_t *)ssock;
+    SecIdentityRef identity = NULL;
+    pj_status_t status;
+
+    pj_assert(ssock->is_server && !ssock->parent);
+
+    if (!ssock->cert)
+        return PJ_SUCCESS;
+
+    status = create_identity_from_cert(dssock, ssock->cert, &identity);
```
sauwming, line 437: Not introduced here, but this change amplifies it enough to be worth raising: `create_data_from_file()` does a single `CFReadStreamRead()` into an 8 KB stack buffer (`data_buf` at line 163, the read at line 212) and silently truncates anything larger. A full-chain PEM bundle, or a `.p12` carrying an intermediate chain, exceeds that routinely.

Previously truncation surfaced as per-connection handshake failures. Now it fails `pj_ssl_sock_start_accept()` and takes the whole listener down at startup. That is the direction #5209 asked for, but it is also the most likely shape of the first "this PR broke my TLS listener" report. Worth either a line in the commit message or, better, a separate fix for the truncating read.

sauwming (commented):
Re-reviewed after the rebase on #5220. The three concrete comments are properly addressed: the `CFRetain()` balances the ref count in both import branches, the ownership contract in the doc comment is now true of the code rather than merely asserted, and the "fails exactly when every connection would have failed" claim is correctly narrowed. The `!options` leak I mentioned in the review body is fixed too. Merges cleanly on master now that #5210 is in.

I have replied on the `ssl_sock_imp_common.c` thread: let's keep the hard fail, with one request about the commit message.

Two remaining items inline, one of which is a new minor defect that #5220 introduces here.

Merge order: this branch carries #5220's commit, so #5220 needs to land first or merging this merges both.

pjlib/src/pj/ssl_sock_imp_common.c
sauwming: Let's keep the hard fail. But I want the record straight on two parts of the argument, because one is weaker than you state and the other is a real gap.

**On symmetry with OpenSSL.** The guard in `ssl_sock_ossl.c` is:

```c
if (!SERVER_SUPPORT_SESSION_REUSE || !ssock->cert)
    return PJ_SUCCESS;
```

That first clause is exactly the gate I was asking for. OpenSSL hard-fails only when the `SSL_CTX` is about to be cached and shared by every connection -- that is, only when the per-connection-reload self-heal does not exist. The same function then sets `ossock->own_ctx = PJ_TRUE`, so the eager path *is* the pinning path there. Darwin caches nothing, so the analogous condition would never be true. The design does distinguish your two cases, just indirectly: not "not yet available" vs "broken", but "recovery is possible" vs "it isn't".

What carries your position is therefore not symmetry but the fact that `SERVER_SUPPORT_SESSION_REUSE` is hardcoded to `1` (`ssl_sock_ossl.c:64`), so every shipped OpenSSL build hard-fails with no self-heal. Practical consistency holds; design symmetry does not.

**On recovery being first-class.** `restart_listener()` does reschedule itself, but it is reachable only from `pjsua_handle_ip_change()`, and its `pjsua_transport_get_info()` guard requires an already-valid transport. There is no auto-retry on the initial `pjsua_transport_create()` path -- a locked keychain at boot means the application has to retry explicitly. Your points about nothing being poisoned are correct (`restart2()` keeps the factory registered, `pjsua_transport_create()` leaves `tpdata[id]` untouched), so a later retry does get a clean slot. But that retry is the application's to write, not something the stack performs on its behalf.

**Why I am accepting it anyway.** Your last argument is the decisive one: nothing in the tree wires up anonymous cipher suites, so a listener with no usable identity cannot serve anybody. A listener that reports ready and then rejects every handshake is #5209 restated, not degraded service. And matching what OpenSSL actually ships is worth more than matching what its guard would do in a configuration nobody builds.

One request: please state the behaviour change in the commit message rather than leaving it implicit -- to the effect that a certificate which only becomes loadable after startup (late keychain unlock, slow-mounting volume, a provisioning script) now needs an explicit `pjsip_tls_transport_restart()`, where this backend previously self-healed on the next connection. That is the one case where this is strictly a regression, and whoever hits it should find it in `git log` rather than in a bug report.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Good catch, and the fix belongs in #5220 rather than here: without this PR, #5220 alone leaves that leak in `set_cert()`. I have added `CFRelease(identity)` on the `CFArrayCreate()` failure path there, called it out in that commit message, and rebased this branch on top, so the refactored `set_cert()` inherits it. Verified present after the rebase.

Your reading of why it was previously harmless is exactly right, and it is a good illustration of why the borrowed reference was so easy to miss: not releasing was *correct* behaviour for a reference we did not own.

pjlib/src/pj/ssl_sock_imp_common.c
laconicman: Taking the correction. You are right that the `SERVER_SUPPORT_SESSION_REUSE` clause is a real gate and not incidental — the eager path in `ssl_sock_ossl.c` is the pinning path, `own_ctx` is set right there, and the analogous condition on this backend is never true. So the design does distinguish "recovery is possible" from "it isn't", and my symmetry argument was the wrong shape. What holds is the narrower practical claim: the macro is hardcoded to `1`, so every shipped OpenSSL build takes the hard-fail path and no self-heal exists in practice.

The `restart_listener()` correction is fair too. I checked the reachability but not the guard: it is `pjsua_handle_ip_change()`-only and wants an already-valid transport, so it is not a retry path for a listener that never came up. The recovery is available but the application has to write it.

I have reworded the commit message to state the behaviour change explicitly, near enough to your wording:

> Behaviour change worth noting: a certificate that only becomes loadable after startup -- a keychain unlocked after launch, a slow-mounting volume, a provisioning script writing the file -- now requires an explicit `pjsip_tls_transport_restart()`. This backend previously self-healed on the next accepted connection, since it caches nothing on the listener and every connection reloads the certificate. That per-connection reload is unchanged; only listener startup is now strict, matching what the OpenSSL backend ships.

I also folded in a sentence about the 8 KB truncation from your other comment, since that is the same "who hits this first" concern.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Agreed on both counts, and I have taken the "better" option rather than only the commit-message line — though it lands as its own PR, since the truncating read is duplicated verbatim in `ssl_sock_apple.m` and fixing it in one place only would leave the Apple backend carrying it.

A line is in this commit message meanwhile:

> Note also that create_data_from_file() silently truncates anything over 8 KB, so an oversized certificate or CA bundle now fails the listener instead of each connection.

Your framing is the right one and I had underweighted it: an 8 KB ceiling is not exotic for a `.p12` carrying an intermediate chain, and turning a per-connection failure into a listener that will not start is exactly where the first complaint comes from. It also reaches further than TLS server certs — `create_data_from_file()` is what reads `CA_file`, so anyone pinning to a CA hits the same ceiling.

sauwming (approved):
need to merge #5220 first

nanangizz (approved):

Issue #5217, opened 2026-08-26
pjlib: --enable-darwin-ssl is a silent no-op, and neither Apple TLS backend is reachable from configure


laconicman:
### Describe the bug

On macOS, `./configure --enable-darwin-ssl` succeeds, says nothing about Darwin SSL, and produces a build with no TLS at all.

Three problems combine.

**1. `--enable-darwin-ssl` does nothing.** `AC_ARG_ENABLE(darwin-ssl, …)` (`aconfigure.ac:2131`) puts the detection in the *action-if-not-given* arm, so passing the flag skips it entirely, and the action-if-given arm only handles `= "no"`. Passing `--enable-darwin-ssl` produces no probe, no `AC_DEFINE`, and no output. The help string documents only `--disable-darwin-ssl` (`:2132`), which matches the implementation, but the generated `configure` still accepts the `--enable-` spelling and silently ignores it.

**2. The autodetect probe can no longer succeed on a current SDK.** Without the flag the probe does run, and fails: it compiles with `CFLAGS="-Werror"` (`:2142`) and calls `SSLReHandshake`, deprecated since macOS 10.15, so `ac_ssl_backend=darwin` is never set on any recent SDK.

**3. `PJ_SSL_SOCK_IMP_APPLE` is not selectable from configure.** The identifier does not appear in `aconfigure.ac` at all. `ssl_sock_apple.o` *is* added to `ac_os_objs` unconditionally on darwin targets (`:670`), but with `PJ_SSL_SOCK_IMP` undefined it compiles to an empty translation unit.

So neither Apple TLS backend can currently be selected through configure, and the failure is silent. `PJ_HAS_SSL_SOCK` stays undefined, so `PJSIP_HAS_TLS_TRANSPORT` (`sip_config.h:955-957`) is undefined, so `sip_transport_tls.c` compiles to nothing (`:33`) and pjsua's TLS paths compile out. The build succeeds; TLS is simply absent, and that is discovered at runtime.

Setting `PJ_SSL_SOCK_IMP` by hand in `config_site.h` is the remaining route, and it does not link. `-framework Security` is added only inside the probe's success branch (`:2163`), and `-framework Network` does not appear anywhere in the build system, so a forced build fails with undefined `_SSLCreateContext`, `_SSLHandshake`, `_SSLClose`, etc.

For contrast, `--disable-ssl` behaves correctly — "no" is the one case the macro implements.

### Steps to reproduce

```
[...]

Issue #3617, opened 2023-06-30
pjsua_transport_create() crashes when using a .p12 certificate on iOS


henlund:
### Describe the bug

Application crashes when setting up transport with TLS and providing a .p12 certificate containing cert, ca list and private key.

### Steps to reproduce

1. Configure transport with TLS using a .p12 certificate containing cert, ca list and private key in Swift:

let url = Bundle.main.path(forResource: "certificate", ofType: "p12")
var tlsSetting = pjsip_tls_setting()
pjsip_tls_setting_default(&tlsSetting)
tlsSetting.cert_file = pj_str(strdup(url))
tlsSetting.password = pj_str(strdup("password"))
transportConfig.tls_setting = tlsSetting
let status = pjsua_transport_create(PJSIP_TRANSPORT_TLS, &transportConfig, &transportId)

2. Application chrashes inside ssl_sock_apple.m at

*p_identity = sec_identity_create(identity);

### PJSIP version

2.13

### Context

- This issue happens on iOS 16.5.1

### Log, call stack, etc

```shell
ssl_sock_apple.m  "Ignoring supplied private key. Private key must be placed in the keychain instead."
```

However it's not clear how the private key should be added to the keychain in order for pjsip to read it. An example in the documentation would be very much appreciated.

sauwming:
Try the steps in:
https://stackoverflow.com/questions/9418661/how-to-create-p12-certificate-for-ios-distribution

henlund:
That link refers to how to use p12 certificates for push notifications when distributing iOS apps. I'm looking for an example how to configure pjsip for iOS using the native TLS support added here https://github.com/pjsip/pjproject/pull/2482.

PR #5220, opened 2026-08-27
pjlib: fix certificate memory handling in the Darwin SSL backend

laconicman:
## Summary

Fixes the memory-handling half of #5215 for `ssl_sock_darwin.c`. The headline is worse than that issue implied: **on iOS this backend segfaults on any valid PKCS#12 today.** Not a latent over-release — a guaranteed crash the moment a TLS connection arrives.

`set_cert()` takes the identity out of the import result with `CFDictionaryGetValue()`, which follows the CoreFoundation Get rule, then releases the array that owns it before using and releasing the identity itself. On iOS `SecPKCS12Import()` is the only import path and always yields dictionaries, so the borrowed branch is the only branch. af82c9085 (#3664, fixing #3617) fixed exactly this in `ssl_sock_apple.m` in 2023; the byte-identical code here never got the backport.

Measured on the iOS Simulator (arm64), master, loading a valid `.p12` with the correct password and driving one TCP connection into the listener:

```
Child process terminated with signal 11: Segmentation fault

EXC_BAD_ACCESS, KERN_INVALID_ADDRESS at 0x0000beadde8d66a0
  libobjc.A.dylib   objc_retain
  CoreFoundation    __NSSingleObjectArrayI_new
  repro             ssl_create
  repro             asock_on_accept_complete2
  repro             pj_ioqueue_poll
```

`0xbeadde…` is freed-memory poison: `CFArrayCreate()` in `set_cert()` retaining an identity that `CFRelease(items)` already deallocated.

## Changes

Three fixes, all backports of what `ssl_sock_apple.m` already does:

1. **`CFRetain()` the borrowed identity** before `CFRelease(items)` drops it — the af82c9085 backport. Guarded against a dictionary carrying no identity, since `CFRetain(NULL)` would itself crash; `ssl_sock_apple.m:579` has no such guard and arguably should.
2. **NULL check on the `SecCertificateCopyValues()` result** in `get_cert_oid()`, present at `ssl_sock_apple.m:1847`, absent here.
3. **Release `cert_data` and `password`** when `CFDictionaryCreate()` fails, instead of leaking both.

Deliberately not included: the 8 KB truncation in `create_data_from_file()`, and deduplication. Both touch `ssl_sock_apple.m` too and so depend on the answer to the open question in #5215 — fix in place twice, or dedupe first and fix once.

## Test plan

Built with `PJ_SSL_SOCK_IMP_DARWIN` forced via `config_site.h`, since `./configure` cannot select this backend on a current SDK (#5217). Same repro on both platforms, five certificate configurations, driving a real TCP connection into each listener.

**iOS Simulator, arm64 — before:** segfault, backtrace above.
**iOS Simulator, arm64 — after:**

| certificate configuration | `start_accept` | first connection |
|---|---|---|
| no certificate configured | `PJ_SUCCESS` | timeout (my client sends no ClientHello) |
| valid .p12, correct password | `PJ_SUCCESS` | timeout — identity loaded, listener healthy |
| valid .p12, wrong password | `PJ_SUCCESS` | OSStatus -25293 |
| unparsable file | `PJ_SUCCESS` | OSStatus -26275 |
| nonexistent file | `PJ_SUCCESS` | `PJ_EINVAL` |

No crash, and a valid `.p12` now loads and serves.

**macOS, arm64:** no behaviour change, before or after. The crash is iOS-specific because the macOS `SecItemImport()` path returns a `SecCertificateRef` and takes the `SecIdentityCreateWithCertificate()` branch, which follows the Create rule and is correctly owned. That is also why the bug survived this long: it is invisible on the platform most likely to be used for development.

Worth recording for anyone testing this backend: the same `.p12` that loads on iOS fails on macOS with `errSecItemNotFound`, because the macOS branch resolves the private key through the keychain rather than from the file. A `.p12` is self-contained on iOS and is not on macOS.

### DIFF
--- a/pjlib/src/pj/ssl_sock_darwin.c
+++ b/pjlib/src/pj/ssl_sock_darwin.c
@@ -273,8 +273,12 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
         options = CFDictionaryCreate(NULL, (const void **)keys,
                                      (const void **)values,
                                      (password? 1: 0), NULL, NULL);
-        if (!options)
+        if (!options) {
+            if (password)
+                CFRelease(password);
+            CFRelease(cert_data);
             return PJ_ENOMEM;
+        }
         
 #if TARGET_OS_IPHONE
         err = SecPKCS12Import(cert_data, options, &items);
@@ -322,6 +326,11 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
                 identity = (SecIdentityRef)
                            CFDictionaryGetValue((CFDictionaryRef) item,
                                                 kSecImportItemIdentity);
+                /* Get rule: this reference is owned by the dictionary, which
+                 * is released along with items below, so take our own.
+                 */
+                if (identity)
+                    CFRetain(identity);
                 break;
             }
 #if !TARGET_OS_IPHONE
@@ -352,8 +361,10 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
         cert_arr[0] = identity;
         cert_refs = CFArrayCreate(NULL, (const void **)cert_arr, 1,
                               &kCFTypeArrayCallBacks);
-        if (!cert_refs)
+        if (!cert_refs) {
+            CFRelease(identity);
             return PJ_ENOMEM;
+        }
     
         err = SSLSetCertificate(dssock->ssl_ctx, cert_refs);
     
@@ -878,6 +889,12 @@ static CFDictionaryRef get_cert_oid(SecCertificateRef cert, CFStringRef oid,
 
     vals = SecCertificateCopyValues(cert, key_arr, NULL);
     dict = CFDictionaryGetValue(vals, key[0]);
+    if (!dict) {
+        CFRelease(key_arr);
+        CFRelease(vals);
+        return NULL;
+    }
+
     *value = CFDictionaryGetValue(dict, kSecPropertyKeyValue);
 
     CFRelease(key_arr);


pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -878,6 +887,12 @@ static CFDictionaryRef get_cert_oid(SecCertificateRef cert, CFStringRef oid,
 
     vals = SecCertificateCopyValues(cert, key_arr, NULL);
     dict = CFDictionaryGetValue(vals, key[0]);
```
sauwming, line 891: The commit message says this "backports the NULL check on the `SecCertificateCopyValues()` result", but the guard added below tests `dict` -- the result of the dictionary *lookup* -- not `vals`. If `SecCertificateCopyValues()` returns NULL, this line dereferences it and crashes before the new check ever runs, and the `CFRelease(vals)` inside that check would too. `key_arr` is unchecked as well.

It is a faithful mirror of `ssl_sock_apple.m:1845-1851`, so keeping the same shape is defensible and I am not asking you to close the pre-existing hole here. But please reword the commit message so it does not claim a NULL check that isn't there -- "backports the missing-value guard in `get_cert_oid()`" or similar.

sauwming (commented):
Thanks -- the retain fix is correct and is the right shape: `CFRetain()` in the dictionary branch balances the single `CFRelease()` in `set_cert()`, and the `if (identity)` guard covers a dictionary that carries no identity. The `cert_data`/`password` release on the `CFDictionaryCreate()` failure path is right too. I compiled the file against both the macOS and iPhoneSimulator SDKs with `PJ_SSL_SOCK_IMP_DARWIN` forced and it produces the same warnings as master, no new ones.

One wording issue on the third change, inline. This should land ahead of #5216, which currently carries this commit.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Reworded, and you are right that the claim was wrong in a way that mattered: the guard tests the lookup result, not `SecCertificateCopyValues()`. The message now reads "backports the missing-value guard in `get_cert_oid()`" and says explicitly that the NULL-`vals` hole is pre-existing and left as it stands in both backends.

I have kept the shape mirroring `ssl_sock_apple.m:1845-1851` rather than closing the hole here, as you suggested. It belongs with the other `create_data_from_file()`/`get_cert_*` work in #5215 — where the real answer is probably to stop having two copies at all.

sauwming (approved):

nanangizz (approved):

PR #5222, opened 2026-08-29
pjlib: read the whole certificate file in the Apple SSL backends

laconicman:
Fixes the truncating read from #5215, and the "or, better, a separate fix" you asked for on #5216.

`create_data_from_file()` issued one `CFReadStreamRead()` into an 8 KB stack buffer and treated whatever came back as the whole file. Anything larger was silently truncated, then failed inside `SecItemImport()`/`SecPKCS12Import()` as an unparsable file rather than an oversized one. `CFReadStreamRead()` is also not obliged to fill the buffer in a single call, so a short read could truncate a smaller file too.

It now reads in a loop into a `CFMutableDataRef` until end of stream, and distinguishes a read error from an empty file.

Two things that make this worth more than its size:

- **It reads `CA_file` as well as `cert_file`**, so the same 8 KB ceiling applied to anyone pinning to a CA, not just to server certificates.
- **A `.p12` carrying an intermediate chain passes 8 KB routinely.** Your point on #5216 was the right one: with eager validation this stops being a per-connection handshake failure and becomes a listener that will not start, which is the most likely shape of the first regression report.

Fixed in both `ssl_sock_darwin.c` and `ssl_sock_apple.m`, which carry byte-identical copies. Fixing one would have left the other with the ceiling — this is the same duplication #5215 is about, and the open question there (fix in place twice, or dedupe first) is unchanged by this PR; I just did not want the fix to wait on it.

## Verified

iOS Simulator, arm64, `PJ_SSL_SOCK_IMP_DARWIN`, measured on top of #5216 so that `pj_ssl_sock_start_accept()` actually loads the certificate:

| certificate | without this fix | with this fix |
|---|---|---|
| 2.4 KB `.p12` | loads | loads |
| 12 KB `.p12` with chain | **fails, OSStatus -26275** | loads |

The 12 KB bundle is a 4096-bit leaf plus a six-certificate chain — nothing exotic.

Worth noting for anyone reproducing on master alone: the test shows nothing there, because without #5216 `pj_ssl_sock_start_accept()` never reads the file, so both sizes report success and the truncation only surfaces later, per connection. I measured against master first and got two false passes before spotting that.

Compile-checked for `PJ_SSL_SOCK_IMP_DARWIN` and `PJ_SSL_SOCK_IMP_APPLE`; same diagnostics as master, no new ones.

### DIFF
--- a/pjlib/src/pj/ssl_sock_darwin.c
+++ b/pjlib/src/pj/ssl_sock_darwin.c
@@ -155,13 +155,22 @@ static pj_ssl_sock_t *ssl_alloc(pj_pool_t *pool)
     return (pj_ssl_sock_t *)PJ_POOL_ZALLOC_T(pool, darwinssl_sock_t);
 }
 
+/* Certificates and PKCS#12 bundles are small. This bound is far above any
+ * legitimate one and exists so that a cert_file or CA_file which is not a
+ * regular file -- a FIFO with no writer, or /dev/zero -- cannot block the
+ * calling thread indefinitely or grow the buffer until the process dies.
+ */
+#define MAX_CERT_FILE_SIZE      (1024 * 1024)
+
 static pj_status_t create_data_from_file(CFDataRef *data,
                                          pj_str_t *fname, pj_str_t *path)
 {
     CFURLRef file;
     CFReadStreamRef read_stream;
+    CFMutableDataRef buf;
     UInt8 data_buf[8192];
     CFIndex nbytes = 0;
+    pj_bool_t too_big = PJ_FALSE;
     
     if (path) {
         CFURLRef filepath;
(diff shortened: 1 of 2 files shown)


pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -209,17 +210,36 @@ static pj_status_t create_data_from_file(CFDataRef *data,
...
+    }
+
+    /* Read until end of stream: one CFReadStreamRead() returns at most
+     * sizeof(data_buf) bytes and is not obliged to fill the buffer even
+     * when more is available, so a single call would silently truncate.
+     */
+    while ((nbytes = CFReadStreamRead(read_stream, data_buf,
+                                      sizeof(data_buf))) > 0)
+    {
+        CFDataAppendBytes(buf, data_buf, nbytes);
```
sauwming, line 246: The loop has no upper bound, where the old single-read had an implicit 8 KB ceiling. If `cert_file`/`CA_file` is misconfigured to point at something that isn't a regular file — a FIFO with no writer, or `/dev/zero`/`/dev/urandom` — `CFReadStreamOpen()` still succeeds, and this loop either blocks the calling thread indefinitely or grows `buf` until the process is OOM-killed (`CFDataAppendBytes()` returns void, so an allocation failure can't be detected and CF aborts). With eager validation this runs on the `pj_ssl_sock_start_accept()`/handshake path, i.e. in a worker thread.

Could you cap the total at a sane maximum (a few hundred KB is far beyond any real `.p12` chain) and return an error past it? Same applies to the copy in `ssl_sock_apple.m`.

laconicman:
Agreed — I removed the only bound the function had and did not replace it. The FIFO case is the one that worries me most: `CFReadStreamOpen()` on a FIFO with no writer succeeds, and the loop then blocks a worker thread indefinitely on the `pj_ssl_sock_start_accept()` path, which after #5216 is reached during `pjsip_tls_transport_start()`.

I will cap the accumulated total and fail past it. Proposed:

```c
/* Certificates and PKCS#12 bundles are small; this is far above any
 * legitimate one and exists to bound a cert_file that is not a regular
 * file, e.g. a FIFO or /dev/zero.
 */
#define MAX_CERT_FILE_SIZE  (1024 * 1024)
```

returning `PJ_ETOOBIG` once `CFDataGetLength(buf)` passes it. A `.p12` carrying a full chain is tens of KB, so 1 MB leaves several orders of magnitude of headroom while still bounding the pathological cases. Say if you would prefer a tighter number. I could not find an existing `PJ_SSL_CERT_MAX_*`-style bound anywhere in the tree to reuse, so this introduces a new named constant rather than following one.

Bounding the total is sufficient on its own — the loop condition is `nbytes > 0`, so a stream that returns 0 terminates it, and any iteration that continues has appended at least one byte. That means the total cap also bounds the iteration count, and no separate iteration limit is needed.

I cannot do anything about `CFDataAppendBytes()` returning void, so allocation failure still aborts inside CoreFoundation rather than surfacing as a `pj_status_t`. The cap is the mitigation: it keeps the buffer far below the size where that becomes plausible.

Applying the same change to both copies in this PR, as you note. If #5224 lands first this collapses to one site, which is rather the point of that PR.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Done — capped at `MAX_CERT_FILE_SIZE` (1 MB), returning `PJ_ETOOBIG` past it, in both copies. Bounding the accumulated total also bounds the iteration count, since the loop condition is `nbytes > 0` so every continuing iteration appends at least one byte. Rationale for the number and the note that `CFDataAppendBytes()` gives no allocation-failure signal are in the PR comment.

sauwming (approved):

nanangizz (approved):

PR #5223, opened 2026-08-29
pjlib: implement the Apple SSL backends in the CMake build

laconicman:
Covers the CMake half of item 3 in #5217. Independent of the autoconf side and of the autodetect question in #5221 — CMake has no autodetect, the backend is named explicitly, so this is the same fix whichever way that decision goes.

Almost everything was already in place: `PJLIB_WITH_SSL` accepts `darwin` and `apple` (`pjlib/CMakeLists.txt:23-26`), `string(TOUPPER "PJ_SSL_SOCK_IMP_${PJLIB_WITH_SSL}")` derives the macro correctly, and `ssl_sock_apple.m` is already compiled under `if(APPLE)`. Both selections then ran into `message(FATAL_ERROR "TODO")`, so neither Apple TLS backend could be built with CMake at all:

```
$ cmake -S . -B build -DPJLIB_WITH_SSL=apple
CMake Error at pjlib/CMakeLists.txt:308 (message):
  TODO
```

The only missing piece was the framework linkage. `Security` for Secure Transport; `Security`, `Network` and `Foundation` for the Network.framework backend. CoreFoundation is already linked for `APPLE` at line 244-250.

The `darwin` branch was also duplicated at 303 and 305, so the second copy was unreachable — dropped.

`schannel` is left as a stub, deliberately. Verifying that linkage needs a Windows host I do not have, and a guess seemed worse than an honest TODO.

## Verified

| `-DPJLIB_WITH_SSL=` | configure | `PJ_SSL_SOCK_IMP` in generated `os_auto.h` | build `pjlib-test` |
|---|---|---|---|
| `darwin` | ok | `PJ_SSL_SOCK_IMP_DARWIN` | links |
| `apple` | ok | `PJ_SSL_SOCK_IMP_APPLE` | links |

Both were `FATAL_ERROR` at configure before. I also checked the selected backend object is not an empty translation unit — `ssl_sock_darwin.c.o` and `ssl_sock_apple.m.o` carry 496 and 559 symbols respectively — since every `ssl_sock_*` source compiles unconditionally and self-gates on the macro, so a wrong macro would still "build" and produce nothing.

`Foundation` is in that list because linking found it, not because I predicted it: configure and the static-library build both passed without it, and only the executable link surfaced

```
  "_OBJC_CLASS_$_NSLock", referenced from:
       in libpjlib.a[56](ssl_sock_apple.m.o)
```

`ssl_sock_apple.m` uses `NSLock` for its event-manager lock (`:140`, `:193`).

### DIFF
--- a/pjlib/CMakeLists.txt
+++ b/pjlib/CMakeLists.txt
@@ -301,11 +301,35 @@ elseif(PJLIB_WITH_SSL STREQUAL "mbedtls")
       MbedTLS::mbedx509
   )
 elseif(PJLIB_WITH_SSL STREQUAL "darwin")
-  message(FATAL_ERROR "TODO")
-elseif(PJLIB_WITH_SSL STREQUAL "darwin")
-  message(FATAL_ERROR "TODO")
+  if(NOT APPLE)
+    message(FATAL_ERROR "PJLIB_WITH_SSL=darwin requires an Apple platform")
+  endif()
+  find_library(Security_LIBRARY Security REQUIRED)
+  target_link_libraries(pjlib-ssl INTERFACE "${Security_LIBRARY}")
 elseif(PJLIB_WITH_SSL STREQUAL "apple")
-  message(FATAL_ERROR "TODO")
+  if(NOT APPLE)
+    message(FATAL_ERROR "PJLIB_WITH_SSL=apple requires an Apple platform")
+  endif()
+  # ssl_sock_apple.m delivers every asynchronous event through
+  # ssl_network_event_poll(), which is only called from ioqueue_select.c.
+  # With any other ioqueue the Network.framework event queue is never
+  # drained and no connection ever completes, so reject the combination
+  # here rather than at runtime.
+  if(NOT PJLIB_WITH_IOQUEUE STREQUAL "select")
+    message(FATAL_ERROR
+      "PJLIB_WITH_SSL=apple requires PJLIB_WITH_IOQUEUE=select "
+      "(got '${PJLIB_WITH_IOQUEUE}')")
+  endif()
+  find_library(Security_LIBRARY Security REQUIRED)
+  find_library(Network_LIBRARY Network REQUIRED)
+  # ssl_sock_apple.m uses NSLock
+  find_library(Foundation_LIBRARY Foundation REQUIRED)
+  target_link_libraries(pjlib-ssl
+    INTERFACE
+      "${Security_LIBRARY}"
+      "${Network_LIBRARY}"
+      "${Foundation_LIBRARY}"
+  )
 elseif(PJLIB_WITH_SSL STREQUAL "schannel")
   message(FATAL_ERROR "TODO")
 endif()


pjlib/CMakeLists.txt
```diff
@@ -301,11 +301,19 @@ elseif(PJLIB_WITH_SSL STREQUAL "mbedtls")
...
+  find_library(Security_LIBRARY Security REQUIRED)
+  find_library(Network_LIBRARY Network REQUIRED)
+  # ssl_sock_apple.m uses NSLock
+  find_library(Foundation_LIBRARY Foundation REQUIRED)
+  target_link_libraries(pjlib-ssl
+    INTERFACE
+      "${Security_LIBRARY}"
+      "${Network_LIBRARY}"
+      "${Foundation_LIBRARY}"
+  )
```
sauwming, line 332: The `apple` backend only works with the `select` ioqueue. `ssl_sock_apple.m` drives all of its async callbacks from `ssl_network_event_poll()`, and the only caller in the tree is `ioqueue_select.c:975` (guarded by `PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_APPLE`); `ioqueue_kqueue.c` and `ioqueue_epoll.c` never call it.

Since `PJLIB_WITH_IOQUEUE` accepts `kqueue` on macOS, `-DPJLIB_WITH_SSL=apple -DPJLIB_WITH_IOQUEUE=kqueue` is now an accepted combination: it configures, compiles and links cleanly, but the Network.framework event list is never drained, so no TLS connection ever completes. This was unreachable while the branch was `FATAL_ERROR "TODO"`. The autoconf path doesn't hit it because it always ends up on `select`.

Could you add a guard here, e.g. a `FATAL_ERROR` when `PJLIB_WITH_IOQUEUE` is not `select`, or force it to `select` for this backend?

pjlib/CMakeLists.txt
```diff
@@ -301,11 +301,19 @@ elseif(PJLIB_WITH_SSL STREQUAL "mbedtls")
       MbedTLS::mbedx509
   )
 elseif(PJLIB_WITH_SSL STREQUAL "darwin")
-  message(FATAL_ERROR "TODO")
-elseif(PJLIB_WITH_SSL STREQUAL "darwin")
-  message(FATAL_ERROR "TODO")
+  find_library(Security_LIBRARY Security REQUIRED)
+  target_link_libraries(pjlib-ssl INTERFACE "${Security_LIBRARY}")
```
sauwming, line 308: Minor: neither this branch nor the `apple` one below is guarded by `if(APPLE)`, and `pj_option` accepts both values on every platform. On Linux the build now fails with `Could NOT find Security_LIBRARY` rather than a message naming the actual mistake.

The sharper edge is that `src/pj/ssl_sock_apple.m` is only added to the target inside `if(APPLE)` (line 207), so if `find_library` does happen to resolve on a non-Apple host, configure succeeds, `PJ_SSL_SOCK_IMP_APPLE` is written into `os_auto.h`, and every `pj_ssl_sock_*` symbol is undefined at link with no explanation. An `APPLE` check around both branches with a clear `FATAL_ERROR` otherwise would rule that out.

laconicman:
Both correct, and the first is a genuinely nasty one — thank you. I confirmed the coupling: `ssl_network_event_poll()` exists in exactly two places, its definition in `ssl_sock_apple.m:276` and its single call in `ioqueue_select.c:975` under `#if PJ_HAS_SSL_SOCK && PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_APPLE`. Neither `ioqueue_kqueue.c` nor `ioqueue_epoll.c` mentions it. So `apple` + kqueue builds, links, and then never completes a handshake, and this PR is what makes that combination selectable.

On where the guard belongs, I would like your steer, because I think the CMake-only version you suggested is the weaker of two options.

**A. CMake guard**, as you proposed — `FATAL_ERROR` when `PJLIB_WITH_IOQUEUE` is not `select`. Local to this PR, friendly message, but it only covers the build system this PR touches.

**B. A compile-time check in `pjlib/include/pj/config.h`**, after both `PJ_SSL_SOCK_IMP` and `PJ_IOQUEUE_IMP` are resolved:

```c
#if defined(PJ_HAS_SSL_SOCK) && PJ_HAS_SSL_SOCK != 0 && \
    PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_APPLE && \
    PJ_IOQUEUE_IMP != PJ_IOQUEUE_IMP_SELECT
#   error Apple Network.framework SSL requires the select ioqueue
#endif
```

That covers autoconf, CMake, and a hand-written `config_site.h` alike — and `config_site.h` is currently the *only* supported way to select this backend (#5217), so it is the path most likely to hit the trap. There is precedent for exactly this shape at `config.h:1390`:

```c
#elif PJ_FILE_IO == PJ_FILE_IO_ANSI && PJ_IOQUEUE_IMP == PJ_IOQUEUE_IMP_IOCP
#   error IOCP ioqueue does not support ANSI file backend
```

B is a change to shared config rather than to the CMake files, so it is a wider blast radius than this PR currently has. I am happy to do B alone, A alone, or both (B for correctness, A to fail earlier with a clearer message) — tell me which you would accept and I will push it.

The `if(APPLE)` point is straightforwardly right and I will add it regardless. Your second-order observation is the sharper half: because `ssl_sock_apple.m` is only added to the target inside `if(APPLE)` at line 207, a non-Apple host where `find_library` happens to resolve gets a successful configure, `PJ_SSL_SOCK_IMP_APPLE` in `os_auto.h`, and then every `pj_ssl_sock_*` symbol undefined at link with nothing pointing at the cause.

pjlib/CMakeLists.txt
laconicman: Guard added, defaulting cleanly: `PJLIB_WITH_IOQUEUE` already defaults to `select`, so it is inert unless someone asks for the broken pair.

```
CMake Error at pjlib/CMakeLists.txt:319 (message):
  PJLIB_WITH_SSL=apple requires PJLIB_WITH_IOQUEUE=select (got 'kqueue')
```

I have implemented the CMake-scoped version you asked for. The stronger `#error` in `config.h` — which would also cover autoconf and hand-written `config_site.h`, the latter being the only supported way to select this backend today — is written and on offer, but it touches shared config so I would rather you chose it than have me widen this PR unasked. Details in the PR comment.

pjlib/CMakeLists.txt
laconicman: Both branches now require `APPLE` with an explicit `FATAL_ERROR`. Your second-order point was the one that decided the wording: the real hazard is not the confusing `find_library` message but a non-Apple host where it happens to resolve, giving a successful configure and `PJ_SSL_SOCK_IMP_APPLE` in `os_auto.h` while `ssl_sock_apple.m` is never added to the target — every `pj_ssl_sock_*` symbol undefined at link with nothing naming the cause.

laconicman:
One scope note I should make explicit rather than leave implied, since it argues for the `config.h` option over the CMake one.

The guard as committed here protects **CMake builds only**. An autoconf build, or a hand-written `config_site.h`, can still select `PJ_SSL_SOCK_IMP_APPLE` alongside a non-`select` ioqueue and get a library that links and never completes a handshake.

That is not hypothetical for me: `config_site.h` is the *only* supported way to select this backend today (#5217), so every build that actually uses it — including my own — is in precisely the unguarded case. The CMake guard protects the path nobody can currently use to select the backend, and misses the path everyone must use.

So my suggestion is firmer than before: the `#error` in `config.h`, after both macros resolve, is the one that covers the real exposure, with precedent both at `config.h:1390` and in `ioqueue_select.c:43`. I have left it out pending your call rather than widening this PR unasked, but I would rather it were in than out.

sauwming (approved):

nanangizz (approved):

PR #5224, opened 2026-08-29
pjlib: share the certificate helpers between the two Apple SSL backends

laconicman:
> **Stacked on #5220 and #5222.** Those remove the divergences between the two copies; this is the move that stops them recurring. Review the top commit only.

This is the concrete proposal for the deduplication question in #5215 — it seemed easier to judge a diff than a description. Happy to close it if you would rather keep the copies.

`ssl_sock_darwin.c` and `ssl_sock_apple.m` carry byte-identical copies of `create_data_from_file()`, `get_info_and_cn()`, `get_cert_oid()` and `get_cert_info()` — close to 400 lines in each file. This moves them into `ssl_sock_apple_common.c`, `#include`d by both, the same way `ssl_sock_imp_common.c` already is.

## Why this is worth doing

Every drift between the two copies has been a defect, and each one had to be found twice:

| defect | fixed in `apple.m` | fixed in `darwin.c` |
|---|---|---|
| identity over-release (crash) | af82c9085, Aug 2023 | #5220, three years later |
| missing-value guard in `get_cert_oid()` | present from the start | #5220 |
| `create_data_from_file()` truncating at 8 KB | #5222 | #5222 — same fix, twice |

The first of those was a segfault on iOS on any valid PKCS#12, sitting in the tree for three years because the fix landed in one copy and nobody diffed the other.

## What is not merged

- **`sslGetCipherSuiteString()`** stays duplicated. Those copies have genuinely diverged — 250 lines against 185, enumerating different cipher-suite sets — and merging them would mean a shared function with backend conditionals inside, which is worse than the duplication.
- **`pj_status_from_err()`** stays per-backend: it takes the backend-specific socket type.

Once #5220 and #5222 are applied, the remaining differences in the moved code were a double space in a `for` statement and a reflowed comment. I kept the `ssl_sock_darwin.c` spelling of both; the diff is otherwise a pure move.

## Build system

Nothing to register. `pjlib/build/Makefile` and `pjlib/CMakeLists.txt` both enumerate sources explicitly, and `ssl_sock_imp_common.c` is absent from both for exactly this reason, so an `#include`-only file is not picked up as its own translation unit. `THIS_FILE` still resolves to the including backend, so log lines continue to name `ssl_sock_darwin.c` or `ssl_sock_apple.m` rather than the shared file.

## Verified

- Compiles for `PJ_SSL_SOCK_IMP_DARWIN` and `PJ_SSL_SOCK_IMP_APPLE` with the same diagnostics as before the move — no new warnings.
- Builds and links `pjlib-test` through CMake for both `-DPJLIB_WITH_SSL=darwin` and `-DPJLIB_WITH_SSL=apple` (with #5223 applied, since those selections are `FATAL_ERROR` stubs on master).
- Functionally exercised on the iOS Simulator, stacking #5216 so that `pj_ssl_sock_start_accept()` actually runs the moved loading path: a 2.4 KB `.p12` and a 12 KB `.p12` with an intermediate chain both load, confirming the moved `create_data_from_file()` still reads past 8 KB.

Net: 751 lines removed, 446 added.


laconicman:
The two failures here look like flakes rather than consequences of this change, and I would like a re-run before I chase them further. Evidence:

**This change cannot reach the failing build.** The failing job is `OpenSSL / pjlib,util,pjnath,pjmedia`, which builds with `PJ_SSL_SOCK_IMP : 1` (OpenSSL). Under that macro both files this PR touches compile to empty translation units, and they are byte-identical to master:

| compiled with `PJ_SSL_SOCK_IMP_OPENSSL` | master | this PR |
|---|---|---|
| `ssl_sock_darwin.c` | 1 symbol | 1 symbol |
| `ssl_sock_apple.m` | 2 symbols | 2 symbols |

`ssl_sock_apple_common.c` is `#include`-only and never becomes its own object — the `ar` line in that job's log lists `ssl_sock_darwin.o` and `ssl_sock_apple.o` but no `ssl_sock_apple_common.o`, exactly as intended.

**The failure does not look backend-related.** It aborts about 30 ms into `ssl_sock_test`:

```
21:25:44.103 Shuffling tests, random seed=102
21:25:44.103 Performing 2 features tests with 2 worker threads
Assertion failed: (handler != ((void*)0)), function pj_throw_exception_, file except.c, line 49.
```

That is a pjlib exception raised on a thread with no `PJ_TRY` handler registered, in a shuffled two-worker run — a threading/ordering problem in the harness rather than anything about certificate loading. The seed is random per run.

**The sibling PRs are green.** #5220 and #5222 touch the same two files and this PR is stacked directly on them; both passed. The only delta here is the code move.

Could you re-run the two failed jobs? I have also opened the same branch on my fork (laconicman/pjproject#11) to get an independent run. If it reproduces there I will dig in properly rather than hand-waving at the seed.

Separately, worth flagging regardless of this PR: if `pj_throw_exception_` can be reached from a `ssl_sock_test` worker thread without a handler, that is a latent abort in the test harness that a different shuffle order could hit on master too.

laconicman:
Pushed a whitespace pass: `git diff --check` was flagging 41 lines in the new file, all trailing whitespace inherited from the sources it was lifted from — `ssl_sock_darwin.c` on master has 164 such lines, they only get flagged here because a new file's lines count as added. Stripped in the new file, plus the extra blank line at EOF; now zero.

That makes the move verbatim *modulo trailing whitespace* rather than byte-identical, so for review: `git diff -w` against the branch parent shows all four functions unchanged. I verified each individually.

Also filed on #5215: the review found a **fourth** duplicated defect, an early-return leak in `get_cert_info()` present in both copies. Kept out of this PR to keep it a pure move.

The re-run request above still stands — the two failures look unrelated to this change.

pjlib/src/pj/ssl_sock_apple_common.c
```diff
@@ -0,0 +1,430 @@
...
+{
+    void *key[1];
+    CFArrayRef key_arr;
+    CFDictionaryRef vals, dict;
+
+    key[0] = (void *)oid;
+    key_arr = CFArrayCreate(NULL, (const void **)key, 1,
+                            &kCFTypeArrayCallBacks);
+
+    vals = SecCertificateCopyValues(cert, key_arr, NULL);
```
sauwming, line 196: Pre-existing, carried over: `SecCertificateCopyValues()` returns NULL when the certificate can't be parsed, and the result is passed straight to `CFDictionaryGetValue()` — NULL deref on a malformed or unsupported certificate. The `!dict` guard just below covers the sibling case but not this one.

Now that there's a single copy, this is a one-line fix here instead of two later:

```c
    vals = SecCertificateCopyValues(cert, key_arr, NULL);
    if (!vals) {
        CFRelease(key_arr);
        return NULL;
    }
    dict = CFDictionaryGetValue(vals, key[0]);
```

pjlib/src/pj/ssl_sock_apple_common.c
```diff
@@ -0,0 +1,430 @@
...
+        serial = SecCertificateCopySerialNumberData(cert, NULL);
+        if (serial) {
+            serial_no = CFDataGetBytePtr(serial);
+            serialsize = CFDataGetLength(serial);
+        }
+    }
+
+    /* Check if the contents need to be updated */
+    update_needed = pj_strcmp2(&ci->issuer.info, buf) ||
+                    pj_memcmp(ci->serial_no, serial_no, serialsize);
```
sauwming, line 267: Pre-existing, carried over: `serialsize` is the raw `CFDataGetLength()` here, but `ci->serial_no` is `pj_uint8_t[20]` (`pjlib/include/pj/ssl_sock.h:270`). The clamp to `sizeof(ci->serial_no)` only happens later, at the `pj_memcpy()` on line 282.

A peer certificate with a serial number longer than 20 octets — non-conforming CAs do emit these — makes this compare read past the array into adjacent struct members, so `update_needed` gets decided on garbage. Moving the existing clamp up to right after `serialsize = CFDataGetLength(serial);` fixes both sites at once.

pjlib/src/pj/ssl_sock_apple_common.c
```diff
@@ -0,0 +1,430 @@
...
+    if (!update_needed) {
```
sauwming, line 268: Pre-existing, carried over: this early return releases only `issuer_info`. `serial` is +1 from `SecCertificateCopySerialNumberData()` and, on macOS, `issuer_cn` is +1 from `CFStringCreateCopy()` — both are released only on the update path below (lines 271/278).

This isn't a cold path: `ssl_update_certs_info()` is called again from `ssl_sock_imp_common.c:996` on every completed renegotiation, and that second call is exactly the `!update_needed` branch. So a long-lived TLS connection leaks a CFData (plus a CFString on macOS) per renegotiation.

pjlib/src/pj/ssl_sock_apple_common.c
```diff
@@ -0,0 +1,430 @@
...
+    /* Serial number */
+    if (serial) {
+        if (serialsize > sizeof(ci->serial_no))
+            serialsize = sizeof(ci->serial_no);
+        pj_memcpy(ci->serial_no, serial_no, serialsize);
+        CFRelease(serial);
+    }
+
+    /* Subject */
+    str = SecCertificateCopySubjectSummary(cert);
```
sauwming, line 311: Pre-existing, carried over: `SecCertificateCopySubjectSummary()` returns NULL for a certificate with an empty subject — common for SAN-only leaf certs — and the result goes straight into `CFStringGetCString()`, which crashes on NULL.

pjlib/src/pj/ssl_sock_apple_common.c
```diff
@@ -0,0 +1,430 @@
...
+#endif
+
+    /* Subject Alternative Name extension */
+#if !TARGET_OS_IPHONE
+{
+    CFArrayRef altname;
+    CFIndex i;
+
+    dict = get_cert_oid(cert, kSecOIDSubjectAltName, (CFTypeRef *)&altname);
+    if (!dict || !CFArrayGetCount(altname))
```
sauwming, line 374: Pre-existing, carried over, two issues in one line:

1. `altname` is the `*value` out-param from `CFDictionaryGetValue(dict, kSecPropertyKeyValue)` in `get_cert_oid()`, which can be NULL even when `dict` is non-NULL. `CFArrayGetCount(NULL)` crashes.
2. When the SAN array is present but empty, this returns without `CFRelease(dict)` — leaks the values dictionary for every such certificate.

sauwming (commented):
Thanks for this — consolidating the two copies is the right move.

I verified the refactor itself is sound:

- **The move is pure.** Extracting the removed lines from `ssl_sock_darwin.c` / `ssl_sock_apple.m` and diffing them against the new `ssl_sock_apple_common.c` shows only trailing-whitespace and one reflowed comment as differences. No logic changed, and no function is lost from either backend.
- **Compiles clean.** `clang -fsyntax-only -Wall` against both the macOS and iPhoneSimulator SDKs, for `PJ_SSL_SOCK_IMP_DARWIN`, `PJ_SSL_SOCK_IMP_APPLE` and `PJ_SSL_SOCK_IMP_OPENSSL`. Warning output is byte-identical to master in every configuration, and both backends still reduce to empty translation units under OpenSSL.
- **Build systems are fine.** Neither `pjlib/build/Makefile` nor `pjlib/CMakeLists.txt` globs sources, so the new `.c` is not picked up as its own TU — same precedent as `ssl_sock_imp_common.c`. The `#include` sits inside each file's `#if PJ_SSL_SOCK_IMP == ...` guard and after `TargetConditionals.h`, so there are no duplicate statics when CMake compiles both files on Apple.

No defect in the refactor. What I did find are pre-existing bugs that the move carries into the shared file verbatim, plus two fixes from d441e5d8a that were applied to `ssl_sock_darwin.c` but never mirrored into `ssl_sock_apple.m` — see the inline comments. None of them block the merge, but since the whole point here is to stop the two copies from drifting, the two `ssl_sock_apple.m` divergences are worth folding in now:

- `ssl_sock_apple.m:514` — `CFRetain(identity)` is unguarded, where d441e5d8a added `if (identity)` to the same code in `ssl_sock_darwin.c:329`. `CFDictionaryGetValue(item, kSecImportItemIdentity)` returns NULL for a PKCS#12 that has certificates but no private key, and `CFRetain(NULL)` aborts — the darwin backend now returns `PJ_EINVAL` with a log line instead.
- `ssl_sock_apple.m:465` — `if (!options) return PJ_ENOMEM;` leaks `cert_data` and `password`; the identical path was fixed in `ssl_sock_darwin.c:211-216`.

(Those two lines fall outside the diff hunks, so I couldn't anchor them inline.)

laconicman:
Thank you for verifying the move independently — extracting and diffing the removed regions is exactly the check I would have wanted a reviewer to run, and it is more convincing than my saying it.

All seven findings confirmed against master. The two `ssl_sock_apple.m` divergences are the ones I want to flag as my own doing:

- `apple.m:579` — `CFRetain(identity)` unguarded. The `if (identity)` in `darwin.c:342` was mine, added in d441e5d8a because `CFDictionaryGetValue` returns NULL for a PKCS#12 with certificates but no key and `CFRetain(NULL)` aborts. I added it to one file.
- `apple.m:530` — `if (!options) return PJ_ENOMEM;` leaking `cert_data` and `password`. Also fixed only in `darwin.c:286-291`, also by me.

So while fixing three-year-old drift between these two files, I introduced two fresh instances of it. That is the argument for this PR in miniature, and I would rather it were in the history than in a review comment.

The five carried-over defects are all real; I checked each:

| line | defect | severity |
|---|---|---|
| :174 | `SecCertificateCopyValues()` NULL dereferenced before the `!dict` guard | crash on an unparsable peer cert |
| :245 | `serialsize` unclamped in `pj_memcmp` against `pj_uint8_t serial_no[20]`; clamp only at the `pj_memcpy` on :282 | over-read; `update_needed` decided on adjacent struct bytes |
| :246 | early return releases `issuer_info` but not `serial` / `issuer_cn` | leak per renegotiation |
| :289 | `SecCertificateCopySubjectSummary()` NULL into `CFStringGetCString()` | crash on a SAN-only leaf |
| :352 | possibly-NULL SAN out-param into `CFArrayGetCount()`; `dict` leaked on the empty-array path | crash, plus leak |

Three of the five are reachable from a peer-supplied certificate during a handshake, which makes them a denial-of-service surface rather than merely untidy.

**Proposed sequencing**, which I think gives the most reviewable history:

1. **A new PR first:** mirror the two fixes above into `ssl_sock_apple.m`. Small, self-contained, verifiable against the existing file.
2. **This PR, rebased on it:** still a pure move, but now of code that is already identical in both files — so the "is this really just a move?" check becomes trivial.
3. **A follow-up after this lands:** the five defects, fixed once in the shared file.

One nuance for that follow-up, which I will put in its description rather than leave implicit: hoisting the serial clamp is not purely a bug fix, it is a small behaviour change. `ci->serial_no` is already truncated to 20 bytes at the `pj_memcpy`, so what changes is the comparison — today it reads `serialsize` bytes from a 20-byte array, so for an over-length serial the result depends on adjacent struct bytes. Afterwards two distinct serials sharing their first 20 octets compare equal and `update_needed` stays false, where the out-of-bounds tail might previously have forced an update. The new behaviour is the correct one for a 20-byte field, but it is observable and should be stated.

That keeps "prove nothing changed" and "change something" in separate commits, and it makes step 3 the demonstration of why step 2 was worth doing — five fixes, one place, instead of ten edits across two files.

I can just as easily fold everything into this PR if you would rather have it in one go; it is your review burden, so your call. Starting on (1) now unless you say otherwise.

laconicman:
All of it is pushed. #5226 is the preparatory PR; this branch is rebased on it and remains a pure move, verified again after the rebase — all four relocated functions are identical to their originals ignoring whitespace.

One correction to what I wrote earlier, because it overstated what #5226 buys: I said rebasing on it would make this "a move of code that is already identical in both files". The four *moved* functions were already identical. What #5226 fixes is drift in `create_identity_from_cert()` / `set_cert()`, which stay per-backend and are not part of the move. So it is a prerequisite for tidiness and a bug fix in its own right, not a precondition for purity. The purity was already there.

Since it may be useful to whoever picks these up next, the misses in this series had a common shape, and it is not one that more care would obviously have caught.

**Everything I got wrong was invisible from the file I was editing.**

- The over-release that #5220 fixed had been sitting in one of two byte-identical copies since 2023. Nothing in `ssl_sock_darwin.c` hints that `ssl_sock_apple.m` contains the same function with an extra line in it.
- Then I did it again in the same series: fixed two paths in `ssl_sock_darwin.c` and left the twins in `ssl_sock_apple.m`. You caught it by diffing the copies, which is the only way it could have been caught.
- The `apple` + kqueue combination is invisible from `pjlib/CMakeLists.txt`, from `ssl_sock_apple.m`, and from `ioqueue_kqueue.c`. It is only visible if you notice that a function defined in one file has exactly one caller in a third file, and then ask what the other two ioqueues do about it.
- The probe overwriting `CFLAGS` looks harmless in `aconfigure.ac` and only matters if you are holding `configure-iphone`'s argument handling in your head at the same time.
- And my `--disable-ssl` guard read correctly on its own; it was wrong only because of an implicit `enable_ssl=no` assigned twenty lines earlier for an unrelated reason.

That is five defects whose common property is that the evidence for each lives in a different file from the code that has to change. It is a fair description of what this corner of the tree costs to work in, and the reason I have been pushing the deduplication: it converts one class of these — the byte-identical twins — into something a compiler can enforce instead of something a reviewer has to remember.

Thank you for the reviews. Extracting the moved regions and diffing them, and compiling across three SDK configurations, is more than I would have expected and it caught things I would not have.

pjlib/src/pj/ssl_sock_apple_common.c
laconicman: Confirmed against master, and deferred deliberately rather than dismissed: it goes in the follow-up PR that fixes all five carried-over defects once in the shared file, after this move lands. Keeping "prove nothing changed" and "change something" in separate diffs is the whole reason the move is reviewable by mechanical diff.

Specifically here: `SecCertificateCopyValues()` NULL is dereferenced by the `CFDictionaryGetValue()` on the next line, before the `!dict` guard that covers the sibling case. Taking your suggested shape.

pjlib/src/pj/ssl_sock_apple_common.c
laconicman: Confirmed against master, and deferred deliberately rather than dismissed: it goes in the follow-up PR that fixes all five carried-over defects once in the shared file, after this move lands. Keeping "prove nothing changed" and "change something" in separate diffs is the whole reason the move is reviewable by mechanical diff.

This one I will flag as a behaviour change rather than a pure fix, in that PR's description: `ci->serial_no` is already truncated at the `pj_memcpy`, so what changes is the comparison. Today it reads `serialsize` bytes from a 20-byte array, so for an over-length serial the result depends on adjacent struct bytes. Afterwards two serials sharing their first 20 octets compare equal and `update_needed` stays false, where the out-of-bounds tail might previously have forced an update. The new behaviour is correct for a 20-byte field, but it is observable.

pjlib/src/pj/ssl_sock_apple_common.c
laconicman: Confirmed against master, and deferred deliberately rather than dismissed: it goes in the follow-up PR that fixes all five carried-over defects once in the shared file, after this move lands. Keeping "prove nothing changed" and "change something" in separate diffs is the whole reason the move is reviewable by mechanical diff.

This is the one I had already filed separately on #5215 before your review; your `ssl_sock_imp_common.c:996` reference is the part I had missed — I had described it as per-renegotiation without pinning down the call site.

pjlib/src/pj/ssl_sock_apple_common.c
laconicman: Confirmed against master, and deferred deliberately rather than dismissed: it goes in the follow-up PR that fixes all five carried-over defects once in the shared file, after this move lands. Keeping "prove nothing changed" and "change something" in separate diffs is the whole reason the move is reviewable by mechanical diff.

Worth noting for severity: SAN-only leaves with an empty subject are entirely normal from modern issuers, so this is not an exotic edge case — it is a crash on a certificate shape in routine use.

pjlib/src/pj/ssl_sock_apple_common.c
laconicman: Confirmed against master, and deferred deliberately rather than dismissed: it goes in the follow-up PR that fixes all five carried-over defects once in the shared file, after this move lands. Keeping "prove nothing changed" and "change something" in separate diffs is the whole reason the move is reviewable by mechanical diff.

Both halves confirmed: the out-param can be NULL while `dict` is not, and the empty-array path returns without releasing `dict`. The short-circuit in `if (!dict || !CFArrayGetCount(altname))` protects the first deref but not the second.

sauwming (approved):

laconicman:
Heads-up before this is merged: **the approved revision was stale and I have rebased it.** Please re-glance rather than taking the earlier approval as carrying over.

When you approved it, the branch carried a snapshot of #5226 taken before its third fix (the `sec_identity_create()` failure path you raised) existed. Merging it as it stood would have landed the older #5226 and quietly dropped that fix. #5222 had also moved underneath it.

It is now rebased onto the current tips of both, and I re-derived the extraction mechanically rather than resolving conflicts by hand. That caught a second problem: #5216 added `create_identity_from_cert()` to `ssl_sock_darwin.c`, so the function following `create_data_from_file()` is no longer `set_cert()` in that file — a naive re-run carried `create_identity_from_cert`'s doc comment into the shared file, orphaned from its function. Fixed; the comment stays with the function.

Re-verified after the rebase: `create_data_from_file` is identical to both originals; `get_info_and_cn`/`get_cert_oid` and `get_cert_info` are identical to `ssl_sock_darwin.c` and differ from `ssl_sock_apple.m` only by the two known cosmetic items (a double space in a `for`, one reflowed comment), resolved toward the darwin spelling as before.

Suggested order: **#5222 → #5226 → #5224**. The first two touch disjoint regions of `ssl_sock_apple.m` (the read loop vs the identity-import path), so they are independent; this one must come last and be re-diffed against whatever actually landed.

sauwming (approved):

nanangizz (approved):

PR #5225, opened 2026-08-31
pjsip: fall back to Via sent-by when "received" param is absent

sauwming:
`pjsip_get_response_addr()` asserts that the Via header of an incoming request carries a `received` parameter, and uses it as the response destination host. The transport layer adds that parameter to every request it accepts, so the assertion holds on the normal receive path, but any caller that builds `rdata` by hand (e.g. after parsing a message with `pjsip_parse_rdata()`) aborts the process instead.

It was reported by OSS-Fuzz because the `fuzz-sip-transaction` harness constructs `rdata` itself and never sets the parameter:

```
fuzz-sip-transaction: src/pjsip/sip_util.c:1780: pjsip_get_response_addr:
    Assertion `rdata->msg_info.via->recvd_param.slen != 0' failed.
    #7 pjsip_get_response_addr sip_util.c:1780
    #8 pjsip_tsx_create_uas2 sip_transaction.c:1840
    #9 test_uas_transaction fuzz-sip-transaction.c:81
```

## Changes

`pjsip/src/pjsip/sip_util.c`:

- Use the Via `sent-by` host when the `received` parameter is missing, as RFC 3261 Section 18.2.2 specifies, so the destination stays correct instead of being an empty host.
- Check `msg_info.via` itself, which was dereferenced unconditionally.

The fuzz harness is deliberately left unchanged, so that the fallback keeps being exercised by the fuzzer.

## Testing

The OSS-Fuzz testcase was run through a standalone driver linked against the unmodified harness: it aborts on master and passes with this change, as does the `fuzz-sip` seed corpus.

### DIFF
--- a/pjsip/src/pjsip/sip_util.c
+++ b/pjsip/src/pjsip/sip_util.c
@@ -1766,6 +1766,7 @@ PJ_DEF(pj_status_t) pjsip_get_response_addr( pj_pool_t *pool,
                                              pjsip_response_addr *res_addr )
 {
     pjsip_transport *src_transport = rdata->tp_info.transport;
+    const pj_str_t *recvd;
 
     /* Check arguments. */
     PJ_ASSERT_RETURN(pool && rdata && res_addr, PJ_EINVAL);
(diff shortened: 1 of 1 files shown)


sauwming:
## Review feedback

**Fuzzer change (`fuzz-sip-transaction.c`) — concern**

Populating `recvd_param` in both `test_uas_transaction()` and `test_create_response()` means the new fallback path (the exact path that crashed) is never exercised by the fuzzer anymore. The library fix alone already resolves the OSS-Fuzz abort, and leaving the harness unchanged would give continuous regression coverage of the fallback instead of making it dead code under fuzzing.

Suggest: drop the harness population in at least one of the two functions (ideally both), so `pjsip_get_response_addr()` fuzzes with a missing `received`. If it's kept for realism, consider a comment acknowledging the coverage tradeoff.

**Minor (not blockers)**

- `rdata->msg_info.via` is dereferenced without a NULL check (`sip_util.c:1782`) — pre-existing behavior (the old assert did the same), not a regression, but worth noting.

sauwming:
## Changes addressing the review

**Harness change dropped entirely.** `tests/fuzz/fuzz-sip-transaction.c` is back to its original form, in both `test_uas_transaction()` and `test_create_response()`, so `pjsip_get_response_addr()` keeps being fuzzed with a missing `received` parameter and the fallback stays covered instead of becoming dead code. The library fix alone is sufficient, as the review notes: the OSS-Fuzz testcase run through a standalone driver linked against the *unmodified* harness aborts on master and passes with this change, and so do all 15 inputs of the `fuzz-sip` seed corpus.

**Via NULL check added.** `PJ_ASSERT_RETURN(rdata->msg_info.via, PJSIP_EMISSINGHDR)`, matching the guard added to `create_tsx_key_2543()` in #5219. Note this is `PJ_ASSERT_RETURN`, so a NULL Via still aborts in assert-enabled builds — the value is documenting the precondition and avoiding the NULL dereference in release builds, not fuzz hardening.

The diff is now library-only:

```
 pjsip/src/pjsip/sip_util.c | 22 +++++++++++++---------
 1 file changed, 13 insertions(+), 9 deletions(-)
```

nanangizz (approved):

PR #5226, opened 2026-08-31
pjlib: apply the Darwin certificate fixes to the Apple backend too

laconicman:
Preparatory PR for #5224, and a fix in its own right.

d441e5d8a (#5220) fixed two defects in `ssl_sock_darwin.c`'s identity import and left the byte-identical code in `ssl_sock_apple.m` untouched — so while fixing three-year-old drift between these two files, I introduced two fresh instances of it. Spotted by @sauwming reviewing #5224.

- **`ssl_sock_apple.m:579`** — `CFRetain(identity)` is unguarded. `CFDictionaryGetValue(item, kSecImportItemIdentity)` returns NULL for a PKCS#12 carrying certificates but no private key, and `CFRetain(NULL)` aborts. `ssl_sock_darwin.c:342` guards it and returns `PJ_EINVAL` with a log line instead.
- **`ssl_sock_apple.m:530`** — `if (!options) return PJ_ENOMEM;` leaks `cert_data` and `password`; fixed in `ssl_sock_darwin.c:286-291`.

With this applied the two copies are functionally identical again, which lets #5224 stay a pure move of code that is already the same in both files, rather than a move that also has to carry fixes. Sequencing per the discussion on that PR: this, then the move, then the five carried-over defects fixed once in the shared file.

Compile-checked for `PJ_SSL_SOCK_IMP_APPLE`; no new diagnostics.

### DIFF
--- a/pjlib/src/pj/ssl_sock_apple.m
+++ b/pjlib/src/pj/ssl_sock_apple.m
@@ -527,8 +527,12 @@ static pj_status_t create_identity_from_cert(applessl_sock_t *assock,
         options = CFDictionaryCreate(NULL, (const void **)keys,
                                      (const void **)values,
                                      (password? 1: 0), NULL, NULL);
-        if (!options)
+        if (!options) {
+            if (password)
+                CFRelease(password);
+            CFRelease(cert_data);
             return PJ_ENOMEM;
+        }
         
 #if TARGET_OS_IPHONE
         err = SecPKCS12Import(cert_data, options, &items);
@@ -576,7 +580,13 @@ static pj_status_t create_identity_from_cert(applessl_sock_t *assock,
                 identity = (SecIdentityRef)
                            CFDictionaryGetValue((CFDictionaryRef) item,
                                                 kSecImportItemIdentity);
-                CFRetain(identity);
+                /* Get rule: this reference is owned by the dictionary, which
+                 * is released along with items below, so take our own. A
+                 * PKCS#12 carrying certificates but no private key yields no
+                 * identity here, and CFRetain(NULL) would abort.
+                 */
+                if (identity)
+                    CFRetain(identity);
                 break;
             }
 #if !TARGET_OS_IPHONE
@@ -605,6 +615,10 @@ static pj_status_t create_identity_from_cert(applessl_sock_t *assock,
         }
 
         *p_identity = sec_identity_create(identity);
+        if (!*p_identity) {
+            CFRelease(identity);
+            return PJ_ENOMEM;
+        }
     
         CFRelease(identity);
     }


pjlib/src/pj/ssl_sock_apple.m
```diff
@@ -576,7 +580,13 @@ static pj_status_t create_identity_from_cert(applessl_sock_t *assock,
...
                            CFDictionaryGetValue((CFDictionaryRef) item,
                                                 kSecImportItemIdentity);
-                CFRetain(identity);
+                /* Get rule: this reference is owned by the dictionary, which
+                 * is released along with items below, so take our own. A
+                 * PKCS#12 carrying certificates but no private key yields no
+                 * identity here, and CFRetain(NULL) would abort.
+                 */
+                if (identity)
+                    CFRetain(identity);
```
sauwming, line 589: Not about this hunk, but about the tail of the same function (line 617) which GitHub won't let me anchor to:

```c
*p_identity = sec_identity_create(identity);

CFRelease(identity);
```

This is the one remaining piece of d441e5d8a that isn't mirrored, so the two functions aren't quite functionally identical yet. On the ownership-transfer failure path Darwin does `CFRelease(identity); return PJ_ENOMEM;`, whereas `sec_identity_create()` returning nil here is not detected: `identity` is released (so no leak) but we fall through to `PJ_SUCCESS` with `*p_identity == NULL`. The caller only checks the status, and `configure_tls()` guards with `if (ssock->cert && assock->identity)`, so the listener would start with no local identity and the failure would only surface later as an opaque handshake error.

```c
*p_identity = sec_identity_create(identity);
if (!*p_identity) {
    CFRelease(identity);
    return PJ_ENOMEM;
}
```

Worth settling here or in #5224, where the two tails have to converge anyway.

sauwming (commented):
Both hunks look correct to me.

- The `!options` path: `cert_data` is guaranteed non-NULL there (the whole block is inside `if (cert_data)`), `password` is properly guarded, and neither object is referenced again after the early return.
- The `CFRetain()` guard: `identity` is initialized to `NULL` and the `break` falls into the existing `if (!identity)` check, so a PKCS#12 with certificates but no private key now returns `PJ_EINVAL` instead of aborting inside `CFRetain(NULL)`. Ownership stays at +1 as in the `SecIdentityCreateWithCertificate()` branch, and the single `CFRelease(identity)` at the tail still balances it.

I diffed the whole function against the copy in `ssl_sock_darwin.c`; after this patch the only remaining differences are the intentional ones (type names, log levels, and the `sec_identity_create()` tail), plus the one noted inline.

pjlib/src/pj/ssl_sock_apple.m
laconicman: Confirmed and fixed — and you are right that it is the piece that made the two functions still not identical.

Traced it through: `sec_identity_create()` returning nil left `*p_identity` NULL, `CFRelease(identity)` ran so nothing leaked, and the function returned `PJ_SUCCESS`. The caller at `:966` checks only the status, so `assock->identity` stayed NULL and `network_start_accept()` proceeded; `configure_tls()` then guards with `if (ssock->cert && assock->identity)` and simply never sets a local identity. The listener comes up, and the first handshake fails with nothing pointing at the cause.

Taken your shape verbatim. That makes three fixes from d441e5d8a mirrored rather than two, and the commit message now says three.

sauwming (approved):

nanangizz (approved):

Issue #5232, opened 2026-09-02
pjlib: register of small Apple TLS backend defects (low priority)


laconicman:
Small defects in the Apple TLS backends found during #5215/#5217/#5229 and the PR series around them, collected here so they are not lost in review threads. **None is urgent** — filing them as a register rather than a request, and happy for any of them to be closed as won't-fix.

Grouped by whether a fix is already committed to.

## Already spoken for — listed only so the register is complete

1. **Five unguarded paths in `get_cert_info()`**, all reachable from a peer-supplied certificate during a handshake: `SecCertificateCopyValues()` NULL dereferenced before its own check; `serialsize` unclamped in the `pj_memcmp` against `pj_uint8_t serial_no[20]`; the `!update_needed` early return leaking `serial` and `issuer_cn` per renegotiation; `SecCertificateCopySubjectSummary()` NULL into `CFStringGetCString()`, reached by SAN-only leaves that modern issuers emit routinely; and a possibly-NULL SAN out-param into `CFArrayGetCount()` plus a `dict` leak on the empty-array path. Raised on #5224; committed to as a follow-up once that lands.

2. **Log level divergence.** `ssl_sock_darwin.c:189` logs "Ignoring supplied private key" at level 3, `ssl_sock_apple.m:434` at level 5 — same condition, same message. On the Apple backend the caller does not get the line at default verbosity, and that is the backend more likely to be in use on Apple platforms. Going into the same follow-up.

3. **`pj_ssl_cert_t` Doxygen** recording which fields each backend honours, and that `cert_file` is imported via `SecPKCS12Import()` and self-contained on iOS while macOS goes through `SecItemImport()` with the key resolved from the keychain. Same follow-up.

4. **TLS version bounds discarded** in `ssl_sock_darwin.c:558,563`. Identified in #5229; `kTLSProtocol11` and `kTLSProtocol12` are documented as legal for `SSLSetProtocolVersionMin`, so an application asking for a 1.2 minimum currently gets a context that will still negotiate 1.0. Will send separately as a behaviour fix.

## Not yet spoken for

[...]

Issue #5229, opened 2026-09-01
pjlib: the TLS configuration surface accepts contradictory combinations, and expresses the same intent differently per build system


laconicman:
### Summary

Working through a series of defects in the Apple TLS backends (#5215, #5217, #5220–#5226) I kept
hitting the same shape of bug rather than unrelated ones. This issue is an attempt to name the
shape, with the evidence gathered along the way, in case it is useful. It proposes one small fix
and sketches two larger ones; the small one is the only thing I would suggest acting on quickly.

The pattern: **TLS configuration is settled by four macros, which four different mechanisms can
set, using three different vocabularies, with no validation that the result is coherent.** Most
invalid combinations are unreachable from any single build path, which is why they have gone
unnoticed — not because they are harmless.

### 1. The surface, counted

Four macros decide whether and how TLS works:

| macro | values | resolved in |
|---|---|---|
| `PJ_HAS_SSL_SOCK` | on/off | `config.h` |
| `PJ_SSL_SOCK_IMP` | 7 (`NONE`, `OPENSSL`, `GNUTLS`, `DARWIN`, `APPLE`, `SCHANNEL`, `MBEDTLS`) | `config.h:1219-1234` |
| `PJ_IOQUEUE_IMP` | 7 (`NONE`, `SELECT`, `EPOLL`, `IOCP`, `KQUEUE`, `UWP`, `SYMBIAN`) | `config.h:727-745` |
| `PJSIP_HAS_TLS_TRANSPORT` | defaults to `PJ_HAS_SSL_SOCK` | `sip_config.h:955-957` |

Four mechanisms can set them, and they do not agree on which backends exist:

| mechanism | backends it can select |
|---|---|
| `./aconfigure` | `DARWIN`, `OPENSSL`, `GNUTLS`, `MBEDTLS` — 4 of 6 (`aconfigure.ac:2162, 2237, 2272, 2307`) |
| CMake `PJLIB_WITH_SSL` | all 6 declared (`pjlib/CMakeLists.txt:23-26`); `darwin`/`apple`/`schannel` were `FATAL_ERROR "TODO"` until #5223, `schannel` still is |
| hand-written `config_site.h` | all 6 |
| `config.h` default | `OPENSSL`, or `NONE` when `PJ_HAS_SSL_SOCK==0` |

Three entry points also compose `CFLAGS` differently before any of this runs: `./aconfigure`
[...]

PR #5233, opened 2026-09-02
pjsip: retry the listener when restarting a TCP or TLS transport that has none

laconicman:
Item 5 of #5232, and the one I would rank highest of the unclaimed items there.

`pjsip_tls_transport_restart2()` returns `PJ_SUCCESS` after updating only the published address whenever `listener->ssock` is NULL. That is right for a `PJSIP_TLS_TRANSPORT_DONT_CREATE_LISTENER` build, which creates a client-only factory on purpose — but the branch is reached in ordinary builds too, where it means something quite different.

In a default build, a *registered* factory with no socket has one cause. The restart path closes the socket, sets it NULL, and keeps the factory registered deliberately so outgoing transports survive — then `pjsip_tls_transport_lis_start()` fails. From that point every later restart takes the address-only branch, reports success, and never rebuilds the listener.

Observed while testing TLS transport restart on iOS: after one restart that failed loading credentials, subsequent restarts returned `PJ_SUCCESS` indefinitely while the listener stayed down. The state is unrecoverable and indistinguishable from a healthy transport.

This got sharper with #5210 and #5216. Now that the certificate is loaded and validated at listener start, a listener whose credentials are briefly unavailable fails to start — and restart is the documented way to recover once they are available. That recovery currently cannot work.

## The change

**One flag, in both transports.** `lis_wanted` is set on entry to `lis_start()` — recording that a
listener was asked for, whether or not that start succeeded and whether or not the socket was later
lost. A restart on a factory that never wanted a listener updates the published address only; a
restart on one that did falls through to the path that reloads credentials, restarts the listener
and returns its real status.

This is applied to `sip_transport_tcp.c` as well as `sip_transport_tls.c`. TCP had the byte-identical
shortcut, `pjsip_tcp_transport_lis_start()` NULLs `asock` on every `on_error` exit, and
`pjsua_transport_lis_restart()` routes TCP restarts into it, so the same bug was there for the same
API call. Fixing only TLS would have left the two transports diverging.

**One credential loader (TLS).** `load_listener_cert()` is the only place that assigns
`listener->cert`, used by both `pjsip_tls_transport_start2()` and `pjsip_tls_transport_restart2()`.
It covers all four sources — files, buffer, store, `cert_direct` — in independent blocks, because
they are cumulative: each loader adds to the same `pj_ssl_cert_t`. Every failure exit clears the
field, so `!listener->cert` reliably means "holds nothing", which is what the reload condition
[...]

PR #5234, opened 2026-09-02
pjlib: apply the TLS version bounds the Darwin backend can actually set

laconicman:
Item 4 of #5232, and the follow-up I promised on #5229 after you corrected my reading of it.

`ssl_create()` computes `min_proto` and `max_proto` from `ssock->param.proto`, then applies them only when they are `<= kTLSProtocol1`:

```c
/* According to the doc, we can't set min/max proto for TLS protocol
 * higher than 1.0. The runtime error given is -9830 (Illegal parameter).
 */
if (min_proto != kSSLProtocolUnknown && min_proto <= kTLSProtocol1) {
```

The -9830 is real, but it applies to one value, not to everything above 1.0. Measured against the current SDK by calling the setter directly:

```
SSLSetProtocolVersionMin(kSSLProtocol3)  -> accepted
SSLSetProtocolVersionMin(kTLSProtocol1)  -> accepted
SSLSetProtocolVersionMin(kTLSProtocol11) -> accepted
SSLSetProtocolVersionMin(kTLSProtocol12) -> accepted
SSLSetProtocolVersionMin(kTLSProtocol13) -> -9830
```

which matches what `SecureTransport.h` documents as legal for `minVersion`. So the guard is three protocol versions too broad, and TLS 1.1 and 1.2 bounds are being dropped for a restriction Secure Transport does not impose.

**Which versions actually lose their floor.** Not the default path: `PJ_SSL_SOCK_PROTO_DEFAULT` expands to `PJ_SSL_SOCK_PROTO_ALL & ~PJ_SSL_SOCK_PROTO_SSL2`, which still carries the SSL3 bit, so `min_proto` is `kSSLProtocol3` and, at `2 <= kTLSProtocol1`'s `4`, the old guard did apply it. The default was never "no minimum" — it was actively pinning the floor to SSL 3.0, below Secure Transport's own default of `kTLSProtocol1`. That is unchanged here.

The floor is genuinely dropped for TLS 1.1 and TLS 1.2 — `PJSIP_TLSV1_2_METHOD`, for instance, maps to a `kTLSProtocol12` minimum that exceeded the guard and was skipped, so an application asking for a TLS 1.2 floor got a context that would still negotiate lower. That is the case this fixes.

## The change

Apply both bounds whenever they are known, and split the TLS 1.3 case in two:

- **A requested floor of TLS 1.3 returns `PJ_EINVAL`.** The caller has stated a policy this
  backend cannot implement. Clamping it down completed a TLS 1.2 handshake while
  `pj_ssl_sock_get_info()` still reported TLS 1.3, so the downgrade was undetectable.
- **A requested ceiling of TLS 1.3 is clamped to 1.2**, since the caller accepts a lower
  version and the result stays inside the requested policy. `PJ_SSL_SOCK_PROTO_DEFAULT` sets
  the TLS 1.3 bit on this backend, making this the common path rather than an anomaly, so it
  logs at level 5.

The first of those is the behaviour change this PR originally flagged as arguably more
correct but did not want to smuggle in. It was requested in review, so it is now what the PR
does.

I checked the other backends before switching, since the argument only holds if it brings
Darwin into line: `ssl_sock_apple.m` and `ssl_sock_mbedtls.c` both return `PJ_EINVAL` for a
protocol selection they cannot provide, while `ssl_sock_ossl.c` (option mask) and
`ssl_sock_gtls.c` (priority string) express the request exactly and never face the choice.
None of them silently narrows a floor.

Failures from `SSLSetProtocolVersionMin/Max` are returned now instead of discarded. A rejected
bound previously left the context on the stack default while `ssl_create()` reported success,
and left a stale `last_native_err` for a later `pj_ssl_sock_get_info()` to report on an
otherwise healthy connection.

Two pjlib-test cases request TLS 1.3 alone and expect success — the TLSv1.3 cipher echo test
and "server TLSv1.2+1.3 vs client TLSv1.3". On this backend both pass today only because of
the silent downgrade, so they are excluded for `PJ_SSL_SOCK_IMP_DARWIN`, alongside the
exclusion already present a few lines below them. The Darwin backend is not the configure
default, so I could not run its suite to confirm their prior state; that part is reasoning
from the code.

**Rejected once, not per connection.** `ssl_create()` runs for every accepted connection, so a server configuration this backend cannot serve used to start cleanly and then fail every inbound handshake — emitting a level 3 log line that any unauthenticated peer could trigger at will, while the operator saw no startup error. The protocol mapping now lives in `get_proto_bounds()` and is also called from `ssl_init_server_ctx()`, the hook this backend already has for failing an unusable server configuration up front.

It validates `newsock_param`, which carries the parameters accepted sockets are actually built from, not `ssock->param` — `pj_ssl_sock_start_accept2()` lets a caller set the two independently. `ssl_init_server_ctx()` therefore takes the parameters as an argument; its declaration is guarded to the OpenSSL and Darwin backends, so that is one call site and two implementations, and the OpenSSL one ignores it. The bounds produced there are discarded: `ssl_create()` still derives its own per connection, so the call is validation only.

Diagnostics: no new warnings against master, and three fewer overall, since the reworked block
names `kTLSProtocol13` fewer times.

### DIFF
--- a/pjlib/src/pj/ssl_sock_darwin.c
+++ b/pjlib/src/pj/ssl_sock_darwin.c
@@ -358,58 +358,18 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
  * still start for a configuration that every connection subsequently
  * rejects -- the check fails open, never closed.
  */
-static pj_status_t ssl_init_server_ctx(pj_ssl_sock_t *ssock)
+/* Map ssock->param.proto onto the Secure Transport version bounds.
+ *
+ * Returns PJ_EINVAL when the requested range cannot be served at all, so an
+ * unusable configuration can be rejected once at listener setup instead of on
+ * every accepted connection.
+ */
+static pj_status_t get_proto_bounds(pj_ssl_sock_t *ssock,
+                                    SSLProtocol *p_min,
+                                    SSLProtocol *p_max)
 {
-    darwinssl_sock_t *dssock = (darwinssl_sock_t *)ssock;
-    SecIdentityRef identity = NULL;
-    pj_status_t status;
-
-    pj_assert(ssock->is_server && !ssock->parent);
-
-    if (!ssock->cert)
-        return PJ_SUCCESS;
-
-    status = create_identity_from_cert(dssock, ssock->cert, &identity);
-    if (identity)
-        CFRelease(identity);
-
-    return status;
-}
-
-/* Create and initialize new Darwin SSL context and instance */
-static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
-{
-    darwinssl_sock_t *dssock = (darwinssl_sock_t *)ssock;
-    SSLContextRef ssl_ctx;
     SSLProtocol min_proto = kSSLProtocolUnknown;
     SSLProtocol max_proto = kSSLProtocolUnknown;
-    OSStatus err;
-    pj_status_t status;
-
-   /* Initialize input circular buffer */
-    status = circ_init(ssock->pool->factory, &ssock->ssl_read_buf, 8192);
-    if (status != PJ_SUCCESS)
-        return status;
-
-    /* Initialize output circular buffer */
-    status = circ_init(ssock->pool->factory, &ssock->ssl_write_buf, 8192);
-    if (status != PJ_SUCCESS)
-        return status;
-
-    /* Create SSL context */
-    ssl_ctx = SSLCreateContext(NULL, (ssock->is_server? kSSLServerSide:
-                                      kSSLClientSide), kSSLStreamType);
-    if (ssl_ctx == NULL)
-        return PJ_ENOMEM;
-
-    dssock->ssl_ctx = ssl_ctx;
-
-    /* Set certificate */
-    if (ssock->cert)  {
-        status = set_cert(dssock, ssock->cert);
-        if (status != PJ_SUCCESS)
-            return status;
-    }
 
     /* Set min and max protocol version */
     if (ssock->param.proto == PJ_SSL_SOCK_PROTO_DEFAULT) {
@@ -469,6 +429,81 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
                               "limiting the maximum version to TLS 1.2"));
         max_proto = kTLSProtocol12;
     }
+
+    *p_min = min_proto;
+    *p_max = max_proto;
+
+    return PJ_SUCCESS;
+}
+
+
+static pj_status_t ssl_init_server_ctx(pj_ssl_sock_t *ssock)
+{
+    darwinssl_sock_t *dssock = (darwinssl_sock_t *)ssock;
+    SecIdentityRef identity = NULL;
+    SSLProtocol min_proto, max_proto;
+    pj_status_t status;
+
+    pj_assert(ssock->is_server && !ssock->parent);
+
+    /* Reject a protocol range this backend cannot serve here, once, rather
+     * than once per accepted connection in ssl_create(). A listener asking
+     * for TLS 1.3 alone would otherwise start successfully and then fail
+     * every inbound handshake, emitting a log line any unauthenticated peer
+     * can trigger while the operator never sees a startup error.
+     */
+    status = get_proto_bounds(ssock, &min_proto, &max_proto);
+    if (status != PJ_SUCCESS)
+        return status;
+
+    if (!ssock->cert)
+        return PJ_SUCCESS;
+
+    status = create_identity_from_cert(dssock, ssock->cert, &identity);
+    if (identity)
+        CFRelease(identity);
+
+    return status;
+}
+
+/* Create and initialize new Darwin SSL context and instance */
+static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
+{
+    darwinssl_sock_t *dssock = (darwinssl_sock_t *)ssock;
+    SSLContextRef ssl_ctx;
+    SSLProtocol min_proto = kSSLProtocolUnknown;
+    SSLProtocol max_proto = kSSLProtocolUnknown;
+    OSStatus err;
+    pj_status_t status;
+
+   /* Initialize input circular buffer */
+    status = circ_init(ssock->pool->factory, &ssock->ssl_read_buf, 8192);
+    if (status != PJ_SUCCESS)
+        return status;
+
+    /* Initialize output circular buffer */
+    status = circ_init(ssock->pool->factory, &ssock->ssl_write_buf, 8192);
+    if (status != PJ_SUCCESS)
+        return status;
+
+    /* Create SSL context */
+    ssl_ctx = SSLCreateContext(NULL, (ssock->is_server? kSSLServerSide:
+                                      kSSLClientSide), kSSLStreamType);
+    if (ssl_ctx == NULL)
+        return PJ_ENOMEM;
+
+    dssock->ssl_ctx = ssl_ctx;
+
+    /* Set certificate */
+    if (ssock->cert)  {
+        status = set_cert(dssock, ssock->cert);
+        if (status != PJ_SUCCESS)
+            return status;
+    }
+
+    status = get_proto_bounds(ssock, &min_proto, &max_proto);
+    if (status != PJ_SUCCESS)
+        return status;
 
     if (min_proto != kSSLProtocolUnknown) {
         err = SSLSetProtocolVersionMin(ssl_ctx, min_proto);
--- a/pjlib/src/pj/ssl_sock_darwin.c
+++ b/pjlib/src/pj/ssl_sock_darwin.c
@@ -445,15 +445,27 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
         min_proto = kTLSProtocol13;
     }
 
-    /* According to the doc, we can't set min/max proto for TLS protocol
-     * higher than 1.0. The runtime error given is -9830 (Illegal parameter).
+    /* SecureTransport.h documents kSSLProtocol3, kTLSProtocol1,
+     * kTLSProtocol11 and kTLSProtocol12 as legal for these two setters.
+     * TLS 1.3 is not among them, so clamp it to 1.2 rather than leaving the
+     * bound unset -- an unset bound falls back to the stack default, which is
+     * lower than whatever the caller asked for.
      */
-    if (min_proto != kSSLProtocolUnknown && min_proto <= kTLSProtocol1) {
+    if (min_proto == kTLSProtocol13 || max_proto == kTLSProtocol13) {
+        PJ_LOG(3, (THIS_FILE, "TLS 1.3 is not supported by this backend, "
+                              "limiting to TLS 1.2"));
+        if (min_proto == kTLSProtocol13)
+            min_proto = kTLSProtocol12;
+        if (max_proto == kTLSProtocol13)
+            max_proto = kTLSProtocol12;
+    }
+
+    if (min_proto != kSSLProtocolUnknown) {
         err = SSLSetProtocolVersionMin(ssl_ctx, min_proto);
         if (err != noErr) pj_status_from_err(dssock, "SetVersionMin", err);
     }
 
-    if (max_proto != kSSLProtocolUnknown && max_proto <= kTLSProtocol1) {
+    if (max_proto != kSSLProtocolUnknown) {
         err = SSLSetProtocolVersionMax(ssl_ctx, max_proto);
         if (err != noErr) pj_status_from_err(dssock, "SetVersionMax", err);
     }
--- a/pjlib/src/pj/ssl_sock_darwin.c
+++ b/pjlib/src/pj/ssl_sock_darwin.c
@@ -446,28 +446,40 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
     }
 
     /* SecureTransport.h documents kSSLProtocol3, kTLSProtocol1,
-     * kTLSProtocol11 and kTLSProtocol12 as legal for these two setters.
-     * TLS 1.3 is not among them, so clamp it to 1.2 rather than leaving the
-     * bound unset -- an unset bound falls back to the stack default, which is
-     * lower than whatever the caller asked for.
+     * kTLSProtocol11 and kTLSProtocol12 as legal for these two setters;
+     * TLS 1.3 is not among them.
      */
-    if (min_proto == kTLSProtocol13 || max_proto == kTLSProtocol13) {
-        PJ_LOG(3, (THIS_FILE, "TLS 1.3 is not supported by this backend, "
-                              "limiting to TLS 1.2"));
-        if (min_proto == kTLSProtocol13)
-            min_proto = kTLSProtocol12;
-        if (max_proto == kTLSProtocol13)
-            max_proto = kTLSProtocol12;
+    if (min_proto == kTLSProtocol13) {
+        /* The caller accepts nothing below TLS 1.3, which this backend
+         * cannot provide. Lowering the floor would complete a TLS 1.2
+         * handshake while pj_ssl_sock_get_info() still reported TLS 1.3,
+         * so refuse rather than downgrade the caller's policy silently.
+         */
+        PJ_LOG(3, (THIS_FILE, "TLS 1.3 is not supported by this backend"));
+        return PJ_EINVAL;
+    }
+
+    if (max_proto == kTLSProtocol13) {
+        /* A lower version is acceptable to the caller, so lowering the
+         * ceiling stays within the requested policy. This is the default
+         * configuration -- PJ_SSL_SOCK_PROTO_DEFAULT sets the TLS 1.3 bit --
+         * so it is a trace, not a warning.
+         */
+        PJ_LOG(5, (THIS_FILE, "TLS 1.3 is not supported by this backend, "
+                              "limiting the maximum version to TLS 1.2"));
+        max_proto = kTLSProtocol12;
     }
 
     if (min_proto != kSSLProtocolUnknown) {
         err = SSLSetProtocolVersionMin(ssl_ctx, min_proto);
-        if (err != noErr) pj_status_from_err(dssock, "SetVersionMin", err);
+        if (err != noErr)
+            return pj_status_from_err(dssock, "SetVersionMin", err);
     }
 
     if (max_proto != kSSLProtocolUnknown) {
         err = SSLSetProtocolVersionMax(ssl_ctx, max_proto);
-        if (err != noErr) pj_status_from_err(dssock, "SetVersionMax", err);
+        if (err != noErr)
+            return pj_status_from_err(dssock, "SetVersionMax", err);
     }
 
     /* SSL verification options */
--- a/pjlib/src/pj/ssl_sock_darwin.c
+++ b/pjlib/src/pj/ssl_sock_darwin.c
@@ -347,16 +347,19 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
  * unusable configuration can be rejected once at listener setup instead of on
  * every accepted connection.
  *
- * Takes the protocol set by value rather than reading it from the socket, so
- * that a caller can validate a parameter set the socket does not own -- the
- * newsock_param of a listener, for instance.
+ * Takes the protocol set through a pointer rather than reading it from the
+ * socket, so that a caller can validate a parameter set the socket does not
+ * own -- the newsock_param of a listener, for instance -- and hands the
+ * expanded set back through it, so PJ_SSL_SOCK_PROTO_DEFAULT is resolved in
+ * exactly one place.
  */
-static pj_status_t get_proto_bounds(pj_ssl_sock_proto proto,
+static pj_status_t get_proto_bounds(pj_ssl_sock_proto *p_proto,
                                     SSLProtocol *p_min,
                                     SSLProtocol *p_max)
 {
     SSLProtocol min_proto = kSSLProtocolUnknown;
     SSLProtocol max_proto = kSSLProtocolUnknown;
+    pj_ssl_sock_proto proto = *p_proto;
 
     if (proto == PJ_SSL_SOCK_PROTO_DEFAULT) {
         /* SSL 2.0 is deprecated. */
@@ -415,6 +418,7 @@ static pj_status_t get_proto_bounds(pj_ssl_sock_proto proto,
         max_proto = kTLSProtocol12;
     }
 
+    *p_proto = proto;
     *p_min = min_proto;
     *p_max = max_proto;
 
@@ -446,16 +450,18 @@ static pj_status_t get_proto_bounds(pj_ssl_sock_proto proto,
  * rejects -- the check fails open, never closed.
  */
 static pj_status_t ssl_init_server_ctx(pj_ssl_sock_t *ssock,
-                                      const pj_ssl_sock_param *newsock_param)
+                                       const pj_ssl_sock_param *newsock_param)
 {
     darwinssl_sock_t *dssock = (darwinssl_sock_t *)ssock;
     SecIdentityRef identity = NULL;
     SSLProtocol min_proto, max_proto;
+    pj_ssl_sock_proto proto;
     pj_status_t status;
 
     pj_assert(ssock->is_server && !ssock->parent);
 
-    status = get_proto_bounds(newsock_param->proto, &min_proto, &max_proto);
+    proto = newsock_param->proto;
+    status = get_proto_bounds(&proto, &min_proto, &max_proto);
     if (status != PJ_SUCCESS)
         return status;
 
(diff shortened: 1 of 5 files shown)


laconicman:
The `OpenSSL / pjlib,util,pjnath,pjmedia` failure here is not caused by this change, and I can show that rather than assert it.

**The change contributes no code to that build.** That job builds with `PJ_SSL_SOCK_IMP : 1` (its own log line), and `ssl_sock_darwin.c` is an empty translation unit under OpenSSL. Compiled both ways:

```
upstream/master                 ssl_sock_darwin.o under OpenSSL: 1 symbol, 336 bytes
fix/darwin-tls-version-bounds   ssl_sock_darwin.o under OpenSSL: 1 symbol, 336 bytes
```

Byte-identical. The linker log in that same job confirms it never reaches the test binary anyway — `ignoring file ... (ssl_sock_darwin.o): found architecture 'arm64', required architecture 'x86_64'`.

**And the failure is not in TLS code.** It aborts about 35 ms into the run:

```
Shuffling tests, random seed=312
Performing 2 features tests with 2 worker threads
Assertion failed: (((alignment)>0) && ((alignment) & ((alignment)-1))==0),
    function pj_pool_alloc_from_block, file pool_i.h, line 48.
Abort trap: 6
```

That is pjlib's pool allocator, not the SSL socket.

I have filed it separately rather than leave it buried in this thread — it looks like a latent defect rather than environmental noise, and it is the third distinct intermittent abort I have seen from that harness (the other two were `pj_throw_exception_` with no handler, and a PESQ failure with port 4000 already in use). I have not re-run this job: a green re-run would only confirm it is intermittent, which the other occurrences already establish, and it spends your runners for no new information.

Happy to re-run or rebase if you would rather see it green before merging — just say which.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -445,15 +445,27 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
...
-     * higher than 1.0. The runtime error given is -9830 (Illegal parameter).
+    /* SecureTransport.h documents kSSLProtocol3, kTLSProtocol1,
+     * kTLSProtocol11 and kTLSProtocol12 as legal for these two setters.
+     * TLS 1.3 is not among them, so clamp it to 1.2 rather than leaving the
+     * bound unset -- an unset bound falls back to the stack default, which is
+     * lower than whatever the caller asked for.
      */
-    if (min_proto != kSSLProtocolUnknown && min_proto <= kTLSProtocol1) {
+    if (min_proto == kTLSProtocol13 || max_proto == kTLSProtocol13) {
+        PJ_LOG(3, (THIS_FILE, "TLS 1.3 is not supported by this backend, "
```
sauwming: This warning will fire on the default configuration, for every socket.

A few lines up, `PJ_SSL_SOCK_PROTO_DEFAULT` is rewritten to `PJ_SSL_SOCK_PROTO_ALL & ~PJ_SSL_SOCK_PROTO_SSL2`, which still has the `TLS1_3` bit set, so `max_proto` becomes `kTLSProtocol13` and this branch is always taken. Both `pj_ssl_sock_param_default()` and the default SSL method mapping in `sip_transport_tls.c` use `PROTO_DEFAULT`, so every TLS connection on macOS would log this at level 3 even though the application never asked for TLS 1.3.

Suggestion: log the ceiling clamp at level 5, and reserve level 3 for the floor clamp -- that is the only case that actually contradicts a caller's policy.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -445,15 +445,27 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
...
+                              "limiting to TLS 1.2"));
+        if (min_proto == kTLSProtocol13)
```
sauwming: Clamping the *minimum* down is a silent policy downgrade.

With `param.proto == PJ_SSL_SOCK_PROTO_TLS1_3` alone, both `min_proto` and `max_proto` are `kTLSProtocol13`, so both get rewritten to `kTLSProtocol12` and `ssl_create()` returns `PJ_SUCCESS`. An application that deliberately configured a TLS 1.3-only transport then completes a TLS 1.2 handshake with no way to detect it -- `ssock->param.proto` is not updated, so `pj_ssl_sock_get_info()` still reports `TLS1_3`.

Returning `PJ_EINVAL` when the requested floor cannot be met (the alternative mentioned in the PR description) seems preferable, and it is consistent with the existing `"Unsupported TLS/SSL protocol"` path a dozen lines above. Clamping the ceiling only is uncontroversial and can stay as is.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -445,15 +445,27 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
...
+            min_proto = kTLSProtocol12;
+        if (max_proto == kTLSProtocol13)
+            max_proto = kTLSProtocol12;
+    }
+
+    if (min_proto != kSSLProtocolUnknown) {
         err = SSLSetProtocolVersionMin(ssl_ctx, min_proto);
         if (err != noErr) pj_status_from_err(dssock, "SetVersionMin", err);
```
sauwming: The setter failure is still swallowed here (and at `SSLSetProtocolVersionMax()` below): `pj_status_from_err()`'s return value is discarded, so a rejected bound leaves the context on the stack default while `ssl_create()` reports success -- the same silent-downgrade shape this PR is fixing.

`pj_status_from_err()` also assigns `dssock->base.last_err = err`, so a rejected bound leaves a stale native error that a later `pj_ssl_sock_get_info()` reports as `last_native_err` on an otherwise successful connection.

The pattern predates this PR, but the change makes it reachable for more values (e.g. `SSLSetProtocolVersionMax(kSSLProtocol3)` where SSLv3 is compiled out), so returning the status here may be worth doing now that the bounds are actually applied.

sauwming:
Thanks for the fix -- the diagnosis is right: the old `<= kTLSProtocol1` guard silently dropped every TLS 1.1/1.2 bound, and `SecureTransport.h` does document `kSSLProtocol3`, `kTLSProtocol1`, `kTLSProtocol11` and `kTLSProtocol12` as legal for both setters.

I left three inline comments. The first one (the TLS 1.3 warning firing on the default configuration, for every socket) should be addressed before merging; the other two are smaller.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Fixed in 6ae6bde — the ceiling clamp logs at level 5 now, and level 3 is reserved for the
floor case, which is an error return rather than a clamp (see the reply on the next thread).

Worth recording why this is Darwin-specific: it is the only backend whose
`PJ_SSL_SOCK_PROTO_DEFAULT` is not `TLS1_2 | TLS1_3`. `ssl_sock_ossl.c`, `ssl_sock_apple.m`,
`ssl_sock_gtls.c` and `ssl_sock_mbedtls.c` all map the default to those two; this one maps it
to `PJ_SSL_SOCK_PROTO_ALL & ~PJ_SSL_SOCK_PROTO_SSL2`, so SSL 3.0, TLS 1.0 and TLS 1.1 are
enabled by default here and nowhere else.

That looks like a separate defect with a security flavour, so I have deliberately left it out
of this PR. Glad to file it as its own issue if you want it tracked.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Implemented in 6ae6bde: a TLS 1.3 floor returns `PJ_EINVAL`, and the ceiling clamp stays as
it was.

I checked the other backends first, since the argument only holds if this brings Darwin into
line rather than making it the odd one out:

- `ssl_sock_apple.m` and `ssl_sock_mbedtls.c` both `return PJ_EINVAL` for a protocol
  selection they cannot provide.
- `ssl_sock_ossl.c` (option mask) and `ssl_sock_gtls.c` (priority string) can express the
  request exactly, so they never face the choice.

None of them silently narrows a floor, so `PJ_EINVAL` is the consistent behaviour rather than
a new one.

One consequence to flag: two pjlib-test cases request TLS 1.3 alone and expect success — the
TLSv1.3 cipher echo test, and "server TLSv1.2+1.3 vs client TLSv1.3". On this backend both
pass today only because of the silent downgrade, so I have excluded them for
`PJ_SSL_SOCK_IMP_DARWIN`, next to the exclusion already a few lines below them. I could not
run the Darwin suite to confirm their prior state, since the backend is not the configure
default — that part is reasoning from the code, not an observed run.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Fixed in 6ae6bde — both setters now `return pj_status_from_err(...)` rather than discarding
it.

Your point about the stale native error holds as well: because `pj_status_from_err()` assigns
`dssock->base.last_err`, swallowing the failure left an error behind for a later
`pj_ssl_sock_get_info()` to report as `last_native_err` on a connection that had otherwise
succeeded. Returning the status makes socket creation fail instead, which matches the shape
of the other early exits in `ssl_create()`.

pjlib/src/pjlib-test/ssl_sock.c
```diff
@@ -2394,12 +2400,18 @@ int ssl_sock_test(void)
         return ret;
 
 #if (PJ_SSL_SOCK_IMP != PJ_SSL_SOCK_IMP_SCHANNEL)
+/* The Darwin backend cannot set a TLS 1.3 floor -- Secure Transport's
+ * SSLSetProtocolVersionMin() does not accept it -- so a TLS 1.3-only
+ * request is rejected at socket creation rather than downgraded.
+ */
+#if (PJ_SSL_SOCK_IMP != PJ_SSL_SOCK_IMP_DARWIN)
```
sauwming, line 2407: While adding Darwin guards here, could you also clean up the existing one 18 lines below (line 2425-2426)?

```c
/* We can't set min/max proto for TLS protocol higher than 1.0. */
#if (PJ_SSL_SOCK_IMP != PJ_SSL_SOCK_IMP_DARWIN && PJ_SSL_SOCK_IMP != PJ_SSL_SOCK_IMP_SCHANNEL)
```

That comment is the exact claim this PR disproves, and the `PJ_SSL_SOCK_IMP_DARWIN` term it guards is now obsolete: the test asserts `if (ret == 0) return PJ_EBUG;`, i.e. it expects the incompatible pair to fail, and with this change the client's TLS 1.3-only request makes `ssl_create()` return `PJ_EINVAL`, so `echo_test()` fails and the assertion holds on Darwin too. The `PJ_SSL_SOCK_IMP_SCHANNEL` term should stay.

Otherwise the file ends up carrying a comment that contradicts the code this PR just landed.

laconicman: Done in 451a4e86 — the `PJ_SSL_SOCK_IMP_DARWIN` term and the stale comment are gone, and the
SChannel term stays.

Agreed on the reasoning, and the mechanism checks out: `ssl_create()` runs per connection on this
backend, from `ssock_on_connect_complete()` for a client, so the TLS 1.3-only client now fails
there with `PJ_EINVAL`, `echo_test()` returns non-zero, and `if (ret == 0) return PJ_EBUG;` holds.

You are right that leaving it would have been the worse outcome — the comment asserted exactly
what the branch disproves.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -445,17 +445,41 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
...
     }
 
-    /* According to the doc, we can't set min/max proto for TLS protocol
-     * higher than 1.0. The runtime error given is -9830 (Illegal parameter).
+    /* SecureTransport.h documents kSSLProtocol3, kTLSProtocol1,
+     * kTLSProtocol11 and kTLSProtocol12 as legal for these two setters;
+     * TLS 1.3 is not among them.
      */
-    if (min_proto != kSSLProtocolUnknown && min_proto <= kTLSProtocol1) {
+    if (min_proto == kTLSProtocol13) {
```
sauwming, line 412: Could this rejection happen at listener setup for the server case?

`ssl_create()` runs per accepted connection on the server side, so a listener configured `PJSIP_TLSV1_3_METHOD` gets `PJ_SUCCESS` from `pj_ssl_sock_start_accept()` and then fails every inbound connection here, emitting this level-3 log line each time -- a log line any unauthenticated peer can trigger at will, while the operator never sees a startup-time error.

This backend already has `ssl_init_server_ctx()` (line 361, called from `ssl_sock_imp_common.c:2242` in `pj_ssl_sock_start_accept2()`), which exists for exactly this: failing an unusable server configuration up front rather than per connection. Hoisting the min/max computation there -- or at least this `min_proto == kTLSProtocol13` check -- would match that pattern.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -445,17 +445,41 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
...
+        /* The caller accepts nothing below TLS 1.3, which this backend
+         * cannot provide. Lowering the floor would complete a TLS 1.2
+         * handshake while pj_ssl_sock_get_info() still reported TLS 1.3,
+         * so refuse rather than downgrade the caller's policy silently.
+         */
+        PJ_LOG(3, (THIS_FILE, "TLS 1.3 is not supported by this backend"));
+        return PJ_EINVAL;
+    }
+
+    if (max_proto == kTLSProtocol13) {
```
sauwming, line 422: One correction to the PR description, which says the floor was previously left unset.

On the default path it wasn't. A few lines above, `PJ_SSL_SOCK_PROTO_DEFAULT` expands to `PJ_SSL_SOCK_PROTO_ALL & ~PJ_SSL_SOCK_PROTO_SSL2`, which still has the `SSL3` bit, so `min_proto` is `kSSLProtocol3`. Checked against the macOS 15 SDK: `kSSLProtocolUnknown=0, kSSLProtocol3=2, kTLSProtocol1=4, kTLSProtocol11=7, kTLSProtocol12=8, kTLSProtocol13=10`, so `kSSLProtocol3 <= kTLSProtocol1` held and the old guard *did* apply it.

So the default was never "no minimum"; it was actively setting the floor to SSL 3.0, below Secure Transport's own default of `kTLSProtocol1`. That is unchanged by this PR. The real win is the TLS 1.1/1.2 case (e.g. `PJSIP_TLSV1_2_METHOD`, where the floor was genuinely dropped) -- worth rewording the description so the motivation matches what the code does.

Separately, and more to the point of the security argument: this backend is the only one whose `PROTO_DEFAULT` is `PROTO_ALL & ~SSL2`. `ssl_sock_ossl.c:1445`, `ssl_sock_gtls.c:475`, `ssl_sock_apple.m:876` and `ssl_sock_mbedtls.c:369` all use `PJ_SSL_SOCK_PROTO_TLS1_2 | PJ_SSL_SOCK_PROTO_TLS1_3`. Aligning Darwin with them is the change that would actually deliver the stated benefit -- happy for that to be the separate issue you offered to file rather than scope creep here.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Done in `b09ece12`. The mapping from `ssock->param.proto` onto the Secure Transport bounds now lives in `get_proto_bounds()`, called from both `ssl_create()` and `ssl_init_server_ctx()`, so an unusable range is rejected once when the listener starts and `ssl_create()` just applies what it returns.

You are right that this is what `ssl_init_server_ctx()` is for — `pj_assert(ssock->is_server && !ssock->parent)` says exactly that it runs for the listener and not for accepted children. I put the check before the existing `if (!ssock->cert) return PJ_SUCCESS;` early return, since a bad protocol range should be rejected whether or not credentials are configured.

The log-flooding angle is the part I had missed entirely. A level 3 line that any unauthenticated peer can trigger at will, on a listener that reported success at startup, is a worse failure than the one I was fixing.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: You are right and the description was wrong. I checked the SDK rather than take it on trust: `kSSLProtocolUnknown=0, kSSLProtocol3=2, kTLSProtocol1=4, kTLSProtocol11=7, kTLSProtocol12=8, kTLSProtocol13=10`, so `kSSLProtocol3 <= kTLSProtocol1` holds and the old guard did apply the floor on the default path. "No minimum" was never the state; an SSL 3.0 floor was, which is below Secure Transport's own default of `kTLSProtocol1` — so the default was actively lowering the floor rather than failing to raise it.

Description reworded to say that, and to put the real win where it belongs: TLS 1.1 and TLS 1.2, where the floor genuinely was dropped.

On `PROTO_DEFAULT`: filed as #5239, with the SDK values and the observation that this is the one backend not using `TLS1_2 | TLS1_3`. I left it as an issue rather than a patch because aligning it is a behaviour change for anyone reaching a pre-TLS 1.2 peer on this backend, and that reads as your call — the three-line patch is in the issue if you want it.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -358,14 +358,104 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
...
 
     pj_assert(ssock->is_server && !ssock->parent);
 
+    /* Reject a protocol range this backend cannot serve here, once, rather
+     * than once per accepted connection in ssl_create(). A listener asking
+     * for TLS 1.3 alone would otherwise start successfully and then fail
+     * every inbound handshake, emitting a log line any unauthenticated peer
+     * can trigger while the operator never sees a startup error.
+     */
+    status = get_proto_bounds(ssock, &min_proto, &max_proto);
```
sauwming: This validates the wrong `param`.

`get_proto_bounds(ssock, ...)` reads the *listener's* `ssock->param.proto`, but accepted child sockets are created from `ssock_parent->newsock_param` (`ssl_sock_imp_common.c:1189`, copied in at `:2323`). `pj_ssl_sock_start_accept()` passes `&ssock->param` (`:2202`) so the two coincide there, but `pj_ssl_sock_start_accept2()` takes an independent `newsock_param`, and then both directions misbehave:

- False negative: `newsock_param->proto = PJ_SSL_SOCK_PROTO_TLS1_3` with a default listener `param.proto` -- `start_accept2()` returns `PJ_SUCCESS` and every accepted connection still fails in `ssl_create()`, which is the case this commit is meant to close.
- False positive: listener `param.proto = PJ_SSL_SOCK_PROTO_TLS1_3` but `newsock_param->proto = PJ_SSL_SOCK_PROTO_TLS1_2` -- `start_accept2()` now returns `PJ_EINVAL` and refuses a configuration in which every accepted connection would have worked.

pjsip's TLS transport `memcpy`s `ssock_param` into `newsock_param` (`sip_transport_tls.c:542`), so it is unaffected, but a direct pjlib caller of `start_accept2()` is. The check wants `newsock_param->proto` -- either doing it in `pj_ssl_sock_start_accept2()` before the copy, or passing the proto value into `ssl_init_server_ctx()`.

Also, `min_proto`/`max_proto` are written and never read here. Worth a brief note that the call is pure validation, otherwise it reads like an oversight.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -358,14 +358,104 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
  * still start for a configuration that every connection subsequently
```
sauwming: Inserting `get_proto_bounds()` here has orphaned this comment.

Lines 344-360 document `ssl_init_server_ctx()` ("Eagerly load and validate the certificate on the listener socket..."), but the new function and its own comment now sit between the two, so the file reads as two stacked doc comments where the first describes a function three definitions away, and `ssl_init_server_ctx()` is left undocumented.

Suggest moving the certificate comment back down to immediately above `ssl_init_server_ctx()`, and extending it to cover the protocol-range check that function now also performs.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Confirmed and fixed in `40c461ff`. This was a real bug and the false negative is the worse half: the case the commit exists to close was the one still slipping through.

I checked the ordering before choosing a fix. `ssl_init_server_ctx()` is called at `ssl_sock_imp_common.c:2242` and `pj_ssl_sock_param_copy(pool, &ssock->newsock_param, newsock_param)` runs at `:2323`, so `ssock->newsock_param` is not populated yet at that point — reading it inside the hook would have validated an empty struct. That ruled out the tempting third option and left your two.

I took the second, passing the parameters in, because the constraint is backend-specific: Secure Transport's inability to set a TLS 1.3 bound is not something `pj_ssl_sock_start_accept2()` should know about. The signature change is contained — the declaration is guarded to `PJ_SSL_SOCK_IMP_OPENSSL || PJ_SSL_SOCK_IMP_DARWIN`, so there are two implementations and one call site, and the OpenSSL one takes a `PJ_UNUSED_ARG`.

`get_proto_bounds()` now takes the protocol set by value rather than reading it off the socket, which is what lets it serve a parameter set the socket does not own. `ssl_create()` keeps writing the expanded `PJ_SSL_SOCK_PROTO_DEFAULT` back to `ssock->param.proto`, since `pj_ssl_sock_get_info()` reports that field at `:1778` and dropping the write would have changed what callers see.

On `min_proto`/`max_proto` being written and never read: fair, it did read like an oversight. The moved comment now says the call is validation only and that the bounds are discarded because `ssl_create()` derives its own per connection.

Verified: `ssl_sock_darwin.c`, `ssl_sock_ossl.c` and `ssl_sock_apple.m` all compile with zero errors and no new warnings — the Darwin count is unchanged from baseline at 51, all Secure Transport deprecations.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Moved and extended in `40c461ff`. The block sits above `ssl_init_server_ctx()` again, and it now covers the protocol check as you asked — including that it validates `newsock_param` rather than `ssock->param`, and that the bounds it produces are discarded because the call is validation only.

Same cause as the one you caught on #5233: I anchored the insertion on the `static pj_status_t ...` signature line, which sits *below* the function's doc comment, so the new function landed between the comment and what it describes. Worse here than there, because that block is upstream's own documentation rather than mine.

I swept both branches for the rest of the class rather than wait to be told a third time — every file we touch, counting stacked comment blocks and comparing against `upstream/master`. All counts match master exactly, so the remaining stacked pairs are upstream's own section headers and these two were the only ones we introduced. Both are now fixed.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -411,51 +504,29 @@ static pj_status_t ssl_create(pj_ssl_sock_t *ssock)
             return status;
     }
 
-    /* Set min and max protocol version */
+    /* Set min and max protocol version. The expansion is written back
+     * because pj_ssl_sock_get_info() reports ssock->param.proto.
+     */
     if (ssock->param.proto == PJ_SSL_SOCK_PROTO_DEFAULT) {
```
sauwming: The `PJ_SSL_SOCK_PROTO_DEFAULT` expansion now exists in two places: here, and again inside `get_proto_bounds()` at line 361. Thanks for documenting why the write-back is needed -- the concern is the duplicated policy constant rather than the write-back itself.

Concrete failure mode: #5239, which you filed to change this very default to `TLS1_2 | TLS1_3`, will be applied to one copy and not the other, and then `pj_ssl_sock_get_info()` reports one protocol set while the Secure Transport bounds are derived from another.

Suggest having `get_proto_bounds()` hand back the expanded set (an out param, or a `pj_ssl_sock_proto *` in/out) and writing that to `ssock->param.proto` here, so the default lives in exactly one place.

pjlib/src/pj/ssl_sock_darwin.c
```diff
@@ -358,14 +445,20 @@ static pj_status_t set_cert(darwinssl_sock_t *dssock, pj_ssl_cert_t *cert)
  * still start for a configuration that every connection subsequently
  * rejects -- the check fails open, never closed.
  */
-static pj_status_t ssl_init_server_ctx(pj_ssl_sock_t *ssock)
+static pj_status_t ssl_init_server_ctx(pj_ssl_sock_t *ssock,
+                                      const pj_ssl_sock_param *newsock_param)
```
sauwming: Continuation line is indented 38 spaces, so `const pj_ssl_sock_param *newsock_param` sits one column left of `pj_ssl_sock_t *ssock`. The identical signature in `ssl_sock_ossl.c:2105` uses 39 and lines up.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Fixed in `85d93ef9`. You are right, and the failure mode you name is the one that would actually have bitten: #5239 is mine, so I would have changed the default in one copy and left `pj_ssl_sock_get_info()` reporting a set the bounds were not derived from.

`get_proto_bounds()` now takes the set through a pointer and hands the expanded value back, so `ssl_create()` writes back exactly what the mapping used rather than computing the same thing separately. `ssl_init_server_ctx()` passes a copy — `newsock_param` is const there and the expansion is discarded, since that call is validation only. `PJ_SSL_SOCK_PROTO_DEFAULT` now appears once in code, in `get_proto_bounds()`.

I took the in/out pointer rather than a separate out param because it makes the write-back at the call site impossible to forget: there is no version of `ssl_create()` that gets the bounds without also getting the set they came from.

pjlib/src/pj/ssl_sock_darwin.c
laconicman: Fixed in `85d93ef9` — 39 columns now, matching `ssl_sock_ossl.c:2105` exactly.

sauwming (approved):

nanangizz (approved):

PR #5238, opened 2026-09-04
Fix Coverity CHECKED_RETURN defects and build warnings

sauwming:
## Coverity CHECKED_RETURN

- **CID 1685791** `pjsip/src/pjsip-ua/sip_100rel.c`: `pjsip_100rel_attach()` ignored the `pjsip_dlg_add_usage()` result; it is now returned on failure, matching the two call sites in `sip_inv.c`.
- **CID 1685939** `pjmedia/src/pjmedia/conference.c`: `pjmedia_conf_get_ports_info()` ignored `pjmedia_conf_get_port_info()` and advanced `count` unconditionally, so a failed slot would be reported as a filled entry. The identical loop in `conf_switch.c` and `conf_thread.c` is fixed as well.
- **CID 1685792** `pjsip/src/test/dlg_target_refresh_test.c`: the three `pjsip_endpt_register_module()` calls are now wrapped in `PJ_TEST_SUCCESS()`, as done elsewhere in the file.

## Build warnings

- `pjlib/src/pjlib-test/ssl_sock_stress.c`: `total_pending`/`total_sent` were accumulated but never used; they are now reported in the existing summary log (`-Wunused-but-set-variable`).
- `pjmedia/src/pjmedia-videodev/darwin_dev.m`: `effective_portrait_angle` is only used in the `TARGET_OS_IPHONE` branch, so its declaration is guarded the same way (`-Wunused-variable` on macOS).
- `pjsip/src/test/pjsua_call_test.c`: `siprec_test_mark_siprec_sessions()` is only called from within `#if PJSUA_HAS_SIPREC`, so the definition is guarded too (`-Wunused-function` when SIPREC is disabled).

## Testing

- `pjlib/build`, `pjmedia/build` and `pjsip/build` rebuild clean (no warnings) on macOS.
- `conf_switch.c` / `conf_thread.c` are not in the default build; syntax-checked with `gcc -fsyntax-only -Wall`.
- `pjsua_call_test.c` also syntax-checked with `-DPJSIP_HAS_SIPREC=0`, the configuration that produced the warning.
- `pjsip-test dlg_target_refresh_test` passes.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_100rel.c
+++ b/pjsip/src/pjsip-ua/sip_100rel.c
@@ -194,6 +194,7 @@ PJ_DEF(pj_status_t) pjsip_100rel_init_module(pjsip_endpoint *endpt)
 PJ_DEF(pj_status_t) pjsip_100rel_attach(pjsip_inv_session *inv)
 {
     dlg_data *dd;
+    pj_status_t status;
 
     /* Check that 100rel module has been initialized */
     PJ_ASSERT_RETURN(mod_100rel.mod.id >= 0, PJ_EINVALIDOP);
@@ -201,7 +202,9 @@ PJ_DEF(pj_status_t) pjsip_100rel_attach(pjsip_inv_session *inv)
     /* Create and attach as dialog usage */
     dd = PJ_POOL_ZALLOC_T(inv->dlg->pool, dlg_data);
     dd->inv = inv;
-    pjsip_dlg_add_usage(inv->dlg, &mod_100rel.mod, (void*)dd);
+    status = pjsip_dlg_add_usage(inv->dlg, &mod_100rel.mod, (void*)dd);
+    if (status != PJ_SUCCESS)
+        return status;
 
     PJ_LOG(5,(dd->inv->dlg->obj_name, "100rel module attached"));
 
--- a/pjlib/src/pjlib-test/ssl_sock_stress.c
+++ b/pjlib/src/pjlib-test/ssl_sock_stress.c
@@ -2104,8 +2104,9 @@ static int concurrent_send_test(void)
         }
 
         PJ_LOG(3, ("", "...concurrent_send_test: threads=%d "
-                   "total_sends=%d recv=%lu/%lu",
-                   CS_SENDER_THREADS, total_sends,
+                   "total_sends=%d pending=%d sent=%lu recv=%lu/%lu",
+                   CS_SENDER_THREADS, total_sends, total_pending,
+                   (unsigned long)total_sent,
                    (unsigned long)state_cli.recv,
                    (unsigned long)total_expected));
     }
--- a/pjmedia/src/pjmedia-videodev/darwin_dev.m
+++ b/pjmedia/src/pjmedia-videodev/darwin_dev.m
@@ -1340,7 +1340,9 @@ static pj_status_t darwin_stream_set_cap(pjmedia_vid_dev_stream *s,
                 const CGFloat cap_ori[4] = { 0, 90, 180, 270};
                 int idx;
                 int rotation;
+#if TARGET_OS_IPHONE
                 int effective_portrait_angle;
+#endif
 
                 if (strm->param.orient < PJMEDIA_ORIENT_NATURAL ||
                     strm->param.orient > PJMEDIA_ORIENT_ROTATE_270DEG)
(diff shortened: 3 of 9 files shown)


pjsip/src/pjsip-ua/sip_100rel.c
```diff
@@ -194,14 +194,17 @@ PJ_DEF(pj_status_t) pjsip_100rel_init_module(pjsip_endpoint *endpt)
...
     /* Check that 100rel module has been initialized */
     PJ_ASSERT_RETURN(mod_100rel.mod.id >= 0, PJ_EINVALIDOP);
 
     /* Create and attach as dialog usage */
     dd = PJ_POOL_ZALLOC_T(inv->dlg->pool, dlg_data);
     dd->inv = inv;
-    pjsip_dlg_add_usage(inv->dlg, &mod_100rel.mod, (void*)dd);
+    status = pjsip_dlg_add_usage(inv->dlg, &mod_100rel.mod, (void*)dd);
+    if (status != PJ_SUCCESS)
+        return status;
```
nanangizz, line 207: The commit message says this new return value "matches the two call sites in sip_inv.c", but both call sites still discard it:

- `pjsip/src/pjsip-ua/sip_inv.c:1151` (UAC) — bare `pjsip_100rel_attach(inv);`
- `pjsip/src/pjsip-ua/sip_inv.c:1955` (UAS) — same

So the CHECKED_RETURN pattern just moves up one frame, and Coverity is likely to re-flag those two lines. It's also inconsistent with the `pjsip_dlg_add_usage()` call ~10 lines above each of them, which does check.

Practical effect is bounded but it does turn a fail-fast into a fail-late: with `PJSIP_INV_REQUIRE_100REL` set and `mod_100rel.mod.id < 0` (100rel module never initialized), `pjsip_100rel_attach()` returns `PJ_EINVALIDOP`, yet `pjsip_inv_create_uas()` still returns `PJ_SUCCESS`. The session then fails per-message in `pjsip_100rel_tx_response()` (`PJ_ASSERT_RETURN(dd != NULL, PJ_EINVALIDOP)`, sip_100rel.c:695) instead of at creation time. No crash — every `dd` consumer in sip_100rel.c is NULL-safe.

Suggest completing the fix at both sites, following the surrounding idiom:

```c
    /* Create 100rel handler */
    status = pjsip_100rel_attach(inv);
    if (status != PJ_SUCCESS) {
        pjsip_dlg_dec_lock(dlg);
        return status;
    }
```

The rest of the PR looks good to me — the conference `get_port_info()` guards are a safe no-op (the loop bound plus the `!conf->ports[i]` skip under the same mutex already exclude both failure conditions), the `ssl_sock_stress.c` summary reads happen after all sender threads are joined, and the `PJSUA_HAS_SIPREC` / `TARGET_OS_IPHONE` guards match every call site.

sauwming: Done in 748bd43 — both call sites now propagate the status.

One deviation from the suggested snippet: the attach is moved *ahead* of `pjsip_dlg_inc_session()` rather than checked in place. Returning after `inc_session()` would leave the dialog with a session reference that nothing releases — the caller only ever sees `*p_inv == NULL`, so e.g. `pjsua_call.c`'s `on_error` just does `pjsip_dlg_dec_lock(dlg)` and the dialog would never be destroyed. Attaching before `inc_session()` puts the new error path on exactly the same footing as the `pjsip_dlg_add_usage()` failure right above it. 100rel attach only calls `pjsip_dlg_add_usage()`, so it has no ordering dependency on `inc_session()` or (UAS) on `inv->invite_tsx`.

Worth noting the behavior change on the UAC side, since that attach is unconditional: an application that never calls `pjsip_100rel_init_module()` will now get `PJ_EINVALIDOP` out of `pjsip_inv_create_uac()` instead of silently continuing. Such an app already trips the `pj_assert()` in `PJ_ASSERT_RETURN(mod_100rel.mod.id >= 0, ...)` in debug builds, and every in-tree sample plus pjsua-lib inits the module, so this looks like the intended fail-fast — but say the word if you'd rather keep UAC lenient and only propagate when `PJSIP_INV_REQUIRE_100REL` is set.

Tested: pjsip-test `inv_offer_answer_test`, `dlg_core_test`, `dlg_target_refresh_test`, `pjsua_call_test` all pass.

nanangizz (approved):

Issue #5100, opened 2026-07-20
pj_hash_calc_tolower can return zero for specific UUIDs


lubosures:
### Describe the bug

I ran into an (seemingly) impossible) assert when calling `pjsip_dlg_create_uac`, stack trace attached below. The assert fires on `tag_hval` being zero.

Looking at the params passed to the call, I could see legal values
```
(gdb) set pr pr
(gdb) p create_param
$2 = {
  ua = 0x55f6f844d020 <mod_ua>,
  local_uri = {
    ptr = 0x7f860c24cba0 "<sip:17372036511@XX.234.119.152:2004>",
    slen = 37
  },
  local_contact = {
    ptr = 0x0,
    slen = 0
  },
  remote_uri = {
    ptr = 0x7f860c24cbe0 "<sip:6680224654500909@YYY.160.251.13:5060;transport=udp;user=phone>",
    slen = 67
  },
  target = {
    ptr = 0x0,
    slen = 0
  },
  grp_lock = 0x0
}
```

Nothing suspicious here. The problem is caused by a specific call-id `d48600d9-514b-4655-a606-43a65dc1f54b`

### Steps to reproduce

If there's a unit test for `pj_hash_calc_tolower` pass `d48600d9-514b-4655-a606-43a65dc1f54b` to it.
I'm attaching a C++ version mimicking the original code  I used to verify the problem is real  for both signed/unsigned 32 bit ints.

[sim_pj_hash.cpp](https://github.com/user-attachments/files/30205305/sim_pj_hash.cpp)

### PJSIP version

2.14.1 

### Context

Ubuntu 24.04, gcc 13 optimized build 

### Log, call stack, etc

```shell
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140213706085952) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140213706085952) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140213706085952, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007f8610ee4476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007f8610eca7f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007f8610eca71b in __assert_fail_base (fmt=0x7f861107f130 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
[...]

PR #5240, opened 2026-09-04
Fix use-after-free in inv_session_destroy()

cyrilleberaud:
## Problem

`inv_session_destroy()` calls `pjsip_dlg_dec_session()` **before** releasing the session's flip-flop pools and reference counter:

```c
    pjsip_dlg_dec_session(inv->dlg, &mod_inv.mod);

    /* Release the flip-flop pools */
    pj_pool_release(inv->pool_prov);      /* reads inv->pool_prov */
    ...
    pj_atomic_destroy(inv->ref_cnt);      /* reads inv->ref_cnt */
```

The `pjsip_inv_session` is allocated from the dialog pool (`PJ_POOL_ZALLOC_T(dlg->pool, ...)`). When the dialog has no other session and no pending transaction, `pjsip_dlg_dec_session()` → `pjsip_dlg_dec_lock()` destroys the dialog and releases its pool right there. The three reads that follow are then made from freed memory.

## Why it usually goes unnoticed

PJSIP itself only drops the last reference from inside `pjsip_dlg_on_tsx_state()`, which brackets the module callbacks with `pjsip_dlg_inc_lock()` / `pjsip_dlg_dec_lock()`. The lock holds a session count of its own, so the dialog survives until the final `dec_lock()`, after `inv_session_destroy()` has returned.

An application that holds its own reference (`pjsip_inv_add_ref()`) and drops it from another thread, outside of any dialog lock, after the BYE transaction has already terminated, does hit it. With a caching pool that actually frees released pools (`max_capacity` exhausted or 0), the memory is reused quickly and `pj_pool_release()` faults on a garbage pool pointer:

```
#00 pj_pool_release+4
#01 pjsip_inv_dec_ref+288
#02 <application code dropping its reference>
```

We are seeing this in the field in [Jami](https://jami.net) on Android.

## Fix

Move `pjsip_dlg_dec_session()` to the very end of `inv_session_destroy()`, so that nothing dereferences `inv` after the dialog may have been destroyed. `inv->dlg` is read as the call argument, before the decrement happens.

No behaviour change for callers that already hold the dialog lock.

### DIFF
--- a/pjsip/src/pjsip-ua/sip_inv.c
+++ b/pjsip/src/pjsip-ua/sip_inv.c
@@ -334,7 +334,6 @@ static void inv_session_destroy(pjsip_inv_session *inv)
     }
     pjsip_100rel_end_session(inv);
     pjsip_timer_end_session(inv);
-    pjsip_dlg_dec_session(inv->dlg, &mod_inv.mod);
 
     /* Release the flip-flop pools */
     pj_pool_release(inv->pool_prov);
@@ -344,6 +343,13 @@ static void inv_session_destroy(pjsip_inv_session *inv)
 
     pj_atomic_destroy(inv->ref_cnt);
     inv->ref_cnt = NULL;
+
+    /* This must be the last thing we do with inv. The session is allocated
+     * from the dialog pool, and decrementing the dialog session count may
+     * destroy the dialog if it has no other session and no transaction,
+     * which frees inv along with it.
+     */
+    pjsip_dlg_dec_session(inv->dlg, &mod_inv.mod);
 }
 
 static pj_status_t inv_async_auth_send_impl(
--- a/pjsip/src/test/inv_offer_answer_test.c
+++ b/pjsip/src/test/inv_offer_answer_test.c
@@ -737,6 +737,77 @@ static void on_new_session(pjsip_inv_session *inv, pjsip_event *e)
 }
 
 
+
+/*
+ * Regression test for the use-after-free in inv_session_destroy() (#5241).
+ *
+ * pjsip_inv_session is zalloc'd from dlg->pool, and so is inv->ref_cnt
+ * (pjsip_inv_create_uac/uas). The session holds exactly one dialog session
+ * count for its whole life, released by inv_session_destroy() through
+ * pjsip_dlg_dec_session(). When that call takes dlg->sess_count to zero with
+ * no transaction outstanding, pjsip_dlg_dec_lock() destroys the dialog and
+ * releases dlg->pool synchronously, via dlg_on_destroy() on the group lock's
+ * destroy list. Everything inv_session_destroy() reads from `inv` after that
+ * point is read from freed memory.
+ *
+ * sip_inv.h documents no locking requirement on pjsip_inv_dec_ref(), so an
+ * application that takes its own reference and drops the last one outside the
+ * dialog lock, once the session is already DISCONNECTED, lands exactly there.
+ * No second thread and no network are involved: this is an ordering defect,
+ * not a race, so the test is deterministic.
+ *
+ * Detection is by AddressSanitizer, which pjsip-test already runs under in
+ * CI. For ASan to see anything the dialog pool must be freed rather than
+ * recycled, so the test first grows dlg->pool past the caching pool's largest
+ * cached size -- see the comment on the pj_pool_alloc() below.
+ */
+static int inv_dec_ref_unlocked_test(void)
+{
+    pjsip_dialog *dlg;
+    pjsip_inv_session *inv;
+    pj_str_t uri = contact_uri;
+
+    PJ_LOG(3,(THIS_FILE, "  dropping the last inv reference unlocked"));
+
+    PJ_TEST_SUCCESS(pjsip_dlg_create_uac(pjsip_ua_instance(),
+                                         &uri, &uri, &uri, &uri, &dlg),
+                    NULL, return -400);
+
+    /* pj_caching_pool recycles a released pool unless its capacity exceeds
+     * the largest cached size, in which case cpool_release_pool() destroys it
+     * outright. A recycled pool is only reset, so the stale reads still
+     * return their old values and the defect stays invisible -- which is why
+     * it has gone unnoticed. Grow the dialog pool past that bound so that the
+     * release is a real free and the sanitizer can see the access.
+     */
+    PJ_TEST_NOT_NULL(pj_pool_alloc(dlg->pool, 128 * 1024), NULL, return -410);
+
+    PJ_TEST_SUCCESS(pjsip_inv_create_uac(dlg, NULL, 0, &inv), NULL,
+                    return -420);
+
+    /* The invite session must be the dialog's only session holder, otherwise
+     * the dec_session() below cannot reach zero and the test proves nothing. */
+    PJ_TEST_EQ(dlg->sess_count, 1, "unexpected dialog session count",
+               return -430);
+
+    /* The application takes its own reference, as sip_inv.h invites it to. */
+    PJ_TEST_SUCCESS(pjsip_inv_add_ref(inv), NULL, return -440);
+
+    /* End the session. This drops the stack's own reference from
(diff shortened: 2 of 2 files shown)


laconicman:
I went through this independently and can confirm the analysis. Reproduced the use-after-free on master under ASan and verified that this patch resolves it, with `dlg_core_test` and `inv_offer_answer_test` still green under ASan afterwards. Details and the full trace are in #5241.

Three things that I think strengthen the case, all of which make the defect look *more* tractable than the description does:

**It is not a race.** No second thread, no network and no BYE are needed — only that the last reference is dropped by a caller not already holding the dialog lock. This is enough, on one thread:

```c
pjsip_dlg_create_uac(pjsip_ua_instance(), &uri, &uri, &uri, &uri, &dlg);
pjsip_inv_create_uac(dlg, NULL, 0, &inv);   /* dlg->sess_count == 1 */
pjsip_inv_add_ref(inv);
pjsip_inv_terminate(inv, PJSIP_SC_REQUEST_TERMINATED, PJ_FALSE);
pjsip_inv_dec_ref(inv);                     /* boom */
```

That matters for reviewability: it makes the bug deterministically testable rather than a field-only crash.

**`inv->ref_cnt` is in the dialog pool too**, not just `inv` itself — `pj_atomic_create(dlg->pool, 0, &inv->ref_cnt)` at `sip_inv.c:1105` and `:1884`. So `pj_atomic_destroy(inv->ref_cnt)` is a second, independent read of freed memory, not only a read of a freed pointer field.

**There is precedent in this same file.** `pjsip_inv_process_redirect()` already brackets its `pjsip_dlg_dec_session()` in `pjsip_dlg_inc_lock()` (`sip_inv.c:3634`), with a comment saying it is there "to prevent the dialog or session from being destroyed while we're waiting for user confirmation". The hazard is already recognised in `sip_inv.c`; `inv_session_destroy()` is the one place that does not guard against it.

On alternatives: I also tested bracketing the whole of `inv_session_destroy()` in `inc_lock`/`dec_lock` instead of reordering. It works, and there is no recursion problem on the group lock. But I would not prefer it to this patch — it adds a lock acquire/release to every teardown and buys nothing the reorder does not. The reorder is the right shape.

One small thing that might be worth a line in the description: the reason this stays latent is not only the locking pattern but the allocator. `cpool_release_pool()` destroys a pool only when its capacity exceeds the largest cached size (64 KiB) or the caching pool's `max_capacity` would be exceeded; otherwise it is just `pj_pool_reset()` and recycled, the first block is retained unscrubbed, and the stale reads return their correct former values. So the failure mode is latent rather than intermittent.

@cyrilleberaud I have a deterministic regression test for `pjsip-test` sitting ready — about 60 lines added to `inv_offer_answer_test.c`, no new file and no build-system change, failing under ASan on master and passing with this patch. Glad to hand it over for you to fold into this PR, or to send it as a small follow-up once this merges — whichever you and the maintainers prefer. It is already written and verified, so saying yes does not commit anyone to open-ended work.

nanangizz:
Thanks, and thanks to @laconicman for the independent confirmation in #5241.

I went through the chain and it holds up: `inv` and `inv->ref_cnt` both come from `dlg->pool` (`sip_inv.c:1102`/`:1105`, `:1885`/`:1888`), and the destruction is synchronous all the way down — `pjsip_dlg_dec_session()` → `pjsip_dlg_dec_lock()` → `unregister_and_destroy_dialog()` → `destroy_dialog()` → `pj_grp_lock_dec_ref()` → `grp_lock_destroy()` calls the handlers inline (`lock.c:397`) → `dlg_on_destroy()` releases the dialog pool. So the three statements after the old `dec_session()` do read freed memory once the session holds the last `sess_count`.

The reorder is the right fix. `inv->dlg` is evaluated before the decrement, so the final line is safe, and it avoids adding a lock round-trip to every teardown. It also matches what `pjsip-simple/evsub.c` already does: `evsub_on_destroy()` has the `dec_session()` as its only statement, with nothing after it.

@laconicman please do send the regression test — folding it into this PR would be better than a follow-up, since it only makes sense together with the fix.

laconicman:
@nanangizz thanks — here it is.

@cyrilleberaud I have no way to push to your branch: "Allow edits by maintainers" only grants that to people with write access on `pjsip/pjproject`, not to another contributor. So the test is a commit on my fork for you to cherry-pick:

```sh
git fetch https://github.com/laconicman/pjproject.git inv-uaf-regression-test
git cherry-pick FETCH_HEAD
```

It is based directly on your `b5c694092`, so it applies cleanly with no conflict — I verified that from a fresh clone of your branch. If you would rather click a button than run a command, say so and I will open it as a PR against `cyrilleberaud:fix-inv-session-destroy-uaf` instead; merging that would land it on the same branch and update this PR.

**What it adds.** 75 lines to `pjsip/src/test/inv_offer_answer_test.c` — one static function plus a call at the end of `inv_offer_answer_test()`. No new file, no `test.c`/`test.h` change, no build-system change.

**What it does.** Creates a UAC dialog and invite session, takes an application reference with `pjsip_inv_add_ref()`, ends the session with `pjsip_inv_terminate()` so the stack drops its own reference under the lock that function holds, then drops the last reference with no dialog lock held. No second thread, no transport, no BYE — the defect is an ordering bug, not a race, so the test is fully deterministic.

**Two things in it that look odd but are load-bearing**, both commented in place:

- It grows the dialog pool past 64 KiB before terminating. `cpool_release_pool()` only calls `pj_pool_destroy_int()` when a pool's capacity exceeds the largest cached size or `max_capacity` would be exceeded; otherwise it is just `pj_pool_reset()` and recycled, the first block is retained unscrubbed, and the stale reads return their correct former values. Without that push past the bound the sanitizer sees nothing and the test passes on a broken tree.
- It lives in `inv_offer_answer_test.c` rather than `dlg_core_test.c`. `pjsip_inv_usage_init()` registers a process-wide singleton, and `inv_offer_answer_test` skips its own init when the module id is already set. Initialising `mod_inv` from another test file with placeholder callbacks makes the offer/answer test hang waiting for events that never arrive, rather than fail — I hit exactly that on the first attempt.

**Verified on your branch**, ASan build (`CFLAGS="-g -O0 -fsanitize=address"`), macOS/arm64:

- On `b5c694092` as it stands: `pjsip-test -w 0 -l 0 dlg_core_test inv_offer_answer_test` → 2/2 pass, 7.8 s, no sanitizer output.
- On `b5c694092` with your fix commit reverted and the test kept: `heap-use-after-free` at `sip_inv.c:340` in `inv_session_destroy()`, freed via `pjsip_dlg_dec_session` → `pjsip_dlg_dec_lock` → `destroy_dialog` → `grp_lock_destroy` → `dlg_on_destroy`, i.e. exactly the chain in the commit message.

So it fails without the fix and passes with it, on your tree rather than on a reconstruction of it.

One small thing I changed from what I described in #5241: the comment heading now cites the issue (#5241) rather than this PR, to match how `dlg_core_test.c` cites issue #5100.

nanangizz:
I took the liberty of including the proposed test code in this PR directly — thanks @laconicman for writing and verifying it, and @cyrilleberaud for the fix.

Verified locally with an ASan build (`-g -O0 -fsanitize=address`), running `dlg_core_test inv_offer_answer_test`:

- with the fix: passes, no ASan findings
- with only `sip_inv.c` reverted to the pre-fix ordering: `heap-use-after-free` at `sip_inv.c:340` in `inv_session_destroy()`, read of size 8, freed via `pjsip_dlg_dec_session()` → `destroy_dialog()` → `dlg_on_destroy()` → `pjsip_endpt_release_pool()`

So the test is load-bearing rather than merely passing. The freed region is reported as 132000 bytes, which confirms the deliberate pool growth in the test is what makes `cpool_release_pool()` actually destroy the pool instead of resetting and recycling it — without that the defect stays invisible to the sanitizer.

For the record: `pjsip-test` does run under ASan in CI (`default-full-bundle-3` on Linux, `video-openh264-3` on macOS), so the test lands in a job that will actually exercise it.

sauwming (approved):

PR #5244, opened 2026-09-08
pjlib: honour type alignment in the pool and fifobuf allocators

nanangizz:
`PJ_POOL_ALIGNMENT` defaults to 4. On a 64-bit target that under-aligns every pool-allocated object that embeds a type with a stricter requirement — an OS synchronisation primitive, or a 64-bit field accessed atomically.

Christian Becker reported a production crash from this. A `CRITICAL_SECTION` inside a pool-allocated `pj_mutex_t` landed at `addr & 7 == 4`, putting its 8-byte `LockSemaphore` (`CS+0x18`) at offset 60 of a cache line, so it straddled the boundary. ntdll writes that field once with `lock cmpxchg` on first contention (`0 -> -1`) but reads it back with a plain `mov`; the torn read is neither `0` nor `-1`, so ntdll calls `NtSetEvent()` on it, gets `STATUS_INVALID_HANDLE` and raises it. Roughly 1 in 8 misaligned mutexes is a latent crash, and it can only fire in the single first-contention window, which is why it presented as rare and patternless.

### Where it actually bites

`aconfigure.ac` and `pjlib/CMakeLists.txt` already set 8 on 64-bit, so this only ever affected builds that run neither — the MSVC project files, which is what the reporter builds. The `pool.h` fallback is now pointer-sized so those get it too. Blast radius is limited to MSVC for the same reason.

### The autoconf test was also incomplete

It matched a list of target triples, which silently gave 4 to any 64-bit target not mentioned: `amd64-*` (the FreeBSD/OpenBSD/NetBSD triple), `ppc64`/`ppc64le`, `riscv64`, `s390x`, `loongarch64`. Replaced with a pointer-size test. It is compile-only, so it stays correct when cross-compiling. Both branches verified here: 64-bit compiles → 8; `-m32` correctly fails to compile → 4.

### The allocator could not honour alignment at all

`PJ_POOL_ALLOC_T` passed only `sizeof(type)`, so a type's declared alignment was discarded. The clearest case is `struct pj_atomic_t`, which is declared `PJ_ALIGN_DATA_SUFFIX(8)` and holds an `_Atomic` value, yet was allocated with `sizeof` alone — the attribute was silently dropped by the allocator.

`PJ_ALIGNOF()` is added (compiler intrinsics, falling back to `alignof`/`_Alignof`, then to the `sizeof(struct { char c; T t; })` idiom), and `PJ_POOL_ALLOC_T`/`PJ_POOL_ZALLOC_T` pass it to `pj_pool_aligned_alloc()`. That function now treats the pool's own alignment as a **floor** rather than only substituting it for `0`, so an allocation is never *less* aligned than before — the change can only ever increase alignment. `pj_pool_aligned_zalloc()` is added for the zeroing variant.

`pj_atomic_slist` already did this by hand with `pj_pool_aligned_alloc(pool, MEMORY_ALLOCATION_ALIGNMENT, ...)` for both the header and the nodes, and was the only place in the tree that did.

### pj_fifobuf had the same defect independently

Chunks are `[size header][payload]` and the header was `sizeof(unsigned)` = 4, so every payload came back 4 bytes past an aligned address, and the next chunk started at an arbitrary offset. UBSan reports this against every `pj_test_log_item` access in `unittest.c`, and `fifobuf_rolling_test` stores `pj_list` nodes (8-byte pointers) in those chunks.

The header is now pointer-sized, each chunk is rounded up to a multiple of it, and `pj_fifobuf_init()` aligns the buffer start; `capacity()` and `available_size()` round down so they stay truthful (allocating exactly `available_size()` still succeeds, which `unittest.c` relies on). This restores the alignment assertion in `pjlib-test/fifobuf.c` that had been commented out with the note *"fifobuf is no longer aligned"* — it is re-enabled here.

### Testing

- UBSan misalignment reports **12 → 0** (clean `-O2 -fsanitize=address,undefined` build, pjlib-test + pjsip-test)
- Clean build, no new warnings; `pjlib-test` 15/15, `pjlib-util-test` 8/8, `fifobuf_test` passes with the re-enabled assertion
- Verified with a probe against the built library that an over-aligned type (64-byte) now gets its alignment through `PJ_POOL_ALLOC_T`, where it was previously capped at the pool default, while `char` still lands on the pool floor
- `pjsip-test` has 4 pre-existing `tsx_uac`/`tsx_uas` timing failures on this machine; the same tests fail (5 of them) on an unmodified clean build of the same base

Not addressed: `pool_alt.h`'s macros still cannot honour over-aligned types, since that allocator is `malloc`-backed and `malloc` only guarantees `max_align_t`. It is a debug-only allocator.

Reported-by: Christian Becker

### DIFF
--- a/aconfigure
+++ b/aconfigure
@@ -5638,20 +5638,34 @@ printf "%s\n" "#define PJ_M_NAME \"$target_cpu\"" >>confdefs.h
 
 { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking memory alignment" >&5
 printf %s "checking memory alignment... " >&6; }
-case $target in
-    sparc64-* | ia64-* | x86_64-* | arm64-* | aarch64-* | mips64* )
-        printf "%s\n" "#define PJ_POOL_ALIGNMENT 8" >>confdefs.h
+cat confdefs.h - <<_ACEOF >conftest.$ac_ext
+/* end confdefs.h.  */
 
-        { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: 8 bytes" >&5
+int
+main (void)
+{
+int t[sizeof(void *) == 8 ? 1 : -1]; (void)t;
+  ;
+  return 0;
+}
+_ACEOF
+if ac_fn_c_try_compile "$LINENO"
+then :
+
+printf "%s\n" "#define PJ_POOL_ALIGNMENT 8" >>confdefs.h
+
+     { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: 8 bytes" >&5
 printf "%s\n" "8 bytes" >&6; }
-    ;;
-    * )
-        printf "%s\n" "#define PJ_POOL_ALIGNMENT 4" >>confdefs.h
 
-        { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: 4 bytes (default)" >&5
-printf "%s\n" "4 bytes (default)" >&6; }
-    ;;
+else case e in #(
+  e)
+printf "%s\n" "#define PJ_POOL_ALIGNMENT 4" >>confdefs.h
+
+     { printf "%s\n" "$as_me:${as_lineno-$LINENO}: result: 4 bytes" >&5
+printf "%s\n" "4 bytes" >&6; } ;;
 esac
+fi
+rm -f core conftest.err conftest.$ac_objext conftest.beam conftest.$ac_ext
 
 
 ac_header= ac_cache=
--- a/aconfigure.ac
+++ b/aconfigure.ac
@@ -140,19 +140,22 @@ AC_MSG_RESULT([Setting PJ_M_NAME to $target_cpu])
 AC_DEFINE_UNQUOTED(PJ_M_NAME,["$target_cpu"])
 
 dnl
-dnl Memory alignment detection
+dnl Memory alignment detection.
+dnl
+dnl Test the pointer size rather than matching the target triple: a list of
+dnl triples silently gives 4 to any 64 bit target it does not mention (e.g.
+dnl amd64-*-freebsd, ppc64, riscv64, s390x), which under-aligns every pool
+dnl allocated object that embeds an OS primitive or a 64 bit field accessed
+dnl atomically. Compile-only, so it stays correct when cross compiling.
 dnl
 AC_MSG_CHECKING([memory alignment])
-case $target in
-    sparc64-* | ia64-* | x86_64-* | arm64-* | aarch64-* | mips64* )
-        AC_DEFINE(PJ_POOL_ALIGNMENT, 8)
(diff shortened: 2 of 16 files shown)


nanangizz:
Thanks — all five points were valid and are fixed in 318faf120.

**fifobuf unsigned truncation (critical).** Correct, and it was a regression I introduced with the chunk rounding: the original compared `size+SZ` as `pj_size_t` and refused the request, while my `unsigned total` wrapped `0xFFFFFFFF` to `8` and then passed the space check. `total` is a `pj_size_t` now and is narrowed only after the bounds check it has to survive. Verified against a 256-byte fifobuf: `alloc(0xFFFFFFFF)` and `alloc(0xFFFFFFF8)` both return NULL, and a normal `alloc(32)` still succeeds.

**pool-debug API.** Correct — and sharper than stated, since `pool_alt.h` already provides `pj_pool_aligned_alloc`, so the new `pj_pool_aligned_zalloc` was the only one missing. Added the macro, the `_imp` declaration and the implementation in `pool_dbg.c`, and confirmed the library builds under `PJ_POOL_DEBUG` with the symbol present. Worth noting separately: `pjlib-test/pool.c` does not compile under `PJ_POOL_DEBUG` either way — it reaches into `pj_pool_t::block_list`, which the alternate pool has no member for. That fails identically on `origin/master`, so it is pre-existing and I left it alone.

**Symbian exports.** Added both at new ordinals. For the record `pj_pool_aligned_alloc` was already absent while already being used by `atomic_slist.c`, so that build was broken before this PR — but the point stands that `PJ_POOL_ALLOC_T` now makes every typed allocation depend on those symbols, which widens it considerably.

**fifobuf contract too broad.** Correct, `long double` is the counter-example. The doc now limits the guarantee to `sizeof(void*)` and says explicitly that fifobuf has no caller-specified alignment. Reworded the internal `SZ` comment the same way.

**Missing regression coverage.** Added `pool_alloc_t_alignment_test()`: a type declared 32-byte aligned allocated through both macros, the zeroing check, and a type needing less than a custom pool alignment still landing on the pool floor.

Writing it caught two mistakes of my own that are worth recording, because the first version was not load-bearing at all:

- `pj_pool_reset()` inside the loop restarted every iteration from the same offset, so the bump pointer never walked the residues and the test only ever probed one of them. The allocations now accumulate.
- the zeroing check was masked by the aligned allocation before it, which left the bump pointer already 32-byte aligned. There is a second odd-sized filler between them now.

With those fixed the matrix behaves:

| | result |
|---|---|
| both fixes present | OK |
| `PJ_POOL_ALLOC_T` reverted | `[Err: -520]` |
| `PJ_POOL_ZALLOC_T` reverted | `[Err: -530]` |
| restored | OK |

Clean build with no new warnings; pjlib-test 10/10 and pjlib-util-test 8/8.

Co-Authored-By Claude Code

sauwming (approved):

PR #5245, opened 2026-09-08
build: track header dependencies and relink executables automatically

nanangizz:
Editing a public header and rebuilding does not recompile the sources that include it.

The GNU makefiles take header dependencies only from a `.depend` file produced by the manual `depend` target, which shells out to `gcc -M` once per source. Nobody runs `make dep` — `CLAUDE.md` explicitly tells you not to, because a partial run leaves a truncated `.depend` that make then reports as `missing separator`. So in practice a header change recompiled **nothing**.

That is worse than a slow rebuild: the resulting binary links objects built against two different versions of the same header. If the header changed an inline function or a macro that expands to a different call, the two halves disagree and it misbehaves somewhere unrelated to the edit. I hit exactly this while working on a `pool.h` change — `pjsua_auth_test` segfaulted 100% of the time on the incremental build, and a clean rebuild of identical source passed. Debugging that cost more than the original change.

### Fix

Let the compiler emit the dependencies. `CC_DEPFLAGS` is `-MMD -MP` for GCC and Clang, so each object's `.d` file is written as a side effect of compiling it and cannot go stale. `-MP` emits a phony target per header, so deleting or renaming one does not break the build with *"No rule to make target"*.

- `build/rules.mak` passes `$(CC_DEPFLAGS)` in every compile rule and does `-include $(OBJS:$(OBJEXT)=.d)`
- Set in `build/cc-gcc.mak`, and in `build/cc-auto.mak.in` from a new `AC_SUBST(CC_DEPFLAGS)` gated on `$GCC` — true for gcc *and* clang, empty otherwise, so MSVC/armcc behaviour is unchanged
- Where `CC_DEPFLAGS` is set, `make dep` becomes a no-op and the old `.depend` is not included at all. That retires the corrupt-`.depend` failure mode entirely.

### The same target also hid the relink dependencies

`depend` was the only place that recorded executable→library dependencies, appending them to `.depend`. So a test binary was not relinked when a library it links against was rebuilt — `make` would report `'../bin/pjsip-test-...' is up to date` even after `sip_inv.o` and `libpjsip-ua.a` had just been rebuilt. Any A/B test of a library change against `pjsip-test` silently exercised the old binary.

These now use `<APP>_EXTRA_DEP`, an existing hook in `rules.mak` that no module was using. The `echo ... >> .depend` lines are deliberately left in place: they are still the mechanism for compilers where `.depend` is used.

### Testing

On a clean build, touching `pjlib/include/pj/pool.h`:

|  | before | after |
|---|---|---|
| sources recompiled | 0 | 372 |
| executables relinked | 0 | 12 |

`make dep` now prints `Dependencies are generated during compilation; nothing to do.` Clean full build with no new warnings; `pjlib-test` 15/15, `pjlib-util-test` 8/8.

`CLAUDE.md` and `.github/copilot-instructions.md` are updated together, since the `make dep` guidance in them describes exactly the behaviour this changes.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -5,7 +5,7 @@ Open-source multimedia communication library in C (SIP, SDP, RTP, STUN, TURN, IC
 ## Build & Test
 
 - **Build**: `./configure && make -j3` (timeout 120s+, NEVER CANCEL)
-- **Skip `make dep`** — it often produces corrupt `.depend` files. `make` generates deps automatically. If build fails with "missing separator" in `.depend`, run `find . -name "*.depend" -delete` and retry.
+- **`make dep` is not needed** — with GCC/Clang the compiler emits a `.d` file per object (`-MMD -MP`), so header dependencies are always up to date and `make dep` is a no-op. On a tree built before that change, a stale `.depend` can still fail with "missing separator": run `find . -name "*.depend" -delete` and retry.
 - **Targeted build**: `cd pjlib/build && make` (or `pjsip/build`, etc.)
 - **Full clean**: `make clean` (required when switching SSL backends or configure options)
 - Get target arch: `make infotarget` (e.g., `x86_64-pc-linux-gnu`)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -5,7 +5,7 @@ Open-source multimedia communication library in C (SIP, SDP, RTP, STUN, TURN, IC
 ## Build & Test
 
 - **Build**: `./configure && make -j3` (timeout 120s+, NEVER CANCEL)
-- **Skip `make dep`** — it often produces corrupt `.depend` files. `make` generates deps automatically. If build fails with "missing separator" in `.depend`, run `find . -name "*.depend" -delete` and retry.
+- **`make dep` is not needed** — with GCC/Clang the compiler emits a `.d` file per object (`-MMD -MP`), so header dependencies are always up to date and `make dep` is a no-op. On a tree built before that change, a stale `.depend` can still fail with "missing separator": run `find . -name "*.depend" -delete` and retry.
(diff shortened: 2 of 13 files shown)


nanangizz:
Thanks — all five points were real and are fixed in ccf9a07a9.

**Missing/incomplete relink prerequisites.** Correct on every count, and the omission of `pjlib-util-test`, `pjsua` and `pjsystest` was the worst of it: those had no `EXTRA_DEP` at all, so disabling `DEP_FILE` left them with no relink edge whatsoever. Also added `PJTURN_SRV_EXTRA_DEP`, and the video/audio device libraries for `pjmedia-test`.

For the executables that link the whole stack I took the suggestion to use `APP_LIB_FILES` / `APP_LIBXX_FILES` rather than hand-written lists — that is what `pjsip-apps` already recorded in its own `.depend` lines, and it removes the possibility of the list drifting from `LDFLAGS` again. I verified coverage mechanically: every `echo '$(BINDIR)/$(X_EXE): ...'` edge in the tree now has a matching `X_EXTRA_DEP` (10 of 10).

**Objects without `.d` files.** Also correct, and worth fixing rather than documenting. The objects now depend on a sentinel file; creating it is newer than any pre-existing object, so a tree built before this change gets exactly one forced rebuild that produces the `.d` files, and it never changes afterwards. On a clean tree it costs nothing.

Verified on a clean build: touching `pjmedia/src/pjmedia-audiodev/audiodev.c` relinks `pjmedia-test`, touching `pjlib-util/src/pjlib-util/dns.c` relinks `pjlib-util-test`, and in a simulated pre-change tree (objects kept, `.d` files and sentinel deleted) a header edit recompiles 52 sources where it previously recompiled none, with the second build a no-op.

Co-Authored-By Claude Code

sauwming (approved):

PR #5246, opened 2026-09-08
pjsip-test: serialize the lazy init of the process wide UA singletons

nanangizz:
`Default / pjmedia,pjsip` fails intermittently with a module registered twice:

```
[ 1/32] msg_err_test                     [OK] [0.001s]
pjsip-test: ../src/pjsip/sip_endpoint.c:182: pjsip_endpt_register_module:
Assertion `pj_list_find_node(&endpt->module_list, mod) == ((void *)0)' failed.
```

Seen on [run 34197847127](https://github.com/pjsip/pjproject/actions/runs/34197847127/job/101969521185) (`-w 3 --shuffle`, seed 669).

### Cause

`dlg_target_refresh_test` and `inv_offer_answer_test` both initialise the UA layer on demand:

```c
if (pjsip_ua_instance()->id == -1)
    pjsip_ua_init_module(endpt, &ua_param);
```

`pjsip_ua_instance()` is a process-wide singleton, and the check is not serialised against the initialisation. With parallel runners both tests can observe `id == -1` and both proceed to register the same static module, which is what the assertion catches. `inv_offer_answer_test` does the same for `pjsip_inv_usage_instance()`.

It only shows up under `--shuffle` with more than one worker, and only when those two tests happen to be co-scheduled, which is why it appears as a rare flake rather than a consistent failure.

### Fix

Follow `register_modules()` in `tsx_uas_test.c`: claim the initialisation under the critical section, perform it outside, and have the callers that did not claim it wait for the module to become ready.

Deliberately **not** simply wrapping the whole thing in the critical section. `pjsip_endpt_register_module()` locks `endpt->mod_mutex` for write, sleeps 100 ms when the endpoint is already running, and runs the module's `load()`/`start()` callbacks. `pj_enter_critical_section()` documents that the caller must not block while holding it, and a 100 ms sleep inside a *global* critical section stalls every other thread that touches one. It would also introduce a `critical_section → mod_mutex(write)` order against the `mod_mutex(read) → critical_section` dispatch path in `PJ_GRP_LOCK_DEBUG` builds. Thanks @sauwming for catching this on the first revision.

The one thing that differs from `tsx_uas_test.c` is that these singletons are shared between two test *files*, so a file-static counter cannot coordinate them the way `modules_reg_cnt` does within a single file. The claim therefore lives in `test.c` as `init_ua_layer()` and `init_inv_usage()`, which both tests call:

- the critical section is held only across an integer
- `pjsip_ua_init_module()` / `pjsip_inv_usage_init()` run outside it
- callers that did not claim it poll for `mod->id >= 0`, with the same 20 x 50 ms and `PJSIP_ENOTINITIALIZED` fallback as `register_modules()`
- a fast path returns `PJ_SUCCESS` when the singleton is already up, so a run that initialised it through pjsua is unaffected

### Testing

Running just the two tests against each other with three workers and a fixed shuffle seed:

| | assertions |
|---|---|
| before | **2 / 24 seeds** |
| after | **0 / 24 seeds** |

A wider shuffled subset (`dlg_target_refresh_test inv_offer_answer_test tsx_basic_test msg_err_test txdata_test`, `-w 3`) passes across 5 seeds, and both tests still pass single-threaded.

### Not addressed

- This stops the crash, not the underlying singleton semantics: whichever test initialises the UA layer first decides the callbacks for the whole process, so `inv_offer_answer_test`'s `on_dlg_forked` is silently absent when `dlg_target_refresh_test` gets there first. That is pre-existing and would want a different fix (per-test endpoints, or an explicit shared init).
- `tsx_uac_test.c:init_test()` holds the critical section across `pjsip_endpt_register_module()` in the same way the first revision of this PR did. It has the same latent problem, but it is not what CI is tripping over and it is not shared across files, so it seems better as its own change.

Co-Authored-By Claude Code

### DIFF
--- a/pjsip/src/test/dlg_target_refresh_test.c
+++ b/pjsip/src/test/dlg_target_refresh_test.c
@@ -433,21 +433,12 @@ int dlg_target_refresh_test(void)
     pj_str_t uac_target_before;
     int rc = 0;
 
-    /* Init UA layer.
-     *
-     * pjsip_ua_instance() is a process wide singleton that other tests
-     * initialize too, so the check and the initialization must not be
-     * interleaved with theirs. Without this, two tests running in parallel
-     * can both see id == -1 and both register the same module, which trips
-     * the assertion in pjsip_endpt_register_module().
-     */
-    pj_enter_critical_section();
-    if (pjsip_ua_instance()->id == -1) {
+    /* Init UA layer */
+    {
         pjsip_ua_init_param ua_param;
         pj_bzero(&ua_param, sizeof(ua_param));
-        pjsip_ua_init_module(endpt, &ua_param);
+        PJ_TEST_SUCCESS(init_ua_layer(&ua_param), NULL, return -1);
     }
-    pj_leave_critical_section();
 
     PJ_TEST_SUCCESS(pjsip_endpt_register_module(endpt, &mod_dlg_tr_test),
                     NULL, return -2);
--- a/pjsip/src/test/dlg_target_refresh_test.c
+++ b/pjsip/src/test/dlg_target_refresh_test.c
@@ -433,12 +433,21 @@ int dlg_target_refresh_test(void)
     pj_str_t uac_target_before;
     int rc = 0;
 
-    /* Init UA layer */
+    /* Init UA layer.
+     *
+     * pjsip_ua_instance() is a process wide singleton that other tests
+     * initialize too, so the check and the initialization must not be
+     * interleaved with theirs. Without this, two tests running in parallel
+     * can both see id == -1 and both register the same module, which trips
+     * the assertion in pjsip_endpt_register_module().
+     */
+    pj_enter_critical_section();
     if (pjsip_ua_instance()->id == -1) {
         pjsip_ua_init_param ua_param;
         pj_bzero(&ua_param, sizeof(ua_param));
         pjsip_ua_init_module(endpt, &ua_param);
     }
+    pj_leave_critical_section();
 
     PJ_TEST_SUCCESS(pjsip_endpt_register_module(endpt, &mod_dlg_tr_test),
                     NULL, return -2);
--- a/pjsip/src/test/inv_offer_answer_test.c
+++ b/pjsip/src/test/inv_offer_answer_test.c
@@ -813,23 +813,16 @@ int inv_offer_answer_test(void)
     unsigned i;
     int rc = 0;
 
-    /* Init UA layer and inv-usage.
-     *
-     * Both are process wide singletons that other tests initialize too, so
-     * the checks and the initializations must not be interleaved with
-     * theirs. Without this, two tests running in parallel can both see
-     * id == -1 and both register the same module, which trips the assertion
-     * in pjsip_endpt_register_module().
-     */
-    pj_enter_critical_section();
-    if (pjsip_ua_instance()->id == -1) {
+    /* Init UA layer */
+    {
         pjsip_ua_init_param ua_param;
         pj_bzero(&ua_param, sizeof(ua_param));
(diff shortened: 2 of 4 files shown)


sauwming:
Issues
1. Blocking while holding the critical section (violates API contract)
The pj_enter_critical_section() docs (pjlib/include/pj/os.h:1000) explicitly state "MUST NOT call any functions that may cause current thread to block (such as allocating memory, performing I/O, locking mutex, etc.)". But pjsip_endpt_register_module internally:
- Takes endpt->mod_mutex (rw write lock)
- Can pj_thread_sleep(100) if the endpoint is already running
- Runs mod->load() / mod->start() callbacks (arbitrary code)
This is the same issue in tsx_uac_test.c:init_test() — the PR is correctly extending an existing pattern — but the existing pattern itself has this latent problem.
2. Lock-order inversion possible in debug builds
critical_section → mod_mutex(write) (the PR's pattern) vs. mod_mutex(read) → critical_section (RX/TX dispatch path when PJ_GRP_LOCK_DEBUG=1). This is an AB-BA inversion. In default release builds the inner path doesn't touch the critical section, so it's not a practical concern today, but it's worth noting.
3. tsx_uas_test.c has a safer pattern that wasn't followed
tsx_uas_test.c:register_modules() uses the critical section only around a counter, releases it, then does registration — never holding the CS across pjsip_endpt_register_module. That's the safer template. However, the atomic check-then-register semantics needed here (UA singleton has no separate counter) make the tsx_uac_test.c style the natural choice.

nanangizz:
Thanks — you are right on all three, and the first is worse than latent in this PR. Reworked in d2d58b738.

**1. Blocking while holding the critical section.** Confirmed against the code: `pjsip_endpt_register_module()` does `pj_rwmutex_lock_write(endpt->mod_mutex)`, then `pj_thread_sleep(100)` when the endpoint is already running, then `mod->load()` / `mod->start()`. Holding a *global* critical section across a 100 ms sleep stalls every other thread that touches one, which is not merely a contract violation on paper. My patch would have made that reachable from two more tests.

**2. Lock-order inversion.** Agreed — `critical_section → mod_mutex(write)` against `mod_mutex(read) → critical_section` in the `PJ_GRP_LOCK_DEBUG` dispatch path is AB-BA. Not reachable in default release builds, but no reason to add a second instance of it.

**3. Following `tsx_uas_test.c` instead.** Done. The one wrinkle you note — the UA singleton has no counter of its own — is really that the singleton is shared between two *files*, so a file-static counter cannot coordinate them the way `modules_reg_cnt` does within `tsx_uas_test.c`. So the claim lives in `test.c` as `init_ua_layer()` and `init_inv_usage()`, and both tests call it:

- claim the initialization under the critical section, holding it only across an integer
- perform `pjsip_ua_init_module()` / `pjsip_inv_usage_init()` outside it
- callers that did not claim it wait for `mod->id >= 0`, the same 20 x 50 ms poll and `PJSIP_ENOTINITIALIZED` fallback as `register_modules()`

There is also a fast path returning `PJ_SUCCESS` when the singleton is already up, so a run that initialised it through pjsua is unaffected — the previous version relied on the `id == -1` check for that and I did not want to lose it.

Re-verified: with three workers and fixed shuffle seeds the assertion fires on 2 of 24 seeds unfixed and 0 of the same 24 with this version; a wider shuffled subset over 5 seeds passes; both tests still pass single-threaded.

I have left `tsx_uac_test.c:init_test()` alone. It has the same latent problem, but it is not what CI is tripping over and it is not shared across files, so it seemed better as its own change than folded in here — happy to do it separately if you would like.

Co-Authored-By Claude Code

sauwming (approved):

Issue #4817, opened 2026-02-24
[pjmedia] Incorrect video from front camera iPhone 17


mobile-sergey:
### Describe the bug

**Description**
When capturing from the front camera on iPhone 17 (and the above variants), outgoing video is severely corrupted on the remote side: image appears rotated ~90°, with horizontal banding, row misalignment, and the picture repeated 2–3 times horizontally. The same build works correctly with the back camera and on other iOS devices.

**Root cause**
Capture uses `CVPixelBuffer` but does not respect its row stride (bytes per row). On iPhone 17 front camera, `CVPixelBufferGetBytesPerRow()` / `CVPixelBufferGetBytesPerRowOfPlane()` can be larger than `width * bytes_per_pixel` (as in Apple QA1829). The code assumed stride equal to width:

- NV12 (I420): For the UV plane, the code used the Y-plane stride when advancing to the next row in the need_clip path, and did not use a separate UV stride. When the UV plane has its own stride (e.g. on iPhone 17 front), this produces wrong row alignment and banding.
- BGRA (packed): The code copied the whole frame with a single `pj_memcpy(..., stream->frame_size)`, effectively assuming contiguous rows with `stride = width`. When the buffer has a larger stride, this reads past the logical row and causes horizontal repetition and corruption.

**Suggested fix**
In `pjmedia/src/pjmedia-videodev/darwin_dev.m`, in the capture path that copies from `CVPixelBuffer` into the pjmedia buffer:

1. NV12 (I420):

- Get stride for both planes, e.g.

```
stride = CVPixelBufferGetBytesPerRowOfPlane(img, 0);
stride_uv = CVPixelBufferGetBytesPerRowOfPlane(img, 1);
```

- In the Y-plane copy, advance the source pointer by stride (not by width) when copying row-by-row.
- In the UV interleaved copy, when copying row-by-row, advance the source pointer by stride_uv after each row (e.g. after each half-height row), not by the Y stride. For example, after the inner UV loop, use:

`p += (stride_uv - stream->vid_size.w);`
so that the next row starts at the correct offset.

2. BGRA (packed):
Do not copy the whole frame in one block. Use row-by-row copy with actual strides:

- src_stride = CVPixelBufferGetBytesPerRow(img);
- dst_stride = stream->bytes_per_row;
- For each row, copy min(src_stride, dst_stride) bytes, then advance src by src_stride and dst by dst_stride.

[...]

PR #5249, opened 2026-09-08
pjmedia: honour CVPixelBuffer stride on the chroma plane and for packed formats

### LINKED ISSUE #5248: pjmedia: darwin_dev.m ignores CVPixelBuffer stride on the chroma plane and for packed formats
## Summary

`darwin_dev.m`'s capture path assumes `CVPixelBuffer` rows are contiguous in two places, and both are still present on `9b3ff706b`. The symptoms were reported and correctly diagnosed by @mobile-sergey in #4817, which was closed pointing at #4751 — that fixed the orientation half, and the two stride symptoms in the same report were never addressed.

I am opening this rather than commenting on #4817 because the analysis there is @mobile-sergey's and I do not want to reopen someone else's issue; the credit for the diagnosis is entirely theirs. What I can add is that we have been running the fix in production on an iPhone 17 since, and a PR follows.

## The two remaining defects

Each plane of a `CVPixelBuffer` has its own bytes-per-row, and padding on one is not derivable from another (Apple QA1829).

**1. The chroma plane is walked with the luma plane's stride.** The capture path reads `CVPixelBufferGetBytesPerRowOfPlane(img, 0)` once and uses it for both planes:

https://github.com/pjsip/pjproject/blob/9b3ff706b/pjmedia/src/pjmedia-videodev/darwin_dev.m#L608-L610

When the chroma plane is padded differently from the luma plane, every UV row starts at the wrong offset — the horizontal banding and row misalignment in #4817.

The clip decision has the same defect. An NV12 chroma row holds `w/2` UV pairs, i.e. `w` bytes, so "unpadded" means `stride_uv == w`; the luma plane can be unpadded while the chroma plane is not, and `need_clip` is computed from the luma stride alone.

**2. Packed formats are copied as one flat block.** The non-planar path is:

```c
} else {
    pj_memcpy(stream->capture_buf, CVPixelBufferGetBaseAddress(img),
              stream->frame_size);
}
```

That assumes `bytes_per_row == width * bpp`. When the buffer is padded, each destination row begins part way into the next source row, so the image repeats horizontally and shears — the "repeated 2–3 times" in #4817, in BGRA.

## Why this is worth reopening as its own report

The rotation fix in #4751 is real and does resolve the orientation symptom, so on a device where only orientation was wrong the report looks resolved. It is not the same root cause: #4751 is about `AVCaptureDeviceRotationCoordinator` and the capture angle, and it merged on 2026-01-23, a month before #4817 was filed. `darwin_dev.m` has had no commits between #4817 being opened and today, so neither stride path was touched.

Anyone hitting this on current master sees banding and horizontal repetition with a closed issue describing exactly their symptoms.

## Fix

PR to follow, two commits — one per defect, so they can be taken separately. The change is the one we ship: it has been running against an iPhone 17 front camera in a production build for months, and is byte-identical to the patch we carry downstream.

## Suggested labels

`component: pjmedia`, `type: bug`.

laconicman:
Fixes #5248. The symptoms were reported and diagnosed by @mobile-sergey in #4817 — the analysis there is theirs, and this is the fix for the two parts of it that #4751 did not cover.

Two commits, one defect each, so they can be taken separately or one reverted without the other.

**`2f22ae3c` — the chroma plane's own stride.** Each plane of a `CVPixelBuffer` has its own bytes-per-row and padding on one is not derivable from another (Apple QA1829). The capture path read the luma stride once and walked both planes with it, so a differently padded chroma plane put every UV row at the wrong offset. The clip decision needed splitting for the same reason: an NV12 chroma row holds `w/2` UV pairs, i.e. `w` bytes, so "unpadded" means `stride_uv == w`, and the luma plane can be unpadded while the chroma plane is not.

**`520660f7` — packed formats copied row by row.** The non-planar path did one `pj_memcpy()` of `frame_size`, which assumes contiguous rows. Where the buffer is padded, each destination row starts part way into the next source row, so the image repeats horizontally and shears.

The row count is bounded by what the destination can hold — `frame_size / dst_stride`, which is exactly what the single `memcpy` of `frame_size` wrote — so the copy stays inside `capture_buf`, and where the two strides are equal it is byte-identical to the old behaviour. Devices that were working keep working.

## Provenance, stated narrowly

The starting point was a patch we have carried downstream since #4817 was closed. Its exact testing scope, because the earlier wording here implied more than it should have:

- Only the **previous** form of this change was exercised — the one @nanangizz has now shown uses the wrong destination pitch, not what is on this branch.
- One narrow use case, and **never with an orientation switch**.
- The corruption did not reproduce before iPhone 17; it was gone after the patch; no degradation was seen on older devices — again, without orientation switching.

That gap is worth naming rather than glossing: the pitch defect found in review only bites when the negotiated orientation is portrait, which is precisely the case never covered. Production use did not validate this fix; it validated a narrower one, on a path where the two pitches happen to agree.

So the argument for these commits is the code, not the field evidence. Hardware coverage across formats, presets, orientations and cameras is being built separately and will be reported here when it exists rather than asserted now.

## Verified

`darwin_dev.m` compiles clean with `-Wall`, zero errors and zero warnings, matching `upstream/master` exactly.

Worth stating precisely, because it is easy to overclaim here: the file is behind `PJMEDIA_HAS_VIDEO && PJMEDIA_VIDEO_DEV_HAS_DARWIN`, and a build without those defined compiles none of this — my first attempt did exactly that and proved nothing. The figures above are from a build with both defined, confirmed by checking that `stride_uv`, `need_clip_uv` and the row-copy loop survive preprocessing.

**Not verified here:** I have no iPhone 17 to re-run the original repro on, and no automated coverage exists for this path — see below. The behavioural evidence is the production use above, not a test in this PR.

## On a regression test

There is none, and I would rather say so than let it pass unmentioned. The two defects are pure arithmetic over strides, so they are testable in principle, but the arithmetic is inline in `darwin_stream_get_frame()` behind an AVFoundation callback and a real `CVPixelBuffer`. Testing it as it stands needs a device with a padded capture buffer.

Extracting the stride arithmetic into a small pure function that both the callback and a `pjmedia-test` case could call would make it testable on any host — a synthetic padded buffer, a known pattern, an assertion on the output rows. That is a larger change than this fix and I did not want to bundle it. Happy to follow up with it if you would like the coverage.

### DIFF
--- a/pjmedia/src/pjmedia-videodev/darwin_dev.m
+++ b/pjmedia/src/pjmedia-videodev/darwin_dev.m
@@ -692,8 +692,49 @@ static pj_status_t darwin_factory_default_param(pj_pool_t *pool,
             }
         }
     } else {
-        pj_memcpy(stream->capture_buf, CVPixelBufferGetBaseAddress(img),
-                  stream->frame_size);
+        /* Packed formats (e.g. BGRA): the buffer's bytes-per-row can exceed
+         * width * bpp for alignment (Apple QA1829), so a single flat copy of
+         * frame_size reads across row boundaries and repeats the image
+         * horizontally. Copy row by row using the real strides.
+         *
+         * The destination pitch comes from the delivered buffer's own width
+         * rather than from stream->bytes_per_row. The converter reads
+         * capture_buf at orig_size -- the landscape swap of the negotiated
+         * size -- while bytes_per_row follows the negotiated orientation, so
+         * for a portrait-negotiated stream the two disagree. Deriving it here
+         * keeps an unpadded buffer byte-identical to the previous flat copy.
+         */
+        pj_size_t bpp8 = stream->size.w?
+                         (stream->bytes_per_row / stream->size.w): 0;
+        pj_size_t src_stride = CVPixelBufferGetBytesPerRow(img);
+        pj_size_t dst_stride = CVPixelBufferGetWidth(img) * bpp8;
+        pj_size_t row_bytes = (dst_stride < src_stride)? dst_stride:
+                                                         src_stride;
+        pj_size_t src_rows = CVPixelBufferGetHeight(img);
+        /* Rows the destination can hold: this is exactly what the previous
+         * single memcpy of frame_size wrote, so the copy stays within
+         * capture_buf and is byte-identical whenever the strides match.
+         */
+        pj_size_t dst_rows = dst_stride ? (stream->frame_size / dst_stride) : 0;
+        pj_size_t rows = (src_rows < dst_rows) ? src_rows : dst_rows;
+        pj_uint8_t *src = (pj_uint8_t*)CVPixelBufferGetBaseAddress(img);
+        pj_uint8_t *dst = (pj_uint8_t*)stream->capture_buf;
+        pj_size_t i;
+
+        for (i = 0; i < rows; ++i) {
+            pj_memcpy(dst, src, row_bytes);
+            dst += dst_stride;
+            src += src_stride;
+        }
+
+        /* Zero what the source could not fill, as the planar branch above
+         * does for a short buffer. capture_buf is allocated once and reused
+         * for every frame, so an unwritten region does not merely expose
+         * uninitialised memory once -- it keeps displaying the previous
+         * frame's content for as long as the shortfall lasts.
+         */
+        if (rows < dst_rows)
+            pj_bzero(dst, (dst_rows - rows) * dst_stride);
     }
     
     status = pjmedia_vid_dev_conv_resize_and_rotate(&stream->conv, 
--- a/pjmedia/src/pjmedia-videodev/darwin_dev.m
+++ b/pjmedia/src/pjmedia-videodev/darwin_dev.m
@@ -608,11 +608,18 @@ static pj_status_t darwin_factory_default_param(pj_pool_t *pool,
              * For example, resolution 352*288 can have a stride of 384.
              */
             pj_size_t stride = CVPixelBufferGetBytesPerRowOfPlane(img, 0);
+            /* The chroma plane has its own stride and its own padding, and
+             * neither is derivable from the luma plane's (Apple QA1829: query
+             * per-plane row bytes). On the iPhone 17 front camera they differ,
+             * and walking UV with the luma stride misaligns every chroma row,
+             * which is the horizontal banding this fix is for.
+             */
+            pj_size_t stride_uv = CVPixelBufferGetBytesPerRowOfPlane(img, 1);
             /* Image height is not always equal to the video resolution.
              * For example, resolution 352*288 can have a height of 264.
              */
             pj_size_t height = CVPixelBufferGetHeight(img);
-            pj_bool_t need_clip;
+            pj_bool_t need_clip, need_clip_uv;
             
             /* Auto detect rotation */
             if ((stream->vid_size.w > stream->vid_size.h && stride < height) ||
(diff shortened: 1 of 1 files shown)


nanangizz (changes requested):
`2f22ae3c` looks correct — the `stride_uv == w` criterion is right for NV12, and `p_len` is assigned before either branch so splitting the flag doesn't leave it stale.

`520660f7` has a problem: `dst_stride` is `stream->bytes_per_row`, but the consumer reads `capture_buf` as `conv->src_size`, which darwin_dev sets to `orig_size` — the *landscape* orientation of `strm->size` (line 1446). For a portrait-negotiated BGRA stream (288x352) the converter reads at 1408 bytes/row while this writes at 1152, so the image shears and 73728 bytes are never written; the old flat `memcpy` matched because it moved the delivered buffer verbatim. Deriving the pitch from `CVPixelBufferGetWidth(img)` instead would keep the unpadded case byte-identical and fix only the padded one.

Second, the `rows`/`row_bytes` clamping can leave part of `capture_buf` unwritten, and it is `pj_pool_alloc`'d and never zeroed. The file's own comment at 618-620 documents that `CVPixelBufferGetHeight()` can be below the configured height, and the planar branch `pj_memset`s the tail for exactly that — the packed branch needs the same, otherwise those rows ship as uninitialised memory and then stay frozen. That also makes "byte-identical whenever the strides match" true only when `src_rows >= dst_rows`.

The two commits are cleanly separable, so `2f22ae3c` needn't wait on this.

Co-Authored-By Claude Code

laconicman:
Both points fixed in `06686843`, which amends the second commit — `2f22ae3c` is untouched, so a force-push rather than a follow-up commit keeps the two atomic as you noted they are. Re-read rather than diff.

**The pitch.** You are right, and I verified the chain rather than taking it: `strm->size = vfd->size` (line 863) keeps the *negotiated* orientation — lines 856-861 snap to a supported resolution but preserve it, they are not a landscape swap — so `bytes_per_row = strm->size.w * bpp/8` follows portrait, while `orig_size` at 1447 is an explicit landscape swap and is what `pjmedia_vid_dev_conv_create_converter()` receives as the source size. 1408 against 1152 for a 288x352 stream, exactly as you calculated.

`dst_stride` now comes from `CVPixelBufferGetWidth(img)`, with the bytes-per-pixel recovered as `bytes_per_row / size.w` so it stays correct for any packed format rather than assuming BGRA. Unpadded buffers remain byte-identical to the old flat copy.

**The tail.** Also right, and worse than uninitialised: `capture_buf` is `pj_pool_alloc`'d once at stream creation and reused for every frame, so an unwritten region does not expose garbage once — it keeps redisplaying the previous frame's content for as long as the shortfall lasts. Zeroed now, mirroring what the planar branch already does when `CVPixelBufferGetHeight()` falls short. Your point that "byte-identical whenever the strides match" was only true for `src_rows >= dst_rows` was correct, and the commit message no longer claims otherwise.

**On the field evidence, which I had overstated.** The PR description said this was running in production. That was true of the *previous* form of the change, not this one, and the testing behind it was narrower than the wording implied: a single use case, and never with an orientation switch. Which is to say it never exercised the case you found — the two pitches agree in landscape, so the defect was invisible to it. I have rewritten that section rather than leave it standing.

Hardware coverage across formats, presets, orientations and cameras is being built as a separate harness. When it produces something it will be reported here; until then the argument for these commits is the code.

Separately, while tracing the capture path I found something worth its own report and have filed it rather than attach it here: `session_runtime_error:` logs at level 3 and returns, the file registers no observer for `AVCaptureSessionWasInterrupted`, and neither path touches `vid_cb` — so a camera lost mid-call is invisible above the videodev layer, with the stream still reporting active while frames simply stop.

nanangizz (approved):

sauwming (approved):

PR #5255, opened 2026-09-09
pjsua2,pjsua,pjmedia: Forward the original SDP as-is during negotiation for signalling-plane B2BUA/relay use cases

wosrediinanatour:
This extension is intended for signalling-plane B2BUA/relay use cases where the SDP is forwarded verbatim between two call legs.

If call option `PJSUA_CALL_SDP_PASSTHROUGH`  is set, PJSUA(2) keeps and forwards the original local and remote SDP as-is during negotiation, bypassing pjmedia's answer/offer content rewriting  (codec/payload type reassignment, format filtering, etc). The SDP negotiator state machine still transitions normally. 

New integration and unit tests are in place.

Note that this commit is related to, but gets far beyond PR #5143 (see comment https://github.com/pjsip/pjproject/pull/5143#issuecomment-5188666789).

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -669,6 +669,8 @@ jobs:
         python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uas-custom-sdp.xml
         python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uas-amr-sdp.xml
         python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uas-static-pt-no-rtpmap.xml
+        python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uac-sdp-passthrough.xml
+        python3 run.py --exe ../../pjsip-apps/src/3rdparty_media_sample/alt_pjsua mod_sipp.py scripts-sipp/alt-pjsua-uas-sdp-passthrough.xml
     - name: upload artifacts on failure
       if: ${{ failure() }}
       uses: actions/upload-artifact@v4
--- a/pjmedia/include/pjmedia/sdp_neg.h
+++ b/pjmedia/include/pjmedia/sdp_neg.h
@@ -720,6 +720,58 @@ PJ_DECL(pj_status_t) pjmedia_sdp_neg_negotiate( pj_pool_t *pool,
 
 
 /**
+ * Negotiate local and remote SDP without inspecting or modifying their
+ * content. Before calling this function, the SDP negotiator must be in
+ * PJMEDIA_SDP_NEG_STATE_WAIT_NEGO state. After calling this function, the
+ * negotiator state will move to PJMEDIA_SDP_NEG_STATE_DONE.
+ *
+ * Unlike #pjmedia_sdp_neg_negotiate(), this function does not run any
+ * offer/answer content negotiation (no codec/format reconciliation, no
+ * payload type reassignment, no SDP version bump): the local and remote
+ * SDP that were fed into the negotiator (via
+ * #pjmedia_sdp_neg_set_local_answer(), #pjmedia_sdp_neg_set_remote_answer(),
+ * #pjmedia_sdp_neg_modify_local_offer2(), or
+ * #pjmedia_sdp_neg_set_remote_offer()) simply become the active local and
+ * remote SDP verbatim. This is intended for applications that relay SDP
+ * on the signalling plane only (e.g. a signalling-plane B2BUA) and want
+ * the negotiator's state machine to progress normally without pjmedia
+ * altering the SDP content that has already been assembled/forwarded.
+ *
+ * Because this function skips pjmedia's internal payload-type bookkeeping
+ * (the per-negotiator dynamic payload type to codec mapping normally
+ * maintained by the negotiator to keep payload types stable across
+ * multiple offer/answer rounds), mixing calls to this function with
+ * #pjmedia_sdp_neg_negotiate() on the same negotiator instance may cause
+ * the payload types chosen by a subsequent #pjmedia_sdp_neg_negotiate()
+ * call to differ from what was actually used on the wire during a
+ * preceding passthrough round. Applications that use this function should
+ * do so for the entire lifetime of the negotiator instance.
+ *
+ * This function also skips the media-count padding that
+ * #pjmedia_sdp_neg_negotiate() normally performs (e.g. inserting a
+ * matching-but-disabled media line when the answer has fewer "m=" lines
+ * than the offer), so the resulting active local and remote SDP may end
+ * up with a different number of media lines from each other. Applications
+ * that inspect the negotiated SDP media array by index (e.g. pjsua, which
+ * assumes as many media lines as the call's configured media count) must
+ * account for this, e.g. by supplying local/remote SDP pairs with a
+ * matching media count, or by bounds-checking both SDPs' media_count
+ * before indexing.
+ *
+ * @param pool          Pool to allocate memory. The pool's lifetime needs
+ *                      to be valid for the duration of the negotiator.
+ * @param neg           The SDP negotiator instance.
+ *
+ * @return              PJ_SUCCESS on success, or the appropriate error
+ *                      code (e.g. if the negotiator is not in
+ *                      PJMEDIA_SDP_NEG_STATE_WAIT_NEGO state, or if the
+ *                      local/remote SDP to negotiate is not available).
+ */
+PJ_DECL(pj_status_t) pjmedia_sdp_neg_negotiate_passthrough(pj_pool_t *pool,
+                                                pjmedia_sdp_neg *neg);
+
+
+/**
  * Enumeration of customized SDP format matching option flags. See
  * #pjmedia_sdp_neg_register_fmt_match_cb() for more info.
  */
--- a/pjmedia/src/pjmedia/sdp_neg.c
+++ b/pjmedia/src/pjmedia/sdp_neg.c
@@ -2331,6 +2331,52 @@ PJ_DEF(pj_status_t) pjmedia_sdp_neg_negotiate( pj_pool_t *pool,
 }
 
 
+/* Negotiate without inspecting/modifying offer/answer content, for
+ * signalling-plane relay (e.g. B2BUA) use cases. */
+PJ_DEF(pj_status_t) pjmedia_sdp_neg_negotiate_passthrough(pj_pool_t *pool,
+                                                          pjmedia_sdp_neg *neg)
+{
+    PJ_ASSERT_RETURN(pool && neg, PJ_EINVAL);
+
+    /* Must be in STATE_WAIT_NEGO state. */
+    PJ_ASSERT_RETURN(neg->state == PJMEDIA_SDP_NEG_STATE_WAIT_NEGO,
+                     PJMEDIA_SDPNEG_EINSTATE);
+
+    /* Must have remote and local SDP to negotiate. */
+    PJ_ASSERT_RETURN(neg->neg_remote_sdp && neg->neg_local_sdp, PJ_EBUG);
+
+    /* Promote the fed-in local/remote SDP to active. Note:
+     * assign_pt_and_update_map() is intentionally not called here.
+     */
+    neg->active_local_sdp = neg->neg_local_sdp;
+    neg->active_remote_sdp = neg->neg_remote_sdp;
+
+    /* Keep the pool used for allocating the active SDPs */
+    neg->pool_active = pool;
+
+    if (!neg->has_remote_answer) {
+        /* We are answering: this answer will be sent, so update the
+         * last sent SDP (mirrors pjmedia_sdp_neg_negotiate()), without
+         * touching its origin.version.
+         */
+        neg->last_sent = neg->neg_local_sdp;
+    }
+
+    /* State is DONE */
+    neg->state = PJMEDIA_SDP_NEG_STATE_DONE;
+
+    /* Save state */
+    neg->answer_was_remote = neg->has_remote_answer;
+
+    /* Clear temporary SDP */
+    neg->initial_sdp_tmp = NULL;
+    neg->neg_local_sdp = neg->neg_remote_sdp = NULL;
+    neg->has_remote_answer = PJ_FALSE;
+
+    return PJ_SUCCESS;
+}
+
+
 static pj_status_t custom_fmt_match(pj_pool_t *pool,
                                     const pj_str_t *fmt_name,
                                     pjmedia_sdp_media *offer,
--- a/pjmedia/src/test/sdp_neg_test.c
+++ b/pjmedia/src/test/sdp_neg_test.c
@@ -1896,6 +1896,161 @@ static int sdp_neg_static_pt_repr_test(pj_pool_t *pool)
     return 0;
 }
 
+/* Real-world SDP offer/answer pair used to verify that
+ * pjmedia_sdp_neg_negotiate_passthrough() promotes the fed-in SDP verbatim,
+ * without the fmtp reconciliation / payload-type bookkeeping that a normal
+ * pjmedia_sdp_neg_negotiate() would apply. The offer carries 6 payload
+ * types (AMR-WB/AMR/telephone-event), while the answer narrows this down
+ * to 2 with different fmtp parameters than the offer's — exactly the kind
+ * of content a normal negotiation would rewrite.
+ */
+static char pt_offer_str[] =
+    "v=0\r\n"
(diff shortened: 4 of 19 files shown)


wosrediinanatour:
PTAL, @sauwming, @nanangizz, or @copilot ;-)

nanangizz:
@wosrediinanatour
Sure, please wait. I’m still trying to recall and digest what you want to achieve/solve by reviewing your PR history and discussing it with the AI. After that, I’ll check possible alternatives (note that pjsua‑lib is already complex, so I’d prefer to avoid adding another one if possible :)

sauwming:
Whoops, I have requested Copilot review.

@nanangizz , please let @wosrediinanatour know whether to act on the Copilot review, or to ignore it first for now.

nanangizz:
Thanks Florian. Looking back over #2770, #2900, #3320, #3336, #4155, #4473, #4598, #4923 and #4938, it is clear these all serve one project, and #4938 already accepted the signalling-plane B2BUA direction. So this is not a pushback on the goal — but this PR adds permanent public API (`PJSUA_CALL_SDP_PASSTHROUGH` in `pjsua.h` and `symbols.i`, a new field in `pjsip_inv_session`), and I would like to settle the shape before we freeze anything that cannot be taken back.

### What I think is right in this PR

The `pjmedia` + `pjsip-ua` part — the *idea*, at least. Even in the cleanest possible design the negotiated local SDP is *what goes on the wire*, so `pjmedia_sdp_neg_negotiate()` will rewrite the relayed body no matter what pjsua does downstream. And the negotiator is still needed as a **state machine** (pjsip-ua drives offer/answer off `pjmedia_sdp_neg_get_state()`), just not as a *content transformer*. Separating those two roles is the durable insight here, and the unit test is good.

I do think the API shape needs to change — see question 1 — but something along these lines belongs in pjmedia regardless of how the pjsua half is decided. The pjsua half is the open question.

### An observation that may make the pjsua half much smaller

#5068 recently added, for T.38, exactly the primitive a signalling-plane B2BUA needs — `pjsua_media.c:4825`:

```c
if (!is_stream_media(remote_sdp->media[mi])) {
    stop_media_stream(call, mi, PJ_FALSE);
    close_call_med_tp(call_med);          /* release pjsua's transport */
    call_med->type  = PJMEDIA_TYPE_UNKNOWN;
    call_med->state = PJSUA_CALL_MEDIA_NONE;
    call_med->dir   = PJMEDIA_DIR_NONE;
    if (local_sdp->media[mi]->desc.port != 0)
        got_media = PJ_TRUE;              /* don't drop the call */
    continue;
}
```

together with `call->offer_app_managed` (`pjsua_call.c:6059`), which already suppresses the "no media in the offer → 488" rejection at `pjsua_media.c:2600`.

Because that path marks the slot `MEDIA_NONE`/`DIR_NONE` and takes the `continue`, it already gives you three things for free that the current PR does not:

- `check_lock_codec()` skips slots with `state == MEDIA_NONE || dir == DIR_NONE`, so no automatic lock-codec re-INVITE built from pjsua's own regenerated SDP;
- `apply_med_update()` is behind the `is_stream_media()` gate, so `pjmedia_stream_info_from_sdp()` never runs on the relayed SDP;
- `pjmedia_transport_media_start()` lives inside `apply_med_update()`, so pjsua never tries to consume relayed `a=crypto` / `a=candidate` as its own SRTP/ICE parameters. **This one matters most**, as far as I can see from the code: `call_med->tp` generated its own crypto / candidates / ufrag / pwd during `encode_sdp`, and the app then replaced the whole body with the relayed one, so `pjmedia_transport_media_start(tp, pool, local_sdp, remote_sdp, mi)` is asked to match transport state against an SDP it did not author. At best a socket stays bound for nothing; at worst `media_start()` fails, `apply_med_update()` errors out, and the call is dropped for having no usable media. With ICE enabled pjsua would also start connectivity checks toward the relayed remote candidates — endpoints a signalling-plane B2BUA has no business contacting. The underlying issue is simply that pjsua does not know it is meant to be absent from the media plane.

The only reason it does not already cover your case is that `is_stream_media()` keys on the media **type** (`audio`/`video`/`text`), so a relayed `m=audio` still looks like pjsua's own media. Generalising that from "this media type is app-managed" to "**all media on this call is app-managed**" looks to me like a much smaller and more coherent change than threading a passthrough flag through the negotiator and then guarding the fallout case by case — and it reuses machinery that is already in master and already tested.

After that generalisation I think what would still be missing is:

1. app-supplied SDP as a first-class input (today the only route is patching a generated SDP inside `on_call_sdp_created`; pjsip has `pjsip_inv_set_local_sdp()`, pjsua has no equivalent);
2. skipping media transport creation entirely — `pjsua_media_channel_init()` still binds RTP/RTCP sockets that the app-managed path then closes again, which with ICE/STUN is also slow;
3. the passthrough negotiation from this PR;
4. `sort_media()` and the rtcp-fb / media-count rewriting in `pjsua_media_channel_update()`, which are still keyed on media type rather than on the app-managed flag.

### Three questions before choosing the shape

**1. SDP involvement — I think we should design for the most demanding case.**

My working assumption is that you need to be able to parse, validate **and modify** the SDP — rewriting `c=`, ports, possibly `a=crypto` to point at your media server — which is also how I read #4938's *"the media server is in another process or even at another node"*. Designing for pure verbatim relay and then discovering that you need to modify would mean changing an API we had already frozen, so I would rather assume the superset now.

That assumption changes the primitive. "Forward the SDP untouched" is not quite what is needed; what is needed is: **pjmedia stops rewriting the media content, but the offer/answer state machine and the RFC 3264 `o=`/version discipline keep working on whatever body the application supplies.** This PR drops both. It promotes the fed-in SDP verbatim and explicitly does not touch `origin.version`, so a modifying application gets no version management on answers — while `pjmedia_sdp_neg_modify_local_offer2()` *still* forces `o=` from the previous active local SDP on re-offers. Passthrough as implemented also only covers the initial offer and answers: every re-INVITE/UPDATE offer still goes through `modify_local_offer2()`, which pads m-lines, rewrites `o=` and reassigns dynamic payload types. (Copilot's `sdp_neg.c:2352` comment makes the same point.)

Concretely, that suggests a negotiator-level **mode** — set once at creation — under which the input paths (`set_local_answer()`, `modify_local_offer2()`, `send_local_offer()`) stop rewriting media content, while `negotiate()` keeps its `last_sent` comparison and version bump. That is a much smaller delta from `pjmedia_sdp_neg_negotiate()` than a second parallel entry point, and far less likely to drift out of sync with it.

Does that match what you need, or do you need something beyond parse / validate / modify?

**2. Would a call-level "all media is app-managed" work for you**, building on #5068's mechanism above, rather than a flag inside the SDP negotiator?

**3. If you do modify the body, who owns `o=`?** Under the assumption in question 1, pjmedia would keep managing the origin line and version for you, over whatever media content you supply. If instead you intend to relay the far end's `o=` verbatim, then pjmedia should stay out of it entirely. Those are two different contracts and I would rather pick one deliberately than end up with the current split, where re-offers get version management and answers do not.

### Remaining scope

One more thing that would help me settle the shape rather than keep deciding increments: after this PR, what else do you still need before you have a working signalling-plane B2BUA on PJSUA2? A rough list is fine, it does not have to be designed yet. Every PR so far has been individually reasonable, which makes each easy to approve — but I have never seen the shape of the whole, and I would rather agree on the destination once.

### On the Copilot review

It landed while I was writing this. Since the pjsua-lib part is still under discussion, please only act on the **pjmedia** comments for now — I would rather you did not churn on code whose shape may still change:

- **`sdp_neg.c:2352`** — worth addressing. It is the same point as question 1 above: passthrough currently covers the initial offer and answers, but `modify_local_offer2()` still rewrites every re-INVITE/UPDATE offer, and `set_local_answer()` replaces `origin.id`. Its suggestion to extend the test to a second offer/answer round is a good one either way.
- **`sdp_neg.h:771` (Symbian `build.symbian/pjsdpU.def`)** — please skip this one. That file was last touched in 2007 (`1f61a8f8a`) and is missing every `sdp_neg` API added since, including `pjmedia_sdp_neg_modify_local_offer2()`, `pjmedia_sdp_neg_register_fmt_match_cb()` and `pjmedia_sdp_neg_set_answer_multiple_codecs()`. It is not maintained, and adding a single new ordinal would suggest otherwise.
- **`pjsua_call.c:790`** — the finding is correct (a rejected re-INVITE leaves `inv->sdp_passthrough` applied while `call->opt` is restored), but it is in the part still under discussion, so please hold it until we have settled the shape.

I have some further code-level findings on the patch; happy to post them once the direction is clearer, so they land on the version we actually want.

Not blocking, and not asking you to redesign anything yet. I would just like the picture before we commit to public API.

wosrediinanatour:
Thank you for your comment. I really appreciate it! I haven't been aware of https://github.com/pjsip/pjproject/pull/5068, which looks quite interesting.

We have had different use cases and, hence, our requirements shifted over time. We are currently working on a B2BUA, where originating and terminating part are not in our system. Thus, the B2BUA should not interfere:
- SDP offer and answer shall be untouched
- Even validation may interfere, if it releases the call.

But - you are right - maybe we could weaken "untouched". It would be OK to have a modified o= line, if the rest is untouched (_relaxed requirements_).
In the following I will address your questions.

**1. SDP involvement — I think we should design for the most demanding case.**:

> My working assumption is that you need to be able to parse, validate and modify the SDP — rewriting c=, ports, possibly a=crypto to point at your media server — which is also how I read https://github.com/pjsip/pjproject/pull/4938's "the media server is in another process or even at another node".

Both, SDP offer and answer are not in our hand. So, from our perspective, SDP is not modified, but rather given.

> "Forward the SDP untouched" is not quite what is needed; what is needed is: pjmedia stops rewriting the media content, but the offer/answer state machine and the RFC 3264 o=/version discipline keep working on whatever body the application supplies.
> ...
> Does that match what you need, or do you need something beyond parse / validate / modify?

If the o= line is the only line that is modified, then it is OK.

**2. Would a call-level "all media is app-managed" work for you**...
> ... building on https://github.com/pjsip/pjproject/pull/5068's mechanism above, rather than a flag inside the SDP negotiator?

Yes, and no. Yes for "all media is app-managed". According to "flag inside the SDP negotiator", you wrote in 1)

> Concretely, that suggests a negotiator-level mode — set once at creation — under which the input paths (set_local_answer(), modify_local_offer2(), send_local_offer()) stop rewriting media content, while negotiate() keeps its last_sent comparison and version bump. That is a much smaller delta from pjmedia_sdp_neg_negotiate() than a second parallel entry point, and far less likely to drift out of sync with it.

Hence, no.

**3. If you do modify the body, who owns `o=`?** 

> Under the assumption in question 1, pjmedia would keep managing the origin line and version for you, over whatever media content you supply. If instead you intend to relay the far end's o= verbatim, then pjmedia should stay out of it entirely. Those are two different contracts and I would rather pick one deliberately than end up with the current split, where re-offers get version management and answers do not.

Due to our _relaxed requirements_, o= line could be modified by the stack.

**Remaining scope**

> One more thing that would help me settle the shape rather than keep deciding increments: after this PR, what else do you still need before you have a working signalling-plane B2BUA on PJSUA2? A rough list is fine, it does not have to be designed yet. Every PR so far has been individually reasonable, which makes each easy to approve — but I have never seen the shape of the whole, and I would rather agree on the destination once.

I completely agree and appreciate it very much!

**An observation that may make the pjsua half much smaller**

The whole section is quite interesting and gives a big shift, I will have to analyze. But - the direction looks very sound.

> 1. app-supplied SDP as a first-class input (today the only route is patching a generated SDP inside on_call_sdp_created; pjsip has pjsip_inv_set_local_sdp(), pjsua has no equivalent);

This was always the preferred way for me.

> 2. skipping media transport creation entirely — pjsua_media_channel_init() still binds RTP/RTCP sockets that the app-managed >path then closes again, which with ICE/STUN is also slow;

Good to know.

> 3. the passthrough negotiation from this PR;

But instead using a flag for the whole lifetime, as discussed in other sections?

> 4. sort_media() and the rtcp-fb / media-count rewriting in pjsua_media_channel_update(), which are still keyed on media type rather than on the app-managed flag.

Good to know.

nanangizz:
Thanks, that clears up a lot.

Noted: pure relay, offer and answer both given, and `o=` may be modified by the stack. That is exactly the contract I had in mind, so question 3 is settled.

On question 2 you are right and my wording was wrong — it is not either/or. Both are needed: call-level "all media is app-managed" on the pjsua side (no transports, no streams, no lock-codec re-INVITE), *plus* a negotiator-level mode so the body that actually goes on the wire is not rewritten. And yes to your follow-up on item 3: a mode set once at negotiator creation and valid for its whole lifetime, not a per-call variant of `negotiate()`.

One thing I want to be explicit about: **I would like this feature to be as generic as possible.** You mentioned your requirements have shifted before, and if this lands in pjsua-lib it has to serve signalling-plane B2BUAs in general, not a single deployment. "Media content untouched, `o=` managed by the stack" happens to cover both your pure relay today and an application that rewrites `c=`/ports later, at no extra cost — so that is the contract I would like to aim for.

Two things still open:

- *"Even validation may interfere, if it releases the call"* — which validation failures have you actually hit? Not rewriting the SDP is one thing; switching off the checks that protect the stack from malformed SDP is another. I would rather add a lenient or app-decides path than remove validation.
- The remaining-scope list — no hurry, but it is still the piece I need before settling the shape.

wosrediinanatour:
> One thing I want to be explicit about: I would like this feature to be as generic as possible. You mentioned your requirements have shifted before, and if this lands in pjsua-lib it has to serve signalling-plane B2BUAs in general, not a single deployment. "Media content untouched, o= managed by the stack" happens to cover both your pure relay today and an application that rewrites c=/ports later, at no extra cost — so that is the contract I would like to aim for.

I fully agree!

> "Even validation may interfere, if it releases the call" — which validation failures have you actually hit? Not rewriting the SDP is one thing; switching off the checks that protect the stack from malformed SDP is another. 

In the past, "Unsupported media type (PJMEDIA_EUNSUPMEDIATYPE)", but probably not anymore with current version.

> I would rather add a lenient or app-decides path than remove validation.

Agreed.

> The remaining-scope list — no hurry, but it is still the piece I need before settling the shape.

I would say, with all discussed points in place, there would be nothing remaining regarding signalling-plane Signaling-only B2BUA (https://www.rfc-editor.org/info/rfc7092/#section-3.1.2). SDP-Modifying Signaling-only (https://www.rfc-editor.org/info/rfc7092/#section-3.1.3) is _not_ in our scope, neither is a Proxy-B2BUA (https://www.rfc-editor.org/info/rfc7092/#section-3.1.1).

Additionally, a 3rd-party call control is something which should be enabled as well (e.g. https://www.rfc-editor.org/info/rfc3725/#section-4.1). Note that a PR about Late SDP is on the way.

I hope I haven't misunderstood your question. Note that a future app could need something else/additionally, to my knowledge not in a foreseeable future.

nanangizz:
Thanks — that is the picture I needed. §3.1.2, with §3.1.3 covered by the contract anyway, plus 3PCC on top: bounded enough for me.

The shape I would like to aim for:

**pjmedia** — a negotiator-lifetime mode rather than a second `negotiate()` entry point:

```c
PJ_DECL(pj_status_t) pjmedia_sdp_neg_set_passthrough(pjmedia_sdp_neg *neg,
                                                     pj_bool_t passthrough);
```

Under the mode, `negotiate()` skips `create_answer()` / `process_answer()` and `assign_pt_and_update_map()`, but keeps its `last_sent` comparison and version bump; `modify_local_offer2()` likewise skips the m-line padding, the `o=` user/id forcing and the PT remap, so re-offers are covered too. That keeps the state bookkeeping in one function instead of two that have to be kept in sync by hand — and cloning the application's SDP into the `negotiate()` pool also keeps `neg->pool_active` honest, which the current `negotiate_passthrough()` does not.

**pjsip-ua** — keep something like your `inv->sdp_passthrough`, and apply it wherever `sip_inv.c` creates a negotiator (about eight places). The application cannot call the setter itself: with late SDP the negotiator is born only when the offer arrives (`sip_inv.c:2782`).

**pjsua** — call-level "all media is app-managed" (building on #5068), app-supplied SDP as a first-class input, and no media transport creation for such calls.

On the Late SDP PR: late offers look orthogonal to me — message shape rather than body content — and both directions already have some support (`PJSUA_CALL_NO_SDP_OFFER` for UAC, the ticket #1526 path in `pjsua_call_answer2()` for UAS). What does your PR add on top? If it is mostly the interaction above, folding it in here may be simpler than coordinating two PRs.

Could you rework this PR toward that shape? No rush. I will hold my remaining code-level comments until then, so they land on the right version.

wosrediinanatour:
Looks good to me! Thank you for the discussion and guidance! 

> On the Late SDP PR: late offers look orthogonal to me — message shape rather than body content — and both directions already have some support (`PJSUA_CALL_NO_SDP_OFFER` for UAC, the ticket #1526 path in `pjsua_call_answer2()` for UAS). What does your PR add on top? If it is mostly the interaction above, folding it in here may be simpler than coordinating two PRs.

I also see it almost orthogonal and more a PJSUA2/PJSUA thing (explicitly acknowledge a call with a SDP). The development is done by my colleagues - I would like to keep it separate to make things easier (if it makes sense, we will be able to merge anytime) .
 
> Could you rework this PR toward that shape? No rush. I will hold my remaining code-level comments until then, so they land on the right version.

Yes, I plan to do it. Probably, most parts will be done completely from scratch.

PR #5260, opened 2026-09-15
pjmedia: use the mbedTLS CSPRNG to generate SDES-SRTP keys

bennylp:
# pjmedia: use the mbedTLS CSPRNG to generate SDES-SRTP keys

## Problem

`generate_crypto_attr_value()` (`transport_srtp_sdes.c`) selects its random
source by TLS backend: OpenSSL gets `RAND_bytes()`, Apple gets
`SecRandomCopyBytes()`, and everything else — including mbedTLS — falls to the
`#else` arm, which builds the key from `pj_rand()` mixed with a timestamp.

So a `--with-mbedtls` build negotiates SDES-SRTP with keys that are not from a
CSPRNG.

`pj_rand()` is libc `rand()` (`pj/compat/rand.h:43`), and pjlib never seeds it
— the `pj_srand()` call in `pj_init()` is commented out
(`os_core_unix.c:286`). An application that does not seed it itself gets the
same `rand()` sequence on every boot, leaving `pj_get_timestamp()` as the only
per-run variation. That is not a basis for an SRTP master key.

## Performance

The fallback is also slow, by design: it calls `pj_thread_sleep()` on roughly 2
of every 7 key bytes. For a 30-byte `AES_CM_128_HMAC_SHA1_80` key that is
~40 ms per key, against **~1.5 ms** for a seed-and-generate from the DRBG —
roughly 25x faster per key.

`pj_thread_sleep()` is a real `nanosleep()` regardless of `PJ_HAS_THREADS`, so
in a single-threaded build it blocks the same thread that runs
`pjsip_endpt_handle_events()`, stalling signalling during call setup. On a live
call the two key generations showed up as a 67 ms and a 31 ms gap in the log.

## Fix

Add an `#elif` arm for `PJ_SSL_SOCK_IMP_MBEDTLS` that seeds a CTR-DRBG from the
mbedTLS entropy source, plus the matching includes in `transport_srtp.c`.

Additive only: 27 insertions, 0 deletions. OpenSSL, Apple and the fallback are
untouched.

The DRBG is seeded per generation rather than held as global state, which
avoids needing a lock (`mbedtls_ctr_drbg_*` is not thread-safe without
`MBEDTLS_THREADING_C`) and any teardown ordering question. At 1.5 ms it is
still far cheaper than what it replaces.

## Testing

Mbed TLS 3.6.7, single-threaded (`PJ_HAS_THREADS 0`) UA on pjsip + pjmedia,
SIP over TLS through Kamailio 5.7.4 with mandatory SDES-SRTP:

- weak-RNG warning gone
- `SRTP started, keying=SDES, crypto=AES_CM_128_HMAC_SHA1_80`, 31 s of G.711,
  audio verified at the peer
- keys differ within and across processes, confirming the DRBG is seeded
- full rebuild clean, no new warnings

Remote is pjsua with OpenSSL backend.

## Note

The `#else` arm still applies when no TLS backend is compiled in, and deserves
a separate look: `pj/compat/rand.h:48` defines `platform_rand()` as the
constant `1` on a platform lacking `rand()`, behind only a `#warning`. Out of
scope here.


sauwming (approved):

nanangizz (commented):
The change is correct and the reasoning holds up. My main comment is about scope. Details below, with what I verified.

## Verified

The problem statement checks out. `pj_srand()` is indeed commented out in `pj_init()`, and `platform_rand()` does degrade to a constant `1` behind only a `#warning` (`pjlib/include/pj/compat/rand.h:48`). One small correction: the commented `pj_srand()` is at `os_core_unix.c:256`, not `:286`.

Mechanically it is sound. `transport_srtp_sdes.c` is `#include`d at `transport_srtp.c:515` and is `ExcludedFromBuild` in the vcxproj, so putting the includes in `transport_srtp.c` works across all three build systems, and no build-system change is needed. `aconfigure.ac:2318-2330` already puts mbedTLS's `-I` into global `CFLAGS` and its libs into global `LIBS`, so pjmedia compiles and applications link. C89 is fine too, since the declarations sit at the top of the `do {` block before any statement.

I compiled the new arm against Mbed TLS **2.28.8** as well (you tested 3.6.7). Clean, so the API is compatible across 2.x and 3.x.

I also measured the number the design rests on, 200 iterations, 30-byte `AES_CM_128_HMAC_SHA1_80` key:

```
Mbed TLS 2.28.8
seed-per-key : 0.237 ms/key  (200 keys)
reuse DRBG   : 0.0001 ms/key
```

So ~0.24 ms, and your ~1.5 ms is conservative. That is roughly 170x cheaper than the ~40 ms fallback it replaces. Seeding per generation to avoid the lock and the teardown ordering question is clearly the right trade here.

## Main comment: three backends are still on the weak generator

The PR's own framing is that "everything else — including mbedTLS — falls to the `#else` arm". It then fixes one of the four affected backends. Still falling through after this patch:

- **`PJ_SSL_SOCK_IMP_GNUTLS`** — `gnutls_rnd(GNUTLS_RND_KEY, key, len)` from `<gnutls/crypto.h>`, returns 0 on success. GnuTLS documents `GNUTLS_RND_KEY` as "fatal in many sessions if broken", which is exactly an SRTP master key. Three lines.
- **`PJ_SSL_SOCK_IMP_DARWIN`** — live and selectable via `aconfigure.ac:2197`. `SecRandomCopyBytes` is available there exactly as on `PJ_SSL_SOCK_IMP_APPLE`. This looks like a pre-existing oversight in the Apple arm rather than anything introduced here: widening its guard to `(PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_APPLE || PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_DARWIN)` covers it, together with the matching guard on the `<Security/SecRandom.h>` include at `transport_srtp_sdes.c:37`. One line each.
- **`PJ_SSL_SOCK_IMP_SCHANNEL`** — `BCryptGenRandom(NULL, key, len, BCRYPT_USE_SYSTEM_PREFERRED_RNG)`.

Since the motivation is security, landing one backend at a time leaves GnuTLS, Darwin and Schannel builds deriving SRTP master keys from an unseeded `rand()`. The Darwin case in particular is a one-line guard widening, and I would fold that in regardless of what happens with the other two.

## Minor

**Include placement.** `transport_srtp_sdes.c:37` already holds the Apple backend's include behind a `PJ_SSL_SOCK_IMP` guard. The mbedTLS symbols are used only in sdes.c, so they would sit more naturally next to it. The two guards also differ in style: sdes.c:37 tests only `PJ_SSL_SOCK_IMP`, the new one also tests `PJ_HAS_SSL_SOCK`. Worth picking one.

**The retry loop re-seeds.** `do { ... } while (!key_ok)` regenerates the whole key if any byte is zero, which is ~11% of the time for a 30-byte key. On the OpenSSL arm that retry is a cheap `RAND_bytes()` call; on the mbedTLS arm it is a full entropy gather and reseed. At 0.24 ms it does not matter in practice, and hoisting the seed above the `do` would mean restructuring across the other arms, so I would leave it. Noting it only for the record.

## Context, not a blocker

The recurring cause is that pjlib has no crypto-grade RNG API. `pj_create_random_string()` (`pjlib/include/pj/string.h:635`) is also built on `pj_rand()`, and it backs ICE ufrag/pwd and SIP branch/tag values. Every call site that needs real entropy ends up re-implementing this `#if` chain. A `pj_crypto_rand_bytes()` in pjlib, implemented once per backend, would let SDES, DTLS, ICE and the rest share one path. Same thread as the `#else`-arm question you already flagged as out of scope, so just noting it here.

nanangizz (approved):
My review can be treated as a separate issue.

amubiera:
I was just about to create a similar PR. (Here it is for reference #5262)

I would suggest adding `sdes_init()` and `sdes_deinit()` called from `pjmedia_srtp_init_lib`:
```C
sdes_init() {
#if ... PJ_SSL_SOCK_IMP_MBEDTLS ...
    mbedtls_ctr_drbg_init(...);
    mbedtls_entropy_init(...);
    mbedtls_ctr_drbg_seed(...);
#endif
}
```

and
```C
sdes_deinit() {
#if ... PJ_SSL_SOCK_IMP_MBEDTLS ...
    mbedtls_entropy_free(...);
    mbedtls_ctr_drbg_free(...);
#endif
}
```

And the `generate_crypto_attr_value` doesn't have to reinitialize the mbedTLS objects every time:
```C
#elif defined(PJ_HAS_SSL_SOCK) && (PJ_HAS_SSL_SOCK != 0) && \
      (PJ_SSL_SOCK_IMP == PJ_SSL_SOCK_IMP_MBEDTLS)
            int err = mbedtls_ctr_drbg_random(&sdes_mbedtls_ctr_drbg,
                                              (unsigned char*)key,
                                              crypto_suites[cs_idx].cipher_key_len);
            if (err != 0) {
                PJ_LOG(4,(THIS_FILE, "Failed generating random key "
                          "(native err=%d)", err));
                return PJMEDIA_ERRNO_FROM_LIBSRTP(1);
            }
#else
```

bennylp:
Thanks for the comments.

There is a correction regarding the performance quotation in the original post.  
The "~1.5 ms per seed-and-generate" was measured badly — it timed a whole
  process that generated two keys, so it was dominated by process startup and
  dynamic linking rather than by the DRBG. Re-measured properly (200 iterations
  of exactly the code in this PR, Mbed TLS 3.6.7, x86-64):

  | | stated | actual |
  |---|---|---|
  | seed-per-key (this PR) | ~1.5 ms | **~0.007 ms** |
  | vs. the `pj_rand()` fallback | ~25x faster | **~5000x faster** |

  So the PR understated the improvement by more than two orders of magnitude.
  The "25x" and "~1.5 ms" in the description should read ~5000x and ~7 µs.

@nanangizz :
agreed, the suggestions for improving other backends are best done on separate PR.

@nanangizz and @amubiera 
a shared DRBG is not thread-safe. `mbedtls_ctr_drbg_random()` is only thread-safe when `MBEDTLS_THREADING_C` is enabled, and that's off by default. So calls to it need to be protected with a mutex. So performance improvement numbers will vary.

I propose we'll put these as *further optimization* for this work. Let's get this on the source first and we'll see how it goes on the device. :)

@amubiera
I didn't take the personalization string approach (as in your PR) since `mbedtls_ctr_drbg_seed()` is called in every attr generation (rather than just once), and that already pulls 48 entropy bytes plus a 24-byte nonce for every key.

PR #5262, opened 2026-09-16
transport_srtp: Use mbedTLS CTR-DRBG for random crypto key.

amubiera:


amubiera:
Similar to #5260 but with added sdes_init/sdes_deinit for better performance.

amubiera:
Closing since #5260 is preferred.

PR #5263, opened 2026-09-17
CMake: support iOS

sauwming:
The CMake build could not produce a usable iOS library, and did not configure for iOS at all. This makes `cmake -DCMAKE_SYSTEM_NAME=iOS` build working libraries, and makes the bundled Xcode sample projects build against them.

## What was wrong

- **Configuring for iOS failed outright.** CMake makes executables bundles there, and the sample apps are command line programs with no bundle destination.
- **No audio device.** `FindCoreAudio.cmake` required AppKit, which does not exist on iOS, so `PJMEDIA_WITH_AUDIODEV_COREAUDIO` was silently forced `OFF`. Worse than a plain failure: CMake passed `PJMEDIA_AUDIO_DEV_HAS_COREAUDIO=0` while a `config_site.h` carrying the `PJ_CONFIG_IPHONE` sample set it to `1`, so the build was internally inconsistent.
- **No camera capture.** Nothing referenced `darwin_dev.m`.
- **No VideoToolbox codec.** `vid_toolbox.h` was listed among the headers, but no option could enable it.
- **Missing frameworks.** pjlib's own `os_info_iphone.m` uses `UIDevice` and `ssl_sock_apple.m` uses `NSLock`, yet pjlib linked only CoreFoundation. Metal and QT had options but linked nothing, noted in the source as `# TODO: add remaining APIs' dependencies`.
- **The exported package did not resolve.** `PjConfig.cmake.in` never called `find_dependency()` for the framework packages, and CMake silently dropped the unresolved targets rather than failing, producing a link line with no frameworks on it.
- **The sample projects could not build.** They include pjproject from the source tree, where the autotools build writes its generated headers; CMake writes them under the build directory, so the bridging header failed on `pj/config.h:79: fatal error: 'pj/compat/os_auto.h' file not found`.

Every one of these is invisible while building static libraries and surfaces only when an application is finally linked.

## What this adds

**Three options**, each defaulting to the platforms where the backend exists:

| Option | Default | Sets |
|---|---|---|
| `PJMEDIA_WITH_VIDEODEV_DARWIN` | ON (Apple) | `PJMEDIA_VIDEO_DEV_HAS_DARWIN` |
| `PJMEDIA_WITH_VID_TOOLBOX_CODEC` | ON (Apple) | `PJMEDIA_HAS_VID_TOOLBOX_CODEC` |
| `PJMEDIA_WITH_VIDEODEV_METAL` | unchanged | now links its frameworks |

VideoToolbox is a codec rather than a video device, so it is named to match `PJMEDIA_WITH_G7221_CODEC` and its siblings and reaches the headers through `pjmedia-codec/config_auto.h.cm` like every other codec.

**Framework resolution.** Find modules in the style of `FindCoreAudio.cmake`, each selecting its frameworks by platform following the lists `aconfigure.ac` links, and each registered both for install and for `find_dependency()` in the exported config:

| Module | macOS adds | iOS adds |
|---|---|---|
| CoreAudio | CoreServices, AudioUnit, AppKit | CoreFoundation, CFNetwork, AVFoundation, UIKit |
| AVFoundation | AppKit | UIKit |
| VideoToolbox | AppKit | UIKit |
| Metal | AppKit | UIKit |

Foundation is common to all four. pjlib links Foundation plus the platform UI framework directly, which is what the autotools build does globally. pjlib also links the C++ runtime, since it compiles `atomic_queue.cpp` and a pure C consumer would otherwise fail on `std::terminate()`. `pjmedia-codec` links AudioToolbox on Apple, for `AudioConverter` used by `ilbc.c` when `PJMEDIA_ILBC_CODEC_USE_COREAUDIO` is set in `config_site.h`, which CMake cannot see.

**iOS sample projects.** `pjsip-apps/build/Makefile` symlinks the built archives into the sample directories after building pjsua; `PJ_IOS_SAMPLE_LIBS` does the same, and also copies the five generated headers into the source tree where the Xcode projects include them. Copied rather than symlinked on purpose: a later `./aconfigure` writes to those exact paths. The library names differ between the two build systems, so it is a mapping — pjlib is `libpj.a` to the Xcode projects, `pjsua-lib` is `libpjsua.a`, and the bundled codecs are `libg7221codec.a`, `libgsmcodec.a` and `libilbccodec.a`.

It is on by default, because a sample that does not work after a build is not much of a sample, but it probes whether the source tree is writable and turns itself off with a message when it is not.

The pjsua2 sample referenced its archives through `../PJSip-Build/pjproject/`, a layout neither build system produces. The link resolved anyway, by name from `LIBRARY_SEARCH_PATHS`, but all twenty references showed as missing in Xcode; they now point at `Libraries/`, which both build systems populate.

On iOS the command line sample applications are skipped. An iOS consumer links the libraries into an application of their own, which is what the autotools iOS build produces as well.

## Verification

Configured and built with no errors for **iOS device (arm64)**, **iOS simulator (arm64 + x86_64 in one fat pass)** and **macOS** as a regression.

A downstream consumer that force-loads every archive links with **zero undefined symbols**, where before it reported `_OBJC_CLASS_$_UIView`, `UIDevice`, `UIApplication`, `NSLock`, `NSNumber`, `NSThread`, `NSAutoreleasePool`, `std::terminate()` and the `AudioConverter` entry points.

All three bundled iOS sample projects build for a device destination against a CMake tree:

```
ipjsua-swift       BUILD SUCCEEDED
ios-swift-pjsua2   BUILD SUCCEEDED
ipjsua             BUILD SUCCEEDED
```

Turning each new option off removes both the generated define and the object code.

## Known gaps

- **No WebRTC on Apple**, so a CMake build has no WebRTC AEC where the autotools build does. This is not a matter of lifting the existing guard: `webrtc_aec3` does not compile on macOS because `PlatformThreadId` is undefined there, and the NEON flags break the x86_64 half of a simulator build. Only iOS arm64 builds. Since the feature is compiled out of the headers too, nothing references its symbols, and the sample step supplies an empty archive so the Xcode projects still link — which also covers any codec a user switches off.
- `PJMEDIA_HAS_VPX_CODEC` and `PJMEDIA_HAS_OPENH264_CODEC` are `set()` in CMake but have no entry in `config_auto.h.cm`, so those settings never reach the headers. Pre-existing and not Apple-specific.
- `find_package(Pj)` fails on any host where FFMPEG is enabled: `FindFFMPEG.cmake` includes `Pj/GetMacroValue`, but only the Find modules are installed, not the `Pj/` helpers two of them include. Pre-existing.


sauwming:
Correct, and fixed in ca6fdc4. I reproduced it with an installed iOS prefix and a downstream consumer.

The failure turned out to be **quieter than described**, which makes it worse. CMake did not fail on the unresolved targets — it silently dropped them:

```
-- TARGET MISSING: AVFoundation::AVFoundation
-- TARGET MISSING: VideoToolbox::VideoToolbox

link line: -lpjmedia-videodev -lpjmedia-codec        # no frameworks at all
```

So `find_package(Pj)` configured cleanly and generated a link line with no frameworks on it, leaving the application to fail later on undefined AVFoundation and VideoToolbox symbols rather than at configure time.

After adding the option-guarded `find_dependency()` blocks alongside the existing ones for CoreAudio, V4L2, OpenGL and the codecs:

```
-- TARGET EXISTS: AVFoundation::AVFoundation
-- TARGET EXISTS: VideoToolbox::VideoToolbox

link line: -framework AVFoundation -framework CoreGraphics -framework QuartzCore
           -framework VideoToolbox -framework CoreMedia -framework CoreVideo
```

One note for whoever reviews this: a downstream project must link the **namespaced** targets (`Pj::pjmedia-videodev`), not the bare names. My first attempt used `pjmedia-videodev`, which CMake treats as a raw library name and turns into `-lpjmedia-videodev`, bypassing the interface entirely and producing the same empty link line for an unrelated reason.

## Unrelated pre-existing bug found while testing this

`find_package(Pj)` fails outright on any host where FFMPEG is enabled:

```
CMake Error at .../FindFFMPEG.cmake:118 (include):
  include could not find requested file:  Pj/GetMacroValue
CMake Error at .../FindFFMPEG.cmake:121 (pj_get_macro_value):
  Unknown CMake command "pj_get_macro_value".
```

`cmake/Pj/GetMacroValue.cmake` exists in the tree but is never installed — `install(FILES ${_pj_find_modules} ...)` installs the Find modules only, not the `Pj/` helper modules two of them include. It predates this PR (`find_dependency(FFMPEG)` is untouched here) and is not Apple-specific, so I have left it alone. It did not affect the verification above because FFMPEG is off on iOS. Happy to fix it separately if you want.

sauwming:
All three suppressed comments were correct, and fixed in 5a7f6fc. I reproduced the failure first — my earlier verification only *built* the static libraries, which as you note proves nothing, so I linked a downstream consumer that force-loads the archives:

```
  "_OBJC_CLASS_$_UIView"              "_OBJC_CLASS_$_NSLock"
  "_OBJC_CLASS_$_UIDevice"            "_OBJC_CLASS_$_NSNumber"
  "_OBJC_CLASS_$_UIApplication"       "_OBJC_CLASS_$_NSThread"
  "_OBJC_CLASS_$_NSAutoreleasePool"   "_OBJC_CLASS_$_NSNotificationCenter"
  "_OBJC_CLASS_$_NSConstantIntegerNumber"
```

After the fix that count is **zero**.

## The larger problem it exposed

Chasing this found something worse than the three modules. `FindCoreAudio.cmake` required **AppKit**, which does not exist on iOS, so `find_package_handle_standard_args` failed and `PJMEDIA_WITH_AUDIODEV_COREAUDIO` was silently forced `OFF`:

```
COREAUDIO_LIBRARY_APPKIT:FILEPATH=COREAUDIO_LIBRARY_APPKIT-NOTFOUND
PJMEDIA_WITH_AUDIODEV_COREAUDIO:BOOL=OFF
```

**The iOS CMake build had no audio device at all.** Worse than a plain failure: CMake passed `PJMEDIA_AUDIO_DEV_HAS_COREAUDIO=0` on the command line while a `config_site.h` carrying the `PJ_CONFIG_IPHONE` sample sets it to `1`, so the build was internally inconsistent rather than broken loudly. CoreAudio is now `ON` for both iOS device and simulator.

All four modules now select frameworks by platform, following the lists `aconfigure.ac` links:

| Module | macOS adds | iOS adds |
|---|---|---|
| CoreAudio | CoreServices, AudioUnit, AppKit | CoreFoundation, CFNetwork, AVFoundation, UIKit |
| AVFoundation | AppKit | UIKit |
| VideoToolbox | AppKit | UIKit |
| Metal | AppKit | UIKit |

Foundation is common to all four. macOS, iOS device and iOS simulator all still build clean.

## Two pre-existing gaps the same test exposed, left alone

The consumer still does not link, for two reasons unrelated to Apple frameworks:

- **`libpjlib.a` needs the C++ runtime.** `atomic_queue.cpp` pulls `std::terminate()` and `___cxa_begin_catch`, but the package does not propagate the C++ standard library, so a C consumer cannot link pjlib from a CMake install on any platform.
- **`libpjmedia-codec.a` needs AudioToolbox** when `PJMEDIA_ILBC_CODEC_USE_COREAUDIO` is set, for `AudioConverterNew` and friends. That comes from `config_site.h`, which CMake cannot see, so it may need an option.

Both predate this PR and are not iOS-specific. Happy to take either separately.

sauwming:
Both addressed in 922c7ba.

## Source-tree writes

Kept on by default — a sample project that does not work after a build is not much of a sample, and making it opt-in means everyone hits `library 'pj' not found` before discovering the flag. The legitimate half of the concern is the read-only tree, so the step now probes writability at configure time and disables itself with a message rather than failing the build:

```
-- [!] the source tree is not writable, so the iOS sample projects will not be
       given the built libraries. The libraries themselves are unaffected.
```

Verified by `chmod a-w pjsip-apps/src` and reconfiguring. The libraries build exactly as before either way, and `PJ_IOS_SAMPLE_LIBS=OFF` still skips it entirely.

## The pjsua2 project's references

The premise is right and worth fixing, but the stated consequence does not hold: the sample builds. `LIBRARY_SEARCH_PATHS` for that target includes `.../ios-swift-pjsua2/Libraries`, so Xcode emits `-l<name>` and the linker resolves the archives from there regardless of what the `PBXFileReference` paths say. I had already built all three samples end to end against a CMake tree before this review.

What it *does* break is the IDE: all twenty references show as missing in Xcode's navigator, which is a poor first impression for anyone opening the sample. `../PJSip-Build/pjproject/` is a layout neither build system produces — the autotools Makefile and the new CMake step both populate `Libraries/`. So the references now point there, which is correct for both.

All three samples still build for a device destination, and macOS and the iOS simulator are unaffected.

sauwming:
All three correct, fixed in c6e7db8.

**Exported SDK paths.** Confirmed against an installed package — it carried the build machine's SDK, version included:

```
/Applications/Xcode.app/.../SDKs/iPhoneOS26.2.sdk/System/Library/Frameworks/UIKit.framework
```

Notably the video frameworks did *not* appear, because those already went through imported targets — which is exactly the difference you point at. `pjlib` and `pjmedia-codec` now do the same via two new installed Find modules, `AppleFoundation` and `AudioToolbox`, with matching `find_dependency()` calls. The exported package now contains **zero** absolute framework paths.

**A bug that fix exposed.** The Find modules created plain `INTERFACE` targets rather than `IMPORTED` ones. That is wrong for a Find module generally, and it broke the moment `pjlib` linked one `PUBLIC`:

```
install(EXPORT "PjTargets" ...) includes target "pjlib" which requires
target "applefoundation" that is not in any export set.
```

The existing modules got away with it only because every consumer linked them `PRIVATE`. All six now create `IMPORTED` targets, which never need exporting.

**Provider-agnostic staging.** Right, and worse than described: the placeholder made it silent. Keying on bundled target names meant a system provider was handed an empty archive for an enabled codec, and the sample would fail on unresolved symbols with nothing explaining why. It now keys on the `Pj::Dep::` aliases, and anything genuinely not built is announced rather than substituted quietly:

```
-- [!] libwebrtc.a is not built by this configuration; the iOS sample
       projects get an empty archive in its place
```

Only `libwebrtc.a` reports, which is the one case that is actually unavailable on Apple.

Verified: macOS, iOS device and iOS simulator all build clean; the exported package has no absolute framework paths; a downstream consumer still links with no undefined symbols; all three iOS sample projects still build.

sauwming:
This has also been tested on Bitrise, which now uses CMake to build PJSIP libraries for iOS.

nanangizz (approved):

PR #5264, opened 2026-09-17
Fixes to support zero thread on Posix

bennylp:
# Fixes to support zero thread on POSIX

`PJ_HAS_THREADS 0` is a documented configuration, but it has had almost no maintenance. So the path is neither compiled nor run upstream, and a range of independent defects had accumulated. A single-threaded SIP user agent built on this configuration aborts within a second of the first inbound RTP packet.

This makes the setting work on POSIX. Windows does not honour it and this is not fixed by this PR -- see the last section.

The threaded build is unaffected.

## Problems fixed

These are independent of one another; none is the "main" one.

**1a. Added `--disable-threads`.** Calling this will set `PJ_HAS_THREADS=0` **and** remove `-lpthread` from linking, to make sure that nothing is using pthread. Previously, we could not exclude `-lpthread` from linking. But this requires manually disabling other components that use thread (such as webrtc), so the full command is:

```
./configure --disable-threads --disable-libwebrtc
```

**1b. Same with CMake.** To disable threading with CMake build system:

```
cmake -S . -B build  -DPJLIB_WITH_THREADS=OFF \
                     -DPJMEDIA_WITH_WEBRTC_AEC=OFF \
                     -DPJMEDIA_WITH_WEBRTC_AEC3=OFF
cmake --build build
```
**2. `pj_thread_this()` asserts and returns NULL**. Some callers already assume this will be NULL in no-thread setting (e.g. in ioqueue_select), which it does, but this is **not** in the spec. Moreover, there is an assertion. 

Fixed by returning a static descriptor of the main thread (same as in `os_core_symbian.cpp`) for the no-thread build. `pj_shutdown()` zeroes the descriptor again, mirroring the `pj_bzero(main_thread_desc)` the threaded path already does, so init/shutdown/init starts clean. `pj_thread_get_name()` returns the real name rather than "".

**3. pthread is still linked.** Many accesses to pthread are not guarded by `#if PJ_HAS_THREADS`, including in upper layer code (such as in `alsa_dev.c`). 

**4. Registering external threads silently succeeded.** Anything that registers external thread must not be allowed.  Allowing it will introduce concurrency problem, because caller thinks that pjlib can be accessed from another thread, but in reality synchronization primitives are no-op.

**5. Thread teardown silently succeeded.** `pj_thread_destroy()` and `pj_thread_resume()` had no `PJ_HAS_THREADS` guard at all and returned `PJ_SUCCESS`. They now follow the same rule as create and join.

**6. Seven assertions could never fire.** They were written `pj_assert("...")` rather than `pj_assert(!"...")`. Two of them also named the wrong function in the string. They are removed rather than corrected. They have never run in any release, so arming them could only turn working builds into aborting ones.

**7. Group lock owner.** Group lock now can use `pj_thread_this()` rather than `(pj_thread_t*)-1` to track group lock owner. Same code for both thread and no-thread build.

**8. An error was reported as a bare integer.** `pj_thread_set_prio()` returned `1` on the no-threads path. `pj_status_t` values below `PJ_ERRNO_START` are interpreted as native OS error codes, so `pj_strerror()` reports nonsense. It now returns `PJ_EINVALIDOP`, matching the Windows implementation.

**9. pjlib-test could not run.** The unit-test library supports single threaded run, but the tests are aborted in the app (pjlib-test) rather than the library. Now it works by excluding some tests when threading is disabled and ignoring `-w N` option (which is the default in CI).

**10. Added CI for PJ_HAS_THREADS=0**. It builds everything but only runs `pjlib-test`. It also checks that none of the upper layer unit test executables nor **pjsua** references a pthread symbol. Note that this doesn't mean the application will work in no-thread setting (e.g. `pjsua` won't).

## Smaller corrections

Header documentation that did not match the implementation:

- `pj_thread_get_prio()`, `pj_thread_get_prio_min()` and `pj_thread_get_prio_max()` all document "-1 on error" but returned 1, 0 and 0 with threading disabled -- values indistinguishable from a valid priority. They now return -1.
- `pj_thread_get_os_handle()` promised a native handle without mentioning that it returns NULL when there is no native thread to refer to.
- `pj_thread_get_prio_max()` described its return value as "Minimum thread priority value", copied from `pj_thread_get_prio_min()`.
- `pj_thread_get_name()` had two `#if` branches that did the same thing, one of them via a redundant self-cast.

## Conventions established

There are some conventions that are already in the code with regard to what should be expected with `PJ_HAS_THREADS=0` setting, but these conventions are not explicitly stated anywhere. This PR writes those conventions as comments in `os.h` and `os_core_unix.c`.

**Guard the code, not the types.** Any calls to pthread/sched_xx/etc are forbidden, but simply declaring pthread types is fine. This rule is stated in `os_core_unix.c`, at the `#include <pthread.h>` site.

**Creating or registering a thread asserts; asking about the current thread does not.** `pj_thread_create()`, `join`, `register` and `init` assert, but functions that only report on the current thread -- `pj_thread_this()`, `is_registered()`, the priority getters and setter, `get_os_handle()` -- just return a value without asserting, so portable code can call them without a `PJ_HAS_THREADS` check at every call site. The rule is documented in `os.h`.

**Synchronisation primitives degrade to accepted no-ops.** Following similar convention in `pj_mutex_t` and `pj_sem_t`, operations will always succeed.

## Windows

A few fixes are included: the dead assertions are removed, and `pj_thread_get_os_handle()` no longer returns NULL for a handle that is actually valid. But `PJ_HAS_THREADS` is still not enforced on Windows -- thread creation and registration ignore it, so most threading functionality keeps working -- and this PR does not change that.

## Note on aconfigure

`aconfigure` was regenerated with autoconf 2.72. Besides the new option it picks up a few pre-existing formatting differences around `PJ_POOL_ALIGNMENT`, where the committed file had drifted from what `aconfigure.ac` generates. They are placement only, with no change in behaviour.


sauwming (approved):

CLAUDE.md
```diff
@@ -125,6 +129,39 @@ https://github.com/pjsip/pjproject_docs
...
+  `pj_mutex_t`, `pj_sem_t`, `pj_rwmutex_t`, `pj_event_t` and `pj_barrier_t`.
+  `pj_event_wait()` returns immediately since blocking could only hang;
+  `pj_barrier_wait()` returns `PJ_TRUE`.
+- **Windows does not honour the setting** — `pj_thread_create()` still creates threads
+  in `os_core_win32.c`. Do not assume the rules above hold there.
+- CI asserts that the unit test executables and `pjsua` reference no pthread symbol:
+  `nm -D -u <binary> | grep -E ' U (pthread_|sched_get)'` must find nothing.
+- pjlib-test: tests that need threads are excluded via the `INCLUDE_*` macros in
+  `pjlib/src/pjlib-test/test.h`, and `-w N` is ignored.
+
```
nanangizz, line 143: - From [here](https://code.claude.com/docs/en/memory#write-effective-instructions): `target under 200 lines per CLAUDE.md file` (not sure if it is still applicable), so perhaps this can be shortened somehow?
- Also update `.github/copilot-instruction.md`? FYI, so far we keep them in sync.

bennylp: Done

nanangizz (approved):

PR #5265, opened 2026-09-17
Added support for late SDP negotiation for INVITE and reINVITE.

oleopold:
Added support for custom ACK sending callback in PJSUA2 for both synchronous and asynchronous. Added tests for late SDP negotiation.


pjsip/src/pjsua2/endpoint.cpp
```diff
@@ -2330,6 +2348,7 @@ void Endpoint::libInit(const EpConfig &prmEpConfig) PJSUA2_THROW(Error)
                                         = &Endpoint::on_call_tsx_terminate_session;
     ua_cfg.cb.on_call_media_state       = &Endpoint::on_call_media_state;
     ua_cfg.cb.on_call_sdp_created       = &Endpoint::on_call_sdp_created;
+    ua_cfg.cb.on_call_send_ack          = &Endpoint::on_call_send_ack;
```
sauwming, line 2351: **1. This installs the callback for every PJSUA2 application.**

`ua_cfg.cb.on_call_send_ack` is set unconditionally, so `pjsua_call_subsys_init()` sets `inv_cb.on_send_ack`, and `inv_send_ack()` then takes the manual branch for *all* 2xx ACKs in every PJSUA2 app — not just ones that override `onCallSendAck()`.

On its own that would be harmless if the default always sent the ACK, but combined with the NULL-`Call` early return below it silently changes ACK behaviour for existing applications. Either install it only when the application actually overrides the callback, or guarantee the default always sends.

sauwming (commented):
Thanks for this — late SDP / offerless INVITE support is a real gap and the negotiation core here is going in the right direction. I'd like to see it land, but there are a few things to sort out first, mostly around the new `on_call_send_ack` callback.

**The main concern is structural.** As designed, `on_call_send_ack` converts a framework-guaranteed action into an application obligation with a silent no-op path. That's why three separate defects cluster on it (comments 1-3 below): the PJSUA2 binding installs it for *every* application, the handler returns without sending when the `Call` lookup fails, and the default implementation can throw a C++ exception through C frames while the dialog lock is held. Net effect is that an existing PJSUA2 application that never asked for this can now fail to ACK a 200/OK.

Two ways out, either is fine by me:

- Make the callback advisory — the application supplies the answer SDP (or a modified tdata) and the framework always sends the ACK. This removes the whole class of failure.
- Keep the current shape, but guarantee the fallback send on every path and wrap the PJSUA2 invocation in try/catch.

If you'd like this to move faster, splitting it would also work: the late-SDP negotiation core in this PR, and `on_call_send_ack` in a follow-up where the API can be discussed on its own.

**Blocking before merge:** comments 1-5 (the ACK regression, the exception escape, the ignored reject `code`) and comment 10 — `on_call_send_ack` is inserted in the middle of `pjsua_callback`, which shifts the offsets of every member after it. That's an ABI break for anything linked against an older shared pjsua, and it's free to avoid by appending at the end of the struct.

**Would like addressed:** comments 6-8. **Follow-ups, not blocking:** comments 9, 11, 12.

Both `sip_inv.c` and `pjsua_call.c` build clean with `-Wall` here. Nice to see the SIPp scenarios included for both UAC and UAS.

pjsip/src/pjsua2/endpoint.cpp
```diff
@@ -1561,6 +1561,20 @@ void Endpoint::on_call_sdp_created(pjsua_call_id call_id,
     }
 }
 
+void Endpoint::on_call_send_ack(pjsua_call_id call_id,
+                                pjsip_rx_data *rdata)
+{
+    Call *call = Call::lookup(call_id);
+    if (!call) {
```
sauwming, line 1568: **2. ACK is never sent when the lookup fails.**

`Call::lookup()` returns NULL whenever `user_data` isn't set — the application destroyed the `Call` before the 200/OK arrived, or this is a transfer/replace call whose `Call` isn't attached yet. We return here without sending anything, and nothing downstream compensates.

Consequence: `inv` stays in `CONNECTING`, the peer retransmits the 200/OK until it times out, and any subsequent BYE is parked indefinitely by the "Delaying BYE request until ACK is received" path in `pjsip_inv_send_msg()`. Before this PR the framework always sent the ACK.

Suggest falling back rather than returning:

```c
    Call *call = Call::lookup(call_id);
    if (!call) {
        pjsua_call_send_ack(call_id, rdata->msg_info.cseq->cseq, NULL);
        return;
    }
```

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -7365,6 +7463,31 @@ static pj_bool_t pjsua_call_on_uac_tsx_terminate_session(
...
+ * send an ACK request for the 2xx response of INVITE, and the application
+ * has asked to handle the ACK transmission manually (by implementing
+ * on_call_send_ack()).
+ */
+static void pjsua_call_on_send_ack(pjsip_inv_session *inv,
+                                   pjsip_rx_data *rdata)
+{
+    pjsua_call *call = (pjsua_call*) inv->dlg->mod_data[pjsua_var.mod.id];
+
+    if (!call)
```
sauwming, line 7477: **3. Same early return at the C level.**

Narrower than the PJSUA2 case — the `else` branch below does fall back to `pjsua_call_send_ack()` when no application callback is set — but if `dlg->mod_data[pjsua_var.mod.id]` has been cleared we still return without ACKing. Worth handling here too, since `rdata` alone is enough to send it.

pjsip/include/pjsua2/call.hpp
```diff
@@ -2103,6 +2133,20 @@ class Call
...
+     * after it receives an incoming 2xx response for INVITE (e.g. when the
+     * 2xx response carries a late SDP offer). The default implementation
+     * simply sends the ACK without any SDP answer, using sendAck().
+     * Application can override this callback to attach an SDP answer to
+     * the ACK (e.g. for a late offer scenario), or to delay sending the
+     * ACK, by calling sendAck() itself.
+     *
+     * @param prm       Callback parameter.
+     */
+    virtual void onCallSendAck(OnCallSendAckParam &prm)
```
sauwming, line 2147: **4. The default implementation can throw through C frames.**

`sendAck()` raises `Error` via `PJSUA2_CHECK_RAISE_ERROR2`, and `Endpoint::on_call_send_ack()` has no try/catch (the only `catch` in endpoint.cpp is at line 700). The exception therefore unwinds through `inv_send_ack` -> `mod_inv_on_tsx_state` -> the tsx/dialog callbacks — C frames, with the dialog lock held — giving `std::terminate` or at best a leaked lock.

This is reachable in normal operation: `acquire_call()` inside `pjsua_call_send_ack()` legitimately fails with `PJ_ETIMEDOUT` or `PJSIP_ESESSIONTERMINATED`. Every other pjsua2 default callback is a no-op and cannot throw, so the surrounding code isn't written to expect this. Please wrap the invocation in `Endpoint::on_call_send_ack()`.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -6133,10 +6196,45 @@ static pj_status_t pjsua_call_on_rx_reinvite(pjsip_inv_session *inv,
...
+
+            status = pjsip_inv_send_msg(inv, response);
+            if (status != PJ_SUCCESS) {
+                PJ_PERROR(3, (THIS_FILE, status,
+                              "Failed to send initial answer"));
+                return status;
+            }
+        }
+
+        call->opt = opt;
```
sauwming, line 6234: **5. The application's rejection code is ignored.**

`code` is passed to `on_call_rx_reinvite()` above but never checked here, unlike the existing paths at lines 6093 and 6108 which do `if (code != PJSIP_SC_OK) goto on_return;`.

So an application that sets `code = 488` or `491` to reject an offerless re-INVITE is overridden: we fall through to `!PJ_SUCCESS`, sip_inv generates a 200/OK with a fresh offer, and `call->opt = opt` is applied even for the request the application just rejected.

pjsip/include/pjsua-lib/pjsua.h
```diff
@@ -1496,6 +1496,23 @@ typedef struct pjsua_callback
...
+     * the ACK request itself, using #pjsua_call_send_ack(), which allows an
+     * SDP answer to be attached to the ACK. This also allows the application
+     * to delay sending the ACK, e.g. while waiting for the SDP answer to be
+     * generated by 3PCC scenarios.
+     *
+     * @param call_id   The call index.
+     * @param rdata     The received 2xx response that triggered the need to
+     *                  send an ACK request.
+     */
+    void (*on_call_send_ack)(pjsua_call_id call_id,
```
sauwming, line 1513: **6. ABI break — please append at the end of the struct.**

Inserting `on_call_send_ack` between `on_call_sdp_created` and `on_stream_precreate` shifts the offsets of every following member of `pjsua_callback`. An application built against an older shared pjsua writes its callbacks at the old offsets, so everything after `on_call_sdp_created` gets misbound at runtime — a nasty failure to debug in the field.

Moving the new member to the end of the struct costs nothing and avoids it entirely.

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -2671,7 +2671,10 @@ static pj_status_t inv_check_sdp_in_incoming_msg( pjsip_inv_session *inv,
      *
      * See also tickets #657, #1644, #1764, and #2123 for more info.
      */
-    if (tsx_inv_data->sdp_done) {
+        if (tsx_inv_data->sdp_done &&
```
sauwming, line 2674: **7. This bypass is wider than it needs to be.**

The new `tsx->role == UAC && !has_sdp && 2xx` condition skips the *entire* multiple-offer/answer guard, including the To-tag/forking check and `accept_multiple_sdp_answers`.

Concrete case it breaks — offerless INVITE with 100rel: the remote puts its offer in a reliable 183 and we answer in the PRACK, so `inv_handle_incoming_reliable_response()` sets `tsx_inv_data->sdp_done = PJ_TRUE` on the INVITE tsx while `has_sdp` stays `PJ_FALSE` (it comes from `tx_data_has_sdp()` at line 4154, and our INVITE carried no SDP). If the 200/OK then repeats the SDP — common, and explicitly meant to be ignored per RFC 6337 3.1.1 — the negotiator is already in `DONE`, so the body is now treated as a brand-new offer: `set_remote_offer()` -> `on_rx_offer2()`, a second offer/answer exchange in the same transaction, and an unnecessary media restart.

I'd narrow the condition to "no SDP has been negotiated yet for this offerless UAC INVITE" rather than skipping the guard wholesale. Worth confirming against a reliable-183 trace before reworking, in case I've misread the interaction.

Minor, same hunk: the reindented `if` sits at 8 spaces while its body stays at the old 4-space level, and the continuation lines use ad-hoc indents. Project style is 4-space/K&R.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -3182,7 +3193,9 @@ PJ_DEF(pj_status_t) pjsua_call_answer2(pjsua_call_id call_id,
      * Or if initial answer is not sent yet, we will answer the call after
      * initial answer is sent (see #1923).
      */
-    if (call->med_ch_cb || !call->inv->last_answer) {
+    if (call->med_ch_cb ||
+        (call->inv->state < PJSIP_INV_STATE_CONFIRMED &&
```
sauwming, line 3197: **8. This can hit an assertion in debug builds.**

Narrowing the condition to `state < PJSIP_INV_STATE_CONFIRMED` means the `CONFIRMED` + `last_answer == NULL` case no longer defers — it falls through to `pjsip_inv_answer()`, which opens with `PJ_ASSERT_RETURN(inv->last_answer, PJ_EINVALIDOP)`. In any build without `NDEBUG` that's an `abort()`.

That state is exactly what an application hits calling `pjsua_call_answer()` from inside `on_call_rx_reinvite()` — `invite_tsx` is the UAS re-INVITE in TRYING with no 100 sent yet — which is the workflow this PR is encouraging. Either keep deferring in that case, or return `PJ_EINVALIDOP` explicitly before reaching `pjsip_inv_answer()`.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -3284,6 +3297,56 @@ pjsua_call_answer_with_sdp(pjsua_call_id call_id,
...
+                     PJ_EINVAL);
+
+    PJ_LOG(4,(THIS_FILE, "Sending ACK for call %d", call_id));
+    pj_log_push_indent();
+
+    status = acquire_call("pjsua_call_send_ack()", call_id, &call, &dlg);
+    if (status != PJ_SUCCESS)
+        goto on_return;
+
+    if (sdp) {
```
sauwming, line 3323: **9. Two guards missing in `pjsua_call_send_ack()`.**

`call->inv` is passed to `pjsip_inv_set_sdp_answer()`/`pjsip_inv_create_ack()` without a NULL check — `acquire_call()` succeeds when only `call->async_call.dlg` is set, so `call->inv` can legitimately be NULL here.

Separately: if the 2xx carried no offer the negotiator is in `DONE`, and `pjmedia_sdp_neg_set_local_answer()` asserts `state == REMOTE_OFFER || WAIT_NEGO`. An application that passes an SDP in that situation gets an `abort()` in a debug build instead of `PJMEDIA_SDPNEG_EINSTATE` back. Worth validating the negotiator state before calling it.

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -3118,6 +3121,56 @@ PJ_DEF(pj_status_t) pjsip_inv_answer(   pjsip_inv_session *inv,
...
     }
 
+    /* An asynchronous confirmed UAS re-INVITE without an SDP offer is
+     * answered with a new local offer in the 2xx response. */
+    if (st_code/100 == 2 && inv->state == PJSIP_INV_STATE_CONFIRMED &&
+        inv->invite_tsx && inv->neg &&
+        pjmedia_sdp_neg_get_state(inv->neg) == PJMEDIA_SDP_NEG_STATE_DONE)
+    {
+        struct tsx_inv_data *tsx_inv_data;
+        pjmedia_sdp_session *offer = (pjmedia_sdp_session*)local_sdp;
```
sauwming, line 3131: **10. Stale offer is re-sent when `on_create_offer()` fails.**

Two things here. The cast discards `const` on a caller-owned parameter, and `on_create_offer()` implementations (`pjsua_call_on_create_offer()`) may leave `*offer` untouched when they fail.

When that happens we fall through to `pjmedia_sdp_neg_send_local_offer()` below and re-offer the stale active SDP, even though media re-initialisation just failed. A failure in the callback should produce an error response rather than an offer the application can't back.

pjsip/src/pjsip-ua/sip_inv.c
```diff
@@ -3118,6 +3121,56 @@ PJ_DEF(pj_status_t) pjsip_inv_answer(   pjsip_inv_session *inv,
...
+                            inv->sdp_neg_flags, offer);
+            } else {
+                status = pjmedia_sdp_neg_send_local_offer(
+                            inv->pool_prov, inv->neg, &local_offer);
+            }
+
+            if (status == PJ_SUCCESS && !local_offer)
+                status = pjmedia_sdp_neg_get_neg_local(inv->neg, &local_offer);
+            if (status == PJ_SUCCESS && local_offer)
+                last_res->msg->body = create_sdp_body(last_res->pool, local_offer);
```
sauwming, line 3159: **11. This assignment is dead.**

`process_answer()` runs immediately after (line 3175) with the negotiator in `LOCAL_OFFER`. It takes the `neg_state == PJMEDIA_SDP_NEG_STATE_LOCAL_OFFER` branch, retrieves the same SDP via `pjmedia_sdp_neg_get_neg_local()`, and unconditionally does `tdata->msg->body = create_sdp_body(tdata->pool, sdp)` — overwriting what's set here.

Harmless beyond a redundant clone into the tdata pool, but it obscures the intent. Not blocking.

pjsip/src/pjsua-lib/pjsua_call.c
```diff
@@ -6133,10 +6196,45 @@ static pj_status_t pjsua_call_on_rx_reinvite(pjsip_inv_session *inv,
...
+            pjsip_tx_data *response;
+
+            status = pjsip_inv_initial_answer(inv, rdata, 100, NULL, NULL,
+                                              &response);
+            if (status != PJ_SUCCESS) {
+                if (response)
+                    pjsip_tx_data_dec_ref(response);
+                PJ_PERROR(3, (THIS_FILE, status,
+                              "Failed to create initial answer"));
+                return status;
```
sauwming, line 6223: **12. The application isn't told that the async answer failed.**

Returning non-`PJ_SUCCESS` here tells sip_inv "the application will not answer", so it generates its own 200/OK. But the application still believes it owns the answer, and its later `pjsua_call_answer()` finds the transaction already completed and fails with `PJ_EINVALIDOP` and no indication why.

Worth clearing the async intent or notifying the application on this path. Same for the `pjsip_inv_send_msg()` failure below. Not blocking.

pjsip/src/pjsua2/endpoint.cpp
oleopold: Code was changed to guarantee to send ACK

pjsip/src/pjsua2/endpoint.cpp
oleopold: fixed

pjsip/src/pjsua-lib/pjsua_call.c
oleopold: fixed

pjsip/include/pjsua2/call.hpp
oleopold: Default implementation does not throw.

pjsip/src/pjsua-lib/pjsua_call.c
oleopold: Code was changed to respect code from callback.

pjsip/include/pjsua-lib/pjsua.h
oleopold: Callback moved to the end of structure.

pjsip/src/pjsip-ua/sip_inv.c
oleopold: Removed

pjsip/src/pjsua-lib/pjsua_call.c
oleopold: Removed

pjsip/src/pjsip-ua/sip_inv.c
oleopold: Removed dead code

pjsip/src/pjsua-lib/pjsua_call.c
oleopold: Added missing guards

nanangizz:
CI failure is legit, it was C90 compatibility check which is only enabled in that CI job, here is the failure:
```
../src/pjsua-lib/pjsua_call.c: In function ‘pjsua_call_on_send_ack’:
../src/pjsua-lib/pjsua_call.c:7540:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
 7540 |     pj_bool_t skip_sending_ack = PJ_FALSE;
      |     ^~~~~~~~~
```

pjsip/src/pjsip-ua/sip_inv.c
oleopold: Instead sending stale sdp offer PJ_EINVALIDOP is returned when on_create_offer() will not return sdp.

pjsip/src/pjsua-lib/pjsua_call.c
oleopold: When fails sending 100 in async ACK application is not notified, but transaction is terminated and framework will not send 200 automaticaly.

Issue #1976, opened 2016-10-31
Support to generate a synthesized IPv6 address from IPv4 address


pjsipbot:
_2016-10-31 01:27:15: @mingteluu_ _created the issue on trac ticket [1976](https://trac.pjsip.org/repos/ticket/1976)_

Add support to generate a synthesized IPv6 address from IPv4 address. The ability to synthesize IPv6 addresses was added to getaddrinfo in iOS 9.2 and OS X 10.11.2 (reference: [Apple's doc](https://developer.apple.com/library/content/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/UnderstandingandPreparingfortheIPv6Transition/UnderstandingandPreparingfortheIPv6Transition.html#//apple_ref/doc/uid/TP40010220-CH213-DontLinkElementID_4)).

Thanks to Jamil Abdullayev for the patch.

___
_2016-10-31 01:27:38: @mingteluu_ _changed **status** from **new** to **closed**_

___
_2016-10-31 01:27:38: @mingteluu_ _set **resolution** to **fixed**_

___
_2016-10-31 01:27:38: @mingteluu_ _commented_

In [r5476](../commit/f7a742bbbf3b02b26f2efbcfb92ed0069392f7bc):
Fixed #1976: Support to generate a synthesized IPv6 address from IPv4 address

PR #5268, opened 2026-09-21
Let an explicit IP address determine the address family in pjsip_resolve()

### LINKED ISSUE #5267: Call fails with EAFNOSUPPORT after IP address change when target resolves to IPv4 but transport still carries the IPv6 flag
# Issue: TLS call fails with EAFNOSUPPORT after IP address change when target resolves to IPv4 but transport still carries the IPv6 flag

## Describe the bug

When a call is re-established after an IP address change (WLAN ↔ cellular handover,
`pjsua_handle_ip_change()` → re-INVITE), the call can fail with `PJ_EAFNOSUPPORT`
(120047 on Windows / `EAFNOSUPPORT` on POSIX) if the SIP proxy is now reached over
IPv4, while the transport type flag still carries `PJSIP_TRANSPORT_IPV6` from the
previous leg.

The resolver synthesizes an IPv4-mapped IPv6 address (`::ffff:<v4>`) for a concrete
IPv4 target, classifies the transport as IPv6, and the subsequent socket creation
fails with `EAFNOSUPPORT`, destroying the transport and killing the call.

## Environment

- PJSIP 2.17, iOS and Android, TLS transport
- `PJ_HAS_IPV6=1`
- Dual-stack network: device previously connected over IPv6, SIP proxy resolves
  to an IPv4 address after the handover

## Root cause

`pjsip_resolve()` in `pjsip/src/pjsip/sip_resolve.c` selects the address family
with:

```c
if ((ip_addr_ver == 6) || (type & PJSIP_TRANSPORT_IPV6))
    af = pj_AF_INET6();
else if (ip_addr_ver == 4)
    af = pj_AF_INET();
```

`get_ip_addr_ver()` has already unambiguously determined the IP version of the
target (4 or 6) — but a stale `PJSIP_TRANSPORT_IPV6` bit in `type`, inherited from
the previous transport leg, takes precedence over the concrete IPv4 address. The
`else if (ip_addr_ver == 4)` branch is never reached.

Observed in a production call with 6 WLAN handovers: the final re-INVITE fails
with 120047 → 503 → media destroyed, every time the previous leg was IPv6 and the
resolver returned an IPv4 address.

## Expected behavior

An explicit IP address determines its own address family. The transport-type
IPv6 flag should only be used as a hint for hostname targets that have not been
resolved to an IP address yet.

## Logs/Screenshots

Excerpt from a device log (production call, WLAN handover, re-INVITE cseq=23118):

```
... TLS transport construction: target resolved to IPv4 <proxy-v4-addr>
... pjsip_resolve: address family selected: IPv6 (transport flag)
... socket(): 120047 (PJ_EAFNOSUPPORT)
... 503 Service Unavailable, media transport destroyed
```

## Suggested fix

Reorder the condition so the IP version wins:

```c
if (ip_addr_ver == 6)
    af = pj_AF_INET6();
else if (ip_addr_ver == 4)
    af = pj_AF_INET();
else if (type & PJSIP_TRANSPORT_IPV6)
    af = pj_AF_INET6();
```

We have this fix running in production and will submit a PR referencing this
issue.

TheOneDenis:
## Description

`pjsip_resolve()` in `pjsip/src/pjsip/sip_resolve.c` selects the address
family with:

```c
if ((ip_addr_ver == 6) || (type & PJSIP_TRANSPORT_IPV6))
    af = pj_AF_INET6();
else if (ip_addr_ver == 4)
    af = pj_AF_INET();
```

`get_ip_addr_ver()` unambiguously determines the IP version when the target
is a literal IP address, but a stale `PJSIP_TRANSPORT_IPV6` bit in `type` —
inherited from the previous transport leg — overrides a concrete IPv4 target:

```c
if (ip_addr_ver == 6)
    af = pj_AF_INET6();
else if (ip_addr_ver == 4)
    af = pj_AF_INET();
else if (type & PJSIP_TRANSPORT_IPV6)
    af = pj_AF_INET6();
```

Now the IP version (when known) always wins, and the transport-type IPv6 flag
is only a hint for hostname targets that have not been resolved yet.

## Motivation and Context

When a re-INVITE is triggered by an IP address change (WLAN ↔ cellular
handover, `pjsua_handle_ip_change()`), the transport type flag may still carry
`PJSIP_TRANSPORT_IPV6` from the previous leg while the new target resolves to
an IPv4 address. The resolver then synthesizes an IPv4-mapped IPv6 address
(`::ffff:<v4>`) and classifies the transport as IPv6. The subsequent socket
creation fails with `EAFNOSUPPORT` (120047), destroying the transport and
killing the call.

Observed in a production call with 6 WLAN handovers: the final re-INVITE
(cseq=23118) failed with 120047 → 503 → media destroyed.

Fixes #5267.

## How Has This Been Tested?

- Full build: `./configure && make -j3`, zero new warnings (existing
  warnings in unrelated files unchanged).
- `pjsip-test`: `resolve_test` passed (15s, full resolver test suite).
- `pjlib-test`: passed.
- Production (2.17 with this patch, iOS): calls survive WLAN handovers
  where the previous leg was IPv6 and the proxy resolves to IPv4; the
  EAFNOSUPPORT failure no longer occurs.

## Types of changes

## Checklist

- [x] My code follows the CODING STYLE of this project (C89, minimal comments).
- [ ] I have added tests to cover my changes. *(requires a transport with a
      stale IPv6 flag resolving to a literal IPv4 address; behavior verified
      via resolver unit suite and production evidence)*


nanangizz:
Thanks for the report and the PR.

Unfortunately this breaks NAT64. When the transport is IPv6 and the target is an IPv4 literal, `af = AF_INET6` is intentional (#1976/#2198). It makes the resolver call `pj_getaddrinfo(AF_INET6, <v4>)` to get the NAT64-synthesized address on IPv6-only networks, which iOS requires. With this change, that block becomes dead code. An example of such a case is an account with `ipv6_sip_use = USE_IPV6_ONLY`, where `sip_util.c` forces IPv6 for in-dialog requests.

The real bug seems to be in the synthesis check. On a network without NAT64, Darwin's `getaddrinfo(AF_INET6, <v4 literal>)` returns a v4-mapped `::ffff:a.b.c.d` instead of failing. The check only looks at `sa_family == AF_INET6`, so it accepts that address, and the connect then fails on the `IPV6_V6ONLY` socket. Errno 47 is `EAFNOSUPPORT` on Darwin.

Suggested fix: reject v4-mapped results so the existing IPv4 fallback is used:

```c
static pj_bool_t is_ipv4_mapped(const pj_in6_addr *a)
{
    static const pj_uint8_t prefix[12] =
        {0,0,0,0,0,0,0,0,0,0,0xff,0xff};
    return pj_memcmp(a->s6_addr, prefix, sizeof(prefix)) == 0;
}
...
if (status2 == PJ_SUCCESS && count > 0 &&
    ai[0].ai_addr.addr.sa_family == pj_AF_INET6() &&
    !is_ipv4_mapped(&ai[0].ai_addr.ipv6.sin6_addr))
```

The type then drops the IPv6 flag and the call continues over IPv4, while NAT64 still works.

Could you please try this in your setup? Also, could you share your `ipv6_sip_use` / NAT64 account settings and a level-5 log of the failing re-INVITE?

TheOneDenis:
Thanks for the detailed analysis! The NAT64 synthesis path explains why the OR condition exists — we clearly missed that. Rejecting the v4-mapped result in the synthesis check looks like the right fix indeed. We'll try your suggestion in our setup and rework the PR.

Our account settings are ipv6_sip_use = PJSUA_IPV6_ENABLED_PREFER_IPV4 with an IPv4-literal proxy, so no NAT64 account config involved. We'll get back to you with the results and the level-5 log.

Issue #2682, opened 2021-03-30
One-Way-Audio (TX only) caused by SRTP ROC reset (media reinit) after IP change


j4nnis:
**Describe the bug**

We are using PJSIP in mobile apps with SRTP and we had a One-Way-Audio issue recently that was hard to debug. 
We were experiencing this issue during quite long (> ~17 min) calls. We also saw a reinvite that was sent from the iOS device. 

After a lot of debugging, we learned that the re-invite must've been caused by an IP change. In this case our app calls 'pjsua_handle_ip_change'. 
Normally (and with shorter calls) this works fine. 
But with longer calls we saw this log from PJSIP: '1412 Failed to unprotect SRTP, pkt size=182' after the reinvite was ACKed (see logs).

Looking into this issue revealed, that decryption fails in this case because PJSIP forgot its SRTP context for the RX media stream. 
While it did reinit with the correct key (found in the SDP in the OK for its reinvite) it did not keep it SRTP rollover counter (ROC). 
Hacking the ROC to the correct value fixed the issue and decryption was once again possible.

After debugging this we found that PJSIPs default setting is to deinit (and later reinit) its media by default when sending a reinvite after an IP change (pjsua_call.c -> pjsua_call_reinvite2() 
Line 3018 -> apply_call_setting() Line 651). 
This discards the RX SRTP context including the current ROC value. 
When the server keeps using the same SRTP stream to send audio to the client, the client will be unable to decrypt it. 

**Expected behavior**

Please correct me if I'm wrong here, but I would expect that the client compares the RX stream to the old stream and only disregards its RX SRTP context if it receives an entirely new stream. 
Otherwise, I would expect it to keep track of its ROC (https://tools.ietf.org/html/rfc3711 3.3.1). 
The only other option would be that it's illegal for the backend side to keep using the same stream in its OK message. 
But I don't believe this is the case. So disregarding the SRTP context when sending the reinvite after an IP change will always be dangerous.

**Logs/Screenshots**

Experiment 23.03.21:
Initial invite:

```
2021-03-23 11:58:08.802101+0100 XXX[6670:3066336] [Call >] 🌘 📚 PJSIP ✳️ LOG #️⃣_0: 11:58:08.802           pjsua_core.c  ....TX 1422 bytes Request msg INVITE/cseq=22002 (tdta0x103ae00a8) 
to TLS 111.111.111.113:5061:
INVITE sip:496@example.com;transport=tls SIP/2.0

[...]

PR #5269, opened 2026-09-21
Fix SRTP ROC retry after transport reinit being a no-op

TheOneDenis:
## Description

After a media reinit following an IP change, the SRTP transport may receive
packets whose sequence numbers have rolled over, so the RX roll over counter
(ROC) carried over from the previous session no longer matches the sender's
actual ROC. PR #2531 added retry logic to `srtp_rtp_cb()` that is supposed to
try the alternative ROC candidate when unprotect fails with
`srtp_err_status_auth_fail`.

However, the retry candidate is selected by comparing against
`srtp_get_stream_roc()`:

```c
srtp_get_stream_roc(srtp->srtp_ctx.srtp_rx_ctx,
                    srtp->setting.rx_roc.ssrc, &roc);
new_roc = (roc == srtp->setting.rx_roc.roc?
           srtp->setting.prev_rx_roc.roc: srtp->setting.rx_roc.roc);
```

`srtp_get_stream_roc()` reports the ROC stored in libsrtp's replay database.
The ROC set via `srtp_set_stream_roc()` (e.g. during `create_srtp_ctx()` when
the carried-over ROC is applied) is kept by libsrtp in a separate *pending*
state and is only folded into the replay database once a packet has been
unprotected successfully.

Since the retry exists precisely for the situation where **no packet
authenticates yet**, the replay database still holds its initial value, the
comparison re-selects the ROC that is already pending, and the alternative
candidate (`prev_rx_roc`) is **never tried**. Every subsequent packet fails
authentication for the rest of the call → one-way audio (RX dead).

This PR tracks the ROC last handed to `srtp_set_stream_roc()` in a new
`rx_roc_pending` field of `transport_srtp`, toggles the retry against that
value instead of the replay-database value, and resets it once a packet
authenticates (mirroring libsrtp folding the pending ROC into the replay
database). The two candidates now actually alternate across retries.

It also adds the RTP sequence number to the "Failed to unprotect SRTP" log
line, which is the key indicator for whether the sender has passed a sequence
number rollover.

## Motivation and Context

Customers lose audio when the network changes mid-call. Measured on a
16-minute production call with three network changes: the two changes at
ROC 0 kept audio working, the one change at ROC 1 killed RX audio — same
call, same SSRC, same key, ROC being the only differing variable.

Related: #2682 (one-way audio after IP change, resolved by #2531) — this
fixes a bug in the retry logic introduced by #2531 itself.

Fixes the case where the retry never tries the alternative ROC.

## How Has This Been Tested?

- Full build: `./configure && make -j3`, zero new warnings (existing
  warnings in unrelated files unchanged).
- `pjmedia-test`: 7/7 passed.
- Production measurement (2.17 with this patch, iOS and Android): a
  16-minute call with three network changes; with the patch, the ROC-1
  handover that previously killed RX audio now recovers on the retry with
  the alternative ROC.

## Types of changes

## Checklist

- [x] My code follows the CODING STYLE of this project (C89, minimal comments).
- [ ] I have added tests to cover my changes. *(the ROC-retry path requires a
      real IP change mid-call and a sender past a sequence rollover; no
      existing unit test covers it, and reproducing it deterministically in
      the test suite is not feasible — see Testing above for production
      evidence)*


nanangizz:
Thank you for your contributions.
Please sign the CLA for this PR as well as the other one you submitted.

nanangizz:
Thanks for the PR, the analysis is correct: `srtp_get_stream_roc()` returns the replay database ROC, not the pending one set by `srtp_set_stream_roc()`, so the retry never alternated. With `rx_roc_pending` the two candidates now alternate as intended, and I don't see a regression in the related paths (SSRC-change restart from #4490, replay restart, remote-address-change case).

A few minor comments:

1. The "Retrying to unprotect SRTP from ROC %d to new ROC %d" log still prints `roc` from `srtp_get_stream_roc()`, which is the misleading value. Please print the previous `rx_roc_pending` instead.
2. The `TEST_ROC` block in `srtp_rtp_cb()` calls `srtp_set_stream_roc()` without updating `rx_roc_pending`. It's debug-only, but it would be good to update it there as well for consistency.
3. In the description: the ROC retry logic was added by #2846 (fixing #2682), not #2531. #2531 only maintains the RX ROC across SRTP restarts.

TheOneDenis:
Thanks for the review and the corrections — good points, especially the #2846 vs #2531 mix-up. We'll address all three comments and update the PR shortly.

